Issue #220, opened 2019-03-27
Error when running training step in tutorial


rpelliard:
Hello,

On this page of the tutorial: https://dvc.org/doc/get-started/pipeline, when running the command 
```
dvc run -d src/train.py -d data/features \
              -o model.pkl \
              python src/train.py -d data/features model.pkl
```

I get the following:

```
Running command:
	python src/train.py -d data/features model.pkl
Arguments error. Usage:
	python train.py features model
```

Looking at `train.py`, it looks like the `-d` causes the mistake and is not necessary.

It also looks like the command is missing a `-f train.dvc` to be consistent with the instructions that follow.

edit: just submitted a PR to fix https://github.com/iterative/dvc.org/pull/221

shcheklein:
Fixed by #221 . Kudos @rpelliard !

PR #221, opened 2019-03-27
Fix training step command in tutorial

rpelliard:
Hello,

This is a proposed fix for the issue https://github.com/iterative/dvc.org/issues/220 i opened
This should fix the error message I get when running the training step command in the tutorial:

```
$ dvc run -d src/train.py -d data/features \
              -o model.pkl \
              python src/train.py -d data/features model.pkl 

Running command:
	python src/train.py -d data/features model.pkl
Arguments error. Usage:
	python train.py features model
Error: failed to run command - stage 'model.pkl.dvc' cmd python src/train.py -d data/features model.pkl failed
```

### DIFF
--- a/static/docs/get-started/pipeline.md
+++ b/static/docs/get-started/pipeline.md
@@ -22,7 +22,7 @@ The third stage, training:
 ```dvc
     $ dvc run -d src/train.py -d data/features \
               -o model.pkl \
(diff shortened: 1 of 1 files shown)


shcheklein:
Thanks @rpelliard ! Good catch!

PR #226, opened 2019-03-28
Fix a mistake in `dvc run --metrics` documentation

### LINKED ISSUE #199: a mistake in `dvc run --metrics` documentation
Hi!

First of all, thank you for developing such a great tool. I'm just starting using it, but it seems to be incredibly useful (to this moment I had massive chaos in my files).

I was reading the documentation (which is great too) when I spotted a mistake I suppose. The description of `-m, --metrics` param of `dvc run` [command](https://dvc.org/doc/commands-reference/run) says near the end:

> [...] Metrics are not cached (put under DVC control), you should see `cache: false` in the stage file. [...]

But from my experiments it doesn't seem to be true. However, it fits `-M, --metrics-no-cache` param. So I think it's some editorial error ;)

Cheers,
Piotr

piojanu:
Fixes #199.

### DIFF
--- a/static/docs/commands-reference/run.md
+++ b/static/docs/commands-reference/run.md
@@ -74,15 +74,14 @@ share them.
 small human readable file (JSON, CSV, text, whatnot) with some numbers or other
 meta-information that describes a model or other outputs. Check `dvc metrics`
 to learn more about tracking metrics and comparing them across different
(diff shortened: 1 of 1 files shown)


shcheklein:
Thanks @piojanu !!

PR #228, opened 2019-03-31
metrics: update metrics.show() documentation

cand126:
Updating the documentation according to this PR: https://github.com/iterative/dvc/pull/1809

### DIFF
--- a/static/docs/commands-reference/metrics:show.md
+++ b/static/docs/commands-reference/metrics:show.md
@@ -32,7 +32,7 @@ command). If `-t` is not specified `-x` is ignored.
 
 * `-t`, `--type` - specify a type of the metric file(s) that will be used to
 determine how to handle `xpath` parameter from down below. Accepted values are:
-`raw`, `json`, `tsv`, `htsv`, `csv`, `hcsv`. If any other value is specified it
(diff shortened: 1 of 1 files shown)


shcheklein:
@cand126 thanks! does it work the same way for `dvc repro -m`? btw? cc @efiop

shcheklein:
@cand126 Also, if we detect type automatically, do we apply `-x` as well or is it ignored it `-t` is not specified explicitly?

efiop:
@shcheklein `-x` overwrites any xpath specified previously. And yes, it works the same way for `dvc repro -m`.

shcheklein:
@efiop what happens if -t is not specified but auto-detected, do we apply -x as well or do we ignore it?

efiop:
@shcheklein yes, we apply `-x`.

shcheklein:
@efiop thanks! we'll need to update the documentation a little bit.

efiop:
@shcheklein Yes, sorry, didn't catch that one right away. Thanks for the heads up!

cand126:
@shcheklein 

- `dvc repro -m` does not detect the file type automatically
- The `--xpath` argument still works with file type detection. For example, `dvc metrics show --xpath=AUC` lists the values of **AUC** column

cand126:
@shcheklein do I also need to update the documentation for `dvc repro -m`?

efiop:
@cand126 We would definitely appreciate it if you'd be so kind :) Issue for that is https://github.com/iterative/dvc.org/issues/229 .

PR #246, opened 2019-04-15
Update example-versioning.md

maggyero:
* Use the same directory layout formatting as in https://dvc.org/doc/get-started/connect-code-and-data.
* Correct the number of validation items (800).
* Add a missing information unit (byte).


shcheklein:
Thanks @maggyero !

Issue #108, opened 2018-11-07
cmd ref: explain remote://myremote notation in `import` & `run`


efiop:
- [x] import/get-url
- [ ] run
- [ ] add

jorgeorpinel:
I don't see this in either cmd ref yet so I unchecked "import" above.

efiop:
Also `dvc add remote://`

jorgeorpinel:
> I unchecked "import" above.

Actually this only applies to import/get-url right? And it's explained in those already. So only `add` and `run` are pending (only mentioned for those in the external data guides ATM).

jorgeorpinel:
> only mentioned for those in the external data guides

After reviewing the add and run refs I see they link to those guides in the appropriate places so I'm not seeing a need to explain `remote://` directly in them. We don't go over all the other URLs either. But lmk if you have a specific idea on how else to address this!

PR #260, opened 2019-04-19
Update dvc install and dvc import documentation

robogeek:
This commit is only the dvc install documentation.  The dvc import documentation is still to come.

### DIFF
--- a/static/docs/commands-reference/checkout.md
+++ b/static/docs/commands-reference/checkout.md
@@ -51,10 +51,15 @@ The output of `dvc checkout` does not list which data files were restored. It
 does report removed files and files that DVC was unable to restore due to it
 missing from the cache.
 
+This command will fail to checkout files that are missing from the cache. In
+such a case, `dvc checkout` prints a warning message. Any files that can be
+checked out without error will be restored.
+
 There are two methods to restore a file missing from the cache, depending on the
 situation. In some cases the pipeline must be rerun using the `dvc repro`
 command. In other cases the cache can be pulled from a remote cache using the
-`dvc pull` command.
+`dvc pull` command.  It may be necessary to use `--all-tags` or `--all-branches`
+with the `dvc pull` command.
 
 ## Options
 
--- a/static/docs/commands-reference/checkout.md
+++ b/static/docs/commands-reference/checkout.md
@@ -58,8 +58,7 @@ checked out without error will be restored.
 There are two methods to restore a file missing from the cache, depending on the
 situation. In some cases the pipeline must be rerun using the `dvc repro`
 command. In other cases the cache can be pulled from a remote cache using the
-`dvc pull` command.  It may be necessary to use `--all-tags` or `--all-branches`
-with the `dvc pull` command.
+`dvc pull` command.
 
 ## Options
 
--- a/static/docs/commands-reference/commit.md
+++ b/static/docs/commands-reference/commit.md
@@ -121,7 +121,7 @@ Now, we can install requirements for the project:
 Then download the precomputed data using:
 
 ```dvc
-    $ dvc pull
+    $ dvc pull --all-branches --all-tags
 ```
 
 This data will be retrieved from a preconfigured remote cache.
--- a/static/docs/commands-reference/import.md
+++ b/static/docs/commands-reference/import.md
@@ -2,15 +2,6 @@
 
 Import file from URL to local directory and track changes in remote file.
 
-Supported schemes:
-
-* `local` - Local path
-* `s3` - URL to a file on Amazon S3
-* `gs` - URL to a file on Google Storage
-* `ssh` - URL to a file on another machine with SSH access
-* `hdfs` - URL to a file on HDFS
-* `http` - URL to a file with a _strong ETag_ served with HTTP or HTTPS
-
 ## Synopsis
 
 ```usage
@@ -21,7 +12,64 @@ Supported schemes:
       out            Output
 ```
 
+## Description
+
+DVC supports `.dvc` files which refer to an external data file, see
+[External Dependencies](/doc/user-guide/external-dependencies). In such a DVC
+file, the `deps` section will list the remote file specification and the `outs`
+section will list the local file name in the workspace. It records enough data
+from the remote file to enable DVC to efficiently check the remote file to
+determine if the local file is out of date.
+
+The `dvc import` command helps the user create such an external data dependency.
+
+It generates a DVC file listing in the `deps` section the resource named in
+the `url` parameter, and in the `outs` section the file named in the `out`
+parameter.
+
+DVC supports several types of remote files:
+
+Type | Discussion | URL format
+-----|------------|------------
+`local` | Local path | `/path/to/local/file`
+`s3` | Amazon S3 | `s3://mybucket/data.csv`
+`gs` | Google Storage | `gs://mybucket/data.csv`
+`ssh` | SSH server | `ssh://user@example.com:/path/to/data.csv`
+`hdfs` | HDFS | `hdfs://user@example.com/path/to/data.csv`
+`http` | HTTP to file with _strong ETag_ | `https://example.com/path/to/data.csv`
+
+In the _External Dependencies_ documentation, an alternative is demonstrated for
+each of these schemes.  Instead of:
+
+```dvc
+    $ dvc import https://example.com/path/to/data.csv data.csv
+```
+
+It is possible to instead run
+
+```dvc
+    $ dvc run -d https://example.com/path/to/data.csv \
+              -o data.csv \
+              wget https://example.com/path/to/data.csv -O data.csv
+```
+
+Both methods generate a DVC file with an external dependency. The `dvc import`
+command saves the user from using the command to copy files from each of the
+remote storage schemes, and from having to install CLI tools for each service.
+
+When DVC inspects a DVC file, one step is inspecting the dependencies to see if
+any have changed. A changed dependency will appear in the `dvc status` report,
+indicating the need to re-run the corresponding part of the pipeline. When DVC
+inspects an external dependency, it uses a method appropriate to that dependency
+to test its current status.
+
 ## Options
+
+* `--resume` - resume previously started download.
+
+* `-f`, `--file` - specify name of the DVC file it generates. It should be 
+  either `Dvcfile` or have a `.dvc` suffix (e.g. `data.dvc`) in order 
+  for `dvc` to be able to find it later. 
 
 * `-h`, `--help` - prints the usage/help message, and exit.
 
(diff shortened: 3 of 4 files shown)


static/docs/commands-reference/commit.md
```diff
@@ -121,7 +121,7 @@ Now, we can install requirements for the project:
 Then download the precomputed data using:
 
 ```dvc
-    $ dvc pull
+    $ dvc pull --all-branches --all-tags
```
shcheklein: 👍

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
 # install
 
-Install dvc hooks into the repository
+Install DVC hooks into the Git repository to automate certain common actions
```
shcheklein: `actions` -> `actions.` ?

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+## Description
+
+As designed DVC combines an intelligent data repository with using a regular SCM
+like Git to store code and configuration files. With `dvc install` the two are
+more tightly integrated, causing certain actions to happen automatically. This
+saves the DVC user from having to remember to perform both DVC and SCM actions.
+
+There are two modes of tracking files in a DVC workspace:
+
+* Data files are managed by DVC in the DVC cache
```
shcheklein: cache. ?

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+* Code, configuration and DVC files are managed by an SCM like Git
```
shcheklein: Git. ?

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+another, the SCM retrieves the corresponding DVC files. By default that leaves
+the workspace in a state where the DVC files refer to data files other than what
+is currently in the workspace. The user at this point should run `dvc checkout`
+so that the data files will match the current DVC files.
+
+**Commit** When committing a change to the SCM repository, that change possibly
+requires rerunning the pipeline to reproduce the workspace results. It is
+helpful to know automatically this as a reminder to run `dvc repro`.
+
+## Installed SCM hooks
```
shcheklein: I would start descrption with this. For most ppl this is the most important information. After that we can do some explanation.

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+
+* Git `pre-commit` hook executes `dvc status` to inform the user about the
+  workspace status.
+* Git `post-checkout` hook executes `dvc checkout` to automatically synchronize
```
shcheklein: May be it makes sense to mention that it happens after "git checkout". Also worth mentioning (here or separately) what happens if `dvc checkout` fails - is `git checkout` rejected?

static/docs/commands-reference/install.md
```diff
@@ -23,13 +50,204 @@ Install dvc hooks into the repository
...
+    $ git checkout master
+    Previous HEAD position was d13ba9a add featurization stage
+    Switched to branch 'master'
+    Your branch is up to date with 'origin/master'.
+
+    $ dvc checkout
+    [##############################] 100% Checkout finished!
+```
+
+Start by resetting the workspace to he at the head commit.
```
shcheklein: Should we move this before the code? I was misled a bit and was expecting `dvc install` to appear before.

static/docs/commands-reference/install.md
```diff
@@ -23,13 +50,204 @@ Install dvc hooks into the repository
...
+     M evaluate.dvc
+     M featurize.dvc
+     M src/featurization.py
+     M train.dvc
+
+    $ git commit -a -m 'updated data after modified featurization'
+    Pipeline is up to date. Nothing to reproduce.
+    [master 78d0c44] modified featurization
+     5 files changed, 12 insertions(+), 12 deletions(-)
+
```
shcheklein: extra space

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+corresponding to that branch or tag. The DVC files in turn refer to data files
+in the DVC cache by checksum. Hence, switching from one SCM branch or tag to
+another, the SCM retrieves the corresponding DVC files. By default that leaves
+the workspace in a state where the DVC files refer to data files other than what
+is currently in the workspace. The user at this point should run `dvc checkout`
+so that the data files will match the current DVC files.
+
+**Commit** When committing a change to the SCM repository, that change possibly
+requires rerunning the pipeline to reproduce the workspace results. It is
+helpful to know automatically this as a reminder to run `dvc repro`.
```
shcheklein: Or it might be useful to see that there are some files are not in the cache yet (and you need to run `dvc commit`)?

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
-- pre-commit : dvc status
-- post-checkout : dvc checkout
+## Description
+
+As designed DVC combines an intelligent data repository with using a regular SCM
+like Git to store code and configuration files. With `dvc install` the two are
+more tightly integrated, causing certain actions to happen automatically. This
+saves the DVC user from having to remember to perform both DVC and SCM actions.
+
+There are two modes of tracking files in a DVC workspace:
```
shcheklein: I'm not sure I understand these two points and their connection to everything else to be honest.

static/docs/commands-reference/install.md
```diff
@@ -23,13 +50,204 @@ Install dvc hooks into the repository
...
+     M train.dvc
+
+    $ git commit -a -m 'updated data after modified featurization'
+    Pipeline is up to date. Nothing to reproduce.
+    [master 78d0c44] modified featurization
+     5 files changed, 12 insertions(+), 12 deletions(-)
+
+```
+
+After rerunning the DVC pipeline, of course the data files are in sync with
```
shcheklein: copy and hight the phrase `Pipeline is up to date. Nothing to reproduce.` once again here? So that user can easily find in the code snippet what is happening.

static/docs/commands-reference/install.md
```diff
@@ -1,16 +1,43 @@
...
+is currently in the workspace. The user at this point should run `dvc checkout`
+so that the data files will match the current DVC files.
+
+**Commit** When committing a change to the SCM repository, that change possibly
+requires rerunning the pipeline to reproduce the workspace results. It is
+helpful to know automatically this as a reminder to run `dvc repro`.
+
+## Installed SCM hooks
+
+* Git `pre-commit` hook executes `dvc status` to inform the user about the
```
shcheklein: does status hook rejects `git commit` if there are some uncommitted DVC files or repro needs to be run?

static/docs/commands-reference/install.md
robogeek: Are you meaning to move the list of installed hooks to the top?

To me that list seemed like a detail, and it was more important to have the high level overview first.

static/docs/commands-reference/install.md
robogeek: If you mean the space on the beginning of the line, that is in the Git output

static/docs/commands-reference/install.md
robogeek: Did some rewording.

static/docs/commands-reference/install.md
robogeek: I've tried a couple ways to induce an error in the Git hook.  Git doesn't seem to notice the error status.

Your question leaves me wondering what happens to the workspace if `dvc checkout` fails.  Does it leave the workspace in a halfway-updated state?  I put a question in Discord.

static/docs/commands-reference/install.md
robogeek: Okay

static/docs/commands-reference/install.md
robogeek: Reworded

static/docs/commands-reference/install.md
robogeek: I think you mean to insert some blank lines making it easier to find the text from `dvc status`?

static/docs/commands-reference/install.md
robogeek: No it doesn't

static/docs/commands-reference/install.md
robogeek: I got the answer that the workspace will be left in an intermediate state - the files which can be checked out will be, others won't.  

It didn't seem the update needed to be in the install documentation but in the checkout documentation.  So I've updated there.

static/docs/commands-reference/install.md
robogeek: That is - I ran a test modifying a file, and ran `git commit`.  The pre-commit hook simply gave `dvc status` and then I was able to commit the file to Git.

Hence there's nothing more to say in the documentation.

static/docs/commands-reference/checkout.md
```diff
@@ -51,10 +51,15 @@ The output of `dvc checkout` does not list which data files were restored. It
...
 
+This command will fail to checkout files that are missing from the cache. In
+such a case, `dvc checkout` prints a warning message. Any files that can be
+checked out without error will be restored.
+
 There are two methods to restore a file missing from the cache, depending on the
 situation. In some cases the pipeline must be rerun using the `dvc repro`
 command. In other cases the cache can be pulled from a remote cache using the
-`dvc pull` command.
+`dvc pull` command.  It may be necessary to use `--all-tags` or `--all-branches`
```
shcheklein: let's remove it, don't see any reason to mention it here. `checkout` and `pull` are made `lazy` for a very good reason. We provide and describe available options to pull missing data. We don't have auto-pull option on checkout yet - but it's another problem and it should be described separately. I think we explicitly specify in `pull`, `fetch`, `status`, `checkout` docs that we analyze only current set of files. If it's not clear enough let's emphasize that better - for example, mention that unlike git DVC does not pull everything at once ...

static/docs/commands-reference/install.md
shcheklein: Yes, I think it's good for a professional to see immediately a low level explanation of what's happening. It should be short enough. Then you provide some motivation. For a lot of users, especially those who are using DVC already it would save a lot of time.

static/docs/commands-reference/install.md
shcheklein: no, I mean an extra new line - 254

static/docs/commands-reference/install.md
shcheklein: that one is good, but I would also mention the phrase in the paragraph itself

static/docs/commands-reference/import.md
```diff
@@ -2,15 +2,6 @@
 
 Import file from URL to local directory and track changes in remote file.
```
shcheklein: local workspace or working space? URL can pattern match with HTTP only. Can we specify somehow that it can be any remote resource - s3, gs, etc?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+
+When DVC inspects a DVC file, one step is inspecting the dependencies to see if
+any have changed. A changed dependency will appear in the `dvc status` report,
+indicating the need to re-run the corresponding part of the pipeline. When DVC
+inspects an external dependency, it uses a method appropriate to that dependency
+to test its current status.
+
 ## Options
 
+* `--resume` - resume previously started download.
```
shcheklein: would be great to give some explanation for this option ... mention that it usefull if network is unstable? or something like that .. other case - can we get them from dvc's github, when the option was initially added? or ask the team?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
       out            Output
 ```
 
+## Description
+
+DVC supports `.dvc` files which refer to an external data file, see
```
shcheklein: Can we add a section before explaining some motivation? Something like - it's handy command to copy the remote file into your workspace, save it in cache to version control and update automatically whenever the remote file changes

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+[External Dependencies](/doc/user-guide/external-dependencies). In such a DVC
+file, the `deps` section will list the remote file specification and the `outs`
+section will list the local file name in the workspace. It records enough data
+from the remote file to enable DVC to efficiently check the remote file to
+determine if the local file is out of date.
+
+The `dvc import` command helps the user create such an external data dependency.
```
shcheklein: it's not only that, it's not clear so far that it brings file to the working space, adds it to the cache, etc

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+
+It generates a DVC file listing in the `deps` section the resource named in
```
shcheklein: It feels like we already mentioned this before, no?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+DVC supports several types of remote files:
+
+Type | Discussion | URL format
+-----|------------|------------
+`local` | Local path | `/path/to/local/file`
+`s3` | Amazon S3 | `s3://mybucket/data.csv`
+`gs` | Google Storage | `gs://mybucket/data.csv`
+`ssh` | SSH server | `ssh://user@example.com:/path/to/data.csv`
+`hdfs` | HDFS | `hdfs://user@example.com/path/to/data.csv`
+`http` | HTTP to file with _strong ETag_ | `https://example.com/path/to/data.csv`
```
shcheklein: there is one more actually - `remote://`. See this ticket - https://github.com/iterative/dvc.org/issues/108. It would be great to add it here and propagate the explanation to the external dependencies section, and dvc run if necessary.

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+
+In the _External Dependencies_ documentation, an alternative is demonstrated for
```
shcheklein: `Another way to understand this command is ... `?

shcheklein: why don't we put a link instead of _italic_?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+inspects an external dependency, it uses a method appropriate to that dependency
+to test its current status.
+
 ## Options
 
+* `--resume` - resume previously started download.
+
+* `-f`, `--file` - specify name of the DVC file it generates. It should be 
+  either `Dvcfile` or have a `.dvc` suffix (e.g. `data.dvc`) in order 
+  for `dvc` to be able to find it later.
```
shcheklein: there are trailing whitespaces, this paragraph can be rewrapped as well

shcheklein: some other paragraphs in this doc could be rewrapped also

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
 * `-v`, `--verbose` - displays detailed tracing information.
 
-* `--resume` - resume previously started download.
+## Example: Initializing a workspace using a remote file
 
-* `-f`, `--file` - specify name of the DVC file it generates. It should be 
-  either `Dvcfile` or have a `.dvc` suffix (e.g. `data.dvc`) in order 
-  for `dvc` to be able to find it later. 
+The [DVC getting started tutorial](/doc/get-started) demonstrates a simple DVC
+pipeline.  In the [Add Files step](/doc/get-started/add-files) we are told to
```
shcheklein: double space

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+        git add data.xml.dvc prepare.dvc
+    
+    $ git add .
+    $ git commit -a -m 'updated data'
+    [master a8d4ce8] updated data
+     2 files changed, 6 insertions(+), 6 deletions(-)
+
+    $ dvc status
+    Pipeline is up to date. Nothing to reproduce.
+
```
shcheklein: extra new line

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
 * `-h`, `--help` - prints the usage/help message, and exit.
 
 * `-q`, `--quiet` - does not write anything to standard output. Exit with 0 if
   no problems arise, otherwise 1.
 
 * `-v`, `--verbose` - displays detailed tracing information.
 
-* `--resume` - resume previously started download.
+## Example: Initializing a workspace using a remote file
```
shcheklein: name is too specific I think. Let's just simplify it to `Tracking a remote file`? And second example - `Detecting remote changes`?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
 
-* `-f`, `--file` - specify name of the DVC file it generates. It should be 
-  either `Dvcfile` or have a `.dvc` suffix (e.g. `data.dvc`) in order 
-  for `dvc` to be able to find it later. 
+The [DVC getting started tutorial](/doc/get-started) demonstrates a simple DVC
+pipeline.  In the [Add Files step](/doc/get-started/add-files) we are told to
+download a file, then use `dvc add` to integrate it with the workspace.
+
+An alternate way to initialize the _Getting Started_ workspace, using
```
shcheklein: it's not exactly alternative, we should specify that it will be tracking changes ... I would say it's an advanced alternative :)

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+`dvc import`, is
```
shcheklein: `is:` ?

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+
+```dvc
+    $ mkdir get-started
+    $ cd get-started
+    $ git init
+    $ dvc init
+    $ mkdir data
+    $ dvc import https://dvc.org/s3/get-started/data.xml data/data.xml
+    Importing 'https://dvc.org/s3/get-started/data.xml' -> '/Volumes/Extra/dvc/get-started/data/data.xml'
```
shcheklein: can we simplify the local path? just for the sake of example and readability - don't show the full local path

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+    [##############################] 100% data.xml
+    Adding 'data/data.xml' to 'data/.gitignore'.
```
shcheklein: btw, it's worth mentioning in the description that it changes `.gitignore` if Git is used

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+    Saving 'data/data.xml' to cache '.dvc/cache'.
+    Saving information to 'data.xml.dvc'.
+
+    To track the changes with git run:
+
+        git add data/.gitignore data.xml.dvc
+```
+
+If you wish, it's possible to set up the other stages from the _Getting Started_
```
shcheklein: let's use links instead of _italic_ everywhere? It's not very usual way to see it this way. We don't use it in other parts of the docs.

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,278 @@ Supported schemes:
...
+```
+
+The `etag` field in the DVC file contains the ETAG recorded from the HTTP
+request. If the remote file changes, the ETAG changes, letting DVC know when
+the file has changed. 
+
+## Example: Remote file that is updated
 
-## Examples
+What if that remote file is one which will be updated regularly? The project
```
shcheklein: Thes idea is great! But let's simplify it - let's use the local "remote" path as an example instead of SSH. The same `data.xml`, and we can skip some outputs in this case then. It looks too big and hard to reproduce now.

static/docs/commands-reference/import.md
robogeek: It's enough to say _This is useful if the   connection to the remote resource is unstable._

static/docs/commands-reference/import.md
robogeek: The remote URL needs to be documented in the `dvc remote` command documentation.  It should then be enough to reference that documentation from here.

static/docs/commands-reference/checkout.md
```diff
@@ -51,6 +51,10 @@ The output of `dvc checkout` does not list which data files were restored. It
 does report removed files and files that DVC was unable to restore due to it
 missing from the cache.
 
+This command will fail to checkout files that are missing from the cache. In
+such a case, `dvc checkout` prints a warning message. Any files that can be
+checked out without error will be restored.
```
shcheklein: `any files that are found in cache` instead of an error. It's not an error usually - it's a warning like you mentioned above.

static/docs/commands-reference/import.md
```diff
@@ -1,15 +1,6 @@
...
-
-Supported schemes:
-
-* `local` - Local path
-* `s3` - URL to a file on Amazon S3
-* `gs` - URL to a file on Google Storage
-* `ssh` - URL to a file on another machine with SSH access
-* `hdfs` - URL to a file on HDFS
-* `http` - URL to a file with a _strong ETag_ served with HTTP or HTTPS
+Import file from any supported URL or local directory to local workspace and track changes in remote file.
```
shcheklein: Probably not 80 symbols here

static/docs/commands-reference/import.md
shcheklein: I would still explain it briefly here - just give an example of the transformation.

static/docs/commands-reference/import.md
```diff
@@ -21,28 +12,288 @@ Supported schemes:
...
+DVC supports several types of remote files:
+
+Type | Discussion | URL format
+-----|------------|------------
+`local` | Local path | `/path/to/local/file`
+`s3` | Amazon S3 | `s3://mybucket/data.csv`
+`gs` | Google Storage | `gs://mybucket/data.csv`
+`ssh` | SSH server | `ssh://user@example.com:/path/to/data.csv`
+`hdfs` | HDFS | `hdfs://user@example.com/path/to/data.csv`
+`http` | HTTP to file with _strong ETag_ | `https://example.com/path/to/data.csv`
```
shcheklein: let's make strong ETag a link. It would be great to mention that it's required to tracked changes.

static/docs/commands-reference/import.md
shcheklein: any changes to this?

Issue #273, opened 2019-04-25
left sidebar: open a section that has a third level hides other sections


shcheklein:
Open `https://dvc.org/doc/commands-reference`

Check the last section in the list - `unprotect` or something.

Click `remote` and see how `unprotect` disappears/\.

shcheklein:
@ProjGoodF please, let's create a PR - it will kick off the new demo environment on Heroku (kudos to @jorgeorpinel ) and we will be able to test it.

Issue #266, opened 2019-04-24
main page title is not displayed properly in Google


shcheklein:
<img width="747" alt="Screen Shot 2019-04-24 at 12 57 40 AM" src="https://user-images.githubusercontent.com/3659196/56642328-0c273100-662c-11e9-90ca-5251deb33967.png">

Should be the one specified here https://github.com/iterative/dvc.org/blob/master/pages/_document.js#L7

Also, this might be a relevant information:

```
Warning: <title> should not be used in _document.js's <Head>. https://err.sh/next.js/no-document-title
Warning: setState(...): Can only update a mounting component. This usually means you called setState() outside componentWillMount() on the server. This is a no-op.
```

shcheklein:
Reopening this for now until we make sure this is fixed.

PR #274, opened 2019-04-26
Gf fixes

ProjGoodF:
1. Edit package.json, script "start" did not work
2. GitHub Issue #266  _document.js file edited according to documentation next.js (https://github.com/zeit/next.js/blob/master/errors/no-document-title.md). At the moment title put only in pages/index.js. We need to use this revision and wait for the repeated site index linking from Google.
3. GitHub Issue #273  The height of the each section sets strongly in pixels. While opening the root section the height was figured according to a formula (quantity of child elements * 31 px), herewith the quantity of elements of the active child section wasn’t considered. Now it’s corrected.


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "node server.js",
     "build": "next build",
-    "start": "NODE_ENV=production node server.js",
```
shcheklein: does Heroku set it by default? it affects certain parts of the code we haven't had any problems running it locally before - I think "dev" key is made for that?

shcheklein:
@ProjGoodF couple minor comments, otherwise looks great! I think if you put Fixed #273 it will close the ticket automatically when PR is merged.

package.json
ProjGoodF: When trying to start up the node wrote that "NODE_ENV" isn't an internal or external command, an executable program or a batch file.

ProjGoodF:
> What is the thing about 4 empty files at the very end of you open the "Files changed" tab?

IDE PHP Storm recorded some changes in these files, so they were included in the commit, I did not create them. Perhaps this is due to the installation of dependencies from package.json

package.json
ProjGoodF: > does Heroku set it by default? it affects certain parts of the code we haven't had any problems running it locally before - I think "dev" key is made for that?

When trying to start up the node wrote that "NODE_ENV" isn't an internal or external command, an executable program or a batch file.

shcheklein: My point is that this env variable is set there for a reason in the production environment. We need to make sure that we are not breaking the production environment behavior with this change (Heroku is where it's being deployed to). 

Are you running this on Windows? Would `npm run-script dev` be enough?

PR #283, opened 2019-05-01
Fixes #213.

### LINKED ISSUE #213: exclude $ when copying
See https://discordapp.com/channels/485586884165107732/485586884165107734/558281550270038037 .

Is this even possible with our md files?

ProjGoodF:
Fixes #213. For all "$" characters has been added custom rules. The rule will be processed if the code fragment has the alias 'dvc'.

### DIFF
--- a/src/Documentation/Markdown/lang/dvc.js
+++ b/src/Documentation/Markdown/lang/dvc.js
@@ -34,7 +34,7 @@ let _javascript = function(hljs) {
     aliases: ['dvc'],
     contains: [
       {
-        begin: /    \$/,
+        begin: /\s*\$/,
         className: 'skipped'
       },
       {
@@ -67,8 +67,8 @@ let _javascript = function(hljs) {
       VAR,
       hljs.HASH_COMMENT_MODE,
       {
-        begin: /^\s*[^\s#$]/,
-        end: /\n|\Z/,
+        begin: /#/,
+        end: /\n/,
         className: 'meta',
       }
     ],
--- a/src/Documentation/Markdown/lang/dvc.js
+++ b/src/Documentation/Markdown/lang/dvc.js
@@ -38,6 +38,11 @@ let _javascript = function(hljs) {
         className: 'skipped',
       },
       {
+        begin: /^\s\s\s\s(?!(\$|\s))/,
+        end: /([^\\]\n)|\Z|\n/,
+        className: 'meta',
+      },
+      {
         begin: /\s/,
         end: /([^\\]\n)|\Z|\n/,
         contains: [
@@ -62,19 +67,13 @@ let _javascript = function(hljs) {
                 'help dvc init add import checkout run pull push fetch status repro remove move gc config remote metrics install root lock unlock pipeline destroy unprotect commit cache pkg tag diff',
             }
           },
-
           QUOTE_STRING,
           APOS_STRING,
           VAR,
           hljs.HASH_COMMENT_MODE
         ]
       },
-      hljs.HASH_COMMENT_MODE,
-      {
-        begin: /^\s*[^\s#$]/,
-        end: /\n|\Z/,
-        className: 'meta',
-      }
+      hljs.HASH_COMMENT_MODE
     ],
 
   };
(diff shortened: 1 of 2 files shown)


shcheklein:
does not support multiple line expressions now:

Screen Shot 2019-05-01 at 11 26 21 AM

src/Documentation/Markdown/lang/dvc.js
```diff
@@ -34,37 +34,37 @@ let _javascript = function(hljs) {
     aliases: ['dvc'],
     contains: [
       {
-        begin: /^\s*\$/,
-        end: /([^\\]\n)|\Z/,
+        begin: /    \$/,
```
shcheklein: let's make it more generic than just 4 spaces

src/Documentation/Markdown/lang/dvc.js
```diff
@@ -34,37 +34,37 @@ let _javascript = function(hljs) {
...
-            },
-          },
-          QUOTE_STRING,
-          APOS_STRING,
-          VAR,
-          hljs.HASH_COMMENT_MODE
-        ]
       },
+      {
+        begin: / dvc [a-z\-]+/,
```
shcheklein: most likely something like this:

```
$ dvc init
 dvc init
 dvc init
```

will highlight `dvc init` three times.

src/Documentation/Markdown/lang/dvc.js
```diff
@@ -34,37 +34,37 @@ let _javascript = function(hljs) {
     aliases: ['dvc'],
     contains: [
       {
-        begin: /^\s*\$/,
-        end: /([^\\]\n)|\Z/,
+        begin: /    \$/,
+        className: 'skipped'
+      },
+      {
+        begin: /\bls\b|\bcat\b|\bvi\b|\bmkdir\b|\bcd\b|\bwget\b|\bdu\b|\bpython\b|\bcp\b|\bexport\b|\becho\b|\bpip\b|\bcurl\b|\btar\b|\bexec\b|\bautoload\b|\bsudo\b|\bunzip\b|\brm\b|\btree\b|\bfile\b|\bmd5\b|\bsource\b|\bvirtualenv\b/,
```
shcheklein: why do we need this? can we actually make `className: 'skipped'` just a sub-contains as `srtrong` and others were before?

ProjGoodF: The documentation for "react-syntax-highlighter" does not contain instructions for customizing dictionaries. The rules can replace each other, on what basis - is unknown. I tried several options, but only this one works correctly.

ProjGoodF:
Please, repeat review

ProjGoodF:
> Rule for the output is not being applied. At least it's rendered in a different way. Check checkout command reference examples. Please, check other cases I mentioned in the comments.

Please show examples of incorrect work, preferably with screenshots. Since what should be syntax highlighting, we understand only intuitively. Now we found and fixed only one mistake.

shcheklein:
Sure! Please, check https://dvc.org/doc/commands-reference/checkout and the deployment for this PR (find the `View deployment` button on this page) - https://dvc-org-pr-283.herokuapp.com/doc/commands-reference/checkout. Go to the examples section and check the differences - almost every example looks different. 

There is a strict logic in the highlighter - it splits a block into multiple sections and applies certain rules to each of them. Before we merge this we need to test every possible case. For now it's enough for me to open a page and see the difference. Please, check other cases before we do next review (I mentioned a few of them in my comments).

shcheklein:
Still the same issue. I haven't checked other cases I mentioned above.

Screen Shot 2019-05-13 at 10 25 14 AM

VS

Screen Shot 2019-05-13 at 10 24 50 AM

shcheklein:
Fixed it here: 

https://github.com/iterative/dvc.org/commit/b50e38c97f8f8433eaeb6df3819aea533c85cbae#diff-81f1c64a4c5307a1dc8df2bd09305929R39

The fix for this issue:

```
returnBegin: true,
```

+

```
{
            begin: /^\s*\$/,
            className: 'skipped',
          },
```

+ the `skipped` definition.

Everything else in that commit is not relevant to this issue and fixes another long standing issue (let me know if you need more details)+ updates the library. 

Closing this. Let's move to something else.

PR #290, opened 2019-05-05
add note about ssh config in commands-reference/remote:modify.md

khamutov:
Documentation for changes in [dvc#1965](https://github.com/iterative/dvc/pull/1965)

### DIFF
--- a/static/docs/commands-reference/remote:modify.md
+++ b/static/docs/commands-reference/remote:modify.md
@@ -176,7 +176,8 @@ For more information about the variables DVC supports, please visit
 
     1) `user` specified in one of the dvc configs;
     2) `user` specified in the url(e.g. `ssh://user@example.com/path`);
-    3) current user;
+    3) `user` specified in `~/.ssh/config` for remote host;
+    4) current user;
 
   ```dvc
     $ dvc remote modify myremote user myuser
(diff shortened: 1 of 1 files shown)


static/docs/commands-reference/remote:modify.md
```diff
@@ -176,7 +176,8 @@ For more information about the variables DVC supports, please visit
 
     1) `user` specified in one of the dvc configs;
     2) `user` specified in the url(e.g. `ssh://user@example.com/path`);
-    3) current user;
+    3) `user` specified in `~/.ssh/config` for remote host;
```
shcheklein: minor question: is there a similar config on Windows? Should we specify a default path here as well?

efiop: @shcheklein Good point. I thought that it would expand it to something like `C:\Users\user\.ssh\config`(which is correct), but apparently it is not able to handle slashes properly, so we also need to adjust that in the patch to use os.path.join.

shcheklein:
@khamutov @efiop could you please update the doc with the latest changes?

efiop:
@shcheklein In general, the proper path for windows would look like `~\.ssh\config`, since `~` might be resolved differently on windows based on a bunch of env vars, so not sure if it is worth bothering with, since `~/.ssh/config` is already descriptive enough.

shcheklein:
Ruslan, sure! Please, feel free to approve and merge it then.

efiop:
Thank you @khamutov ! :tada:

PR #295, opened 2019-05-05
Added -f option to `dvc remote add` #231

### LINKED ISSUE #231: add -f option to `dvc remote add`

kamdarkarnavee:
Fixes #231

### DIFF
--- a/static/docs/commands-reference/remote:add.md
+++ b/static/docs/commands-reference/remote:add.md
@@ -11,7 +11,7 @@ remotes.
 ## Synopsis
 
 ```usage
-    usage: dvc remote add [-h] [-q | -v] [-d]
+    usage: dvc remote add [-h] [-q | -v] [-d] [-f]
                           [--global] [--system] [--local]
                           name url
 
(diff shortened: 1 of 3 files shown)


shcheklein:
@kamdarkarnavee workflow comment - if you write "Fixes #231" in the PR description Github can do some magic - like close the issue when PR is merged.

static/docs/commands-reference/remote:list.md
```diff
@@ -18,7 +18,7 @@ remotes.
 
 ## Description
 
-Reads DCV configuration files and prints the list of available remotes. 
+Reads DVC configuration files and prints the list of available remotes.
```
shcheklein: thanks!

static/docs/commands-reference/remote:add.md
```diff
@@ -24,7 +24,7 @@ remotes.
 
 `name` and `url` are required. `url` specifies a location to store your data. It
 could be S3 path, SSH path, Azure, Google cloud, local directory, etc - see more
-examples below. If `url1` is a local relative path, it will be resolved relative
+examples below. If `url` is a local relative path, it will be resolved relative
```
shcheklein: 👍

static/docs/user-guide/autocomplete.md
```diff
@@ -29,6 +29,23 @@ Depending on what you typed on the command line so far, it completes:
...
+    $ cat /etc/shells
+
+    # /etc/shells: valid login shells
+      /bin/sh
+      /bin/bash
+      /bin/rbash
+      /bin/dash
+```
+
+For instance, here since Zsh is unavailable, you only need to configure Bash.
```
shcheklein: this change makes sense, but it's not exactly how it works though. Usually engineers are using the same shell they prefer. Either zsh or bash. It's very rare that you need to install it for both at once. What would be useful here is to mention a command that you can run to check what shell is active now. For less experienced users (who don't care is it a bash or zsh) it might save some time.

PR #302, opened 2019-05-07
Fixes #272

### LINKED ISSUE #272: left sidebar: does not support long section names
If section name does not fit into one line it hides another section that follows it.

ProjGoodF:
Fixes #272. The best option is to use jquery.slideToggle, since this function can calculate the height of the element in the case of height: auto

### DIFF
--- a/src/Documentation/SidebarMenu/SidebarMenu.js
+++ b/src/Documentation/SidebarMenu/SidebarMenu.js
@@ -1,4 +1,5 @@
 import React, { Fragment } from 'react'
+import $ from 'jquery'
 // components
 import DownloadButton from '../../DownloadButton'
 // utils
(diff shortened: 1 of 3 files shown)


src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -9,6 +10,24 @@ import styled from 'styled-components'
 import { media, OnlyDesktop } from '../../styles'
 
 export default class SidebarMenu extends React.Component {
+    constructor(props) {
+        super(props);
+        this.collapse = this.collapse.bind(this);
+    }
+    collapse(){
+        setTimeout(function () {
+            $('[data-open=true]').slideDown();
```
shcheklein: please, let's use 2 spaces as in other code

ProjGoodF: Ok, there were 4 spaces in the source code

shcheklein: Yep. Looks there was a mix. Let's change it please, so that we have it consistent. Thanks!

shcheklein:
I'm getting a warning in the logs:

```Warning: Each child in an array or iterator should have a unique "key" prop.

Check the top-level render call using <styled.div>. See https://fb.me/react-warning-keys for more information.
    in React.Fragment
    in SidebarMenu
    in div
    in styled.div
    in div
    in styled.div
    in div
    in styled.div
    in div
    in styled.div
    in Layout
    in withRouter(Layout)
    in Page
    in Documentation
    in StyleSheetManager
    in Unknown
    in Container
    in App
    in Capture```

PR #303, opened 2019-05-07
Fixes #286

### LINKED ISSUE #286: confusing behavior on network fail
When trying to click on a link in menu on dvc.org while network failed it changes address and then does nothing. If you click on some top level menu item other than one you are in then menu works as expected and hide/show animation is shown for a content, but it leaves the content if the last page not new one.

ProjGoodF:
Fixes #286. It seems that with no connection there should be no changes.

### DIFF
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -138,7 +138,7 @@ export default class Documentation extends Component {
         )
       })
     }).catch(()=>{
-      window.location.reload();
+      window.location.reload()
     })
   }
 
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -111,10 +111,6 @@ export default class Documentation extends Component {
     const file = indexFile || flatten(files)[0]
     e && this.setCurrentPath(section, indexFile ? undefined : file)
     this.loadFile({ file, section, parseHeadings: false })
-    this.setState({
-      currentSection: section,
-      pageNotFound: false
-    })
   }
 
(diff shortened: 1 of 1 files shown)


ProjGoodF:
> Either I'm missing something or behavior is the same as before. When there is no network connection it changes the URL but does not show a new page.

we disabled section selection

shcheklein:
> we disabled section selection

When?

I still can click a section when network is not available. Check the deployment for this PR. It changes the URL and does not change the content:


Screen Shot 2019-05-07 at 11 03 01 PM

pages/doc.js
```diff
@@ -141,6 +137,8 @@ export default class Documentation extends Component {
           }
         )
       })
+    }).catch(()=>{
```
shcheklein: will it catch any error? can it cause an infinite loop? can we actually detect a specific error to be safe here?

ProjGoodF: This will catch any http request error.
This cannot cause an endless loop.
Yes, here we can see the specific error of the http request: the code and text of the error.

shcheklein: 👍 My bad. Didn't recognize this is a promise catch.

pages/doc.js
```diff
@@ -141,6 +137,8 @@ export default class Documentation extends Component {
...
+      window.location.reload();
```
shcheklein: let's remove `;` we don't use them in this code base

PR #304, opened 2019-05-08
Addresses "local remote" and S3 remote example confusions

### LINKED ISSUE #284: get started: get rid of local remotes
Here: https://dvc.org/doc/get-started/configure

Remote is a complicated concept for newbies. Local remote is even harder to understand. It is better to use S3\SSH or even with no remotes at all.

EDITED: Local remote breaks project shareability - one of the brightest DVC features.

jorgeorpinel:
Fixes #284

### DIFF
--- a/static/docs/commands-reference/remote:add.md
+++ b/static/docs/commands-reference/remote:add.md
@@ -26,7 +26,7 @@ remotes.
 could be S3 path, SSH path, Azure, Google cloud, local directory, etc - see more
 examples below. If `url1` is a local relative path, it will be resolved relative
 to the current directory and saved to config relative to the config file
-location(see LOCAL example below). Whenever possible DVC will create a remote
+location (see LOCAL example below). Whenever possible DVC will create a remote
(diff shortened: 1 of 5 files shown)


static/docs/commands-reference/remote:add.md
```diff
@@ -75,31 +75,36 @@ is specified for them.
...
+> While the term may seem contradictory, it doesn't have to be.
+The "local" part refers to the machine where the project is stored, so it can be
+any directory accessible to the same system. The "remote" part refers
+specifically to the project/repository itself.
+
+Using an absolute path (recommended):
 
 ```dvc
-    $ dvc remote add myremote ../dir
+    $ dvc remote add myremote /path/to/dir
```
shcheklein: let's use something like `/tmp/dvc-example-storage`

so that command is actually executable

jorgeorpinel: Went for `/tmp/my-dvc-storage`

PR #308, opened 2019-05-10
Added --downstream option to  #271

ameyanator:


static/docs/commands-reference/repro.md
```diff
@@ -97,6 +97,9 @@ local cache and updates stage files with the new checksum information.
 
 * `-v`, `--verbose` - displays detailed tracing information.
 
+- `--downstream` - rerun the commands present in the downstream of the
```
shcheklein: I think we can improve it a little bit. It's not clear what does "downstream of the pipeline means". @mroutis can you give a little bit more details here? How does it exactly limits the scope of stages to run.

mroutis: Sure, @shcheklein ; I would use the same definition that we currently have in the _help_ section:
`Reproduce the pipeline starting from the specified stage.`

Also, I don't know how great is to just copy paste what we have on the CLI, as it doesn't add anything for the user. I left a lot of comments on the code related to how `downstream` works, maybe we can add them in a section about pipelines.

shcheklein:
@ameyanator hey, Ameya! do you need more information or some help to get it done?

shcheklein:
Merging as is, but it requires more work, will keep the ticket open.

PR #312, opened 2019-05-10
Add note about reflinks in tutorial

### LINKED ISSUE #139: add a note about reflinks and how space is calculated in the tutorial
In our article https://dvc.org/doc/tutorial/define-ml-pipeline , we use `du` to show that space taken by the repo didn't double. That works with hardlinks, but `du` is not smart enough when it comes to reflinks, which it counts as separate full-blown filse and accounts for their size two times(for the cache file and for the reflink), which makes it confusing for the mac users https://github.com/iterative/dvc/issues/1442 . We should add a note about it.

kurianbenoy:
Fix #139 

If you want to change the message, please let me know. Also should we mention about the innacurate space calculation by du command?
I am GSOD applicant as well 😊

### DIFF
--- a/static/docs/tutorial/define-ml-pipeline.md
+++ b/static/docs/tutorial/define-ml-pipeline.md
@@ -15,11 +15,6 @@ to download `data.xml`and save it into the `data` subdirectory.
 
 </details>
 
-> Note: In case of systems supporting reflink, use `df` utility to see that free 
-space on the drive didn't decline by the file size that we are adding and no 
-duplication takes place. As `du` is inaccurate in modern file systems using
-reflinks.
-
 ```dvc
     $ mkdir data
     $ wget -P data https://dvc.org/s3/so/100K/Posts.xml.zip
@@ -106,6 +101,11 @@ repository is still 41MB. Both of the files correspond to the same `inode` (file
 meta-data record) in a file system. Use `ls -i` to see file system inodes. If
 you are using a modern file system with reflinks you might see different inodes,
 still only one copy if the actual file data is stored.
+
+> Note: In case of systems supporting reflink, use `df` utility to see that free 
+space on the drive didn't decline by the file size that we are adding and no 
+duplication takes place. As `du` is inaccurate in modern file systems using
+reflinks.
 
 ```dvc
     $ ls -i data/Posts.xml.zip
--- a/static/docs/tutorial/define-ml-pipeline.md
+++ b/static/docs/tutorial/define-ml-pipeline.md
@@ -15,6 +15,11 @@ to download `data.xml`and save it into the `data` subdirectory.
 
 </details>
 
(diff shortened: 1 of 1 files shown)


static/docs/tutorial/define-ml-pipeline.md
```diff
@@ -118,6 +118,9 @@ and Windows. Some implementation details (like inodes) might differ, but the
 overall DVC behavior is the same. See also [DVC Files and
 Directories](/doc/user-guide/dvc-files-and-directories)
```
shcheklein: this note above should be update to mention reflinks on modern file systems. I'm not sure why do we need the reference to the DVC files and directories here.

static/docs/tutorial/define-ml-pipeline.md
```diff
@@ -118,6 +118,9 @@ and Windows. Some implementation details (like inodes) might differ, but the
...
 
+> Note: to prove that no duplication occurs, we could use df utility to see that
```
shcheklein: we need to mention that we need to refer to df only in case of reflinks.

shcheklein: I would also put this note above the example, so that a user can see it in advance

shcheklein:
> Fix #139
> 
> If you want to change the message, please let me know. Also should we mention about the innacurate space calculation by du command?
> I am GSOD applicant as well 😊

Yes, I would mention that du is inaccurate in case of a modern file system that is using reflinks.

static/docs/tutorial/define-ml-pipeline.md
kurianbenoy: Can I remove the reference to DVC files and directories?

shcheklein: @kurianbenoy Yes! I think it made it here by mistake. This link is not relevant. There should be a link to some doc that explains everything about link types, something like that one I shared below.

static/docs/tutorial/define-ml-pipeline.md
```diff
@@ -15,6 +15,11 @@ to download `data.xml`and save it into the `data` subdirectory.
 
 </details>
 
+> Note: In case of systems supporting reflink, use `df` utility to see that free 
+space on the drive didn't decline by the file size that we are adding and no 
+duplication takes place. As `du` is inaccurate in modern file systems using
+reflinks.
```
shcheklein: Looks great! But it should go right before the code block that includes `du -sh .`. The point is to show that even we now have a file in the working space *and* in the `.dvc/cache` they still take the same amount of space.

kurianbenoy: Thanks, incorporated the feedback.

kurianbenoy:
> Just move the paragraph a bit. Otherwise looks great! Thank you.

Made the changes, @shcheklein I highly value the quick response in dvc.org :smile:

PR #319, opened 2019-05-12
Format code according to prettier

### LINKED ISSUE #318: add pre-commit hook to format code
`.prettierrc` file needs to be updated with configurations that we follow to write code. Like we use 2 spaces for indentation, no trailing white space, etc.
Then these configurations can be used to format code before committing.

algomaster99:
- [x] Modify `.prettierrc`
- [x] Add a pre-commit hook

`printWidth: 80` sets the rule for a line to not exceed `80` character limit.

I will add the pre-commit hook after the configuration is approved so that all the changes in formatting of JS files are committed too.

Fix #318


package.json
```diff
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "NODE_ENV=production node server.js",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "prettier-src": "prettier --no-semi --write '{pages,src}/**/*.{js,jsx}'"
+    "prettier-src": "prettier --write '{pages,src}/**/*.{js,jsx}'"
```
algomaster99: Flag `--no-semi` is not required since it is already specified in `.prettierrc` as `semi: false`.

.prettierrc
```diff
@@ -1,3 +1,6 @@
 semi: false
 singleQuote: true
 trailingComma: none
+printWidth: 100
+tabWidth: 2
+useTabs: false
```
shcheklein: let's add newline at the end

algomaster99:
@shcheklein @mroutis 
* I have added the prettier hook. I am looking for a hook which will format markdown files as well. But I think the markdown linter shouldn't be strict enough since, sometimes, markdown gets rendered wrongly if formatted correctly.
* Next, I will commit the JS files with formatting.
* This will require changes in contribution guide for which should a separate PR be opened?

shcheklein:
@algomaster99 Thanks! And yes, let's create a PR to the contributing guide that explains how this pre-commit hook can be installed and how to run it.

PR #325, opened 2019-05-13
Update metrics.md

sudodoki:
Change eval.txt references to eval.json in metrics file - just based on content seemed like this is leftover from previous versions


shcheklein:
@sudodoki thanks!

PR #328, opened 2019-05-13
Format markdown files according to prettier

algomaster99:
This pull request formats all markdown files according to prettier version 1.17.0. It was done by running the command `npm run prettier-md`.


static/docs/commands-reference/import.md
```diff
@@ -34,25 +34,25 @@ The `dvc import` command helps the user create such an external data dependency.
...
-`remote` | Remote path (see explanation below) | `remote://myremote/path/to/file`
+| Type     | Discussion                                              | URL format                                 |
+| -------- | ------------------------------------------------------- | ------------------------------------------ |
+| `local`  | Local path                                              | `/path/to/local/file`                      |
+| `s3`     | Amazon S3                                               | `s3://mybucket/data.csv`                   |
+| `gs`     | Google Storage                                          | `gs://mybucket/data.csv`                   |
+| `ssh`    | SSH server                                              | `ssh://user@example.com:/path/to/data.csv` |
+| `hdfs`   | HDFS                                                    | `hdfs://user@example.com/path/to/data.csv` |
+| `http`   | HTTP to file with _strong ETag_ (see explanation below) | `https://example.com/path/to/data.csv`     |
+| `remote` | Remote path (see explanation below)                     | `remote://myremote/path/to/file`           |
```
shcheklein: ❤️

static/docs/commands-reference/metrics:add.md
```diff
@@ -21,33 +21,30 @@ Alternatively, an output file could be made a metric via `-M` or `-m` parameter
...
-
-* `-x`, `--xpath` - specify a path within a metric file to get a specific metric
-value. Should be used if metric file contains multiple numbers and you need to
-get a only one of them. Only single path is allowed. This path will be saved
-into the corresponding `.dvc` file and will be used automatically in `dvc
-metrics show`. Accepted value depends on the metric file type (`-t` option):
+- `-t`, `--type` - specify a type of the metric file. Accepted values are:
+  `raw`, `json`, `tsv`, `htsv`, `csv`, `hcsv`. Type will be used to determine how
+  `dvc metrics show` handles displaying it. This type will be saved into the
+  corresponding `.dvc` file and will be used automatically in the `dvc metrics show`. `htsv`/`hcsv` are the same `tsv`/`csv` but the values in the first row of
```
shcheklein: a bug?

algomaster99: @shcheklein seems like it. Let me check on this.

algomaster99:
@shcheklein 
>     not sure what is the policy for code blocks - 2 spaces, no spaces, 4 spaces?
- [ ] Let's wait for [this issue](https://github.com/prettier/prettier/issues/6117) to be resolved to address this.

>     it looks like it does not preserve 80 symbols and does not wrap lines automatically

- [ ] Apparently, these is markdown support for word wrap. It should have been formatted. I will look into this problem as soon as possible.

For now, let's wait for these subtasks to resolve.

algomaster99:
@shcheklein Using `proseWrap: always` did the trick. Now prettier will wrap the markdown code if the length increases `100` characters. It takes `100` since it is configured in `.prettierrc`.
I suggest we can explicitly mention `--print-width 100` in the `npm run prettier-js` command and change it to default in `.prettierrc`.

algomaster99:
@shcheklein 
- Paragraphs are wrapped at 80 characters.
- Indentation of code blocks will have to be fixed manually. :disappointed:
- Since `printWidth: 80`, we will have to format all JS files again.

kurianbenoy:
If that is the case, I can manually fix it if needed. Or should we leverage
the use of some scripts?

algomaster99:
@kurianbenoy I would recommend writing a bash script. Let's do smart work. :wink:. But still, wait for @shcheklein to respond.

static/docs/changelog/0.35.md
```diff
@@ -1,74 +1,75 @@
...
 
-* 📖 [Get started guide](https://dvc.org/doc/get-started/agenda) has been
+- 📖 [Get started guide](https://dvc.org/doc/get-started/agenda) has been
   simplified (e.g. to use tags instead of branches) and extended. We have also
-  prepared a [Github DVC project
-  ](https://github.com/iterative/example-get-started)that reflects the sequence
-  of steps in the “get started” guide. You can now download the whole project
-  and reproduce all the models.
+  prepared a
+  [Github DVC project ](https://github.com/iterative/example-get-started)that
```
shcheklein: _minor_: add a space after `get-started)`

.pre-commit-config.yaml
```diff
@@ -3,3 +3,4 @@ repos:
     rev: 1.17.0
     hooks:
       - id: prettier
+        args: ['--prose-wrap=always']
```
shcheklein: _minor_: do we need it here if it's already specified in the config file?

shcheklein:
> @kurianbenoy I would recommend writing a bash script. Let's do smart work. 😉. But still, wait for @shcheklein to respond.

agreed! script makes sense in this case.

.pre-commit-config.yaml
algomaster99: @shcheklein I tested this a little bit. I figured out that the pre-commit hook wasn't wrapping the text which could be because it was following the default config which is `proseWrap: preserve`. So I needed to explicitly pass the argument.
So yes, pre-commit hook doesn't read from `.prettierrc` file.

algomaster99:
@kurianbenoy Are you working on the bash script?

kurianbenoy:
> @kurianbenoy Are you working on the bash script?

No, I am not working on the bash script @algomaster99.

PR #330, opened 2019-05-13
[WIP] Fix #313

### LINKED ISSUE #313: use css for padding instead of 4 leading spaces in code blocks
From https://github.com/iterative/dvc.org/issues/310

kurianbenoy:
@ivan I have been thinking about using css padding by calling a html element like this:
```
<div class="intent">
$ sudo yum update
$ sudo wget https://dvc.org/rpm/dvc.repo -O /etc/yum.repos.d/dvc.repo
 $ sudo yum install dvc
$ sudo yum update
$ sudo yum install dvc
</div>
```

with .intent having property padding-left:4em as css attribute.

 I am not familiar with React, so I am not sure, what is correct change to be done there. Yet I hope this approach work, as its done for tag <detail> in docs.

Thanks

### DIFF
--- a/src/Documentation/Markdown/Markdown.js
+++ b/src/Documentation/Markdown/Markdown.js
@@ -61,7 +61,7 @@ const HtmlRenderer = props => {
 const CodeBlock = ({ value, language }) => {
   const dvcStyle = Object.assign({}, docco)
   dvcStyle['hljs-comment'] = { color: '#999' }
-  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px' }
+  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px', paddingLeft: '8em' }
   return (
     <SyntaxHighlighter language={language} style={dvcStyle}>
       {value}
--- a/src/Documentation/Markdown/Markdown.js
+++ b/src/Documentation/Markdown/Markdown.js
@@ -61,7 +61,8 @@ const HtmlRenderer = props => {
 const CodeBlock = ({ value, language }) => {
   const dvcStyle = Object.assign({}, docco)
   dvcStyle['hljs-comment'] = { color: '#999' }
-  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px', paddingLeft: '8em' }
+  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px' }
+  dvcStyle['hljs']['padding'] = '0.5em 0.5em 0.5em 2em'
   return (
     <SyntaxHighlighter language={language} style={dvcStyle}>
       {value}
--- a/static/docs/get-started/add-files.md
+++ b/static/docs/get-started/add-files.md
@@ -16,15 +16,15 @@ link as`(Chrome) or `Save object as`(Firefox).
 </details>
 
 ```dvc
-    $ mkdir data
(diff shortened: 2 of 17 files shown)


kurianbenoy:
@shcheklein I guess the change I made in `src/Documentation/Markdown/Markdown.js` in line 64 fixes the padding. Can I remove the 4 leading spaces in all documents in Get started section?

src/Documentation/Markdown/Markdown.js
```diff
@@ -61,7 +61,7 @@ const HtmlRenderer = props => {
 const CodeBlock = ({ value, language }) => {
   const dvcStyle = Object.assign({}, docco)
   dvcStyle['hljs-comment'] = { color: '#999' }
-  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px' }
+  dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px', paddingLeft: '8em' }
```
shcheklein: I think you should try something like this: `dvcStyle['hljs']['padding']='0.5em 0.5em 0.5em 8em'` - it's a more general solution that would apply padding to the whole code block

kurianbenoy:
I have incorporated the changes you suggested.

shcheklein:
@kurianbenoy check the deployment:

Screen Shot 2019-05-13 at 10 24 50 AM

It looks like 2em should be enough.

src/Documentation/Markdown/Markdown.js
```diff
@@ -62,6 +62,7 @@ const CodeBlock = ({ value, language }) => {
   const dvcStyle = Object.assign({}, docco)
   dvcStyle['hljs-comment'] = { color: '#999' }
   dvcStyle['hljs-meta'] = { color: '#333', fontSize: '14px' }
+  dvcStyle['hljs']['padding']='0.5em 0.5em 0.5em 8em'
```
shcheklein: space before and after `=`

shcheklein: like I mentioned, let's try `2em`, it should be enough

kurianbenoy:
Screenshot-2019-5-14 Data Science Version Control System(1)

kurianbenoy:
I feel it should have been better to use 3 or 4 em to be as aligned to current docs

shcheklein:
@kurianbenoy yep, 3em looks like a good option. We def don't need more.

shcheklein:
@kurianbenoy actually, even 2em is enough ) Up to you, 2em or 3em :)

kurianbenoy:
Ok let's stick to 2em then @shcheklein

shcheklein:
@algomaster99 @jorgeorpinel I'm going to merge this. Just keep in mind guys, that we don't need spaces any more. @algomaster99 when you run the prettier next time (if we can figure out all the caveats with it), please, make sure that it removes all spaces everywhere.

@kurianbenoy thank you! :)

algomaster99:
@shcheklein all the spaces should be removed inside code blocks know? There should be no indentation right?

shcheklein:
@algomaster99 correct! no need for the manual indentation anymore. @kurianbenoy already did this across the get started section. Would be great if `prettier` can handle everything else. Please, let us know.

@kurianbenoy let's first wait on @algomaster99 if it can be done automatically. If `prettier` can't handle this we can probably write a script do this across all `.md` files.

PR #341, opened 2019-05-15
Format JS files according to prettier

algomaster99:
- All the prose are wrapped at 80 characters.


shcheklein:
Thanks @algomaster99 !

Issue #265, opened 2019-04-24
add a user guide page about the optimization (advanced mode)


shcheklein:
In the recent DVC release we are changing the way we are linking DVC controlled files with the working space via https://github.com/iterative/dvc/pull/1841, https://github.com/iterative/dvc/pull/1914 and a potentially a few other PRs.

The motivation for this change could be found here: https://github.com/iterative/dvc/issues/1599

This is a major change and affect the user workflow, cache settings, performance, etc. Sometimes is a way that can make users unhappy - since copying is happening.

Things we need to do here:

- [x] we should write a user guide section about the links (some information can be taken from - the https://github.com/iterative/dvc/issues/1599);
- [x] we need to write a section "advanced mode" or "performance optimization" which should be a brief guide on how to switch DVC into using `hardlinks`, `symlinks`, etc, what caveats ppl should expect (that it's recommended to use protected mode by default and `dvc unprotect` in those cases);
- [x] we should put links to these sections wherever it's appropriate;
- [x] we should update the [config](https://dvc.org/doc/commands-reference/config) doc - introduce new flags;
- [x] we should review and update existing mentions of the `dvc unprotect`;
- [x] [Update a Tracked File](https://dvc.org/doc/user-guide/update-tracked-file) section itself should be updated;

mroutis:
Related: https://github.com/iterative/dvc/pull/1997/files#r283492357

jorgeorpinel:
Thanks for merging #345 Ivan. I still have to double check the last 2 check boxes in this issue though. I'll close again when I do so if you don't mind.

shcheklein:
oops, Github closed it automatically :)

PR #344, opened 2019-05-15
`dvc cache` and `dvc cache dir` command references

jorgeorpinel:

PR #345, opened 2019-05-17
New cache file linking guide and related updates to existing docs

jorgeorpinel:
Partially addresses #265

### DIFF
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -87,8 +87,8 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
   The default value is `cache`, which resolved relative to the default project
   config location results in `.dvc/cache`.
   > See also helper command `dvc cache dir` that properly transform paths
-  relative to the present working directory into relative to the project config
-  file.
+  > relative to the present working directory into relative to the project
+  > config file.
 
 - `cache.protected` - makes files in the workspace read-only. Possible values
   are `true` or `false` (default). Run `dvc checkout` for the change go into
@@ -103,15 +103,14 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
 
 - `cache.type` - link type that DVC should use to link data files from cache to
   your workspace. Possible values: `reflink`, `symlink`, `hardlink`, `copy` or a
-
   combination of those, separated by commas e.g: `reflink,hardlink,copy`.  
   By default, DVC will try `reflink,copy` link types in order to choose the most
   effective of those two. DVC avoids `symlink` and `hardlink` types by default
   to protect user from accidental cache and repository corruption.  
   > **Note!** If you manually set `cache.type` to `hardlink` or `symlink`, **you
-  will corrupt the cache** if you edit data files in the workspace. See the
-  `cache.protected` config option above and corresponding `dvc unprotect`
-  command to modify files safely.  
+  > will corrupt the cache** if you modify tracked data files in the workspace.
+  > See the `cache.protected` config option above and corresponding
+  > `dvc unprotect` command to modify files safely.  
   There are pros and cons to different link types. Refer to [Cache File
   Linking](/docs/user-guide/cache-file-linking) for a full explanation of each
   one.
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -73,7 +73,7 @@ This is the main section with the general config options:
 These are sections in the config file that describe particular remotes. These
 sections contain a `url` value, and can also specify `user`, `port`, `keyfile`,
 `timeout`, `ask_password`, and other cloud-specific key/value pairs for each
-remote. See `dvc remote` for more info.
+remote. See `dvc remote` for more information.
 
 ### cache
 
@@ -92,69 +92,41 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
 
 - `cache.protected` - makes files in the workspace read-only. Possible values
   are `true` or `false` (default). Run `dvc checkout` for the change go into
-  effect. (It affects only files that are under DVC control.)
+  effect. (It affects only files that are under DVC control.)  
   Due to the way DVC handles linking between the data files in the cache and
   their counterparts in the working directory, it's easy to accidentally corrupt
   the cached version of a file by editing or overwriting it. Turning this config
   option on forces you to run `dvc unprotect` before updating a file, providing
-  an additional layer of security to your data.
+  an additional layer of security to your data.  
   It's highly recommended to enable this mod when `cache.type` is set to
   `hardlink` or `symlink`.
 
 - `cache.type` - link type that DVC should use to link data files from cache to
   your workspace. Possible values: `reflink`, `symlink`, `hardlink`, `copy` or a
-  combination of those, separated by commas: `reflink,symlink`.
-  By default, DVC will try `reflink` and `copy` link type in order to choose the
-  most effective of those two. DVC avoids `symlink` and `hardlink` types by
-  default to protect user from accidental cache and repository corruption.
-  > **Note!** Unless your workspace supports `reflinks` – if you are on a recent
-  Mac chances are you are using `reflinks` – or you've manually specified
-  `cache.type copy` **you are corrupting** the cache if you edit data files in
-  the workspace. See the `cache.protected` config option above and corresponding
-  `dvc unprotect` command to modify files safely.
 
-  There are pros and cons to different link types. Each type is explained below,
-  from the best and most efficient to the least efficient:
-
-  1. **`reflink`** - this is the best link type that could be. It is as fast as
-     hard/symlinks, but doesn't carry a risk of cache corruption, since
-     filesystem takes care of copying the file if you try to edit it in place,
-     thus keeping a linked cache file intact.
-     Unfortunately reflinks are currently supported on a limited number of
-     filesystems (Linux: Btrfs, XFS, OCFS2; MacOS: APFS), but they are coming to
-     every new filesystem and in the future will be supported by the majority of
-     filesystems in use.
-
-  2. **`hardlink`** - the most efficient way to link your data to cache if both
-     your repo and your cache directory are located on the same
-     filesystem/drive.
-     Please note that hardlinked data files should never be edited in place, but
-     instead deleted and then replaced with a new file, otherwise it might cause
-     cache corruption and automatic deletion of a cache file by dvc.
-
-  3. **`symlink`** - The most efficient way to link your data to cache if your
-     repo and your cache directory are located on different filesystems/drives
-     (i.e. repo is located on ssd for performance, but cache dir is located on
-     hdd for bigger storage).
-     Please note that data file linked with symlink should never be edited in
-     place, but instead deleted and then replaced with a new file, otherwise it
-     might cause cache corruption and automatic deletion of a cache file by dvc.
-
-  4. **`copy`** - The most inefficient link type, yet the most widely supported
-     for any repo/cache FS combination. Suitable for scenarios with relatively
-     small data files, where copying them is not a performance/storage concern.
+  combination of those, separated by commas e.g: `reflink,hardlink,copy`.  
+  By default, DVC will try `reflink,copy` link types in order to choose the most
+  effective of those two. DVC avoids `symlink` and `hardlink` types by default
+  to protect user from accidental cache and repository corruption.  
+  > **Note!** If you manually set `cache.type` to `hardlink` or `symlink`, **you
+  will corrupt the cache** if you edit data files in the workspace. See the
+  `cache.protected` config option above and corresponding `dvc unprotect`
+  command to modify files safely.  
+  There are pros and cons to different link types. Refer to [Cache File
+  Linking](/docs/user-guide/cache-file-linking) for a full explanation of each
+  one.
 
 - `cache.slow_link_warning` - used to turn off the warnings about having a slow
   cache link type. These warnings are thrown by `dvc pull` and `dvc checkout`
   when linking files takes longer than usual, to remind them that there are
-  faster cache link types available than the defaults (`reflink` or `copy` – see
+  faster cache link types available than the defaults (`reflink,copy` – see
   `cache.type`). Accepts values `true` and `false`.
   > These warnings are automatically turned off when `cache.type` is manually
   > set.
 
 - `cache.local` - name of a local remote to use as local cache. This will
   overwrite the value provided to `dvc config cache.dir` or `dvc cache dir`.
-  Refer to `dvc remote` for more info on "local remotes".
+  Refer to `dvc remote` for more information on "local remotes". 
 
 - `cache.ssh` - name of an [SSH remote to use as external
   cache](/doc/user-guide/external-outputs#ssh).
--- a/static/docs/get-started/add-files.md
+++ b/static/docs/get-started/add-files.md
@@ -40,10 +40,11 @@ $ git commit -m "add source data to DVC"
 
 ### Expand to learn about DVC internals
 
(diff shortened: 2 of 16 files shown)


jorgeorpinel:
@shcheklein the guide isn't here yet but you may want to double check the file changes in this PR so far (reflecting https://github.com/iterative/dvc/pull/1841), which are needed for consistency with recent updates I made (#344) and with the new user guides I'll push up here soon. Thanks ✌️

static/docs/get-started/add-files.md
```diff
@@ -41,7 +41,7 @@ $ git commit -m "add source data to DVC"
 ### Expand to learn about DVC internals
 
 You can see that actual data file has been moved to the `.dvc/cache` directory
-(usually hardlink or reflink is created, so no physical copying is happening).
+(usually a reflink is created, so no file content copying is happening).
```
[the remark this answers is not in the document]
shcheklein: I would actually make an details section - something like: `IMPORTANT: Performance on large data or something` ...

static/docs/get-started/example-pipeline.md
```diff
@@ -167,8 +167,8 @@ intermediate result.
 Second, you should see by now that the actual data is stored in the `.dvc/cache`
 directory, each file having a name in a form of an md5 hash. This cache is
 similar to Git's internal objects store but made specifically to handle large
-data files. DVC is using reflinks, hardlinks and other optimizations to manage
-your actual workspace without copying every time object from/to the cache.
+data files. DVC uses reflinks, hardlinks, or other optimizations to manage your
```
shcheklein: the same, we should be emphasizing this. It was really painful decision to disable hardlinks and symlinks. A lot of ppl might be confused and disappointed to see that we are copying their files. We should give a lot of warnings and docs so that they understand right away that there is a way to enable an optimization.

static/docs/philosophy/related-technologies.md
```diff
@@ -81,7 +81,8 @@ process.
      want to see in your Git repository) in a local key-value store and use file
      symlinks instead of the actual files.
 
-   - DVC uses hardlinks instead of symlinks to make user experience better.
+   - DVC uses reflinks (or hardlinks) instead of symlinks to make user
```
shcheklein: I think the previous paragraph should be updated. I don't see how two of them make sense together. @dmpetrov @efiop what did we want to actually say here? are you familiar with git-annex internals?

static/docs/philosophy/related-technologies.md
```diff
@@ -111,10 +112,9 @@ process.
      `git clone` command. It gives more granularity on managing data and code
      separately. Hooks could be configured to make workflow simpler.
 
-   - DVC creates hardlinks (or even reflinks if they are supported). The
-     `dvc checkout` command does not actually copy data files from cache to the
-     workspace, as copying files is a heavy operation for large files (30
-     GB+).
+   - DVC creates reflinks. The `dvc checkout` command does not actually copy
```
shcheklein: reflinks are rare. This phrase could be misleading. Btw, in places like this I would use a term copy-on-write links (aka reflinks)  and would put a link to some docs about them. It's not a very well known stuff.

static/docs/philosophy/related-technologies.md
efiop: @shcheklein Ah, right, this is an old one. Not sure why the first paragraph is that way. I would probably remove it for now.

Also, would rephrase a little: `DVC can use reflinks or hardlinks instead of symlinks to make user experience better.`

static/docs/get-started/add-files.md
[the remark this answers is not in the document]
jorgeorpinel: OK. I added that tradeoff information to the new file link type guide (now pushed) and linked to it here instead.

static/docs/get-started/example-pipeline.md
jorgeorpinel: Split into special note.

static/docs/philosophy/related-technologies.md
jorgeorpinel: Updated.

static/docs/philosophy/related-technologies.md
jorgeorpinel: Added as a foot note in this document but what would be a good source to link to about it? I haven't found anything that looks authoritative enough. Wikipedia? https://en.wikipedia.org/wiki/Copy-on-write

jorgeorpinel:
OK guys I addressed the feedback but also had to push up all the other changes I have which add the new /doc/user-guide/cache-file-linking guide (See it [live in review app](https://dvc-org-pr-345.herokuapp.com/doc/user-guide/cache-file-linking)). Hopefully it's not too many changes to review now!

static/docs/user-guide/cache-file-linking.md
```diff
@@ -0,0 +1,93 @@
+# File Link Types for the DVC Cache
+
+File links are entries in the file system that don't necessarily hold the file
+contents, but which point to where the file is actually stored. DVC uses file
+links in your project's workspace to avoid copying them from/to the DVC cache.
```
shcheklein: them - sounds like "to avoid copying links"

static/docs/user-guide/cache-file-linking.md
```diff
@@ -0,0 +1,93 @@
...
+> [ReFS](https://docs.microsoft.com/en-us/windows-server/storage/refs/refs-overview)
+> file systems.
+
+## Supported file links types and their tradeoffs
+
+There are pros and cons to the 3 supported link types (`reflink`, `hardlink`,
+`symlink` or soft link). While reflinks have all the benefits and none of the
+worries, they're not commonly supported in most platforms yet. Hard/soft links
+also optimize speed and space in the file system, but carry the risk of breaking
+your workflow ,since updating tracked files in the workspace causes data
```
shcheklein: ` ,` -> `, `

static/docs/commands-reference/config.md
```diff
@@ -103,58 +103,29 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
...
-     repo and your cache directory are located on different filesystems/drives
-     (i.e. repo is located on ssd for performance, but cache dir is located on
-     hdd for bigger storage).  
-     Please note that data file linked with symlink should never be edited in
-     place, but instead deleted and then replaced with a new file, otherwise it
-     might cause cache corruption and automatic deletion of a cache file by dvc.
-
-  4. **`copy`** - The most inefficient link type, yet the most widely supported
-     for any repo/cache FS combination. Suitable for scenarios with relatively
-     small data files, where copying them is not a performance/storage concern.
```
shcheklein: ❤️

static/docs/get-started/example-pipeline.md
```diff
@@ -167,8 +167,12 @@ intermediate result.
 Second, you should see by now that the actual data is stored in the `.dvc/cache`
 directory, each file having a name in a form of an md5 hash. This cache is
 similar to Git's internal objects store but made specifically to handle large
-data files. DVC is using reflinks, hardlinks and other optimizations to manage
-your actual workspace without copying every time object from/to the cache.
+data files.
+
+> **Note!** For performance with large data files, DVC can use file links from
+the cache to the workspace to avoid copying actual file contents. Refer to
```
shcheklein: did you run the pre-commit hook we have now? I think it puts `>` before each line

static/docs/get-started/example-versioning.md
```diff
@@ -227,9 +227,10 @@ $ python train.py
 $ dvc add model.h5
 ```
 
-Note! `dvc remove` or `dvc unprotect` is required, otherwise `python train.py`
-will overwrite the existing file and may corrupt the cached version. Check this
-[guide](/doc/user-guide/update-tracked-file) to learn more.
+> **Note!** `dvc remove` or `dvc unprotect` is required, otherwise `python
```
shcheklein: it's not required anymore. It's required only in the protected mode. By default it's disabled and it's safe to modify/overwrite files.

static/docs/get-started/add-files.md
```diff
@@ -41,7 +41,9 @@ $ git commit -m "add source data to DVC"
 ### Expand to learn about DVC internals
 
 You can see that actual data file has been moved to the `.dvc/cache` directory
-(usually hardlink or reflink is created, so no physical copying is happening).
+(ideally with reflinks if available on the system, otherwise by file copy – see
```
shcheklein: let's make it a separate `<details>` section - something like `Important performance caveats` or something like this.

static/docs/user-guide/update-tracked-file.md
```diff
@@ -1,10 +1,14 @@
...
 Due to the way DVC handles linking between the data files in the cache and their
-counterparts in the working directory (see
-[#799](https://github.com/iterative/dvc/issues/799) and
-[#599](https://github.com/iterative/dvc/issues/599) for example), updating
-tracked files has to be carried out with caution.
+counterparts in the working directory (refer to [Cache File
+Linking](/docs/user-guide/cache-file-linking)), updating tracked files has to be
+carried out with caution to avoid data corruption when the DVC config option
+`cache.type` is set to `hardlink` or/and `symlink`. (See `dvc config cache` for
+more details on setting the cache file link types.)
```
shcheklein: is it `).`?

static/docs/user-guide/cache-file-linking.md
```diff
@@ -0,0 +1,93 @@
+# File Link Types for the DVC Cache
```
shcheklein: I'm thinking if we should rename it to more user friendlier version - "Performance with large data" or something similar? Or are you thinking to put together a separate sections for that?

static/docs/philosophy/related-technologies.md
jorgeorpinel: Link added.

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: Changed to "avoid duplicating contents that will exist in the DVC cache."

static/docs/get-started/example-pipeline.md
jorgeorpinel: I had not but just installed the pre-commit hook (which uses `prettier`) and it does add them indeed. BTW Rewrap also puts `>` before each line but only if a long comment has no line breaks yet, otherwise it doesn't...

Also, `prettier` re-wraps my paragraphs differently than VSCode Rewrap. (The pre-commit hook doesn't line-break links.)

So there's inconsistencies throughout the docs with these 2 things at least. Maybe I should just run `prettier` on everything?

jorgeorpinel: OK... I just ran ``npm bin`/prettier --write --prose-wrap=always` on all the files altered by this PR and the funny thing is that when `pre-commit` ran prettier again the results were different. So I'm not sure how to run `prettier` so it has the same settings as when it is ran by `pre-commit`.

Anyway, I noticed some problems with the auto-reformatting like that it changes how we wrote sidebar.json and it inserts more spaces to indent `yaml` code-blocks showing DVC-file examples so I'm turning it off for now. I'll ask about all this on Discord.

static/docs/user-guide/update-tracked-file.md
jorgeorpinel: In this case the entire sentence is inside the parenthesis so `.)` is correct 😃

I double checked just now and found several sources verifying this e.g. https://www.quickanddirtytips.com/education/grammar/periods-and-parentheses

```
If the words inside the parentheses are a complete sentence, the period, question
mark, or exclamation point that ends the sentence goes inside the parenthesis:

Bring chocolate. (Squiggly likes sweets.)
```

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: Yes! Actually this is not even the main new guide you requested in #265 **add a user guide page about the optimization (advanced mode)**. My plan was to add another guide for that matter, which will link back to this one.

Maybe I've added too much optimization info already to this one guide so no longer sure what will the optimization guide will contain. Is there anything else to optimize besides cache file link types?

static/docs/get-started/example-versioning.md
jorgeorpinel: Good point. Note removed.

static/docs/get-started/add-files.md
jorgeorpinel: Ok, done. This doc may have too many expandable detail sections now though, you may want to see it in the review app: https://dvc-org-pr-345.herokuapp.com/doc/get-started/add-files

static/docs/commands-reference/config.md
```diff
@@ -87,74 +87,45 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
...
   > See also helper command `dvc cache dir` that properly transform paths
-  relative to the present working directory into relative to the project config
-  file.
+  > relative to the present working directory into relative to the project
+  > config file.
 
 - `cache.protected` - makes files in the workspace read-only. Possible values
   are `true` or `false` (default). Run `dvc checkout` for the change go into
-  effect. (It affects only files that are under DVC control.)
+  effect. (It affects only files that are under DVC control.)
```
shcheklein: remove trailing whitespaces

static/docs/commands-reference/config.md
```diff
@@ -87,74 +87,45 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
...
 - `cache.protected` - makes files in the workspace read-only. Possible values
   are `true` or `false` (default). Run `dvc checkout` for the change go into
-  effect. (It affects only files that are under DVC control.)
+  effect. (It affects only files that are under DVC control.)  
   Due to the way DVC handles linking between the data files in the cache and
   their counterparts in the working directory, it's easy to accidentally corrupt
   the cached version of a file by editing or overwriting it. Turning this config
   option on forces you to run `dvc unprotect` before updating a file, providing
-  an additional layer of security to your data.
+  an additional layer of security to your data.
```
shcheklein: trailing whitespaces

static/docs/user-guide/cache-file-linking.md
shcheklein: no, not that much .. I think this guide alone is probably enough. I would rename, may be write a brief motivation - that space and speed are important and we can't just copy, etc, etc. and call it a section :)

static/docs/commands-reference/config.md
jorgeorpinel: They are in on purpose. In markdown they help to insert a line break within a paragraph.

Explained here: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lists

static/docs/commands-reference/config.md
jorgeorpinel: It's another line break. Should I remove these and add full paragraphs breaks instead?

jorgeorpinel:
Ready, @shcheklein

jorgeorpinel:
Ready, @shcheklein. Just a couple discussions still open above for a final decision on your part please. Thanks

static/docs/commands-reference/config.md
shcheklein: Ah, np then. I'm totally fine with two paragraph when you think it's appropriate.

static/docs/commands-reference/config.md
shcheklein: So, is this an actual whitespace? I'm fine with that but I would expect it to be fragile - some editors remove trailing whitespaces on save and prettier won't return them most likely. Not a big deal though, let's not worry about this now.

shcheklein:
Great stuff, @jorgeorpinel ! 🎉

static/docs/commands-reference/config.md
jorgeorpinel: You're right. We may want to avoid these in the future... I'll search all the docs to see if there's many occurrences.

Issue #256, opened 2019-04-16
Replace the CSV file with the Pickle file in the documentation


maggyero:
https://dvc.org/doc/dvc-philosophy/how-it-works

At step 2, `cnn_train.py` is run and yields the `results.csv` file (???) instead of the `output.p` file mentioned at step 4.

shcheklein:
Fixed by https://github.com/iterative/dvc.org/commit/9194fe303cc9a9a7dd31b133abd8b2afdb6501f6

Thanks @maggyero for the attention to these small details!

Issue #363, opened 2019-05-20
pre-commit hook prettier (vs npx prettier) questions


jorgeorpinel:
~~First, running `npm run prettier-md` does not yield the same results as what the pre-commit hook does (hook installation explained in https://dvc.org/doc/user-guide/contributing-documentation). Should it? And how do we equalize them if so?~~

~~For example with `npm` the bullet lists are standardized to `*` while with the pre-commit hook they're `-`. I understand we want the latter.~~

UPDATE: I needed to run `npm install` as my node_modules `prettier` version was outdated. `npx prettier --write <file pattern>`. Apparently oesn't necessarily behave the same as the pre-commit hook (which specifies a prettier version in **.pre-commit-config.yaml**) so

- [x] I think we should remind people in the contribution guides to `npm i` often.

---

Secondly, the pre-commit hook introduces extra indentation in YAML block e.g.
 ```yaml
 md5: df06d8d51e6483ed5a74d3979f8fe42e
 outs:
-- cache: true
-  md5: b8f4d5a78e55e88906d5f4aeaf43802e.dir
-  metric: false
-  path: pics
+  - cache: true
+    md5: b8f4d5a78e55e88906d5f4aeaf43802e.dir
+    metric: false
+    path: pics
 wdir: .
 ```
(from [.../commands-reference/add.md](/static/docs/commands-reference/add.md)) but those are just copy-paste of actual DVC-files so

- ~~[ ] I think those should not be changed by the linter.~~ No longer needed per https://github.com/iterative/dvc.org/pull/378#pullrequestreview-241516633

---

- [x] Last, we need to run `prettier` once we're sure how to and that it's behaving correctly on all .md files with
```console
$ npm run prettier-md
```

jorgeorpinel:
On question 1, from https://discordapp.com/channels/485586884165107732/565759186693128202/580164996198301696

> ...how will the configurations of pre-commit hook and running prettier from command line match(?)
> their version will always need to be in sync. Currently my prettier version is 1.17.1. So in my YAML, I changed it to 1.17.1 so they both match.

So should we then establish a strict version of prettier in the `package.json` file? It currently is` "^1.17.0"`

jorgeorpinel:
@algomaster99 would you like to briefly mention the option to use `husky` here? Thanks

algomaster99:
@jorgeorpinel 
```
"husky": {
  "hooks": {
    "pre-commit": "lint-staged" 
    },
},
"lint-staged": {
[...]

PR #368, opened 2019-05-21
Completes optimization guide related updates.

### LINKED ISSUE #265: add a user guide page about the optimization (advanced mode)

jorgeorpinel:
- Removes trailing white spaces per https://github.com/iterative/dvc.org/pull/345#pullrequestreview-239795703
- Updates `dvc unprotect` cmd ref.

Fixes #265

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,6 @@
   "devDependencies": {
     "babel-plugin-styled-components": "^1.5.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
-    "prettier": "^1.17.0"
+    "prettier": "~1.17.0"
   }
 }
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -119,8 +119,8 @@ details.)
   > `dvc unprotect` command to modify files safely.
 
   There are pros and cons to different link types. Refer to
-  [Cache File Linking](/docs/user-guide/cache-file-linking) for a full
-  explanation of each one.
+  [File link types](/docs/user-guide/cache-file-linking) for a full explanation
+  of each one.
 
 - `cache.slow_link_warning` - used to turn off the warnings about having a slow
   cache link type. These warnings are thrown by `dvc pull` and `dvc checkout`
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -64,9 +64,9 @@ This is the main section with the general config options:
   each stage in `dvc repro`. By default this behavior requires the use of option
   `-i` in that command. Accepts values `true` and `false`.
 
-- `core.analytics` - used to turn off [anonymized usage
-  statistics](/doc/user-guide/analytics). Accepts values `true` (default) and
-  `false`.
+- `core.analytics` - used to turn off
+  [anonymized usage statistics](/doc/user-guide/analytics). Accepts values
+  `true` (default) and `false`.
 
 ### remote
 
@@ -79,13 +79,15 @@ remote. See `dvc remote` for more information.
 
 The DVC cache is a hidden storage (by default located in the `.dvc/cache`
 directory) for files that are under DVC control, and their different versions.
-(See `dvc cache` and [DVC internal
-files](/doc/user-guide/dvc-files-and-directories) for more details.)
+(See `dvc cache` and
+[DVC internal files](/doc/user-guide/dvc-files-and-directories) for more
+details.)
 
 - `cache.dir` - set/unset cache directory location. A correct value must be
   either an absolute path or a path **relative to the config file location**.
   The default value is `cache`, which resolved relative to the default project
   config location results in `.dvc/cache`.
+
   > See also helper command `dvc cache dir` that properly transform paths
   > relative to the present working directory into relative to the project
   > config file.
@@ -116,15 +118,16 @@ files](/doc/user-guide/dvc-files-and-directories) for more details.)
   > See the `cache.protected` config option above and corresponding
   > `dvc unprotect` command to modify files safely.
 
-  There are pros and cons to different link types. Refer to [Cache File
-  Linking](/docs/user-guide/cache-file-linking) for a full explanation of each
-  one.
+  There are pros and cons to different link types. Refer to
+  [Cache File Linking](/docs/user-guide/cache-file-linking) for a full
+  explanation of each one.
 
 - `cache.slow_link_warning` - used to turn off the warnings about having a slow
   cache link type. These warnings are thrown by `dvc pull` and `dvc checkout`
   when linking files takes longer than usual, to remind them that there are
   faster cache link types available than the defaults (`reflink,copy` – see
   `cache.type`). Accepts values `true` and `false`.
+
   > These warnings are automatically turned off when `cache.type` is manually
   > set.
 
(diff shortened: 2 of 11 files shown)


static/docs/commands-reference/config.md
```diff
@@ -79,74 +79,82 @@ remote. See `dvc remote` for more information.
...
   > will corrupt the cache** if you modify tracked data files in the workspace.
   > See the `cache.protected` config option above and corresponding
-  > `dvc unprotect` command to modify files safely.  
-  There are pros and cons to different link types. Refer to [Cache File
-  Linking](/docs/user-guide/cache-file-linking) for a full explanation of each
-  one.
+  > `dvc unprotect` command to modify files safely.
+
+  There are pros and cons to different link types. Refer to
+  [Cache File Linking](/docs/user-guide/cache-file-linking) for a full
```
shcheklein: name is different now?

static/docs/user-guide/cache-file-linking.md
```diff
@@ -41,61 +40,73 @@ system.
...
-Each file linking option is further detailed below, in order of efficiency:
+File link type benefits summary:
+
+| `cache.type` | speed | space | safe |
+|---|---|---|---|---|
+| `reflink` | x | x | x |
+| `hardlink` | x | x | |
+| `symlink` | x | x | |
+| `copy` | | | x |
+
```
jorgeorpinel: FYI added the table above @shcheklein

static/docs/commands-reference/config.md
jorgeorpinel: Updated here and in other docs in 0556cd0

jorgeorpinel:
OK I think this PR is finished and will actually close #256 😃

static/docs/user-guide/contributing-documentation.md
```diff
@@ -78,22 +78,23 @@ $ source .env/bin/activate
...
 
 It's highly recommended to run the Node.js docs app locally to test significant
 changes to the docs before submitting them, and its very much needed in order to
 make changes to the docs JavaScript engine itself (rare). To do so, please
 follow the steps below:
 
 - Make sure you have the latest version of [Node.js](https://nodejs.org/en/)
-installed. Install dependencies by running the command `npm install`.
+  installed. Install and keep the dependencies up to date by running
+  `npm install` often.
```
jorgeorpinel: Note: I updated the sentence above per #363 

All other changes to this file are by linting it with `npx prettier`.

package.json
```diff
@@ -59,6 +59,6 @@
   "devDependencies": {
     "babel-plugin-styled-components": "^1.5.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
-    "prettier": "^1.17.0"
+    "prettier": "~1.17.0"
```
jorgeorpinel: Changed per https://github.com/iterative/dvc.org/issues/363#issuecomment-494557734

static/docs/commands-reference/config.md
```diff
@@ -79,74 +79,82 @@ remote. See `dvc remote` for more information.
...
   > will corrupt the cache** if you modify tracked data files in the workspace.
   > See the `cache.protected` config option above and corresponding
-  > `dvc unprotect` command to modify files safely.  
-  There are pros and cons to different link types. Refer to [Cache File
-  Linking](/docs/user-guide/cache-file-linking) for a full explanation of each
-  one.
+  > `dvc unprotect` command to modify files safely.
+
+  There are pros and cons to different link types. Refer to
+  [File link types](/docs/user-guide/cache-file-linking) for a full explanation
```
shcheklein: change the name or the link?

static/docs/user-guide/cache-file-linking.md
shcheklein: 👍 can we rename the `safe` column somehow? it feels like hardlinks and symlinks are really bad - unsafe :) while they just need extra measures.

static/docs/user-guide/cache-file-linking.md
```diff
@@ -41,61 +40,73 @@ system.
...
+   cache file intact.
+
+   > Unfortunately reflinks are currently supported on a limited number of file
+   > systems only (Linux: Btrfs, XFS, OCFS2; MacOS: APFS), but in the future
+   > they will be supported by the majority of file systems in use.
 
 2. **`hardlink`** - hard links are the most efficient way to link your data to
    cache if both your repo and your cache directory are located on the same file
-   system/drive.  
+   system/ drive.
```
shcheklein: remove an extra space here?

package.json
shcheklein: if this change fixes that ticket, please add it to the PR description - like Fixes #... so that we don't forget to close it. Thanks :)

static/docs/user-guide/contributing-documentation.md
ryokugyu: @shcheklein @jorgeorpinel is it needed to mention first run 
`npm install -g npm` first?
then run 'npm install'

jorgeorpinel: Having the latest Node.js installed is mentioned in the previous paragraph.

package.json
jorgeorpinel: It doesn't completely fix it, these changes only address the first checkbox from that issue.

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: How about "editable" ?

static/docs/commands-reference/config.md
jorgeorpinel: Do you mean "File link types" or `cache-file-linking` ?

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: Maybe I'll change the phrase completely to "same partition or storage device" to be more correct.

static/docs/user-guide/cache-file-linking.md
```diff
@@ -41,61 +40,73 @@ system.
...
+
+File link type benefits summary:
+
+| `cache.type` | speed | space | editable |
+| ------------ | ----- | ----- | -------- |
+| `reflink`    | x     | x     | x        |
+| `hardlink`   | x     | x     |          |
+| `symlink`    | x     | x     |          |
+| `copy`       |       |       | x        |
+
```
jorgeorpinel: How about now @shcheklein ? Please notice the use of the same words in the paragraph above in **bold**.

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: Done.

static/docs/commands-reference/config.md
shcheklein: I mean - should we do something like [File link types](/doc/user-guide/cache-file-linking#file-link-types-for-the-dvc-cache) or keep it as is - I'm not sure to be honest. Even though neither link, nor title match the content of the link it does not necessary mean that we should change it. So, up to you to decide on this - thus the ? mark at my comment :)

static/docs/user-guide/cache-file-linking.md
shcheklein: I don't know ... it still creates an effect that they are not editable at all, don't you think? may be we should just use a phrase - not nice, but at least will be easier to understand for users

static/docs/commands-reference/config.md
jorgeorpinel: I think adding the #anchor is best. I didn't do it because it goes over 80 columns in the .md file but that's OK right? I'll do the change in all the docs where it applies (its a few).

jorgeorpinel: Done

static/docs/user-guide/cache-file-linking.md
jorgeorpinel: OK changed in paragraph above and table to "no protected mode"

shcheklein:
thanks @jorgeorpinel 🎉

Issue #370, opened 2019-05-22
pipeline: show: dot: check that user doesn't accidentally write to dvc file


TomHortons:
Issue: https://github.com/iterative/dvc/issues/1743

Is this still opened?

shcheklein:
Closed by #371

PR #371, opened 2019-05-22
pipeline: show: dot: check that user doesn't accidentally write to dvc file

TomHortons:
Related to https://github.com/iterative/dvc.org/issues/370


shcheklein:
thanks, @TomHortons 🎉

PR #378, opened 2019-05-24
Regular updates to docs

### LINKED ISSUE #363: pre-commit hook prettier (vs npx prettier) questions

jorgeorpinel:
Given different conversations I've had or noticed on Discord and GH

Also fixes #363

### DIFF
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -41,12 +41,12 @@ references the DVC cache entry using the checksum. See
 [DVC File Format](/doc/user-guide/dvc-file-format) for the detailed description
 of the DVC _metafile_ format.
 
-By default DVC tries using reflinks (See [Cache File
-Linking](/docs/user-guide/cache-file-linking)) to avoid copying any file
-contents and to optimize DVC file operations for large files. DVC also supports
-other link types for use on file systems without `reflink` support, but they
-have to be specified manually. Refer to the `cache.type` config option in `dvc
-config cache` for more information.
+By default DVC tries using reflinks (See
+[Cache File Linking](/docs/user-guide/cache-file-linking)) to avoid copying any
+file contents and to optimize DVC file operations for large files. DVC also
+supports other link types for use on file systems without `reflink` support, but
+they have to be specified manually. Refer to the `cache.type` config option in
+`dvc config cache` for more information.
 
 A `dvc add` target can be an individual file or a directory. There are two ways
 to work with directory hierarchies with `dvc add`.
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -119,6 +119,7 @@ $ tree
 
 Let's check the `data.xml.dvc` file inside:
 
+<!-- prettier-ignore -->
 ```yaml
 md5: aae37d74224b05178153acd94e15956b
 outs:
@@ -181,6 +182,7 @@ are all added to the DVC cache. DVC prints a message to that effect, saying that
 `md5` values are computed for each directory. A DVC file is generated for the
 top-level directory, and it contains this:
 
+<!-- prettier-ignore -->
 ```yaml
 md5: df06d8d51e6483ed5a74d3979f8fe42e
 outs:
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -119,14 +119,13 @@ $ tree
 
 Let's check the `data.xml.dvc` file inside:
 
-<!-- prettier-ignore -->
 ```yaml
 md5: aae37d74224b05178153acd94e15956b
 outs:
-- cache: true
-  md5: d8acabbfd4ee51c95da5d7628c7ef74b
-  metric: false
-  path: data.xml.jpg
+  - cache: true
+    md5: d8acabbfd4ee51c95da5d7628c7ef74b
+    metric: false
+    path: data.xml.jpg
 ```
 
 This is a standard DVC stage file with only an `outs` entry. The checksum should
@@ -182,14 +181,13 @@ are all added to the DVC cache. DVC prints a message to that effect, saying that
 `md5` values are computed for each directory. A DVC file is generated for the
 top-level directory, and it contains this:
 
-<!-- prettier-ignore -->
 ```yaml
 md5: df06d8d51e6483ed5a74d3979f8fe42e
 outs:
-- cache: true
-  md5: b8f4d5a78e55e88906d5f4aeaf43802e.dir
-  metric: false
-  path: pics
+  - cache: true
+    md5: b8f4d5a78e55e88906d5f4aeaf43802e.dir
+    metric: false
+    path: pics
 wdir: .
 ```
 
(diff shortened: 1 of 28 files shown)


static/docs/commands-reference/add.md
```diff
@@ -119,6 +119,7 @@ $ tree
 
 Let's check the `data.xml.dvc` file inside:
 
+<!-- prettier-ignore -->
```
shcheklein: why do we need to ignore this?

static/docs/user-guide/contributing.md
```diff
@@ -281,9 +281,13 @@ To run DVC from its Git repository you need to setup your environment:
 
 ## Code style guidelines
 
-We are using [PEP8](https://www.python.org/dev/peps/pep-0008/?) and are checking
+We are using [PEP8](https://www.python.org/dev/peps/pep-0008/?) and checking
 that our code is formatted with [black](https://github.com/ambv/black).
 
+For [docstrings](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring),
```
shcheklein: probably not only docstrings? cc @efiop

static/docs/commands-reference/add.md
jorgeorpinel: Otherwise the YAML blocks (DVC-file samples) get indented by prettier. See the 2nd point in the description of #363

static/docs/user-guide/contributing.md
jorgeorpinel: @mroutis mentioned the docstrings to me recently but yes maybe more aspects? Please lmk and I'll update the doc. Thanks

static/docs/commands-reference/add.md
shcheklein: I'm not sure this is a problem - looks like this is a standard prettier indentation for YAML: 

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBadAdKBLKA3CbMOAAjKTJIGYAWADhqqwBMBDeSiygJgAZeAjKkGpuTKACNsAG2moYELmQBk2ZvwFZKJAObY8CbRQDCACwBO2AM5bKAM1YBbGQE9yJACIBXR63M2obVZmZnM4KyskW20yaVxwigAfaJjtGgBWOhIAdVZpAGtfQI9zADoU1LIAZS9sDgBibgBObgrKMDq3SooAJQgXPJIAeVZ8trIrGHZSGIoAWQBJcZIABwhJwZIKel4aADYsK1NsFflFEgAqNQ0sFfMIZi8wGCjAylQSFLarfK9K9wAQgAZKhNGgeNoAfhIAEcvKxYJ1tG1tm1mOEwJYVjBsNAtiQAawfnAYBI8tIoatLMRKiiSBleKV+G0WW8yB9vr9-u4yMCaIw6AAJNpwhE4mBdIxkTRsyjoqyYk446BKSg1FZwcwkQUQCArNp3IgzbpkMQtJm8VmpLBTAAePO46QEpRorVgECm0nx-MYpXSbsgjkcCBgkRIAD4UkDpiRWHZ4OYoLrAtYSBJwjByrLCWBfisSJBYKxniRUwA5BFgFwpAEyaRWYNagAC1CodFQjDbWZAABoQHrlVArMhQH57gB3AAKfgQw5QrAIal7IAk5mL+RJVRWxdwOmQMHMXjgfdMMEc0myx3gVm3xCqs7q+k6yHAEWXuCsmpgk7XOl8yAcetjxAAArKxbQBNdc03Jw4CBeIALyT8+zA20ql3aQ4AARS8D04EQoC+23fxNRfAZz2XQ1YGyNQYFMZA6F4Ij7k-bI1xWF87nCTUDGXOE8J-PU5xAIlUCgOA4HRZhlzCOFsDCH9WD-VgCOQkBP2cfdD2AqwMOw3D4FU4CpgkGjmDo5BuD7A9WBkXdjAgIMVJQJNxOXLxPwAFVYCQ50Az8AF8AqAA

jorgeorpinel: Yes its standard prettier indentation and can't be configured otherwise, only ignored.

I thought we wouldn't want the YAML DVC-file blocks to be changed by prettier since they are just textual copy-pastes of how the real files look. prettier changes them in a way that DVC does never produce them.

BTW we may have similar problems with other code blocks, I just ran `npm run prettier-md` and I'm examining each change... Will report

shcheklein: I think it's fine. Don't has to be exactly the same including indents. It's easier to let it format this - so that we always have the same standard representation everywhere.

jorgeorpinel: OK will un-do then 👍 

p.s. the rest of the changes prettier does are fine, but I had to manually fix https://github.com/jorgeorpinel/dvc.org/blob/master/static/docs/changelog/0.35.md which is all messed up apparently 😬

static/docs/get-started/connect-code-and-data.md
```diff
@@ -102,6 +102,7 @@ This is how the result should look like now:
 
 This is how `prepare.dvc` looks like internally:
 
+<!-- prettier-ignore -->
```
shcheklein: remove?

shcheklein: all of them below as well?

static/docs/commands-reference/add.md
jorgeorpinel: Un-done. And prettier-md script ran on everything.

static/docs/get-started/connect-code-and-data.md
jorgeorpinel: All removed now. Also ran prettier-md on everything (notice the yaml code blocks have been formatted).

static/docs/user-guide/contributing.md
efiop: @shcheklein We don't follow google coding style in full, so It is hard to recommend it here, since black will probably reformat it anyway. Let's start with this docstring reference, we do use those.

shcheklein: 👍

PR #382, opened 2019-05-25
embed the videos

### LINKED ISSUE #349: embed the video on the 'doc' page itself
Stumbling upon this page https://dvc.org/doc/understanding-dvc/resources.
dvc

The links look like a static image. And, also upon clicking on the image, we are redirecting to YouTube page. 

We can embed the video on the DVC page itself, rather than opening a new target. 
It will impact user browsing experience.

kurianbenoy:
FIx #349

### DIFF
--- a/static/docs/understanding-dvc/resources.md
+++ b/static/docs/understanding-dvc/resources.md
@@ -4,17 +4,23 @@
 
 **A fun video about Reproducibility in #MachineLearning using DVC**
 
-[![Image Machine Learning Model and Dataset versioning
-practises](https://img.youtube.com/vi/jkfh2PM5Sz8/0.jpg)](https://www.youtube.com/watch?v=jkfh2PM5Sz8)
+<iframe width="560" height="315" src="https://www.youtube.com/embed/4h6I9_xeYA4"
+frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope;
+picture-in-picture" allowfullscreen></iframe>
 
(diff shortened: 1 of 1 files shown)


shcheklein:
thanks @kurianbenoy !

ryokugyu:
@kurianbenoy good job there!
But, in the articles section, whenever I click on some link. I opens a new page on the same tab.
imo, the page must open in another tab for better user experience. As, after reading one article user needs to check some other article.He/She must have to go back and look other articles.

Moving forward, in case of user wants to compare two articles/ read them side by side. He/She again have to perform 2-3 steps. It will make the user experience not so good.

kurianbenoy:
That's a great idea @ryokugyu, even though I usually open in a new tab with a cool keyboard hack. 
Should we use:
`[link](url){:target="_blank"}`
or 
HTML for it
`<a href="http://example.com/" target="_blank">Hello, world!</a>`

shcheklein:
@kurianbenoy Markdown syntax, please 👍

ryokugyu:
{:target="_blank"} is working? @kurianbenoy 
I tried this, but it doesn't help.

shcheklein:
@ryokugyu @kurianbenoy this issue should be solved with JS code if `{:target="_blank"}` is not supported. Check the documentation for the react markdown plugin. It has `linkTarget` callback to determine what target we would like to use with a particular link. For every external link we should be using `_blank`.

PR #384, opened 2019-05-26
_blank target  in resources.md

ryokugyu:
[link](url){:target="_blank"} didn't worked out.
So, sticked to the basic html syntax.


ryokugyu:
> [link](url){:target="_blank"}
> So, sticked to the basic html syntax.

PR #389, opened 2019-05-30
Transform cursor to pointer when hovering over a collapsible

### LINKED ISSUE #362: css: make expandable <detail> sections `cursor: pointer` on `:hover`
> May also fix #120 

For example 
image
from https://dvc.org/doc/get-started/add-files

They currently have normal mouse cursors but could use 👆

algomaster99:
### Demo
hover

Fixes #362

### DIFF
--- a/src/Documentation/Markdown/Markdown.js
+++ b/src/Documentation/Markdown/Markdown.js
@@ -213,6 +213,7 @@ const Content = styled.article`
     display: block;
     position: relative;
     opacity: 0.9;
+    cursor: pointer;
 
     &:after {
       position: absolute;


jorgeorpinel:
What do you use to create screenshot gifs @algomaster99 ? Those are useful 😃

algomaster99:
@jorgeorpinel I use [Simple Screen Recorder](https://www.maartenbaert.be/simplescreenrecorder/) for recording screens but its output is `.mp4`. You can easily convert it into `.gif` using this [link](https://ezgif.com/video-to-gif).

jorgeorpinel:
I see. You may want to try cmd+shift+5 if you're on Mac in that case. ezgif.com/video-to-gif is nice too but I like to convert videos with https://handbrake.fr/ (downloaded)

PR #397, opened 2019-06-02
Fix #394

### LINKED ISSUE #394: mention in tutorial about ls -i meaning
```
$ ls -i data/Posts.xml.zip
78483929 data/Posts.xml.zip

$ ls -i .dvc/cache/ec/
78483929 88519f8465218abb23ce0e0e8b1384
```

In [tutorial page](https://dvc.org/doc/tutorial/define-ml-pipeline), where we mention this code block, we can tell that it prints the index no of each file, and inode in cache and data/ is same. 

Should I mention @shcheklein this as a comment in code block, or in documentation?

kurianbenoy:
#394

### DIFF
--- a/static/docs/commands-reference/repro.md
+++ b/static/docs/commands-reference/repro.md
@@ -8,7 +8,7 @@ dependencies to find only those that have to be rerun.
 
 ```usage
 usage: dvc repro [-h] [-q | -v]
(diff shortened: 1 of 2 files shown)


kurianbenoy:
Oh wait, my other commit is also coming up, sorry about

PR #402, opened 2019-06-03
Add contributing note for zsh users

pared:


static/docs/user-guide/contributing.md
```diff
@@ -48,6 +48,9 @@ contributing!
   $ source .env/bin/activate
   $ pip install -e .[all,tests]
   ```
+**NOTE** When using zsh shell, remember to escape square brackets.
```
shcheklein: Ugh, we have in quite a lot of other places - git grep "\[s3\]" . Even in DVC repo itself. Should ZSH users be aware about specifics of their shell without us? It's not scalable to put notes like this everywhere.

Also, Please, use `>`, without **Note**

shcheklein:
Closing in favor of https://github.com/iterative/dvc.org/issues/414

PR #403, opened 2019-06-03
Fix configuration in ci tests

algomaster99:
- There is no `yarn.lock` file since we don't use yarn so `yarn install` will fail. `npm install` will work.
- Since there are no tests, `npm run test`/`yarn test` would fail.


shcheklein:
Thanks, @algomaster99 ! This yaml file was an autogenerated default template. The idea that we need to fix/adjust it now to run linter, prettier, etc to do all the checks.

PR #408, opened 2019-06-05
Fix #404: Refactor root and improve its style.

### LINKED ISSUE #404: cmd ref: review `root` ref to be aligned with other commands
See https://dvc.org/doc/commands-reference/root.

i.

- [x] The layout of this command reference is outdated.

- [x] It should be the same as for the [cache](https://dvc.org/doc/commands-reference/cache) for example.

ii.

- [x] Description should include a better motivation for this command - mostly ease of generating stage files no matter where you run the `dvc run`.

- [x] Examples should be actionable and meaningful. It's should be easy to run commands to see the results.

iii.

- ~~A new example about shell scripts to generate dvc.yaml files is needed to complete this update. See https://github.com/iterative/dvc.org/pull/1637#discussion_r463117077~~ Now part of #2076

vibhor98:
Fixes #404 to refactor `root` command reference guide and improve its style to be aligned with other commands.

### DIFF
--- a/static/docs/commands-reference/root.md
+++ b/static/docs/commands-reference/root.md
@@ -2,11 +2,17 @@
 
 Returns relative path to project's directory.
 
-Can be used to build a path to a dependency, command, or output.
+## Synopsis
 
 ```usage
 usage: dvc root [-h] [-q] [-v]
 ```
+
+## Description
+
+While modifying code in different sub-directories, there can be cases when
+developers require relative path to the DVC's root directory. So, this command
+can be used to build a path to a dependency, command, or output.  
 
 ## Options
 
--- a/static/docs/commands-reference/root.md
+++ b/static/docs/commands-reference/root.md
@@ -10,9 +10,10 @@ usage: dvc root [-h] [-q] [-v]
 
 ## Description
 
-While modifying code in different sub-directories, there can be cases when
-developers require relative path to the DVC's root directory. So, this command
-can be used to build a path to a dependency, command, or output.  
+While in project's sub-directory, sometimes developers may want to refer some
+file belonging to another directory. This command returns relative path to the
+DVC project's root directory from the current working directory. So, this
+command can be used to build a path to a dependency file, command, or output.  
 
 ## Options
 


vibhor98:
Hi @shcheklein, kindly review! Thanks!

static/docs/commands-reference/root.md
```diff
@@ -2,12 +2,18 @@
...
-Can be used to build a path to a dependency, command, or output.
+## Synopsis
 
 ```usage
 usage: dvc root [-h] [-q] [-v]
 ```
 
+## Description
+
+While modifying code in different sub-directories, there can be cases when
```
shcheklein: I'm not sure I understand this to be honest, could you elaborate please? Even better can you ask a question on our Discord chat to get more information and insights?

vibhor98: Ok, I'll ask on Discord. Thanks!

vibhor98: Done, PTAL!

vibhor98:
Hi @shcheklein, I've improved the motivation after referring the original feature request  [#779](https://github.com/iterative/dvc/issues/779). Also, I installed dvc on my Mac and checked the examples. They are easily runnable. Please review!

static/docs/commands-reference/root.md
```diff
@@ -2,12 +2,19 @@
...
 ```usage
 usage: dvc root [-h] [-q] [-v]
 ```
 
+## Description
+
+While in project's sub-directory, sometimes developers may want to refer some
+file belonging to another directory. This command returns relative path to the
+DVC project's root directory from the current working directory. So, this
+command can be used to build a path to a dependency file, command, or output.
```
shcheklein: So, am I correct that this command was not built to be used inside DVC-files? (cc @efiop). The primary use case is to simplify developers' life when they run `dvc run`. And make `dvc run` executable effectively from any subdirectory. So, primarily it can be useful if I use a script that generates some stage files?

shcheklein:
Looks great, @vibhor98 ! A quick question to discuss, mostly to @efiop :) Just trying to better understand the motivation for the command, in what user scenarios it gives value.

vibhor98:
Hi @shcheklein, shall I take up other issue while waiting for @efiop reply? Thanks!

static/docs/commands-reference/root.md
efiop: @shcheklein Correct. But it can also be used in dvc files as well, just by using it in the command itself, so that it is evaluated each time the command is ran.

efiop:
Guys, sorry for the delay.

jeremydesroches:
@jorgeorpinel I’ll work on closing this out.

PR #417, opened 2019-06-07
WIP: `dvc pkg` cmd ref (and other misc updates)

jorgeorpinel:
Review app may be at https://dvc-org-pr-417.herokuapp.com/ (enable at [CI pipeline](https://dashboard.heroku.com/pipelines/7a105706-a0cc-4e66-a022-2ea44c7cf238))

---
Closes #385

Issue #321, opened 2019-05-13
standard way to refer to DVC-files and other terms


jorgeorpinel:
There 's several ways in which the docs refer to **DVC-files** (meaning the stage files created by `dvc add`, `dvc run`, etc, with the `.dvc` file extension).

Standardize everything to "DVC-file(s)" throughout the docs EXCEPT if the phrase is used for OTHER DVC-related files (not stage files) i.e. in "[DVC Files and Directories](https://dvc.org/doc/user-guide/dvc-files-and-directories)":

- [x] "DVC file" and plural "DVC files". <-- Do NOT use for stage files

- [x] "DVC-file" and its plural "DVC-files". <-- Use this one

- [x]  "`.dvc` file(s)" or "`.dvc` file extension" <-- Avoid except in single doc where DVC-files are described
- [x] "stage file(s)" <-- Avoid except when in the context of data pipelines
   > ^ Partially extracted this case to #423
- [x] DVC "metafile(s)"/"meta file(s)" <-- Avoid except in single doc where DVC-files are described

~~UPDATE: Other interesting phrases to review are:~~ See #432

jorgeorpinel:
Side note: I'm already changing `.dvc` for `.dvc/` (see https://github.com/jorgeorpinel/dvc.org/commit/cda4582e6d4479fe96cc74a6224c5b2e5670f0f5) when it refers to the directory to help ease "DVC" naming confusions throughout the docs.

shcheklein:
Sounds good, @jorgeorpinel . I think all these small things worth mentioning in some section of the Contributing guide? What do you think? Something like a "terminology guideline" (is there a better name for this)?

jorgeorpinel:
Good idea. Maybe "terminology conventions" ...

OR it could be a new doc called **Glossary** and in the docs contribute guidelines we just mention to "utilize the terminology listed in the glossary(linked) consistently". Please assign me here to confirm (:

jorgeorpinel:
p.s. I just noted the term "stage file(s)" is also widely used in the docs for this one.

jorgeorpinel:
Sorry, 2 of the checkboxes in the original description weren't completed with the PR that closed this. Reopened this issue and opened PR #433

PR #418, opened 2019-06-07
regular update to docs (early June)

jorgeorpinel:
Extracted from changes in #417 (unrelated to `pkg`)


jorgeorpinel:
@shcheklein here are some changes that I extracted from the now separate PR about `dvc pkg` commands. Please check when you can. I have more changes related to #321 but it's a looot of files so it'd be best if this PR is merged before I push those. Thanks

shcheklein:
Thanks, @jorgeorpinel !

PR #428, opened 2019-06-08
Elaborate working of downstream flag in command `dvc repro`

### LINKED ISSUE #271: repro: improve information about --downstream

algomaster99:
These changes show the working of command `dvc repro --downstream`

Fixes #271

### DIFF
--- a/static/docs/commands-reference/repro.md
+++ b/static/docs/commands-reference/repro.md
@@ -188,6 +188,8 @@ Saving information to 'Dvcfile'.
 You can check now that `Dvcfile` and `count.txt` have been updated with the new
 information, new `md5` checksums and a new result respectively.
 
+## Examples: Downstream
+
 There is also an option which allows one to reproduce results from a specific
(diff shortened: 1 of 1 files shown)


static/docs/commands-reference/repro.md
```diff
@@ -186,3 +187,42 @@ Saving information to 'Dvcfile'.
 
 You can check now that `Dvcfile` and `count.txt` have been updated with the new
 information, new `md5` checksums and a new result respectively.
+
+There is also an option which allows one to reproduce results from a specific
```
shcheklein: let's create a separate section `## Examples: Downstream` or something like that. Check `dvc add` to see an example of this split. It easier to navigate, read, and make sense of example.

algomaster99:
@shcheklein I thought this example would be great since it is not from the data science perspective so anyone can understand it. Also, we are just reusing the example above it so users won't have to switch tabs to refer to the example in `dvc checkout`.

PR #434, opened 2019-06-12
Modify installation command to make it work on zsh as well

### LINKED ISSUE #414: use single quotes for `pip install` commands with square brackets
Related to #402 and #412 .

`pip install -e dvc[all]` fails when ZSH is being used

`pip install -e 'dvc[all]'` should be working everywhere sh, bash, zsh.

Check that it's actually working and update our docs.

algomaster99:
Fixes #414


algomaster99:
@shcheklein I did `git grep 'pip install -e'` instead.
Will this `pip install -U dvc[s3]` also require changes?

shcheklein:
@algomaster99 the problem is that square brackets have a special meaning on zsh, so any square brackets stuff should be put into quotes

Issue #103, opened 2018-10-30
guide: using NFS as a remote storage


efiop:
I.e. mount it to some dir and specify it as a local dir repository for dvc.

Related to https://github.com/iterative/dvc.org/issues/4303

dashohoxha:
I have generalized this as "Mounted DVC Storage", which includes not only NFS, but also SSHFS, Samba, or any other data storage that can be network-mounted on a local directory.

I have also built an interactive example that explains how to do it for SSHFS (since this is easier to set up for an interactive example): https://katacoda.com/dvc/courses/examples/mounted-storage

jorgeorpinel:
@efiop did you mean in https://dvc.org/doc/command-reference/remote/add though? Or somewhere else.

efiop:
@jorgeorpinel I guess a note there would also be pretty nice as a way of using any storage that you can mount to your local fs.

jorgeorpinel:
OK. So while #784 greatly simplifies this issue, it won't completely solve it. I'm updating the PR's description to remove the link. Or if you prefer to also add the note Ruslan suggests Dashamir, please reestablish the link to close this issue automatically from your PR. 🙂

iesahin:
I read the comments in #784. Is there anything that should be done here?

shcheklein:
It feels that we don't have a perfect place at the moment to put this note. I think there should be something about remotes in the User Guide in the first place. Starting from basic concepts, expanding into a proper section, where we can put a note in the introductory part about "Unsupported directly remotes" with a solutions like sshfs, mounted NFS, etc. For now, we can probably create a section and a note in the existing `remote add`, `remote modify`, `remote index` with the same meaning.

jorgeorpinel:
QQ @dberenbaum: This has a low priority now but https://github.com/iterative/dvc.org/issues/4303 (part of this one) a high priority. 🤔 Does that mean you want to focus on the latter without addressing the former (this one) ?

dberenbaum:
@jorgeorpinel In https://dvc.org/doc/user-guide/data-management/remote-storage#file-systems-local-remotes, you added:

> You can also use system directories, mounted drives, network resources e.g. network-attached storage (NAS), and other external devices as storage. 

Is that enough to close this one? Do we need anything else for this issue specifically?

jorgeorpinel:
[...]

Issue #429, opened 2019-06-10
cases: update Shared Dev Server with an optimization link


shcheklein:
UPDATE: See summary in https://github.com/iterative/dvc.org/issues/429#issuecomment-553118884 below.

More context here - https://discuss.dvc.org/t/share-nas-data-in-server/180/4?u=shcheklein

jorgeorpinel:
@shcheklein what optimization link? Sorry, I read the discussion not sure what's missing from the shared dev server use case. Do you mean this part mostly?

```console
$ dvc config cache.type "reflink,symlink,hardlink,copy" # to enable symlinks to avoid copying
$ dvc config cache.protected true # to make links RO so that we you don't corrupt them accidentally
```

(Perhaps in an expandable details section?)

shcheklein:
@jorgeorpinel yes, mostly enabling different link types. It's a very important part of that case. Without enabling symlinks and understanding consequences (needed protection) that case does not make much sense. I would not make it expandable, but try to keep it within a paragraph and link to other docs.

jorgeorpinel:
We could even rethink the name of this use case, think about the actual use case.

PR #442, opened 2019-06-19
[WIP] shared storage on nfs article introduced

ryokugyu:


ryokugyu:
Fix #103 and #429

static/docs/use-cases/multiple-data-scientists-on-a-single-machine.md
```diff
@@ -1,5 +1,3 @@
-# Shared Development Server
```
shcheklein: why removing this?

static/docs/use-cases/shared-storage-on-nfs.md
```diff
@@ -0,0 +1,254 @@
...
+> With large data files it is better to set the cache directory to NFS. Not only
+> just it will cache the data faster but also version the data. Thus you would
+> avoid copying large files from you NFS server to the machine.
+
+Lets configure NFS server and client to see how DVC can help to make the
+workflow faster.
+
+<details>
+
+# Setting up the NFS server
```
shcheklein: Is there a link we can share instead? Setting up NFS server/client is not DVC-specific and I would avoid including this into the doc.

static/docs/use-cases/shared-storage-on-nfs.md
```diff
@@ -0,0 +1,254 @@
...
+get the results. It became necessary that data is accessible to everyone and
+every team member has a same updated dataset. For this example, we will be using
+NFS (Network File System) for storing and sharing files on the network.
+
+With DVC, we need not to copy the dataset on our local machine everytime when
+new data is added to dataset. We can set the `cache directory` on NFS server.
+The cached data will be present in the NFS server, which in turn will be fast to
+access and process requests faster.
+
+> With large data files it is better to set the cache directory to NFS. Not only
```
shcheklein: It should be a regular paragraph that explains _why_ it's beneficial to use DVC in this case. Namely, for people their workspace looks the same way as they are alone, DVC "magically" manages links from the workspace to cache. People can work with multiple versions of a datasets without affecting each other and without creating duplicates (like a full copy of a folder with 1M images)

static/docs/use-cases/shared-storage-on-nfs.md
```diff
@@ -0,0 +1,254 @@
...
+> just it will cache the data faster but also version the data. Thus you would
+> avoid copying large files from you NFS server to the machine.
+
+Lets configure NFS server and client to see how DVC can help to make the
```
shcheklein: Again, not sure we need this. I would assume that it's configured already.

static/docs/use-cases/shared-storage-on-nfs.md
```diff
@@ -0,0 +1,254 @@
...
+you and your colleagues are members of the same group (e.g. 'users') and that
+your shared directory is owned by that group and has respective permissions.
+
+Let's create a mount point of client side.
+
+```dvc
+mkdir -p /mnt/dataset/project1_data
+```
+
+### Configure Cache
```
shcheklein: we def need more explanation behind the commands below

static/docs/use-cases/shared-storage-on-nfs.md
```diff
@@ -0,0 +1,254 @@
...
+location.
+
+> It will be beneficial to have a shared cache location on the server. It avoid
+> copying large files from shared server to the local machine.
+
+After configuring NFS on both server and client side. Let's create an export
+directory on server side where all data will be stored.
+
+```dvc
+mkdir -p /project1_data
```
shcheklein: no need to copy the name from discuss - just `/storage` should be enough

ryokugyu:
@shcheklein please have a look at the changes and let me know any other changes required.

PR #449, opened 2019-06-23
Creating a bucket is necessary before adding it to remote storage

### LINKED ISSUE #415: missing information about s3 usage
Docs don't explicitly say that when one is trying to use s3 or s3 compatible storage there should already exist bucket.

Then, if you would try to push data to s3 and bucket wouldn't exist, you would get the following:
```
Preparing to upload data to 's3://my/remote/'
Preparing to collect status from s3://my/remote/
[############                  ] 40% Collecting information
ERROR: failed to push data to the cloud - An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: Unknown
``` 

So I propose to explicitly say in docs that you should already have existing bucket to be able to push to s3.
Following docs are related to s3 usage:
https://dvc.org/doc/get-started/configure
https://dvc.org/doc/commands-reference/remote-add
Related discussion in ods:
https://opendatascience.slack.com/archives/CGGLZJ119/p1559831715009900

Naba7:
Fix #415 
I just added the information to create a bucket with the links to it in commands-reference/config and commands-reference/remote .

### DIFF
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -181,9 +181,10 @@ $ dvc config core.loglevel debug
 
 Add an S3 remote and set it as the project default:
 
-Before,adding a new remote be sure to login into AWS services and follow
-instructions at [Create a Bucket](https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html)
-to create your bucket. 
+> **Note!** Before,adding a new remote be sure to login into AWS services and
+> follow instructions at
+> [Create a Bucket](https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html)
+> to create your bucket.
 
 ```dvc
 $ dvc remote add myremote s3://bucket/path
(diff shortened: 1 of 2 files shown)


static/docs/commands-reference/config.md
```diff
@@ -181,6 +181,10 @@ $ dvc config core.loglevel debug
 
 Add an S3 remote and set it as the project default:
 
+Before,adding a new remote be sure to login into AWS services and follow
```
shcheklein: use `>` before each line to make it a "note"

Naba7:
> have you run the pre-commit hook? or prettier manually? I think links should be on a separate line.

No,I hadn't run pre-commit hook. Yes, I ran it just now, and I am changing the link to a separate line.

static/docs/commands-reference/config.md
Naba7: ok

static/docs/commands-reference/config.md
```diff
@@ -181,6 +181,11 @@ $ dvc config core.loglevel debug
 
 Add an S3 remote and set it as the project default:
 
+> **Note!** Before,adding a new remote be sure to login into AWS services and
```
shcheklein: Before,adding -> Before adding

Naba7: fixed it

shcheklein:
@Naba7 thanks 🎉 !

PR #456, opened 2019-06-25
docs: regular update (late June) [WIP]

### LINKED ISSUE #425: user-guide/get-started: reorganize "Understanding DVC" section
Files currently located in `/static/docs/understanding-dvc/`

Remove all the **Understanding DVC** docs and just rescue all that is unique and valuable into the get started and user guide docs.

UPDATES:
- See example actions in https://github.com/iterative/dvc.org/issues/425#issuecomment-500141659
- See deeper discussion (still example based) in https://github.com/iterative/dvc.org/issues/425#issuecomment-633453572

jorgeorpinel:
Misc. needed updates/improvements to several docs.

### DIFF
--- a/static/docs/commands-reference/pipeline.md
+++ b/static/docs/commands-reference/pipeline.md
@@ -18,9 +18,10 @@ positional arguments:
 ## Description
 
 A data pipeline, in general, is a chain of commands that process data files. It
-produces intermediate data and a final result. Machine Learning (ML) pipelines
-typically start a with large raw datasets, include featurization and training
-intermediate stages, and produce a final model, as well as certain metrics.
+produces intermediate data and a final result. For example, Machine Learning
+(ML) pipelines typically start a with large raw datasets, include featurization
+and training intermediate stages, and produce a final model, as well as certain
+metrics.
 
 In DVC, pipeline stage files and commands, their data I/O, interdependencies,
 and results (intermediate or final) are defined with `dvc add` and `dvc run`,
--- a/static/docs/commands-reference/pipeline.md
+++ b/static/docs/commands-reference/pipeline.md
@@ -1,6 +1,6 @@
 # pipeline
 
-A set of commands to manage pipelines:
+A set of commands to manage [pipelines](/doc/get-started/pipeline):
 [show](/doc/commands-reference/pipeline-show) - visualize or
 [list](/doc/commands-reference/pipeline-list) - list existing pipelines.
 
@@ -17,7 +17,18 @@ positional arguments:
 
 ## Description
 
-Manage pipeline.
+A data pipeline, in general, is a chain of commands that process data files. It
+produces intermediate data and a final result. Machine Learning (ML) pipelines
+typically start a with large raw datasets, include featurization and training
+intermediate stages, and produce a final model, as well as certain metrics.
+
+In DVC, pipeline stage files and commands, their data I/O, interdependencies,
+and results (intermediate or final) are defined with `dvc add` and `dvc run`,
+among other commands. This allows us to form one or more pipelines of stages
+connected by their dependencies and outputs.
+
+`dvc pipeline` commands help users display the existing project pipelines in
+different ways.
 
 ## Options
 
--- a/static/docs/commands-reference/repro.md
+++ b/static/docs/commands-reference/repro.md
@@ -57,6 +57,11 @@ specified), and updates stage files with the new checksum information.
   reproduced as part of the pipeline in the parent directory, or as an
   independent unit.
 
+- `-R`, `--recursive` - `targets` is expected to contain directory path(s).
+  Determines the stages to reproduce by searching each target directory and its
+  subdirectories for DVC-files to inspect. Along with providing `targets`, this
+  is another way to limit the scope of DVC-files to reproduce.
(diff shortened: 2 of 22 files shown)


static/docs/commands-reference/repro.md
```diff
@@ -57,6 +57,11 @@ specified), and updates stage files with the new checksum information.
   reproduced as part of the pipeline in the parent directory, or as an
   independent unit.
 
+- `-R`, `--recursive` - `targets` is expected to contain directory path(s).
```
shcheklein: it actually accepts files as well now, just to simplify and make the workflow flexible

static/docs/commands-reference/repro.md
```diff
@@ -57,6 +57,11 @@ specified), and updates stage files with the new checksum information.
...
+  Determines the stages to reproduce by searching each target directory and its
+  subdirectories for DVC-files to inspect. Along with providing `targets`, this
```
shcheklein: the last sentence sounds strange. Should we change it to something like: `Along with --with-deps and other similar options it is a way to set the scope (DVC-files) to reproduce`?

static/docs/commands-reference/pipeline.md
```diff
@@ -17,7 +17,18 @@ positional arguments:
 
 ## Description
 
-Manage pipeline.
+A data pipeline, in general, is a chain of commands that process data files. It
+produces intermediate data and a final result. Machine Learning (ML) pipelines
```
shcheklein: `For example, Machine Learning (ML) pipelines ...`

static/docs/commands-reference/run.md
```diff
@@ -12,24 +12,27 @@ usage: dvc run [-h] [-q | -v] [-d DEPS] [-o OUTS] [-O OUTS_NO_CACHE]
                [--ignore-build-cache] [--remove-outs] [--no-commit]
                [--outs-persist OUTS_PERSIST]
                [--outs-persist-no-cache OUTS_PERSIST_NO_CACHE]
-               command
+               ...
```
shcheklein: I would keep the `command`. Especially considering that we have it two-three lines below. No matter what we have in the DVC itself

static/docs/commands-reference/repro.md
jorgeorpinel: Hmm but if `targets` only includes files and `-R` is used, it's basically ignored, right? I'll update the paragraph for your consideration assuming so for now...

shcheklein: right, it will be just ignored.

jorgeorpinel: OK paragraph changed (for all cmd refs) please check it out and lmk if you think it needs improvement in a new comment. (Resolving this one since it will be marked as outdated by Github.)

static/docs/commands-reference/repro.md
jorgeorpinel: Actually in my latest update to the -R option explanation I thought about it and decided to remove it as recursiveness does not limit any scope, if anything it expands the number of DVC-files that are checked or produced.

static/docs/commands-reference/pipeline.md
jorgeorpinel: Updated

static/docs/commands-reference/run.md
jorgeorpinel: But the actual usage output of `dvc run` has `...`, not `command` in that line. My suggestion is to improve the description to explain the ellipsis symbol, see 988dc79. Lmk if you'd like me to revert that last change and just change `...` for `command` though, as you prefer )

shcheklein: I would keep command ... it's easier to understand. We don't need to be precise with DVC. I would try to fix it on the DVC side as well.

shcheklein:
How does it fix #425 btw?

static/docs/commands-reference/run.md
jorgeorpinel: I tried fixing it on the core repo but it's how argparse outsputs it automtically since the argument takes all remanent command-line inputs. (It's an option in argparse and `dvc run` is the only one that uses it.)

Anyway, I added `command` in the doc and remove the `...` "remainder" explanation in the description.

jorgeorpinel:
Shoot, I haven't gotten to the "Understanding DVC" stuff yet. Let me remove that mention in this PR's description and will use the next PR for that for sure – sorry I've derailed from getting to it for some time...  Will prioritize right after this.

shcheklein:
@jorgeorpinel np! it's not a high priority. Let's prioritize now "dvc import/dvc get" - it's super important :)

jorgeorpinel:
Got it. And probably `pkg` commands too?

p.s. can this PR be merged? 😬

shcheklein:
@jorgeorpinel sure! there are no pkg commands anymore, that what I meant - dvc import, dvc get, dvc get-url

also, we will need to document DVC API after that  - dvc.api.open ... dvc.api.get_content

please, let's remove WIP from the PR :)

shcheklein:
cool stuff @jorgeorpinel 🎉

jorgeorpinel:
Understood!

PR #459, opened 2019-06-27
few improvements for use case

ryokugyu:
few improvements and remove `#` for right side menu


shcheklein:
good stuff @ryokugyu 🎉

PR #464, opened 2019-06-29
cmd ref: doc `import`, `get`, and `get-url`; other misc. updates

### LINKED ISSUE #385: document `dvc import` and `dvc get` (previously `pkg` commands)
UPDATE: See description in https://github.com/iterative/dvc/pull/2160 instead.

---
OLD: Some introductory descriptions for DVC packages (from Discord):

> You have project named `awesome_project`, which has `awesome_data` in  it
  pkg will allow you to "import" the project
  `dvc pkg import awesome_project://awesome_data` (syntax may be incorrect)

> 1. Any DVC repo is a _package_. Outputs inside packages can be called _artifacts_ (that we'll also be able to work with)
> 2. There will be a set of commands to import/export these _packages_ across different repositories and from a repository to a file system ~~(just a data artifact)~~
> 3. basic use cases include things like: dataset registry (reuse a single dataset in multiple project, while having a single point of responsibility for the dataset itself), easy mechanism to pull a data artifact w/o doing `git clone` + `dvc pull data.dvc`, etc
> 4. in the future (no need to focus on this for now) any set of DVC files will be considered as package, thus making pipelines reusable

jorgeorpinel:
Fix #385 based on desc in https://github.com/iterative/dvc/pull/2160


jorgeorpinel:
This is still WIP @shcheklein but you may want to review the changes from `term: ...` commits. Those are just misc. doc improvements that affect several files. Sorry its all mixed up in this PR...

static/docs/commands-reference/cache_dir.md
```diff
@@ -1,4 +1,4 @@
-# dir
+# cache dir
```
shcheklein: cache dir? :)

static/docs/commands-reference/config.md
```diff
@@ -187,7 +187,7 @@ Add an S3 remote and set it as the project default:
 > to create your bucket.
 
 ```dvc
-$ dvc remote add myremote s3://bucket/path
```
shcheklein: why do we change this? even though it's a key in S3 it's easier to understand as a path I believe

static/docs/commands-reference/get-url.md
```diff
@@ -0,0 +1,176 @@
+# get-url
+
+Download or copy file or directory from any supported URL (for example
+`http://`, `s3://`, `ssh://`, and other protocols) or local directory to the
+<abbr>workspace</abbr>.
```
shcheklein: it's actually not necessarily the workspace. It's more like a wget/rclone/curl. You don't even need a repo to get the data. The same for regular get. We might even even make a direct reference to import-url - that it's the same but does not require a repo, does not create DVC-files. It's single purpose is just download a data artifact.

static/docs/commands-reference/remote.md
```diff
@@ -43,11 +43,7 @@ Using DVC with a remote data storage is optional. By default, DVC is configured
 to use a local data storage only (usually `.dvc/cache` directory inside your
 repository), which enables basic DVC usage scenarios out of the box.
 
-[Add](/doc/commands-reference/remote-add),
```
shcheklein: why removing this reference?

static/docs/commands-reference/remote_add.md
```diff
@@ -18,18 +18,18 @@ usage: dvc remote add [-h] [--global] [--system] [--local] [-q | -v]
...
 ## Description
 
 `name` and `url` are required. `url` specifies a location to store your data. It
 could be S3 path, SSH path, Azure, Google cloud, Aliyun OSS local directory,
 etc. (See more examples below.) If `url` is a local relative path, it will be
-resolved relative to the current directory but saved **relative to the config
-file location** (see LOCAL example below). Whenever possible DVC will create a
-remote directory if it doesn't exists yet. It won't create an S3 bucket though
-and will rely on default access settings.
+resolved relative to the present working directory but saved **relative to the
```
shcheklein: double check that it's actually "current directory", i.e. some directory inside the workspace

static/docs/commands-reference/remote_add.md
```diff
@@ -253,7 +255,7 @@ $ dvc remote add myremote hdfs://user@example.com/path/to/dir
 >
 > - `pull`
 > - `fetch`
-> - `import`
+> - `import-url`
```
shcheklein: get, get-url?

static/docs/commands-reference/root.md
```diff
@@ -12,7 +12,7 @@ usage: dvc root [-h] [-q | -v]
 
 While in project's sub-directory, sometimes developers may want to refer some
 file belonging to another directory. This command returns relative path to the
-DVC project's root directory from the current working directory. So, this
```
shcheklein: `current working directory` or cwd is a know term

static/docs/commands-reference/status.md
```diff
@@ -91,15 +91,14 @@ cache. For the typical process to update workspaces, see
   name defined using the `dvc remote` command. Implies `--cloud`.
 
 - `-a`, `--all-branches` - compares cache content against all Git branches.
-  Instead of checking just the currently checked out workspace, it checks
-  against all other branches of this workspace. The corresponding branches are
-  shown in the status output. Applies only if `--cloud` or a remote is
-  specified.
+  Instead of checking just the workspace, it checks against all other branches
+  of this workspace. The corresponding branches are shown in the status output.
```
shcheklein: of his workspace -> of this repo

shcheklein: i'm not sure if we need use "against" as well. It just runs the same status command for all branches.

static/docs/commands-reference/version.md
```diff
@@ -24,7 +24,7 @@ system/environment:
 | `Filesystem type`                           | Shows the filesystem type (eg. ext4, FAT, etc.) and mount point of workspace and the cache directory                                                      |
 
 > If `dvc version` is executed outside a DVC workspace, the command outputs the
-> filesystem type of the current working directory.
```
shcheklein: same as above, cwd is a standard term

static/docs/user-guide/external-dependencies.md
```diff
@@ -28,6 +28,10 @@ $ dvc run -d /home/shared/data.txt \
           cp /home/shared/data.txt data.txt
 ```
 
+> ```yml
+> TODO: What does the DVC-file looks like?
```
shcheklein: WIP?

static/docs/commands-reference/config.md
jorgeorpinel: It's just a more correct term in the context of S3 but I can revert if you prefer.

Justification: Being used to S3 I always think about "keys" not "paths". These docs for ex. only talk about keys (and don't even mention paths): https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html and https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html. The term "path" is used in the S3 documentation in the context of bucket access URLs, which is not what we're referring to in the example above, e.g. https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

Lmk )

shcheklein: I still think it's better to keep path here. We don't talk about a specific key. It's more like a prefix for all object we will put later. So, it's confusing to see a key here. Also, my intuition is that people less familiar with S3 internals.

jorgeorpinel: Sure, reverted.

static/docs/commands-reference/get-url.md
jorgeorpinel: Got it! Good to know, will update.

jorgeorpinel: Updated in 54ca3ca. BTW re

> We might even even make a direct reference to import-url - that it's the same but does not require a repo, does not create DVC-files. Its single purpose is just download a data artifact.

And also `get` commands don't support `remote://` URLs right? See https://github.com/iterative/dvc/pull/2160#pullrequestreview-252712391

The import* vs get* refs are not that similar though, every section is different. I don't know if just making a ref from get to import will suffice as documentation.

static/docs/commands-reference/remote.md
jorgeorpinel: All the links are already on the top of the document. This reference isn't specific to any one commands, its jus a list of all commands again so I changed to simply "`dvc remote` commands".

static/docs/commands-reference/remote_add.md
jorgeorpinel: I changed all the occurrences of the term "current directory" with "present working directory" as I think they are synonyms (in order to standardize, using typical Linux jargon). Are they not synonyms?

jorgeorpinel: As for the description yes that's how it works, I just double checked.

static/docs/commands-reference/root.md
jorgeorpinel: Hmmm my life has been a lie! I always thought the standard term was "present working directory" from `pwd` but it seems the P there means "print"! Will revert all these changes back to "current"...

jorgeorpinel: Fixed in bc80b3d

static/docs/commands-reference/remote_add.md
jorgeorpinel: `get-url`, yes. good catch, thanks. Updated.

static/docs/commands-reference/status.md
jorgeorpinel: Updated, please check. I removed one of the `against` words but left the first one which I think make sense.

static/docs/user-guide/external-dependencies.md
jorgeorpinel: Woops, yes accidental add. Note removed!

static/docs/commands-reference/get-url.md
shcheklein: Definitely not enough. Just as an additional reference if you see it makes sense. Since they are similar after all.

re remote:// - yes, like @efiop mentioned in that comment.

static/docs/commands-reference/remote.md
shcheklein: I would keep ... it's just a nicer UI :) also gives a perspective at what's going on in the description

static/docs/commands-reference/get-url.md
```diff
@@ -0,0 +1,166 @@
...
+Download or copy file or directory from any supported URL (for example `s3://`,
+`ssh://`, and other protocols) or local directory to the local file system.
+
+> Like `dvc init`, this is one of the few commands that doesn't require an
+> existing DVC project to run.
+
+> See `dvc get` to download data from other DVC repositories (e.g. GitHub URLs).
+
+> Unlike `dvc import-url`, this command does not track the downloaded data
+> file(s) (does not create a DVC-file).
```
shcheklein: too many notes I think. let's keep the summary simple, put some of them into description. Some of them (like dvc get reference I would embed into the summary).

static/docs/commands-reference/get-url.md
```diff
@@ -0,0 +1,166 @@
...
+positional arguments:
+  url            (See supported URLs in the description.)
+  out            Destination path to put data to.
+```
+
+## Description
+
+In some cases it's convenient to get a data file or directory from a remote
+location. The `dvc get-url` command helps the user do so. The `url` argument
+should provide the location of the data to be imported, while `out` can be used
```
shcheklein: imported -> downloaded

static/docs/commands-reference/get-url.md
```diff
@@ -0,0 +1,166 @@
...
+DVC will be using default AWS credentials file to access S3. To override some of
+these settings, you could the options described in `dvc remote modify`.
+
+We use `boto3` library to set up a client and communicate with AWS S3. The
+following API methods are performed:
+
+- `list_objects_v2`, `list_objects`
+- `head_object`
+- `download_file`
+- `upload_file`
```
shcheklein: upload, delete, copy are not performed for sure, it might be that we only need the read permissions for the key

static/docs/commands-reference/import-url.md
```diff
@@ -1,8 +1,14 @@
...
-workspace and track changes in remote file or directory.
+Download or copy file or directory from any supported URL (for example `s3://`,
+`ssh://`, and other protocols) or local directory to the <abbr>workspace</abbr>,
+and track changes in the remote source with DVC. Creates a DVC-file.
+
+> See `dvc import` to download and tack data from other DVC repositories (e.g.
+> GitHub URLs).
+
+> See also `dvc get-url` which corresponds to the first step this command
+> performs (just download the data).
```
shcheklein: same here, looks complicated when there are too many notes.

static/docs/commands-reference/import-url.md
```diff
@@ -87,7 +93,7 @@ from having to install CLI tools for each service.
 
 When DVC inspects a DVC-file, its dependencies will be checked to see if any
 have changed. A changed dependency will appear in the `dvc status` report,
-indicating the need to reproduce this import stage. When DVC inspects an
```
shcheklein: "import stage" is kind of a noun phrase . It comes from code again. I'm not sure if there is a better term. But imported is not the right word for sure.

static/docs/user-guide/dvc-file-format.md
```diff
@@ -34,7 +34,7 @@ outs:
 locked: True
 
 # Comments like this line persist through multiple executions of
-# dvc repro/commit but not through dvc run/add/import commands.
+# dvc repro/commit but not through dvc run/add/import-url commands.
```
shcheklein: get should be added here as well

static/docs/user-guide/external-dependencies.md
```diff
@@ -90,7 +90,7 @@ $ dvc run -d remote://example/data.txt \
 Please refer to `dvc remote add` for more details like setting up access
 credentials for certain remotes.
 
-## Using import
+## Using import-url
```
shcheklein: we should update this page eventually, and include the regular import also

static/docs/commands-reference/get-url.md
jorgeorpinel: Sure, I did add the links between import<->import-get as well as import<->get, and all other related commands (:

static/docs/commands-reference/remote.md
jorgeorpinel: OK, reverted.

static/docs/commands-reference/get-url.md
jorgeorpinel: Agreed. Fixed in 4d236f8

static/docs/commands-reference/get-url.md
jorgeorpinel: Also updated in 4d236f8

static/docs/user-guide/dvc-file-format.md
jorgeorpinel: Updated in dae938c

static/docs/commands-reference/get-url.md
jorgeorpinel: Hey yeah why does `get-url` need to upload, delete, or copy anything within S3? I've removed the writing operations and permissions from this doc for now... Not making sense to me, can you confirm @efiop ? Thanks

static/docs/commands-reference/import-url.md
jorgeorpinel: You're right. "import stage" is fine. Reverted in 2915e33

static/docs/user-guide/external-dependencies.md
jorgeorpinel: Will do

jorgeorpinel:
Finally removing `WIP` from this PR title as first versions of `dvc import` and `dvc get` are now included in the changes. A few questions (in HTML comments) and examples pending.

static/docs/commands-reference/get-url.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Description
+
+In some cases it's convenient to get a data file or directory from a remote
+location. The `dvc get-url` command helps the user do so. The `url` argument
+should provide the location of the data to be downloaded, while `out` can be
+used to specify the (path and) file name desired for the downloaded data file or
+directory.
+
+> Like `dvc init`, this is one of the few commands that doesn't require an
```
shcheklein: it's still feels strange to reference `dvc init` here tbh. It's not also that important that there are a few commands. I would skip the first half of the sentence. And it feels that we can make it a regular paragraph, not a note. It's important information that it's a "global" command.

static/docs/commands-reference/get-url.md
shcheklein: does not require DVC repo to be run, and thus do not track ....

static/docs/commands-reference/get-url.md
shcheklein: @efiop please, can you provide more context here.

static/docs/commands-reference/get.md
```diff
@@ -0,0 +1,45 @@
+# get
+
+Download or copy file or directory from another DVC repository into the local
```
shcheklein: Should we clarify it a little bit:

`... DVC repository (usually, a link to Github or any other git server) ... `

or `... DVC repository (link to a Git server) ... `

or `... DVC repository in Git ...`

static/docs/commands-reference/get.md
```diff
@@ -0,0 +1,45 @@
...
+> Unlike `dvc import`, this command does not track the downloaded data file(s)
+> (does not create a DVC-file).
+
+## Synopsis
+
+```usage
+usage: dvc get [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
+
+positional arguments:
+  url                   DVC repository URL to download data from.
```
shcheklein: we should update this, it's not clear that DVC repo is Git repo. And it's easier to get the concept when you see that you can pass a link to a Git server.

static/docs/commands-reference/get.md
```diff
@@ -0,0 +1,45 @@
...
+  path                  Path to data within DVC repository.
+```
+
+## Description
+
+In some cases it's convenient to get a <abbr>data artifact</abbr> from another
+DVC repository. The `dvc get` command helps the user do so. The `url` argument
+should provide the external DVC project's Git repository URL, while `path` is
```
shcheklein: `DVC project's Git repository URL` - this is kind of a clarification I was mentioning above

static/docs/commands-reference/get.md
```diff
@@ -0,0 +1,45 @@
...
+used to specify the path to the data to be downloaded within the repo.
+
+<!-- A file of the same name is then created in the working directory? -->
```
shcheklein: some TODOs to be addressed?

static/docs/commands-reference/import-url.md
```diff
@@ -215,8 +222,9 @@ file has changed.
 ## Example: Detecting remote file changes
 
 What if that remote file is one which will be updated regularly? The project
-goal might include regenerating some artifact based on the updated data. A
-pipeline can be triggered to re-execute based on a changed external dependency.
+goal might include regenerating a <abbr>data artifact</abbr> based on the
+updated source. A pipeline can be triggered to re-execute based on a changed
```
shcheklein: we are changing this semantics ... we can update the docs and examples a bit later when we do docs for the `dvc update`

static/docs/commands-reference/import.md
```diff
@@ -0,0 +1,62 @@
+# import
+
+Download or copy file or directory from another DVC repository into the
```
shcheklein: same here as in `dvc get`

static/docs/commands-reference/import.md
```diff
@@ -0,0 +1,62 @@
...
+> See also `dvc get` which corresponds to the first step this command performs
+> (just download the data).
+
+## Synopsis
+
+```usage
+usage: dvc import [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
+
+positional arguments:
+  url                   DVC repository URL.
```
shcheklein: same here as in `dvc get`

static/docs/commands-reference/init.md
```diff
@@ -1,6 +1,6 @@
 # init
 
-This command initializes a DVC environment in a current Git repository.
+This command initializes a DVC environment in a local Git repository.
```
shcheklein: local is confusing term here. What if it's not local to me in a sense that it's located on remote machine - some EC2 instance? Either keep the previous version or just remove it.

static/docs/commands-reference/remote_remove.md
```diff
@@ -36,8 +36,8 @@ possible to edit config files manually.
 
 - `--local` - remove remote specified in the
   [local](/doc/user-guide/dvc-files-and-directories) configuration file
-  (`.dvc/config.local`). Local configuration files stores private settings or
-  local environment specific settings that should not be tracked by Git.
+  (`.dvc/config.local`). Local config files stores private configuration that
```
shcheklein: stores -> store

static/docs/commands-reference/version.md
```diff
@@ -11,6 +11,9 @@ usage: dvc version [-h] [-q | -v]
 
 ## Description
 
+> Like `dvc init`, this is one of the few commands that doesn't require an
```
shcheklein: I'm not sure this is really important )

static/docs/commands-reference/import.md
```diff
@@ -0,0 +1,62 @@
...
+> See `dvc import-url` to download and tack data from other supported URLs.
+
+The `dvc import` command helps the user create such an external data dependency.
+The `url` argument should provide the external DVC project's Git repository URL,
+while `path` is used to specify the path to the data to be imported within the
+repo. An import stage (DVC-file) is then created with the name of the data
+artifact, similar to having used `dvc run` to generate the same output as done
+in the external DVC project.
+
+<!-- SSH Git URL supported or only HTTPS? -->
```
shcheklein: TODOs?

static/docs/commands-reference/import.md
```diff
@@ -0,0 +1,62 @@
...
+usage: dvc import [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
+
+positional arguments:
+  url                   DVC repository URL.
+  path                  Path to data within DVC repository.
+```
+
+## Description
+
+In some cases it's convenient to add a <abbr>data artifact</abbr> from another
```
shcheklein: need to come with a better articulation here. For import and get. The biggest use case is that it's easier to re-use datasets/models created in one repo within another repo. For example, one of the actual cases - datasets registry - a github repo ppl use to store and update datasets. This command provides a way to "use" this dataset from that Gtihub repo.  Will think about better message here.

static/docs/commands-reference/import.md
```diff
@@ -0,0 +1,62 @@
...
+while `path` is used to specify the path to the data to be imported within the
+repo. An import stage (DVC-file) is then created with the name of the data
+artifact, similar to having used `dvc run` to generate the same output as done
+in the external DVC project.
+
+<!-- SSH Git URL supported or only HTTPS? -->
+
+<!-- External path to another DVC repo supported? -->
+
+## Options
```
shcheklein: let's create a ticker for examples

static/docs/commands-reference/get-url.md
jorgeorpinel: Updated in 88e434e but I did leave it as a note because I think it's not crucial information you need to know. I think of notes as things that the reader may chose to skip reading. You don't really need to know this but it's a nice-to-know. But I can make it a normal paragraph if you insist 😄

static/docs/commands-reference/get.md
jorgeorpinel: Yes, good call. Updated in e00738e

static/docs/commands-reference/get.md
jorgeorpinel: Ok but that's a change for the core `dvc` repo. So I'll update here and open an issue there...

jorgeorpinel: Ok updated in f3f57c1. What do you think about the text? `DVC repository URL (Git server link).`

static/docs/commands-reference/get.md
jorgeorpinel: Yes. I added a note that both HTTP and SSH [protocols](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols) are supported in 50f7c6e. Please check.

static/docs/commands-reference/get.md
jorgeorpinel: Getting there (:

static/docs/commands-reference/import-url.md
jorgeorpinel: I'm not sure what you mean. Can you clarify?

p.s. I didn't change that sentence, it's just a rewrap due to the `<abbr>`

static/docs/commands-reference/get-url.md
shcheklein: Yep, my point is that it's important to understand in case of `dvc get` or `dvc get-url` that they do not require repo. They are solving different use cases - all other commands are about organizing the ML/data project, these two are about getting you model, for example in the production environment.

static/docs/commands-reference/get.md
shcheklein: how about - `Git repository URL with DVC project to download from`?

static/docs/commands-reference/get.md
shcheklein: 👍 also, see the version of the phrase above

static/docs/commands-reference/init.md
jorgeorpinel: > What if it's not local to me in a sense that it's located on remote machine?

If I understand correctly, you would be SSH'd into that remote machine so from that terminal session's perspective it's a local repo. "Git repository on your local machine" is the phrase most used in Git docs (see [this search](https://git-scm.com/search/results?search=local%20machine)) but "local repository" is also used a couple times (see [this other search](https://git-scm.com/search/results?search=local%20repo)).

Anyway, I changed it to "Git repository where you're working from" in bc69724. Like?

static/docs/commands-reference/import-url.md
shcheklein: yes, sure. So, by default all external dependencies are going to be locked (`lock: true` and `-rev-lock: <git-commit-hash>` in the DVC-file). It means when you run `dvc repro` or `dvc status` they are not going to check the changes automatically. You will need to explicitly call `dvc update` to fetch the new data, update the `-rev-lock` field

static/docs/commands-reference/init.md
shcheklein: Yep, sounds good. Actually, reading this again, I realized that presence of Git is optional. So, it's more like `in a project directory`. That should be enough.

static/docs/commands-reference/remote_remove.md
jorgeorpinel: Fixed in ddece21

jorgeorpinel: Actually... For some reason this option had a different working in `remote list` and `remote remove`. I standardized it with the rest of the `remote` subcommands in a69f0ae.

static/docs/commands-reference/get-url.md
jorgeorpinel: Fair enough 🙂  Updated in e609e87

static/docs/commands-reference/get.md
jorgeorpinel: Yes, better. Went for `URL of Git repository with DVC project to download from` in c4444c6.

static/docs/commands-reference/get.md
jorgeorpinel: I didn't understand your last comment ): What version? What phrase?

If you're referring to a previous comment in this PR, a lot of them I've resolved and even so this PR is gotten so long that almost everything is hidden by default when the page loads. So please repeat the comment in the code if needed, thanks!

static/docs/commands-reference/import-url.md
jorgeorpinel: OK great added note in https://github.com/iterative/dvc.org/issues/478#issuecomment-510681023.

static/docs/commands-reference/init.md
jorgeorpinel: It's not optional.

```consone
$ mkdir empty
$ cd empty
empty$ dvc init
ERROR: failed to initiate dvc - /Users/drbidibombom/empty is not tracked by any
supported scm tool (e.g. git). Use '--no-scm' if you don't want to use any scm.
...
```

shcheklein: there is `--no-scm` option for that

static/docs/commands-reference/get.md
shcheklein: Ha :) It's hard to track things for me as well. I probably meant `URL of Git repository with DVC project to download from`

static/docs/commands-reference/version.md
jorgeorpinel: You're right. I guess it's obvious. Deleted in a7ee6aa

static/docs/commands-reference/import.md
jorgeorpinel: Yes, those are temporal. Only one left at this point and will remove soon.

jorgeorpinel:
@shcheklein please notice changes into info about `boto3` requirements and DVC installation in 9d0819d (related to https://discordapp.com/channels/485586884165107732/563406153334128681/599168364086165505)

static/docs/commands-reference/get-url.md
jorgeorpinel: Finally figured this out. See commit 9d0819d with message "s3: update info on boto3 methods and permissions required..."

static/docs/commands-reference/init.md
jorgeorpinel: True! Hmmmm actually this doc needed a bunch of updates once I read it slowly. See 0afdc5d )

static/docs/commands-reference/import.md
jorgeorpinel: Good point. OK, I did a considerable refactoring of this in 91b4e10, what do you think?

static/docs/commands-reference/import.md
jorgeorpinel: Good idea. Done, https://github.com/iterative/dvc.org/issues/487

static/docs/user-guide/external-dependencies.md
jorgeorpinel: in #487

static/docs/commands-reference/remote_add.md
```diff
@@ -31,11 +31,11 @@ config file location** (see LOCAL example below). Whenever possible DVC will
 create a remote directory if it doesn't exists yet. It won't create an S3 bucket
 though and will rely on default access settings.
 
-> If you installed DVC via `pip`, and depending on the remote type you plan to
-> use you might need to install optional dependencies: `s3`, `gs`, `azure`,
-> `ssh`. Or `all_remotes` to include them all. The command should look like
-> this: `pip install -U "dvc[s3]"` - it installs `boto3` library along with DVC
-> to support AWS S3 storage.
+> If you installed DVC via `pip`, depending on the remote type you plan to use
+> you might need to install optional dependencies: `[s3]`, `[gs]`, `[azure]`,
```
shcheklein: there should be also `[oss]` btw :)

static/docs/commands-reference/get-url.md
```diff
@@ -27,6 +27,8 @@ directory.
 Note that this command doesn't require an existing DVC project to run in. It's a
 single-purpose command that can be used out of the box after installing DVC.
 
+> See `dvc get` to download data from other DVC repositories (e.g. Github URLs).
```
shcheklein: in this specific case I would mention everywhere `data or model file or directory`. We want for ppl to pattern match on the case when model is needed in production environment.

static/docs/commands-reference/get.md
```diff
@@ -18,19 +18,23 @@ positional arguments:
...
 ## Description
 
-In some cases it's convenient to get a <abbr>data artifact</abbr> from another
-DVC repository. The `dvc get` command helps the user do so. The `url` argument
-should provide the external DVC project's Git repository URL (both HTTP and SSH
-protocols supported, e.g. `[user@]server:project.git`), while `path` is used to
-specify the path to the data to be downloaded within the repo.
+DVC provides an easy way to reuse datasets, intermediate results, ML models, or
+other files and directories tracked in another DVC repository into the present
+<abbr>workspace</abbr>. The `dvc get` command downloads such a <abbr>data
```
shcheklein: it's not even about present workspace only, it's about getting a file to deploy it for example. Or slice and dice with other tools, in an ad-hoc way. It not like you need a project/workspace for that.

static/docs/commands-reference/remote_add.md
jorgeorpinel: Added in 6d0d0e0

static/docs/commands-reference/get-url.md
jorgeorpinel: Understood. Updated in 2b1008a

static/docs/commands-reference/get.md
jorgeorpinel: You're right. Updated in d53434d, check it out )

jorgeorpinel:
Merging is blocked for me @shcheklein . You have to accept the changes, please check 🙂

jorgeorpinel:
Yay finally haha 🦉

PR #467, opened 2019-06-30
include information about disabling git-hooks

mroutis:
Related to:
https://stackoverflow.com/questions/56818930/dvc-push-after-several-local-commits/56820897#comment100200632_56820897

### DIFF
--- a/static/docs/commands-reference/install.md
+++ b/static/docs/commands-reference/install.md
@@ -52,6 +52,28 @@ The installed Git hook automates executing `dvc push`.
 - Git `pre-push` hook executes `dvc push` before `git push` to upload files and
   directories under DVC control to remote.
 
+For more information about git hooks, refer to the
+[git-scm documentation](https://git-scm.com/docs/githooks).
+
+## Disable Git hooks
+
+When you run `dvc install`, it creates three files under the `.git/hooks`
+directory:
+
+```
+.git/hooks
+├── post-checkout
+├── pre-commit
+└── pre-push
+```
+
+To disable them, you need to **remove** or **edit** those files (i.e.
+`rm .git/hooks/post-checkout`, `vim .git/hooks/pre-commit`).
+
+_If a hook already exists, DVC will try to append the corresponding content, so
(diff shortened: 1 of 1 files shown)


static/docs/commands-reference/install.md
```diff
@@ -52,6 +52,21 @@ The installed Git hook automates executing `dvc push`.
 - Git `pre-push` hook executes `dvc push` before `git push` to upload files and
   directories under DVC control to remote.
 
+For more information about git hooks, refer to the
+[git-scm documentation](https://git-scm.com/docs/githooks).
+
+## Disable Git hooks
+When you run `dvc install`, it creates three files under the `.git/hooks` directory:
```
shcheklein: please, install and run the pre-commit hook to format it properly

static/docs/commands-reference/install.md
```diff
@@ -52,6 +52,21 @@ The installed Git hook automates executing `dvc push`.
...
+
+```
+.git/hooks
+├── post-checkout
+├── pre-commit
+└── pre-push
+```
+
+To disable them, you need to remove those files (i.e. `rm .git/hooks/post-checkout`).
```
shcheklein: are we sure that those files are DVC-specific? do we overwrite them when we install the hooks or append to them?

mroutis: We append to them, @shcheklein , I'll add a note about it.

static/docs/commands-reference/install.md
```diff
@@ -52,6 +52,28 @@ The installed Git hook automates executing `dvc push`.
...
+.git/hooks
+├── post-checkout
+├── pre-commit
+└── pre-push
+```
+
+To disable them, you need to remove those files (i.e.
+`rm .git/hooks/post-checkout`).
+
+_If a hook already exists, DVC will try to append the corresponding content, so
```
shcheklein: Let's make the previous paragraph - "remove or edit"-style instead of this note. It's a way safer option.

mroutis: Agree :+1:

PR #470, opened 2019-07-02
config: add core.checksum_jobs config option

Witiko:
Documents the `core.checksum_jobs` config option added by https://github.com/iterative/dvc/pull/2197.

### DIFF
--- a/static/docs/commands-reference/config.md
+++ b/static/docs/commands-reference/config.md
@@ -71,6 +71,9 @@ This is the main section with the general config options:
   [anonymized usage statistics](/doc/user-guide/analytics). Accepts values
   `true` (default) and `false`.
 
+- `core.checksum_jobs` - number of threads for computing checksums. Accepts
+  positive integers. The default value is `max(1, min(4, cpu_count() // 2))`.
+
 ### remote
 
 These are sections in the config file that describe particular remotes. These


static/docs/commands-reference/config.md
```diff
@@ -71,6 +71,9 @@ This is the main section with the general config options:
   [anonymized usage statistics](/doc/user-guide/analytics). Accepts values
   `true` (default) and `false`.
 
+- `core.checksum_jobs` - number of threads for computing checksums. Accepts
```
shcheklein: @Witiko can we mention what is the default value for this, please? (even if it's a cpu()*something formula)

Witiko: Would it be ok if we just say that it is “some sane value”? The default value is `None` and although `RemoteBASE._collect_dir` uses the `max(1, min(4, cpu_count() / 2))` formula, future parts of the code may use other sane default.

shcheklein: Let's just mention this - `By default it's calculated by this (or similar) formula: ...? "`. It should give enough info - that it's something sane, conservative and related to cpu count.

Witiko:
I updated the text as suggested.

shcheklein:
@Witiko thanks! 🎉

PR #479, opened 2019-07-08
[WIP]cmd ref: improve run and commit

### LINKED ISSUE #460: cmd ref: improve `run` and `commit` / how to add outs/deps without re-running stage?

>

_UPDATE_ (From #612)

ryokugyu:
Fix #460


shcheklein:
@ryokugyu hey, any progress on this? do you need some help?

ryokugyu:
@shcheklein working on it. 
busy with college evaluations.
I need a bit of help with an example.

Naba7:
You can use the NLP tutorial for examples.

ryokugyu:
> You can use the NLP tutorial for examples.

thinking about the same, maybe it will be better if we keep more simple examples such as in `dvc add`

shcheklein:
Closing as stale.

PR #505, opened 2019-07-26
metrics show: add note about xpath overwriting xpath defined in stage file

pared:
Related to iterative/dvc#2285

### DIFF
--- a/static/docs/commands-reference/metrics_show.md
+++ b/static/docs/commands-reference/metrics_show.md
@@ -22,10 +22,16 @@ The optional `path` argument can represent a DVC metric file or a directory. If
 `path` is a directory, recursively search and process all metric files in it
 with the `-R` option.
 
-If metric type is specified (via `-t`) it overrides the type and `--xpath`
-defined (if any) in the corresponding DVC-file (usually, using
-`dvc metrics modify` command). If `-t` is not specified `-x` is ignored. The
-type can be detected by the file extension automatically if the type is
+- If type (via `-t`) and xpath (`-x`) are specified, they override the type and
+  xpath defined (if any) in the corresponding DVC-file (usually, using
+  `dvc metrics modify` command).
+- If type (via `-t`) is specified and xpath (`-x`) is not, type will overwrite
+  the type specified in DVC-file, no xpath will be considered, which will result
+  in displaying all metrics from file.
+- If type (via `-t`) is not specified and xpath (`-x`) is, DVC will try to read
+  type from DVC-file and apply xpath provided in `-x`.
+
+The type can be detected by the file extension automatically if the type is
 supported.
 
 ## Options
@@ -47,7 +53,8 @@ supported.
   value. Should be used if metric file contains multiple numbers and you need to
   get a only one of them. Only single path is allowed. If multiple metric files
   exist in the project, the same parser and path will be applied to all of them.
-  If xpath for particular metric has been set using [`dvc metrics modify`](https://dvc.org/doc/commands-reference/metrics-modify#options)
+  If xpath for particular metric has been set using
+  [`dvc metrics modify`](https://dvc.org/doc/commands-reference/metrics-modify#options)
   xpath passed in this option will owervrite it, only for current command run.
   It may fail to produce any results or parse files that are not in a
   corresponding format in this case. Accepted value depends on the metric file
--- a/static/docs/commands-reference/metrics_show.md
+++ b/static/docs/commands-reference/metrics_show.md
@@ -47,6 +47,8 @@ supported.
   value. Should be used if metric file contains multiple numbers and you need to
   get a only one of them. Only single path is allowed. If multiple metric files
   exist in the project, the same parser and path will be applied to all of them.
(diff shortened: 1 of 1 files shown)


pared:
@shcheklein I think we need to keep this logic. Type and xpath are entangled and we will probably not avoid this logic without some hard assumption, like "`-t` can be used only in `metrics modify`". I will try to make it a little bit more clear.

shcheklein:
@pared I'm not sure I understand this point `like "-t can be used only in metrics modify"`. Right now as far as I understand, documentation is wrong after the last PR was merged. Specifically, `If `-t` is not specified `-x` is ignored.` part. I'm fine to keep the logic if you think it makes sense, but docs should be updated appropriately, and comments in the code.

static/docs/commands-reference/metrics_show.md
```diff
@@ -22,11 +22,18 @@ The optional `path` argument can represent a DVC metric file or a directory. If
 `path` is a directory, recursively search and process all metric files in it
 with the `-R` option.
 
-If metric type is specified (via `-t`) it overrides the type and `--xpath`
-defined (if any) in the corresponding DVC-file (usually, using
-`dvc metrics modify` command). If `-t` is not specified `-x` is ignored. The
-type can be detected by the file extension automatically if the type is
-supported.
+- If metric type (via `-t`) and xpath(`-x`) is specified, it overrides the
+type and xpath
```
shcheklein: extra index that is not needed

static/docs/commands-reference/metrics_show.md
```diff
@@ -22,11 +22,18 @@ The optional `path` argument can represent a DVC metric file or a directory. If
 `path` is a directory, recursively search and process all metric files in it
 with the `-R` option.
 
-If metric type is specified (via `-t`) it overrides the type and `--xpath`
-defined (if any) in the corresponding DVC-file (usually, using
-`dvc metrics modify` command). If `-t` is not specified `-x` is ignored. The
-type can be detected by the file extension automatically if the type is
-supported.
+- If metric type (via `-t`) and xpath(`-x`) is specified, it overrides the
```
shcheklein: as far as I understand it's enough to specify _only_ `-t` to override the xpath (and make effectively empty if it's not provided explicitly)

static/docs/commands-reference/metrics_show.md
```diff
@@ -22,11 +22,18 @@ The optional `path` argument can represent a DVC metric file or a directory. If
...
+type and xpath
+defined(if any) in the corresponding DVC-file (usually, using `dvc metrics
```
shcheklein: extra space is needed before (

static/docs/commands-reference/metrics_show.md
shcheklein: oh, I see the comment below

static/docs/commands-reference/metrics_show.md
shcheklein: please use `.` at the end

static/docs/commands-reference/metrics_show.md
```diff
@@ -47,6 +54,8 @@ supported.
   value. Should be used if metric file contains multiple numbers and you need to
   get a only one of them. Only single path is allowed. If multiple metric files
   exist in the project, the same parser and path will be applied to all of them.
+  If xpath for particular metric has been set using [`dvc metrics modify`](https://dvc.org/doc/commands-reference/metrics-modify#options)
+  xpath passed in this option will owervrite it, only for current command run.
```
shcheklein: indent

static/docs/commands-reference/metrics_show.md
shcheklein: are specified

shcheklein: the override

static/docs/commands-reference/metrics_show.md
```diff
@@ -22,11 +22,18 @@ The optional `path` argument can represent a DVC metric file or a directory. If
...
-If metric type is specified (via `-t`) it overrides the type and `--xpath`
-defined (if any) in the corresponding DVC-file (usually, using
-`dvc metrics modify` command). If `-t` is not specified `-x` is ignored. The
-type can be detected by the file extension automatically if the type is
-supported.
+- If metric type (via `-t`) and xpath(`-x`) is specified, it overrides the
+type and xpath
+defined(if any) in the corresponding DVC-file (usually, using `dvc metrics
+modify` command).
+- If metric type is specified and xpath is not, type will overwrite the type
```
shcheklein: let's use `xpath` and `type` everywhere

shcheklein: let's mention (via `-t`)

static/docs/commands-reference/metrics_show.md
```diff
@@ -22,11 +22,18 @@ The optional `path` argument can represent a DVC metric file or a directory. If
...
+specified in DVC-file, no xpath will be considered, which will result in
+displaying all metrics from file.
+- If metric type is not specified and xpath is, DVC will try to read type from
+associated DVC-file and apply xpath provided in `-x`.
```
shcheklein: `associated` is probably not the right term.

just `from the DVC-file`

PR #509, opened 2019-07-30
imports: examples for new `import`, `update` cmd refs; et al [WIP]

### LINKED ISSUE #487: cmd ref: advanced examples for import (and get)
> The previous PR was too long. We're extracting this part, which is pending and hopefully the last related to `dvc import` and `dvc get`. See https://github.com/iterative/dvc.org/pull/464#pullrequestreview-259871929

- [x] Note that the `import-url` ref and `external-dependencies` guide also needs an update (See https://github.com/iterative/dvc.org/pull/464#pullrequestreview-257721375).

- [x] Create more sample repos for advanced `import` and possibly `get` command example/s. See https://github.com/iterative/dvc.org/issues/487#issuecomment-517520838 below.

- [x] Actually create/update advanced examples for import/get commands.

- ~~Check the `example-pipeline` carefully. It looks like it might be broken now. (Check both examples in fact.)~~ Moved to #564

- [x] Lets also update get started, work on use cases, etc. to incorporate these important new commands (include also `dvc update`).

jorgeorpinel:
Fix #487

### DIFF
--- a/src/Documentation/glossary.js
+++ b/src/Documentation/glossary.js
@@ -44,6 +44,13 @@ export default {
       desc:
         'A file or a directory that is under DVC control. See `dvc add`, ' +
         '`dvc run`, `dvc import`, `dvc import-url` commands.'
+    },
+    {
+      name: 'External Dependency',
+      match: ['external dependency', 'external dependencies'],
+      desc:
+        'A DVC-file dependency with origin in an external source. See ' +
+        '[External Dependencies](/doc/user-guide/external-dependencies).'
     }
   ]
 }
--- a/src/Documentation/glossary.js
+++ b/src/Documentation/glossary.js
@@ -42,8 +42,8 @@ export default {
       name: 'Output',
       match: ['output', 'outputs'],
       desc:
-        'A file or a directory that is under DVC control. See `dvc add`,' +
-        ' `dvc run`, `dvc import`, `dvc import-url` commands.'
+        'A file or a directory that is under DVC control. See `dvc add`, ' +
+        '`dvc run`, `dvc import`, `dvc import-url` commands.'
     }
   ]
 }
--- a/src/Documentation/glossary.js
+++ b/src/Documentation/glossary.js
@@ -50,8 +50,8 @@ export default {
       match: ['external dependency', 'external dependencies'],
       desc:
         'A DVC-file dependency with origin in an external source, for ' +
-        'example HTTP, SSH, Amazon S3, or Google Cloud Storage remote ' +
-        'locations. See ' +
+        'example HTTP, SSH, Amazon S3, Google Cloud Storage remote ' +
+        'locations, or even other DVC repositories. See ' +
         '[External Dependencies](/doc/user-guide/external-dependencies).'
     }
   ]
--- a/src/Documentation/glossary.js
+++ b/src/Documentation/glossary.js
@@ -49,7 +49,9 @@ export default {
       name: 'External Dependency',
       match: ['external dependency', 'external dependencies'],
       desc:
-        'A DVC-file dependency with origin in an external source. See ' +
+        'A DVC-file dependency with origin in an external source, for ' +
+        'example HTTP, SSH, Amazon S3, or Google Cloud Storage remote ' +
+        'locations. See ' +
         '[External Dependencies](/doc/user-guide/external-dependencies).'
     }
   ]
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -115,7 +115,10 @@ $ dvc add data.xml
 Adding 'data.xml' to '.gitignore'.
 Saving 'data.xml' to cache '.dvc/cache'.
 Saving information to 'data.xml.dvc'.
-...
+
+To track the changes with git run:
+
+	git add .gitignore data.xml.dvc
 ```
 
 As the output says, a DVC-file has been created for `data.xml`. Let us explore
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -162,7 +162,7 @@ $ file .dvc/cache/d8/acabbfd4ee51c95da5d7628c7ef74b
 
 What if you have not one dog picture, but hundreds of pictures of dogs and cats?
 Your goal might be to build an algorithm to identify dogs and cats in pictures,
-and this is your training data set:
+and this is your training dataset:
 
 ```dvc
 $ tree pics
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -115,10 +115,7 @@ $ dvc add data.xml
 Adding 'data.xml' to '.gitignore'.
 Saving 'data.xml' to cache '.dvc/cache'.
 Saving information to 'data.xml.dvc'.
-
-To track the changes with git run:
-
-	git add .gitignore data.xml.dvc
+...
 ```
 
 As the output says, a DVC-file has been created for `data.xml`. Let us explore
@@ -191,10 +188,7 @@ Linking directory 'pics'.
 [##############################] 100% pics
 
 Saving information to 'pics.dvc'.
-
-To track the changes with git run:
-
-  	git add pics.dvc
+...
 ```
 
 There are no DVC-files generated within this directory structure, but the images
--- a/static/docs/commands-reference/import-url.md
+++ b/static/docs/commands-reference/import-url.md
@@ -95,9 +95,9 @@ Both methods generate an equivalent [stage file](/doc/commands-reference/run)
 user from having to manually copy files from each of the remote storage schemes,
 and from having to install CLI tools for each service.
 
-Note that by default, import stages are locked in their DVC-files (with
-`locked: true`). Use `dvc update` manually on them to force updating the
-downloaded file or directory from the external data source.
+Note that by default, import stages are locked in their DVC-files (with field
+`locked: true` in this case). Use `dvc update` manually on them to force
+updating the downloaded file or directory from the external data source.
 
 > If a stage is unlocked (editing the `lock` value in its DVC-file, for example
 > using `dvc unlock`), they will start to be checked by `dvc status`, and
--- a/static/docs/commands-reference/import-url.md
+++ b/static/docs/commands-reference/import-url.md
@@ -95,9 +95,9 @@ Both methods generate an equivalent [stage file](/doc/commands-reference/run)
 user from having to manually copy files from each of the remote storage schemes,
 and from having to install CLI tools for each service.
 
-Note that by default, import stages are locked in their DVC-files (with field
-`locked: true` in this case). Use `dvc update` manually on them to force
-updating the downloaded file or directory from the external data source.
+Note that import stages are locked by default. Use `dvc update` manually on them
+to force updating the downloaded file or directory from the external data
+source.
 
 > If a stage is unlocked (editing the `lock` value in its DVC-file, for example
 > using `dvc unlock`), they will start to be checked by `dvc status`, and
@@ -170,7 +170,6 @@ Let's take a look at the resulting stage file (DVC-file) `data.xml.dvc`:
 ```yaml
 md5: 61e80c38c1ce04ed2e11e331258e6d0d
 wdir: .
-locked: true
 deps:
 - etag: '"f432e270cd634c51296ecd2bc2f5e752-5"'
   path: https://dvc.org/s3/get-started/data.xml
@@ -232,7 +231,6 @@ example:
 ```yaml
 md5: eca0a296d67781cc488c6ffd1cc63b8e
 wdir: .
-locked: true
 deps:
 - md5: a304afb96060aad90176268345e10355
   path: /tmp/dvc-import-url-example/data.xml
--- a/static/docs/commands-reference/import-url.md
+++ b/static/docs/commands-reference/import-url.md
@@ -90,10 +90,10 @@ $ dvc run -d https://example.com/path/to/data.csv \
           wget https://example.com/path/to/data.csv -O data.csv
 ```
 
-Both methods generate an equivalent stage file (DVC-file) with an external
-dependency. The `dvc import-url` command saves the user from having to manually
-copy files from each of the remote storage schemes, and from having to install
-CLI tools for each service.
+Both methods generate an equivalent [stage file](/doc/commands-reference/run)
+(DVC-file) with an external dependency. The `dvc import-url` command saves the
+user from having to manually copy files from each of the remote storage schemes,
+and from having to install CLI tools for each service.
 
 Note that by default, import stages are locked in their DVC-files (with
 `locked: true`). Use `dvc update` manually on them to force updating the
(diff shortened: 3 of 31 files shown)


static/docs/commands-reference/add.md
```diff
@@ -115,10 +115,7 @@ $ dvc add data.xml
 Adding 'data.xml' to '.gitignore'.
 Saving 'data.xml' to cache '.dvc/cache'.
 Saving information to 'data.xml.dvc'.
-
-To track the changes with git run:
```
shcheklein: It might be beneficial to keep at least once per example? It might click for someone on how DVC works when they see this command.

static/docs/commands-reference/import-url.md
```diff
@@ -90,10 +90,10 @@ $ dvc run -d https://example.com/path/to/data.csv \
...
 ```
 
-Both methods generate an equivalent stage file (DVC-file) with an external
-dependency. The `dvc import-url` command saves the user from having to manually
-copy files from each of the remote storage schemes, and from having to install
-CLI tools for each service.
+Both methods generate an equivalent [stage file](/doc/commands-reference/run)
+(DVC-file) with an external dependency. The `dvc import-url` command saves the
+user from having to manually copy files from each of the remote storage schemes,
+and from having to install CLI tools for each service.
```
shcheklein: few typos below:

1. If `dvc unlock` (of if we edit value manually) ...
2. -f option  mentions `dvc run` while it should be `dvc import-url`

static/docs/commands-reference/import-url.md
```diff
@@ -119,195 +119,155 @@ downloaded file or directory from the external data source.
...
-
-```dvc
-    $ git clone https://github.com/iterative/example-get-started
-```
-
-Second, let's install the requirements. But before we do that, we **strongly**
-recommend creating a virtual environment with
-[virtualenv](https://virtualenv.pypa.io/en/stable/) or a similar tool:
+Follow these instructions before each example if you actually want to try them
+on your system. First, download the project and move into it:
```
shcheklein: `move into it` sounds a it weird :)

static/docs/commands-reference/import-url.md
```diff
@@ -119,195 +119,155 @@ downloaded file or directory from the external data source.
...
 
 ```dvc
+$ git clone https://github.com/iterative/example-get-started
 $ cd example-get-started
-$ virtualenv -p python3 .env
```
shcheklein: it installs dvc as well, so .. I do understand though that we can keep it out of scope

static/docs/commands-reference/import-url.md
```diff
@@ -119,195 +119,155 @@ downloaded file or directory from the external data source.
...
+  persist: false
 ```
 
-The `etag` field in the DVC-file contains the ETag recorded from the HTTP
-request. If the remote file changes, the ETag changes, letting DVC know when the
-file has changed.
+The `etag` field in the DVC-file contains the
+[ETag](https://en.wikipedia.org/wiki/HTTP_ETag) recorded from the HTTP request.
+If the remote file changes, its ETag will be different, letting DVC know whether
+its necessary to download the it again.
```
shcheklein: `the it` - typo?

static/docs/commands-reference/import-url.md
```diff
@@ -119,195 +119,155 @@ downloaded file or directory from the external data source.
...
-
-```dvc
-$ git add data/.gitignore data.xml.dvc
+...
 ```
 
-> Note that it's possible to set up the other
-> [stages](/doc/commands-reference/run) from the _Getting Started_ example, but
-> since we don't need them for this example, we'll skip it.
+> Note that it's possible to set up the other stages from the _Get Started_
```
shcheklein: is actually important to mention?

static/docs/commands-reference/import-url.md
```diff
@@ -119,195 +119,155 @@ downloaded file or directory from the external data source.
...
-Let us again use the [Getting Started](/doc/get-started) example, in a way which
-will mimic an updated external data source.
+Let's use the [Get Started](/doc/get-started) project again, simulating an
+updated external data source.
 
-To make it easy to experiment with this, let us use a local directory as our
-remote data location. In real life the data file will probably be on a remote
-server, of course. Run these commands:
+To make it easy to experiment with this, let's use a local directory as our
+remote data location. (In real life, the data file will probably be on a remote
```
shcheklein: `remote data` can we come up with a better term here? it can be confused with remote data storage

static/docs/commands-reference/import-url.md
```diff
@@ -345,63 +298,71 @@ $ tree
...
-noticed by the `dvc status` command. Running `dvc repro` then ran both stages of
-this pipeline, and if we had set up the other stages they also would have been
-run. It first downloaded the updated data file. And then noticing that
-`data/data.xml` had changed, that triggered the `prepare.dvc` stage to execute.
+`dvc repro` ran again both stages of this pipeline (and if we had set up more
+stages, they also would have been run again). It first downloaded the updated
+data file. And then, noticing that `data/data.xml` had changed, that triggered
+the `prepare.dvc` stage to execute.
+
+> Feel free to clean up after this example get out of and remove the
```
shcheklein: I'm not sure it's important

static/docs/user-guide/external-dependencies.md
```diff
@@ -101,30 +101,28 @@ credentials for certain remotes.
...
 
-<details>
+The command above creates an <abbr>import stage</abbr> specified in DVC-file
+`data.xml.dvc` that uses an external dependency (in this case of HTTP type).
 
-### Expand to learn more about DVC internals
+<details>
 
-The resulting DVC-file will contain something like this:
+### Expand to see the (partial) resulting DVC-file
```
shcheklein: why partial?

static/docs/commands-reference/add.md
jorgeorpinel: I thought they weren't relevant to the example so removed, but you're right. Restored in 6863597

static/docs/commands-reference/import-url.md
jorgeorpinel: All examples in all cmd refs assume you have DVC installed on the system right? So this can be kept out?

static/docs/commands-reference/import-url.md
jorgeorpinel: Thanks. Updated text in 1a214ea

static/docs/commands-reference/import-url.md
jorgeorpinel: OK how about "`cd` into it"? Also updated in 1a214ea

static/docs/commands-reference/import-url.md
jorgeorpinel: Yes, thanks. Also updated in 1a214ea

static/docs/commands-reference/import-url.md
jorgeorpinel: Good point. Not important for the example. Removed it in 1a214ea as well.

static/docs/commands-reference/import-url.md
jorgeorpinel: How about:

"let's use a local machine directory (external to the sample DVC project) to simulate a remote data source location" ?

also updated in 1a214ea

static/docs/commands-reference/import-url.md
jorgeorpinel: Me neither. I guess not! Removed in 1a214ea as well.

jorgeorpinel:
> Is it still WIP?

Yes, haven't included the examples for `import` and `update` yet.

static/docs/user-guide/external-dependencies.md
jorgeorpinel: Because I'm only showing the `deps` section of the DVC-file. I guess the wording is weird. Changed to "Expand to see resulting dependencies in the DVC-file" in 6ec8f30.

What do you think?

static/docs/commands-reference/import-url.md
shcheklein: 👍

static/docs/commands-reference/import-url.md
```diff
@@ -99,16 +99,18 @@ Note that by default, import stages are locked in their DVC-files (with
...
 ## Options
 
 - `-f`, `--file` - specify name of the DVC-file it generates. By default the
   DVC-file name generated is `<file>.dvc`, where `<file>` is file name of the
-  output (`out`). The stage file is placed in the same directory where `dvc run`
-  is run by default, but `-f` can be used to change this location, by including
-  a path in the provided value (e.g. `-f stages/stage.dvc`).
+  output (`out`). The stage file is placed in the same directory where
+  `dvc import-url` is run by default, but `-f` can be used to change this
+  location, by including a path in the provided value (e.g.
```
shcheklein: `location and file name`

static/docs/user-guide/external-dependencies.md
shcheklein: yeah, I understand. I mean why do we want to show partials only. I think we can certainly remove some default values but I would show more or less all parts of the file.

shcheklein: or we can put `...` in the YAML block or even `(not relevant parts are omitted)` ... but again, I would use it carefully and try to keep the high level structure for the file - it can give more intuition to a user

jorgeorpinel: True. I guess we don't use many partial DVC-file code blocks in other docs. I restored more complete versions with `...` instead of initial default values in 93f6fe7

static/docs/commands-reference/import-url.md
jorgeorpinel: Also added in 93f6fe7.

jorgeorpinel:
37784e8 introduces a fist, basic example for the `dvc import` cmd ref. Basically the same as the corresponding example in the **External Dependencies** guide.

static/docs/commands-reference/import.md
```diff
@@ -72,3 +73,47 @@ force updating the downloaded data artifact from the external DVC repo.
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
+
+## Example
+
+An obvious case for this command is to import a complete ML model from an
```
shcheklein: in this case it's better to pull dataset - data.xml from the repo

shcheklein: we'll need to come up with a better example repo for this when we start doing use cases and updating get started

jorgeorpinel:
07b4250 adds a super simple example for the `dvc update` cmd ref. Not sure it's needed as I plan to make another one right after with an actually changed data artifact...

shcheklein:
@jorgeorpinel it looks a bit strange, but let's keep it. better than nothing. Let me know when I can merge this. Then we can think about more advanced examples, get started, use cases.

static/docs/commands-reference/import.md
jorgeorpinel: changed to download dataset instead in d739685.

I agree it would be nice to have other repo examples. What do you have in mind? I'm almost ready to start updating the _Get Started_ and _Use Cases_.

jorgeorpinel:
Everything has basic examples now @shcheklein. I think we're ready for the advanced ones and related _Get Started_ and _User Guide_ updates. I'll message you on Discord about the plan for that.

I was planning to use this same PR for the advanced stuff too but feel free to merge it now if you prefer 🙂

static/docs/commands-reference/import.md
```diff
@@ -72,3 +73,45 @@ force updating the downloaded data artifact from the external DVC repo.
...
+[external dependency](/doc/user-guide/external-dependencies). Check
+`data.xml.dvc`:
+
+```yaml
+md5: 7de90e7de7b432ad972095bc1f2ec0f8
+wdir: .
+locked: true
+deps:
+- path: data/data.xml
+  repo:
```
shcheklein: we should update `https://dvc.org/doc/user-guide/dvc-file-format` btw :)

static/docs/commands-reference/update.md
```diff
@@ -37,3 +37,32 @@ their DVC-files (via fields `lock: true` and `rev_lock`). For a locked stage,
...
+`dvc update` on it instead:
+
+```dvc
+$ dvc update model.pkl.dvc
+Stage 'model.pkl.dvc' didn't change.
+Output 'model.pkl' didn't change. Skipping saving.
+Saving information to 'model.pkl.dvc'.
+```
+
+This time nothing has changed (since our Get Started sample repo is pretty
```
shcheklein: link to Get started repo, also sometimes you use it all lower case

static/docs/get-started/example-pipeline.md
```diff
@@ -68,63 +68,62 @@ that are described in earlier [get started](/doc/get-started) chapters.
...
-  When we run `dvc add` `Posts.xml.zip`, DVC creates a
-  [DVC-file](/doc/user-guide/dvc-file-format).
+```dvc
+$ mkdir data
+$ wget -P data https://dvc.org/s3/examples/so/Posts.xml.zip
+$ dvc add data/Posts.xml.zip
+```
 
-  ### Expand to learn more about DVC internals
+<details>
```
shcheklein: looks like a bug in structure

static/docs/user-guide/contributing.md
```diff
@@ -31,39 +31,39 @@ contributing!
...
+```
+
+Install coding style pre-commit hooks with:
+
+```dvc
+$ pip install pre-commit
+$ pre-commit install
+```
+
+> Note: the double quotes (`""`) used to wrap the dependencies, which are to be
```
shcheklein: should we move this Note up a bit?

static/docs/commands-reference/import.md
jorgeorpinel: Updated requirements in https://github.com/iterative/dvc.org/issues/487#issuecomment-517520838

static/docs/commands-reference/import.md
jorgeorpinel: Good point. Updated in 710b95c, do you think that's enough of a change?

static/docs/commands-reference/update.md
jorgeorpinel: Hmmm. It's already linked at the beginning of the example (link 44 right now). But I agree this line is strange, changed in 20b8f0d

static/docs/get-started/example-pipeline.md
jorgeorpinel: Yeah good find. I only unindented that file, so that error was in before I think, but I just fixed it in f8bbf8f

static/docs/user-guide/contributing.md
jorgeorpinel: Yes. Done in 9e2ada7

jorgeorpinel:
re `check the example-pipeine carefully`

Moved to https://github.com/iterative/dvc.org/issues/487#issuecomment-517520838

static/docs/commands-reference/update.md
```diff
@@ -37,3 +37,31 @@ their DVC-files (via fields `lock: true` and `rev_lock`). For a locked stage,
...
+
+```dvc
+$ dvc import git@github.com:iterative/example-get-started model.pkl
+Importing 'model.pkl (git@github.com:iterative/example-get-started)' -> 'model.pkl'
+...
+Saving information to 'model.pkl.dvc'.
+...
+```
+
+As DVC mentions, the import stage (DVC-file) `model.pkl.dvc` is created. this
```
shcheklein: this -> This

static/docs/commands-reference/update.md
```diff
@@ -37,3 +37,31 @@ their DVC-files (via fields `lock: true` and `rev_lock`). For a locked stage,
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
+
+## Example
```
shcheklein: we should remember to update the description above and mention that `import-url` locks artifacts as well. and that `dvc update` updates both now.

shcheklein: also, second Note about dvc repro looks redundant to me.

also, `This is the only command that does so with locked stages, and import stages are locked by default.` - not sure we need this.

it's better to describe what's happening here from the user perspective in the first place and leave internal mechanics to `>` or expandable notes. From a user perspective it's simple - download a new version and update DVC-file. I don't care if it unlocks file, if it runs repro inside, etc - these details should go to notes.

src/Documentation/glossary.js
```diff
@@ -42,8 +42,15 @@ export default {
...
-        'A file or a directory that is under DVC control. See `dvc add`,' +
-        ' `dvc run`, `dvc import`, `dvc import-url` commands.'
+        'A file or a directory that is under DVC control. See `dvc add`, ' +
+        '`dvc run`, `dvc import`, `dvc import-url` commands.'
+    },
+    {
+      name: 'External Dependency',
+      match: ['external dependency', 'external dependencies'],
+      desc:
+        'A DVC-file dependency with origin in an external source. See ' +
```
shcheklein: it worth specifying some list of external storages, just to give more sense of what is happening here

static/docs/commands-reference/import.md
shcheklein: yep, should be enough for now

static/docs/commands-reference/update.md
jorgeorpinel: Thanks. Fixed.

static/docs/commands-reference/update.md
jorgeorpinel: Updated! And yes that note was totally redundant. Reorganized the whole description in 77f9533 but let me think about your comment on the user's perspective...

jorgeorpinel: OK. I like this idea to focus command references (and docs in general) to user-focused descriptions and explanations. BUT we currently don't really have that in many places. For example:

- The description in https://dvc.org/doc/commands-reference/add goes into detail on what happens "under the hood" after the 1st paragraph.
- https://dvc.org/doc/commands-reference/checkout also has a pretty detailed and long description
- https://dvc.org/doc/commands-reference/commit also has an under-the-hood style bullet list in its description
  etc.

So seems to me that for consistency's sake we'd need to make a major review of all command refs in order to take this new direction?

All that said, I did rewrite this one description focusing on the user's perspective in 54fbeb3. (So the summarized changes are https://github.com/jorgeorpinel/dvc.org/compare/b8d4512...54fbeb3.)

shcheklein: I would keep all these details. Let me correct my previous comment. I think description can and should include as much details as possible, but it should be user-centric anyway. We should start them with some high-level description - `dvc commit` is more or less good example. And it's totally fine to include more under-the-hood details. At least when they describe some user-visible things - like a file moved to cache, DVC-file created with a specific field. I would avoid describing specific mechanism certain behavior implemented with. I think this is similar to public vs private methods in software engineering :)

In our specific case. Let's remove this lock/unlock stuff just because we are going to change it soon, locking of these import/import-url stages won't depend on `lock: true` - it will be by definition for those stages.

jorgeorpinel: > I would avoid describing specific mechanism certain behavior implemented with. I think this is similar to public vs private methods in software engineering :)

Got it!

> Let's remove this lock/unlock stuff just because we are going to change it soon

OK, removed in debb558

src/Documentation/glossary.js
jorgeorpinel: Agree. Added in 634ed43

static/docs/commands-reference/update.md
```diff
@@ -13,22 +13,20 @@ positional arguments:
...
-bring the imported <abbr>data artifact</abbr> up to date. This is the only
-command that does so with locked stages, and import stages are locked by
-default.
+After creating <abbr>import stages</abbr>
+([DVC-files](/doc/user-guide/dvc-file-format)) with `dvc import` or
+`dvc import-url`, the external data source can change. Use `dvc update` to bring
+these imported file, directory, or <abbr>data artifact</abbr> up to date.
+
+Note that import stages are locked by default. `dvc update` is the only single
+command that updates locked stages.
```
jorgeorpinel: @shcheklein re

> For each DVC-file:
>
> 1. It pings the URL (import-url) or checks remote Git repo (in case of import) that are specified in DVC-file created by dvc import/dvc import-url to see if the new version available (by comparing artifact checksum or rev-lock).
> 2. If the new version available it downloads and replaces the existing one with a new one.
> 3. It saves data to cache. Depending on cache strategy either by copying or by utilizing links.
> 4. It updates the DVC-file with the new checksum and rev-lock in case of dvc import.
>
> ... remove it, just because we are going to change the implementation logic - it won't depend on lock: true anymore soon.

OK note that I already removed the references to `locked: true` in 634ed43. But it seems I should just remove these notes completely. I'm just curious: how will import stages be locked moving forward if no `locked: true` will be used? For `dvc status` and `dvc repro` I mean.

jorgeorpinel: Moved discussion to https://github.com/iterative/dvc.org/pull/509#pullrequestreview-270476714

shcheklein: @jorgeorpinel they will be locked by "definition". DVC is aware about their type, that they are created with import/import-url and repro won't affect them no matter what you do. At least for now. The only way to update them will be `dvc update`.

src/Documentation/glossary.js
```diff
@@ -42,8 +42,17 @@ export default {
...
-        'A file or a directory that is under DVC control. See `dvc add`,' +
-        ' `dvc run`, `dvc import`, `dvc import-url` commands.'
+        'A file or a directory that is under DVC control. See `dvc add`, ' +
+        '`dvc run`, `dvc import`, `dvc import-url` commands.'
+    },
+    {
+      name: 'External Dependency',
+      match: ['external dependency', 'external dependencies'],
+      desc:
+        'A DVC-file dependency with origin in an external source, for ' +
```
shcheklein: should it include Git repo as a source?

static/docs/commands-reference/update.md
jorgeorpinel: I see. I'm still curious how DVC knows they're import stages. I guess given the fields used in the `deps` section of the DVC-file. Alright, Updated in f9e1d76! It was a relatively big change because one of the examples in `dvc import-url` had to be reviewed and partially re-written.

src/Documentation/glossary.js
jorgeorpinel: Hmmm how about "or even other DVC repositories" ? Updated in dc439ac

jorgeorpinel:
This should be ready for merge.

static/docs/commands-reference/import-url.md
```diff
@@ -95,13 +95,9 @@ Both methods generate an equivalent [stage file](/doc/commands-reference/run)
...
 and from having to install CLI tools for each service.
 
-Note that import stages are locked by default. Use `dvc update` manually on them
-to force updating the downloaded file or directory from the external data
-source.
-
-> If a stage is unlocked (editing the `lock` value in its DVC-file, for example
-> using `dvc unlock`), they will start to be checked by `dvc status`, and
-> updated by `dvc repro`.
+Note that import stages are considered always locked. (See `dvc lock`.) They can
```
shcheklein: let's remove `See dvc lock` and that they cannot be unlocked. It will be a different type of lock. 

Something like this:

Note that import stages are considered always "locked" - it means that if you run `dvc repro` DVC won't try to update them. You must run `dvc update` explicitly ...

shcheklein: the same for import

static/docs/commands-reference/import-url.md
```diff
@@ -151,18 +149,25 @@ the _Add Files_ step mentioned before.
...
 ```dvc
 $ dvc import-url https://dvc.org/s3/get-started/data.xml data/data.xml
 Importing 'https://dvc.org/s3/get-started/data.xml' -> 'data/data.xml'
 [##############################] 100% data.xml
-...
+[##############################] 100% data.xml
+
+Adding 'data/data.xml' to 'data/.gitignore'.
+
+WARNING: data 'data/data.xml' exists. Removing before checkout.
```
shcheklein: @efiop is not fixed yet?

efiop: Not yet fixed. We are working on it in https://github.com/iterative/dvc/pull/2358. Let's not include this into the command-reference, since it is a bug.

static/docs/commands-reference/import-url.md
```diff
@@ -308,56 +300,41 @@ Pipelines are up to date. Nothing to reproduce.
...
 
 ```dvc
-$ dvc status
-data.xml.dvc:
-	changed deps:
-		modified:           .../tmp/dvc-import-url-example/data.xml
+$ dvc update data.xml.dvc
+WARNING: Dependency '.../tmp/dvc-import-url-example/data.xml' of 'data.xml.dvc' changed because it is 'modified'.
+WARNING: Stage 'data.xml.dvc' changed.
+Reproducing 'data.xml.dvc'
```
shcheklein: really bad UI, created a ticket https://github.com/iterative/dvc/issues/2367

src/Documentation/glossary.js
shcheklein: sounds good to me!

static/docs/commands-reference/import-url.md
jorgeorpinel: Same for `update`. Updated in 5f14b34

static/docs/commands-reference/import-url.md
jorgeorpinel: Removed in 0f50522

jorgeorpinel:
Merge? 🙂

PR #514, opened 2019-07-31
remove show-checksums reference from doc

kurianbenoy:
Documentation updated to Fix the issue in DVC:
https://github.com/iterative/dvc/pull/2331


shcheklein:
@kurianbenoy thanks!

PR #527, opened 2019-08-07
Fix #305

### LINKED ISSUE #305: support directories for the third level documentation
To support tickets like #144 we need to add a real third level support. Files will be stored in subdirectories.

iAdramelk:
This PR should fix #305. It contains following changes.

New format for `sidebar.json`. It removes distinction between sections and files and allows any number of nesting levels.

New format (full):

```
{
  "slug": "metrics",
  "label": "metrics",
  "source": "metrics/index.md",
  "children": []
}
```

`slug` - **Required**. URL part for the current link, will be also used as URL prefix for all children of the current category. Example: `metrics` -> `/doc/metrics`. For the child `add`of metrics -> `/doc/metrics/add`.

`label` - name of the menu item. If not set, it will be generated from `slug` using `startCase` from lodash. Generating from slug example: `get-started` -> `Get Started`.

`source` -> name of the corresponding .md file in the filesystem. If not set, it will be generated from `slug`. Generating from slug example: `metrics` -> `metrics.md`. For legacy reasons you can explicitly set this field to `false`, in this case after user select this item, its first child will be shown instead.

`children` - optional array of child items.

You can also use short syntax. `metrics` is the same as `{ "slug": "metrics" }`.

This config is parsed using new `src/Documentation/SidebarMenu/helper.js` file. It will normalize data from "sidebar.json" and generate all necessary field. It's output structure is:

```
{
  label: "Add Files or Directories",
  path: "/doc/get-started/add-files",
  source: "/static/docs/get-started/add-files.md",
  prev: "/doc/get-started/configure",
  next: "/doc/get-started/share-data",
  children: []
}
```

This way we don't need to calculate them in react component anymore.

This file also exports two new helper functions:

`getItemByPath` - recursively search for items in sidebar, that matches given path. It also resolve some edge cases (like `source = false`, and `/doc/` route).
`getParentsListFromPath` - returns all parents paths from current path. Example: `/doc/commands-reference/destroy` will return `['/doc/commands-reference', '/doc/commands-reference/destroy']`. Used to make nescessary menu items active.

`pages/doc.js`, `src/Documentation/Markdown/Markdown.js` and `src/Documentation/SidebarMenu/SidebarMenu.js` are updated accordingly, all business logic related to parsing `sidebar.json` are removed from them and replaced with data from new normalized menu items.

Difference from the current version:

1. Prev/Next are generated for all pages, not only for pages inside the same section.
2. Right now each nesting level in menu has different left margin. In this version all nested levels always add 20px.


pages/doc.js
```diff
@@ -8,27 +8,22 @@ import { RightPanel } from '../src/Documentation/RightPanel/RightPanel'
 import Page from '../src/Page'
 import SearchForm from '../src/SearchForm'
 import Page404 from '../src/Page404'
-import PerfectScrollbar from 'perfect-scrollbar'
```
shcheklein: let's keep it for now. Scrolling behave differently if you remove it and `wheelPropagation: true` stuff. Also, would need to check different browsers.

iAdramelk: It's moved to the `/src/Documentation/SidebarMenu/SidebarMenu.js`, because it requires  `#sidebar-menu` that is created in that file.

src/Documentation/SidebarMenu/helper.js
```diff
@@ -0,0 +1,136 @@
...
+    next: "/doc/get-started/share-data",
+    children: []
+  }
+*/
+
+const PATH_ROOT = '/doc/'
+const FILE_ROOT = '/static/docs/'
+const FILE_EXTENSION = '.md'
+
+let prevReference // We will save prev item reference here to use for the next and prev fields
```
shcheklein: it's definitely > 80 symbols. I would recommend to install the pre-commit hook: https://dvc.org/doc/user-guide/contributing-documentation#development-environment

shcheklein: and run the prettier from the command line once again to reformat what has been committed already

iAdramelk: I think prettier ignores comment lengths, so it probably didn't help. But I will fix comment lengths.

shcheklein: you might be right - it just does not know what can be done with it.

pages/doc.js
```diff
@@ -109,47 +66,37 @@ export default class Documentation extends Component {
...
 
-  onFileSelect = (file, section, e) => {
-    e && e.preventDefault()
-    this.setCurrentPath(section, file)
-    this.loadFile({ file, section, parseHeadings: true })
-  }
+    if (!item) {
+      this.setState({ pageNotFound: true, currentItem: {} })
+
+      return
```
shcheklein: it's probably better to use `else` instead of an explicit `return`

src/Documentation/SidebarMenu/helper.js
```diff
@@ -0,0 +1,149 @@
...
+    path: PATH_ROOT + parentPath + slug,
+    source: source === false ? false : sourcePath,
+    label: label ? label : startCase(slug),
+    prev,
+    next: undefined
+  }
+}
+
+function normalizeSidebar(data, parentPath, currentSidebar) {
+  data.forEach(item => {
```
shcheklein: ie11? not that it's very important to keep supporting it, but if it's the only that breaks it I would do some workaround

src/Documentation/SidebarMenu/helper.js
```diff
@@ -0,0 +1,149 @@
...
+const normalizedSidebar = [] // Current state of sidebar to search for prev
+
+// Normalization
+
+function normalizeItem(item, parentPath) {
+  validateRawItem(item)
+
+  const { label, slug, source } = item
+
+  // If prev item don't have source we need to recirsively search for it
```
shcheklein: minor: don't -> doesn't

src/Documentation/SidebarMenu/helper.js
```diff
@@ -0,0 +1,149 @@
...
+    throw Error(
+      "If you set 'source' to false, you had to add at least one child"
+    )
+  }
+}
+
+// Global cache vars used in normalization
+
+let prevReference // Save last item here to generate the prev field
+const normalizedSidebar = [] // Current state of sidebar to search for prev
```
shcheklein: `Current state of sidebar to search for` - a bit misleading. It's not only a "cache" to search for prev. This is what we export then as a final result, right?

src/Documentation/SidebarMenu/helper.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  if (isSourceDisabled && (!children || !children.length)) {
+    throw Error(
+      "If you set 'source' to false, you had to add at least one child"
+    )
+  }
+}
+
+// Global cache vars used in normalization
```
shcheklein: can we pass them between functions, do we actually need a global prevReference, for example?

shcheklein:
@iAdramelk https://dvc-org-pr-527.herokuapp.com/doc/understanding-dvc/what-is-dvc.md - not found page is broken + the page itself is broken?

pages/doc.js
iAdramelk: Ok, I will change it.

src/Documentation/SidebarMenu/helper.js
iAdramelk: I have an idea how it can be done, but I'm not sure that it will make code easier to understand, but I will try today.

src/Documentation/SidebarMenu/helper.js
iAdramelk: Will try to move them inside function and if I wasn't able to, will change text.

src/Documentation/SidebarMenu/helper.js
iAdramelk: Fixed

src/Documentation/SidebarMenu/helper.js
iAdramelk: forEach works in IE starting with IE9 according to https://caniuse.com/#feat=es5 , so it should not be a problem.

iAdramelk:
@shcheklein Thie problem with https://dvc-org-pr-527.herokuapp.com/doc/understanding-dvc/what-is-dvc.md is that I forgot to remove `.md` extension form the `slug` field, my bad. Fixed it and checked other pages to make sure that this is the only one.

iAdramelk:
Yep, removed prevReference and current result from global scope successfully.

src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -1,13 +1,44 @@
...
-import startCase from 'lodash.startcase'
-import includes from 'lodash.includes'
 // styles
 import styled from 'styled-components'
 import { media, OnlyDesktop } from '../../styles'
+// sidebar helpers
+import { getParentsListFromPath } from './helper'
+
+function SidebarMenuItem({ children, label, path, activePaths, onNavigate }) {
+  const isActive = activePaths && activePaths.includes(path)
```
shcheklein: use lodash.includes, otherwise we are breaking compatibility

iAdramelk: You are right, will change.

PR #536, opened 2019-08-08
Update links to subcommands

iAdramelk:
- Chech if page exists and only create link it if it exists
- If subcommand page exists – link to page, if not, create anchor

### DIFF
--- a/src/Documentation/Markdown/utils/remark-linker.js
+++ b/src/Documentation/Markdown/utils/remark-linker.js
@@ -3,7 +3,7 @@
 import visit from 'unist-util-visit'
 import { getItemByPath } from '../../SidebarMenu/helper'
 
-const DVC_REGEXP = /dvc\s+[a-z][a-z-.]*\s/
+const DVC_REGEXP = /dvc\s+[a-z][a-z-.]*/
 const COMMAND_REGEXP = /^[a-z][a-z-]*$/
 const COMMAND_ROOT = '/doc/commands-reference/'
 
--- a/src/Documentation/Markdown/utils/remark-linker.js
+++ b/src/Documentation/Markdown/utils/remark-linker.js
@@ -3,25 +3,30 @@
 import visit from 'unist-util-visit'
 import { getItemByPath } from '../../SidebarMenu/helper'
 
+const DVC_REGEXP = /dvc\s+[a-z][a-z-.]*\s/
+const COMMAND_REGEXP = /^[a-z][a-z-]*$/
+const COMMAND_ROOT = '/doc/commands-reference/'
+
 function linker() {
   function transformer(tree) {
     visit(tree, 'inlineCode', function(node, index, parent) {
-      if (parent.type !== 'link' && /dvc\s+[a-z-.]+/.test(node.value)) {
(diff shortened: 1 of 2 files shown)


src/Documentation/Markdown/utils/remark-linker.js
```diff
@@ -1,23 +1,36 @@
...
+        let tmpUrl = '/doc/commands-reference/'
+        let url
+
+        if (getItemByPath(tmpUrl + parts[1])) {
+          url = tmpUrl = tmpUrl + parts[1]
+        }
 
         if (parts.length > 2) {
-          url += '#' + parts[2]
+          if (getItemByPath(tmpUrl + parts[2])) {
```
shcheklein: a little bit concerned re the performance of this if page is large enough, and sidebar will become bigger

also, why do we need to do this check - what are the cases when we want to be "smart"?

also, what are the cases with `#`?

src/Documentation/Markdown/utils/remark-linker.js
```diff
@@ -1,23 +1,36 @@
 ;`use strict`
 
 import visit from 'unist-util-visit'
+import { getItemByPath } from '../../SidebarMenu/helper'
 
 function linker() {
   function transformer(tree) {
     visit(tree, 'inlineCode', function(node, index, parent) {
       if (parent.type !== 'link' && /dvc\s+[a-z-.]+/.test(node.value)) {
```
shcheklein: do we need to change the regex here to match three-parts nodes?

shcheklein:
https://dvc-org-pr-536.herokuapp.com/doc/commands-reference/remote/default - `dvc remote modify` has a wrong link link with `#`

src/Documentation/Markdown/utils/remark-linker.js
iAdramelk: It's basically just object search in the list with 200 nodes without computations. And we only run it at max twice per node with dvc command. Right now all pages have at max 10 commands, so this search run max 20 times. It should be pretty fast.

But I think that I can change Makrdown.js to `PureComponent` from `Component` to prevent rerenders without markdown content changing. This way search will be run only on initial render and on page change.

About why we need it. We basically have this possible cases:

1. `dvc command`.
2. `dvc command subcommand` - subcommand page exists. E. g. `dvc metrics add`.
3. `dvc command subcommand` - no subcommand page, but parent page has header with the command name. E. g. `dvc config cache ` (there is a #cache header on the config page).
4. `dvc command subcommand` - no subcommand page and parent page don't have corresponding header.

`getItemByPath` is used to check if subcommand page exists or not and if not set `#` instead. It's possible that there is no such anchor on the page, but unfortunately I can't check it without downloading .md file.

src/Documentation/Markdown/utils/remark-linker.js
iAdramelk: Not necessary. Right now this regexp just check if this is a `dvc` command or not. Breaking it to parts happens at the next line `let parts = node.value.split(/\s+/)`. We can add additional check to `part[2]` to check if it matches command syntax, but we can leave regexp itself as is.

PR #537, opened 2019-08-09
Fix: #525: How to fix autocomplete on debian/ubuntu

### LINKED ISSUE #525: autocomplete: fix instructions for enabling on Ubuntu 18.04 (bionic)
The instructions for enabling shell autocompletion in bash are not complete (https://dvc.org/doc/user-guide/autocomplete). It only says:
> On a current Linux OS (in a non-minimal installation), bash completion should be available;

This is true, but I also had to enable it by commenting out these lines on `~/.bashrc`:
```
# enable programmable completion features (you don't need to enable
# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
  if [ -f /usr/share/bash-completion/bash_completion ]; then
    . /usr/share/bash-completion/bash_completion
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi
```
I am actually using LinuxMint 19, but it is based on Ubuntu 18.04, so this behavior should be the same for all the **bionic** systems (Ubuntu 18.04 and all the variations based on it).

dashohoxha:
Instructions for fixing bash autocomplete on debian/ubuntu. Fix #525

### DIFF
--- a/static/docs/user-guide/autocomplete.md
+++ b/static/docs/user-guide/autocomplete.md
@@ -101,6 +101,7 @@ steps:
   ```
 
 - Exit from the shell and open a new one, or just reload `~/.bashrc`:
+
   ```dvc
   $ source ~/.bashrc
   ```
--- a/static/docs/user-guide/autocomplete.md
+++ b/static/docs/user-guide/autocomplete.md
@@ -77,13 +77,18 @@ completion.
 
 ### Click to expand if it doesn't work on Debian/Ubuntu
 
-As mentioned above, it should work out of the box. But if it does not, try these steps:
+As mentioned above, it should work out of the box. But if it does not, try these
+steps:
 
 - Make sure that the package `bash-completion` is installed:
-  ```bash
-  sudo apt install --reinstall bash-completion 
+
+  ```dvc
+  $ sudo apt install --reinstall bash-completion
   ```
-- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines are there:
+
+- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines
+  are there:
+
   ```bash
   # enable bash completion in interactive shells
   if ! shopt -oq posix; then
@@ -94,9 +99,10 @@ As mentioned above, it should work out of the box. But if it does not, try these
     fi
   fi
   ```
+
 - Exit from the shell and open a new one, or just reload `~/.bashrc`:
-  ```bash
-  source ~/.bashrc
+  ```dvc
+  $ source ~/.bashrc
   ```
 
 For more details see: https://linuxhandbook.com/enable-tab-completion/
--- a/static/docs/user-guide/autocomplete.md
+++ b/static/docs/user-guide/autocomplete.md
@@ -73,6 +73,36 @@ fi
 You can `source` your `~/.bash_profile` or launch a new terminal to utilize
 completion.
 
+<details>
+
+### Click to expand if it doesn't work on Debian/Ubuntu
+
+As mentioned above, it should work out of the box. But if it does not, try these steps:
+
+- Make sure that the package `bash-completion` is installed:
+  ```bash
+  sudo apt install --reinstall bash-completion 
+  ```
+- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines are there:
+  ```bash
+  # enable bash completion in interactive shells
+  if ! shopt -oq posix; then
+    if [ -f /usr/share/bash-completion/bash_completion ]; then
+      . /usr/share/bash-completion/bash_completion
+    elif [ -f /etc/bash_completion ]; then
+      . /etc/bash_completion
+    fi
+  fi
+  ```
+- Exit from the shell and open a new one, or just reload `~/.bashrc`:
+  ```bash
+  source ~/.bashrc
+  ```
+
+For more details see: https://linuxhandbook.com/enable-tab-completion/
+
+</details>
+
 ## Configure Zsh
 
 The DVC specific completion script is located in this path of our main


static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
...
+
+### Click to expand if it doesn't work on Debian/Ubuntu
+
+As mentioned above, it should work out of the box. But if it does not, try these steps:
+
+- Make sure that the package `bash-completion` is installed:
+  ```bash
+  sudo apt install --reinstall bash-completion 
+  ```
+- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines are there:
```
shcheklein: please, follow the documentation contributing guide to enable the pre-commit hook to indent and wrap things properly for you. And please follow the conventions mentioned there.

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
 You can `source` your `~/.bash_profile` or launch a new terminal to utilize
 completion.
 
+<details>
+
+### Click to expand if it doesn't work on Debian/Ubuntu
+
+As mentioned above, it should work out of the box. But if it does not, try these steps:
```
shcheklein: does not -> doesn't ... I think we take the "less official" path with those. Am I right, @jorgeorpinel ?

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
...
+
+- Make sure that the package `bash-completion` is installed:
+  ```bash
```
shcheklein: put extra space before and after code blocks

static/docs/user-guide/autocomplete.md
shcheklein: `it is` -> it's

static/docs/user-guide/autocomplete.md
jorgeorpinel: You are correct Ivan. I mean.. "You're correct" 😋 

I ty to use "does not" only in explanation about DVC commands behavior, and when special emphasis is needed.

TBH it's probably too much to expect perfect consistency with this very common term though, so either way should be OK.

static/docs/user-guide/autocomplete.md
jorgeorpinel: Also it should be ```dvc please, not `bash`. It's a peculiarity of our docs engine.

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
...
+  if ! shopt -oq posix; then
+    if [ -f /usr/share/bash-completion/bash_completion ]; then
+      . /usr/share/bash-completion/bash_completion
+    elif [ -f /etc/bash_completion ]; then
+      . /etc/bash_completion
+    fi
+  fi
+  ```
+- Exit from the shell and open a new one, or just reload `~/.bashrc`:
+  ```bash
```
jorgeorpinel: This one should also be ```dvc please, since it's a console command, not really `bash` language.

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
...
+
+- Make sure that the package `bash-completion` is installed:
+  ```bash
+  sudo apt install --reinstall bash-completion 
+  ```
+- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines are there:
+  ```bash
+  # enable bash completion in interactive shells
+  if ! shopt -oq posix; then
+    if [ -f /usr/share/bash-completion/bash_completion ]; then
```
jorgeorpinel: Unsure about this case of ```bash. Isn't that technically Shell script? I.e. ```sh

static/docs/user-guide/autocomplete.md
shcheklein: yep, only for cli. And use `$ cmd` syntax

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,36 @@ fi
...
+### Click to expand if it doesn't work on Debian/Ubuntu
+
+As mentioned above, it should work out of the box. But if it does not, try these steps:
+
+- Make sure that the package `bash-completion` is installed:
+  ```bash
+  sudo apt install --reinstall bash-completion 
+  ```
+- Make sure that it is enabled. Edit `~/.bashrc` and make sure that these lines are there:
+  ```bash
```
shcheklein: here `bash` or `sh` make sense

static/docs/user-guide/autocomplete.md
jorgeorpinel: Isn't that technically Shell script? I.e. ```sh
Up to you, I guess `bash` could make sense given the context.

static/docs/user-guide/autocomplete.md
shcheklein: or `shell` .. I'm not sure if the difference is important, check some other places in the docs. We should use the one the highlighter is on for.

dashohoxha:
I fixed the formatting (with prettier) and also fenced code blocks (although I think they are not necessary in this case, because there are no `dvc` commands).

But please let's keep "it is" and "does not" as they are. I know English as well as you do, and I know the difference between them and their short forms. It is not only about more/less formal but it is also about emphasis. They sound better to me the way they are.

static/docs/user-guide/autocomplete.md
```diff
@@ -73,6 +73,42 @@ fi
...
+    if [ -f /usr/share/bash-completion/bash_completion ]; then
+      . /usr/share/bash-completion/bash_completion
+    elif [ -f /etc/bash_completion ]; then
+      . /etc/bash_completion
+    fi
+  fi
+  ```
+
+- Exit from the shell and open a new one, or just reload `~/.bashrc`:
+  ```dvc
```
shcheklein: one more extra line above, please :)

dashohoxha: This somehow has escaped from Prettier (it is not infallible after all :) ).

jorgeorpinel:
My only comment on the "does not" and "is not" etc is that we want the docs, as much as possible, to sound like the same person wrote it all for familiarity. It's about consistency, not correctness. Thus I'm changing them in 2b96e0e (for my next PR). Thanks

PR #544, opened 2019-08-10
Small fix

dashohoxha:


dashohoxha:
> @dashohoxha please, check the documentation contributing guide. There are certain style we are enforcing. Please. install the pre-commit hook to do this for you.

I did it by clicking on "Edit on Github", as a normal user would do for submitting a minor change.
If this may break the rules of formatting, then probably we should remove this button from the page and discourage this way of contribution, because it seems that the contributor cannot run the Prettier in such a workflow.

By the way, the contribution page (https://dvc.org/doc/user-guide/contributing-documentation) says:

> Make sure you have the latest version of Node.js installed. Install and keep the dependencies up to date by running npm install often. This will also enable the Git pre-commit hook that will be formatting your code and documentation files automatically.

However I had to run prettier manually before committing (`npx prettier --write <file path(s)>`) because it didn't work when I staged the files. Could I be missing something?

shcheklein:
@dashohoxha 

> I did it by clicking on "Edit on Github", as a normal user would do for submitting a minor change.
If this may break the rules of formatting, then probably we should remove this button from the page and discourage this way of contribution, because it seems that the contributor cannot run the Prettier in such a workflow.

No easy workaround here. We can probably, eventually setup an automation to run the prettier and push an automated commit every time there are some changes in master. For now I usually do final style edits when it's an external contributor. All "internal" contributors use pre-commit.

> Could I be missing something?

Did you run `npm install`? Was it successful?

dashohoxha:
> Did you run `npm install`? Was it successful?

It was, but I also got some warnings about security issues and I had to do a kind of update. As a result `package-lock.json` has some differences, which seem to be higher versions of packages and some settings.

> No easy workaround here. We can probably, eventually setup an automation to run the prettier and push an automated commit every time there are some changes in master.

For small contributions this may not be necessary. We may also accept contributions that are not 100% compliant to the formatting rules, and once in a while (for example once a week) prettify all the doc pages. This can be easily done with a simple bash script.

PR #555, opened 2019-08-13
get-started: update Get Started accord to most recent example-get-started repo

### LINKED ISSUE #487: cmd ref: advanced examples for import (and get)

### LINKED ISSUE #554: lint: inconsistent fenced code block `-`line indentation
A problem after implementing #372 is that some lines inside md fenced code blocks starting with `-` get indented for unknown reasons by Prettier.

E.g. by finding the regex `^- cache` in all files we can see that static/docs/commands-reference/commit.md has the unindented lines:

image
> Similar cases in static/docs/get-started/connect-code-and-data.md and static/docs/tutorial/define-ml-pipeline.md

However static/docs/commands-reference/add.md has them indented:

image

I don't see a clear pattern explaining this apparently erratic behavior but it's all over the docs.

Again, this i work of Prettier which has been run on all md files in #512. None of the DVC-file samples had indentation before `-` lines originally. We decided that the indentation was OK in https://github.com/iterative/dvc.org/issues/512#issuecomment-518878493 but the fact that it indents some and unindents others troubles me a little. Perhaps we should use `<!-- prettier-ignore -->`s before all YAML code blocks after all?

All these code samples are still valid YAML though so please just close this issue if the decision is to live with it and let `prettier` do whatever it wants 😋 

---

Originally reported partially as [lint: don't indent fenced code block contents](https://github.com/iterative/dvc.org/issues/512)

jorgeorpinel:
Creating https://github.com/iterative/example-repos-dev and derivate repos.

Review app docs @ https://dvc-org-pr-555.herokuapp.com/

---

Also fix #554 (See https://github.com/iterative/dvc.org/pull/555#issuecomment-520871308)


jorgeorpinel:
Note: the last couple commits here are to fix #554

jorgeorpinel:
Review app docs @ https://dvc-org-pr-555.herokuapp.com/

static/docs/get-started/connect-code-and-data.md
```diff
@@ -48,17 +48,19 @@ recommend creating a virtual environment with a tool such as
 $ virtualenv -p python3 .env
 $ source .env/bin/activate
 $ echo ".env/" >> .gitignore
-$ pip install -r requirements.txt
+$ pip install -r src/requirements.txt
```
shcheklein: check other places in command references where we use example-get-started repo, sometimes we provide instruction on how to install it in the details section

static/docs/get-started/connect-code-and-data.md
```diff
@@ -48,17 +48,19 @@ recommend creating a virtual environment with a tool such as
...
 Save the progress to Git:
 
 ```dvc
 $ git add .
-$ git commit -m "add code"
+$ git commit -m "Add source code files to repo"
 ```
 
-</details>
+## Create a first data transformation stage
```
shcheklein: I would avoid using second level titles. If there is a temptation to use it it means that section has become too complicated. The ideal size of the chapter for the get started - one screen. It should be simple, just enough to give some sense of the idea.

static/docs/get-started/connect-code-and-data.md
jorgeorpinel: Another good catch, thanks! Fixed in 9885f13

static/docs/get-started/connect-code-and-data.md
jorgeorpinel: Updated in 60a998b per chat IRL.

PR #556, opened 2019-08-13
Script that runs 'npx prettier' on all .js and .md files.

dashohoxha:
Simple bash script that prettifies all the files (as discussed here: https://github.com/iterative/dvc.org/pull/544#issuecomment-520161436)


misc/prettier.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
shcheklein: - let's use `scripts` directory
- can we make accept a single file as well?
- let's update REAME + docs contributing guide to include this script?

dashohoxha: > can we make accept a single file as well?

For a single file it is simply `npx prettier --write <file>`. This is already documented on the docs contributing guide.

This script does it for all the files (which may not be so convenient to do with commands), and is intended to be run periodically (for example once a week or once a month), in order to fix any small contributions coming from people who click the button "Edit on GitHub", who cannot run the command `prettier` themselves. We have discussed it here: https://github.com/iterative/dvc.org/pull/544#issuecomment-520160121

dashohoxha: > let's update REAME + docs contributing guide to include this script?

Fixed the script and updated the guide.

PR #562, opened 2019-08-15
Removed wrong statement

dashohoxha:
The command `dvc add` is not related somehow to a `Dvcfile`; that is only for stages.

### DIFF
--- a/static/docs/commands-reference/add.md
+++ b/static/docs/commands-reference/add.md
@@ -98,9 +98,11 @@ and make your project reproducible.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
-- `-f`, `--file` - specify name of the DVC-file it generates. By default the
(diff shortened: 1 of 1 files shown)


static/docs/commands-reference/add.md
```diff
@@ -100,10 +100,7 @@ and make your project reproducible.
 
 - `-f`, `--file` - specify name of the DVC-file it generates. By default the
   DVC-file name generated is `<file>.dvc`, where `<file>` is file name of the
-  first output (from `targets`). The stage file is placed in the same directory
-  where `dvc run` is run by default, but `-f` can be used to change this
-  location, by including a path in the provided value (e.g.
-  `-f stages/stage.dvc`).
+  `<target>` (this option is not allowed with multiple targets).
```
shcheklein: it's important for user to understand that this option can change the location of the file generated

PR #572, opened 2019-08-19
Case insensitive matching with glossary terms

### LINKED ISSUE #547: glossary: case insensitive terms
The match terms (defined in `glossary.js`) should be case insensitive so that for example, both `Workspace` and `workspace` match "workspace" (without having to also specify Worspace in the match array.

> Extracted form https://github.com/iterative/dvc.org/issues/518#issuecomment-519812152

algomaster99:
Fixes #547

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -27,7 +27,12 @@ class Tooltip extends Component {
 
   componentDidMount() {
     glossary.contents.forEach((glossaryItem, index) => {
-      if (includes(glossaryItem.match.map(word => word.toLowerCase()), this.props.text.toLowerCase())) {
+      if (
+        includes(
+          glossaryItem.match.map(word => word.toLowerCase()),
+          this.props.text.toLowerCase()
+        )
+      ) {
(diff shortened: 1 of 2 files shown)


src/Tooltip/index.js
```diff
@@ -27,7 +27,7 @@ class Tooltip extends Component {
 
   componentDidMount() {
     glossary.contents.forEach((glossaryItem, index) => {
-      if (includes(glossaryItem.match, this.props.text)) {
+      if (includes(glossaryItem.match.map(word => word.toLowerCase()), this.props.text.toLowerCase())) {
```
shcheklein: please, run prettier or pre-commit to fix the style

jorgeorpinel:
If you could add <abbr> to some term with a different case than the value in the glossary `match` so we can test this on the review app (https://dvc-org-pr-572.herokuapp.com/doc) that would be nice.

For example `Output` in static/docs/commands-reference/status.md

shcheklein:
thanks, @algomaster99 !

PR #585, opened 2019-08-23
get-started: update sample commands using dataset-registry repo; other related changes

jorgeorpinel:
(Soon at) https://github.com/iterative/dataset-registry

Per https://github.com/iterative/dvc.org/issues/487#issuecomment-517520838


jorgeorpinel:
@shcheklein notice I had to update package-lick.json (ran `npm install`) because the app was crashing for me locally. The [deployment](https://dvc-org-pr-585.herokuapp.com/doc) seems good.

jorgeorpinel:
@shcheklein FYI only the 3rd and last commit (3b3a92f) here actually addresses the topic of this PR 😋 (to match the changes in https://github.com/iterative/example-repos-dev/pull/7). The 2nd commit is just a misc. doc update.

package-lock.json
```diff
@@ -3830,8 +3830,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
```
shcheklein: @jorgeorpinel what version of npm do you have? it looks like the difference here is only in some defaults. It can happen if a different version of npm is used.

static/docs/get-started/add-files.md
```diff
@@ -2,24 +2,21 @@
...
 DVC allows storing and versioning data files, ML models, directories,
 intermediate results with Git, without checking the file contents into Git.
-Let's get a sample dataset to play with:
+Let's get a dataset example to play with:
 
 ```dvc
 $ mkdir data
-$ wget https://data.dvc.org/get-started/data.xml -O data/data.xml
+$ cd data
+dvc import https://github.com/iterative/dataset-registry \
```
shcheklein: the problem with this is that we don't need `dvc add` after that, here we can use `dvc get` and clarify that it's similar to `wget` but allows to use Git repo as a source ... for import we need to come with a separate section or use it in some other place ...

package-lock.json
jorgeorpinel: 6.4.1

jorgeorpinel: How do you manage your Node version? Seems like the official way to upgrade is to download the installer every time... Have you tried [n](https://www.npmjs.com/package/n)?

iAdramelk: @shcheklein Interesting. What I can get from the changes is that `package-lock.json` on my machine marked dependencies for `fsevents` as optional, but this one marked them as required. Question is why it marked them as optional in the first place.

`fsevents` is a library used on MacOS to detect file changes then you for example need to rebuild your bundle is some files in FS changes. It is used only on Mac, requires creating C bindings with `node-gyp` and if bindings are not created it can usually still work without them using manual polling. So yes, if @jorgeorpinel is on Mac and tried to run `npm run dev` it can potentially break it.

I don't have a good explanation to why have they been marked optional in the first place through. According to `git blame` it was added in my commit f9c1415e5aac0b262279570dd4c8bffa121f7aa3, but this commit only added `lodash.debounce` so theoretically it shouldn't affect `fsevents` dependencies at all. There is also no info about package-lock.json bugs in `npm` changelog.

So yes, this version without `optional` flag is correct version and we should used it, there is no need to change npm vesrion.

About what to do with it to prevent bugs like hat in the future my only idea in to move to `yarn` from `npm`. We are using yarn and `yarn.lock` files on our projects by default for the last two years or so and there was no such problems with them.

About updating node and npm, and using different node version: On Mac and Linux using package managers like https://brew.sh/ should solve problem with updates. For Windows there is https://chocolatey.org/ , but I never used it myself, so I can't comment on it. And to have multiple versions at the same time we usually use https://github.com/nvm-sh/nvm and not `n`.

iAdramelk: Found relevant issue in NPM by the way: https://github.com/npm/npm/issues/17722

iAdramelk: Created ticket #595

jorgeorpinel:
Note that my last commit here moves the https://dvc.org/doc/get-started/example-pipeline datasets into https://github.com/iterative/dataset-registry (corresponding to https://github.com/iterative/dataset-registry/commit/7476a858f6200864b5755863c729bff41d0fb045)

jorgeorpinel:
Alright, I finished reviewing the existing Get Started [Example: Pipelines](https://dvc.org/doc/get-started/example-pipeline) in this PR as well. Please review! Hopefully we can merge this so I can move on with #487 🙂

static/docs/get-started/add-files.md
```diff
@@ -2,24 +2,22 @@
...
 DVC allows storing and versioning data files, ML models, directories,
 intermediate results with Git, without checking the file contents into Git.
-Let's get a sample dataset to play with:
+Let's get a dataset example to play with:
 
 ```dvc
 $ mkdir data
-$ wget https://data.dvc.org/get-started/data.xml -O data/data.xml
+$ cd data
+$ dvc get https://github.com/iterative/dataset-registry \
```
shcheklein: maybe just name it iterative/datasets?

static/docs/get-started/example-versioning.md
```diff
@@ -220,7 +230,7 @@ Let's commit the second version:
 
 ```dvc
 $ git add model.h5.dvc data.dvc metrics.json
-$ git commit -m "model second version, 2000 images"
+$ git commit -m "Second model, trained with2000 images"
```
shcheklein: `with 2000`

package-lock.json
shcheklein: Should be resolved by migrating to yarn. Please, merge and resolve the conflicts.

static/docs/get-started/data-artifacts.md
```diff
@@ -0,0 +1,45 @@
...
+inspects
+[get-started/data.xml.dvc](https://github.com/iterative/dataset-registry/blob/master/get-started/data.xml.dvc)
+and tries to retrieve the file using the external project's default remote
+(configured
+[here](https://github.com/iterative/dataset-registry/blob/master/.dvc/config)).
+
+</details>
+
+This is similar to manually downloading the data and using `dvc add` to track it
+with DVC. With an import stage however, we can run `dvc update` to checks for
```
shcheklein: `to check` typo

static/docs/get-started/data-artifacts.md
```diff
@@ -0,0 +1,45 @@
+# Reusing Data Artifacts
```
shcheklein: let's skip artifacts .. how about just `reusing data` (I still don't like the name, haha :) )

static/docs/get-started/data-artifacts.md
```diff
@@ -0,0 +1,45 @@
...
+That's why we created `dvc import`! This commands can download data from any DVC
+project hosted on a Git repository. It also tracks the imported data by creating
+an import stage ([DVC-file](/doc/user-guide/dvc-file-format)). Example:
+
+```dvc
+$ dvc import https://github.com/iterative/dataset-registry \
+           get-started/data.xml
+```
+
+This downloads `data.xml` (adding it to `.gitignore`) and creates the
```
shcheklein: Not sure how does it fit in the Get Started "story" .. this command will create a DVC-file that won't be used anywhere, right?

static/docs/get-started/data-artifacts.md
```diff
@@ -0,0 +1,35 @@
+# Data Artifacts
```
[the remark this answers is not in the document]
jorgeorpinel: I assume this name is OK.

static/docs/get-started/data-artifacts.md
```diff
@@ -0,0 +1,35 @@
...
+and tries to retrieve the file using the external project's default remote
+(configured
+[here](https://github.com/iterative/dataset-registry/blob/master/.dvc/config)).
+
+Finally, `dvc update` allows us to "force" updating the imported data without
+having to reproduce the pipeline.
+
+```dvc
+$ dvc update data.xml.dvc
+```
```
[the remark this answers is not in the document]
jorgeorpinel: I presume the new version is more acceptable.

static/docs/get-started/add-files.md
jorgeorpinel: Maybe. Opened https://github.com/iterative/dataset-registry/issues/1

static/docs/get-started/example-versioning.md
jorgeorpinel: Fixed

static/docs/get-started/data-artifacts.md
jorgeorpinel: Fixed

static/docs/get-started/data-artifacts.md
jorgeorpinel: Sure, changed. Maybe **Reusing data among different projects" ?

static/docs/get-started/data-artifacts.md
jorgeorpinel: Good point, didn't think about that... How about just adding a note like at the bottom of https://dvc.org/doc/get-started/configure? Suggested change in 1b3bdfa

static/docs/get-started/example-pipeline.md
```diff
@@ -246,122 +252,123 @@ actual commands instead of DVC-files):
...
-transparently.
+The key detail to notice is that DVC automatically derives the dependencies
+between the pipeline stages and builds a dependency graph (DAG) transparently.
 
-Not only can DVC streamline your work into a single, reproducible environment,
-it also makes it easy to share this environment by Git including the
-dependencies — an exciting collaboration feature which gives the ability to
-reproduce the research easily in a myriad of environments.
+DVC streamlines all of your experiments into a single, reproducible
+<abbr>project</abbr>, and it makes it easy to share it on Git, including the
```
shcheklein: not sure you can say `on Git`

static/docs/get-started/example-pipeline.md
```diff
@@ -246,122 +252,123 @@ actual commands instead of DVC-files):
...
+dependencies. This is an innovative collaboration feature which provides the
```
shcheklein: let's remove `innovative` as well. This "marketing"-style (like myriad of environments as well) language looks not appropriate for the get started and tutorials, let's save it for blog posts :)

static/docs/get-started/reusing-data.md
```diff
@@ -0,0 +1,48 @@
...
+
+> This command is only shown for informational purposes. No need to actually run
+> it in order to continue with this guide.
+
+This downloads `data.xml`, adds it to `.gitignore`, and creates the
+`data.xml.dvc` stage file.
+
+<details>
+
+### Expand to learn about how importing works
```
shcheklein: title is a bit misleading? Explanation covers only one specific detail.

shcheklein: it also out of place here - should be before get?

static/docs/get-started/reusing-data.md
```diff
@@ -0,0 +1,48 @@
...
+That's why we created `dvc import`! This commands can download data from any DVC
+project hosted on a Git repository. It also tracks the imported data by creating
+an import stage ([DVC-file](/doc/user-guide/dvc-file-format)). Example:
+
+```dvc
+$ dvc import https://github.com/iterative/dataset-registry \
+           get-started/data.xml
+```
+
+> This command is only shown for informational purposes. No need to actually run
```
shcheklein: may be we can say something like "Alternatively, get started could have been initialized with `dvc import` to ..." and actually show that path (remove dvc-file and data/data.xml, etc)

static/docs/get-started/reusing-data.md
```diff
@@ -0,0 +1,48 @@
...
+
+This is similar to manually downloading the data and using `dvc add` to track it
+with DVC. With an import stage however, we can run `dvc update` to check for
+updates in the external data source before
+[reproducing](/doc/get-started/reproduce) the <abbr>pipeline</abbr> that depends
+on the import.
+
+> `dvc get` can also download a file from an external DVC project, as shown in
+> the [Add Files](/doc/get-started/add-files) chapter. But it does not track the
+> downloaded data. In fact this command can be used outside of a DVC project.
```
shcheklein: `dvc import` does not track data either. `dvc import` just keeps the connection between the projects.

jorgeorpinel:
> package-lock.json should not be part of the project anymore

Fixed. Deployment now works. Will address the remaining comments ASAP. Please notice there are some questions still open above: https://github.com/iterative/dvc.org/pull/585#discussion_r319781831 and https://github.com/iterative/dvc.org/pull/585#discussion_r319781968.

static/docs/get-started/data-artifacts.md
shcheklein: May be we can say something like - "An alternative way to initialize the project would be to use `dvc import` ... " and actually show how `add` can be replaced with import and what are the benefits?

static/docs/get-started/data-artifacts.md
shcheklein: nah, let's keep just reusing data for now .. will figure out something else later if needed

static/docs/get-started/add-files.md
```diff
@@ -2,24 +2,22 @@
 
 DVC allows storing and versioning data files, ML models, directories,
 intermediate results with Git, without checking the file contents into Git.
-Let's get a sample dataset to play with:
+Let's get a dataset example to play with:
 
 ```dvc
 $ mkdir data
-$ wget https://data.dvc.org/get-started/data.xml -O data/data.xml
+$ cd data
```
shcheklein: let's use `-o` option for the `dvc get` instead of doing `cd data ... cd ..`

static/docs/get-started/data-artifacts.md
jorgeorpinel: Yes, great idea. Improved in cb42034.

static/docs/get-started/example-pipeline.md
jorgeorpinel: True, it's not a platform. Fixed

static/docs/get-started/example-pipeline.md
jorgeorpinel: OK. It was "exciting" before which I also thought was more like marketing. Removed altogether now.

static/docs/get-started/reusing-data.md
jorgeorpinel: Previously addressed.

static/docs/get-started/add-files.md
jorgeorpinel: Yes. Also addressed previously.

static/docs/get-started/reusing-data.md
jorgeorpinel: Hmmm yes you're right... OK, I just rectified this in my latest commit.

static/docs/get-started/add-files.md
```diff
@@ -26,6 +23,10 @@ example:
 $ dvc add data/data.xml
 ```
 
+> For an alternative and faster way to download the project's dataset in a
```
shcheklein: I think this is misleading - import serves a different purpose, it's not just a faster (btw why is it faster?) way. Also, I would not include too many notes, especially that go forward. Let's keep it simple for now.

static/docs/get-started/reusing-data.md
```diff
@@ -1,26 +1,33 @@
...
 
 That's why we created `dvc import`! This commands can download data from any DVC
 project hosted on a Git repository. It also tracks the imported data by creating
-an import stage ([DVC-file](/doc/user-guide/dvc-file-format)). Example:
+an import stage ([DVC-file](/doc/user-guide/dvc-file-format)).
+
+If you remember the [Add Files](/doc/get-started/add-files) chapter where we
+download the raw data for this example project, the `dvc get` command is used
+(which unlike `dvc import` does not track the downloaded artifact). Let's say we
+wanted to download and track the data with a single command:
```
shcheklein: not sure again that it's right to emphasize that it's a single command vs two commands here. It's not that important as a difference.

static/docs/get-started/reusing-data.md
```diff
@@ -8,17 +8,26 @@ different machines). But what if we wanted to reuse a dataset from a project in
 a different one?
 
 That's why we created `dvc import`! This commands can download data from any DVC
-project hosted on a Git repository. It also tracks the imported data by creating
-an import stage ([DVC-file](/doc/user-guide/dvc-file-format)).
+project hosted on a Git repository. It also records the connection between this
+and the external project by creating an import stage
+([DVC-file](/doc/user-guide/dvc-file-format)). To track the the imported data,
```
shcheklein: the the -> the

static/docs/get-started/reusing-data.md
shcheklein: I see that this is addressed already

static/docs/commands-reference/get.md
```diff
@@ -53,15 +53,16 @@ created in the current working directory, with its original file name.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
-## Examples
+## Example: Machine learning model deployment
```
shcheklein: again, I think we should make titles for examples less specific ... at least for this one, we need to provide one very generic example

static/docs/get-started/example-versioning.md
```diff
@@ -295,8 +305,16 @@ our example, please notice that `train.py` produces binary files (e.g.
...
+> the [Switching between versions](#switching-between-versions) section with:
+>
+> ```dvc
+> $ git checkout master
+> $ dvc checkout
+> ```
+
 ```dvc
-$ dvc remove -p model.h5.dvc
+$ dvc remove -pf model.h5.dvc
```
shcheklein: typo?

static/docs/get-started/reusing-data.md
jorgeorpinel: True. Updated title AND added more useful details about the importing process.

static/docs/commands-reference/get.md
jorgeorpinel: > Also replying to comment from https://github.com/iterative/dvc.org/commit/8e9e94725675e5285d4d36f0a4fd8ae77c45e95b#r35084861:
>
> > let's use something more generic - Getting (or accessing? or downloading) model or data... Deployment scenario is too specific

OK. You had asked for that specific example before I think 😋 but agree, it's better as a side note, which I added in 5c3812e.

static/docs/commands-reference/get.md
```diff
@@ -91,3 +92,65 @@ can be automated leveraging DVC with
 The same example applies to raw or intermediate data files as well, of course,
 for cases where we want to download those files and perform some analysis on
 them.
+
+## Example: Compare different versions of the same experiment
```
jorgeorpinel: > Compare different versions of data or model?

Also updated in 5c3812e.

static/docs/commands-reference/get.md
```diff
@@ -91,3 +92,65 @@ can be automated leveraging DVC with
...
+know that each chapter has a corresponding
+[tag](https://github.com/iterative/example-get-started/tags). Tag `7-train` is
+where we train a first version of the example model, and tag `9-bigrams-model`
+has an improved model (trained using bigrams). What if we wanted to have both
+versions of the model "checked out" at the same time? `dvc get` provides an easy
+way to do this:
+
+```dvc
+$ dvc get . model.pkl --rev 7-train --out model.monograms.pkl
+```
```
jorgeorpinel: > since we cloned it, can we do `dvc get . --rev ...`?

Yep, thanks. I had already noticed and improved that one.

static/docs/get-started/add-files.md
jorgeorpinel: OK, removed note (:

jorgeorpinel: p.s. by "faster" I meant that typing a single command consumes less of the user's time.

static/docs/get-started/example-versioning.md
jorgeorpinel: Hmm no, that was part of my review of actually trying the example. without `-f` the sample would need to include more the confirmation prompt and somehow indicate to the user to answer it.

static/docs/get-started/reusing-data.md
jorgeorpinel: Fixed, thanks.

static/docs/tutorial/define-ml-pipeline.md
```diff
@@ -11,9 +11,9 @@ files and download a 40MB data archive into this directory.
 ### Expand to learn how to download on Windows
 
 Windows doesn't include the `wget` utility by default, so you'll need to use the
-browser to download `data.xml`. Save it into the `data` subdirectory.
+browser to download `data.xml`. Save it into the `data/` subdirectory.
 (Right-click [this link](https://data.dvc.org/tutorial/ver/data.zip) and click
-`Save link as`(Chrome) or `Save object as`(Firefox)).
+`Save Link As...`(Chrome).
```
shcheklein: why removing the Firefox version? :)

static/docs/get-started/example-versioning.md
shcheklein: 👍

static/docs/tutorial/define-ml-pipeline.md
jorgeorpinel: I couldn't find the exact text that the Firefox context menu has corresponding to "Save Link As...". I was going to install it to check and then I realized FF has fallen to under 10% global browser usage. Google Chrome is totally dominating the market! So I decided it's not really worth pointing out FF (over other small options at least). https://www.w3schools.com/browsers/default.asp

But lmk if you prefer it back and I'll find out the exact current text for this function.

shcheklein: 👍

PR #597, opened 2019-08-28
Replace npm with yarn

iAdramelk:
- Update contributing guide
- Replace package-lock.json with yarn.lock
- Update CircleCI commands

Heroku and CircleCI images includes yarn out of the box so there is no need to update them.


iAdramelk:
> Question: just to make sure does it support pre-commit hook via husky that we have specified?

Yep, it sets them correctly. We use them on our project all the time.

shcheklein:
@jorgeorpinel @dashohoxha @algomaster99 mind this change, guys! please let us know if you hit any issues.

dashohoxha:
On Ununtu-18.04 I get these warnings:
```bash
$ yarn
yarn install v1.17.3
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "linux" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > react-collapsible@2.6.0" has unmet peer dependency "prop-types@^15.5.8".
warning " > react-ga@2.6.0" has unmet peer dependency "prop-types@^15.6.0".
warning " > babel-jest@24.9.0" has unmet peer dependency "@babel/core@^7.0.0".
warning "babel-jest > babel-preset-jest@24.9.0" has unmet peer dependency "@babel/core@^7.0.0".
[4/4] Building fresh packages...
Done in 63.36s.

$ yarn
yarn install v1.17.3
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.81s.
```

I hope they are normal, because everything else (for example `yarn dev`) works as expected.

iAdramelk:
@dashohoxha Thanks! They are installed via submodules of other packages right now, so there should not be any practical problems. But to prevent problems in the future it is better to explicitly add them. Made it in separate PR because this was already merged: https://github.com/iterative/dvc.org/pull/600

jorgeorpinel:
Works well for me.

Can you remind us what benefits does this bring though? Is it just for the MacOS inconsistencies when running `npm install`?

iAdramelk:
@jorgeorpinel Not exactly on Mac. Problem looks like this:

1. Some dependencies in `package.json` can be optional and only used on one OS and not used on other OSes. In our case this dependency is `fsevents` – library to detect file changes on Mac OS.
2. When `npm install` in run on Mac `fsevents` is correctly marked as required. But then somebody run `npm install` on Linux or Windows after that in some cases npm updates `package-lock.json` and mark `fsevents` as optional (because it is not needed on this operating systems).
3. Now if Windows/Linux user commit this changes, it can break builds for users on Mac, because `fsevents` will not be installed.

It can potentially be any library and any OS combination, not only `fsevents` and Mac OS.

Most other differences between `yarn` and `npm` are not as important right now as they where important in the past. But on plus side we have better interface for updating dependencies with `yarn upgrade-interactive` and `yarn upgrade-interactive --latest`, but this is not that important.

jorgeorpinel:
I see, thanks. So `yarn` always mark `fsevents` as required even when it's only needed on Mac?

This is what I see from my Mac:
```console
$ yarn why fsevents
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "fsevents"...?
...
info Reasons this module exists
   - "babel-jest#@jest#transform#jest-haste-map" depends on it
   - Hoisted from "babel-jest#@jest#transform#jest-haste-map#fsevents"
   - Hoisted from "next#webpack#watchpack#chokidar#fsevents"
```

iAdramelk:
@jorgeorpinel yep, `yarn` will always add `fsevents` to the `yarn.lock`.

PR #591, opened 2019-08-26
regular updates and fixes (late August) (2)

jorgeorpinel:

Issue #540, opened 2019-08-10
term: SCM -> Git


jorgeorpinel:
Git is the only SCM supported on DVC and probably will be for a long time. Also, most of the docs use likely Git-specific language such as "tags" and "branches", as well as `git ...` commands in sample code blocks.

- [ ] Try to get rid of "SCM", except perhaps in very general descriptions of data versioning problems. Use "Git" or \`git\` otherwise.
- [ ] Review also sentences/paragraphs where "tag" and "branch" is used for terminology correctness.

dashohoxha:
Also I think that it might be better if the option `--no-scm` of `dvc init` is renamed to `--no-git`, since Git is the only SCM supported and there are no plans (yet) to support any other SCM.

shcheklein:
@dashohoxha good point, we can create an alias probably. Create aa ticket with this suggestion on DVC core repo, let's see what the @iterative/engineering thinks about this.

jorgeorpinel:
Yes, this issue I wrote a while ago before it was decided that only Git would be supported (for the time being). Now I think we should replace all mentions of "SCM" for "Git" or `git`.

Zenix27:
I can work on this.

jorgeorpinel:
Sure @Zenix27 this is a simple one to test our process, thanks! Please make sure you've read https://dvc.org/doc/user-guide/contributing/docs first.

srishti-nema:
Hi @jorgeorpinel , is this issue available?

pared:
Hi @srishti-nema! Seems like as of today, @Zenix27 is working on it now. @Zenix27 can you confirm?

Zenix27:
@pared Not now, @srishti-nema  you can move forward on this for now, I am working on understanding the repro.

pared:
@jorgeorpinel can we assign @srishti-nema to that one?

jorgeorpinel:
Sure, go for it @srishti-nema, thanks.

PR #601, opened 2019-09-01
regular updates and fixes (early September)

### LINKED ISSUE #448: term: clarifications (Aug)
~~Continuation of #432~~

- [x] "remote cache" -> "remote storage"
- [x] Avoid "local cache". (It may be external). Use "cache directory" or maybe "project cache" instead.
- ~~Review usage of "DVC" branding of terms (i/[^`]DVC / regex): Avoid "DVC cache, "DVC pipeline" (others) but keep when necessary, specially "DVC project", "DVC remote" (others)~~
  UPDATE: Done up to static/docs/commands-reference/metrics, moved to #608.
- [x] Review usage of term "step" (use "stage" when referring to pipelines).
- [x] "step by step" -> "step-by-step"
- [x] Use either "dataset(s)" or "data set(s)".
- [x] Find regex `r[au]ns? again`, consider term "regenerate" instead.
- [x] Review usage of "chain of command(s)". It does not mean what we intend it to. Check also the term "graph" and related (DAG).

jorgeorpinel:
Fix #448, and brings other ad hoc changes.


static/docs/commands-reference/get.md
```diff
@@ -42,7 +42,8 @@ created in the current working directory, with its original file name.
   isn't used) is the current working directory (`.`) and original file name.
 
 - `--rev` - specific Git revision of the DVC repository to import the data from.
-  `HEAD` by default.
+  [`HEAD`](https://git-scm.com/book/en/v2/Git-Internals-Git-References#ref_the_ref)
+  is used by default when this option is not specified.
```
jorgeorpinel: > Moved from https://github.com/iterative/dvc.org/pull/591#discussion_r319361679

Maybe it's more correct to say `master`? Or "tip of the default branch", @efiop?

static/docs/commands-reference/install.md
```diff
@@ -285,6 +285,6 @@ Data and pipelines are up to date.
 
 After reproducing this pipeline up to the "evaluate" stage, the data files are
 in sync with the code/config files, but we must now commit the changes to the
-Git repository. Looking closely we see that `dvc status` is run again, informing
-us that the data files are synchronized with the `Pipelines are up to date.`
-message.
+Git repository. Looking closely we see that `dvc status` is used again,
+informing us that the data files are synchronized with the
+`Pipelines are up to date.` message.
```
shcheklein: please, check the message. I think it's a different one now.

shcheklein:
run -> regenerate. Not sure this is a good idea after all. It sounds like stage files are being regenerated in this context w/o specifying explicitly the that outputs are regenerated. So, wither we should come up with something else or keep the previous one. @jorgeorpinel what was the motivation for this change, btw?

static/docs/commands-reference/commit.md
```diff
@@ -49,23 +49,23 @@ DVC can't guarantee reproducibility in those cases – You commit any data you
...
+to the <abbr>cache</abbr> after creating a DVC-file. What _commit_ means is that
+DVC:
 
 - Computes a checksum for the file/directory
 - Enters the checksum and file name into the DVC-file
 - Tells the SCM to ignore the file/directory (e.g. add entry to `.gitignore`)
   (Note that if the <abbr>workspace</abbr> was initialized with no SCM support
   (`dvc init --no-scm`), this does not happen.)
-- Adds the file/directory or to the DVC cache
+- Adds the file/directory or to the cache directory
```
shcheklein: Why should we explicitly specify that this is a directory?

shcheklein: it sounds not like we always just copy files to some directory, while it's more complicated

shcheklein: I would try to use cache directory when we explaining it's structure, or some internal and it's clear from the context that there is some directory indeed

static/docs/tutorial/index.md
```diff
@@ -25,7 +25,7 @@ and this approach will not require storing binary files in your Git repository.
 
 ## DVC Workflow
 
-The diagram below describes all the DVC commands and relationships between local
-cache and remote storage.
+The diagram below describes all the DVC commands and relationships between a
+local cache and remote storage.
```
shcheklein: local - forgot to remove?

static/docs/commands-reference/get.md
jorgeorpinel: Ping @efiop

static/docs/commands-reference/install.md
jorgeorpinel: True. Updated to `Data and pipelines are up to date.` in 1665812.

jorgeorpinel:
@shcheklein 

> run -> regenerate. Not sure this is a good idea after all... what was the motivation for this change, btw?

It's one of the checkboxes I added in #448. The reason is that I never liked the previous phrase "run again", which was a replacement for "rerun".

I'm using this term in the context of `dvc repro`, in addition to the "reproduce", of course, but you don't always want to explain or define `repro` with the same word. It should mostly refer to the outputs or results of pipeline stages though, for stage commands you're right that it's not correct so I'm using the word "execute" instead. I made some wording improvements in 492cfc6. What do you think?

static/docs/get-started/reproduce.md
```diff
@@ -1,11 +1,12 @@
...
 
 In the previous chapters, we described our first
-[pipeline]](/doc/commands-reference/pipeline). Basically, we created a number of
-[stage files](/doc/commands-reference/run). Each of these
-[DVC-files](/doc/user-guide/dvc-file-format) describes single stage we need to
-run towards a final result (a pipeline). Each depends on some data (either raw
-data files or intermediate results from previous stages) and code files.
+[pipeline]](/doc/commands-reference/pipeline). Basically, we generated a number
+of [stage files](/doc/commands-reference/run)
+([DVC-files](/doc/user-guide/dvc-file-format)). Each of these stages define
```
shcheklein: `defines a single` ?

shcheklein: defines a single command

static/docs/commands-reference/run.md
```diff
@@ -125,12 +125,11 @@ pipeline.
   for confirmation.
 
 - `--ignore-build-cache` - if an exactly equal DVC-file exists (same list of
-  outputs and inputs, the same command to run) which has been already executed,
-  and is up to date, with option `dvc run` won't execute the command again by
-  default (thus "build cache"). This option gives a way to forcefully run the
-  command anyway. It's useful if the command is considered non-deterministic for
-  some reason (meaning it produces different outputs from the same list of
-  inputs).
+  outputs and inputs, the same command to run which has been already executed),
```
shcheklein: I think it's import that ) is after run here

static/docs/commands-reference/add.md
```diff
@@ -16,31 +16,34 @@ positional arguments:
...
+[external outputs](/doc/user-guide/external-outputs).
 
-The result is data file is added to the DVC cache, and DVC-files can be tracked
-via Git or other version control system. The DVC-file lists the added file as an
-output (`out`), and references the DVC cache entry using the checksum. See
+Unless the `-f` options is used, the DVC-file name generated is `<file>.dvc` by
+default, where `<file>` is file name of the first output (from `targets`).
+
+The result is data file is placed in the cache directory, and DVC-files can be
+tracked via SCM. The DVC-file lists the added file as an output (`outs` field),
```
shcheklein: SCM -> Git?

static/docs/commands-reference/add.md
```diff
@@ -16,31 +16,34 @@ positional arguments:
...
-Unless the `-f` options is used, by default the DVC-file name generated is
-`<file>.dvc`, where `<file>` is file name of the first output (from `targets`).
+Note that `targets` outside the current workspace are supported, creating
+[external outputs](/doc/user-guide/external-outputs).
 
-The result is data file is added to the DVC cache, and DVC-files can be tracked
-via Git or other version control system. The DVC-file lists the added file as an
-output (`out`), and references the DVC cache entry using the checksum. See
+Unless the `-f` options is used, the DVC-file name generated is `<file>.dvc` by
+default, where `<file>` is file name of the first output (from `targets`).
```
shcheklein: `output` - may be just first target? I'm not sure we should use output term here.

static/docs/commands-reference/commit.md
jorgeorpinel: Hmmm good points. I was just reviewing the DVC "branding" of terms like DVC project and DVC remote, trying to avoid it sometimes. I'm going to just reduce this term to "cache" or some times "project's cache" except when explaining cache internals or when related to the `dvc cache dir` command (and corresponding config option).

jorgeorpinel: Done! (Along with related improvements) in 3f7884b.

static/docs/commands-reference/get.md
jorgeorpinel: OK well, I went for "tip of the default branch" in 5fe82e7.

static/docs/get-started/reproduce.md
jorgeorpinel: > defines a single command

Yes that is more correct syntax, thanks. But I wend for "These stages define individual commands to execute" because I think we want to talk about the commands in plural. See a5019ef.

static/docs/commands-reference/run.md
jorgeorpinel: Good catch, thanks. Fixed in 46aa961

static/docs/commands-reference/add.md
jorgeorpinel: Already fixed, but note that this has its own issue: #540.

static/docs/commands-reference/add.md
jorgeorpinel: This is an interesting question. Firstly, that whole sentence ("Unless the `-f` option... first output (from `targets`)." was already in the document, a bit further down. I just thought the explanation made more sense up here.

That said, I also noticed the word "output" isn't very helpful the way its currently used in this document. BUT I think it actually is very important to use it here, maybe just explain earlier and better that `dvc add` DVC-files are special because they define outputs so the `targets` become outputs.

Any further thoughts? Let me see if I can improve it...

jorgeorpinel: Please see the changes in 8753d05. Is it better explained now?

jorgeorpinel:
Sorry for the delay here @shcheklein. I was a bit stuck in a bunch of other discussions and trying to finish all of #448 here... So many file changes... Sry 😅

static/docs/tutorial/index.md
jorgeorpinel: No, this one I left on purpose because the image below reads "Local Cache":

image

jorgeorpinel:
p.s. I'm done adding commits here. Maybe one more involving the term "project metrics" but that can always go into a future PR.

static/docs/commands-reference/repro.md
```diff
@@ -166,7 +166,7 @@ $ tree
 ├── count.txt      <---- result: "2"
 ├── filter.dvc     <---- first stage
 ├── numbers.txt    <---- intermediate result of the first stage
-├── process.py     <---- code that runs some transformation
+├── process.py     <---- code that causes data transformation
```
shcheklein: causes is not the best term either. If we want to change this we can use `implements` for example or even simple `does`.

static/docs/tutorial/index.md
shcheklein: 👍

shcheklein:
huge stuff :) congrats 🎉

static/docs/commands-reference/repro.md
jorgeorpinel: I see you've fixed this already (:  Thanks

PR #602, opened 2019-09-02
Trim newline characters in tooltip texts

### LINKED ISSUE #518: glossary: no tooltip when there's break lines in <abbr> tags
For example the lines

```md
- `etag`: strong ETag response header (only HTTP <abbr>external
  dependencies</abbr> created with `dvc import-url`)
```

won't produce a glossary tooltip. But they should.

If we require `<abbr>` terms to be in a single line, then maybe the automatic linter (`pretty-quick`) that runs on a Git pre-commit hook should enforce that format, like it enforces `quoted text` to be in a single line?

algomaster99:
Fixes #518 

To test these changes, please edit the `add.md` file as follows
Screenshot from 2019-09-02 17-23-04

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -32,7 +32,7 @@ class Tooltip extends Component {
           glossaryItem.match.map(word => word.toLowerCase()),
           // In v4 text field in a React.Node,
           // so to get string we need to use it's children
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Thanks a lot 🎉

PR #604, opened 2019-09-02
Fix #598: Hidden items in sidebar if submenu is opened

### LINKED ISSUE #598: opening pipeline "hides" version item
Open https://dvc.org/doc/commands-reference/pipeline. You won't see `version` command reference at the end of the list.

iAdramelk:
Problem with the current version is an incorrect calculation of opened item height (we need to supply height to make animations work). Right now height is calculated without considering height of opened children. This PR fix it.

P. S. Old branch had name with `$` symbol which prevented circle.ci from working. I recreated PR with correct name.

### DIFF
--- a/src/Documentation/SidebarMenu/SidebarMenu.js
+++ b/src/Documentation/SidebarMenu/SidebarMenu.js
@@ -24,8 +24,6 @@ function calculateHeight({ activePaths, path }) {
   for (let i = 0; i < reversePaths.length; i++) {
     const current = reversePaths[i]
 
-    console.log(current, heightMap[current])
-
     height += heightMap[current]
 
     if (path === current) break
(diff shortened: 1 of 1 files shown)


src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -12,21 +12,37 @@ import { getParentsListFromPath } from './helper'
...
+
+// Helper function to calculate element height with all opened children
+function calculateHeight({ activePaths, path }) {
+  let height = 0
+  const reversePaths = [...activePaths].reverse()
+
+  for (let i = 0; i < reversePaths.length; i++) {
+    const current = reversePaths[i]
+
+    console.log(current, heightMap[current])
```
shcheklein: remove this?

iAdramelk: Oops

shcheklein:
thanks, @iAdramelk !

PR #607, opened 2019-09-03
Update preparation.md

lion19:
* reflect the fact that requirements.txt was unzipped to a `code/` directory


jorgeorpinel:
Yes, this is correct! Thanks  @lion19 🎉

PR #610, opened 2019-09-06
Next update with possible fix for global styles blinking

iAdramelk:
- Update next.js to 9.x
- Move redirects to `server.js`.
- Move global styles and reset to layout component instead of `_document.js`.


server.js
```diff
@@ -21,7 +21,15 @@ app.prepare().then(() => {
...
     // Special URL host redirects
-    if (req.headers.host === 'man.dvc.org') {
+    if (
+      req.headers['host'].match(/^www/) !== null ||
+      (req.headers['x-forwarded-proto'] !== 'https' && !dev)
+    ) {
+      res.writeHead(301, {
+        Location:
+          'https://' + req.headers['host'].replace(/^www\./, '') + req.url
+      })
```
shcheklein: `res.end()`  is missing?

iAdramelk: Yes, my bad. Added it.

shcheklein:
thanks, @iAdramelk !

Issue #461, opened 2019-06-27
glossary: add <abbr> tags everywhere!


jorgeorpinel:
~~Once #431 is merged,~~

There will be an initial glossary.js file with a set of terms we want to add `<abbr>` tags around in the md files so they display glossary tooltips on hover and dotted underline links to click on (for mobile).

- [x] Implement mobile view ~~per https://github.com/iterative/dvc.org/pull/431#pullrequestreview-263985560~~ @algomaster99 

- [ ] Add missing glossary terms listed on #424.
  
- [ ] We have to use discretion to chose at least one occurrence of each term in all the docs to have this functionality. Note that some of these occurrences are currently linked to some other doc and would need to be unlinked (but an equivalent link should be in the glossary tooltip).  
  Note that the "workspace" and "DVC project" terms are already fully applied as well as "cache directory" and "the project" (but maybe not simply "cache", so start from there).

  > Per https://github.com/iterative/dvc.org/pull/431#issuecomment-506012482 and https://github.com/iterative/dvc.org/pull/431#issuecomment-506078052

- ~~There should be a section or a few sections called Basic Concepts in the user guide that is/are properly written on a user-level language explaining some missing concepts. The glossary tooltips could point to these sections among other things. An example would a DVC-file term and this link: https://dvc.org/doc/user-guide/dvc-file-format~~ Extracted to #550 

- [ ] Review the terms and their descriptions in the glossary.js file, vs. all the existing definitions in md docs, possibly making the ones in docs shorter now that we can have the glossary tooltip over those occurrences of each term.

  > Per https://github.com/iterative/dvc.org/pull/431#discussion_r298302607

---

~~Possibly also add functionality to the glossary engine module so all the `abbr` terms in each doc are displayed on the right side bar (when visible, for wider screens) beneath the edit and chat buttons.~~

algomaster99:
@jorgeorpinel Let's start working on this issue now. :)

jorgeorpinel:
Feel free to get going with fixing the mobile view, sure! Thanks @algomaster99

jorgeorpinel:
Any luck so far with the mobile view @algomaster99 ?

Of all the `glossary:` issues, this I think is the most important one so I'm giving it p1 label @shcheklein but please correct me if i'm wrong.

jorgeorpinel:
[...]

Issue #424, opened 2019-06-07
create glossary


jorgeorpinel:
From https://github.com/iterative/dvc.org/issues/321#issuecomment-491665471

- [x] Create new **Glossary** ~~doc~~ with definitions to relevant DVC concepts and terms. Use appropriate links to other docs in these definitions.
- [ ] Mention to "utilize the terminology listed in the glossary(linked) consistently" in the docs contrib guide.

---
UPDATE: Main terms that can be grouped in single entries, and related terms to connect with or disambiguate from (ideas):
- Workspace, DVC project - related: SCM repository, working directory
- DVC meta files and directories (in '.dvc/' dir) - related: DVC config file, DVC cache
- Data artifact, data - related: versioning
- DVC cache, cache, "the cache" - related: file checksum, file link, innode, remote cache
- DVC-file - related: stage file, `.dvc` file extension, `Dvcfile` file name, _orphan_ stage file
- DVC remote, remote cache - related: remote location (see external deps/outs guides), cache
- Pipeline, data pipeline - related: command chain, pipeline stage
- Stage, pipeline stage - related: stage file, DVC-file
- DAG, computational graph, command chain - related: pipeline
- Metric?
- DVC workflow? - related: workflow, Git workflow, experiment workflow

(but in alphabetical order)

kurianbenoy:
I would like to work on this after completing #396

shcheklein:
@kurianbenoy sure! but let's coordinate here first with @jorgeorpinel, what terms we should put into Glossary, their definitions, etc.

algomaster99:
To start the list, here are a few suggestions:
- Pipeline/computation graph
- Cache and workspace
- Metric
- Dvcfile

ryokugyu:
@kurianbenoy let's work together on it. I am also working on a tutorial which is about to complete.

ryokugyu:
> To start the list, here are a few suggestions:
> 
> * Pipeline/computation graph
> * Cache and workspace
> * Metric
> * Dvcfile

Further, 
* repro
* staging
* remote
* lock 
* workflow

shcheklein:
Related #296 . I would say glossary should be very simple and it's main purpose probably to serve as an index and as a guide for contributors. While #296 should introduce DVC concepts like cache, workspace, etc in more details.

algomaster99:
@shcheklein  Who is working on this? I can take this up if it is unassigned. :)

shcheklein:
[...]

PR #611, opened 2019-09-09
term: "metric" and related; other regular updates

jorgeorpinel:


static/docs/commands-reference/metrics/add.md
```diff
@@ -13,14 +13,20 @@ positional arguments:
 
 ## Description
 
-Sets a special flag in the relevant [DVC-file](/doc/user-guide/dvc-file-format)
-to identify a specified output as a metric file. Alternatively, an output file
-can be marked as a metric via `-M` or `-m` parameter of the `dvc run` command.
+Sets a special field, in the [DVC-file](/doc/user-guide/dvc-file-format) that
```
shcheklein: probably should remove `special` here as well

static/docs/user-guide/contributing.md
```diff
@@ -91,8 +91,8 @@ $ python -m tests
 ============= 434 passed, 6 skipped, 8 warnings in 131.43 seconds ==============
 ```
 
-Otherwise, for each failed test you should see and output like this that should
-help you identify the problem:
+Otherwise, for each failed test you should see and <abbr>output</abbr> like this
```
shcheklein: not a right `abbr` here I think

static/docs/get-started/compare-experiments.md
```diff
@@ -2,9 +2,10 @@
 
 DVC makes it easy to iterate on your project using Git commits with tags or Git
 branches. It provides a way to try different ideas, keep track of them, switch
-back and forth. To find the best performing experiment or track the progress, a
-special _metric_ output type is supported in DVC (described in one of the
-previous chapters).
+back and forth. To find the best performing experiment or track the progress,
+special [project metrics](/doc/commands-reference/metrics) (a type of
```
shcheklein: remove `special`?

static/docs/get-started/example-pipeline.md
```diff
@@ -224,7 +225,8 @@ $ dvc run -d code/train_model.py -d data/matrix-train.pkl \
           python code/train_model.py data/matrix-train.pkl 20170426 data/model.pkl
 ```
 
-Finally, evaluate the model on the test dataset and get the metrics file:
+Finally, evaluate the model on the test dataset and get the
+[metric](/doc/commands-reference/metrics) file:
```
shcheklein: I might make sense to do `abbr` for metrics term instead?

static/docs/commands-reference/get.md
```diff
@@ -1,7 +1,7 @@
 # get
 
-Download or copy file or directory from another DVC repository (on a Git server
-e.g. Github) into the local file system.
+Download or copy file or directory from an external <abbr>DVC project</abbr>
+hosted on a Git repository online (e.g. Github) into the local file system.
```
jorgeorpinel: @shcheklein to your comment in https://github.com/iterative/dvc.org/commit/1fd6c115210ae422b71c9eba52b94e4d761ced91#r35008637,

> This not fully correct... I believe we can use `dvc get` on "offline" projects, not online only.
Most likely we can use `dvc get` with `.` (internal project) - a sample use case is to get an artifact from a specific revision to then diff it.

Indeed. I just confirmed all this. Added info explaining that `url` can be a local file system path and removed the term "external".

> `hosted on Git` sounds not common, better: `from Git`, `specified by Git URL`, etc.

Improved all, see my latest changes to the file (: 

BUT, do we want `dvc get` to work with `.`? Seems confusing and unnecessary to me. That's what `dvc checkout` is for... If so maybe I should add a note that you can even use this same current repo as a source and an example on how to replace checkout with get.

static/docs/commands-reference/get.md
```diff
@@ -18,19 +18,26 @@ positional arguments:
...
 
+The `url` argument specifies the address of the Git repository containing the
+source <abbr>DVC project</abbr>. Both HTTP and SSH protocols are supported for
+online repositories (e.g. `[user@]server:project.git`). `url` can also be a
+local file system path to an "offline" repository. The `path` argument of this
+command is used to specify the path of the data to be downloaded within the
+repo.
+
+Note that the required `path` is expected to point to a file or directory
+defined in one of the [DVC-files](/doc/user-guide/dvc-file-format) in the source
```
jorgeorpinel: @shcheklein to your comment in https://github.com/iterative/dvc.org/commit/1fd6c115210ae422b71c9eba52b94e4d761ced91#r35008668,

> defined -> controlled? referenced? I don't like that it sounds here like DVC-files come first and data second, while we always should be emphasizing that DVC-files are helpers/meta for data.

Hmmm OK updated here but we use this language in a bunch of other places, let me review...

static/docs/commands-reference/get.md
shcheklein: > BUT, do we want dvc get to work with .? Seems confusing and unnecessary to me. That's what dvc checkout is for... If so maybe I should add a note that you can even use this same current repo as a source and an example on how to replace checkout with get. 

The difference is that you can get two (or more) versions of the same file/directory simultaneously with `dvc get. It can be useful to dig deeper, compare with some diff tool, reuse for whatever reason. On the other hand, with `dvc checkout` you can _checkout_ only one version at a time, also it's quite cumbersome and requires good knowledge of Git.

static/docs/commands-reference/get.md
jorgeorpinel: OK, reviewed throughout in my latest commit in this PR.

shcheklein: I'm not sure we need to make a strict rule or fix other places. It's just in this case when we try to emphasize and explain that it expects data reference I would be careful with wording and usage of DVC-file stuff.

static/docs/commands-reference/metrics/add.md
jorgeorpinel: Removed and reviewed throughly in my last commit.

static/docs/user-guide/contributing.md
jorgeorpinel: Fixed along with related changes in my latest commit.

static/docs/get-started/compare-experiments.md
jorgeorpinel: Also fixed in the latest commit.

static/docs/get-started/example-pipeline.md
jorgeorpinel: Agreed, already requested in #461 (listed in #424).

static/docs/commands-reference/get.md
jorgeorpinel: Hmm well it's not obvious to me when wording changes should or not propagate throughout all docs. (I like to do this generally, for consistency.) Do you think the default should be to avoid that? It's typically a quick "grep", but some times it can become pretty time consuming to check everything.

static/docs/commands-reference/get.md
jorgeorpinel: > you can get two (or more) versions of the same file/directory simultaneously with `dvc get`.

Good point... (Using the `--out` option.) OK added an advance example about this to `dvc get` for PR #585. See https://github.com/iterative/dvc.org/pull/585/commits/8e9e94725675e5285d4d36f0a4fd8ae77c45e95b

shcheklein:
@jorgeorpinel is it ready to merge? I see a few comments are not still addressed, right?

static/docs/commands-reference/get.md
```diff
@@ -18,19 +18,26 @@ positional arguments:
...
 single-purpose command that can be used out of the box after installing DVC.
 
+The `url` argument specifies the address of the Git repository containing the
+external <abbr>project</abbr>. Both HTTP and SSH protocols are supported for
+online repositories (e.g. `[user@]server:project.git`). `url` can also be a
+local file system path to an "offline" repository.
+
+The `path` argument of this command is used to specify the location of the data
+to be downloaded within the source project. It should point to a data file or
+directory tracked by that project – specified in one of the
```
jorgeorpinel: > Still a bit too complicated and hard to read tbh... not sure we need stuff about visibility...

@shcheklein This I had previously revised myself but just did some more improvements in 39a19f9.

> On the other hand, no matter how well explained, the commands get and import are intrinsically a bit confusing...

At least the arguments are somewhat hard to explain for sure, even if they do make sense but as mentioned before there should probably be a command to list data artifacts in an external project. I don't remember if there's an issue or discussion about it open still...

jorgeorpinel: > May it's better to show a simple code snippet...
> ```
> $ dvc get example-get-started model.pkl
> ```

I agree we shouldn't shy away from short command examples in descriptions but that is already the exact first example so the user only has to scroll a little bit to find it. I would use a different command but actually I don't think this command helps explain the fact that `path` is not actually a file in the Git repo, but a record in one of its DVC-files.

static/docs/commands-reference/get.md
```diff
@@ -18,19 +18,26 @@ positional arguments:
...
-DVC provides an easy way to reuse datasets, intermediate results, ML models, or
-other files and directories tracked in another DVC repository into the current
-working directory, regardless of whether it's a DVC project. The `dvc get`
-command downloads such a <abbr>data artifact</abbr>.
-
-The `url` argument specifies the address of the Git repository containing the
-external <abbr>DVC project</abbr> (both HTTP and SSH protocols supported, e.g.
-`[user@]server:project.git`). `path` is used to specify the path of the data to
-be downloaded within the repo.
+Provides an easy way to reuse datasets, intermediate results, ML models, or
```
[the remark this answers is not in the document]
jorgeorpinel: Changed.

static/docs/user-guide/dvcignore.md
```diff
@@ -36,9 +36,9 @@ Keep in mind, that when you add to `.dvcignore` entries that affect one of the
 existing <abbr>outputs</abbr>, its status will change and DVC will behave as if
 that affected files were deleted.
 
-If DVC finds a `.dvcignore` file inside a dependency or an <abbr>output</abbr>
-directory, it raises an error. Ignoring files inside such directory should be
-handled from `.dvcignore` file from upper levels of the project tree.
+If DVC finds a `.dvcignore` file inside a dependency or in an output directory,
```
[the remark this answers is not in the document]
jorgeorpinel: Yes but it's used in the paragraph above. I only use `<abbr>` about once per section max.

static/docs/tutorial/define-ml-pipeline.md
```diff
@@ -380,8 +380,8 @@ $ git add .
 $ git commit -m Evaluate
 ```
 
-The <abbr>output</abbr> of the evaluation stage contains the target metrics
-value in a simple text form:
+The <abbr>output</abbr> of the evaluation stage contains the target value in a
```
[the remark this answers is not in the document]
jorgeorpinel: Agree. But added in an occurrence a few paragraphs above in the same section instead. (See ab95e5f.)

jorgeorpinel:
Sorry for the delay. Yes, some conversations are still open, please review. Hopefully it's ready or near ready for merge.

static/docs/commands-reference/get.md
jorgeorpinel: > there should probably be a command to list data artifacts in an external project

Opened feature request in https://github.com/iterative/dvc/issues/2509.

shcheklein:
@jorgeorpinel I think it's more or less ready to be merged - please resolve the conflicts and let me know when it's ready :)

jorgeorpinel:
Resolved (:

PR #629, opened 2019-09-17
install: update description about behaviour on existing hook

pared:
Merge after https://github.com/iterative/dvc/pull/2505


static/docs/commands-reference/install.md
```diff
@@ -51,6 +51,8 @@ The installed Git hook automates executing `dvc push`.
 - A `pre-push` hook executes `dvc push` before `git push` to upload files and
   directories under DVC control to remote storage.
 
+_If a hook already exists, DVC will raise exception._
```
shcheklein: _an_ exception - I think 

Clarify a bit what should I do if I see this exception. Edit manually? Can we put a link to Git docs at least so that people can start from there?

PR #630, opened 2019-09-19
Fix: typo dvc run -R to -M

artmadeit:
In get started > metrics.
Says:
using the `-R` option of `dvc run`.
when there isn't any -R option. So maybe ir refers to the -M option.

- Have you followed the guidelines in our
  [Contributing Documentation](https://dvc.org/doc/user-guide/contributing-documentation)?
Yes, it's a small one


shcheklein:
thanks @artmadeit !

PR #631, opened 2019-09-19
get-started: summarize Import Data chapter

jorgeorpinel:
and related updates


src/Documentation/sidebar.json
```diff
@@ -8,19 +8,19 @@
       "initialize",
       "configure",
       "add-files",
-      "share-data",
+      "store-data",
```
shcheklein: please, grep and update the links to this chapter if we have any

src/Documentation/sidebar.json
```diff
@@ -8,19 +8,19 @@
...
       "retrieve-data",
+      {
+        "label": "Import Data",
```
shcheklein: you don't need to specify a label if it's the same as title in .md file

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,69 @@
...
+[stored](/doc/get-started/store-data) for a single <abbr>DVC project</abbr>.
+`dvc import` also lets us download <abbr>data artifacts</abbr>, but from a
+different project that is hosted on a Git repository, and creates an import
+stage (DVC-file) to track changes in this
+[external dependency](/doc/user-guide/external-dependencies).
+
+For example, in the [Add Files](/doc/get-started/add-files) chapter we download
+raw data using the `dvc get` command – which unlike `dvc import` does not create
+an import stage. Let's say we wanted to download the same dataset, while keeping
+the connection to its source project:
```
shcheklein: To be honest I looks like both paragraphs are saying more or less the same. Can we try to merge them?

Also a few comments:

`download data for a single DVC project` - not clear enough is it for that project or is it data defined/tracked in that project? 

first paragraph is complicated and confusing still :( took me a while to understand that it's about push/pull. Can we just write something like this:

`We've see how to use push and pull to store and download back data for a project. But what if we want to plug some data or ML model produced in one project into another one? To enable reusability of data artifacts.

One option, of course is to download it (wget or `dvc get`) and use `dvc add`. But this way the connection is lost, we don't know where that data came from, how it was produced, are there new versions of it, etc. `dvc import` similar to what `dvc get` (or wget/curl) + `dvc add` can do but it also saves some additional context into the DVC-file it creates:
``` 

`import stage` - let's try to avoid this terminology here - just a DVC-file that has some addition information inside

File: src/Documentation/sidebar.json (hunk 1 above)
Comment (jorgeorpinel) [reply src/Documentation/sidebar.json]: I did look, no links to update (:

File: src/Documentation/sidebar.json (hunk 2 above)
Comment (jorgeorpinel) [reply src/Documentation/sidebar.json]: Ah true. This used to be different than it's title, that's why it was expanded here. Fixed.

File: static/docs/get-started/import-data.md (hunk 3 above)
Comment (jorgeorpinel) [reply static/docs/get-started/import-data.md]: OK. I did an almost complete rewrite based in 6126652 on these comments. What do you think?

File: static/docs/get-started/import-data.md [hunk 4]
```diff
@@ -0,0 +1,70 @@
...
+`dvc add` to track it locally, but the connection between projects would be
+lost: we couldn't tell where the data came from or whether there are new
+versions available. We offer the `dvc import` command instead (for projects
+hosted on a Git repository). It downloads the data, and creates a
+[DVC-file](/doc/user-guide/dvc-file-format) to track changes in the imported
+data.
+
+<details>
+
+### Expand to for an example
```
shcheklein: `to for` typo? to lear more?

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
+# Import Data
+
+We've seen how to [push](/doc/get-started/store-data) and
+[pull](/doc/get-started/retrieve-data) data a <abbr>DVC project</abbr>'s
+[remote](/doc/commands-reference/remote). But what if we wanted to integrate a
+dataset or ML model produced in one project into another project? (As an
+[external dependency](/doc/user-guide/external-dependencies))
```
shcheklein: `))` ->`).)` (end of sentence)

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
...
+
+One way is to download the data (with `wget` or `dvc get`, for example) and use
+`dvc add` to track it locally, but the connection between projects would be
```
shcheklein: locally -> within the project?

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
...
+lost: we couldn't tell where the data came from or whether there are new
+versions available. We offer the `dvc import` command instead (for projects
```
shcheklein: `We offer` something less salesy? :)

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
...
+hosted on a Git repository). It downloads the data, and creates a
```
shcheklein: I think we can omit that it's hosted on Git here

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
...
+[DVC-file](/doc/user-guide/dvc-file-format) to track changes in the imported
+data.
+
+<details>
```
shcheklein: I think it went to far a bit :) Let's show the command at least with some, may be with this intro: `In the Add Files chapter we download raw data using the dvc get command, and then track it with dvc add. Let's see how this would be improved by, instead, just running:` you have

static/docs/get-started/import-data.md
jorgeorpinel: Ugh yes. **~~to~~ for**. Fixing, thanks.

static/docs/get-started/import-data.md
jorgeorpinel: Hmmm it's strange indeed but wasn't a full sentence so no period `.`. I'm moving the question mark `?` to the end instead (: Thanks

static/docs/get-started/import-data.md
jorgeorpinel: Which project? There are 2 in question. Changing to just "track it".

static/docs/get-started/import-data.md
jorgeorpinel: True. Rephrasing!

jorgeorpinel: p.s. Note that due to semantics I'm splitting this paragraph into 2 smaller ones. I know you only wanted 1 or 2 paragraphs but it just wouldn't be good writing. But it's the same number of words after all (:

static/docs/get-started/import-data.md
jorgeorpinel: I guess "hosted" is not the clearest term, changed to "for projects in a Git repository" to match the (new) [`dvc import` command reference](https://dvc-org-pr-631.herokuapp.com/doc/commands-reference/import). But I think it's an important note to keep, it only takes 6 words.

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,72 @@
+# Import Data
+
+We've seen how to [push](/doc/get-started/store-data) and
+[pull](/doc/get-started/retrieve-data) data a <abbr>DVC project</abbr>'s
```
shcheklein: `in` or `within` is missing?

static/docs/get-started/import-data.md
jorgeorpinel: p.s. Actually I figured out a way to keep all this within 3 paragraphs:  by moving the last paragraph after the example up to the last one introducing `dvc import` (split from the one above in this review comment).

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,72 @@
...
+[remote](/doc/commands-reference/remote). But what if we wanted to integrate a
+dataset or ML model produced in one project into another project (as an
+[external dependency](/doc/user-guide/external-dependencies))?
+
+One way is to download the data (with `wget` or `dvc get`, for example) and use
+`dvc add` to track it, but the connection between projects would be lost. We
+wouldn't be able to tell where the data came from or whether there are new
+versions available.
+
+There's a better alternative (for projects in a Git repository): the
```
shcheklein: a few problems here - `for projects` - does it mean that import does not work for other projects? or that the project we are importing from should be a git project. 

I would imagine 99% of the DVC projects are Git projects - that's why I'm not sure we need to mention this note here. May be it's better to mention something like `... by providing a Git URL to a DVC project`?

static/docs/get-started/import-data.md
```diff
@@ -0,0 +1,70 @@
...
+`dvc import` command. It downloads the data, and creates a
+[DVC-file](/doc/user-guide/dvc-file-format) to track changes in the imported
+data. This is better than using `dvc add` to track downloaded data, because with
+_imports_ we can run `dvc update` to check for changes in the external data
+source before [reproducing](/doc/get-started/reproduce) the
+<abbr>pipeline</abbr> that depends on this data.
```
shcheklein: this last part can go after the code snippet?

static/docs/get-started/import-data.md
jorgeorpinel: ✔️

static/docs/get-started/import-data.md
jorgeorpinel: Good point. How about "It downloads data from any DVC project in a Git repository...'? (Again, this wording is copied from the `dvc import` cmd ref.) Changed in fb4efc8.

static/docs/get-started/import-data.md
jorgeorpinel: Not sure I understand, please be specific (:

jorgeorpinel: Also, I still haven't rewritten the whole expandable section (per https://github.com/iterative/dvc.org/pull/631#pullrequestreview-291330688) but working on it.

static/docs/get-started/import-data.md
jorgeorpinel: Show the command with some ?
I have ?

jorgeorpinel: I think I may know what you mean. It's too detailed. I slashed it by half now. Better?

static/docs/get-started/import-data.md
jorgeorpinel: I simplified this paragraph in my latest commit.

static/docs/get-started/import-data.md
jorgeorpinel: Pushed new version we worked on together.

static/docs/get-started/example-pipeline.md
```diff
@@ -41,6 +41,12 @@ $ git add code/
 $ git commit -m "Download and add code to new Git repo"
 ```
 
+> `dvc get` can download <abbr>data artifacts</abbr> from any <abbr>DVC
```
shcheklein: remove one extra new line before

jorgeorpinel:
Woo! 🎉

PR #565, opened 2019-08-16
Fix #563:  Managing Data Storage On An External Hard Drive

dashohoxha:
Fix #563

PR #455, opened 2019-06-25
Shared cache on NFS Introduced

ryokugyu:
fix #103

PR #637, opened 2019-09-24
use-cases: review of existing material; et al.

jorgeorpinel:


jorgeorpinel:
@shcheklein is the index for this section necessary? Perhaps we could remove it like https://dvc.org/doc/user-guide which just jumps to the first guide.

shcheklein:
@jorgeorpinel I would say it can be useful as an extended list of content. May be we should simplify it to be precisely a list of use cases with some brief descriptions (ideally just excerpts from the cases). What do you think?

jorgeorpinel:
Yes  I think it's useful. I think the way it is now (in this PR) works well but as I add more use cases I'll start adding the descriptions here are you suggest.

Really the question is why doesn't the User Guide have an index 😋 - but that's besides the point here.

jorgeorpinel:
> I would say it can be useful as an extended list of content. May be we should simplify it to be precisely a list of use cases with some brief descriptions

Applied this now actually, @shcheklein. See https://dvc-org-pr-637.herokuapp.com/doc/use-cases

src/Documentation/sidebar.json
```diff
@@ -43,6 +43,10 @@
         "label": "Data & Model Files Versioning",
         "slug": "data-and-model-files-versioning"
       },
+      {
+        "label": "Update a Tracked File",
+        "slug": "update-tracked-file"
```
shcheklein: do you need a slug here?

jorgeorpinel: Will be changed to `update-tracked-files` in an upcoming commit anyway.

jorgeorpinel: Updated.

src/Documentation/sidebar.json
```diff
@@ -47,15 +47,13 @@
         "label": "Share Data & Model Files",
         "slug": "share-data-and-model-files"
       },
-      {
-        "label": "Shared Development Machine",
-        "slug": "multiple-data-scientists-on-a-single-machine"
-      }
+      "update-tracked-files",
```
jorgeorpinel: Not sure whether [Update Tracked Files](https://dvc-org-pr-637.herokuapp.com/doc/use-cases/update-tracked-files) as is belongs to Use Cases or User Guide. I think it's more of a use case. @shcheklein thinks it's more of a User Guide. Informal definitions:

> Use Cases: (the most blog post-like like content, lots of text about motivation, titles that match real-life workflow/high level cases, has links to 1+ user guide sections and 1+ links to commands reference)

> User Guide: like a car owner manual: not a goal to explain even why do you need a car, but explain different parts, how to use them. Of course it can include some motivation (like why would you want to manage external data?). But in general it’s more technical, dry.

jorgeorpinel: p.s. note that I've updated the index.md of each doc section to include a formal description of what "use cases" and "user guides" are. (See [UC](https://dvc-org-pr-637.herokuapp.com/doc/use-cases) and [UG](https://dvc-org-pr-637.herokuapp.com/doc/user-guide), respectively.)

shcheklein: just to clarify, do we need to clarify further the difference between UC/UG? is it clear enough now why "updating a tracked file" is more about UG and not UC? if not, could you please give a bit more context why do you consider it as a UC?

src/Documentation/glossary.js
```diff
@@ -13,8 +13,8 @@ code, ML models, etc. A workspace becomes a **DVC project** when
 [\`dvc init\`](/doc/commands-reference/init) is run, and
 [DVC-files](/doc/user-guide/dvc-file-format) or stage files are created in it.
 
-Note that [external outputs](/doc/user-guide/external-outputs) also form part
-of your expanded workspace, technically.
+Note that [external outputs](/doc/user-guide/external-cache-and-outputs) also
```
shcheklein: I wonder if we can create a script that checks every link as a test on CI system? :)

src/Documentation/sidebar.json
```diff
@@ -65,22 +63,19 @@
         "label": "File Format (.dvc)",
         "slug": "dvc-file-format"
       },
+      "dvcignore",
       "large-dataset-optimization",
       "external-dependencies",
       {
         "label": "Managing External Data",
-        "slug": "external-outputs"
+        "slug": "external-cache-and-outputs"
```
shcheklein: I would actually not change it now. Let's wait when we have more clarify on how do we want to structure this. Otherwise, why not `managing-external-data`?

static/docs/commands-reference/gc.md
```diff
@@ -91,3 +99,5 @@ Let's check the size now:
 $ du -sh .dvc/cache/
 3.1G    .dvc/cache/
 ```
+
+<!-- TODO: Add `dvc gc -aT` example. -->
```
shcheklein: let's not commit TODOs please :) create a ticket/checkbox for that

static/docs/use-cases/index.md
```diff
@@ -1,15 +1,28 @@
...
-This list is _not_ exhaustive. We keep growing it and try to include interesting
-scenarios from the community. Please, [contact us](/support) if you have some
-suggestions!
-
-If you want to get some hands-on experience we recommend to follow along
-[Get Started](/doc/get-started) and try examples at the very end of it. They
-include prepared datasets and code to run. Use cases are not written to be run
-end to end, specifically we don't provide prepared datasets and code to train
-models.
+We provide short articles on common DVC scenarios in this section, with a
```
shcheklein: better to say something like: "ML workflow or data management scenarios that DVC can help with or improve"

static/docs/use-cases/index.md
```diff
@@ -1,15 +1,28 @@
...
+justified context (usually extracted from real-life support cases). Their
```
shcheklein: nor sure we should mention that it comes from support cases. It comes from different sources, not only support cases.

static/docs/use-cases/index.md
```diff
@@ -1,15 +1,28 @@
...
+solution approaches combine several features and commands of DVC, for didactic
```
shcheklein: it should not be necessarily even DVC commands only - in case of "Training models remotely" it can be DVC + another tool that helps orchestrating cloud machines

static/docs/use-cases/index.md
```diff
@@ -1,15 +1,28 @@
...
+- [Shared Development Server](/doc/use-cases/shared-development-server.md)
+  describes a single development machine setup for teams that prefer so.
+
+This list of use cases is _not_ exhaustive. We keep reviewing our docs and will
+include interesting scenarios that surface in our community. Please,
+[contact us](/support) if you need help or have suggestions!
+
+Use cases are not written to be run end-to-end. For more general hands-on
+experience with DVC, we recommend to follow the [Get Started](/doc/get-started),
+and/or [Tutorial](/doc/tutorial)] first – These include prepared datasets,
```
shcheklein: These -> these?

static/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -4,7 +4,7 @@ Once initialized in a <abbr>project</abbr>, DVC populates its installation
 directory (`.dvc/`) with the internal files and directories needed for DVC
 operation:
 
-### DVC internal files and directories
+### DVC internals
```
shcheklein: probably not a good naming here ... I would read this as internals of the DVC as a program, not files ... to be honest "internal" is a pretty bad term here in general ... `.dvc/config` is not internal, it provides a public API. Let's take a look if Git has a similar section or how do people refer to `.git` directory content.

static/docs/user-guide/dvcignore.md
```diff
@@ -1,15 +1,15 @@
 # .dvcignore File
 
-Marks which files and/or directories should be ignored when traversing
-repository.
+Marks which files and/or directories should be excluded when traversing a
```
shcheklein: why not ignored? feels like it can be a better term since it's _dvcignore_ :)

static/docs/user-guide/dvcignore.md
```diff
@@ -1,15 +1,15 @@
...
-many files in its data directory, you might encounter extended execution time
-for operations that are as simple as `dvc status`. In other cases you might want
-to omit files or folders unrelated to the project (like `.DS_Store` on Mac). To
-address these requirements we are implementing `.dvcignore` files handling.
-`.dvcignore` by design works similar way as `.gitignore` does.
+project. For example, when working in a <abbr>workspace</abbr> directory with a
+large number of data files, you might encounter extended execution time for
+operations as simple as `dvc status`. In other case you might want to omit files
+or folders unrelated to the project (like `.DS_Store` on Mac). To address these
+scenarios, DVC supports optional `.dvcignore` files. By default, `.dvcignore`
```
shcheklein: By default means usually that there is a second option or some way to make it behave differently. Not the case now, probably.

static/docs/user-guide/external-cache-and-outputs.md
```diff
@@ -14,7 +14,7 @@ You can take under DVC control files on an external storage with `dvc add` or
 specify external files as outputs for
 [DVC-files](/doc/user-guide/dvc-file-format) created by `dvc run` (stage files)
 DVC will track changes in those files and will reflect so in your pipeline
-[status](/doc/commands-reference/status). Currently, the following types of
+[status](/doc/command-reference/status). Currently, the following types of
```
shcheklein: are we renaming it? it would be painful and I'm not sure it's needed

static/docs/user-guide/index.md
```diff
@@ -0,0 +1,33 @@
+# User Guide
+
+This section describes the main DVC concepts and features comprehensively,
+explaining when and how to use them, as well as connections between them. These
+guides don't focus on specific scenarios, but have a general scope – like a user
+manual. Their topics range from more technical basics, which impact more parts
+of DVC, to more advanced things you can do. We also include a few guides related
+to contributing to this [open-source project](https://github.com/iterative/dvc).
+
+- [DVC Files and Directories](/doc/user-guide/dvc-files-and-directories)
```
shcheklein: I hope that we'll have a proper third level soon to cluster some of these by topics. Not for this PR for sure.

src/Documentation/glossary.js
dashohoxha: > I wonder if we can create a script that checks every link as a test on CI system? :)

It is easy to create a script that extracts all the URL-s from a file. Also it is easy to check each of them if it exists. I am not sure about how to use this on a CI system.

Can we create an issue for implementing this?

There is also this: https://wummel.github.io/linkchecker/ But it seems to be a bit old and bloated.

jorgeorpinel: Yeah good idea. Who implemented the initial CI pipeline and yarn tests? Let's ask them for now (: Opened https://github.com/iterative/dvc.org/issues/652, please tag them there.

static/docs/commands-reference/gc.md
jorgeorpinel: Accidental commit. Removing...

static/docs/use-cases/index.md
jorgeorpinel: Updating

static/docs/use-cases/index.md
jorgeorpinel: Removing "support".

static/docs/use-cases/index.md
jorgeorpinel: Adding ", as well as other appropriate tools".

static/docs/use-cases/index.md
jorgeorpinel: Changing to "which" (lower case, yes), actually. Thanks

jorgeorpinel: Updates to this file in 2e26095.

static/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: `.git/`:
- "contains all of your necessary repository files — a Git repository skeleton" ([Git Basics](https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository) official docs)
- "where almost everything that Git stores and manipulates is located" ([Git **Internals**](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain#ch10-git-internals) official docs)
- "Git directory structure... special folder where all the git stuff is" ([Inside Git: .Git directory](https://githowto.com/git_internals_git_directory))
- "subfolder with all the git related revision history of your repo" ([What is a bare git repository?](http://www.saintsjd.com/2011/01/what-is-a-bare-git-repository/))

`.git/config`:
- "config(uration) file" 
- "contains your project-specific configuration options" ([Git **Internals**](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain#ch10-git-internals) official docs)

The **Git Internals** chapter linked above desribes itself as "... this is where we’ll go over the inner workings and implementation of Git."

static/docs/user-guide/dvcignore.md
jorgeorpinel: The term "ignore" is everywhere on that document. I added a couple synonyms like "exclude" and "omit" for user search purposes. The former ("exclude") is also used in other docs in reference to `.dvcignore` usage, but mostly for other purposes. "Omit" is not used anywhere else.

If you'd like I can revise all the occurrences of these terms to make sure only "ignore" is used for Dvcignore stuff, and that "exclude" is also used consistently (in some other context). Maybe incorporate "omit" if needed. Lmk

jorgeorpinel: p.s. It's also good writing practice to have varied lexicon.

static/docs/user-guide/dvcignore.md
jorgeorpinel: True! Removing "by default".

static/docs/user-guide/external-cache-and-outputs.md
jorgeorpinel: Fixed back to `commands-reference` (in a few places).

But actually I do want to change it because the most used term is "Command Reference". Examples:

image
image
image

image

image

jorgeorpinel: p.s. in fact I already have the change stashed on my branch but indeed it's pretty huge so I decided not to include it in this PR. Would need a separate one. Waiting for this one to merge first. Do you agree @shcheklein?

src/Documentation/sidebar.json
jorgeorpinel: re User Guide vs. Use Cases:

I see why you don't want **Update Tracked Files** in Use Cases. It's not specific enough real-file example. But TBH the other Use Cases are also general in some ways, especially [Share Data and Model Files](https://dvc.org/doc/use-cases/share-data-and-model-files), which example S3 could be replaced by any other supported remote type. The existing use cases just happen to include a real-life example and some diagrams. Yet, for consistency (since UTF has no example or diagrams) I agree with you... So I'll move it back for now.

BUT I also don't think it really belongs in the User Guide, which "describes the main DVC concepts and features comprehensively... These guides don't focus on specific scenarios, but have a general scope".

Anyway. I'm opening a new issue to solve this confusion once and for all! See #654

src/Documentation/sidebar.json
jorgeorpinel: I don't really like **Managing External Data** actually. Why isn't **External Dependencies** included in that guide? I think **External Cache and Outputs**, which is what this article is about, should be not only the slug but the title (except that it's too long for the menu/nav bar). Maybe just **External Outputs** in the menu?

shcheklein: it was before. Yes, it's confusing for now. The idea is to eventually (and I was waiting for the engine to properly support 3 levels) - is to have one single Manage External Data second level section, that will include External Outputs, External Dependencies, External Cache, Shared Cache (or a better title to explain cases like, https://github.com/iterative/dvc.org/pull/565 by @dashohoxha , https://github.com/iterative/dvc.org/pull/455), other related articles.

We'll have to go three levels to put more structure and scale growing UG.

shcheklein: For now I've changed it to Manage External Data to simplify a new user experience - External Outputs do not mean anything to them, especially the name is extremely confusing when you use External _output_ to deal with _input_ (source, raw) data - this a very common case.

src/Documentation/sidebar.json
shcheklein: Agreed regrading S3, it was used for illustration only and probably this document should mention that there are other remotes.

Also, agreed that sometimes it's hard to make it precisely "Use Casey".

Regarding the User Guide - there are sections like "Large Files Optimization" - it's not a use case by itself, it's a technical description how to fine fine tune DVC in certain situations. I think UG should include things like `Pipelines -> How to Edit, How to Create`, etc. And Updating Tracked Files is very similar in my mind to those.

static/docs/user-guide/external-cache-and-outputs.md
shcheklein: Yep, let's change it if you see that it's more common. Please, do `git grep commands-reference` to double check that everything is handled.

shcheklein: And, yes, it's better to do a separate PR 👍

static/docs/user-guide/dvcignore.md
shcheklein: 👍

static/docs/user-guide/dvc-files-and-directories.md
shcheklein: so, this should be specifically `Inside DVC: .dvc directory` then? Not more general and broad "DVC internals"?

static/docs/user-guide/index.md
jorgeorpinel: Good idea! Moved your remarks to https://github.com/iterative/dvc.org/issues/654#issuecomment-536330090

src/Documentation/sidebar.json
jorgeorpinel: Understood. I renamed it `managing-external-data` for now then (in 3d486bc).

I moved your other remarks to https://github.com/iterative/dvc.org/issues/654#issuecomment-536330090.

src/Documentation/sidebar.json
jorgeorpinel: > Regarding the User Guide...

Good points. Moved discussion to https://github.com/iterative/dvc.org/issues/654#issuecomment-536330090 🙂

static/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Maybe I'll just remove this H2! It's not really needed is it? But I also need to review the usage of the term "internal" throughout the docs...

src/Documentation/sidebar.json
jorgeorpinel: > ...and probably this document should mention that there are other remotes.

 Added note in e29233b

static/docs/user-guide/dvc-files-and-directories.md
shcheklein: Sounds good to me!

static/docs/user-guide/index.md
shcheklein: 👍

src/Documentation/sidebar.json
shcheklein: 👍

static/docs/commands-reference/gc.md
```diff
@@ -11,33 +11,41 @@ usage: dvc gc [-h] [-q | -v] [-a] [-T] [-c] [-r REMOTE] [-f] [-j JOBS]
...
-specified. If the cache holds a few cached versions of a data file, all except
-the current one will be removed. You can `dvc fetch` all the needed files back
-anytime you want **as long as you have previously pushed them to a remote**.
-(See `dvc remote` and `dvc push` for more information.)
+This command deletes (garbage collects) data files or directories that may exist
+in the cache (or [remote storage](/doc/commands-reference/remote)) but no longer
+referred to in [DVC-files](/doc/user-guide/dvc-file-format) currently
+[checked out](/doc/commands-reference/checkout) in the <abbr>project</abbr>. By
+default this command only cleans up the local cache, which is typically located
+on the same machine as the project in question; This usually helps to free up
```
shcheklein: `; This` --> `. This`?

static/docs/commands-reference/gc.md
```diff
@@ -11,33 +11,41 @@ usage: dvc gc [-h] [-q | -v] [-a] [-T] [-c] [-r REMOTE] [-f] [-j JOBS]
...
-Note that unless `--cloud` is specified, this action does not remove data files
-from remote storage. This command is just a way to clean the working cache which
-is usually located on the machine your are running experiments on, so this
-usually helps to save some space.
+Data added to DVC in a different branch or version than the current Git commit
+(`HEAD`), it will be deleted by `dvc gc` unless the `--all-branches` or
+`--all-tags` options are used. Note that if the cache/remote holds several
+versions of the same data file, all except the current one will be deleted.
+
+Unless the `--cloud` option is used, this action does not remove data files from
```
shcheklein: I think `--remote` as well

static/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Done. And reviewed term in e17ab1a.

static/docs/commands-reference/gc.md
jorgeorpinel: Agree. Fixing

static/docs/commands-reference/gc.md
jorgeorpinel: Hmm. No, `-r` is to specify which remote (other than the default) when `-c` is used. Clarified in 0813129.

jorgeorpinel:
This may be ready to merge. Please check @shcheklein 

I'll then open a PR for https://github.com/iterative/dvc.org/pull/637#pullrequestreview-293496064 (renaming `commands-reference`) and the main purpose of this review: new use case with uncompressed cat and dog image dataset versions!

static/docs/commands-reference/gc.md
```diff
@@ -47,9 +47,10 @@ anytime you want **as long as they have previously been pushed**. (See
   this option can be used to specify a list of them (each project is a path) to
   keep data that is currently referenced from them.
 
-- `-c`, `--cloud` - also remove files in the remote storage. _This operation is
-  dangerous._ It removes datasets, models, other files that are not linked in
-  the current branch/commit (unless `-a` or `-T` is specified).
+- `-c`, `--cloud` - also remove files in the default remote storage. _This
+  operation is dangerous._ It removes datasets, models, other files that are not
+  linked in the current branch/commit (unless `-a` or `-T` is specified). Use
+  `-r`to specify which remote to collect from (instead of the default).
```
shcheklein: `r`specify -> `r` specify

static/docs/commands-reference/gc.md
shcheklein: Yep, you are right! We changed the `dvc status` behavior in regards of `--cloud` and `--remote` Probably it's better to make them consistent as part of `dvc gc` redesign (cc @pared)

shcheklein:
@jorgeorpinel done!

static/docs/user-guide/external-cache-and-outputs.md
jorgeorpinel: Done in 060c700.

PR #649, opened 2019-09-27
documentation example gives error

amundv:
dvc version v0.6 currently gives error message
'ERROR: ambiguous option: --a could match --all-branches, --all-tags'
This fix clears that error.


shcheklein:
@amundv thanks! in the docs it is `-a` not `--a`. Have you tried to run it exactly like in the document?

amundv:
@shcheklein wops thanks :)

PR #653, opened 2019-09-27
Add mobile view for tooltip

algomaster99:
Reference #461 

Screenshot from 2019-09-28 05-02-34

### DIFF
--- a/src/Layout/index.js
+++ b/src/Layout/index.js
@@ -34,6 +34,7 @@ class Layout extends Component {
           {children}
           <Footer isDocPage={this.isDocPage} />
         </Bodybag>
+        <ModalRoot id="modal-root"></ModalRoot>
       </Wrapper>
     )
   }
@@ -69,4 +70,10 @@ const Bodybag = styled.div`
     scroll-behavior: smooth;
     will-change: scroll-position;
   `}
+`
+
+const ModalRoot = styled.div`
+  position: fixed;
+  z-index: 100000;
+  height: 100%;
 `
--- a/src/Layout/index.js
+++ b/src/Layout/index.js
@@ -75,5 +75,4 @@ const Bodybag = styled.div`
 const ModalRoot = styled.div`
   position: fixed;
   z-index: 100000;
-  height: 100%;
 `
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -143,12 +143,13 @@ class DesktopView extends Component {
           </TooltipContainer>
         )}
         <HighlightedText
+          id={`tooltip-text-${this.props.id}`}
           onMouseOver={this.hoverIn}
           onMouseLeave={this.hoverOut}
           onFocus={this.hoverIn}
           onBlur={this.hoverOut}
         >
-          <span id={`tooltip-text-${this.props.id}`}>{this.props.text}</span>
+          {this.props.text}
         </HighlightedText>
       </>
     )
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -1,0 +1,221 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import ReactMarkdown from 'react-markdown'
+import styled from 'styled-components'
+
+class DesktopView extends Component {
+  state = {
+    hover: false,
+    margin: -70,
+    pointBorderAfter: 'white transparent transparent transparent',
+    pointBorderBefore: '#d1d5da transparent transparent transparent',
+    pointMargin: -15,
+    pointTop: 100,
+    pointTopAfter: -14,
+    pointTopBefore: 16,
+    top: 'unset',
+    width: 400
+  }
+
+  tooltipPositionEval = () => {
+    const headerHeight = document.getElementsByClassName('header')[0]
+      .offsetHeight
+    const markdownBody = document.getElementsByClassName('markdown-body')[0]
+    const tooltipBoundary = document
+      .getElementById(`tooltip-text-${this.props.id}`)
+      .getBoundingClientRect()
+    const tooltipBoxHeight = document.getElementById('tooltip-box').offsetHeight
+    const tooltipHeight = tooltipBoundary.top - tooltipBoxHeight
+    const maxWidth = markdownBody.offsetLeft + markdownBody.clientWidth
+    const container = document.getElementsByClassName('tooltip-container')[0]
+    const tooltipWidth = container.offsetLeft + this.state.width
+    const vertical = tooltipHeight > headerHeight ? 'top' : 'bottom'
+    const horizontal = tooltipWidth > maxWidth ? 'right' : 'left'
+
+    switch (`${horizontal} ${vertical}`) {
+      case 'left top':
+        this.setState({
+          margin: -10,
+          pointBorderAfter: 'white transparent transparent transparent',
+          pointBorderBefore: '#d1d5da transparent transparent transparent',
+          pointMargin: -15,
+          pointTop: 100,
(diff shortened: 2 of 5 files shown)


src/Tooltip/mobile-view.js
```diff
@@ -0,0 +1,123 @@
...
+    })
+  }
+
+  render() {
+    return (
+      <>
+        <HighlightedText onClick={this.openTooltip}>
+          {this.props.text}
+        </HighlightedText>
+        {this.state.visible && (
```
shcheklein: this will be inserting `div` into DOM after the term, right? It breaks the layout sometimes in case like this `term.`. Browsers stop treating `word.` as a single non-wrappable entity and treat `word` and `.` as two separate items and in certain cases it leads to re-layout when `word` moves to the previous line and `.` stays on the same when tooltip `div` is visible. The workaround is to move hidden div _before_ the term. Not ideal, but will solve 99% of the problems.

src/Tooltip/desktop-view.js
```diff
@@ -0,0 +1,222 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import ReactMarkdown from 'react-markdown'
+import styled from 'styled-components'
+
+class DesktopView extends Component {
+  state = {
+    hover: false,
+    margin: -70,
```
shcheklein: probably not related to this PR even, but it's a really bad practice to hard code constants like this. At the very least there should be constants defined in this file with some explanations where do they come from. Values should be defined as formulas that use those constants. Ideally, certain constant values should be taken from other globals - with of the sidebar, for example, etc.

shcheklein:
@algomaster99 take a look at the CodeClimate complaints. Probably some parts could be generalized, deduplicated.

src/Tooltip/mobile-view.js
```diff
@@ -0,0 +1,123 @@
+import React, { Component } from 'react'
+import ReactMarkdown from 'react-markdown'
+import ReactDOM from 'react-dom'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+class MobileView extends Component {
```
shcheklein: @iAdramelk Alex, you have mentioned that it's better to use pure react component, could you please share the link again?

src/Tooltip/mobile-view.js
algomaster99: No, it won't. I have made a `div` tag with ID - `modal-root` which is at the same level as `bodybag`. It inserts the modal inside it. I have followed [React Portals](https://reactjs.org/docs/portals.html) tutorial for this.

src/Tooltip/desktop-view.js
algomaster99: @shcheklein I agree. I will define them as constant and write comments for explanation.

src/Tooltip/mobile-view.js
algomaster99: @shcheklein This component is a pure component. But it is not inherited from one.

src/Tooltip/mobile-view.js
shcheklein: 👍

algomaster99:
@shcheklein 
Screenshot from 2019-09-28 06-22-34
How this is similar isn't making sense to me.

shcheklein:
@algomaster99 codeclimate produces false positives indeed, treat it as warnings/recommendations. If it's clear that it is a false positive, not new or something the is a button per each item to close the issue (Invalid, Not New, etc)

jorgeorpinel:
Looks great, thanks @algomaster99 

I wasn't really thinking on a typical popup with margins everywhere though, but more of a "frame" design where the tooltip takes the full lower 1/3 (more or less) of the screen, kind of like this:

image

What do you think? Also the x (to close tooltip) should definitely be larger please, for easier touch.

algomaster99:
@jorgeorpinel Might be a great idea but since there is an overlay (translucent black), scrolling won't be possible. I will have to see. Thanks for the suggestion though!

shcheklein:
`X` (close button) - yes, we can make it larger, or write somewhere that touch to any part of the screen closes the message.

@algomaster99 @jorgeorpinel I totally fine with current implementation - it's simple and doing it's job. I've asked @iAdramelk to review as well. As soon as it's done will be happy to merge. 

@algomaster99 please, create a ticket to refactor hard coded constants.

src/Tooltip/mobile-view.js
iAdramelk: Hi. I think we can leave <Tooltip> as is here. They are used inside of <Markdown> and it is already a PureComponent, so there will not be unnecessary rerenders anyway.

shcheklein:
@algomaster99 it looks like styles are broken, inline code sections do not work in the mobile view. I haven't checked other stuff.

src/Layout/index.js
```diff
@@ -70,3 +71,9 @@ const Bodybag = styled.div`
     will-change: scroll-position;
   `}
 `
+
+const ModalRoot = styled.div`
+  position: fixed;
+  z-index: 100000;
+  height: 100%;
```
iAdramelk: Can we remove `height: 100%` from here and move it to Tooltip? We can use portal for more than one type of modals in the future and not all of them may need 100%.

src/Tooltip/mobile-view.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+MobileView.propTypes = {
+  description: PropTypes.string.isRequired,
+  header: PropTypes.string.isRequired,
+  text: PropTypes.node.isRequired
+}
+
+class Modal extends Component {
+  constructor(props) {
```
iAdramelk: Can we use function component here? There is no state or lifecycle methods in it.

iAdramelk: Also can we move portal creation logic to separate component? I think we will want to use it for other modals in the future too, and we will probably will rewrite desktop tooltips with portals too.

algomaster99:
@shcheklein which sections? I don't get you. Can you also share a screenshot?

shcheklein:
@algomaster99 

Screen Shot 2019-09-30 at 2 25 17 PM

algomaster99:
@shcheklein Oh okay. I don't know why that isn't working. I am confused too. :confused:

iAdramelk:
> @shcheklein Oh okay. I don't know why that isn't working. I am confused too. 😕

I think that this is because it is now rendered outside of markdown component. You need to add `className="markdown-body"` to the markdown root, to make global markdown styles work.

src/Tooltip/desktop-view.js
algomaster99: Moved conversation here: #663

algomaster99:
@iAdramelk 
I added `markdown-body` to `<ReactMarkdown className="markdown-body" source={this.props.description} />` which fixes the styles of inline-code but removes the font.
Screenshot from 2019-10-02 18-44-35

Should I add another style, something like `.portal` and hardcode the font-family there?

iAdramelk:
> Should I add another style, something like `.portal` and hardcode the font-family there?

@lgomaster99  yep, I think we need to add second class with font here. Not sure about the name through. I think portal and it's content should be separate entities:

1. Portal is just a mount point where we can add arbitrary html. It shouldn't have it's own styles.
2. Tooltip visual styles that can be reused between desktop and mobile version. In perfect world they didn't have ant position logic inside and can be placed anywhere.
3. Tooltip positioning logic. It can be either separate, or the part of the unified tooltip component.

In this logic font is a part of the Tooltip visual styles and should be placed near them.

algomaster99:
@shcheklein This can be merged now. 💯

PR #659, opened 2019-09-30
soft-code all style constants in tooltip file

### LINKED ISSUE #658: glossary: soft code style constants in tooltip file
Soft code all style constants inside `state`.
https://github.com/iterative/dvc.org/blob/c049220f41616384cb85e695af869056c8e82e32/src/Tooltip/index.js#L11

algomaster99:
Fixes #658

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -7,21 +7,79 @@ import includes from 'lodash.includes'
 import glossary from '../Documentation/glossary'
 import { OnlyDesktop, OnlyMobile } from '../styles'
 
+/**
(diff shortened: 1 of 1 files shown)


algomaster99:
@shcheklein I have derived all those constants from hit and trial so I don't have an exact idea how everything works.
When I was creating it, I didn't know if something like react portals exist. I am thinking to make.this again but using React Portals this time.

shcheklein:
@algomaster99 then, let's close this and create a new to migrate to portals. Probably we'll share some logic with the new mobile implementation? Also, @iAdramelk can have some ideas how it can refactored.

algomaster99:
Created a ticket - #663

PR #661, opened 2019-10-01
cmd ref: change "Commands Reference" for "Command Reference" (all singular)

jorgeorpinel:


jorgeorpinel:
Do we need a (temporary) redirect in the DNS or app for old `/doc/commands-reference/...` paths?

shcheklein:
👍 @jorgeorpinel yes, let's add a redirect in server.js for convenience

jorgeorpinel:
Redirect added @shcheklein. Try https://dvc-org-pr-661.herokuapp.com/doc/command-reference fore example. Added some comments in the latest commit, which adds the redirect. Please review.

server.js
```diff
@@ -39,16 +41,22 @@ app.prepare().then(() => {
...
       res.writeHead(301, {
         Location:
           'https://s3-us-east-2.amazonaws.com/dvc-public/' +
           req.headers.host.split('.')[0] +
           pathname
       })
       res.end()
-
-      // Special URL path redirects below
+    } else if (/^\/doc\/commands-reference.*/.test(pathname)) {
```
shcheklein: it will match `/commands-referencesdsdsdsd/something`, let's make it something like

`/^\/doc\/commands-reference(\/.*/)?`

?

jorgeorpinel: Sure, just need to make that `/` optional for path `/doc/commands-reference`. Fixed in 1eaca99.

jorgeorpinel: Oh wait I see your regex also achieves that. Looks better let me try that...

jorgeorpinel: Yes. (Except your last `/` should be after `?` 😋  Updated and tried with paths `/doc/commands-reference`, `/doc/commands-reference/add`, and `/doc/commands-referenceadd`. see latest commit.

Issue #608, opened 2019-09-05
term: misc. clarifications


jorgeorpinel:
> Continuation of #448

- ~~Make consistent use of "`name.dvc` stage file" vs "'name' stage" phrases.~~ (no longer applies in DVC 1.0)
- [x] Look up for phrases like ~~"should",~~ ~~"obvious"~~, "you probably noticed", "as seen here" (without explanation), etc. to make the wording friendlier. Especially in introductory docs and guides (e.g. Get Started) - continued in #1394
- [ ] "data remote" / "DVC remote" -> "remote" / "remote storage"?
- ~~"stage file (DVC-file)" -> "stage file (DVC-file with dependencies)"~~
- [x] Avoid "entry" when in the context of cache files. Use "data", "files", "directories" as appropriate.
- [ ] Review usage of "DVC" branding of terms (i/[^`]DVC / regex): Avoid "DVC cache, "DVC pipeline" (others) but keep when necessary, specially "DVC project", "DVC remote" (others)
  UPDATE: Done up to contents/docs/commands-reference/metrics
- [x] review usage of "which"
- [x] avoid "flag" in the context of command options. Favor "option".
- [ ] Review usage of "version", "revision", and "reference" per https://github.com/iterative/dvc.org/issues/605#issuecomment-544187967
- [x] GitHb vs. Github (look for similar cases)
- [ ] stage **output**(s) vs. command **output** (what's printed to stdout)
- [x] stage output vs. data artifacts: they shouldn't be synonyms. Let's just remove "artifact" from the glossary? #1887
- [x] Double check correct usage of singular "metric" vs plural "metrics"
- [ ] DAG vs dependency graph and which link to use? https://en.wikipedia.org/wiki/Directed_acyclic_graph vs https://en.wikipedia.org/wiki/Dependency_graph
- [ ] Double check correct usage of "metric file" vs "metrics file" (both singular)
- [ ] "with Git" -> "to Git" https://trends.google.com/trends/explore?geo=US&q=add%20to%20git,add%20with%20git
- [ ] status vs. state
- [ ] Decide whether to use term pipeline, DAG, dependency graph, and when, and apply. See https://github.com/iterative/dvc.org/pull/1383#discussion_r433398546
- [x] Eliminate "import stage" term. Just use imported `.dvc` file or something like that

jorgeorpinel:
See also #552.

shcheklein:
Can be closed by now?

jorgeorpinel:
[...]

PR #667, opened 2019-10-02
docs: regular updates (early Oct)

jorgeorpinel:


jorgeorpinel:
Thanks! That was fast... I wanted to include some of the stuff from #608, but OK I'll do that later and focus on finishing the new use case now.

PR #647, opened 2019-09-26
glossary: make tooltip IDs unique

algomaster99:
Fixes #644

PR #671, opened 2019-10-03
Even more regular updates (early Oct) towards new Use Case

jorgeorpinel:


package.json
```diff
@@ -6,10 +6,9 @@
   "scripts": {
     "dev": "node server.js",
     "build": "next build",
-    "start": "NODE_ENV=production node server.js",
     "test": "jest",
-    "prettier-js": "prettier --write '{pages,src}/**/*.{js,jsx}'",
-    "prettier-md": "prettier --write 'static/docs/**/*.md'",
+    "start": "NODE_ENV=production node server.js",
+    "pretty": "pretty-quick --staged --no-restage --bail",
```
jorgeorpinel: As noted in https://github.com/iterative/dvc.org/pull/647#issuecomment-538110156, I realized we have 2 scripts for similar purposes: `yarn pretty` (uses `pretty-quick`) and `yarn lint` (uses `eslint`). Both are enforced locally with a Git pre-commit hook (see line 71 of this same file) but this could be disabled by the developer. CircleCI only enforces the latter, `yarn lint`.

Is this correct? Seems strange to me @shcheklein and @iAdramelk.

jorgeorpinel: p.s. I cleaned up some old linting scripts as you can see in e74cbc1.

package.json
```diff
@@ -6,10 +6,9 @@
...
     "lint": "yarn run eslint src"
```
jorgeorpinel: As noted in https://github.com/iterative/dvc.org/pull/647#issuecomment-538110156, I realized we have 2 scripts for similar purposes: `yarn pretty` (uses `pretty-quick`) and `yarn lint` (uses `eslint`). Both are enforced locally with a Git pre-commit hook (see line 71 of this same file) but this could be disabled by the developer. CircleCI only enforces the latter, `yarn lint`.

Is this correct? Seems strange to me @shcheklein and @iAdramelk.

---

p.s. I cleaned up some old linting scripts as you can see in e74cbc1.

static/docs/use-cases/shared-development-server.md
```diff
@@ -69,10 +69,10 @@ $ git commit -m "cleanup raw data"
 $ git push
 ```
 
-Your colleagues can pull the project, and have both `raw` and `clean` data files
-instantly appear in their workspace without moving anything manually. After
-this, they could decide to continue building this pipeline and process the
-cleaned up data:
+Your colleagues can [checkout](/doc/command-reference/checkout) the project
+data, and have both `raw` and `clean` data files instantly appear in their
```
jorgeorpinel: BTW, "instantly appear"? Actually DVC has been becoming slower recently. Should we use the term "quickly" instead on "instantly" in most docs? (except perhaps marketing-oriented pages) or is this slowing temporary @shcheklein ? (In any case it will never be truly instantaneous.)

package.json
iAdramelk: Yes, we should add prettier to the CI too. But it should be validate-only task, without attempts to fix code like it pre-commit hooks.

jorgeorpinel: OK, changed in 0e82cd7 can you please check? Seems to work based on [the corresponding CI job](https://circleci.com/gh/iterative/dvc.org/992?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link):

image

But maybe we should test committing badly formatted code...

jorgeorpinel: Wait actually why did it find 0 changed files? That's not right 🙁 ...

jorgeorpinel: Oh wait maybe that's OK because there are no JS or MD file changes in that commit 😋

static/docs/use-cases/shared-development-server.md
shcheklein: in case of shared dev server they supposed to use some kind of links (should we review and update the page?). In this case it should more or less instant. I agree though that `quickly` is a safer choice though. No copying is happening. @efiop could you chime in please? Is it instant now if it's a shared cache?

jorgeorpinel: I thought about it and adding a note about file links in this document didn't seem necessary. It doesn't change anything, it would just be a disclaimer for the term "instant" which seems like we want to change anyway, right?

BTW I'm on a Mac so it should be using reflinks. Still all `dvc` commands take several seconds, even `dvc version`. There's some sort of noticeable overhead both in my pip-installed `dvc` and on the actual core repo in dev mode.

jorgeorpinel: p.s. updated wording in 4c42677 so feel free to resolve this conversation if there's nothing more to add 🙂

package.json
jorgeorpinel: Hmmm. This isn't working actually. See https://circleci.com/gh/iterative/dvc.org/1000. Any thoughts on how to setup `pretty-quick` correctly here @iAdramelk?

- [ ] Remove test commit be661b5 from this branch before trying to merge PR...

static/docs/use-cases/shared-development-server.md
shcheklein: @jorgeorpinel in case of data if time is a few seconds we can def consider that as instantaneous :)

package.json
shcheklein: @jorgeorpinel should you specify a key to force it run the check on all files (vs only changed - as it seems that's what it's trying to do - and thus there are 0 of them)

jorgeorpinel: When I run `yarn pretty-quick --check --verbose` on this branch locally, even with a clean Git working tree, it finds 11 changed files. I'm not sure what it's comparing it to. Maybe master? But there's 13 files changed in this PR so. Actually it's pretty mysterious... I wonder how the CI sets up the git repo actually, let me check.

jorgeorpinel: The CI seems to setup the git repo correctly, I even tried replicating how it checks out code, but I still see 12 changed files when running `yarn format-check`, while CircleCI sees 0. 🤷‍♂

static/docs/use-cases/shared-development-server.md
jorgeorpinel: Idk. "instantaneous" seems misleading to me.

image

And `dvc version` takes 13 seconds on my Mac BTW ):  But that's a separate problem I guess will be addressed when a performance profiler system is established (:

shcheklein: `dvc version` 13 seconds is a bug!!! Please, create a ticket at `iterative/dvc` to investigate this.

```
$ time dvc versions

real    0m0.140s
user    0m0.117s
sys     0m0.024s
```

on https://katacoda.com/dvc/courses/basics/data

package.json
shcheklein: 👀

shcheklein: @jorgeorpinel added an extra step to fix this

static/docs/use-cases/shared-development-server.md
jorgeorpinel: Opened https://github.com/iterative/dvc/issues/2582

package.json
jorgeorpinel: Thank you!

jorgeorpinel:
Note: I reverted be661b5 with 41c49ef instead of removing via force-push (since Ivan added a commit in between). Better to keep history linear (:

scripts/prettier.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/bash
```
shcheklein: why removing this? 

Check if scripts is needed at all after that.

Check if there are references to prettier.sh somewhere in docs.

package.json
```diff
@@ -6,11 +6,11 @@
   "scripts": {
     "dev": "node server.js",
     "build": "next build",
-    "start": "NODE_ENV=production node server.js",
     "test": "jest",
-    "prettier-js": "prettier --write '{pages,src}/**/*.{js,jsx}'",
```
shcheklein: after removing the script below and these targets there will be no way to run the prettier on some files before committing.

Also, we are removing patterns - does it mean that prettier will be on any files at all? Not sure if it's a big deal or now.

scripts/prettier.sh
jorgeorpinel: I'm pretty sure it was an old way to do what we have in yarn scripts now. Do you want it back just in case? It's not used anywhere in the repo anymore.

I had forgotten to update the contributing docs page though, just did in 38ec8ae.

package.json
jorgeorpinel: Prettier can still be run with `yarn yarn format-staged` as explained in the docs contrib guide or `yarn pretty-quick ...` or even `yarn prettier ...` But again, if you prefer to have these here I can bring them back. We should just document their existence and purpose somewhere, at least in the README file?

scripts/prettier.sh
shcheklein: I'm not sure yarn scripts can do that (especially the way they are written now - only run on staged files).I think the idea was to have a mechanism to run prettier on a specific files/files w/o running `git commit` or even `yarn something` (which is not possible now since it's only prettier-staged which is the same as running git hook?)

jorgeorpinel: Is it a common need to run prettier on arbitrary files in our repo? I'm not sure it is but I would just document Prettier usage examples in the docs contrib guide such as `yarn prettier --write '{pages,src}/**/*.{js,jsx,md}'`. But is it goal of this repo to teach Prettier usage?

jorgeorpinel: I added the note in 9ccfb51 (and made some other improvements about this in general). What do you think?

package.json
jorgeorpinel: This discussion is the same as in https://github.com/iterative/dvc.org/pull/671#discussion_r331810661 🙂

scripts/prettier.sh
dashohoxha: > Is it a common need to run prettier on arbitrary files in our repo?

At some point in time the hooks that fix the formatting were not working properly. The reason for writing this script was to merge contributions from community even if they were not 100% conforming to the formatting rules, then run this script manually once in a while to correct any formatting issues. I am not sure if this is needed anymore.

But I think that it still might be useful to be able to check and fix the files before staging them.

> I would just document Prettier usage examples in the docs contrib guide such as `yarn prettier --write '{pages,src}/**/*.{js,jsx,md}'`

If you write a small script that does this, then there would be no need to document its usage.

jorgeorpinel: > If you write a small script that does this, then there would be no need to document its usage.

Yes, but that's not the only possible usage of Prettier, it could be something else you want to do (a specific folder, only md files, only md and jss, etc.) and having an entire sh script to wrap a simple `prettier` call with a few arguments seems overkill to me. Check out the note about using Prettier via `yarn` I added to the docs contrib guide. I hope this, the link to `pettier` usage, and the `js,jsx,md` example should be enough 🙂

shcheklein: k, let's simplify this and see if we get some feedback that something is not clear/convenient.

Issue #531, opened 2019-08-07
glossary tooltip: side menu button wrong z-positioning


shcheklein:
Hey, @algomaster99 . Please, check it out:

<img width="637" alt="Screen Shot 2019-08-07 at 4 19 14 PM" src="https://user-images.githubusercontent.com/3659196/62664159-5f00cc00-b92f-11e9-852d-4e50598127ed.png">

shcheklein:
Latest valid example - https://dvc-org-pr-606.herokuapp.com/doc/commands-reference/gc

shcheklein:
Fixed by the migration to Gatsby.

PR #675, opened 2019-10-04
Rewrite tooltip for desktop view using React Portals

### LINKED ISSUE #663: glossary: rewrite tooltip (desktop-view) using "Portals"?
React [Portals](https://reactjs.org/docs/portals.html) is a good way to create elements which hover above the DOM. Currently, it is created using `position: fixed` and other relevant styles. I think implementing this will make it more polished.

algomaster99:
Fixes #663 

### Improvements
1. Written using react portals.
2. The pointer looks uniform now.

##### Old

|   | Top | Bottom |
|---|---|---|
| **Right** | right-top | right-bottom |
| **Left** | left-top | left-bottom |

##### New

|   | Top | Bottom |
|---|---|---|
| **Right** | right-top | right-bottom |
| **Left** | left-top | left-bottom |

3. `<HighlightedText>` has been moved to `index.js` since it is common for both views.

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -68,6 +68,7 @@ Tooltip.propTypes = {
 
 export const HighlightedText = styled.span`
   border-bottom: 1px black dotted;
+  white-space: nowrap;
 `
 
 export default Tooltip
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import includes from 'lodash.includes'
+import styled from 'styled-components'
 
 import glossary from '../Documentation/glossary'
 import { OnlyDesktop, OnlyMobile } from '../styles'
@@ -64,5 +65,9 @@ Tooltip.propTypes = {
   id: PropTypes.string.isRequired,
   text: PropTypes.node.isRequired
 }
+
+export const HighlightedText = styled.span`
+  border-bottom: 1px black dotted;
+`
 
 export default Tooltip
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -12,6 +12,25 @@ class DesktopView extends Component {
     width: 400
   }
 
+  componentDidMount() {
+    document.getElementById('bodybag').addEventListener('scroll', () => {
+      if (this.state.hover) {
+        const headerHeight = document.getElementsByClassName('header')[0]
+          .offsetHeight
+        const tooltipBoundary = document
+          .getElementById(`tooltip-text-${this.props.id}`)
+          .getBoundingClientRect()
+        if (tooltipBoundary.top < headerHeight) {
+          this.setState({
+            hover: false
+          })
+        } else {
+          this.tooltipPositionEval()
+        }
+      }
+    })
+  }
+
   tooltipPositionEval = () => {
     const headerHeight = document.getElementsByClassName('header')[0]
       .offsetHeight
@@ -94,12 +113,13 @@ class DesktopView extends Component {
   }
 
   hoverOut = () => {
+    const TIMEOUT = 50 // time in milliseconds
     this.setState({
       interval: setTimeout(() => {
         this.setState({
           hover: false
         })
-      }, 100)
+      }, TIMEOUT)
     })
   }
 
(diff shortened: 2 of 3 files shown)


algomaster99:
@shcheklein Please merge this PR first before merging #653 .

shcheklein:
Cools stuff! 

Screen Shot 2019-10-04 at 10 29 11 AM
Screen Shot 2019-10-04 at 10 29 23 AM
Screen Shot 2019-10-04 at 10 29 35 AM

@algomaster99 are those intended?

algomaster99:
@shcheklein In the first screenshot, the pointer shouldn't be so much above the word. You would have scrolled upwards a little.
All other overlapping behavior is intended.

shcheklein:
@algomaster99 nope, not scrolled.

algomaster99:
@shcheklein Okay, will look into it. Thanks!

algomaster99:
Screenshot from 2019-10-05 02-33-25

It calculates the position according to the location of the highlighted text. Since the selected element here is spanning over the whole paragraph, it fails to calculate it correctly. How can I find the exact position of the highlighted text?
@shcheklein @iAdramelk 

Look at this for example:-
Screenshot from 2019-10-05 02-41-19

algomaster99:
@shcheklein You can test it now :)

src/Tooltip/index.js
```diff
@@ -65,4 +66,9 @@ Tooltip.propTypes = {
   text: PropTypes.node.isRequired
 }
 
+export const HighlightedText = styled.span`
+  border-bottom: 1px black dotted;
+  white-space: nowrap;
```
shcheklein: nice trick, haha :)

shcheklein:
@algomaster99 is actually possible to hide it right after we scroll the pointer out?

shcheklein:
Btw, it probably fixes #531

algomaster99:
@shcheklein 
> is actually possible to hide it right after we scroll the pointer out?

I don't get it. Do you mean that we hover over the text once and then scrolling should hide the pointer?

shcheklein:
@algomaster99 yes, right now if I scroll, the tip scrolls with me and stays open for a while. Ideally it should be hidden sooner, or at least stay attached to the text.

algomaster99:
@shcheklein Can you check it now? I think the performance decreases a little while scrolling. Let me know if you can notice.

shcheklein:
@algomaster99 scroll just does not work for me now if mouse pointer is on top of a tooltip area (and it's very easy to get into it and get stuck). Also, is there a simpler way to achieve this?

algomaster99:
@shcheklein scroll can't work inside because tooltip technically is outside the DOM tree.

shcheklein:
@algomaster99 I'm not an expert in React to judge the specific implementation, but in this case I think then the portal transition defeats the purpose? Check the wikipedia site and the tooltips (page previews) they have. This is the behavior we would like to achieve - multiline support, smooth scroll, ideally no overlap with boundaries - it's better to always show it over the text area, etc.

algomaster99:
@shcheklein Yeah, I saw it on GitHub too. I don't think it is possible to make `modal-root` move relative to `<HighlightedText>`. If @iAdramelk has some suggestions, please drop it here. Otherwise, we might need to revert to previous view. :confused:

iAdramelk:
@algomaster99 Sorry, just returned to the places that have internet. Let me think about it a little. I have couple of ideas, but want to test them first.

algomaster99:
@iAdramelk What do you suggest?

iAdramelk:
@algomaster99 sorry for delay. I tried different options. But with didn't find solution that didn't require global update to the layout. So we probably should fallback for now.

What I tried:

1. Updating position on scroll: Either freezing browser if we update to often or we have visual lag between scroll and tooltip position update.
2. Moving portal inside markdown container and use absolute for desktop and fixed for mobile: Not working good because it's parent and body also has fixed styles and body has to be under the header.

We can probably fix it if we remove fixed from the content container and do some other changes to layout, but not sure that we should do it as the part of the current issue.

algomaster99:
@iAdramelk moving portal inside the markdown container will affect the mobile view so it is not a great idea to proceed with.
The tooltip container has to be absolutely positioned with the text and I think that is not possible if we wire it up inside the portal whose position is at root level in DOM.

iAdramelk:
@algomaster99 yep, this ideas didn't work. :) So we should return original version for now.

algomaster99:
@shcheklein @iAdramelk closing this for now.

PR #680, opened 2019-10-08
More regular updates (Oct)

jorgeorpinel:


static/docs/command-reference/remote/index.md
```diff
@@ -67,9 +67,7 @@ For the typical process to share the <abbr>project</abbr> via remote, see
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
-## Examples
-
-1. Let's for simplicity add a default local remote:
+## Example: Add a default local remote:
```
shcheklein: I would say that's not what we usually do in other place - we don't use second level headers to split steps. Rather we use them to split multiple examples.

static/docs/understanding-dvc/how-it-works.md
```diff
@@ -1,93 +1,93 @@
...
-   $ git pull # download tracked data from remote storage
-   $ dvc checkout # checkout data files
-   $ ls -l data/ # You just got gigabytes of data through Git and DVC:
-
-   total 1017488
-   -r--------  2 501  staff   273M Jan 27 03:48 Posts-test.tsv
-   ```
-
-8. DVC works on Mac, Linux, and Windows.
+- DVC is a command line tool that works on top of Git:
```
shcheklein: we should really get reed of all this understanding-dvc section. A lot of these materials overlaps with other tutorials, a lot of them should go into user guide.

static/docs/understanding-dvc/related-technologies.md
```diff
@@ -1,22 +1,24 @@
...
+comparison to existing technologies and tools.
 
-DVC combines a number of existing technologies and ideas into a single product
-with the goal of bringing the best engineering practices into the data science
-process.
+DVC combines a number of existing ideas into a single product, with the goal of
+bringing best practices from software engineering into the data science field.
 
-1. **Git**. The difference is:
+## Differences with related tools
```
shcheklein: it repeats the main title?

shcheklein: we can easily do `##` (second level) here instead of numbered list

static/docs/command-reference/remote/index.md
jorgeorpinel: I don't think these are steps. I see these as totally separate/unrelated examples (except for sharing the name of the remote between the 2nd and 3rd ones). That's why I moved to the "Examples: ..." convention. They're pretty short examples, granted. Not sure they're worth putting here in the index even.

I could consolidate them into a single example that it a little more connected though. Would that be better?

static/docs/understanding-dvc/how-it-works.md
jorgeorpinel: True. That's a separate issue though, that I haven't been able to get to yet: #425

static/docs/understanding-dvc/related-technologies.md
jorgeorpinel: H2 looks too "important" for a series of short sections, it even has a horizontal bar underneath. That's why I wanted to use H3 below (e.g. `### Git`) but I didn't want to skip from H1 to H3 so I came up with `## Differences with related tools` which now that I think about it I agree sounds repetitive. I guess it doesn't matter to skip, I'll just remove it! We're getting rid of this entire section of the docs (hopefully) soon anyway.

static/docs/command-reference/remote/index.md
shcheklein: 👍 k, makes sense. I was probably confused by the colons at the end of the titles. Probably we should remove those. I think it's fine to keep them separate (and revisit maybe at some point to make them more meaningful, but it applies to a lot of other DVC command reference examples.)

static/docs/understanding-dvc/related-technologies.md
shcheklein: 👍

static/docs/command-reference/remote/index.md
jorgeorpinel: Oh yeah definitely should remove those colons! Will do. Thanks for the heads up.

jorgeorpinel: Done.

jorgeorpinel:
All done here I think.

PR #683, opened 2019-10-09
get-started: agenda copy edits

gagejustins:
Made a few copy edits!

### DIFF
--- a/static/docs/get-started/agenda.md
+++ b/static/docs/get-started/agenda.md
@@ -1,36 +1,35 @@
-# Agenda
+# Agenda
 
-In the next few sections we'll build a simple natural language processing (NLP)
-project from scratch. If you'd like to get the final result or have any issues
-along the way, you can download the fully reproducible
-[GitHub project](https://github.com/iterative/example-get-started) by running:
+In the next few sections we'll build a simple natural language processing (NLP)
+project from scratch. If you'd like to get the final result or have any issues
+along the way, you can download the fully reproducible
+[GitHub project](https://github.com/iterative/example-get-started) by running:
 
 ```dvc
-$ git clone https://github.com/iterative/example-get-started
+$ git clone https://github.com/iterative/example-get-started
 ```
 
-Otherwise, bear with us and we'll introduce some basic DVC concepts and get to
-the same results together!
+Otherwise, bear with us and we'll introduce some basic DVC concepts to get to
+the same result together!
 
-The idea of the project is a simplified version of our
-[simple text classification tutorial](/doc/tutorials/deep). It explores the NLP
-problem of predicting tags for a given StackOverflow question. For example,
-let's say we want a classifier that can identify posts about Python by tagging
-them with `python`.
+The idea of the project is a simplified version of our
+[text classification tutorial](/doc/tutorials/deep). It explores the NLP problem
+of predicting tags for a given StackOverflow question. For example, we might
+want a classifier that can flag (or predict) posts about Python by tagging them
+with `python`.
 
 ![](/static/img/example-flow-2x.png)
 
-This problem is natural language processing related, but NLP isn't the only Data
-Science context that DVC can help with. DVC is designed to be agnostic to
-frameworks, languages, and problem sets. If you _have_ data files or datasets
-and/or you _produce_ data files, models, or datasets and you want to:
+This is a natural language processing context, but NLP isn't the only area where
+DVC can help. DVC is designed to be agnostic of frameworks, languages, etc. If
+you _have_ data files or datasets and/or you _produce_ data files, models, or
+datasets and you want to:
 
-- Capture and save those <abbr>data artifacts</abbr> the same way you capture
-  code
-- Track and switch between different versions of data easily
-- Be able to understand how data artifacts (e.g. ML models) were built in the
-  first place
-- Be able to compare models to one another
-- Bring software best practices to your team and get everyone on the same page
+- Capture and save those <abbr>data artifacts</abbr> the same way you capture
+  code
+- Track and switch between different versions of data easily
+- Understand how data artifacts (e.g. ML models) were built in the first place
+- Be able to compare models to each other
+- Bring software best practices to your team and get everyone on the same page
 
-Then you're in the right place! Click the `Next` button below to start ↘
+Then you're in the right place! Click the `Next` button below to start ↘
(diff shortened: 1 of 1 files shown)


static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
 # Agenda
 
-In the next few sections we will build a simple natural language processing
```
jorgeorpinel: Question for you @gagejustins, what criteria do you recommend to decide on abbreviations such as "we'll", "they're", "we've", etc? We haven't come up with a definite "policy" so this is currently a (minor) problem throughout our docs. I think that we should NOT (vs. shouldn't 😋 ) abbreviate when the sentence needs to emphasize the 2nd word in the abbreviation, as long as it's an idea directly related to the document's topic. What do you think?

> We also some times may use large caps as in "should NOT" which I think should be avoided in the future once this policy is decided.

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
+project from scratch. If you'd like to get the final result or have any issues
+along the way, you can download the fully reproducible
+[GitHub project](https://github.com/iterative/example-get-started) by running:
 
 ```dvc
 $ git clone https://github.com/iterative/example-get-started
 ```
 
-Otherwise, bear with us and we will introduce the basic DVC concepts and get to
+Otherwise, bear with us and we'll introduce some basic DVC concepts to get to
```
jorgeorpinel: How about "to get the same results" instead? Two to's too many 😋

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
 the same result together!
 
-The idea of the project is a simplified version of the
-[Tutorial](/doc/tutorials/deep). It explores the NLP problem of predicting tags
-for a given StackOverflow question. For example, we want a classifier that can
-predict posts about the Python language by tagging them `python`.
+The idea of the project is a simplified version of our
+[text classification tutorial](/doc/tutorials/deep). It explores the NLP problem
```
jorgeorpinel: Thanks for the update. Let's call this "Deep Dive Tutorial" please, as in the navigation.

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
+of predicting tags for a given StackOverflow question. For example, we might
+want a classifier that can flag (or predict) posts about Python by tagging them
+with `python`.
```
jorgeorpinel: How about just "predict posts about Python by tagging them with `python`"? I'd rather avoid introducing the concept of flagging in this context (it's used for other purposes in our docs).

> In fact I'm going to add reviewing the usage of the term "flag" to #608. (Letting you know in case you want to address that one checkbox as part of this PR 🙂).
> image

static/docs/get-started/agenda.md
gagejustins: Great question! I think abbreviations for most pronouns tend to be more personal and inviting, which definitely vibes with DVC's brand. For emphasis I'd use italics, and agree that we shouldn't abbreviate in that case

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
 
 ![](/static/img/example-flow-2x.png)
 
-Do not let the NLP nature of the example discourage you from using DVC in other
-Data Science areas. There was no strong reason behind picking the NLP area. On
-contrary, DVC is designed to be agnostic of frameworks, languages, etc. If you
-have data files or datasets and/or you produce other data files, models,
+This is a natural language processing context, but NLP isn't the only area where
```
jorgeorpinel: Definitely better, thanks! Please add "....only area **of data sciences** where..."

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
+DVC can help. DVC is designed to be agnostic of frameworks, languages, etc. If
+you _have_ data files or datasets and/or you _produce_ data files, models, or
```
jorgeorpinel: No need to add _italic_ formatting here. We don' really use that much and I'm not seeing the special importance of terms "have" and "produce". Thanks for the suggestion though!

static/docs/get-started/agenda.md
jorgeorpinel: Agree! Let's use this practice @shcheklein? Except... not sure we want to start using _italics_ (as mentioned in one of my review comments below) – mainly because there's so many other docs we'd have to review and add them to in order to be consistent 😅. Oh, and we tend to use **bold** words more, although also not super consistently...

gagejustins: Gotcha, will adjust to using bold! An important thing to add for docs (at least in my experience) is that we'll never be perfect, and users rarely expect you to be. In fact, DVC's docs are _remarkably_ good and y'all are a tiny company

jorgeorpinel: Deciding a word-emphasizing policy/ best practice and reviewing existing docs seems to me an issue of its own so I opened #686 for now.

static/docs/get-started/agenda.md
gagejustins: That doesn't work well english wise - how about classify posts about python?

jorgeorpinel: Hm OK sure. I'm feeling apprehensive of the word "predict" here but I guess "classify" works. Or "predict which posts ... {something else}". Thanks!

static/docs/get-started/agenda.md
jorgeorpinel: I see this is addressed now. Resolving

static/docs/get-started/agenda.md
```diff
@@ -1,36 +1,35 @@
...
+[GitHub project](https://github.com/iterative/example-get-started) by running:
 
 ```dvc
 $ git clone https://github.com/iterative/example-get-started
 ```
 
-Otherwise, bear with us and we will introduce the basic DVC concepts and get to
-the same result together!
+Otherwise, bear with us and we'll introduce some basic DVC concepts to get the
+same results together!
```
jorgeorpinel: Not crazy about the "together" part now that I notice it. What do you think?

gagejustins: I think it's good! Makes it sound collaborative

gagejustins:
@jorgeorpinel we good to merge here? I can't because I'm not an authorized user

shcheklein:
thanks, @gagejustins ! 🎉

PR #688, opened 2019-10-09
cmd ref: fix some phrasing issues in gc

peter-schmidbauer:
Fixed some ambiguous and wrong phrasing for the dvc gc command reference.

### DIFF
--- a/static/docs/command-reference/gc.md
+++ b/static/docs/command-reference/gc.md
@@ -19,10 +19,11 @@ default this command only cleans up the local cache, that is typically located
 on the same machine as the project in question. This usually helps to free up
 disk space.
 
-Data added to DVC in a different branch or version than the current Git commit
-(`HEAD`), it will be deleted by `dvc gc` unless the `--all-branches` or
-`--all-tags` options are used. Note that if the cache/remote holds several
-versions of the same data file, all except the current one will be deleted.
+All cache data added to DVC in all branches other than the current Git commit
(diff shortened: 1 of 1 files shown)


static/docs/command-reference/gc.md
```diff
@@ -12,18 +12,17 @@ usage: dvc gc [-h] [-q | -v] [-a] [-T] [-c] [-r REMOTE] [-f] [-j JOBS]
...
-disk space.
-
-All cache data added to DVC in all branches other than the current Git commit
-(`HEAD`) will be deleted by `dvc gc` unless at least one of the options 
-`--all-branches` or `--all-tags` is used. Note that if the cache/remote holds 
-several versions of the same data file, all except the current one will be 
-deleted.
+in the cache (or [remote storage](/doc/command-reference/remote) if `-c` is
+used) but no longer referenced in [DVC-files](/doc/user-guide/dvc-file-format)
+currently [checked out](/doc/command-reference/checkout) in the
```
shcheklein: @jorgeorpinel 
`checked out` reference should go to Git, not DVC

phrase is hard to read. I would split it into two parts an explain in simpler word. `but no longer referenced in the workspace. It means that there are no DVC-file in the current tree that have pointers to this cached data inside them.` Or something along those lines.

jorgeorpinel: Good catch. I'm just going to remove "checked out" because we don't need to assume there's a Git repo involved in this first paragraph. This made me think of further rephrasing here, please see the latest version.

PR #689, opened 2019-10-09
Add styles to render captions for images

### LINKED ISSUE #685: image captions in Markdown
It would be cool to have the ability to add captions to images in the docs through Markdown. A caption would: 

- Be centered directly under an image
- Be smaller text than the normal paragraph font size
- Be italicized
- Have narrower margins than the normal paragraph margins

algomaster99:
Fixes #685 

## Caption preview
Screenshot from 2019-10-10 00-46-48

## How to put one?
```
![](/static/img/cats-and-dogs.jpg) _Dataset to classify cats and dogs_
```
This generates:
```
<p>
  <img src="/static/img/cats-and-dogs.jpg" />
  <em>Dataset to classify cats and dogs</em>
</p>
```

Now I styled the element using `img + em` selector.

## Styles

- `margin-top` is given as `-6px` so that there is `2px` space between the image and the caption. It is not noticeable here since the background colour of image is white.

- The colour is the same as the one used in quotes.
Screenshot from 2019-10-10 00-54-12

### DIFF
--- a/src/styles.js
+++ b/src/styles.js
@@ -30,6 +30,14 @@ export const global = `
 	*:focus {
     outline: 0;
   }
+
+  img + em {
+    color: #6a737d;
+    font-size: 0.9em;
+    display: block;
+    margin-top: -6px;
+    text-align: center;
+  }
 `
 
 const maxWidth = 1005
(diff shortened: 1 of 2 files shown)


jorgeorpinel:
Smart use of CSS [adjacent selectors](https://www.w3.org/TR/CSS21/selector.html#adjacent-selectors)! Good thinking @algomaster99, thanks!

p.s. Could maybe use `p > img + em` to be even more specific, but I don't foresee any issues with the current selector.

Issue #545, opened 2019-08-10
simplify get-started by removing references/links to advanced doc pages


dashohoxha:
After finishing the “Get Started” tutorial, I have a feeling that it is a bit more complex than it should be for a beginner. For example the links and references to more advanced and complex explanations only distract the reader and are not much useful.

My opinion is that a good first tutorial should be consistent, self-contained and sequential (reading it from the beginning to the end, like an article, without jumping forth and back to more advanced doc pages). If there is something that needs to be explained, it is much better to include a paragraph or a note to the text of the tutorial itself, instead of sending the reader to another page (“for more details click here”).

For example what is the point or benefit of referring the reader to a future chapter/section of the tutorial itself, when the reader will eventually get there if he just reads in order. (However it makes sense to refer back to past sections of the tutorial, to remind the reader something, if necessary).

So, my opinion is that Get Started may need to be simplified a bit.

shcheklein:
Agreed. The initial idea was to keep as simple as possible. But it has been growing with more and more details over the time. @dashohoxha it would be great to put a list of things you think can be simplified/removed so that we can discuss.

dashohoxha:
> @dashohoxha it would be great to put a list of things you think can be simplified/removed so that we can discuss.

Can I work on simplifying/improving it instead? The things that I will try to do:
- Move detailed installation instructions on a dedicated page (under User Guide).
- Remove references to more advanced docs or pages. Maybe simplify any explanations too.
- Put links to the interactive lessons (https://katacoda.com/dvc/courses/get-started).
- However this can be properly done if the structure is changed a bit. Also 16 sections are too much, I will try to reduce them to 8 (using subsections if needed).
- etc.

shcheklein:
@dashohoxha 

> Move detailed installation instructions on a dedicated page (under User Guide).

Yep.

> Remove references to more advanced docs or pages. Maybe simplify any explanations too.

Can you give some examples? I would say it depends. Like in the previous case - I think even if we move stuff out, it would makes sense to keep a link to the User Guide page.

> Put links to the interactive lessons (https://katacoda.com/dvc/courses/get-started).

Good. May be let's decide on when we have a run on Katacoda button? I think it'll solve this problem.

> However this can be properly done if the structure is changed a bit. Also 16 sections are too much, I will try to reduce them to 8 (using subsections if needed).

[...]

PR #695, opened 2019-10-14
user-guide: reorganize installation page into this section

### LINKED ISSUE #656: make a top-level (user-guide level?) installation page
Right now the installation page is a subsection of Get Started: https://dvc.org/doc/get-started/install
I think it should be a top-level section, that includes all the details about installation that are scattered on the other pages (like Bash autocompletion, installation just for testing, installation for development, IDE integrations, etc.

dashohoxha:
Fix #656
Related: https://github.com/iterative/dvc.org/issues/545#issuecomment-541577615, https://github.com/iterative/dvc/issues/2567#issuecomment-538624649


dashohoxha:
> 1. The index page (index.md or something) that also serves as a main installation guide - usually it serves a an overview, contents, etc. It feels weird that I see this page and Mac OS simultaneously - should I read this and/or open Mac OS - it's confusing.

But `pip` and `conda` are system agnostic. They work on MacOS as well (I assume, because I have never tried it). I would also say that `pip` is the recommended installation, so it is ok that it is the first thing that you see. On the other hand, the other installation pages (including MacOS) are listed under the section "Other installation options" (maybe we should call it "Alternative installation ways", to make it more clear that they are optional?)

> 2. Linux - seems we can merge into one page with H2 (`##`) titles. It would be natural to have Mac, Windows, Linux (otherwise it's like apples + oranges).

It seems normal to me to keep them separate as they have different installation instructions (one with `deb` and one with `rpm`). The same thing is done on the big Download button on the front page (https://dvc.org/): we have "MacOS", "Windows", "Linux DEB" and "Linux RPM", we don't have a single "Linux" item.

Other tools have different installation pages for different Linux families, for example Docker: https://docs.docker.com/install/linux/docker-ce/debian/
But they have nested them under "Linux". Do you suggest that we nest them in another level? I don't think it is necessary, since we have only a few of them (three, including Arch).

> 3. please, use `dvc` syntax highlighter in any place when you have command line that starts with `$`

It does not seem to make any difference, since there are no `dvc` commands on those blocks. But this is something that can be fixed easily.

> 4. I would not repeat bash completion content on Mac in the Mac OS section. Why not ZSH? Why repeating instead of providing a link?

I thought that Bash is more commonly used than ZSH. Besides, Bash instructions are a bit different for MacOS and for Linux (while ZSH instructions are the same). Also I repeated them for convenience, so that all the steps for one system are in one page (convenience for the users, not convenience for the doc writer, because for the doc writter it actually is more work having to maintain both of the copies).
But I don't have a strong opinion on this. I can replace the repeated parts with a link.

> The bottom line - the structure is confusing for now.

Well, it _seems_ confusing to you, because obviously it does not seem confusing to me, since I have tried my best to make it clear. But I am willing to change it to make you happy, as long as you are consistent in your requests, it doesn't matter what i think.

> Let's think how can we make it better. Any good examples of tools?

I have already mentioned Docker: https://docs.docker.com/install/linux/docker-ce/debian/

By the way, I have the impression that most of the tools have the installation instructions on the top level of the doc structure. Let's think about this too. I think that it should be immediately after Get Started.

dashohoxha:
> Let's think how can we make it better.

I wish I could hide the entries for MacOS, Windows, etc. on the left sidebar. This might help avoid some confusion. However removing them from `sidebar.json` makes these paths inaccessible. Is there some way to keep the paths, but not show them on the left side?

Another option might be to include everything on the main page of installation, but make the sections for MacOS, Windows, etc. collapsed (similar to what is in the current page: https://dvc.org/doc/get-started/install).

Which one do you think would be better?
Or maybe we should try both (on separate PR's) and see which one looks better?

dashohoxha:
A third option would be to remove pip+conda instructions from the index page and include (repeat) them on each of the pages for MacOS, Windows, etc.

static/docs/user-guide/install/completion.md
```diff
@@ -0,0 +1,154 @@
+# Tab Completion
+
+Enjoy working with DVC faster and with fewer typos!
```
jorgeorpinel: Let's preserve the original name, ** Shell Autocomplete**. Or perhaps "Shell Auto Completion" for an improvement. These seem to be the most used terms out there.

Also, should remove the original doc, or they will be duplicated. Still here: https://dvc-org-pr-695.herokuapp.com/doc/user-guide/autocomplete

static/docs/user-guide/install/dev.md
```diff
@@ -0,0 +1,45 @@
+# Install for testing and development
+
```
jorgeorpinel: This is also duplicated in https://dvc-org-pr-695.herokuapp.com/doc/user-guide/contributing#development-environment, but with totally different methods...

static/docs/user-guide/install/dev.md
```diff
@@ -0,0 +1,45 @@
...
+```dvc
+$ virtualenv -p python3 .env
+$ source .env/bin/activate
+$ pip install gitpython
+$ pip install git+https://github.com/iterative/dvc.git
+```
+
+> **Note**! `gitpython` should be installed first to allow `setup.py` to
+> dynamically generate dvc version from the current git commit SHA. It allows us
```
jorgeorpinel: I'm not getting why recommend this `gitpython` method instead of the existing regular dev install method from https://dvc.org/doc/user-guide/contributing#development-environment.

jorgeorpinel: Also, please capitalize DVC or use backquotes (`dvc`) if referring to the command.
And please use _articles_: "**a** DVC version", "**the** official DVC release" 🙂

static/docs/user-guide/install/index.md
```diff
@@ -0,0 +1,62 @@
+# Install
+
+There are a few ways to install DVC: using `pip` or `conda`, which are system
+agnostic, and using packages/installers for specific systems.
```
jorgeorpinel: and using -> or using

jorgeorpinel: system -> platform (both occurrences)
I would even add a note at the end: "for specific platforms (Linux/Windows/Mac)." to be extra clear what we're referring to.

static/docs/user-guide/install/index.md
```diff
@@ -0,0 +1,62 @@
...
+
+> It may happen that you install DVC first for testing and then try to install
+> it again via a different way. This may cause any conflicts or confusion. So,
+> before trying to install it, make sure that it is not already installed, using
+> a command like: `which dvc`.
```
jorgeorpinel: Good note! But let's make it more compact? Something like:

Please double check you don't already have DVC before trying to install it (again), with a command like `which dvc`.

static/docs/user-guide/install/index.md
```diff
@@ -0,0 +1,62 @@
+# Install
+
```
jorgeorpinel: A very similar page exists in https://dvc-org-pr-695.herokuapp.com/doc/get-started/install which should probably be rewritten much shorter and just link to this guide for more details?

static/docs/user-guide/install/mac.md
```diff
@@ -0,0 +1,20 @@
...
+
+   ```dvc
+   $ brew install iterative/homebrew-dvc/dvc
+   ```
+
+   or:
+
+   ```dvc
+   $ brew cask install iterative/homebrew-dvc/dvc
+   ```
```
jorgeorpinel: Seems like we're deprecating this option per https://discordapp.com/channels/485586884165107732/565699007037571084/633702914103443458 (and https://github.com/iterative/homebrew-dvc/issues/16)

static/docs/user-guide/install/plugins.md
```diff
@@ -0,0 +1,30 @@
+# IDE Plugins & Syntax Highlighting
+
```
jorgeorpinel: Also exist duplicated in https://dvc-org-pr-695.herokuapp.com/doc/user-guide/plugins. Let's merge?

jorgeorpinel:
Seems like we'll want to stop recommending `pip` as a default installation method 😅 
Please review/participate on conversation in https://discordapp.com/channels/485586884165107732/565699007037571084/633698351883354124 to decide the corresponding changes to this PR. Thanks!

static/docs/user-guide/install/completion.md
dashohoxha: > Let's preserve the original name, ** Shell Autocomplete**. Or perhaps "Shell Auto Completion" for an improvement.

Or perhaps "Shell Tab Completion".

> should remove the original doc, or they will be duplicated

Let's postpone this for another PR that follows this one.

static/docs/user-guide/install/dev.md
dashohoxha: Maybe I should remove the installation for development and testing, to avoid duplication: https://dvc-org-pr-695.herokuapp.com/doc/user-guide/install/dev

static/docs/user-guide/install/dev.md
dashohoxha: > I'm not getting why recommend this gitpython method instead of the existing regular dev install method from https://dvc.org/doc/user-guide/contributing#development-environment.

I don't know, I tried to consolidate this page: https://dvc-org-pr-695.herokuapp.com/doc/user-guide/development
which is actually linked from here: https://dvc-org-pr-695.herokuapp.com/doc/get-started/install

static/docs/user-guide/install/index.md
dashohoxha: > A very similar page exists in https://dvc-org-pr-695.herokuapp.com/doc/get-started/install which should probably be rewritten much shorter and just link to this guide for more details?

Let's postpone this for another PR after this one.

static/docs/user-guide/install/plugins.md
dashohoxha: > Also exist duplicated in https://dvc-org-pr-695.herokuapp.com/doc/user-guide/plugins. Let's merge?

Next PR.

dashohoxha:
> Seems like we'll want to stop recommending pip as a default installation method ￼

I did not get this impression from the discussion. Could you please check again?

static/docs/user-guide/install/completion.md
jorgeorpinel: Or **Command-line completion**. Nah let's just get it done 🙂 @shcheklein which term do you vote for?

src/Documentation/sidebar.json
```diff
@@ -82,6 +82,26 @@
...
+        "source": "install/index.md",
+        "children": [
+          {
+            "label": "MacOS",
+            "slug": "macos"
+          },
+          "windows",
+          "debian",
+          "fedora",
+          "arch",
```
jorgeorpinel: These 3 may need labels: `Debian/ Ubuntu`, `Fedore/ CentOS`, `Arch Linux` to match https://dvc-org-pr-695.herokuapp.com/doc/user-guide/install.

jorgeorpinel: And/or actually may need to merge them per https://github.com/iterative/dvc.org/pull/695#pullrequestreview-301594486?

> Linux - seems we can merge into one page with H2 (##) titles...

static/docs/user-guide/install/completion.md
```diff
@@ -0,0 +1,154 @@
...
+one you are using (Bash or Zsh).
+
+Depending on the shell that you are using, follow the instruction below to
+install and configure shell completion.
+
+## Bash completion on Mac OS
+
+1. First, make sure that Bash completion support is installed:
+
+   ```shell
```
jorgeorpinel: `dvc`

static/docs/user-guide/install/completion.md
shcheklein: I would vote for something like https://medium.com/@dilankam/enable-auto-completion-of-git-commands-on-mac-os-mojave-10-14-2d692c100bc - `Enable Auto Completion` or ``Enable Autocomplete` (I'm not sure which one is a better form for a native speaker) - I think most users should understand this. We can also, mention on the index page and explain it using other words (like Tab Completion). @gagejustins what would be you take on this, I wonder? :)

static/docs/user-guide/install/plugins.md
```diff
@@ -0,0 +1,30 @@
...
+[DVC-file](/doc/user-guide/dvc-file-format) that contains all the needed
+information to track your data and transformations.
+
+The file itself is in a simple YAML format.
+
+## Vim
+
+In order to recognize DVC-files as YAML in Vim, you should add:
+
+```vim
```
jorgeorpinel: Maybe `usage`, `dvc`, or `yaml` (the 3 formats we have custom CSS for) will look better, please try.

static/docs/user-guide/install/completion.md
jorgeorpinel: I guess the currently use terms **Shell Completion** nav link and title **Shell Tab Completion* work fine.

static/docs/user-guide/install/dev.md
jorgeorpinel: Yes, thanks.

static/docs/user-guide/install/dev.md
jorgeorpinel: Woah! I never noticed that content was duplicated before (even when I've participated in the file's [history](https://github.com/iterative/dvc.org/commits/master/static/docs/user-guide/development.md) 😆 ! Thanks for pointing this out... I'm opening a separate issue to address this: #700

static/docs/user-guide/install/index.md
jorgeorpinel: I don't think we should merge a PR that duplicates a bunch of content. If you'd like you can remove all these semi-duplicated pages from the PR so we can get it done, and add them in a next one where the old ones can be deleted at the same time, then we'll just review that the migration is good 🙂

static/docs/user-guide/install/plugins.md
jorgeorpinel: Same as https://github.com/iterative/dvc.org/pull/695#discussion_r335282084

static/docs/user-guide/install/index.md
jorgeorpinel: @shcheklein seems to think this is OK though, per https://github.com/iterative/dvc.org/pull/695#pullrequestreview-302319707. I vote against it, but you're 
the merge master Ivan 😎  Your call.

jorgeorpinel:
p.s. UPDATE: If you open an immediate PR to address the duplicated docs, you could also address this last checkbox there.

static/docs/user-guide/install/completion.md
jorgeorpinel: Reopening due to Ivan's last comment, which I just noticed. Also curious about @gagejustins' input here.

static/docs/user-guide/install/index.md
dashohoxha: @jorgeorpinel Simplifying `/get-started/install` is not a simple task. After simplification you may need to merge it with `/get-started/initialize`. Then you have to change all the links that refer to `/get-started/install` to link to the new installation page. This may potentially cause any conflicts with any other PRs.
Let's keep PRs small, focused and quick, to avoid potential conflicts.

dashohoxha:
> 1. Merge all Linuxes. I really doubt that Arch deserves a full section on it's own.

What if we introduce another level for Linux distros? It seems like merging them would make the page a bit messy. In the future we may need to add different instructions for other distros (or maybe split Fedora from CentOS for some reasons). Let's be forward-looking and keep them in separate pages.

> 2. Mentions that it's a good practice to use some virt env to install pip packages.

This is needed when installing from the development code. But for normal installation it shouldn't be needed. It feels like installing Git for each repo. I think everyone agrees with this (except you).

static/docs/user-guide/install/plugins.md
dashohoxha: None of them looks better.

dashohoxha:
I introduced another level in the hierarchy of the menu. It seems cool to me.

static/docs/user-guide/install/index.md
jorgeorpinel: Well. I think you can basically remove the get-started/install and just add a link on the top of initialize, so maybe it's not that complicated. But as you guys choose.

Let's leave these unresolved though, and/or open a new issue detailing all these duplication problems and linking back to these discussions?

shcheklein: I think @dashohoxha already opened a new issue #701. May be it worth putting the link to this discussion in that ticket.

jorgeorpinel: Linked

jorgeorpinel:
> I introduced another level in the hierarchy of the menu.

image

Thanks for the initiative but I don't know... 5 clicks to navigate to them. I think already 3 levels is a little too much some times.

jorgeorpinel:
> In the future we may need to add different instructions for other distros...

This can't be done with H2s?

shcheklein:
@dashohoxha 

> In the future we may need to add different instructions for other distros (or maybe split Fedora from CentOS for some reasons). Let's be forward-looking and keep them in separate pages.

Would really really try to avoid this. Linux - we should aim for simplicity and unification. We def don't want to have a page per linux distro. If it's so specific that deserves a note - make an expandable section of H2. Let's do a single page. Too much overlap, too low unique information for distros like Arch - I don't think Arch even deserves it's own H2 to be honest.

> re Extras

I think it's getting out hands and we are complicating things too much (4 levels? - too many).
I think we can keep Shell Autocompletion and Plugins & Intergations as part of the Installation guide.

> This is needed when installing from the development code. But for normal installation it shouldn't be needed. It feels like installing Git for each repo. I think everyone agrees with this (except you).

No. It's _always_, _always_ better to not touch you system's Python packages. Especially on Mac. https://opensource.com/article/19/4/managing-python-packages and can share a lot of other materials.

> Development version (user focused one)

Sure, let's do as part of other PR, but I would really try to merge it into the Install guide. The easiest way to do this now is to just move first under Install.

shcheklein:
@jorgeorpinel 

adding updating notes sounds cool, but probably should be done separately

jorgeorpinel:
> 2. Mentions that it's a good practice to use some virt env to install pip packages.

On this also I don't think the note is necessary, but for a different reason. I think we should avoid teaching other technologies in our docs that aren't DVC. Users should be responsible for  knowing their `git`, `pip`, etc. best practices. (Similar argument as when I removed the Prettier script file.)

Then again if it's a short one-line note (linking to dev install doc) it probably doesn't hurt either. So I'm neutral on this.

static/docs/user-guide/install/extra/code.md
```diff
@@ -0,0 +1,51 @@
+# Install from code
```
shcheklein: let's just move the Development section as is for now. No need to move here parts of the Contributing Guide. They serve two very different purposes.

static/docs/user-guide/install/plugins.md
jorgeorpinel: Linke to #701 desc.

shcheklein:
@jorgeorpinel 

> On this also I don't think the note is necessary, but for a different reason. I think we should avoid teaching other technologies in our docs. Users should be responsible for knowing their git, pip, etc. usage and best practices. (Similar argument as when I removed the Prettier script file.)

Level of mistake is way higher in this case and it's extremely hard to fix it.

> Then again if it's a short one-line note (linking to dev install doc) it probably doesn't hurt either. So I'm neutral on this.

It's just a simple note indeed (not going into specifics how to install and/or use virtualenv/pyenv). Not sure how is it related to dev install docs.

jorgeorpinel:
> Not sure how is it related to dev install docs.

@shcheklein I meant the link can go to https://dvc.org/doc/user-guide/contributing#development-environment which has more detail an sample commands to use `virtualenv`.

shcheklein:
>  I meant the note can link to https://dvc.org/doc/user-guide/contributing#development-environment which has more detail and sample commands to use virtualenv.

yep, I got the idea. But I think it can be confusing because that page is made with a very specific goal in mind (not educate about virtualenv). Though we can point to it with some explanation. Alternatively, we can point to some other docs that have virtualenv setup.

jorgeorpinel:
p.s. I pushed a commit with copy edits instead of leaving a more detailed review. Please pull branch and feel fee to review them.

dashohoxha:
> I pushed a commit with copy edits

Thanks @jorgeorpinel.
I pushed a commit that fixed the structure.
If you don't mind, could you also add the notes about the virtual env etc. because you can always formulate them better.

static/docs/user-guide/install/extra/code.md
jorgeorpinel: I addressed this in 1774355e.

shcheklein:
@jorgeorpinel @dashohoxha so, here is one good example of what I'm referring to https://www.tensorflow.org/install/pip . We don't have to go that lengthy at this stage - we just don't resources for this. But being a bit mindful about regular user problems is the way we write our docs.

For us I would just a put a note with `>` that we recommend to use virtualenv (or other env isolation tool), also if you don't use virtualenv.

dashohoxha:
@jorgeorpinel I reverted your last commit because it contains changes that are unrelated to this PR. Thanks for your understanding.

> I'm not sure I understand exactly what Ivan envisions. You may have to work it out with him.

That's the difficult part, because I am not able to understand Ivan either.

shcheklein:
@dashohoxha have you had time to take a look at the example I shared? A simplified version of it is exactly thing I meant.

static/docs/user-guide/install/completion.md
```diff
@@ -0,0 +1,154 @@
...
+   if [ -f $(brew --prefix)/etc/bash_completion ]; then
+       . $(brew --prefix)/etc/bash_completion
+   fi
+   ```
+
+4. Finally, `source ~/.bash_profile` or launch a new terminal to activate it.
+
+## Bash completion on Ubuntu
+
+On a Linux system bash-completion should work out of the box, but if you are
```
shcheklein: is it true, @efiop ? or is it only if you install it as a binary?

static/docs/user-guide/install/completion.md
```diff
@@ -0,0 +1,154 @@
...
+Then reload your shell:
+
+```dvc
+$ exec $SHELL -l
+```
+
+This step is optional but will make the DVC output look much nicer, by adding
+more colors to it. Add the following to your `~/.zshrc`:
+
+```bash
```
shcheklein: is there zsh?

static/docs/user-guide/install/linux.md
```diff
@@ -0,0 +1,104 @@
...
+
+> Requires [Anaconda Distribution](https://docs.anaconda.com/anaconda/).
+
+```dvc
+$ conda install -c conda-forge dvc
+```
+
+> Currently, it supports only Python versions 2.7, 3.6, and 3.7.
+
+## Install from repo
```
shcheklein: repo -> repository? I'm not sure what is the reason to cut it here. If take a look at Ubuntu's docs they don't use "repo" term as far as I can tell.

static/docs/user-guide/install/linux.md
```diff
@@ -0,0 +1,104 @@
...
+
+### On Debian/Ubuntu
+
+```dvc
+$ sudo wget \
+       https://dvc.org/deb/dvc.list \
+       -O /etc/apt/sources.list.d/dvc.list
+
+$ sudo apt update
+
```
shcheklein: should we remove these extra spaces? we don't put them in other places.

static/docs/user-guide/install/linux.md
dashohoxha: This is how it is called in the url: `https://dvc.org/rpm/dvc.repo`. Anyway, fixed it.

static/docs/user-guide/install/completion.md
dashohoxha: There is, but the result is the same. Anyway fixed it.

static/docs/user-guide/install/linux.md
dashohoxha: It seemed toe me that it makes the content more readable. But I don't have a string opinion on this. Fixed it.

static/docs/user-guide/install/code.md
```diff
@@ -0,0 +1,51 @@
+# Install from code
```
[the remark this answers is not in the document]
dashohoxha: This is actually the Development section moved here and modified.
I am removing the last part that clones the code and installs from it, since this seems related to development, not related to usage.

I am also renaming it to "Install latest version".
A name that includes "development" would still give the wrong impression that this is for developers.

dashohoxha:
> have you had time to take a look at the example I shared? A simplified version of it is exactly thing I meant.

That example does not explain anything to me. First of all DVC is a CLI tool, not a ML library, why should we take  tensorflow as an example? Anyway, the installation page is this one: https://www.tensorflow.org/install and it mentions only `pip install tensorflow`, nothing else. I think that we should not tell users how to install `pip`.

> Also, we already have a phrase:
>
> In this case it is highly recommended to
use [virtualenv](https://virtualenv.pypa.io/en/stable/), in order to encapsulate
your testing environment. 
>
> this is along the lines I was talking about.

Yes, we already have it on the page that shows how to install and test the latest version from the code on GitHub. So, the problem is already solved, we don't need to do anything about it.

>  current version of "Install from code" is very confusing - name (we can discuss this in details if needed), mixing different audiences - users and developers, which is really bad idea.

> The only thing that matters in this section more or less - how do I install the "preview", "development", "alpha", "nightly build" version if I want to be early adopter of it. And we should name the section that is very clear that it's about that thing, content should be simple and serve one purpose (not from local code, for example, if we really need this - just put a link).

I have already made some changes. Let's discuss and improve further if needed.

static/docs/user-guide/install/latest.md
```diff
@@ -0,0 +1,35 @@
+# Install latest version
```
shcheklein: the latest version is usually means the latest releases version. It's still a confusing name. Why don't you like Development version?

dashohoxha: Fixed.

> Why don't you like Development version?

No particular reason, just seemed confusing.

static/docs/user-guide/install/completion.md
efiop: @shcheklein only if you install from our deb package.

efiop: Also not sure what "VPS" has to do with anything here. @dashohoxha Could you clarify?

static/docs/user-guide/install/completion.md
jorgeorpinel: The generic one is `sh`

static/docs/user-guide/install/linux.md
jorgeorpinel: I had addressed this in 1774355.

static/docs/user-guide/install/completion.md
dashohoxha: > Also not sure what "VPS" has to do with anything here. @dashohoxha Could you clarify?

The Ubuntu server image in some cloud providers is minimized (to save time in deployment), and the package `bash-completion` is missing. More details here: https://linuxhandbook.com/enable-tab-completion/

shcheklein: @dashohoxha so, should we update to clarify that it's not out of the box in certain cases? And I still don't understand what VPS has do with it.

dashohoxha: > should we update to clarify that it's not out of the box in certain cases

I think the explanation is correct because it says "In any case, follow these steps to make sure that it is installed". So, if it works already that's fine, if not follow these steps to make sure that everything is setup correctly.

> I still don't understand what VPS has do with it.

If you have installed DVC in a VPS on the cloud, the package `bash-completion` probably is not installed, and without it tab completion cannot work. This is one of the cases when bash completion does not out of the box.

Maybe we should not mention it at all and just say: "In case that bash completion does not work, follow these steps...". What is your take on it?

shcheklein: Explanation is might be correct, but the statement `On a Linux system bash-completion should work out of the box` is not. It's just false and misleading.

I get the cloud part, I don't understand what _VPS_ has to do with it. Could you elaborate? Give an example?

static/docs/user-guide/install/completion.md
```diff
@@ -65,10 +65,6 @@ install and configure shell completion.
 
 ## Bash completion on Ubuntu
 
-On a Linux system bash-completion should work out of the box, but if you are
```
shcheklein: not kinda the way to address this I expected :) Why can't we mention that it should work out of the box if it was installed from the package?

static/docs/user-guide/install/completion.md
dashohoxha: > I get the cloud part, I don't understand what VPS has to do with it. Could you elaborate? Give an example?

I removed the paragraph that mentions VPS because it seems superfluous. Anyway, if you are confused by it, I can imagine that a normal user would be much more confused. I hope that now it is simpler and more logical.

shcheklein: Yep, thanks. I would say you removed too much. I would keep a mention that it should be working out of the box on Linux if it's installed from the package (manually or from the custom repository).

static/docs/user-guide/install/completion.md
dashohoxha: > Why can't we mention that it should work out of the box if it was installed from the package?

If it works out of the box, these instructions still don't harm anything, because they just say "make sure that this is so and so". So, everything is still ok,

static/docs/user-guide/install/completion.md
dashohoxha: >  I would keep a mention that it should be working out of the box on Linux if it's installed from the package (manually or from the custom repository).

Installing from the package still does not guarantee that bash completion should work out of the box. If you install it on a VPS on the cloud this would not be enough.

static/docs/user-guide/install/completion.md
shcheklein: > because they just say, make sure that this is so and so

they (almost all of them) do not, they specifically mention to do a step a step installation. It can save my time by not even reading the first step and not running the command. You went from one extreme to another for a reason I don't quite understand to be honest.

dashohoxha: >  for a reason I don't quite understand to be honest.

Installing from the package still does not guarantee that bash completion should work out of the box. If you install it on a VPS on the cloud this would not be enough.

shcheklein:
@dashohoxha 

> First of all DVC is a CLI tool, not a ML library

it does not matter. The problem is the same. You have to use `sudo` and you are breaking your system potentially.

> https://www.tensorflow.org/install

It's the analog of the index page or get started install we have. The page I sent is the analog of the detailed installation instruction.

> it mentions only pip install tensorflow, nothing else

this is not true - it mentions three commands, including upgrading pip? Why? For the user convenience. Probably they have see that the failure rate is high if don't upgrade it.

> Yes, we already have it on the page that shows how to install and test the latest version from the code on GitHub. So, the problem is already solved, we don't need to do anything about it.

Don't see any connection here to be honest? How does it "solve" the problem? I also don't see any difference - if we mention that you should isolate dev version, for the same reasons it's better to isolate the stable one. Also, the way dev version is written is still wrong and should be simplified. I would keep only a mention about virtualenv/pyenv - a recommendation.

static/docs/user-guide/install/completion.md
shcheklein: thus `_should_ be working`, not `must` - it's enough to try it before jumping into the instruction.

I still don't understand why do you use a very specific term `VPS`. I can imagine that `bash-completion` is not installed by default on a variety of Linuxes. Or am I missing something? 

Btw, @efiop, can we install (and should?) the necessary deb packages automatically to make it work? Also, will it start working automatically if I install the necessary deb packages after I installed DVC deb?

dashohoxha:
>  Also, the way dev version is written is still wrong and should be simplified. I would keep only a mention about virtualenv/pyenv - a recommendation.

Sure, I can do this.

> if we mention that you should isolate dev version, for the same reasons it's better to isolate the stable one

Sorry, I don't understand this, and I cannot write about things that I am not convinced.
Let's merge this PR and then someone else who understands it better than me can fix/improve this part. I have my own limitations. Thanks for your understanding.

static/docs/user-guide/install/linux.md
```diff
@@ -0,0 +1,100 @@
...
+$ pip install 'dvc[s3]'
+```
+
+In this case it installs `boto3` library as well, besides DVC.
+
+</details>
+
+## Install with conda
+
+> Requires [Anaconda Distribution](https://docs.anaconda.com/anaconda/).
```
shcheklein: would be great to mention that it requires Miniconda (the package manager) https://docs.conda.io/projects/continuumio-conda/en/latest/user-guide/install/macos.html (find the right link per platform) or Anaconda. Anaconda is scary to be honest.

shcheklein:
@dashohoxha I think it also would be beneficial to actually recommend conda before pip for all systems. Pip is the most problematic one (unless you are installing a library and is using virtualenv). Similar to what PyTorch is doing on the installation page.

static/docs/user-guide/install/completion.md
dashohoxha: > I still don't understand why do you use a very specific term VPS. I can imagine that bash-completion is not installed by default on a variety of Linuxes. Or am I missing something?

It **is** installed by default on almost all versions of Linux. Surprisingly it is not installed on ubuntu servers  offered on some clouds.

>  will it start working automatically if I install the necessary deb packages after I installed DVC deb?

There is also the part of enabling it on `~/.bashrc`. Installing `bash-completion` does not guarantee that your `~/.bashrc` will have the right configuration.

shcheklein:
> Sorry, I don't understand this, and I cannot write about things that I am not convinced.
Let's merge this PR and then someone else who understands it better than me can fix/improve this part. I have my own limitations. Thanks for your understanding.

Sure! Absolutely.

static/docs/user-guide/install/linux.md
dashohoxha: > would be great to mention that it requires Miniconda (the package manager)

Maybe someone else knows better than me about anaconda/minicoda. I have never used them and whatever I recommend I doubt that might be wrong.

static/docs/user-guide/install/dev.md
```diff
@@ -2,14 +2,13 @@
 
 If you want to test the latest version of DVC, before the official release, you
 can install it from the code on GitHub. In this case it is highly recommended to
-use [virtualenv](https://virtualenv.pypa.io/en/stable/), in order to encapsulate
-your testing environment.
+use [virtualenv](https://virtualenv.pypa.io/en/stable/) in order to encapsulate
+your testing environment, otherwise it will automatically upgrade your system's
```
shcheklein: it depends on how DVC is installed, so this is not correct. Also, the whole point of using the preview version usually is to use it as a replacement, not along with an old one. It's the case when early adopters what the most recent shiny features.

static/docs/user-guide/install/linux.md
shcheklein: Sure! I'll take care of this when it's merged.

static/docs/user-guide/install/dev.md
```diff
@@ -19,7 +14,9 @@ $ pip install git+https://github.com/iterative/dvc.git
 > (e.g. `0.24.3-9c7381`). For more information on naming convention, refer to
 > [dvc version](/doc/command-reference/version).
 
-## Install branch from GitHub
+Note that this will automatically upgrade your system's DVC version to the
```
shcheklein: this is not true in many many case when DVC was not installed via pip.

static/docs/user-guide/install/completion.md
gagejustins: Hey all, sorry for the delay - was at our company offsite. What we're talking about here is traditionally referred to as some variation of "auto complete" - using the "tab" key is often the *way* that you do it in an editor, but it's not part of the name

static/docs/user-guide/install/completion.md
efiop: @dashohoxha 

> It is installed by default on almost all versions of Linux. Surprisingly it is not installed on ubuntu servers offered on some clouds.

I suppose you are talking about bash-completion package, not dvc-specific completion. That  makes sense.

> Btw, @efiop, can we install (and should?) the necessary deb packages automatically to make it work? Also, will it start working automatically if I install the necessary deb packages after I installed DVC deb?

Yes, we could totally specify `bash-completion` package as a dependency. Probably as simple as `--deb-pre-depends bash-completion`. But I'm not sure if it is a common practice to include that as a mandatory dependency, since bash completion is optional. Need to do the research there.

static/docs/user-guide/install/dev.md
shcheklein: @dashohoxha are you going to address this? I'm not merging other PRs until we merge this one to avoid the conflicts, but we can't wait too long. Let me know if you need our help with this.

shcheklein:
@jorgeorpinel @dashohoxha btw, this https://github.com/iterative/dvc/issues/2582#issuecomment-544589995 is exactly the issue with installing DVC with a system wide pip - it's a never ending story of getting ERRORs on dependencies. It's almost impossible to fix, recover, etc. Thus I _always_ recommend to _never_ _ever_ install _anything_ system wide. Either use brew or conda. In some rare cases if you know what are doing I would install aws cli, or virtualenv itself globally. Or, if it's a remote machine to run something before you shut it down pip can be an option .

static/docs/user-guide/install/dev.md
dashohoxha: > this is not true in many many case when DVC was not installed via pip.

@shcheklein It is a verbatim copy of this: https://dvc.org/doc/user-guide/development

> Let me know if you need our help with this.

I'd prefer if someone else rewords it, because I am afraid that no matter what I do, it will never be good enough.

> I'm not merging other PRs until we merge this one to avoid the conflicts

In this case there are not likely to be any conflicts, since this PR only adds new pages, without touching any existing ones.

shcheklein: @dashohoxha 

> @shcheklein It is a verbatim copy of this: https://dvc.org/doc/user-guide/development

kk. I think that document is just outdated a bit (or wrong that happens a lot).

> I'd prefer if someone else rewords it

kk

>

putting this PR on hold until I have time to address the comments that left. Merging everything other PRs that are ready.

dashohoxha: > this is not true in many many case when DVC was not installed via pip

I have tried to fix it. I am not sure this is best way to fix it, but maybe it is acceptable.

PR #704, opened 2019-10-17
Darken the color in tooltips

algomaster99:
Screenshot from 2019-10-17 23-32-24

Hovering over keywords inside blockquotes took their text colour. These changes are introduced to fix that.


shcheklein:
@jorgeorpinel yeah, I think the change itself make sense no matter where we decide to use tips.

PR #715, opened 2019-10-20
Wrap strings at 80 characters

### LINKED ISSUE #548: lint: wrap js strings (in src/)
UPDATE: See https://github.com/iterative/dvc.org/issues/548#issuecomment-544209275.

---

OUTDATED:

String values in JavaScript files are currently not automatically wrapped to 80 chars by the pre-commit hook that runs `pretty-quick`, like all other lines of code. If possible, have the linter do this. E.g.

```es6
const strVar  = "Really long string going over 80 characters. Blah blah blah blah blah blah blah"
let templateLit = `
Another long text
Including lines that go over 80 chars. Blah blah blah blah blah blah blah blah blah blah blah blah
`
```

This would be pretty useful to make sure the `glossary.js` strings are properly wrapped, for example.

algomaster99:
Fixes #548 

1. For `styled` blocks, I have disabled eslint max-len checking because `eslint` was automatically converting
```
@media only screen and
(min-device-width: 768px) and
(max-device-width: 1024px) {
    padding: 0px 61px 0px 67px;
    max-width: auto;
  }
```
to
```
 @media only screen and (min-device-width: 768px) and (max-device-width: 1024px) {
    padding: 0px 61px 0px 67px;
    max-width: auto;
  }
```
2. For `svg` tag, I thought to let `d` attributes as it is.
3. In `helper.test.js`, eslint was not allowing to break parametes into multiple line.


shcheklein:
> For styled blocks, I have disabled eslint max-len checking because eslint was automatically converting

so, I can write a styled component 100 symbols long (I think it's not good). Is there an option for `eslint` to respect the existing formatting (it's it's below 80).

> For svg tag, I thought to let d attributes as it is.

👍 

> In helper.test.js, eslint was not allowing to break parametes into multiple line.

could you elaborate, please? give more details?

algomaster99:
@shcheklein `ESLint` automatically formatted
```
  it(
      "Throws error if item has source: false and doesn't have children",
      () => {
        const rawData = [{ slug: 'item-name', source: false }]

        jest.doMock('../sidebar.json', () => rawData)

        expect(() => require('./helper')).toThrow(
          new Error(
            "If you set 'source' to false, you had to add at least one child"
          )
        )
      }
    )
  })
```
to this
```
it("Throws error if item has source: false and doesn't have children", () => {
      const rawData = [{ slug: 'item-name', source: false }]

      jest.doMock('../sidebar.json', () => rawData)

      expect(() => require('./helper')).toThrow(
        new Error(
          "If you set 'source' to false, you had to add at least one child"
        )
      )
    })
  })
```
The latter exceeds the 80 character limit where `it` is defined.

jorgeorpinel:
Hm... The longest line there,

`it("Throws error if item has source: false and doesn't have children", () => {`

is exactly 80 chars long.

src/Diagram/index.js
```diff
@@ -160,6 +160,7 @@ export class DiagramSection extends Component {
 
 export default DiagramSection
 
+/* eslint-disable max-len */
 const Diagram = styled.section`
   padding-top: 80px;
```
jorgeorpinel: I also don't agree this is necessary. Is there a problem with how ESLint formats this? I see the `@media...` line 210 below is already in a single line anyway.

jorgeorpinel: ...which BTW goes over 80 chars, which is the problem `eslint` reported for this file. See https://github.com/iterative/dvc.org/issues/548#issuecomment-544209275

jorgeorpinel: Same for the other 3 files where this is done.

algomaster99: @jorgeorpinel So should I let it stay above 80 characters and remove the disabling comments or shall I stick with disabling linting on media queries?

jorgeorpinel: I don't understand. If you disable it, then the linter won't break the line. Simply do not disable the linter and let it break the line?

algomaster99: @jorgeorpinel linter wasn't breaking the line and was throwing an error. But when I manually fixed it, it reverted the changes forcefully and threw a `max-len` error.

jorgeorpinel: The linter only examines Git changes, not all the files every time. You would have to run it manually like `yarn prettier --write '{static,src}/**/*.{js,jsx,md}`. (Mentioned in https://dvc.org/doc/user-guide/contributing-docs#doc-style-guidelines-and-tips-for-java-script-and-markdown)

jorgeorpinel: So can you just break those lines manually and remove all these `/* eslint-disable max-len */`? 🙂

algomaster99: @jorgeorpinel I did break those lines. But when I ran the linter, it reverted my changes and put everything in one line.
I broke the media queries and the tests function.

jorgeorpinel: Ohhhh! This is not CSS, it's a template literal in JavaScript... But still I didn't know Prettier would elongate some lines. That's not nice! Seems like it's a known problem though... Check prettier/prettier/issues/3368, prettier/prettier/issues/6192, and prettier/prettier/issues/6262 for possible workarounds.

Try https://github.com/sgtpep/prettier-plugin-html-template-literals...

Not sure what to do otherwise, I guess just let those line be long. We definitely don't want `eslint-disable max-len`.

algomaster99: @jorgeorpinel 
https://github.com/prettier/prettier/pull/5979
I think this PR is forcing prettier to make a template literal into one line :man_facepalming: 
Check out this [changelog](https://github.com/prettier/prettier/pull/5979/files#diff-d310d8d326c0b4526a13ac21c13110a2).
[This guy's](https://github.com/prettier/prettier/issues/6192#issuecomment-525723143) current workaround is using `// prettier-ignore`

shcheklein:
@algomaster99 any progress on this? do you need any help?

algomaster99:
@shcheklein https://github.com/iterative/dvc.org/pull/715#discussion_r339018414 see this

jorgeorpinel:
To summarize, the conversation in https://github.com/iterative/dvc.org/pull/715#pullrequestreview-304672758 (now resolved), Prettier has a weird behavior (introduced in prettier/prettier/pull/5979) that elongates some lines inside js template literals during formatting. But then hen ESLint reports an error in those lines...

1. @algomaster99 maybe report that bug on the prettier repo?
1. Possible workarounds are described in prettier/prettier/issues/3368, prettier/prettier/issues/6192, and prettier/prettier/issues/6262 - have we tried them?
1. Aman is suggesting we use special comments for Prettier or ESLint to ignore those blocks.

I think that we should not include special ignore comments because we'll forget to remove them later. Unless, perhaps, they only affect that one specific line and we keep track of this bug on Prettier. However we may be forced to do this unless there is some way to force merging this without CI checks having passed. Is this possible @shcheklein?

src/Diagram/index.js
jorgeorpinel: OK just remove the eslint ignore comments for now please, and merge master so this PR is mergeable (except for the CI lint test which will not pass).

shcheklein:
> @algomaster99 maybe report that bug on the Prettier repo?

+1

> I think that we should not include special ignore comments because we'll forget to remove them later.

+1

> CI checks having passed. Is this possible @shcheklein?

it's technically possible, but we are not going to do this since it will affect all other PRs after that - and the general status of the repo (README.md badge)

jorgeorpinel:
Do you see any other alternatives/ workarounds @algomaster99? Not requiring `prettier-ignore` comments.

algomaster99:
@jorgeorpinel 

> Hm... The longest line there

Screenshot from 2019-10-26 19-38-42
Screenshot from 2019-10-26 19-39-32
It is actually indented.

> Do you see any other alternatives/ workarounds

No, not yet. Thinking to open issues in prettier.

algomaster99:
@jorgeorpinel 
> * Possible workarounds are described in [prettier/prettier/issues/3368](https://github.com/prettier/prettier/issues/3368), [prettier/prettier/issues/6192](https://github.com/prettier/prettier/issues/6192), and [prettier/prettier/issues/6262]
(https://github.com/prettier/prettier/issues/6262) - have we tried them?

See https://github.com/prettier/prettier/issues/6262#issuecomment-507533012.
Anyway, I don't think so there other workarounds than `prettier-ignore` :confused:

algomaster99:
> No, not yet. Thinking to open issues in prettier. Here it is: [prettier/prettier#6716](https://github.com/prettier/prettier/issues/6716).

@jorgeorpinel @shcheklein They have replied. I say we go with `eslint-disable-next-line`. What do you suggest?

shcheklein:
@algomaster99 

So, first of all we have two different problems, right?

> They have replied. I say we go with eslint-disable-next-line. What do you suggest?

Have you tried to google if there is a solution on the eslint side? It might be that we will have to got with that eslint-disable-next-line clause. Or alternatively just cut the description length.

Have we decided what do we do with styled blocks? And what the problem with them in the first place?

shcheklein:
@algomaster99 please, check also the latest reply in the issue you created in the Prettier repo. There is an option to use a common eslint-prrettier config. I haven't check which of our problems would is solve, which would it ignore, etc.

Btw, eslint only checks the correctness it doesn't modify code, right? Prettier modifies it.

algomaster99:
@shcheklein 
> So, first of all we have two different problems, right?

Yes.
1.  https://github.com/iterative/dvc.org/pull/715#issuecomment-546606946 the string after `it` is automatically formatted by `prettier` to be in one line with the argument `() => {}`
2. The media queries is styled divs are forcefully formatted by prettier to be in one line. For which I suggested `max-len: ["error", {"code": 80, "ignoreTemplateLiterals": true}]`. This will override can help with overriding prettier's behavior in a cleaner way. See this - https://eslint.org/docs/rules/max-len#options.

> Have we decided what do we do with styled blocks?

What do you think of point 2 above?

> Have you tried to google if there is a solution on the eslint side?

There was one option available for `max-len` called [`ignoreStrings`](https://eslint.org/docs/rules/max-len#options) but it can make the code uglier in few other places where strings could be formatted to 80 characters.

algomaster99:
> Btw, eslint only checks the correctness it doesn't modify code, right?

Eslint is a code linter while prettier is a formatter. Eslint understands our code and throws errors which are more based on the syntax like Eslint will throw errors/warnings if there are unused imports, etc. But to answer your question, they both can modify it.
1. Eslint needs a flag `--fix` for modifying it.
2. Prettier doesn't need any.

shcheklein:
would be great to see if there is a global `eslint` option to ignore  `it(...)` lines. Otherwise, let's go with the `eslint-disable-next-line` flag. It's fine if we don't have to cover a lot of tests with it and there are no other options.

> What do you think of point 2 above?

I'm not sure I understand how eslint's option helps overriding prettier's behavior.

> Eslint needs a flag --fix for modifying it.

do we use it?

algomaster99:
> do we use it?

No, pre-commit hook runs `yarn lint` which only checks it. There is no `--fix` flag.

I will update you when I progress on this issue. Will try turning off all the conflicting rules.

shcheklein:
> There is no --fix flag.

`because eslint was automatically converting` - is it a typo, then?

`This will override can help with overriding prettier's behavior` - still don't quite understand how then eslint's option can overrider prettier's behavior

> Will try turning off all the conflicting rules.

sounds good ... I would focus though on the problems we have and cherry-pick relevant rules if possible

algomaster99:
@shcheklein 
> because eslint was automatically converting

When I run `yarn lint`, it only shows the errors and warnings. How come is it fixing errors for you?
Screenshot from 2019-10-27 12-04-26

shcheklein:
`because eslint was automatically converting` - this you phrase, not mine, see here https://github.com/iterative/dvc.org/pull/715#issue-330162018

algomaster99:
@shcheklein sorry, my mistake. It was `prettier` not `ESLint`. :sweat_smile:

algomaster99:
@shcheklein 
> still don't quite understand how then eslint's option can override prettier's behavior

So there are two packages installed in our project called - `eslint-config-prettier` and `eslint-plugin-prettier`.
1. `eslint-config-prettier` - Its purpose is to switch off all rules which may conflict with prettier configuration. https://github.com/prettier/eslint-config-prettier#special-rules
2. `eslint-plugin-prettier` - Whenever you run `eslint`, this plugin makes sure it will run prettier too. For example, let's say we leave out a semi-colon in our code. Running eslint will return the following log.
Screenshot from 2019-10-27 12-40-30

Now to include these two configurations, we simply need to add one line in `.eslintrc.json`
```json
{
  "extends": [
    "plugin:prettier/recommended"
  ]
}
```

Now getting back to the problem, [`max-len`](https://github.com/prettier/eslint-config-prettier#max-len) was already switched off because it knew conflicts due to `styled-div` and description of test case will arise. But there were some lines which could be formatted to 80 characters and ESlint was ignoring that too. For example, https://github.com/iterative/dvc.org/pull/715/files#diff-2d4c3e3ce4857c5300e16a64af06b639. This is why we needed to enable `max-len`.

algomaster99:
@shcheklein 
So what do you suggest for the [two problems above](https://github.com/iterative/dvc.org/pull/715#issuecomment-546642569)?.

shcheklein:
> we simply need to add one line in .eslintrc.json

it's already here, right?

> So what do you suggest for the two problems above?

@algomaster99 as I mentioned your plan `I will update you when I progress on this issue. Will try turning off all the conflicting rules.` looks good to me.

algomaster99:
@shcheklein the rule are already off since we have already written the line `plugin:prettier/recommended`. 

We just turned on one - `max-len`.

> Now getting back to the problem, [`max-len`](https://github.com/prettier/eslint-config-prettier#max-len) was already switched off because it knew conflicts due to `styled-div` and description of test case will arise. But there were some lines which could be formatted to 80 characters and ESlint was ignoring that too. For example, https://github.com/iterative/dvc.org/pull/715/files#diff-2d4c3e3ce4857c5300e16a64af06b639. This is why we needed to enable `max-len`.

shcheklein:
@algomaster99 I meant this line: `"plugin:prettier/recommended"` is already there? Was you referring to it in your `we simply need to add one line in .eslintrc.json`?

shcheklein:
so, do we need to do more changes? 

what about "/* eslint-disable max-len */"? as we discussed we don't consider this as a solution, right? Could you once again summarize the problems and solutions you want to apply to each of them and describe what is still missing?

algomaster99:
@shcheklein 

### Problem 1
`prettier` won't **allow** us to break the media queries in styled-components. https://github.com/prettier/prettier/pull/5979
Screenshot from 2019-10-28 23-21-00
_The small vertical line on the left indicates the 80 character limit_

#### Proposed solution
1. Change `"max-len": ["error", { "ignoreStrings": false }],` to `"max-len": ["error", { "ignoreStrings": false, "ignoreTemplateLiterals": true }],`. ESLint simply ignore **all template literals** in our codebase and let prettier format it the way it they do. Since `prettier` is an _opinionated code formatter_ we can't do anything about it and they don't have an option for it too. Using this method will **have** the media query line **above 80** since `prettier` would have formatted the code before `eslint` is run according to our pre-commit hook configuration.
2. Use `/* eslint-disable max-len */` and `/* eslint-enable max-len */`. This will format the code using `prettier` and when `eslint` is run, it will simply ignore the block enclosed between these comments. This method is currently used and it **enforces** us to have the media query line length **above 80** since `prettier` formats it and then `eslint` ignores it.
3. Use `// prettier-ignore` just above those styled-components and then break it manually.
Screenshot from 2019-10-28 23-34-39
Screenshot from 2019-10-28 23-34-48. 
Doing this wouldn't throw any errors when running `yarn lint`. Using this method will **wrap** the media query line since we did it manually and `// prettier-ignore` will simply ignore formatting this block.

P.S. I also checked if any rule of `stylelint` could fix this. I couldn't find any.

### Problem 2
`prettier` **won't allow** us to break the arguments of a test's function. https://github.com/prettier/prettier/pull/953 https://github.com/prettier/prettier/issues/6716
Screenshot from 2019-10-26 19-38-42
_The small vertical line on the left indicates the 80 character limit_

#### Proposed solution
1. Use `// prettier-ignore` and format it manually. The will wrap it and `eslint` won't throw any `max-len` errors.
2. Use `// eslint-disable-next-line max-len`. This will simple accept `prettier`'s formatting and ignore the length of line when `eslint` is run..

shcheklein:
@algomaster99 thanks for the great summary!

I vote for the solutions 1 and 2 for problems 1 and 2 respectively. This way we always accept the way prettier formats the code and we adjust ESLint to be compatible with it. It's also aligned with the way we already do this by applying `"plugin:prettier/recommended"`.

algomaster99:
> It's also aligned with the way we already do this by applying `"plugin:prettier/recommended"`

So that running `eslint` will show errors reported by `prettier`?

shcheklein:
> So that running eslint will show errors reported by prettier?

To be honest I'm not sure how does exactly `plugin:prettier/recommended` affect the behavior of these two. My point was that I would expect that at least it applies some set of config options to ESLint to make it compatible with what prettier expects/complaints about. Similar to applying `"ignoreTemplateLiterals": true ` from the above manually, but `prettier/recommended` is doing this automatically. May be I'm wrong on this.

Since it's already enabled (I mean `plugin:prettier/recommended`) may be it would be great to understand what does it change. Does it have any effect on the CI runs. Etc.

jorgeorpinel:
> Will try turning off all the conflicting rules.

Did this not fix **probelm 1**? I also think this is the best strategy. See [recommended ESLint config](https://prettier.io/docs/en/integrating-with-linters.html#eslint). Or are you saying their recommended config doesn't already disable `ignoreTemplateLiterals` @algomaster99?

I also vote for solution #1 here since it doesn't require maintaining several `/* eslint-disable max-len */` and `/* eslint-enable max-len */` comments.

---

For **problem 2**, could the assertion strings simply be re-written to be shorter so we avoid this problem? E.g. `"Fails if source:false in leaf item"`

---

> Since it's already enabled (I mean plugin:prettier/recommended) may be it would be great to understand what does it change.

+1

jorgeorpinel:
Oh no! I'm 10 minutes late to comment on this PR after it was merged! Took me like 15 to read everything and write that last note 😋

src/Documentation/SidebarMenu/helper.test.js
```diff
@@ -257,6 +257,7 @@ describe('SidebarMenu/helper', () => {
       )
     })
 
+    // eslint-disable-next-line max-len
     it("Throws error if item has source: false and doesn't have children", () => {
```
jorgeorpinel: I think we could've just shortened this string to something like `"Throws error  if source:false in leaf item"`.

src/Documentation/glossary.js
```diff
@@ -33,8 +33,8 @@ Initialized by running \`dvc init\` in the **workspace**. It will contain the
       desc: `
 The DVC cache is a hidden storage (by default located in the \`.dvc/cache\`
 directory) for files that are under DVC control, and their different versions.
-For more details, please refer to this
-[document](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory).
+For more details, please refer to this [document]
+(/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory).
```
jorgeorpinel: Weird change. Unexpected side-effect of our ESLint config change @algomaster99 @shcheklein ? Happens again twice in the [changes to this file](https://github.com/iterative/dvc.org/pull/715/files#diff-e09d8d350a751ef31fb5badbfa968bf0) below.

src/Documentation/Markdown/Markdown.js
```diff
@@ -55,6 +55,7 @@ const HeadingRenderer = ({ level, children }) => {
         >
           <path
             fillRule="evenodd"
+            // eslint-disable-next-line max-len
             d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"
```
jorgeorpinel: Quick question @algomaster99 . If we put this string in back quotes (`) would Prettier maybe wrap it under 80 chars?

src/Documentation/glossary.js
jorgeorpinel: Ah never mind, these are MD links inside a template literal.

src/Documentation/Markdown/Markdown.js
algomaster99: @jorgeorpinel No, it doesn't.

Issue #714, opened 2019-10-20
js: refactor all constants into a single file


algomaster99:
It is nice to have a single source for all constants like `ROOT_ELEMENT`, etc. to make it feasible for resuing at different places.

shcheklein:
not sure this a good first issue. @algomaster99 do you have a list of constants we want to move?

algomaster99:
- `ROOT_ELEMENT`
- `SIDEBAR_MENU`
- `MIN_HEIGHT`

And there are a few more. These are a few examples.

jorgeorpinel:
Nice. If you can list them all in as checkboxes (and check off the ones you fix with PRs), then it will be a good first issue 🙂

shcheklein:
there also should be a list of where to _use_ these constants, right?

algomaster99:
@shcheklein We can determine it by grepping through the whole project for the particular constant.
Although, there are some constants which are used only once in a project. So I will have to decide whether they should really go into the file or not.

shcheklein:
@algomaster99 sometimes even grep can be not enough if it's already a value that we calculated using a few constants (like in your tooltip code as far as I understand).

algomaster99:
@shcheklein in my case, we are just movie the header element' class to a constant file. Not the height because we can only find its height after it has mounted. It is similar to what we have done for `bodybag`.

> sometimes even grep can be not enough

Might be. I will have to look carefully and jot them down here. :sweat_smile:

jorgeorpinel:
> there are some constants which are used only once in a project

My 2c on this is that either the constant goes in the shared file like all the others (in case it's reused later), or the value is hardcoded as a literal where it's used if you're sure no one else will ever need it.

jorgeorpinel:
Where should the constants file be located? Is there a best practice for React or Next project structures?

See https://github.com/iterative/dvc.org/pull/717/files#diff-7c51329997e372ad3b55f13cecdd28ef

shcheklein:
let's revisit this, might be not relevant with the latest migration

PR #717, opened 2019-10-21
Use a unique `id` to access the app header and content body instead of a `className`

### LINKED ISSUE #681: react: unexpected URL anchor behavior when reloading a page
For example, go to https://dvc.org/doc/user-guide/managing-external-data in a new tab and then scroll down to the examples, and click the 🔗  icon that appears when hovering the cursor over the **Examples** header. This fixes the scroll position to that anchor and adds `#examples` to the URL. Now, reload the page or even select eh address bar and press Enter. The scroll is reset to the top when the page reloads, however the URL remains as https://dvc.org/doc/user-guide/managing-external-data#examples

Only if you completely close the tab and restore it (open it again) will it scroll to the correct location.

> I'm on Version 77.0.3865.90 (Official Build) (64-bit) for Mac.

algomaster99:
Fixes #681 
Partially addresses #714

### DIFF
--- a/consts/index.js
+++ b/consts/index.js
@@ -1,0 +1,1 @@
+export const HEADER = 'header'
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -20,6 +20,8 @@ import styled from 'styled-components'
 import { media } from '../src/styles'
 // sidebar data and helpers
 import sidebar, { getItemByPath } from '../src/Documentation/SidebarMenu/helper'
+// constants
+import { HEADER } from '../consts'
 
 const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
@@ -153,7 +155,7 @@ export default class Documentation extends Component {
     const element = document.getElementById(hash.replace(/^#/, ''))
 
     if (element) {
-      const headerHeight = document.getElementsByClassName('header')[0]
+      const headerHeight = document.getElementsByClassName(HEADER)[0]
         .offsetHeight
       const elementBoundary = element.getBoundingClientRect()
       const rootElement = document.getElementById(ROOT_ELEMENT)
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -155,8 +155,7 @@ export default class Documentation extends Component {
     const element = document.getElementById(hash.replace(/^#/, ''))
 
     if (element) {
-      const headerHeight = document.getElementsByClassName(HEADER)[0]
-        .offsetHeight
+      const headerHeight = document.getElementById(HEADER).offsetHeight
       const elementBoundary = element.getBoundingClientRect()
       const rootElement = document.getElementById(ROOT_ELEMENT)
       rootElement.scroll({ top: elementBoundary.top - headerHeight })
--- a/src/Documentation/RightPanel/RightPanel.js
+++ b/src/Documentation/RightPanel/RightPanel.js
@@ -6,7 +6,7 @@ import { LightButton } from '../LightButton'
 import throttle from 'lodash.throttle'
 
 const ROOT_ELEMENT = 'bodybag'
-const MARKDOWN_ROOT = '.markdown-body'
+const MARKDOWN_ROOT = '#markdown-root'
 
 const imageChecker = (images, callback) => {
   // IE can't use forEach on array-likes
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -20,7 +20,7 @@ class DesktopView extends Component {
   }
 
   tooltipPositionEval = () => {
-    const headerHeight = document.getElementsByClassName(HEADER)[0].offsetHeight
+    const headerHeight = document.getElementById(HEADER).offsetHeight
     const markdownBody = document.getElementsByClassName('markdown-body')[0]
     const tooltipBoundary = document
       .getElementById(`tooltip-text-${this.props.id}`)
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types'
 import ReactMarkdown from 'react-markdown'
 import styled from 'styled-components'
 
+import { HEADER } from '../../consts'
+
 class DesktopView extends Component {
   state = {
     hover: false,
@@ -18,8 +20,7 @@ class DesktopView extends Component {
   }
 
   tooltipPositionEval = () => {
-    const headerHeight = document.getElementsByClassName('header')[0]
-      .offsetHeight
+    const headerHeight = document.getElementsByClassName(HEADER)[0].offsetHeight
     const markdownBody = document.getElementsByClassName('markdown-body')[0]
     const tooltipBoundary = document
       .getElementById(`tooltip-text-${this.props.id}`)
--- a/src/Tooltip/desktop-view.js
+++ b/src/Tooltip/desktop-view.js
@@ -21,7 +21,7 @@ class DesktopView extends Component {
 
   tooltipPositionEval = () => {
     const headerHeight = document.getElementById(HEADER).offsetHeight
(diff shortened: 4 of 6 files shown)


algomaster99:
@shcheklein Oh I forgot about this. :sweat_smile:  I think @iAdramelk meant to replace `className` by `id` so that it can be unique.

shcheklein:
@algomaster99 so, if you actually introduce the change required (unique id), I think this PR will become obsolete.

algomaster99:
@shcheklein I can push one more commit and change the title? :laughing:

shcheklein:
@algomaster99 👍

jorgeorpinel:
Doesn't seem to work as perfectly as the previous version. It doesn't scroll all the way to the header, the lower you go, the greater the offset. So at some point the CONTENTS table in the right side bar doesn't select the correct header either. Seems like a linear error in the offset math.

image
> https://dvc-org-pr-717.herokuapp.com/doc/tutorials/versioning#automating-capturing

jorgeorpinel:
Also, this contains changes for #714 right? Just curious.

pages/doc.js
```diff
@@ -20,6 +20,8 @@ import styled from 'styled-components'
 import { media } from '../src/styles'
 // sidebar data and helpers
 import sidebar, { getItemByPath } from '../src/Documentation/SidebarMenu/helper'
+// constants
+import { HEADER } from '../consts'
 
 const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
```
iAdramelk: Maybe we should move this id's to the consts file too?

src/Tooltip/desktop-view.js
```diff
@@ -18,8 +20,7 @@ class DesktopView extends Component {
   }
 
   tooltipPositionEval = () => {
-    const headerHeight = document.getElementsByClassName('header')[0]
-      .offsetHeight
+    const headerHeight = document.getElementById(HEADER).offsetHeight
     const markdownBody = document.getElementsByClassName('markdown-body')[0]
```
iAdramelk: Theoretically we can have more than 1 element with `markdown-body` class because it is the class that applies markdown css rules and there is no guaranties that  thepage body will always be the first one. Maybe we need to add unique id there and place it to consts file?

pages/doc.js
algomaster99: Yeah, I have plans to. But not in this PR.

algomaster99:
@jorgeorpinel This existed before changing it to `id` as well.
The problem is really weird though.
On checking the output of 
```
 console.log(elementBoundary.top, '= top')
```
it shows two different results when hard-reloaded and then reloading by server.
Screenshot from 2019-10-23 20-51-46

The `elementBoundary.top` signifies the position of the header w.r.t. the viewport. I am confused how it shows two different outputs.

P.S. This is only happening with markdown files with large text inside.

shcheklein:
@algomaster99 this a known issue with pages that have images - it takes time to load them. I think it can be fixed as a separate issue (by providing a hint along with the image? I don't know yet).

@iAdramelk @algomaster99 either I don't understand what the _unique_ id mean or this is not done yet?

algomaster99:
@shcheklein this is done. It was on hold because of https://github.com/iterative/dvc.org/pull/717#issuecomment-545170093

algomaster99:
@shcheklein Can you open a separate issue for this delay in image loading? Also, have you thought of any solution to this?

shcheklein:
@algomaster99 I think you should have more context to describe it - so, feel free to create a ticket :) No, I don't have any good solution in mind, except some complicated logic that will be calculating height of every # before loading. I would also take a look if it's a problem other people have tried to or solved already. I can imagine that any website that has images should have a similar issue, right?

src/TopMenu/index.js
```diff
@@ -50,7 +52,7 @@ class TopMenu extends Component {
     return (
       <Wrapper>
         <Container
-          className="header"
```
shcheklein: are there any css rules by chance behind this class?

src/Tooltip/desktop-view.js
shcheklein: @algomaster99 are you going to address this?

algomaster99: @shcheklein i think it is better to address it here https://github.com/iterative/dvc.org/issues/714

src/TopMenu/index.js
algomaster99: @shcheklein I have checked it. There are none. Only place the header class is used now is in the tooltip component to style the header inside the tooltip.

src/Tooltip/desktop-view.js
shcheklein: @algomaster99 It does not seem to be related to #714 . It's about changing Class to unique Id in the first place, right? and only after that about moving something to the const file.

src/Documentation/RightPanel/RightPanel.js
```diff
@@ -6,7 +6,7 @@ import { LightButton } from '../LightButton'
 import throttle from 'lodash.throttle'
 
 const ROOT_ELEMENT = 'bodybag'
-const MARKDOWN_ROOT = '.markdown-body'
+const MARKDOWN_ROOT = '#markdown-root'
```
shcheklein: if it's not used yet, let's remove it from this PR

algomaster99: @shcheklein Used what? Can you please elaborate? :)

algomaster99:
@shcheklein Since this PR doesn't only involve changing `className` to `id` for the header, I think I should also add an `id` for [`tooltip-container`](https://github.com/iterative/dvc.org/blob/593c020d3c84d4bfad077625f35c7bfff30fdb72/src/Tooltip/desktop-view.js#L30).

src/Documentation/RightPanel/RightPanel.js
shcheklein: yep, my bad, I see.

shcheklein:
@algomaster99 let's do as a separate one.

iAdramelk:
@algomaster99 we had similar problem on with right panel. We solved it using this code https://github.com/iterative/dvc.org/blob/master/src/Documentation/RightPanel/RightPanel.js#L11-L43

But in only solves it for images, if there is some other elements that can change container height, there is still be a bug.

jorgeorpinel:
Can't the scrolling (re)run after `window.onload` with some vanilla JS?

iAdramelk:
@jorgeorpinel if I remember correctly `window.onload` triggers before images are loaded.

iAdramelk:
Easiest way to solve this problem would be by using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe), but unfortunately Safari and IE didn't support it and it's polyfills are large and slow.

jorgeorpinel:
> The load event fires at the end of the document loading process. At this point, all of the objects in the document are in the DOM, and all the images, scripts, links and sub-frames have finished loading.

From https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onload

iAdramelk:
@jorgeorpinel Yes, you are right. I misremembered what the problem was. `window.onload` is indeed fires after initial page load. But in our case we downloading and rendering markdown files after initial page load so we can't use window.onload to find if images are loaded or not.

jorgeorpinel:
Ah, makes sense thanks for the reminder. Oh well, scrolling seems to work great now 🙂

jorgeorpinel:
p.s. never mind, reloading URLs with #anchors in pages with images before the anchor (e.g. https://dvc.org/doc/tutorials/versioning#preparation) still misses the offset. Should we open a separate issue for this?

jorgeorpinel:
UPDATE: created #750 for this.

consts/index.js
```diff
@@ -0,0 +1 @@
+export const HEADER = 'header'
```
jorgeorpinel: I think this file should be located in src/ or in the root folder. Why consts/?

Issue #720, opened 2019-10-21
user-guide: rename user-manual?


jorgeorpinel:
- [ ] Rename **User Guide** section to **User Manual**? Better naming because we can call each document inside a "guide" without causing confusion with the parent section (the manual).

- [x] And update the guide title "Update Tracked Files" to "Updating".

Would also need to setup a (tmp) redirect in server.js, update the GitHub label, et al.

shcheklein:
@jorgeorpinel up to you :) but, I don't see a strong reason to do so. What sections do you want to call "guide" inside?

jorgeorpinel:
- [ ] and update the guide title "Update Tracked Files" to "Updating".

jorgeorpinel:
Yeah, not a huge deal, not sure it's worth the trouble but I do think it would be an improvement.

> What sections do you want to call "guide" inside?

All the documents inside would be "the guides". The ones listed in https://dvc.org/doc/user-guide.

shcheklein:
@jorgeorpinel if all of them are "guides" in an implicit sense, then it's correct to have a "User Guide" name. I think it's better to not repeat Guide multiple times inside.

jorgeorpinel:
We don't really have to call them "guides". In fact we don't often do. But if we're going to, at least call the parent section "User Guide**s**"? (plural). But TBH I don't think this section is much of a guide. Guiding information doesn't do into detail, just uses the most convenient steps to get from A to B. I think the Get Started section could be seen as a guide, but the stuff here is some of the most technical, more like a manual.

The usage of the term "guide" should definitely be reviewed though. We don't have much consistency there. I'll take a look...

shcheklein:
I think "* Guide" is a pretty established naming in tech docs. Check HBase docs, look for Administrator Guide, Programmer's Guide, etc. All of them consist internally out of other smaller "guides".

shcheklein:
also - https://en.wikipedia.org/wiki/User_guide

jorgeorpinel:
Good arguments. But still I'm not convinced.

[...]

Issue #702, opened 2019-10-16
install: add instructions to update/uninstall DVC


jorgeorpinel:
At least for the recommended install methods for each platform

> First mentione in https://github.com/iterative/dvc.org/pull/695#pullrequestreview-302330591

dashohoxha:
@jorgeorpinel Don't you think that the user should know how to use his commands (`apt`, `yum`, `pip`, etc.)? So, do we need to tell him how to update/upgrade a package?

I think that the same argument goes for telling users when and how to use a virtualenv. They should know when and how to use it, it should not be our concern. But this is a different topic.

jorgeorpinel:
Yes, it's a good point @dashohoxha. But at least for `pip`, since upgrading can be different if you install from PyPi vs. if you install from local code, maybe we should include the commands in our docs.`

What do you think @shcheklein?

jorgeorpinel:
On second thought I don't think my argument agains teaching other technologies really applies to installing and upgrading DVC, even if it's done with a 3rd party tool.

In fact maybe we should even include instructions to **uninstall** DVC as well somewhere. (Other similar tools offer this kind of information, probably because users look for it e.g https://devcenter.heroku.com/articles/heroku-cli#uninstalling-the-heroku-cli.)

Also, what should be the right term be: "update" or "upgrade"? 😋

algomaster99:
@jorgeorpinel oh-my-zsh also does this. I think we should also have a command like `uninstall_dvc` which removes dvc installed from source or binary. Otherwise, there will be many separate instructions to uninstall it. If you think this is valid, please make it a separate issue.

jorgeorpinel:
Not a bad idea. Opened iterative/dvc/issues/2637 in parallel.

The question is should we provide instructions in the docs for now?

shcheklein:
@jorgeorpinel nice, but not a top priority I would say. Uninstall can be really useful in some case (Mac OS package for example).

jorgeorpinel:
From https://github.com/iterative/dvc.org/issues/65#issuecomment-504713595:

Related discussion: https://discordapp.com/channels/485586884165107732/485596304961962003/574941227624169492

For Mac (installed from pkg):

```
$ which dvc
/usr/local/bin/dvc -> /usr/local/lib/dvc/dvc

$ ls -la /usr/local/bin/dvc
/usr/local/bin/dvc -> /usr/local/lib/dvc/dvc

sudo rm -f /usr/local/bin/dvc
[...]

PR #725, opened 2019-10-23
docs: regular updates (late Oct)

### LINKED ISSUE #698: user-guide: improve Windows POSIX/terminal tips
For windows users is necessary past to the same line the command when has `\` in the end of line.

For example:
`dvc get https://github.com/iterative/dataset-registry \
          tutorial/ver/data.zip`

to

`dvc get https://github.com/iterative/dataset-registry tutorial\ver\data.zip`

**It's also necessary use `\` in the path on windows.**

### LINKED ISSUE #701: user-guide: merge/remove duplicate pages from new Installation subsection
- [x] Move installation page to the top level of the docs (currently it is under the user-guide).
- [x] Remove https://dvc.org/doc/user-guide/plugins (which is now located under installation). See https://github.com/iterative/dvc.org/pull/695#pullrequestreview-302052801
- [x] Update https://dvc.org/doc/user-guide/development. See details in #700.
- [x] Simplify https://dvc.org/doc/get-started/install and maybe merge it with the next section (https://dvc.org/doc/get-started/initialize). See https://github.com/iterative/dvc.org/pull/695#pullrequestreview-302050578
- [x] Find all links to the removed pages and redirect them to the new pages.

Depends on: #695
Related: #700 , #660

### LINKED ISSUE #736: scripts: update autocomplete instructions
We are now including bash and zsh completion scripts into our deb, rpm and brew packages. So users of those don't need to get the scripts by-hand.

jorgeorpinel:
Addresses #720
Fixes #701
Fixes #698
Fixes #736
Addresses #702


static/docs/use-cases/data-registry.md
```diff
@@ -0,0 +1,113 @@
+# Data Registry
```
shcheklein: should stay in a separate PR?

jorgeorpinel: Oh shoot. That shouldn't be here... Let me check what happened.

jorgeorpinel: Ok removed. That was an error, thanks for noticing!

static/docs/install/macos.md
```diff
@@ -20,7 +20,7 @@ Get the PKG package (binary) from the big "Download" button on the
 > [virtual environment](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments)
 > or
 > [pipx](https://packaging.python.org/guides/installing-stand-alone-command-line-tools/)
-> for Python 3.6+ to keep your system environment safe.
+> (on Python 3.6+) to encapsulate your local environment.
```
shcheklein: I'm fine with either, but the meaning is a little bit different. The previous version is stronger in a sense that it answers second _why_ - `you want to encapsulate your local environment to keep your system environment safe`. One of the main reasons to do this.

static/docs/install/pre-release.md
```diff
@@ -1,29 +1,27 @@
...
 
 > It's highly recommended using
 > [virtual environment](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments)
 > or
 > [pipx](https://packaging.python.org/guides/installing-stand-alone-command-line-tools/)
-> for Python 3.6+ to keep your system environment safe.
+> (on Python 3.6+) to encapsulate your local environment.
 
 ```dvc
+$ virtualenv -p python3 .env
```
shcheklein: may be we should exclude those since this instruction is not special comparing to the regular one with pip

May be we can add a separate paragraph: "If you want to keep your previous version installed with pip, don't forget ..."

static/docs/install/macos.md
jorgeorpinel: To me the word safe here may imply that if you don't use a virtual env your computer may be corrupted, hacked or something scary. To me, the main reason for virtual Python envs is avoid dependency conflicts. Not sure that qualifies as safety. 🙂

jorgeorpinel: Maybe there's a good term in between that will convey this better... How about "keep your global Python environment healthy"?

shcheklein: nah, it's fine as you changed it :) let's not overthink this. link and note should be enough. Just wanted to mention since to me it's close to "your computer may be corrupted" - a lot of users are not able untangle that hell pip can create with their system environment.

static/docs/install/pre-release.md
jorgeorpinel: Removed.

static/docs/install/macos.md
jorgeorpinel: Especially on Mac where there are site-packages in 3 different system dirs, some of which are protected by special MacOS file security 😋

static/docs/install/index.md
```diff
@@ -7,8 +7,17 @@ Please double check that you don't already have DVC (for example running
...
+## Advanced options
 
-It is recommended to setup [shell completion](/doc/install/completion) as well.
-And check out these [plugins](/doc/install/plugins), which might be useful too.
+- Shell completion is included with certain installation methods
+  ([Debian/UbuntuFedora/CentOS Linux repositories](/doc/install/linux#install-from-repository)
+  and
+  [MacOS Homebrew](http://localhost:3000/doc/install/macos#install-with-brew)),
+  and it can be easily [setup manually](/doc/install/completion) for most
+  others.
```
jorgeorpinel: @efiop @mroutis can you confirm this is correct? Specifically the "for most others" part. By this I mean that we don't have instructions to setup Bash completion except for MacOs and Debian/Ubuntu. (See https://dvc.org/doc/install/completion.) Or is this doc incomplete? Thanks!

shcheklein: `http://localhost:3000/doc/install/macos#install-with-brew` - localhost? )

Also, I would simplify it on the index page - "Shell completion is enabled by default in certain cases, if it does not work please see this instruction". No need to go into details too much

static/docs/install/completion.md
```diff
@@ -27,15 +27,24 @@ Depending on what you typed on the command line so far, it completes:
...
 Use the command `echo $0` to check your active shell, if you are not sure which
 one you are using (Bash or Zsh).
 
 Depending on the shell that you are using, follow the instruction below to
 install and configure shell completion.
 
-## Bash completion on Mac OS
+## Manual setup
+
+Note that Shell completion is automatically included when DVC is installed on
```
shcheklein: Again, I would simplify and put it as a first sentence into each section below. And would keep H2 (for better navigation)

static/docs/tutorials/deep/define-ml-pipeline.md
```diff
@@ -6,24 +6,28 @@ To include a data file into your data science environment, you need to copy the
...
-</details>
-
 ```dvc
 $ mkdir data
 $ wget -P data https://data.dvc.org/tutorial/nlp/100K/Posts.xml.zip
 $ du -sh data/*
  41M data/Posts.xml.zip
 ```
 
+<details>
```
shcheklein: can we do the same for the get started - wrap the same info into expandable section?

static/docs/install/index.md
jorgeorpinel: Oops. fixed `localhost` URLs.

OK, simplified note, then.

static/docs/install/completion.md
jorgeorpinel: This doc is specifically about completion, so maybe we can keep the details here? Or are you thinking in the future more installation methods will include shell completion so these comments may not be needed?

shcheklein: yeah, but having something like `Install on Mac Os`, `Install on Linux`, etc seems to be more relevant/actionable than just a single `Manual Setup` item on the right side menu - that was my thinking.

jorgeorpinel: Ah yes well that's an engine thing. I didn't think to check the side bar, which only cares about H2, it seems.. I'm changing the headers again then... Please check! https://dvc-org-pr-725.herokuapp.com/doc/install/completion

static/docs/tutorials/deep/define-ml-pipeline.md
jorgeorpinel: I'm not following. The only reference to downloading with Windows in the get-started is in https://dvc.org/doc/get-started/connect-code-and-data, already collapsed in the **Expand to prepare example code** block.

static/docs/install/completion.md
```diff
@@ -2,6 +2,12 @@
 
 Enjoy working with DVC faster and with fewer typos!
 
+> Note that shell completion is automatically enabled when DVC is installed on
+> MacOS [with Homebrew](/doc/install/macos#install-with-brew), or on
+> Debian/Ubuntu Linux
+> [from deb repository](/doc/install/linux#install-from-repository). Please
```
shcheklein: what about rpm?

static/docs/install/completion.md
```diff
@@ -2,6 +2,12 @@
 
 Enjoy working with DVC faster and with fewer typos!
 
+> Note that shell completion is automatically enabled when DVC is installed on
```
shcheklein: We can make it just a regular paragraph, what do you think?

static/docs/install/index.md
```diff
@@ -3,12 +3,18 @@
...
 - [Install on Windows](/doc/install/windows)
 - [Install on Linux](/doc/install/linux)
 
-You can also [install the development version](/doc/install/pre-release), before
-the official release.
+## Advanced options
 
-It is recommended to setup [shell completion](/doc/install/completion) as well.
-And check out these [plugins](/doc/install/plugins), which might be useful too.
+- Shell completion is automatically enabled by certain installation methods. If
```
shcheklein: 👍

static/docs/tutorials/deep/define-ml-pipeline.md
shcheklein: ah, I see. My bad. It is indeed a part of the larger expandable section.

static/docs/install/completion.md
jorgeorpinel: According to #736, these are the only packages where it's included:

> We are now including bash and zsh completion scripts into our deb, rpm and brew packages.

shcheklein: so, yes _rpm_ is here :) it's a separate Linux repo for packages, similar to Debian/Ubuntu, but different. 

So, either `Debian/Ubuntu Linux` and `from deb repository` or there should be a second link to the `install from rpm repository`

static/docs/install/completion.md
jorgeorpinel: You're right, updated!

static/docs/install/completion.md
jorgeorpinel: Ah yes sorry I was confused haha, adding... Hold on.

jorgeorpinel: OK, added. Please check.

jorgeorpinel:
Please note I also just commited d557fe1 @shcheklein. Not sure how you'll feel about this.

shcheklein:
@jorgeorpinel tbh, I think it's a signal that some refactoring is needed when we have two of three notes in a row. I would try to avoid them in the first place. Otherwise I'm totally fine with those delimiters, though don't quire understand if they are here to make it easier to maintain these files and avoid mistakes or there are some other reasons?

jorgeorpinel:
> don't quire understand if they are here to make it easier to maintain these files

The reason for those comments is so the quotes are not merges into one when the Markdown is rendered. This way there is an empty line between notes (signaling these are different topics).

Issue #727, opened 2019-10-23
docs: remove unnecessary index pages that serve as a table of contents


shcheklein:
**Blocked by #738**

E.g. https://dvc.org/doc/user-guide/contributing

It'll be very painful to support them. We will need to keep titles up to date, list of them, snippets, etc, etc.

Let's not use pages like this at all or come with a way to generate them automatically.

cc @dashohoxha

jorgeorpinel:
The point of having them is that we can add some text about the purpose of the section, and summary of each page inside. All of those texts are custom, they don't come from inside any of the other pages. So auto-generating them won't be able to get these same results, so I think the decision is either keep them or remove them. If we don't want to have to maintain them, let's just delete all these index.md files.

shcheklein:
@jorgeorpinel yep, I don't mean pages that have some unique content or some introductory text. I specifically mean pages https://dvc.org/doc/user-guide/contributing that do not have any new content but repeat the structure. Moreover I really like that page, but I understand that it'll be a nightmare to keep those in sync w/o proper tooling.

dashohoxha:
> Let's not use pages like this at all

I have tried ` "source": false,` on `sidebar.json`, but it doesn't work (unless it is the top level, like `understanding-dvc`). Maybe it should be fixed on the engine.

> come with a way to generate them automatically

Again this seems like it needs to be fixed on the engine. I could be specified with a field like `"source": "auto",` or `"source": "index",`

dashohoxha:
> If we don't want to have to maintain them, let's just delete all these index.md files.

@jorgeorpinel As I said above, currently this is not possible, as it breaks the engine. Maybe this could be regarded as bug too, because I would expect it to work automatically without index pages.

> so I think the decision is either keep them or remove them

I agree with @shcheklein , sometimes they are useful, sometimes they are a burden. I think that autogenerating an index page should be the default, unless the `"source"` is given explicitly, or is specified as `false`.

shcheklein:
@dashohoxha 

> As I said above, currently this is not possible

It should be possible - check understanding DVC and/or changelog sections.

jorgeorpinel:
> yep... I specifically mean pages https://dvc.org/doc/user-guide/contributing

Ah, Ok. Never mind Ivan.

> I have tried ` "source": false,` on `sidebar.json`, but it doesn't work (unless it is the top level, like `understanding-dvc`)... it breaks the engine. Maybe this could be regarded as bug...

[...]

PR #732, opened 2019-10-24
Fix #563: How to keep data and cache on external drive

### LINKED ISSUE #563: how: use DVC when data is stored in an external drive
E: Check whether #520 was done first... See also #899

---

This doc should explain the best solution (or a couple of possible solutions) for this situation.

Example: the data is located in a partition of size 16TB on an external drive, while the DVC project is on `/home` of a partition of size 320GB.

> Context: https://discordapp.com/channels/485586884165107732/485596304961962003/611244643685892153

dashohoxha:
Fix #563
Replaces https://github.com/iterative/dvc.org/pull/565


src/Documentation/sidebar.json
```diff
@@ -132,6 +132,17 @@
       {
         "label": "Anonymized Usage Analytics",
         "slug": "analytics"
+      },
+      {
+        "label": "How To",
+        "slug": "howto",
+        "source": "howto/index.md",
+        "children": [
```
jorgeorpinel: Do we plan to have more how to guides soon? This is probably a good idea but just checking whether there is an immediate need for a 3rd level for this one doc.

static/docs/user-guide/howto/index.md
```diff
@@ -0,0 +1,3 @@
+# How To
+
+- [Keep Data and Cache on External Drive](/doc/user-guide/howto/keep-data-and-cache-on-external-drive)
```
jorgeorpinel: See https://github.com/iterative/dvc.org/issues/727#issuecomment-545756811.

jorgeorpinel:
Cool. Does this PR include my commits from the other one, abd2446, b3d5c6b, and f345245?

image

Please `git cherry-pick` them if not, they address several issues that were long standing in that PR.

src/Documentation/sidebar.json
dashohoxha: > Do we plan to have more how to guides soon?

Sure. I am planning to add some more. You can also add some if you wish.

> This is probably a good idea but just checking whether there is an immediate need for a 3rd level for this one doc.

I think that the "How To" section is going to become a long one (having lots of items), and eventually it might need to be split from the User Guide as a top-level section on its own. For the time being it is Ok nested under the User Guide.

static/docs/user-guide/howto/index.md
dashohoxha: I agree. But for the time being let's keep `index.md`, until the engine is fixed, and remove it later. It should not be difficult to remove it.

dashohoxha:
> Does this PR include my commits

Sure, it does include all the commits. I have also made some other small modifications.

static/docs/user-guide/howto/index.md
jorgeorpinel: Yes. You're right.

src/Documentation/sidebar.json
jorgeorpinel: Sounds good but I think it's best to create the 3rd level when more guides are added. What do you think @shcheklein?

dashohoxha:
> I haven't done an extensive check but just looking at the first file in my first commit, lines 40-43 (abd2446#diff-278c4173e3a65d532906f6a613dc7ce0R40-R43) for example are not included in this PR.

I did some regular updates and I found that windows link superfluous and out of place, since this HowTo does not cover at all the windows case, there are Linux commands everywhere (which might also work in Mac, but no guarantee about it). If you want to cover the Windows case maybe you need to add another HowTo which does it thoroughly.

> I guess there was a reason to so it, but having closed the original PR and opened a new one from a different branch makes it extra time consuming to track and review the continuity in these changes. I tried just doing git cherry-pick to apply my work again, but since the file is located somewhere else, this method does not work.

You don't need to track the commits. Just read the HowTo and see whether it is in a good shape or not: https://dvc-org-pr-732.herokuapp.com/doc/user-guide/howto/keep-data-and-cache-on-external-drive

> We'll probably have to manually compare and reapply my changes (combined here: 482e1c7...f345245) vs those in this PR.

If I have removed something and you re-apply your changes (that you don't even remember why they were added in the first place), then you are going to undo my work.
Remember that those changes were meant for a UseCase, and this is a HowTo, and I believe you know the difference, don't you?

In any case, whatever changes/improvements you may think of, I am sure that they are not blocking changes. So let's merge first this PR, and make the changes later during regular updates.

jorgeorpinel:
> I did some regular updates and I found that windows link superfluous and out of place...

You may find it out of place but I added that note because Ivan suggested it was necessary. Refer to https://github.com/iterative/dvc.org/pull/565#pullrequestreview-276195808. I think we should discuss and decide, not just discard previous work.

> You don't need to track the commits... you don't even remember why they were added in the first place...

All of my changes had a similar reason, which is why I'm tracking them. It's also how we collaborate. But don't worry, I am going to apply everything again myself when I get a chance (keeping in mind this is now a how-to, not a use-case), and I'll be happy to discuss any review comments you open in this PR on things you think are not needed – you're often right. 

> you are going to undo my work...

I don't see how, but yes, we're gonna have to do some double work unfortunately, given the PR switch. But again, I do understand it helps us have a clean slate, which is nice, since the previous PR was stale and too long. So let's just focus on getting it done.

> they are not blocking changes...

Again, this we all have to decide together.

> p.s. There are some pending review comments by Ivan in the old PR you may want to double check, for example: https://github.com/iterative/dvc.org/pull/565#pullrequestreview-276195595, https://github.com/iterative/dvc.org/pull/565#pullrequestreview-276196319, and https://github.com/iterative/dvc.org/pull/565#pullrequestreview-278196181 for example.

jorgeorpinel:
(I'm hiding these couple long comments above ^ to simplify reading this PR.)

src/Documentation/sidebar.json
shcheklein: yes, I think we'll have to introduce a how to section soon. There are a few tickets that start with "How to ..." already - probably good candidates for that. May be we can move things like "Running on Windows" into this section>

shcheklein:
@dashohoxha it has really created a problem reviewing this PR. Please, move some of the concerns I and/or @jorgeorpinel raised in the initial one with your responses. Otherwise you force us to do this. I'm not sure why there was a need to create a second one instead of iterating on the first since we effectively lost a lot of context by doing this.

src/Documentation/sidebar.json
jorgeorpinel: OK resolving this then. Thanks

jorgeorpinel:
> - This https://github.com/iterative/dvc.org/pull/732#pullrequestreview-307512983 I'll address myself ASAP.

Done! Most of my old changes were indeed already in this new version but I had to review line by line to double check. A few were gone and I added again. I also took the opportunity to make small misc. wording improvements, etc. Feel free to review and comment on this latest version and I'll address.

Glad we're finally getting it over with. Thanks.

static/docs/user-guide/howto/keep-data-and-cache-on-external-drive.md
```diff
@@ -0,0 +1,200 @@
...
+$ chmod u+rw -R /mnt/external-drive/
+```
+
+> Or refer to
+> [User Account Control](https://docs.microsoft.com/en-us/windows/security/identity-protection/user-account-control/user-account-control-overview)
+> for Windows.
+
+## Create a DVC project with external cache
+
+An [external cache](/doc/user-guide/external-outputs) resides outside of the
```
shcheklein: link is broken

static/docs/user-guide/howto/keep-data-and-cache-on-external-drive.md
```diff
@@ -0,0 +1,200 @@
+# Keep Data and Cache on External Drive
+
+Sometimes the data may be stored on an
+[external hard drive](https://whatis.techtarget.com/definition/external-hard-drive).
+Usually such data is huge, which means that it won't fit on our local drive, and
```
shcheklein: This motivation is not enough. Do you remember where did this case come from? It's not clear why do I need two drives, what advantages does this configuration has if I store everything - input data, results, etc outside? I understand the case when we do this with S3, for example.

static/docs/user-guide/howto/keep-data-and-cache-on-external-drive.md
```diff
@@ -0,0 +1,200 @@
...
+usual.
+
+For example, let's say that the raw data files are in
+`/mnt/external-drive/data/raw/`, and you are cleaning them up. You could do it
+like this:
+
+```dvc
+$ dvc add /mnt/external-drive/data/raw
+...
+$ dvc run \
```
shcheklein: if you do this, it's better to have a separate directory per project on the external drive. Or, again, I don't quite understand the motivation and use case behind this configuration.

static/docs/user-guide/howto/keep-data-and-cache-on-external-drive.md
dashohoxha: > Do you remember where did this case come from?

Normally I would not remember, however the issue #563 has some context about it. A user has requested help for this situation on the chat. I am not sure why he has to keep the data on an external drive, but keep the project on his computer. But this is how he prefers to organize his work.

Recently I have seen another discussion where a user wants to keep all the data and cache on an external directory, on a fixed absolute path, so that he can search easily the code of the project. Again I am not sure whether this is the most efficient solution for him, but this is what he does. (Sorry, I cannot find a reference for this discussion.)
I would say that the external-data case is very similar to the external-drive case.

> I understand the case when we do this with S3, for example.

I am not sure about the benefits of using this approach for the S3 case, because the data would still have to be downloaded, in order to be processed locally, and it would still have to be uploaded, in order to be saved remotely. The benefits are not so clear, in my opinion. You might be able to save some local storage space, but this comes at the cost of increased downloads and uploads.

dashohoxha:
> The biggest concern is that I don't understand the motivation for this combination.

I left a comment here: https://github.com/iterative/dvc.org/pull/732#discussion_r339773638

static/docs/user-guide/howto/keep-data-and-cache-on-external-drive.md
dashohoxha: > if you do this, it's better to have a separate directory per project on the external drive

I don't understand your request here. It is just an example, why should we talk about several projects?

> I don't quite understand the motivation and use case behind this configuration.

I have tried to explain it here: https://github.com/iterative/dvc.org/pull/732#discussion_r339773638

I will try to revise the motivation including this one too:

> Recently I have seen another discussion where a user wants to keep all the data and cache on an external directory, on a fixed absolute path, so that he can search easily the code of the project.

dashohoxha:
I changed the title and rewrote the motivation. I hope it looks better now.

I was tempted again to remove the reference about Windows (https://github.com/iterative/dvc.org/pull/732/files#diff-e67a9ecb076e21d068e0f5559ed64582R30) but I didn't dare. @jorgeorpinel it really seems out of place to me, please think about removing it again.

shcheklein:
btw, @dashohoxha how does this one correlate with the new data sharing section?

dashohoxha:
> how does this one correlate with the new data sharing section?

I don't see any correlation.
Maybe there is some correlation with the external data management (it could be an example of it), but I think that it deserves to be a standalone HowTo.

static/docs/user-guide/howto/external-data-and-cache.md
```diff
@@ -0,0 +1,151 @@
...
+> guide about sharing data with a
+> [mounted DVC storage](/doc/user-guide/data-sharing/mounted-storage).
+
+## Make the data directory accessible
+
+Let's assume that the data of our project is located on `/data/project1/`. First
+we have to make sure that we can read and write the `/data/` directory:
+
+```dvc
+$ sudo chown <username>: -R /data/
```
shcheklein: I think this is not right to provide chown, chmod and other commands to manage permissions. They are not actionable anyway. Exact commands you need to run depend on the case. Some general suggestions to check the permissions should be enough?

static/docs/user-guide/howto/external-data-and-cache.md
```diff
@@ -0,0 +1,151 @@
+# Keep Data and Cache Outside the Project
+
+Sometimes we would like to keep the data and cache outside the project
```
shcheklein: not sure how is it related to being able to search code tbh, could you elaborate?

static/docs/user-guide/howto/external-data-and-cache.md
```diff
@@ -0,0 +1,151 @@
...
+directory, so that we can search easily the code of the project.
+
+Keeping the data outside of the project, on a fixed absolute path, is also
```
shcheklein: it's a a very tricky case - if multiple projects use the same data how do they switch different versions?

shcheklein:
I don't see how does it replace #565 or fixes #563. It's a very specific how-to focused on external dependencies and output on a local machine. Also, it reminds me a lot some of the other data managements sections you put together.

src/Documentation/sidebar.json
shcheklein: @jorgeorpinel this is the discussion around How to that happened a while ago ^^

shcheklein:
closing this as stale

PR #756, opened 2019-10-29
Update metrics.md

TobiRoby:
rewording sentence

### DIFF
--- a/static/docs/get-started/metrics.md
+++ b/static/docs/get-started/metrics.md
@@ -1,6 +1,6 @@
 # Experiment Metrics
 
-The last stage we would like to add to our pipeline is its the evaluation. Data
+As a final stage we would like to add an evaluation part to our pipeline. Data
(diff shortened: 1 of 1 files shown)


static/docs/get-started/metrics.md
```diff
@@ -1,6 +1,6 @@
 # Experiment Metrics
 
-The last stage we would like to add to our pipeline is its the evaluation. Data
+As a final stage we would like to add an evaluation part to our pipeline. Data
```
jorgeorpinel: Thanks for noticing the typos! Definitely better. But let's avoid the "part" term and keep "stage" related to "pipeline". How about "Finally, we would like to add an evaluation stage to our pipeline"?

jorgeorpinel: I'l address this.

jorgeorpinel:
Thanks @TobiRoby!

PR #769, opened 2019-11-03
hotfix: server redirect of man.dvc.orl/import-url and

### LINKED ISSUE #768: man.dvc.org/import-url redirect is broken
It supposed to behave in the same way as `man.dvc.org/get-url`

The fix should be applies to the server.js file.

jorgeorpinel:
Fixes #768


shcheklein:
thanks, @jorgeorpinel !

jorgeorpinel:
Oops this may be wrong haha let me hotfix the hotfix...

PR #770, opened 2019-11-03
hotfix previous hotfix! (#769)

### LINKED ISSUE #768: man.dvc.org/import-url redirect is broken

jorgeorpinel:
Actually fixes #768
(Finishes #769)


jorgeorpinel:
Can't test this on the review app unfort. Logic should be correct according to my tests on `node` console...

jorgeorpinel:
Also not working. Weird!

PR #775, opened 2019-11-05
Updated the last stage

hiteshsp:
The last stage value was showing "model.pkl.dvc" instead of "train.dvc"
https://dvc.org/doc/get-started/visualize

Stages > dvc pipeline show --ascii train.dvc


shcheklein:
thanks @hiteshsp !

PR #776, opened 2019-11-05
Basic React code refactoring & linting

### LINKED ISSUE #716: lint: resolve errors in pages/ and update lint target
UPDATE: See latest known state in https://github.com/iterative/dvc.org/issues/716#issuecomment-549693282.

jorgeorpinel:
Reorganizes code in first pages and a few components so order makes more sense (e.g. imports used in order, general components before specific ones, etc), adds basic code documentation (not sure needed), and fixes #716.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
     "format-check": "pretty-quick --check --verbose",
-    "lint": "eslint src pages"
+    "lint-check": "eslint src pages"
   },
   "repository": {
     "type": "git",
@@ -64,12 +64,16 @@
     "eslint-plugin-react": "^7.14.3",
     "husky": "^3.0.4",
     "jest": "^24.9.0",
+    "lint-staged": "^10.0.0-1",
     "prettier": "^1.18.2",
     "pretty-quick": "^1.11.1"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn format-staged && yarn lint"
+      "pre-commit": "yarn format-staged && lint-staged"
     }
+  },
+  "lint-staged": {
+    "*.{js,jsx}": "eslint"
   }
 }
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
     "format-check": "pretty-quick --check --verbose",
-    "lint": "eslint src"
+    "lint": "eslint src pages"
   },
   "repository": {
     "type": "git",
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -1,6 +1,7 @@
+import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
-import Router from 'next/router'
+// FIXME: import Router from 'next/router'
 
 const DESCRIPTION =
   `Open-source version control system for Data Science and` +
@@ -25,7 +26,7 @@ export default class MyDocument extends Document {
   render() {
     return (
       <>
-        <html>
+        <html lang="en">
           <Head>
             <meta charSet="utf-8" />
             <meta name="description" content={DESCRIPTION} />
@@ -43,7 +44,9 @@ export default class MyDocument extends Document {
             <meta
               prefix="og: http://ogp.me/ns#"
               property="og:description"
-              content="DVC is designed to handle large data files, models, and metrics as well as code. DVC is an open-source framework and distributed version control system for machine learning projects."
+              content="DVC is designed to handle large data files, models, and
+              metrics as well as code. DVC is an open-source framework and
+              distributed version control system for machine learning projects."
             />
             <meta
               prefix="og: http://ogp.me/ns#"
@@ -57,11 +60,14 @@ export default class MyDocument extends Document {
             />
             <meta
               name="twitter:description"
-              content="DVC is designed to handle large data files, models, and metrics as well as code. DVC is an open-source framework and distributed version control system for machine learning projects."
+              content="DVC is designed to handle large data files, models, and
+              metrics as well as code. DVC is an open-source framework and
+              distributed version control system for machine learning projects."
             />
             <meta
               name="viewport"
-              content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"
+              content="width=device-width, initial-scale=1.0, maximum-scale=1.0,
+              user-scalable=0"
             />
             <link
               rel="shortcut icon"
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -11,7 +11,7 @@ const inject = str => (
   <div className="inject" dangerouslySetInnerHTML={{ __html: str }} />
 )
 
-export default class Documentation extends Document {
+export default class Page extends Document {
   static getInitialProps({ renderPage }) {
     const sheet = new ServerStyleSheet()
     const page = renderPage(App => props =>
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -1,8 +1,3 @@
-/*
- * HTML document
- * Uses Next.
- */
-
 import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
@@ -16,7 +11,7 @@ const inject = str => (
   <div className="inject" dangerouslySetInnerHTML={{ __html: str }} />
 )
 
-export default class MyDocument extends Document {
+export default class Documentation extends Document {
   static getInitialProps({ renderPage }) {
     const sheet = new ServerStyleSheet()
     const page = renderPage(App => props =>
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -6,7 +6,6 @@
 import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
-// import Router from 'next/router'
 
 const DESCRIPTION =
   `Open-source version control system for Data Science and Machine Learning ` +
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -1,12 +1,16 @@
+/*
+ * HTML document
+ * Uses Next.
+ */
+
 import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
-// FIXME: import Router from 'next/router'
+// import Router from 'next/router'
 
 const DESCRIPTION =
-  `Open-source version control system for Data Science and` +
-  ` Machine Learning projects. Track your data, models, and experiments with` +
-  ` a Git-like tool.`
+  `Open-source version control system for Data Science and Machine Learning ` +
+  `projects. Track your data, models, and experiments with a Git-like tool.`
 const KEYWORDS = `data version control machine learning models management`
 
 const inject = str => (
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -1,7 +1,8 @@
 /* global docsearch:readonly */
 
 /*
- * Documentation "page" (dynamic section at /doc)
+ * Doc (documentation) page
+ * Dynamic section at /doc
  */
 
 import React, { Component } from 'react'
@@ -30,7 +31,7 @@ const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
 
 /**
- * Documentation Component
+ * <Documentation> Component
  */
 export default class Documentation extends Component {
   constructor() {
@@ -239,9 +240,6 @@ export default class Documentation extends Component {
   }
 }
 
-/**
- * Container styled <div> component
- */
 const Container = styled.div`
   display: flex;
   flex-direction: row;
@@ -264,9 +262,6 @@ const Container = styled.div`
   }
 `
 
-/**
- * Backdrop styled <div> component
- */
 const Backdrop = styled.div`
   display: none;
 
@@ -293,9 +288,6 @@ const Backdrop = styled.div`
   `};
 `
 
-/**
- * Side styled <div> component
- */
 const Side = styled.div`
   width: 280px;
   background-color: #eef4f8;
@@ -324,9 +316,6 @@ const Side = styled.div`
   `};
 `
 
-/**
- * SearchArea styled <div> component
- */
 const SearchArea = styled.div`
   height: 60px;
   display: flex;
@@ -346,9 +335,6 @@ const SearchArea = styled.div`
   }
 `
 
-/**
- * SideToggle styled <div> component
- */
 const SideToggle = styled.div`
   display: none;
   position: fixed;
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -1,10 +1,5 @@
 /* global docsearch:readonly */
 
-/*
- * Doc (documentation) page
- * Dynamic section at /doc
- */
-
 import React, { Component } from 'react'
 // components
 import Page from '../src/Page'
@@ -30,9 +25,6 @@ import { media } from '../src/styles'
 const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
 
-/**
- * <Documentation> Component
- */
 export default class Documentation extends Component {
   constructor() {
     super()
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -1,5 +1,9 @@
 /* global docsearch:readonly */
 
+/*
+ * Documentation "page" (dynamic section at /doc)
+ */
+
 import React, { Component } from 'react'
 // components
 import Page from '../src/Page'
--- a/pages/doc.js
+++ b/pages/doc.js
@@ -1,31 +1,33 @@
 /* global docsearch:readonly */
 
 import React, { Component } from 'react'
-// nextjs
-import { HeadInjector } from '../src/Documentation/HeadInjector'
 // components
+import Page from '../src/Page'
+import { HeadInjector } from '../src/Documentation/HeadInjector'
+import Hamburger from '../src/Hamburger'
+import SearchForm from '../src/SearchForm'
 import SidebarMenu from '../src/Documentation/SidebarMenu/SidebarMenu'
+import Loader from '../src/Loader/Loader'
+import Page404 from '../src/Page404'
 import Markdown from '../src/Documentation/Markdown/Markdown'
 import RightPanel from '../src/Documentation/RightPanel/RightPanel'
-import Page from '../src/Page'
-import SearchForm from '../src/SearchForm'
-import Page404 from '../src/Page404'
-import Loader from '../src/Loader/Loader'
-import Hamburger from '../src/Hamburger'
 // utils
 import fetch from 'isomorphic-fetch'
 import kebabCase from 'lodash.kebabcase'
+// constants
+import { HEADER } from '../consts'
+// sidebar data and helpers
+import sidebar, { getItemByPath } from '../src/Documentation/SidebarMenu/helper'
 // styles
 import styled from 'styled-components'
 import { media } from '../src/styles'
-// sidebar data and helpers
-import sidebar, { getItemByPath } from '../src/Documentation/SidebarMenu/helper'
-// constants
-import { HEADER } from '../consts'
 
 const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
 
+/**
+ * Documentation Component
+ */
 export default class Documentation extends Component {
   constructor() {
     super()
@@ -233,6 +235,9 @@ export default class Documentation extends Component {
   }
 }
 
+/**
+ * Container styled <div> component
+ */
 const Container = styled.div`
   display: flex;
   flex-direction: row;
@@ -254,6 +259,10 @@ const Container = styled.div`
     pointer-events: none;
   }
 `
+
+/**
+ * Backdrop styled <div> component
+ */
 const Backdrop = styled.div`
   display: none;
 
@@ -275,11 +284,14 @@ const Backdrop = styled.div`
       background-color: rgba(0, 0, 0, 0.4);
       z-index: 1;
       opacity: 1;
-      pointer-events: all;    
-    `} 
+      pointer-events: all;
+    `}
   `};
 `
 
+/**
+ * Side styled <div> component
+ */
 const Side = styled.div`
   width: 280px;
   background-color: #eef4f8;
@@ -304,10 +316,13 @@ const Side = styled.div`
       props.isOpen &&
       `
       transform: translateX(0);
-    `} 
+    `}
   `};
 `
 
+/**
+ * SearchArea styled <div> component
+ */
 const SearchArea = styled.div`
   height: 60px;
   display: flex;
@@ -327,6 +342,9 @@ const SearchArea = styled.div`
   }
 `
 
+/**
+ * SideToggle styled <div> component
+ */
 const SideToggle = styled.div`
   display: none;
   position: fixed;
--- a/pages/features.js
+++ b/pages/features.js
@@ -11,154 +11,157 @@ import Hero from '../src/Hero'
 import FeaturesHero from '../src/FeaturesHero'
 import TrySection from '../src/TrySection'
 
-export const FeaturesPage = () => (
-  <Page stickHeader={true}>
-    <HeadInjector />
-    <Hero>
-      <FeaturesHero />
-    </Hero>
-    <Container>
-      <Features>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/git-icon.svg"
-              alt="Git compatible"
-            />
-          </Icon>
-          <Name>Git-compatible</Name>
-          <Description>
-            DVC runs on top of any Git repository and is compatible with any
-            standard Git server or provider (GitHub, GitLab, etc). Data file
-            contents can be shared by network-accessible storage or any
-            supported cloud solution. DVC offers all the advantages of a
-            distributed version control system — lock-free, local branching, and
-            versioning.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/storage-icon.svg"
-              alt="Storage agnostic"
-            />
-          </Icon>
-          <Name>Storage agnostic</Name>
-          <Description>
-            Use S3, Azure, GCP, SSH, SFTP, Aliyun OSS rsync or any
-            network-attached storage to store data. The list of supported
-            protocols is constantly expanding.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/repro.svg"
-              alt="Reproducibility"
-            />
-          </Icon>
-          <Name>Reproducible</Name>
-          <Description>
-            The single &#39;dvc repro&#39; command reproduces experiments
-            end-to-end. DVC guarantees reproducibility by consistently
-            maintaining a combination of input data, configuration, and the code
-            that was initially used to run an experiment.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/branching.svg"
-              alt="Low-friction branching"
-            />
-          </Icon>
-          <Name>Low friction branching</Name>
-          <Description>
-            DVC fully supports instantaneous Git branching, even with large
-            files. Branches beautifully reflect the non-linear structure and
-            highly iterative nature of a ML process. Data is not duplicated —
-            one file version can belong to dozens of experiments. Create as many
-            experiments as you want, instantaneously switch back and forth, and
-            save a history of all attempts.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img src="/static/img/features/icons/storage-icon.svg" alt="" />
-          </Icon>
-          <Name>Metric tracking</Name>
-          <Description>
-            Metrics are first-class citizens in DVC. DVC includes a command to
-            list all branches, along with metric values, to track the progress
-            or pick the best version.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/ml-pipe.svg"
-              alt="ML pipelines framework"
-            />
-          </Icon>
-          <Name>ML pipeline framework</Name>
-          <Description>
-            DVC has a built-in way to connect ML steps into a DAG and run the
-            full pipeline end-to-end. DVC handles caching of intermediate
-            results and does not run a step again if input data or code are the
-            same.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/languages-icon.svg"
-              alt="Language & framework agnostic"
-            />
-          </Icon>
-          <Name>Language- & framework-agnostic</Name>
-          <Description>
-            No matter which programming language or libraries are in use or how
-            code is structured, reproducibility and pipelines are based on input
-            and output files. Python, R, Julia, Scala Spark, custom binary,
-            Notebooks, flatfiles/TensorFlow, PyTorch, etc. are all supported.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/cluster.svg"
-              alt="HDFS, Hive & Apache Spark"
-            />
-          </Icon>
-          <Name>HDFS, Hive & Apache Spark</Name>
-          <Description>
-            Include Spark and Hive jobs in the DVC data versioning cycle along
-            with local ML modeling steps or manage Spark and Hive jobs with DVC
-            end-to-end. Drastically decrease a feedback loop by decomposing a
-            heavy cluster job into smaller DVC pipeline steps. Iterate on the
-            steps independently with respect to dependencies.
-          </Description>
-        </Feature>
-        <Feature>
-          <Icon>
-            <img
-              src="/static/img/features/icons/failures.svg"
-              alt="Failure tracking"
-            />
-          </Icon>
-          <Name>Track failures</Name>
-          <Description>
-            Bad ideas can sometimes spark more ideas among colleagues than
-            successful ones. Retaining knowledge of failed attempts can save
-            time in the future. DVC is built to track everything in a
-            reproducible and easily accessible way.
-          </Description>
-        </Feature>
-      </Features>
-    </Container>
-    <TrySection title="Ready to give it a try?" />
-  </Page>
-)
+export default function FeaturesPage() {
+  return (
+    <Page stickHeader={true}>
+      <HeadInjector />
+      <Hero>
+        <FeaturesHero />
+      </Hero>
+      <Container>
+        <Features>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/git-icon.svg"
+                alt="Git compatible"
+              />
+            </Icon>
+            <Name>Git-compatible</Name>
+            <Description>
+              DVC runs on top of any Git repository and is compatible with any
+              standard Git server or provider (GitHub, GitLab, etc). Data file
+              contents can be shared by network-accessible storage or any
+              supported cloud solution. DVC offers all the advantages of a
+              distributed version control system — lock-free, local branching,
+              and versioning.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/storage-icon.svg"
+                alt="Storage agnostic"
+              />
+            </Icon>
+            <Name>Storage agnostic</Name>
+            <Description>
+              Use S3, Azure, GCP, SSH, SFTP, Aliyun OSS rsync or any
+              network-attached storage to store data. The list of supported
+              protocols is constantly expanding.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/repro.svg"
+                alt="Reproducibility"
+              />
+            </Icon>
+            <Name>Reproducible</Name>
+            <Description>
+              The single &#39;dvc repro&#39; command reproduces experiments
+              end-to-end. DVC guarantees reproducibility by consistently
+              maintaining a combination of input data, configuration, and the
+              code that was initially used to run an experiment.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/branching.svg"
+                alt="Low-friction branching"
+              />
+            </Icon>
+            <Name>Low friction branching</Name>
+            <Description>
+              DVC fully supports instantaneous Git branching, even with large
+              files. Branches beautifully reflect the non-linear structure and
+              highly iterative nature of a ML process. Data is not duplicated —
+              one file version can belong to dozens of experiments. Create as
+              many experiments as you want, instantaneously switch back and
+              forth, and save a history of all attempts.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img src="/static/img/features/icons/storage-icon.svg" alt="" />
+            </Icon>
+            <Name>Metric tracking</Name>
+            <Description>
+              Metrics are first-class citizens in DVC. DVC includes a command to
+              list all branches, along with metric values, to track the progress
+              or pick the best version.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/ml-pipe.svg"
+                alt="ML pipelines framework"
+              />
+            </Icon>
+            <Name>ML pipeline framework</Name>
+            <Description>
+              DVC has a built-in way to connect ML steps into a DAG and run the
+              full pipeline end-to-end. DVC handles caching of intermediate
+              results and does not run a step again if input data or code are
+              the same.
+            </Description>
+          </Feature>
+          <Feature>
+            <Icon>
+              <img
+                src="/static/img/features/icons/languages-icon.svg"
+                alt="Language & framework agnostic"
+              />
+            </Icon>
+            <Name>Language- & framework-agnostic</Name>
+            <Description>
+              No matter which programming language or libraries are in use or
+              how code is structured, reproducibility and pipelines are based on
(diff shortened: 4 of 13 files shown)


pages/_document.js
```diff
@@ -1,6 +1,7 @@
+import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
-import Router from 'next/router'
+// FIXME: import Router from 'next/router'
```
jorgeorpinel: This doesn't seem to be used in the file. Can I just delete the statement, or is there a special reason for this?

package.json
```diff
@@ -10,7 +10,7 @@
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
     "format-check": "pretty-quick --check --verbose",
-    "lint": "eslint src"
+    "lint": "eslint src pages"
```
jorgeorpinel: I still need to make one for the pre-commit hook using https://github.com/okonet/lint-staged to check only staged files, and use this one for the CI checks.

pages/_document.js
jorgeorpinel: This doesn't seem to be used in the file. Can I just delete the statement, or is there a special reason for this @iAdramelk @shcheklein?

pages/doc.js
```diff
@@ -1,31 +1,33 @@
...
-// constants
-import { HEADER } from '../consts'
 
 const ROOT_ELEMENT = 'bodybag'
 const SIDEBAR_MENU = 'sidebar-menu'
 
+/**
+ * Documentation Component
+ */
 export default class Documentation extends Component {
```
jorgeorpinel: Do you think we need some code documentation like this, at least on the main class/function component in each React file?

pages/doc.js
```diff
@@ -233,6 +235,9 @@ export default class Documentation extends Component {
   }
 }
 
+/**
+ * Container styled <div> component
+ */
 const Container = styled.div`
```
jorgeorpinel: These I now realize may be too much to document each, but what do you guys think?

pages/features.js
```diff
@@ -1,6 +1,10 @@
+/*
+ * Features page
+ */
```
jorgeorpinel: I also added a doc block before imports to each file I worked with. I think this is useful, but what do you think?

src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -19,7 +19,10 @@ const blankStyle = {}
 // Element weight for animations
 const heightMap = {}
 
-// Helper function to calculate element height with all opened children
+/**
+ * Helper function to calculate element height with all opened children
+ * @param {*} param0
```
jorgeorpinel: Not sure what values for `@param` to use here. What's a good code doc format to use in a Next/React js code base?

src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -35,6 +38,9 @@ function calculateHeight({ activePaths, path }) {
   return height
 }
 
+/**
+ * <SidebarMenuItem> (Pure) Pomponent for <SidebarMenu>
+ */
 class SidebarMenuItem extends React.PureComponent {
```
jorgeorpinel: Why is this a `PureComponent`? I'm not sure this component satisfies the guidelines in [this SO answer](https://stackoverflow.com/a/43936258/761963) (assuming it's correct).

src/PromoSection/index.js
```diff
@@ -1,36 +1,53 @@
+/*
+ * React fn component `PromoSection`
+ * FIXME: Similar to ../Subscribe and ../TrySection
```
jorgeorpinel: I found that some components are very similar and could probably be merged (using code such as Props for the subtle differences). Is such a refactoring desirable, or is it common practice in React to just copy-paste components (some times exactly the same) @iAdramelk? Thanks

pages/doc.js
algomaster99: @jorgeorpinel Why is this being added? :thinking:

pages/features.js
```diff
@@ -11,13 +15,10 @@ import Hero from '../src/Hero'
...
-  <Head>
-    <title>Features | Machine Learning Version Control System</title>
-  </Head>
-)
-
-export default () => (
+/**
+ * FeaturesPage fn component
+ */
+const FeaturesPage = () => (
```
jorgeorpinel: Is there any reason I shouldn't turn these into an `export default function...` statement instead of an arrow function that gets exported later? To be consistent with components in src/

src/Documentation/SidebarMenu/SidebarMenu.js
algomaster99:
```
@param {Object} ...
@param {string[]} activePaths - ...
@param {string} path - ...
```
You can use this as reference - https://jsdoc.app/tags-param.html

src/Documentation/SidebarMenu/SidebarMenu.js
algomaster99: @jorgeorpinel which of the guidelines does it violate? Its state anyway doesn't change.

pages/_document.js
shcheklein: probably not, let's remove and just test it

pages/doc.js
shcheklein: I would aim in keeping code as self-documenting as possible. I definitely against comments like above - in this case it's literally a repetition of the next line. It makes sense to write some high level descriptions for components, functions that are hard to understand what are they doing w/o understanding a bigger context. But it should be rare.

pages/doc.js
shcheklein: def no comments like this :)

pages/features.js
shcheklein: not useful in this way. See my comment above.

pages/features.js
```diff
@@ -11,13 +15,10 @@ import Hero from '../src/Hero'
 import FeaturesHero from '../src/FeaturesHero'
 import TrySection from '../src/TrySection'
 
-const HeadInjector = () => (
```
shcheklein: why do we move it down?

pages/index.js
```diff
@@ -12,6 +16,29 @@ import PromoSection from '../src/PromoSection'
...
+    <PromoSection />
+    <UseCases />
+    <Subscribe />
+  </Page>
+)
+
+/**
+ * HeadInjector fn component
+ */
 const HeadInjector = () => (
```
shcheklein: again, not clear why this change - I mean switching the order

pages/support.js
```diff
@@ -13,21 +16,17 @@ import Hero from '../src/Hero'
 import TrySection from '../src/TrySection'
 import Popover from '../src/Popover/Popover'
 
-const HeadInjector = () => (
-  <Head>
-    <title>Support | Machine Learning Version Control System</title>
-  </Head>
-)
-
-export default () => (
```
shcheklein: is lint complaining about export default?

src/PromoSection/index.js
shcheklein: def not a common practice to copy stuff blindly. Hard to say anything about this case specifically. Like in any language it's a balance of time, effort, need to generalize, etc.

Please, don't use FIXME, TODO, etc - if see a thing that should be fixed but in the scope of this PR, create a ticket instead.

src/TrySection/index.js
```diff
@@ -35,11 +48,50 @@ const Wrapper = styled.section`
...
+  `};
+
+  ${media.phablet`
+    display: none;
+  `};
+`
+
+/**
+ * Title styled <h3> component
+ * FIXME: Similar to the one in ../PromoSection
```
shcheklein: see a comment above, don't use FIXME

package.json
jorgeorpinel: OK, this was applied per https://github.com/iterative/dvc.org/issues/716#issuecomment-549943912.

package.json
```diff
@@ -64,12 +64,16 @@
...
     "pretty-quick": "^1.11.1"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn format-staged && yarn lint"
+      "pre-commit": "yarn format-staged && lint-staged"
     }
+  },
+  "lint-staged": {
+    "*.{js,jsx}": "eslint"
```
jorgeorpinel: Please notice `eslint` is being run on staged files only at pre-commit hook! CI will continue to eslint everything under src/ + pages/

pages/doc.js
jorgeorpinel: @algomaster99 source code documentation is a best practice. Found a nice introductory article here: https://www.valentinog.com/blog/jsdoc/

BTW I rec [JSDoc](https://jsdoc.app/) here.

@shcheklein I agree. I didn't mean to just add obvious comments like `<Documentation> Component` above. These are all placeholders for more informative descriptions. For example, here we could spec the `this.state` object. Maybe some components don't anything to spec, in which case I agree no JSDoc comment would be needed, as long as their names are descriptive.

I'll make some adjustments to my suggestions and get back to you here...

pages/_document.js
jorgeorpinel: Removed.

pages/doc.js
jorgeorpinel: Agree. Removed all the styled component doc comments.

pages/doc.js
shcheklein: @jorgeorpinel that's what I was referring to - https://blog.codinghorror.com/coding-without-comments/ (founder of SO) - it's a good practice to write code that does not need _any_ comments. Again, not saying that comments are not needed - it's just keep in mind that article before adding them.

Let's def do not merge stubs like above and in other places.

JSDoc, and other explanations are really needed when you write a library with an external API, that will be used by others. They can serve as a replacement for the API reference. Really not sure we need this.

pages/features.js
jorgeorpinel: Agree. This placeholder would be changed. A good example is what these guys use: https://docs.reactioncommerce.com/docs/jsdoc-style-guide#document-a-react-component (but not as complicated, probably only `@file` or `@module` and optional `@link`. I'll make some suggestions...

src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -19,7 +19,12 @@ const blankStyle = {}
...
 const heightMap = {}
 
-// Helper function to calculate element height with all opened children
+/**
+ * Calculate element height with all open children.
+ * @param {Object} params - Parameters needed to calculate height.
+ * @param {Object} params.activePaths - ???.
+ * @param {string} params.path - ???.
+ */
 function calculateHeight({ activePaths, path }) {
```
jorgeorpinel: Need help for names and descriptions of these params. I don't think an Object here is really needed also: Shouldn't `activePaths` and `path` be separate params?

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: Agree. Let's go with JSDoc. 🙂

I've updated this, let's continue this in https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312201679 please.

src/Documentation/SidebarMenu/SidebarMenu.js
shcheklein: let's keep that simple comment? even that one is already stretching it in my mind ) Using that heavy jsdoc style is probably to much for helpers ... and probably too much in general )

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: Please correct me if I'm wrong, still learning React, but I don't see any of them here:

✖️ State/Props should be an immutable object
✖️ State/Props should not have a hierarchy (not sure what this means actually)
✖️ You should call `forceUpdate` when data changes

But anyway, I'm not suggesting to change this. Mainly I want to know why this component is a `PureComponent`. Do you know @algomaster99?

p.s. There's another 2 `PureComponent` subclasses in the docs we could check.

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: BTW this suggestion is based off an example in https://jsdoc.app/tags-param.html#parameters-with-properties.

src/PromoSection/index.js
jorgeorpinel: Sure. I don't think it was done blindly and [React docs](https://reactjs.org/docs/rendering-elements.html) do state that "React elements are ... cheap to create" so maybe people tend interpret that as copy-paste a lot is OK?

Anyway, removed all code tags and opened #777 (lucky number 🍀).

pages/features.js
jorgeorpinel: For example I know `this` behaves differently in arrow functions, but not sure whether that or something else is a reason for this. I'd prefer to have consistency if not (switch to `export default function` which seems to be more widely used in our React code base).

cc @iAdramelk @shcheklein thanks for feedback is possible, still learning React over here.

pages/features.js
jorgeorpinel: I moved it under the main component, `FeaturesPage`, like the rest of the private components in this module. Also reorganized them into the order they appear in the JSX returned by `FeaturesPage` (and moved the `export` statement to the end of the file.)

pages/index.js
jorgeorpinel: Same as https://github.com/iterative/dvc.org/pull/776#discussion_r342929399. Just looking for a consistent organization of code in component files. Sorry, should've done this in a separate PR though... It may be a little difficult to un-do just these changes at this point, but I can manage to, if you prefer. Please lmk.

pages/support.js
jorgeorpinel: It did complain that the export needs a name, yes. So I named it `const SupportPage` and moved the `export` statement to the bottom of the file.

pages/features.js
shcheklein: kk, was asking if there was a specific reason for this (linter complaining for example). Not sure what is the best way (if it's important at all) to do this. I would check some style guides from Facebook or Google (?) or some other big techs.

pages/index.js
shcheklein: no, not a problem see my comment above. Just check if there is an actual style guide exists around organizing exports. It might that both ways are just fine.

pages/_document.js
```diff
@@ -1,11 +1,15 @@
+/*
+ * HTML document
```
shcheklein: again, please let's not do comments like this

yarn.lock
```diff
@@ -1009,6 +1009,34 @@
     "@types/istanbul-reports" "^1.1.1"
     "@types/yargs" "^13.0.0"
 
+"@nodelib/fs.scandir@2.1.3":
```
shcheklein: @jorgeorpinel what happened to this file?

src/Documentation/SidebarMenu/SidebarMenu.js
iAdramelk: @jorgeorpinel It's not exactly works like this. Basic idea is that every time parent component is rerendered, it will cause all it's child components to rerender too. Sometimes this rerender process is cheap, but in some other cases it can be quire resource consuming (for example if we call markdown component on each rerender it will try to parse/stringify document with remark every time, which is not fast).

In this case we have some listeners for the scroll in sidebar and there is some changes that fired on page scroll in `docs.js`, because of that most docs page children components are optimized to prevent unnecessary rerenders.

I have an article about the React render process details, if you are interested: https://evilmartians.com/chronicles/optimizing-react-virtual-dom-explained

pages/features.js
iAdramelk: We should change them all to the same styles, yes. There is no practical reason there to use arrow function instead of normal function.

pages/doc.js
jorgeorpinel: OK I got you now Ivan. But

> it's a good practice to write code that does not need any comments 

Unfortunately I'm seeing lots of non-readable code. In those cases adding some code docs would be a first good step towards improving the code base, don't you think? Even when really they should be refactored. For now I'll remove all my doc code suggestions from this PR since it was never its main purpose, and open another one with some more subtle comments or refactoring of unreadable code.

> the code already tells us how it works; we need the comments to tell us why it works... to write good comments you have to be a good writer

Challenge accepted 😎

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: Super informatie. Thanks @iAdramelk !

pages/features.js
jorgeorpinel: Actually yes this was prompted by the linter demanding a name for the export, same as in https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312085056. 🙂

The Airbnb React/JSX style guide seems interesting. I think in this case I'll switch to direct `export` statements ([example](https://github.com/airbnb/javascript/tree/master/react#declaration)), as also confirmed by Alexey in https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312006609 above...

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: Yes, you're right. Will revert... But this function needs refactoring for readability :P

yarn.lock
jorgeorpinel: This must have been added when I installed `lint-staged` for the change explained in https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312194024 (per https://github.com/iterative/dvc.org/issues/716#issuecomment-549939176).

jorgeorpinel: Does that answer your question?

jorgeorpinel:
> please, test carefully - mobile, Chrome, FF (from https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312088354)

Is there a good way to do this that reduces human error @iAdramelk (other than automating UI tests)? Any tips? Thanks in advance.

src/Documentation/SidebarMenu/SidebarMenu.js
shcheklein: > But this function needs refactoring for readability.

It might be indeed and it had even a comment :). But let me give you my 2c on how I usually read code:

I would never go into reading a file "as-is", meaning I would never even bother reading all these helpers (why would I care about them unless I need to fix a bug in one of them?). I usually start from some let's call it "entry-point" depending on a task you are trying to solve. And at this stage it's important to have a good project structure that somehow correlates with the real-world concepts so that it's easy to find that entry point. In most cases it means you go and see a component. I don't care about all the stuff around initially. Components are by definition declarative and you can make a lot of sense reading them w/o going too deep. 

Only after that (and having some context) you start jumping into particular functions. Since you have more context it helps to read these helpers in most cases.

So, the point here is that helpers, smaller components sometimes are not meant to be understood w/o understanding the larger context.

yarn.lock
shcheklein: looks like too many changes just to this file for a single library. Also, is there a split between dev/test libraries and production?

jorgeorpinel:
```console
$ yarn why @nodelib/fs.scandir 
yarn why v1.19.1
[1/4] 🤔  Why do we have the module "@nodelib/fs.scandir"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@nodelib/fs.scandir@2.1.3"
info Reasons this module exists
   - "lint-staged#del#globby#fast-glob#@nodelib#fs.walk" depends on it
   - Hoisted from "lint-staged#del#globby#fast-glob#@nodelib#fs.walk#@nodelib#fs.scandir"
info Disk size without dependencies: "108KB"
info Disk size with unique dependencies: "200KB"
info Disk size with transitive dependencies: "200KB"
info Number of shared dependencies: 2
✨  Done in 1.05s.
```

> is there a split between dev/test libraries and production?

Yes, `lint-staged` is in `devDependencies`.

yarn.lock
```diff
@@ -1009,6 +1009,34 @@
...
+  resolved "https://registry.yarnpkg.com/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz#3a582bdb53804c6ba6d146579c46e52130cf4a3b"
+  integrity sha512-eGmwYQn3gxo4r7jdQnkrrN6bY478C3P+a/y72IJukF8LjB6ZHeB3c+Ehacj3sYeSmUXGlnA67/PmbM9CVwL7Dw==
+  dependencies:
+    "@nodelib/fs.stat" "2.0.3"
+    run-parallel "^1.1.9"
+
+"@nodelib/fs.stat@2.0.3", "@nodelib/fs.stat@^2.0.2":
+  version "2.0.3"
+  resolved "https://registry.yarnpkg.com/@nodelib/fs.stat/-/fs.stat-2.0.3.tgz#34dc5f4cabbc720f4e60f75a747e7ecd6c175bd3"
+  integrity sha512-bQBFruR2TAwoevBEd/NWMoAAtNGzTRgdrqnYCc7dhzfoNvqPzLyqlEQnzZ3kVnNrSp25iyxE00/3h2fqGAGArA==
```
jorgeorpinel: @shcheklein 

> looks like too many changes just to this file for a single library.

I tested some more new ones. Summary:

```console
$ yarn why @nodelib/fs.stat
   - "lint-staged#del#globby#fast-glob" depends on it
$ yarn why @nodelib/fs.walk
   - "lint-staged#del#globby#fast-glob" depends on it
$ yarn why @samverschueren/stream-to-observable
   - "lint-staged#listr" depends on it
$ yarn why @types/events
   - "lint-staged#del#globby#@types#glob" depends on it
$ yarn why @types/glob
   - "lint-staged#del#globby" depends on it
$ yarn why @types/minimatch
   - "lint-staged#del#globby#@types#glob" depends on it
$ yarn why @types/node
   - "lint-staged#del#globby#@types#glob" depends on it
$ yarn why aggregate-error
   - "lint-staged#del#p-map" depends on it
$ yarn why any-observable
   - "lint-staged#listr#@samverschueren#stream-to-observable" depends on it
$ yarn why array-union
   - Hoisted from "pretty-quick#multimatch#array-union"
```

Only at `array-union` did the reason stop being `lint-staged`, but it comes from `pretty-quick` instead, which we also use. (Not sure why it wasn't there before.)

There's many other new ones below it seems, but I haven't tried them all... Will need more time to finish investigating this.

iAdramelk:
> Is there a good way to do this that reduces human error @iAdramelk (other than automating UI tests)? Any tips? Thanks in advance.

@jorgeorpinel Checklists help here. If you try to remember things from the top of your head every time, you will always forget something. Better create a checklist with all the things you want to check and follow. Every time bugs happen because you forgot something – add it to the list and test next time.

pages/doc.js
iAdramelk: My 5 cents abut comments. We need to distinguish two different types o comments: Comments describing what code doing and comments explaining why do we need to do this stuff at all and why we choose this way to do it and not another.

First type of comments are usually redundant – what code is doing should be understandable just by reading the code. By comments about why can be pretty useful, because why is not necessary obvious from code itself.

jorgeorpinel: OK, all doc comments removed now!

pages/features.js
jorgeorpinel: Removed for now.

pages/features.js
jorgeorpinel: Changing according to https://github.com/airbnb/javascript/tree/master/react#naming and [...#parentheses](https://github.com/airbnb/javascript/tree/master/react#parentheses)...

pages/features.js
jorgeorpinel: Changing same as https://github.com/iterative/dvc.org/pull/776#discussion_r343470755.

pages/index.js
jorgeorpinel: This was actually not an export, just a minor (kind of unnecessary) private component, so I moved it under the main export in the file.

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: > I would never go into reading a file "as-is", meaning I would never even bother reading all these helpers... smaller components sometimes are not meant to be understood w/o understanding the larger context

Somewhat agree. Let's try this approach and see how one would find about `function calculateHeight`: See the `SidebarMenu` Component and assume it renders the left-hand site navigation. It uses 1+ `SidebarMenuItem` like this:

https://github.com/iterative/dvc.org/blob/be3fb0ea4a0a6217bd54e01338f83f8f5bfca8d4/src/Documentation/SidebarMenu/SidebarMenu.js#L129-L138

We can then assume the `SidebarMenuItem` PureComponent renders each entry in the nav (sometimes recursively nested other `SidebarMenuItem` instances). SidebarMenuItems contain:

https://github.com/iterative/dvc.org/blob/be3fb0ea4a0a6217bd54e01338f83f8f5bfca8d4/src/Documentation/SidebarMenu/SidebarMenu.js#L62-L72

This is where it gets especially hard to read. I can guess it calculates the height of the collapsable parts (nested SidebarMenuItems). It also explains where the function argument `{ activePaths, path }` comes from.

But at this point it's basically reverse engineering.

pages/_document.js
jorgeorpinel: Removed

yarn.lock
jorgeorpinel: Continued in https://github.com/iterative/dvc.org/pull/776#pullrequestreview-312802656.

pages/doc.js
jorgeorpinel: @iAdramelk thanks for your comments. (I just noticed them now in the notification email.)

It's an interesting topic. I don't think there's no absolute rules here. Even this article (from the same author as the one Ivan shared): https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/ tries to encourage the comments that explain _why_ but ends up giving an example of a comment that explains _what_:

> You may very well know that
>
> `$string = join('',reverse(split('',$string)));`
>
> reverses your string, but how hard is it to insert
>
> `# Reverse the string`
>
> into your Perl file?

Anyway. I'm going to open a separate issue or PR to continue this exploration regarding comments and code documentation. Thanks!

src/Documentation/SidebarMenu/SidebarMenu.js
jorgeorpinel: Anyway, this is probably too deep right now. This change has been reverted.

package.json
shcheklein: 👍

pages/_document.js
```diff
@@ -1,18 +1,17 @@
...
+  `Open-source version control system for Data Science and Machine Learning ` +
+  `projects. Track your data, models, and experiments with a Git-like tool.`
 const KEYWORDS = `data version control machine learning models management`
 
 const inject = str => (
   <div className="inject" dangerouslySetInnerHTML={{ __html: str }} />
 )
 
-export default class MyDocument extends Document {
+export default class Documentation extends Document {
```
shcheklein: this is probably not a good name - sounds like a plural term, while it's actually a singe page

pages/index.js
shcheklein: another thing to keep in mind is that it complicates reading the code a bit, you don't expect something being used defined after the usage - it's not natural

pages/doc.js
iAdramelk: Well, some examples from my experience. :)

We were changing our analytics code from using google Analytics directly to using it through GTM, and after the migration, our old event started to behave differently. I spend around 10 hours reading the documentation for GA, GTM, and their counter source code, but the final fix only consisted of adding one param to the GTM function call.

Here I could just commit this change, but the next guy who sill refactor it will not have my knowledge about why this param matter and will need to spend 10 hours to find it out again. In my opinion, this is the right place for the comment and some documentation links. In the end, commit consisted of one line of code changes and around 20 lines of comments that described how GA and GTM work together and what needs to be done for them to work correctly.

Another good example of a useful comment from the article above is:

```
/* A binary search turned out to be slower than the Boyer-Moore algorithm for the data sets of interest, thus we have used the more complex, but faster method even though this problem does not at first seem amenable to a string search technique. */
```

Because this comment describes why this algorithm was used here instead of a more popular binary search.

pages/_document.js
jorgeorpinel: I thought about `DocumentationDocument` but it seemed too long. `DocumentationDoc` is confusing since both words start with "doc".

We agree that `MyDocument` is a terrible option right? Seems like it was copy-pasted from a Next tutorial and never updated 😋

jorgeorpinel:
Yes. Still investigating the yarn file changes.

pages/_document.js
shcheklein: > We agree that MyDocument is a terrible option right? Seems like it was copy-pasted from a Next tutorial and never updated

🤝 

How about Page, or SitePage - it's a general component for all pages, not only docs as far as I understand.

pages/index.js
jorgeorpinel: Yes but we define all the styled components under the main exports. Should we also place those above? Then you'd have to scroll a lot to get to the export, also complicating the reading. Not sure which option is best or whether arbitrarily putting some on top and some later was actually the best choice...

I personally like having the export all the way on top for readability, it probably be the best entry point to reading a file in most cases anyway.

shcheklein: > Should we also place those above?

Def not. `styled` - they are not so much about code even, it is a very specific thing and there are specific conventions around it. I would not use them as an example. Usually styles are located in a separate file.

> I personally like having the export all the way on top for readability, it probably be the best entry point to reading a file in most cases anyway.

it makes sense. I'm just sharing that for me it would a bit counterintuitive to expect definitions after the code. I don't have a very strong opinion though and fine with either. @iAdramelk do you have any preferences or some JS wisdom to share? :)

yarn.lock
jorgeorpinel: OK I tested down to `path-key` and actually it seems indeed everything new is somehow related to `lint-staged`. Sometimes they're also related to other previously installed packages so maybe this file just hasn't been updated in the repo for a long time. But in any case `lint-staged` does have a ridiculous amount of dependencies...

<details>

<summary>Summary of yarn.lock package reasons (`array-union` to `p-map`)</summary>

```console
$ yarn why array-union
=> Found "array-union@1.0.2"
   - Hoisted from "pretty-quick#multimatch#array-union"
   - Hoisted from "next#autodll-webpack-plugin#del#globby#array-union"
=> Found "globby#array-union@2.1.0"
info This module exists because "lint-staged#del#globby" depends on it.
$ yarn why braces
=> Found "braces@2.3.2"
   - Hoisted from "micromatch#braces"
   - Hoisted from "next#fork-ts-checker-webpack-plugin#chokidar#braces"
=> Found "lint-staged#braces@3.0.2"
   - "lint-staged#micromatch" depends on it
=> Found "fast-glob#braces@3.0.2"
   - "lint-staged#del#globby#fast-glob#micromatch" depends on it
$ yarn why clean-stack
   - "lint-staged#del#p-map#aggregate-error" depends on it
$ yarn why cli-truncate
   - "lint-staged#listr#listr-update-renderer" depends on it
$ yarn why cross-spawn
=> Found "cross-spawn@6.0.5"
   - Hoisted from "execa#cross-spawn"
   - Hoisted from "eslint#cross-spawn"
=> Found "pretty-quick#cross-spawn@5.1.0"
   - "pretty-quick#execa" depends on it
=> Found "lint-staged#cross-spawn@7.0.1"
   - "lint-staged#execa" depends on it
$ yarn why date-fns
   - "lint-staged#listr#listr-verbose-renderer" depends on it
$ yarn why dedent
   - "lint-staged" depends on it
$ yarn why del
=> Found "del@5.1.0"
   - "lint-staged" depends on it
=> Found "autodll-webpack-plugin#del@3.0.0"
info This module exists because "next#autodll-webpack-plugin" depends on it.
$ yarn why dir-glob
   - "lint-staged#del#globby" depends on it
$ yarn why elegant-spinner
   - "lint-staged#listr#listr-update-renderer" depends on it
$ yarn why execa
=> Found "execa@1.0.0"
   - Hoisted from "husky#execa"
   - Hoisted from "jest#jest-cli#@jest#core#jest-changed-files#execa"
   - Hoisted from "babel-jest#@jest#transform#jest-haste-map#sane#execa"
=> Found "pretty-quick#execa@0.8.0"
info This module exists because "pretty-quick" depends on it.
=> Found "lint-staged#execa@2.1.0"
info This module exists because "lint-staged" depends on it.
$ yarn why fast-glob
   - "lint-staged#del#globby" depends on it
$ yarn why fastq
   - "lint-staged#del#globby#fast-glob#@nodelib#fs.walk" depends on it
$ yarn why figures
=> Found "figures@2.0.0"
   - Hoisted from "eslint#inquirer#figures"
   - Hoisted from "lint-staged#listr#listr-verbose-renderer#figures"
   - Hoisted from "styled-reset#opencollective#inquirer#figures"
=> Found "listr-update-renderer#figures@1.7.0"
info This module exists because "lint-staged#listr#listr-update-renderer" depends on it.
$ yarn why fill-range
=> Found "fill-range@4.0.0"
   - "braces" depends on it
   - Hoisted from "braces#fill-range"
=> Found "lint-staged#fill-range@7.0.1"
   - "lint-staged#micromatch#braces" depends on it
=> Found "fast-glob#fill-range@7.0.1"
   - "lint-staged#del#globby#fast-glob#micromatch#braces" depends on it
$ yarn why get-own-enumerable-property-symbols
   - "lint-staged#stringify-object" depends on it
$ yarn why get-stream
=> Found "get-stream@4.1.0"
   - "execa" depends on it
=> Found "pretty-quick#get-stream@3.0.0"
   - "pretty-quick#execa" depends on it
=> Found "lint-staged#get-stream@5.1.0"
   - "lint-staged#execa" depends on it
$ yarn why globby
=> Found "globby@10.0.1"
   - "lint-staged#del" depends on it
=> Found "autodll-webpack-plugin#globby@6.1.0"
   - "next#autodll-webpack-plugin#del" depends on it
$ yarn why graceful-fs
=> Found "graceful-fs@4.2.2"
   - Hoisted from "babel-jest#@jest#transform#graceful-fs"
   - Hoisted from ...
   - Hoisted from "next#webpack#terser-webpack-plugin#cacache#move-concurrently#fs-write-stream-atomic#graceful-fs"
   - Hoisted from "babel-jest#babel-plugin-istanbul#test-exclude#read-pkg-up#read-pkg#load-json-file#graceful-fs"
=> Found "del#graceful-fs@4.2.3"
info This module exists because "lint-staged#del" depends on it.
$ yarn why ignore
=> Found "ignore@3.3.10"
   - "pretty-quick" depends on it
   - Hoisted from "pretty-quick#ignore"
=> Found "eslint#ignore@4.0.6"
info This module exists because "eslint" depends on it.
=> Found "globby#ignore@5.1.4"
info This module exists because "lint-staged#del#globby" depends on it.
$ yarn why indent-string
=> Found "indent-string@3.2.0"
   - "lint-staged#listr#listr-update-renderer" depends on it
=> Found "aggregate-error#indent-string@4.0.0"
info This module exists because "lint-staged#del#p-map#aggregate-error" depends on it.
$ yarn why is-number
=> Found "is-number@3.0.0"
   - Hoisted from "braces#fill-range#is-number"
   - Hoisted from "braces#fill-range#to-regex-range#is-number"
   - Hoisted from "braces#snapdragon#base#cache-base#has-value#has-values#is-number"
=> Found "lint-staged#is-number@7.0.0"
   - "lint-staged#micromatch#braces#fill-range#to-regex-range" depends on it
=> Found "fast-glob#is-number@7.0.0"
   - "lint-staged#del#globby#fast-glob#micromatch#braces#fill-range#to-regex-range" depends on it
$ yarn why is-obj
   - "lint-staged#stringify-object" depends on it
$ yarn why is-observable
   - "lint-staged#listr" depends on it
$ yarn why is-path-cwd
=> Found "is-path-cwd@2.2.0"
   - "lint-staged#del" depends on it
=> Found "autodll-webpack-plugin#is-path-cwd@1.0.0"
   - "next#autodll-webpack-plugin#del" depends on it
$ yarn why is-path-inside
=> Found "is-path-inside@3.0.2"
   - "lint-staged#del" depends on it
=> Found "is-path-in-cwd#is-path-inside@1.0.1"
info This module exists because "next#autodll-webpack-plugin#del#is-path-in-cwd" depends on it.
$ yarn why is-regexp
   - "lint-staged#stringify-object" depends on it
$ yarn why is-stream
=> Found "is-stream@1.1.0"
   - Hoisted from "execa#is-stream"
   - Hoisted from ...
   - Hoisted from "styled-reset#opencollective#node-fetch#is-stream"
=> Found "lint-staged#is-stream@2.0.0"
   - "lint-staged#execa" depends on it
$ yarn why lint-staged
info Has been hoisted to "lint-staged"
$ yarn why listr...
   - "lint-staged..." depends on it
$ yarn why log-symbols
=> Found "log-symbols@3.0.0"
   - "lint-staged" depends on it
=> Found "listr-update-renderer#log-symbols@1.0.2"
info This module exists because "lint-staged#listr#listr-update-renderer" depends on it.
$ yarn why merge2
   - "lint-staged#del#globby" depends on it
$ yarn why micromatch
=> Found "micromatch@3.1.10"
info Has been hoisted to "micromatch"
info Reasons this module exists
   - Hoisted from "next#webpack#micromatch"
   - Hoisted from ...
   - Hoisted from "babel-jest#@jest#transform#jest-haste-map#sane#micromatch"
=> Found "lint-staged#micromatch@4.0.2"
info This module exists because "lint-staged" depends on it.
=> Found "fast-glob#micromatch@4.0.2"
info This module exists because "lint-staged#del#globby#fast-glob" depends on it.
$ yarn why mimic-fn
=> Found "mimic-fn@2.1.0"
   - "lint-staged#execa#onetime" depends on it
=> Found "restore-cursor#mimic-fn@1.2.0"
   - "eslint#inquirer#cli-cursor#restore-cursor#onetime" depends on it
$ yarn why npm-run-path
=> Found "npm-run-path@2.0.2"
   - Hoisted from "execa#npm-run-path"
   - Hoisted from "pretty-quick#execa#npm-run-path"
=> Found "lint-staged#npm-run-path@3.1.0"
   - "lint-staged#execa" depends on it
$ yarn why onetime
info Reasons this module exists
   - "lint-staged#execa" depends on it
=> Found "restore-cursor#onetime@2.0.1"
info This module exists because "eslint#inquirer#cli-cursor#restore-cursor" depends on it.
$ yarn why p-finally
=> Found "p-finally@1.0.0"
   - Hoisted from "execa#p-finally"
   - Hoisted from "pretty-quick#execa#p-finally"
=> Found "lint-staged#p-finally@2.0.1"
   - "lint-staged#execa" depends on it
$ yarn why p-map
=> Found "p-map@2.1.0"
   - "lint-staged#listr" depends on it
=> Found "del#p-map@3.0.0"
info This module exists because "lint-staged#del" depends on it.
=> Found "autodll-webpack-plugin#p-map@1.2.0"
   - "next#autodll-webpack-plugin#del" depends on it
$ yarn why p-map
=> Found "p-map@2.1.0"
   - "lint-staged#listr" depends on it
=> Found "del#p-map@3.0.0"
info This module exists because "lint-staged#del" depends on it.
=> Found "autodll-webpack-plugin#p-map@1.2.0"
   - "next#autodll-webpack-plugin#del" depends on it
```

</details>

shcheklein: 👍

jorgeorpinel:
@shcheklein on the yarn.lock fie changes, I think `lint-staged` simply has all these deps (see https://github.com/iterative/dvc.org/pull/776#discussion_r343826354) but also many of them should have been in there before from things like `pretty-quick` and `eslint` but weren't. Probably these other libs were put manually in the package.json file instead of using `yarn install` like I did now for `lint-staged`, so the lock file was outdated.

As for whether this will cause Yarn to install all of this on prod, it shouldn't:

> Yarn will not install any package listed in `devDependencies` if the `NODE_ENV` environment variable is set to production.
(See https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false)

shcheklein:
> @jorgeorpinel Checklists help here. If you try to remember things from the top of your head every time, you will always forget something. Better create a checklist with all the things you want to check and follow. Every time bugs happen because you forgot something – add it to the list and test next time.

Good idea. Would be great to create a wiki page on Github with a checklist like this.

pages/index.js
iAdramelk: I agree that functions and methods should be defined before they are used; it helps readability. But I'm conflicted about styled-components. On the one hand, they will be in the way if we place them above the code as they didn't have any logic in them. On the other hand, I often feel confused, then I see something then looks like a react component, but I didn't see import or definition. Then I scroll, and I understand that this is just a styled component. Maybe some prefix to the styled component name could help there like starting their names with capital S or Styled.

shcheklein:
@jorgeorpinel hey, it looks like it's almost done! only one component to come with a good name for (the most difficult problem in software engineering :)), and may be some reordering. Do you need our help or more reviews, comments?

jorgeorpinel:
Hey @shcheklein I got this. I was just distacted with other stuff but getting back to it now. Yes, almost done. I'll get back to you ASAP.

pages/index.js
jorgeorpinel: OK I reordered the pages/ files again to have local components defined before the export.

jorgeorpinel: > I often feel confused, then I see something then looks like a react component, but I didn't see import or definition.... Maybe some prefix to the styled component name could help there like starting their names with capital S or Styled.

Well for example, if you see `<Container>` in a component's render() method, and didn't see the definition, you can assume its defined later (else React will throw an error at runtime) – so it's probably a styled component. I don't oppose revising names though! I would recommend just we can just add the name of the styled HTML tag, in this case `<ContainerDiv>`. Should we create a separate issue for this?

pages/doc.js
jorgeorpinel: (This conversation is resolved but I'll leave it "unresolved" so we can later find this text if we want to, as it was a good discussion.)

jorgeorpinel:
This is done in terms of code changes and reviews. I also did basic sanity checks and things seem to work in general but not a detailed testing of every web page that these changes may impact, on every relevant browser (Chrome, FF, mobile...) Still working on that one...

pages/_document.js
jorgeorpinel: Oh, I almost forget about this one. Ah OK good point. Changed to `Page`.

jorgeorpinel:
>  please, test carefully - mobile, Chrome, FF

OK @shcheklein, I checked the all the pages of https://dvc-org-pr-776.herokuapp.com/ on Chrome, FF, and Safari, both in their desktop and simulated mobile modes. They look the same as on dvc.org and there's no Console errors.

p.s. I do get some console errors in my local site though, but they're also on the master branch. Opened #791 for this.

pages/index.js
jorgeorpinel: Extracted to #792.

PR #784, opened 2019-11-12
Data sharing scenarios

### LINKED ISSUE #54: add "How to store data on your own server" tutorial/guide
Probably needs a better name.

### LINKED ISSUE #103: guide: using NFS as a remote storage

### LINKED ISSUE #194: cases: training on a remote machine (SSH)
for example, `ssh` to a remote machine, `git clone / pull` to get your code, then `dvc pull` to download your dataset, and `dvc repro` or `dvc run` to start training your models

### LINKED ISSUE #429: cases: update Shared Dev Server with an optimization link

### LINKED ISSUE #648: user-guide: document data sharing with unsupported storage types
Since a remote cache is basically a mirror of the local cache, then any tool (not just DVC) can be used to synchronize them.

This can be another workaround for storage types that are not yet supported by DVC (besides mounting the remote storage on a local directory). Maybe this is not as efficient as DVC, and some things are not going to work (for example `dvc status --cloud`), but it still should allow people to share data.

Should we create a page (on UC or UG) that documents this workaround?

Related issues:
- https://github.com/iterative/dvc/issues/1940
- https://github.com/iterative/dvc/issues/1183
- https://github.com/iterative/dvc/issues/1696

dashohoxha:
Explain the different data sharing scenarios:
- Remote DVC Storage
- Shared Development Server
- Mounted DVC Storage
- Synced DVC Storage

Deprecates:
- ~~https://dvc.org/doc/use-cases/sharing-data-and-model-files~~
- ~~https://dvc.org/doc/use-cases/shared-development-server~~
- https://github.com/iterative/dvc.org/pull/455

Related: https://github.com/iterative/dvc.org/issues/103, https://github.com/iterative/dvc.org/issues/648#issuecomment-535885723, https://github.com/iterative/dvc.org/issues/54, https://github.com/iterative/dvc.org/pull/455, https://github.com/iterative/dvc.org/issues/194

Fix #54, address #103, fix #194, fix #648


jorgeorpinel:
Looking really cool! Probably Ivan will review this? Feel free to request my review otherwise. Just a Q, how does this address #429 **use-case: update shared dev server case with an optimization link?**

shcheklein:
@dashohoxha looks great! will be definitely reviewing this. It will take a while though :)

dashohoxha:
> @dashohoxha looks great! will be definitely reviewing this. It will take a while though :)

Please take your time. It took me a long time to develop it, so it would seem fair if you just did a quick review in 5 min :)

dashohoxha:
> How does this fix #429 use-case: update shared dev server case with an optimization link?

@jorgeorpinel I have moved the "shared dev server case" to the User Guide, under the section "Data Sharing", along with other data sharing possibilities (or scenarios). I have also revised it a bit to use a local DVC storage instead of a shared cache. In this context, the suggested optimization is using a deduplicating filesystem.

So, my assumption is that that issue is oudated/deprecated and maybe should be closed (as soon as this PR is merged).

> Feel free to request my review otherwise.

I am sure that it is not perfect and there are things that can be improved, but please let's merge it first and do any improvements later (in other PRs).

shcheklein:
@dashohoxha @jorgeorpinel I still think Jorge should review it. We all need to see what's happening, we can fix some small issues right away, and I would love to know his opinion.

jorgeorpinel:
> I have moved the "shared dev server case" to the User Guide

OK. Need to make a more careful review but for one one comment, why is the use case still there if you're suggesting to move it to the user guide? Same for /doc/use-cases/sharing-data-and-model-files

> I have also revised it a bit to use a local DVC storage instead of a shared cache. In this context, the suggested optimization is using a deduplicating filesystem... So, my assumption is that that issue is oudated/deprecated.

I see. In that case case we could just close it now, please comment on there for Ivan to confirm. Tbh I don't think this change addresses it. Removing "fix #429" from here, for now.

> please let's merge it first and do any improvements later

Review first, work on changes until we agree its mergeable, merge, then continue to improve, for sure. If that's what you meant, I agree.

shcheklein:
> please let's merge it first and do any improvements later

my 2c on this. The problems with this that it never happens. Making small improvements is considered "unsexy" for some reason (I have no idea why) and people just move on to some new stuff. Just one example - we've created a few tickets after we merged the Install page and they are still not resolved, even those that have p-1 priority, like the first get started page.

I'm not saying that we should be polishing to perfection, but we def should review, come us with a reasonable changes, create follow up tickets and agreed on resolving them, then merge.

dashohoxha:
> OK. Need to make a more careful review but for now one comment: Why is the use case still there if you're suggesting to move it to the user guide? Same for /doc/use-cases/sharing-data-and-model-files.

Because I would prefer to remove them on the next PR (after this one is merged). Removing them maybe needs also any URL fixing on other pages where they might be used. Maybe any redirections too, I am not sure.

jorgeorpinel:
That would produce duplicate content in between and as Ivan mentioned from the previous experience the priorities are hard to manage this way. I think the full move and redirections should be included along with all this new content, which is great BTW.

dashohoxha:
> I think the full move including redirections (glad you remembered) should be included along with all this new content

I would prefer to do it separately, but if Ivan also thinks that it should be included in this PR, I'd be glad to do it.

jorgeorpinel:
I understand. And it would make the PR easier to review, but from my PoV it just kicks the can down the road. Really appreciate your help with this, Dashamir! 😄

dashohoxha:
@jorgeorpinel I created an issue as a follow-up of this PR: https://github.com/iterative/dvc.org/issues/793

src/Documentation/sidebar.json
```diff
@@ -119,6 +119,29 @@
...
+        "label": "Data Sharing",
+        "slug": "data-sharing",
+        "source": "data-sharing/index.md",
+        "children": [
+          {
+            "label": "Remote DVC Storage",
+            "slug": "remote-storage"
+          },
+          {
+            "label": "Shared Development Server",
```
shcheklein: I'm not sure this one fits into this section. From the technical perspective it's the same as mounted DVC storage, right?

shcheklein: in general last three look superficial, too specific. We should try and better generalize them (it's all about where cache is located, whether it is shared via `dvc cache dir` and whether remote is used/needed). It would be impossible to writer an article per each case and it's not needed. I would try to generalize this properly.

static/docs/user-guide/data-sharing/remote-storage.md
```diff
@@ -0,0 +1,193 @@
+# Sharing Data Through a Remote DVC Storage
```
shcheklein: it is strange to see that it contains only examples, no explanation of what's happening whatsoever. I would expect it too explain remotes way better - this is a primary purpose of this.

SSH example is too complicated - 

- the same dir for git and dvc on remote is too specific, very uncommon and distracts a lot
- name of the remote storage should not use cache in it

static/docs/user-guide/data-sharing/shared-server.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Create accounts for each user and add them to groups for accessing the Git
+  repository and the DVC storage.
+- Create a bare git repository (for example on `/var/local/data/project.git/`)
+  and an empty directory for the DVC storage (for example on
+  `/var/local/data/project.cache/`).
+
+- Grant users read/write access to these directories (through the groups).
+
+### Configure local DVC storage
```
shcheklein: this is not how it should be done. The whole purpose of the shared dev server is to share cache not remote. The difference is that cache support different cache link types that saves us from copying files. In that scenario it's pretty common to actually use S3 as a remote storage.

static/docs/user-guide/data-sharing/mounted-storage.md
```diff
@@ -0,0 +1,115 @@
+# Sharing Data Through a Mounted DVC Storage
```
shcheklein: again, see my other comments. There are at least two possibilities - shared cache or shared remote. In case of NAS it's actually beneficial to share cache (and use some regular cloud remote to still do backups).

src/Documentation/sidebar.json
dashohoxha: > From the technical perspective it's the same as mounted DVC storage, right?

It seems to me that it is not the same as "mounted DVC storage", because in the case of mounted storage the project and the DVC storage are located on different file systems, so it can't be as efficient as the case of "shared development server".

The difference is subtle but the consequences are considerable (especially when we work with big data).

dashohoxha: > in general last three look superficial, too specific.

The first one seems the same to me as the last three (in terms of being detailed or superficial).

> We should try and better generalize them (it's all about where cache is located, whether it is shared via `dvc cache dir` and whether remote is used/needed). It would be impossible to writer an article per each case and it's not needed. I would try to generalize this properly.

I agree that the differences between them are subtle, but I think that they should be discussed as separate cases. At least I don't see how to generalize and merge them into a single page.

static/docs/user-guide/data-sharing/remote-storage.md
dashohoxha: > it is strange to see that it contains only examples, no explanation of what's happening whatsoever

It is a modified version of this page: https://dvc.org/doc/use-cases/sharing-data-and-model-files
That one does not have much explanations either and things are explained mostly by the example.

Actually I don't find it feasible to explain a solution without using at least a few DVC commands, and for those commands to make sense they have to be used in the context of an example. So, the description mainly describes the situation, and the solution is described by the examples. The hope is that once the reader has understood the solution he can generalize and adopt it for his own case.

> I would expect it too explain remotes way better - this is a primary purpose of this.

I am planning to explain the details of the remotes (and their types) on another section. This section is about data sharing scenarios, so let's just refer to the remote details, but not include them here.

> the same dir for git and dvc on remote is too specific, very uncommon and distracts a lot

Yes, the Git repository is usually located on GitHub. But this is just an example, an assumption to keep things simple and interactive.

> name of the remote storage should not use cache in it

I tried to keep the analogy with Git. In Git a central bare repository is usually named `project.git`. So, a central DVC storage/cache is name `project.cache`.

static/docs/user-guide/data-sharing/shared-server.md
dashohoxha: > this is not how it should be done. The whole purpose of the shared dev server is to share cache not remote. The difference is that cache support different cache link types that saves us from copying files.

But in a shared development server we can use a deduplicating filesystem (which supports xref copy), so we don't need to resort to cache link types.

Besides this, using a shared cache requires special care when working with commands that modify the cache (for example `dvc gc`). When we use a local DVC storage we don't need to do that, and we just use the normal workflow with `dvc push` and `dvc pull`.

> In that scenario it's pretty common to actually use S3 as a remote storage.

It can still be used to backup and restore the local DVC storage.

static/docs/user-guide/data-sharing/mounted-storage.md
dashohoxha: > In case of NAS it's actually beneficial to share cache (and use some regular cloud remote to still do backups).

Sharing cache in the case of a NAS may cause problems when we try to use `dvc gc`. I remember seeing some discussions about this.

shcheklein: Yes, and we even introduced a special flag - to pass multiple projects at once to `dvc gc`. Gc in DVC is a big pain still but it does not change the fact I mentioned above.

static/docs/user-guide/data-sharing/shared-server.md
shcheklein: probably I'm missing something here, I'm not sure. When you do remote DVC won't be using any linking between project's cache and remote right now. Because that's the purpose of having cache - cache is about using reflinks or other link types to avoid duplication and make checkouts fast.

jorgeorpinel:
Checking in on this. Any news?

BTW I re-deployed the review app to check out the diagrams. Cool! What kind of UML diagrams are they? (class, sequence, etc.) Are all of them done with https://www.umlet.com/? http://www.umletino.com/? How do you export .uxf ?

Visible in:
- https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/remote-storage#ssh-remote-example
- https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/shared-server#shared-server-example
- https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/mounted-storage#sshfs-mounted-storage-example
- https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/synced-storage#synced-storage-example

static/img/user-guide/data-sharing/shared-server.uxf
```diff
@@ -0,0 +1,297 @@
...
+// Drag a whole relation at its central square icon
+//
+// Press Ctrl+C to copy the whole diagram to the system clipboard (then just paste it to, eg, Word)
+// Edit the files in the "palettes" directory to create your own element palettes
+//
+// Select "Custom Elements &gt; New..." to create new element types
+//////////////////////////////////////////////////////////////////////////////////////////////
+
+
+// This text will be stored with each diagram;  use it for notes.</help_text>
```
jorgeorpinel: Probably don't need these comments in any of the uxf files?

static/docs/user-guide/data-sharing/mounted-storage.md
dashohoxha: > we even introduced a special flag - to pass multiple projects at once to `dvc gc`

The option `-p, --projects` of [`dvc gc`](https://dvc.org/doc/command-reference/gc) gets a path to a project (at least this is how I understand the man page, I have never tried it).

In the case of a NAS mounted storage I assume that the collaborating projects are located on different machines, isn't it? So, the option `-p, --projects` cannot be used in this case.

static/docs/user-guide/data-sharing/shared-server.md
dashohoxha: > probably I'm missing something here, I'm not sure. When you do remote DVC won't be using any linking between project's cache and remote right now. Because that's the purpose of having cache - cache is about using reflinks or other link types to avoid duplication and make checkouts fast.

If we use a deduplicating filesystem (which supports reflink copy) copying a cached file to the storage (with `dvc push`) will be done instantaneously and it will take 0 (zero) additional space on disk.

Please check again the interactive tutorial: https://katacoda.com/dvc/courses/examples/shared-server It does mention that `dvc push` and `dvc fetch` are very efficient (and it uses a 10G data file as an example).

dashohoxha:
> Cool! What kind of UML diagrams are they? (class, sequence, etc.) Are all of them done with https://www.umlet.com/? http://www.umletino.com/? How do you export .uxf ?

They are not UML diagrams. They might be called deployment diagrams, but they are not UML deployment diagrams. I used Umlet because I find it handy and efficient (and it is also open source).
The format `.uxf` is the basic format of Umlet, and the images (`.png`) are exported from it.

static/docs/user-guide/data-sharing/mounted-storage.md
shcheklein: There are probably ways still to run GC - clone projects to a single machine. It's not ideal (all about GC not) but it's a maintenance operations vs day to day workflow that is being optimized with links if you share cache vs sharing remote directly.

Also, I think it is the same problem with other your cases and in one of them it's about people sharing the same machine, right?

static/img/user-guide/data-sharing/shared-server.uxf
dashohoxha: > Probably don't need these comments in any of the uxf files?

Most probably they are not needed, but they don't harm. UXF files can be edited with Umlet, they are not supposed to be read or edited as a source code.

static/docs/user-guide/data-sharing/shared-server.md
shcheklein: It's a very strong assumption - that all the people share the same mount point. It's in an optimization that we have indeed that works only on certain file systems.

Also, some feedback regarding the tutorial. It's feels super advanced. DVC part starts only in the middle and before that a lot of magic that I think can be simplified.

static/docs/user-guide/data-sharing/mounted-storage.md
dashohoxha: > It's not ideal (all about GC not) but it's a maintenance operations vs day to day workflow that is being optimized with links if you share cache vs sharing remote directly.

I agree that if we disregard any issues about GC, then sharing the cache would indeed optimize data sharing for the NAS case and would require less storage (by using symlinks).
I may extend the tutorial and the user-guide page to explain this optimization as well.

> Also, I think it is the same problem with other your cases and in one of them it's about people sharing the same machine, right?

I don't think it applies to the other cases because it does not bring any improvements (in terms of optimization: faster and smaller disk space).

static/docs/user-guide/data-sharing/shared-server.md
dashohoxha: > It's a very strong assumption - that all the people share the same mount point. It's in an optimization that we have indeed that works only on certain file systems.

It's an example that shows how to make things efficient even if the filesystem does not support deduplication (aka reflink copy). It is much more easy to assume that the shared server supports deduplication natively, so no extra configurations or tricks are needed to make things efficient. 

> Also, some feedback regarding the tutorial. It's feels super advanced. DVC part starts only in the middle and before that a lot of magic that I think can be simplified.

Do you think that it would become simpler if the magic that sets up the server and user account is played automatically? This way the people that might be interested in them (linux beginners) would still see them, but the professionals (who already know how to do it) can skip them quickly.

jorgeorpinel:
OK, thanks for the info. Looks like a handy tool. Ivan is the diagram king so I guess I'm sure he'll have something to say about them eventually.

static/docs/user-guide/data-sharing/shared-server.md
shcheklein: > It's an example that shows how to make things efficient even if the filesystem does not support deduplication (aka reflink copy)

I'm not sure how would optimize anything if the filesystem does not support deduplication (aka reflink copy). You would need to share the cache - and it's a completely different setup, right?

>  It is much more easy to assume that the shared server supports deduplication natively, so no extra configurations or tricks are needed to make things efficient.

it's easier, but it's very very specific. Especially the assumption that people work on the same machine + the same FS is way too strong.

> Do you think that it would become simpler if the magic that sets up the server and user account is played automatically? 

I prefer to focus on the actual stuff and hide all auxiliary stuff. There are few mechanisms to hide that - expandable section, extra section at the end with a link to it, FAQ at the end, separate setup section, etc. Not sure what is the best mechanism here. 

> beginners

I'm not sure this is the right audience for these guides in the first place. At least not now. I'm not saying that docs should be complicated, dry and do not provide any help at all. On contrary, even professional would appreciate if we care about them and put some links, materials about setup. It's more about focus, structure, etc.

static/docs/user-guide/data-sharing/mounted-storage.md
shcheklein: > I may extend the tutorial and the user-guide page to explain this optimization as well.

I first understand the options in terms of organizing data, understand which of them are more general then others, then would try to come up with a couple of sections that explain them in a general way. And by general I mean concepts like -  cache is shared or not? people use a single machine or not? etc

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
+# Sharing Data Through a Mounted Cache
```
shcheklein: this one replace the shared server use case? I think the explanation in the use case article is better in a sense that it gives the context why a single shared server is used in the first place

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
...
+appear automatically to the cache of all the others. As a result, no `dvc push`
+and `dvc pull` are needed to share the data, just a `dvc checkout` will be
+sufficient.
+
+> ** ❗ Caution:** Deleting data from the cache will also make it disappear from
+> the cache of the other users. So, be careful with the command `dvc gc` (which
+> cleans obsolete data from the cache) and consult the other users of the
+> project before using this command.
+
+The optimization in data management comes from using the _symlink_ cache type.
```
shcheklein: it can be hardlinks and reflinks

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
...
+You can find more details about it in the page of
+[Large Dataset Optimization](https://dvc.org/doc/user-guide/large-dataset-optimization).
+
+## Mounted Cache Example
```
shcheklein: I still think that explaining SSHFS here is way too much - too specific, takes too much time and distracts from the point even if expandable sections are being used

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
...
+In this example we will see how to share data with the help of a cache directory
+that is mounted through SSHFS. We are using a SSHFS example because it is easy
+to network-mount a directory with SSHFS. However once you understand how it
+works, it should be easy to implement it for other types of network-mounted
+storages (like NFS, Samba, etc.).
+
+> For more detailed instructions check out this
+> [interactive example](https://katacoda.com/dvc/courses/examples/mounted-cache).
+
+<p align="center">
```
shcheklein: that's not what we usually do I believe (not sure why) to center the images ... is there a special Markdown syntax to center an image? In general it's better to keep as less HTML as possible.

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
...
+remote directory to `.dvc/cache/` of the project like this:
+
+```dvc
+$ mkdir -p ~/project/.dvc/cache
+$ sshfs \
+      dvc-server:/srv/project.cache/ \
+      ~/project/.dvc/cache/
+```
+
+### Optimize data management
```
shcheklein: protected mode is missed here

static/docs/user-guide/data-sharing/mounted-cache.md
```diff
@@ -0,0 +1,144 @@
...
+This configuration is the same for all the users, so we can add it to Git in
+order to share it with the other users:
+
+```dvc
+$ git add .dvc/config
+$ git commit -m "Use symlinks if reflinks are not available"
+$ git push
+```
+
+### Sharing data
```
shcheklein: I think the primary point of using shared cache usually is optimize resources - not sharing data - in sense to pass some data to another user

static/docs/user-guide/data-sharing/mounted-storage.md
shcheklein: I don't see how my initial concern is resolved or addressed here. Please 🙏 , don't resolve them on your own - it makes it extremely hard to do reviews (check and follow up the previously raised concerns).

shcheklein: To be precise - I don't see much value in three (two?) sections that explain different variations of the mounted remote. And to be even more precise - I haven't see a mounted share remote case. The only benefit I see - unsupported storage type. I would just create a How to or FAQ or something with one-two paragraphs explanation on options - mount, use rsync/rclone, etc.

static/docs/user-guide/data-sharing/remote-storage.md
shcheklein: > I am planning to explain the details of the remotes (and their types) on another section. This section is about data sharing scenarios, so let's just refer to the remote details, but not include them here.

I don't think we need both that was point. I'm confused why do we need both. I think the remote section is enough. There should be some "DVC workflow" section that explains from a high level perspective the workflow.

shcheklein:
@jorgeorpinel I would say I have mixed feeling about the use cases like versioning data and sharing data, so I understand where the intention to deprecate them comes from :)

I agree that the way they are written not the are too low level, do not serve as a good "landing page" as @jorgeorpinel  called them. But I think we should be improving them in that direction (the same way we do with data registry) and write more use cases like this if possible. This section should answer the question of an infrastructure, DevOps, MLOps folks on how DVC can help them. How can it be used, in what high level scenarios. And it's even okay if they overlap in this case. I totally can see that data versioning is to some extent about data management as data registry is. Or data sharing overlaps potentially with data registry in a sense that data registry can be used to share data in a little bit different scenarios.

To reiterate on the Use Cases vs User Guide. 

Users Guide is like a car owner's manual - a lot of details, how tos, etc.
Use Cases would be like explaining how that car can be used (potentially referring a lot to the owner's manual to cover the technical details) - driving from home to work, renting, taxi, etc

Hope that makes sense :)

jorgeorpinel:
OK, so here are the decisions I propose:

> the way they are written not the are too low level, do not serve as a good "landing page"... But I think we should be improving them (and write more use cases)

I agree. So the first decision 🥇 is that we're not removing the use-cases section. ✔️ 

> it's even okay if they overlap in this case. 

Again agree. If we can come up with a good balance on how much overlap is good to have between use cases and these deeper new docs in /doc/user-guide/data-sharing/, then I think we can keep both (having in mind we will improve the use cases in the future to be more high level).

I think that this specifically means that 🥈 the new docs should try to not repeat the more high level parts of the use cases. ✔️
I just hope we do review this now, to prevent merging the PR with lots of repetition vs. existing use cases. I could focus on reviewing this aspect if you agree, @shcheklein. (So far I haven't looked at the detailed changes here.)

> > The first one is too generic to be a use-case. Data Versioning is rather a core feature of DVC...
>
> data versioning is to some extent about data management as data registry is.

>> The second page is about the case when DVC is used only as a dataset catalog...
>
> data sharing overlaps potentially with data registry in a sense that data registry can be used to share data

That level of overlap between use cases is fine, I agree. And also that data versioning is a core feature of DVC. Data Registry is our newest use case and we're already working on its 2nd iteration so its possibly the best use case we have.

So 🥉 let's keep these 2 use cases where they are as well. ✔️

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > this one replace the shared server use case?

Sorry, you got this one wrong. This one is replacing the shared server use case: https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/shared-server

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > it can be hardlinks and reflinks

No, if the cache is mounted from a NAS, neither hardlinks nor reflinks work (because they don't work across different filesystems).

I think this is a good example of why these cases need to be explained separately and they cannot be consolidated further. If you can be confused on such cases, then normal users might be much more confused than you. So, it can't be simplified further.

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > I still think that explaining SSHFS here is way too much - too specific, takes too much time and distracts from the point even if expandable sections are being used

This looks like your personal opinion. I don't think the same. I think that without some specific details and a concrete example, the explanation would too abstract and much more difficult to understand.

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > that's not what we usually do I believe (not sure why) to center the images ... 

You are using very large images, which are scaled down by the browser and so they are displayed on the center of the page. I am not sure whether this is a good or bad practice...

> is there a special Markdown syntax to center an image?

As far as I know there isn't.

> In general it's better to keep as less HTML as possible.

I certainly agree with this.

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > protected mode is missed here

No it is not missing, maybe you have missed it: https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/mounted-cache#optimize-data-management

static/docs/user-guide/data-sharing/mounted-cache.md
dashohoxha: > I think the primary point of using shared cache usually is optimize resources - not sharing data - in sense to pass some data to another user

Sorry, I don't understand what you are trying to say here. Can you please explain further what is the problem that you perceive?

dashohoxha:
> It's too complicated. It puts too much focus on some strange cases like mounted remotes. Not clear why do we need remotes here if we have a new PR about them. Examples with SSHFS are too specific and complicated.

I have tried to answer some of the concerns on the other comments. User Guide pages are supposed to be details oriented (as opposed to Use Cases, which are supposed to be high level -- at least this is what I gather from the other discussions). In this sense I don't think these pages are too complicated.

I am afraid that you are not getting my vision, and I don't get yours either. At this point I am stuck as I don't see how this PR can be improved further.

> @jorgeorpinel:  So ￼ let's keep these 2 use cases where they are as well.

If we agree on this, then it seems to me that no major changes need to be added on this PR.

> @jorgeorpinel: I could focus on reviewing this aspect if you agree

I would be more than happy if @jorgeorpinel takes over making further improvements to these pages.
But I would be even more happy if we merged this PR first and make further improvements on another PR.

static/docs/user-guide/data-sharing/remote-storage.md
dashohoxha: > I don't think we need both

I don't think that these UG pages:
https://dvc-org-pr-807.herokuapp.com/doc/user-guide/external-data  (from another PR)
can be merged with the pages of this PR. I think they should be separate sections.

static/docs/user-guide/data-sharing/mounted-storage.md
dashohoxha: > I don't see how my initial concern is resolved or addressed here.

I have added this page:
- https://dvc-org-pr-784.herokuapp.com/doc/user-guide/data-sharing/mounted-cache

and this interactive example:
- https://katacoda.com/dvc/courses/examples/mounted-cache

that explain the case of mounted cache, which is more efficient if we share data through a NAS (with caveat of being careful with the command `dvc gc`).

static/docs/user-guide/data-sharing/mounted-cache.md
shcheklein: @dashohoxha then, I'm even more confused since the `shared-server.md` is still part of this PR.

static/docs/user-guide/data-sharing/mounted-cache.md
shcheklein: If users use the same mount point for their workspaces the way your example is written DVC might pick hardlinks or  reflinks. Thinks a single EC2 box with a single huge SSD xfs on it. And it's being used by multiple people.

static/docs/user-guide/data-sharing/mounted-cache.md
shcheklein: > I think that without some specific details and a concrete example

I don't even disagree with this. It's a matter of number of details and also a matter of using 1% concrete example vs using 80% example. SSHFS is way too specific and we have too many details. It's like 50% of the page is about setting up this.

static/docs/user-guide/data-sharing/mounted-cache.md
shcheklein: 👍 probably worth mentioning why is it needed. Was mostly reading the text that's probably missed this.

static/docs/user-guide/data-sharing/mounted-cache.md
shcheklein: Kk. Let me step back a bit then. May be I'm missing the whole point of this PR still. In the index page for this new section, you write:

```
Like Git, DVC facilitates collaboration and data sharing on a distributed
environment. It makes it easy to consistently get all your data files and
directories to any machine, along with the source code.
```

And I understand this more or less. And the primary mechanism for this is just regular remotes (that's why I still confused considering that you have second PR that is about external data but also includes remotes. So both overlap in this sense but not quite focus on remotes management.

So, could you reiterate (and fix index.md?) to communicate what is the goal here? What part of the DVC workflow for end users do we cover? How does it relate to the second PR?

static/docs/user-guide/data-sharing/mounted-storage.md
shcheklein: It does not answer my question unless again I'm missing the whole point of this PR. Could you please elaborate on this:

> To be precise - I don't see much value in three (two?) sections that explain different variations of the mounted remote. And to be even more precise - I haven't see a mounted share remote case. The only benefit I see - unsupported storage type. I would just create a How to or FAQ or something with one-two paragraphs explanation on options - mount, use rsync/rclone, etc.

shcheklein:
@dashohoxha see my comments on some discussions on this ticket.

> I am afraid that you are not getting my vision, and I don't get yours either.

yep. 👍 And that prevents it from being merged or improved. I asked about the vision in some comments above. Please share your thoughts and we'll see if I can understand the purpose and end goal of this PR.

shcheklein:
Even though some parts could be useful, I'm closing this :( :

* it feels that it was created without understanding of the use cases which translates in an overcomplicated structure, some artificial examples, and some non relevant statements
* it looks like author does not like to continue iterating with us on this
* two things above mean for us that it would take less time to re-write certain relevant materials from scratch then addressing them here

Issue #670, opened 2019-10-03
Create an interactive button to run google colabs, Katacode , kaggle kernels, github button for Get-started tutorial


kurianbenoy:
This is to create an easier run time environment for our tutorials like how pytorch docs tutorial has: 

https://pytorch.org/tutorials/beginner/blitz/tensor_tutorial.html
![tutorial](https://user-images.githubusercontent.com/24592806/66139916-d8aaf280-e61e-11e9-9e3b-a1827f054a67.png)

shcheklein:
We can reuse Katacoda's scenarios @dashohoxha is building for this.

dashohoxha:
Yes, I am working on an interactive version of Get-Started: https://github.com/iterative/dvc.org/issues/664
I don't know whether it will be linked with a button or with a simple link, but this does not seem crucial to me.

dashohoxha:
> I am working on an interactive version of Get-Started

Here it is: https://katacoda.com/dvc/courses/get-started

Actually I did it as a single tutorial first: https://katacoda.com/dvc/courses/tutorials/get-started
But it seemed to be too long, taking more than 60min to finish, so I split it in smaller lessons.

I am not sure yet how to best link it from the website.

@kurianbenoy Please feel free to give a try to a Colab version of the get-started, some people may prefer it better (refering to this: https://discordapp.com/channels/485586884165107732/565759186693128202/628298295063740436)

shcheklein:
@dashohoxha this is great! I'll take a look and review the scenarios, also we should start integrating it with the engine (add a button). Any suggestions on how can it look like? I would prefer it to stand out as much as possible, a full text message something like "Run an interactive version". Would be great to emphasize it somehow when you hover code blocks?

@algomaster99 @jorgeorpinel any suggestions on this ^^

jorgeorpinel:
Super cool to have all those Katacoda scenarios! I'd love to help reviewing them too. Are they created in a way that we can easily give feedback on @dashohoxha? For example in a Git repo perhaps? (Not familiar with Katacoda.) Otherwise how can we leave comments on specific commands in certain scenarios? Thanks

> Google Colabs, Katacoda, Kaggle kernels

So we're discarding Kaggle kernels for now right?

> feel free to give a try to a Colab version of the get-started

Er... Not sure we want so many versions of the same material, already it may be a challenge to maintaining both our dvc.org docs and Katacoda updated and synchronized. Perhaps we should also discard that option for now?

[...]

PR #786, opened 2019-11-14
Add references to the interactive lessons and examples

dashohoxha:


static/docs/command-reference/fetch.md
```diff
@@ -120,6 +120,8 @@ pipeline stages, as well as a few Git tags, such as our
 [get started example repo](https://github.com/iterative/example-get-started).
 Then we can see what happens with `dvc fetch` as we switch from tag to tag.
 
+> [Try these examples interactively](https://katacoda.com/dvc/courses/examples/fetch).
```
shcheklein: let's use ▶ instead of `>` to make it more visible.

dashohoxha:
> We should not forget remove/update these references after we get #670 solved.

It seems to me that the current references are visible and good enough.
So, maybe we should close #670.

static/docs/get-started/index.md
```diff
@@ -24,3 +24,6 @@ motivation and what's happening internally.
 You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
 Also, if DVC is not installed, please follow these [instructions](/doc/install)
 to do so.
+
+▶ For a quick start you can also
+**[`try these hands-on interactive lessons`](https://katacoda.com/dvc/courses/get-started/initialize)**
```
shcheklein: I think too much emphasis here.

static/docs/get-started/index.md
```diff
@@ -24,3 +24,6 @@ motivation and what's happening internally.
 You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
 Also, if DVC is not installed, please follow these [instructions](/doc/install)
 to do so.
+
+▶ For a quick start you can also
```
shcheklein: that's sad, I though they would be rendered as here https://unicode.org/emoji/charts/full-emoji-list.html

not sure why are they not

static/docs/command-reference/fetch.md
```diff
@@ -120,6 +120,8 @@ pipeline stages, as well as a few Git tags, such as our
 [get started example repo](https://github.com/iterative/example-get-started).
 Then we can see what happens with `dvc fetch` as we switch from tag to tag.
 
+#### [`▶ Try examples interactively`](https://katacoda.com/dvc/courses/examples/fetch)
```
shcheklein: I still don't get why do make emphasis on fetch at all. It's one the most rare commands. Can we just delete/hide for now the interactive tutorial? remove it from the list of interactive tutorials? Otherwise when you google `dvc docs` now fetch is shown as one of the big sections

static/docs/get-started/index.md
shcheklein: let's make it a small black triangle or figure out how to render it properly

dashohoxha: Let's not try to make it perfect since it is only temporary and it will be replaced by the button (#670).

static/docs/command-reference/fetch.md
dashohoxha: > when you google `dvc docs` now fetch is shown as one of the big sections

Certainly there must be some other reasons, not the link to the interactive tutorial, since this change is not published yet.

dashohoxha:
Closing in favor of #824

PR #790, opened 2019-11-14
chore: 🤖 run lint-staged on pre-commit hook

### LINKED ISSUE #788: lint: switch pretty-quick for lint-staged in pre-commit hook?
> Extacted from https://github.com/iterative/dvc.org/issues/716#issuecomment-550666828

We're now using `lint-staged` for ESLint in the pre-commit hook (see #776) and it seems it also supports Prettier so it would be cleaner to only use this tool and stop using `pretty-quick`.

Please refer to the `lint-staged` [README](https://github.com/okonet/lint-staged#automatically-fix-code-style-with-prettier-for-javascript--flow-typescript-markdown-or-html) and figure out the necessary changes to package.json by testing locally. (And remove `pretty-quick` dependency, update yarn.lock.)

theashraf:
- remove `pretty-quick` from `pre-commit` hooks and replace it with `lint-staged`
- configure `lint-staged` linters for `.js, .json, .md` files
- add format script to format the whole codebase (to be used when needed)

closes #788


theashraf:
The CI failed cause I've removed the `format-check` script from `package.json`,
I think it's fine to just run `Eslint` in the CI pipelines since it's already configured to check prettier rules as well via the `eslint-plugin-prettier`.
any suggestions?

jorgeorpinel:
Oh. Hmmm shoot. I guess we should've waited until https://github.com/iterative/dvc.org/pull/776/files was merged because it introduces `lint-staged` without `.lintstagedrc.json` (configured in package.json) and changes the `lint` script name to `lint-check`.

- [ ] But OK let's keep working on it and I'll take care of resolving conflicts later...

.lintstagedrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "*.js": ["eslint --fix", "jest --findRelatedTests", "git add"],
+  "*.{json,md}": ["prettier --write", "git add"]
```
jorgeorpinel: - We want ` "*.{js,jsx}"` for ESLint and ` "*.{js,json,md}"` for Prettier, please.
- What is `jest --findRelatedTests` for?
- What does `git add` do? We don't want it to stage automatic formatting at pre-commit hook, just to fail committing so the user can check what Prettier did (with `git diff`) and stage themselves. Thanks!

package.json
```diff
@@ -8,8 +8,7 @@
     "build": "next build",
     "test": "jest",
     "start": "NODE_ENV=production node server.js",
-    "format-staged": "pretty-quick --staged --no-restage --bail",
-    "format-check": "pretty-quick --check --verbose",
+    "format": "prettier --write \"**/*.+(js|md|json)\"",
```
jorgeorpinel: - OK to merge those into a single `format` script, but it be used at CI, so it should not `--write`, just check and fail if it finds a problem. (Equivalent at `--check --verbose` for the former `format-check": "pretty-quick` script. Thanks.
- Please use `(js|json|md)` (alphabetical order).

jorgeorpinel:
@theashraf first of all thanks again for your contribution!

> The CI failed cause I've removed the `format-check` script...

On this, please notice https://github.com/iterative/dvc.org/pull/790#pullrequestreview-317185114, and then we just need to remove that step from .circleci/config.yml for the CI to work again 🙂

jorgeorpinel:
Ah wait no. I need to think about removing the `format-check` script actually... I'll get back to you.

package.json
jorgeorpinel: Actually... I'm not sure we can remove `format-check`. Let me think...

jorgeorpinel:
> I think it's fine to just run `Eslint` in the CI pipelines...

OK yeah all good. We do want to run Prettier separately as a CI check, but we simply need to update the `format-check` run stage in .circleci/config.yml to use `format` now. Please. Thanks!

package.json
jorgeorpinel: Yes, never mind. It's OK, we just need to use `format` in the CI config now. See https://github.com/iterative/dvc.org/pull/790#issuecomment-554034488 🙂

jorgeorpinel: I addressed this in my recent merge from master.

jorgeorpinel: Nevermind, part of this still requires addressing.

jorgeorpinel:
> we simply need to update the format-check run stage in .circleci/config.yml to use format now.

I've done this in my recent merge from master.

.lintstagedrc.json
jorgeorpinel: I removed `git add`.

theashraf: > * We want ` "*.{js,jsx}"` for ESLint and ` "*.{js,json,md}"` for Prettier, please.

I agree with linting `.jsx` files using `eslint` but I don't think we need to run `prettier` on `.js,jsx` files since `eslint --fix` will do the same as `prettier --write` 

> * What is `jest --findRelatedTests` for?

It looks up every source and test file that requires your files

package.json
theashraf: the `format` script is intended to format the whole codebase, I think if we want to check the formatting it's better to rename the script to `check-format` instead of just `format`.
So I suggest to rename the `format` script to `check-format` and run prettier with the parameters you've mentioned.

.lintstagedrc.json
jorgeorpinel: > `eslint --fix` will do the same as `prettier --write` 

Hmmm. You did mention about [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier) earlier but I didn't quite understand we were doing this until now that you bring it up. Cool! So I guess just add jsx here. Maybe check that other changes I made are consistent with this fact also?

jorgeorpinel: > It looks up every source and test file that requires your files

https://jestjs.io/docs/en/cli#findrelatedtests-spaceseparatedlistofsourcefiles reads "Find and run the tests that cover a space..." Are you suggesting we replace the `test` script from package.json (and corresponding step in .circleci/config.yml) with `link-staged`?

jorgeorpinel: Wait no. I keep getting confused between pre-commit hook and CI checks. Never mind... I guess this would just add test enforcement for changed files at `git commit`? If so, let's not do this now. It will make committing too slow.

package.json
jorgeorpinel: Hmmm good point. But we also have the `lint` script, what would we rename it to? `check-lint`? That doesn't sound as nice as `check-format`. I think just go back to `format-check` and `lint-check` respectively, is best, please.

jorgeorpinel: p.s. Actually, we don't intend to have scripts to actually format the code. We have information on how to do that in https://dvc.org/doc/user-guide/contributing/docs#doc-style-guidelines-and-tips-for-java-script-and-markdown.

.lintstagedrc.json
theashraf: > I guess this would just add test enforcement for changed files at `git commit`? If so, let's not do this now. It will make committing too slow.

it will not slow down the commiting cause `jest --findRelatedTests` will run only the tests on the changed staged files not the whole test files, also it makes sense to test before commiting to ship the code with confidence and avoiding catching failed tests in CI instead you can catch it locally before

jorgeorpinel: It's already kind of slow. Running Jest will definitely add more wait time for commits, which may push people to start skipping the hook. Tests will be run at CI anyway so there's no need for this now and it's beyond the scope of the issue at hand. But I'll mention this to the team and check what they think. Let's not add now though, thanks!

package.json
theashraf: I don't see there is a reason to rename the `lint` script since its name gives the exact intent of it, I'll only rename the `format` script to `check-format` and replace the `--write` arg with `--check`

> p.s. Actually, we don't intend to have scripts to actually format the code. We have information on how to do that in https://dvc.org/doc/user-guide/contributing/docs#doc-style-guidelines-and-tips-for-java-script-and-markdown.

Good Point!

.lintstagedrc.json
theashraf: Agreed !!

jorgeorpinel: cc @shcheklein @algomaster99 @iAdramelk what do you think about adding `jest` to our pre-commit hook?

.lintstagedrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "*.{js,jsx}": ["eslint --fix", "git add"],
+  "*.{json,md}": ["prettier --write", "git add"]
```
jorgeorpinel: Hm I see you added `git add` again. I'll remove this again.

static/docs/user-guide/large-dataset-optimization.md
```diff
@@ -117,8 +117,8 @@ $ dvc config cache.protected true
 
 Setting `cache.protected` is important with `hardlink` and/or `symlink` cache
 file link types. Please refer to the
-[Update a Tracked File](/doc/user-guide/updating-tracked-files) on how to
-manage tracked files under these cache configurations.
+[Update a Tracked File](/doc/user-guide/updating-tracked-files) on how to manage
+tracked files under these cache configurations.
```
jorgeorpinel: Thanks!

.lintstagedrc.json
```diff
@@ -0,0 +1,4 @@
...
+}
```
jorgeorpinel: I tried changing a js file so a code line goes over 80 chars and Prettier did reformat it, however the commit didn't fail. We need to figure out a way to stop the commit if any tool makes changes to the files ([`--bail` behavior](https://github.com/azz/pretty-quick#--bail) of `pretty-quick`). (Would also be good to keep [`--verbose`](https://github.com/azz/pretty-quick#--verbose) too).

jorgeorpinel: Wait. This code is outdated, let me comment again with the most current code....

.lintstagedrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "*.{js,jsx}": ["eslint --fix"],
+  "*.{json,md}": ["prettier --write"]
+}
```
jorgeorpinel: I tried changing a js file so a code line goes over 80 chars and Prettier did reformat it, however the commit didn't fail. We need to figure out a way to stop the commit if any tool makes changes to the files ([`--bail` behavior](https://github.com/azz/pretty-quick#--bail) of `pretty-quick`). (Would also be good to keep [`--verbose`](https://github.com/azz/pretty-quick#--verbose) too).

> From https://github.com/iterative/dvc.org/pull/790#discussion_r346641520

theashraf: The commit didn't fail because the format issues have been fixed automatically using `eslint --fix`. Do you want the developer to fix the format issues manually?

jorgeorpinel: They were fixed but not staged because we don't want automatic staging of fixes. (At least that's how it worked with `pretty-quick`.) This allows the developer to review the automatic formatting with `git diff` before adding them manually and trying to commit again. I wonder if this will be possible with `lint-staged`... Maybe this is helpful: https://medium.com/hackernoon/announcing-lint-staged-with-support-for-partially-staged-files-abc24a40d3ff ?

Re `--verbose`: perhaps `lint-staged --debug`? See https://github.com/okonet/lint-staged#command-line-flags

Also, probably simpler to put this [in package.json](https://github.com/okonet/lint-staged#packagejson-example) instead of a standalone config file?

jorgeorpinel: Thanks for moving to package.json!

> I wonder if this will be possible...

Possibly also related/useful: okonet/lint-staged/issues/550, https://github.com/okonet/lint-staged/issues/594#issuecomment-474304561. I also opened okonet/lint-staged/issues/730, let's see what they think.

iAdramelk:
@jorgeorpinel I think that using `git add` here is ok. `prettier` shouldn't try to change anything that will break the code, and I don't see a lot of meaning for the developer to manually check `prettier` fixes, because we are forcing them anyway. So even if they didn't like the changes, it's not like that they can do something about it.

.lintstagedrc.json
jorgeorpinel: Thanks for the feedback Alexey. I agree.

jorgeorpinel:
Thanks Alexey. Agree with you again. Let's just see if `lint-staged` provides a way to "bail", like `pretty-quick` did. (Waiting for okonet/lint-staged/issues/730.) Although, yet another option is simply to keep `pretty-quick`... I'l check with Ivan directly about this.

.lintstagedrc.json
jorgeorpinel: Resolving in favor of https://github.com/iterative/dvc.org/pull/790#issuecomment-554811425.

jorgeorpinel:
UPDATE: No updates 🙁 

I just pinged the `lint-staged` guys again (https://github.com/okonet/lint-staged/issues/730#issuecomment-559260063). If there's no answer this week we'll just make an executive decision here.

jorgeorpinel:
UDPATE: The `lint-staged` guys are actually implementing this 😬 See okonet/lint-staged/pull/745

jorgeorpinel:
So unfortunately the `lint-staged` guys ended up changing their minds about implementing this so let's keep `pretty-quick --bail` for now? I think it's important being able to know what you're committing without automatic `git add`. We can keep this PR alive (will get stale but oh well) to check for future versions of `lint-staged` though, as they hinted maybe they'll try to implement in a future version.

jorgeorpinel:
p.s. apologies to @theashraf who was great at submitting this PR. This is very rare that we don't end up merging them, especially when they come from an open issue (which I'm closing for now, BTW).

Abdelrahman, if you're interested in any other issue please feel free to check with me or anyone on the team (#dev-docs channel of our [chat](https://dvc.org/chat) before having a go just so we double check the issue validity. Again, this situation is very rare, in fact it's the first time it happens that I'm aware. But still, sorry!

shcheklein:
Closing this, @jorgeorpinel , right?

jorgeorpinel:
OK in that case I'll reopen the issue and relabel it. I guess that makes more sense.

PR #799, opened 2019-11-18
Remove confusing deletion instruction on get-started

skshetry:
The `Retrieve Data` section already has a code snippet to delete `data/data.xml`, but the notice seems to have remained from the old days when there didn't use to be deletion code snippet. I found this very confusing, and maybe others did as well. Again, this is redundant, anyway.


shcheklein:
thanks @skshetry !!

PR #800, opened 2019-11-18
Improve re-importing example in `dvc import`

### LINKED ISSUE #796: improve dvc import fixed revisions example
[This section](https://dvc.org/doc/command-reference/import#example-fixed-revisions-re-importing) is confusing, probably because it misses a step. These are the steps I expect:
- importing a particular version (this one is missing)
- updating
- rebinding a revision

That would read as a story unfolding.

The other thing is that we start from some yaml, not from a command, which is also confusing. Should we include those yamls at all? Or include but hide in a bubble.

jorgeorpinel:
and improve refs among this example and `dvc update` cmd ref.

Closes #796

### DIFF
--- a/static/docs/command-reference/import.md
+++ b/static/docs/command-reference/import.md
@@ -121,8 +121,9 @@ outs:
 ```
 
 Several of the values above are pulled from the original stage file
-`model.pkl.dvc` in the external DVC repository. `url` and `rev_lock` subfields
-under `repo` are used to save the origin and version of the dependency.
+`model.pkl.dvc` in the external DVC repository. The `url` and `rev_lock`
+subfields under `repo` are used to save the origin and version of the
+dependency.
 
 ## Example: fixed revisions & re-importing
 
@@ -154,8 +155,6 @@ deps:
 If the
 [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
 moves (e.g. branches), you may use `dvc update` to bring the data up to date.
-This will update `rev_lock` in the import stage (DVC-file).
-
 However, for typically static references (e.g. tags), or for SHA commits, in
 order to actually "update" an import, it's necessary to **re-import the data**
 instead, by using `dvc import` again without or with a different `--rev`. This
--- a/static/docs/command-reference/import.md
+++ b/static/docs/command-reference/import.md
@@ -148,12 +148,18 @@ deps:
       rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
 ```
 
-If the Git revision tends to move, i.e. branches, this doesn't have much of an
-effect on the import/update workflow. However, for typically static references
-i.e. tags, or for SHA commits, `dvc update` will not have any effect on the
-import. In this cases, in order to actually "update" an import, it's necessary
-to **re-import the data** instead, by using `dvc import` again without or with a
-different `--rev`. For example:
+If the
+[Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+tends to move, i.e. branches, this doesn't have much of an effect on the
+import/update workflow. However, for typically static references i.e. tags, or
+for SHA commits, `dvc update` will not have any effect on the import.
+
+In this cases, in order to actually "update" an import, it's necessary to
+**re-import the data** instead, by using `dvc import` again without or with a
+different `--rev`. This will overwrite the import stage (DVC-file), either
+removing or replacing the `rev` field, respectively. This can produce an import
+stage that is able to be updated normally with `dvc update` going forward. For
+example:
 
 ```dvc
 $ dvc import --rev master \
@@ -161,9 +167,9 @@ $ dvc import --rev master \
              use-cases/cats-dogs
 ```
 
-This will overwrite the import stage (DVC-file) either removing or replacing the
-`rev` field. This can produce an import stage that is able to be updated
-normally with `dvc update` going forward.
+> In the above example, the value for `rev` in the new import stage will be
+> `master`, which happens to be the default branch in this Git repository, so
+> the command is equivalent to not using `--rev` at all.
 
 ## Example: Data registry
 
--- a/static/docs/command-reference/import.md
+++ b/static/docs/command-reference/import.md
@@ -154,7 +154,7 @@ deps:
 
 If the
 [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
-moves (e.g. branches), you may use `dvc update` to bring the data up to date.
+moves (e.g. a branch), you may use `dvc update` to bring the data up to date.
 However, for typically static references (e.g. tags), or for SHA commits, in
 order to actually "update" an import, it's necessary to **re-import the data**
 instead, by using `dvc import` again without or with a different `--rev`. This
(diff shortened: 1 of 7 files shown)


static/docs/command-reference/import.md
```diff
@@ -121,14 +121,23 @@ outs:
...
 
 Several of the values above are pulled from the original stage file
-`model.pkl.dvc` in the external DVC repository. `url` and `rev_lock` fields are
-used to specify the origin and version of the dependency.
+`model.pkl.dvc` in the external DVC repository. `url` and `rev_lock` subfields
+under `repo` are used to save the origin and version of the dependency.
 
 ## Example: fixed revisions & re-importing
 
+```dvc
```
Suor: I would add an enter phrase above: 

"To import a specific revision of an artifact you may use:"

static/docs/command-reference/update.md
```diff
@@ -32,9 +32,11 @@ Another detail to note is that when the `--rev` (revision) option of
...
 kind of
 [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References) this
-is, for example a branch or a tag. For static refs such as tags (unless manually
-updated), or for SHA commits, `dvc update` will not have any effect on the
-import.
+is, for example a branch or a tag. For static references such as tags (unless
+manually updated), or for SHA commits, `dvc update` will not have any effect on
+the import. Refer to the
+[re-importing example](/doc/command-reference/import#example-fixed-revisions-re-importing)
+to learn how to "update" fixed-revision imports.
```
Suor: I think we can assume that people understand what tags and branches are and they should get how updating and switching differ. So this paragraph might be shortened.

Or am I wrong? Do people understand switching is a kind of update? If they don't then us mixing these two might confuse instead of helping.

static/docs/get-started/import-data.md
```diff
@@ -67,9 +67,9 @@ outs:
     persist: false
 ```
 
-The `url` subfield points to the source project, while `rev_lock` lets DVC know
-which Git repository version did the data come from. Note that `dvc update`
-updates the `rev_lock` value.
+The `url` subfield under `repo` points to the source project, while `rev_lock`
+lets DVC know which Git repository version did the data come from. Note that
+`dvc update` updates the `rev_lock` value.
```
Suor: Do we need to say that `dvc update` updates `rev_lock`? Or anything about `rev_lock` for that matter? Maybe leave it as an implementation detail.

We really don't expect people to edit that manually, while editing `rev` might make some sense.

static/docs/command-reference/import.md
```diff
@@ -24,7 +24,7 @@ DVC provides an easy way to reuse datasets, intermediate results, ML models, or
 other files and directories tracked in another <abbr>DVC repository</abbr> into
 the workspace. The `dvc import` command downloads such a <abbr>data
 artifact</abbr> in a way that it is tracked with DVC, so it can be updated when
-the data source changes.
```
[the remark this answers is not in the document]
jorgeorpinel: Parentheses are used for parts of a sentence of paragraph that could be omitted without affecting the meaning. This cmd ref is about import, not update, so it makes sense to keep them. We also use this form "(See [link].)" in many places in the docs already, so for consistency I'd also prefer to keep it.

static/docs/command-reference/import.md
jorgeorpinel: Agree. Added.

static/docs/command-reference/update.md
jorgeorpinel: I don't understand your comment. No switching mentioned in this paragraph (lines 31-39).

static/docs/get-started/import-data.md
jorgeorpinel: This IS an implementation detail. You're commenting on a paragraph inside a `<details>` section in https://dvc.org/doc/get-started/import-data.

static/docs/command-reference/import.md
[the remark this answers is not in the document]
jorgeorpinel: I assume you don't have a strong opinion here, so resolving this one.

static/docs/command-reference/update.md
jorgeorpinel: Perhaps the text was a little bit confusing? I re-phrased this, please review again.

static/docs/get-started/import-data.md
jorgeorpinel: > We really don't expect people to edit that manually

As for this: I agree. But I don't think that's the only reason we provide implementation details in our docs.

jorgeorpinel: This comment did made me realize we needed to update other docs that explain about `rev` and `rev_lock` though, so thanks! See my latest commit, 89026ced.

static/docs/command-reference/import.md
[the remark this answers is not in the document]
Suor: I don't insist, it's your call here. I think that this inhibits readability though.

static/docs/command-reference/import.md
```diff
@@ -139,22 +152,25 @@ deps:
...
 
-If the Git revision moves, such as a branch, this doesn't have much of an effect
-on the import/update workflow. However, for static refs such as tags (unless
-manually updated), or for SHA commits, `dvc update` will not have any effect on
-the import. In this cases, in order to actually "update" an import, it's
-necessary to **re-import the data** instead, by using `dvc import` again without
-or with a different `--rev`. For example:
+If the
+[Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+moves (e.g. branches), you may use `dvc update` to bring the data up to date.
```
Suor: -> e.g. a/some branch (to make it play along with singular the Git revision and moves) 

Also, if you want to say it might be a branch you probably meant "a Git revision" above. Not sure I am right here, use your own sense of the language :)

static/docs/command-reference/import.md
```diff
@@ -139,22 +152,25 @@ deps:
...
+However, for typically static references (e.g. tags), or for SHA commits, in
```
Suor: (e.g. tags), -> , like tags,

Again, not insisting. It simply sounds more fluent to me this way.

static/docs/command-reference/import.md
jorgeorpinel: Agree, probably both singular and plural are valid here but singular sounds better. Updated.

> you probably meant "a Git revision" above.

I meant **the** specific revision used in the command example above (`--rev cats-dogs-v1`).

static/docs/command-reference/import.md
jorgeorpinel: I like "(e.g. tags)" because we have "(e.g. a branch)" in the previous sentence. And commit hashes (SHA-1) are not [references](https://git-scm.com/book/en/v2/Git-Internals-Git-References) technically speaking. But thanks!

shcheklein:
@jorgeorpinel @Suor thanks!

static/docs/command-reference/import.md
Suor: But the for revision above is a tag, and you say it might be a branch in the same sentence.

jorgeorpinel: Correct. Still, it refers to **the** revision used in the `dvc import --rev ...` command, whatever it happens to be.

Suor: If you change `the` to `a` then it won't be refering to anything above. That would be just some revision, this is what I meant from the start.

jorgeorpinel: I would then need to say "a revision used with the `--rev` option" which is much longer. Seems unnecessary since "the" is also correct.

PR #801, opened 2019-11-19
Generalize, review and fix meta fields

shcheklein:


consts/index.js
```diff
@@ -1 +1,10 @@
 export const HEADER = 'header'
+
+export const META_BASE_TITLE = 'Data Version Control · DVC'
```
jorgeorpinel: This seems to be causing build error https://reactjs.org/docs/error-decoder.html/?invariant=31&args[]=object%20with%20keys%20%7BMETA_BASE_TITLE%7D&args[]=.

See https://dashboard.heroku.com/apps/dvc-org-pr-801/activity/builds/c32b9281-33f1-4ae5-b4a3-2168468b6633. Probably easier to debug locally though (with the React Developer Tools browser extension).

jorgeorpinel: Also, I wonder whether we should just put this file in the root dir or in src/ as consts.js

shcheklein: It was caused by a wrong JSX expression, addressed in https://github.com/iterative/dvc.org/pull/801/commits/18b22324e2a02a5fe774f0fb5b33a1d4493b62e2

> Also, I wonder whether we should just put this file in the root dir or in src/ as consts.js

👍 done in https://github.com/iterative/dvc.org/pull/801/commits/690ec582e32a8f85e2e29528c63a9f3111bfd13d

pages/features.js
```diff
@@ -10,10 +10,11 @@ import Page from '../src/Page'
 import Hero from '../src/Hero'
 import FeaturesHero from '../src/FeaturesHero'
 import TrySection from '../src/TrySection'
+import { META_BASE_TITLE } from '../src/consts'
 
 const HeadInjector = () => (
   <Head>
-    <title>Features | Machine Learning Version Control System</title>
+    <title>Features | {META_BASE_TITLE}</title>
```
jorgeorpinel: We're removing "Machine Learning" from all these titles? I guess it makes sense, just noticing.

shcheklein:
> Looks good. One comment left, but not a blocker. Mergeable.

Yep. After discussion with @dmpetrov and @SvetaGr decided to simplify and optimize titles for SEO.

Thanks, @jorgeorpinel for reviewing this :)

PR #805, opened 2019-11-20
use-cases: improvements to data-registry case per Alex' review

### LINKED ISSUE #795: use-cases: improve Data Registry case
I took a look at the new Data Registry page. Congrats @jorgeorpinel on compiling it and please don't be angry if you'll get a bit more than you expected :) 

Here go some considerations big and small as they occur in the text. 

---

> Furthermore, the version of the data file imported to B can be an older iteration than what's currently used in A.

- [x] Time is somewhat non-linear when branches are involved, so it might not be correct to talk about "older iteration". Should we just say that B might refer to any version of an artifact from A.

> Keeping this in mind, we could build a

could -> can

> This way we would have a repository that has all the metadata and change history for the project's data.

- [x] I don't understand this sentence. "we would ... change history", what does it mean? Why do we want to change any history? What does "history of projects data" mean?

Also, why do we push data-registry from the start? Can't one repo simply depend on the other repo? Registry is only one scenario. The majority of advantages listed do not require data registry. I understand that this use case is named "Data Registry", but maybe we can lay some trail to it? 

> features such as change history

- [x] what does this mean "change history feature"?

> # Example

- [ ] This story about wget and the past brings confusion. May we just state the problem (the current state of it) and solve it? The whole story then starts from the middle, we present commands that do something else then expected and then say about that in some buble. I needed to reread it several times to get what is this about. 

> A dataset we use for several of our examples and tutorials is one containing

- [x] ... tutorials contains ...

> We partitioned the dataset in two ... 

- [x] We split the dataset into two ... (See https://github.com/iterative/dvc.org/issues/795#issuecomment-556114361)

jorgeorpinel:
Fix #795

### DIFF
--- a/static/docs/use-cases/data-registry.md
+++ b/static/docs/use-cases/data-registry.md
@@ -45,28 +45,29 @@ Advantages of using a DVC **data registry** project:
 ## Example
 
 A dataset we commonly use for several of our examples and tutorials contains
-2800 images of cats and dogs. We split it in two for our
+2800 images of cats and dogs, which was split it in two for our
 [Versioning Tutorial](/doc/tutorials/versioning). Originally, the parts were
-backed up on a storage server, and downloaded with `wget`. This setup was then
-revised to download the dataset sing `dvc get` instead, so we created the
-[dataset-registry](https://github.com/iterative/dataset-registry)) repository, a
-<abbr>DVC project</abbr> hosted on GitHub, to version the dataset (see its
+backed up on a storage server, and downloaded with
+[`wget`](https://www.gnu.org/software/wget/). This was then revised in order to
+download the parts with `dvc get` instead, so we created the
+[dataset-registry](https://github.com/iterative/dataset-registry)
+<abbr>project</abbr> to version the dataset (in the
 [`tutorial/ver`](https://github.com/iterative/dataset-registry/tree/master/tutorial/ver)
 directory).
 
-However, there are a few problems with the way that dataset is structured. Most
-importantly, this single dataset is tracked by 2 different
-[DVC-files](/doc/user-guide/dvc-file-format), instead of 2 versions of the same
-one, which would better reflect the intentions of this dataset... Fortunately,
-we have also prepared an improved alternative in the
+However, there's a few problems with the way that dataset is versioned. Most
+importantly, this split dataset is tracked by 2 different
+[DVC-files](/doc/user-guide/dvc-file-format) (one for each part), instead of 2
+versions of a single DVC-file. An initial version could have the first part
+only, while an update would have the entire, unified dataset. Fortunately, we
+have also prepared this improved alternative in the
 [`use-cases/`](https://github.com/iterative/dataset-registry/tree/master/use-cases)
 directory of the same <abbr>DVC repository</abbr>.
 
-To create a
-[first version](https://github.com/iterative/dataset-registry/tree/cats-dogs-v1/use-cases)
+To create the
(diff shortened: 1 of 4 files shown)


static/docs/use-cases/data-registry.md
```diff
@@ -10,27 +10,26 @@ different projects (similar to package management systems, but for data), DVC
...
-the same way you do with code; and we don't need ad-hoc conventions to store
-different data versions. Other projects can share the data in the registry by
-downloading (`dvc get`) or importing (`dvc import`) them for use in different
-data processes.
+have a repository with all the metadata and history of changes in the project's
+data. We could see who updated what, and when, use pull requests to update data
+(the same way we do with code), and avoid ad-hoc conventions to store different
+data versions. This is what we call a data registry. Other projects can share
+datasets in a registry by downloading (`dvc get`) or importing (`dvc import`)
+them for use in different data processes.
```
jorgeorpinel: @Suor here's the change related to "can" vs. "could".

static/docs/use-cases/data-registry.md
```diff
@@ -7,30 +7,25 @@ tracking of datasets and any other <abbr>data artifacts</abbr>.
...
-also includes the `dvc get`, `dvc import`, and `dvc update` commands. For
-example, project A may use a data file to begin its data
-[pipeline](/doc/command-reference/pipeline), but project B also requires this
-same file; Instead of
-[adding it](/doc/command-reference/add#example-single-file) it to both projects,
-B can simply import it from A. Furthermore, the version of the data file
-imported to B can be an older iteration than what's currently used in A.
+also includes the `dvc get`, `dvc import`, and `dvc update` commands. This means
+that a project can depend on data from an external <abbr>DVC project</abbr>, but
+chaining several projects this way can easily become messy...
```
jorgeorpinel: Simplified project inter-dependency mention and introduced problem (for data registry solution).

shcheklein: is it still WIP?

jorgeorpinel: Nope. Please review.

shcheklein: I think `...` is a bit confusing. Why `.` is not enough here?

jorgeorpinel: Good point. Reviewed usage of `...` in 55ab7571 now.

static/docs/use-cases/data-registry.md
```diff
@@ -7,30 +7,25 @@ tracking of datasets and any other <abbr>data artifacts</abbr>.
...
-also includes the `dvc get`, `dvc import`, and `dvc update` commands. For
-example, project A may use a data file to begin its data
-[pipeline](/doc/command-reference/pipeline), but project B also requires this
-same file; Instead of
-[adding it](/doc/command-reference/add#example-single-file) it to both projects,
-B can simply import it from A. Furthermore, the version of the data file
-imported to B can be an older iteration than what's currently used in A.
+also includes the `dvc get`, `dvc import`, and `dvc update` commands. This means
+that a project can depend on data from an external <abbr>DVC project</abbr>, but
+chaining several projects this way can easily become messy.
```
shcheklein: see my comment here https://github.com/iterative/dvc.org/issues/795#issuecomment-557943717

jorgeorpinel: OK I just removed the remark about getting messy for now. It will be introduced again in the next iteration but about having no system to version data.

jorgeorpinel:
OK. But I'm moving this it to the `use-cases/data-registry` branch. Unfortunately, had to create a new PR for this: #818.

PR #810, opened 2019-11-22
docs: update notes about cache dir

### LINKED ISSUE #806: tutorials: confusing code sample and note about .dvc/cache dir
I was following the tutorial at [https://dvc.org/doc/tutorials/deep/preparation](https://dvc.org/doc/tutorials/deep/preparation) to start exploring DVC's functionalities. However, when I issue the command `dvc init` inside the `classify/` directory, I don't get the expected results.
In fact, `ls -a .dvc` should output the following:
`./      ../     .gitignore  cache/  config`
whereas my output is:
`./      ../     config .gitignore  tmp/`
so basically the `cache/` directory doesn't get created, but a `tmp/` directory does. This wouldn't be a big deal if the tutorial didn't mention that _"The .dvc/cache directory is one of the most important parts of any DVC repositories"_

jorgeorpinel:
Fix #806

### DIFF
--- a/static/docs/command-reference/init.md
+++ b/static/docs/command-reference/init.md
@@ -14,11 +14,10 @@ usage: dvc init [-h] [-q | -v] [--no-scm] [-f]
 ## Description
 
 After DVC initialization, a new directory `.dvc/` will be created with the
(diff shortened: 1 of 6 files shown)


shcheklein:
👍

Issue #759, opened 2019-10-30
md: support displaying command snippets with Windows/OS/Linux tabs to select


dashohoxha:
> Extracted from https://github.com/iterative/dvc.org/issues/698#issuecomment-546284166 +

It would be nice to display command snippets with tabs "Linux/Mac" and "Windows", if they are different (similar to this example: https://dagshub.com/docs/pipeline/#featurization-or-pre-processing).

This page has some ideas about how it can be done: https://stackoverflow.com/questions/37606292/how-to-create-tabbed-code-blocks-in-mkdocs-or-sphinx
This one also has an example (`class="codetabs"`): https://raw.githubusercontent.com/apache/spark/master/docs/quick-start.md

We might try a syntax like this:

`````
````dvc
```Linux/Mac
$ cat .gitignore
```
```Windows
$ type .gitignore
```
````
`````

There are 4 backticks for the enclosing `dvc` block. It is a valid markdown syntax and normally it would render like this:

````dvc
```Linux/Mac
$ cat .gitignore
```
```Windows
$ type .gitignore
```
````

The engine should render it with tabs, but even with normal markdown rendering it would appear somewhat acceptable.

dashohoxha:
I am not sure how many users are using DVC on Windows, but it seems that there is a workaround by using a POSIX-like command line shell: https://dvc.org/doc/user-guide/running-dvc-on-windows#posix-like-command-line-shell
So, I would say that the priority of this issue is `nice-to-have`.

jorgeorpinel:
at least one Windows user recently reported confusion that would be cleared with this improvement. See https://discordapp.com/channels/485586884165107732/485596304961962003/647534970272743424

jorgeorpinel:
In order to move away from the research phase here, what do we need to decide? I think these are the questions:

1. What platforms/tabs will we support?
  I agree that 2 suffice: **Linux/Mac terminal**, and **Windows cmd**.
1. What format will we use in Markdon files to provide the different command samples?
  Here I think we just need to use new highlight formats like the ones suggested above. I'd go for `POSIX` and `cmd` though. I don't think we need to nest them in `dvc` blocks though, we can just add them one after another and show only the one corresponding to the user's platform.
1. How do we determine the user's platform?
  Obviously there should be a tab system so it can be chosen manually, but it could also be detected automatically via browser (defaulting to `POSIX`).

Thoughts @dashohoxha @shcheklein @leehanchung?

[...]

PR #812, opened 2019-11-22
Update the linebreak symbol

leehanchung:
The original doc comment of this is confusing for Python users:
```dvc
$ mkdir data
$ dvc get https://github.com/iterative/dataset-registry \ 
get-started/data.xml -o data/data.xml
```
So changed this into one single command line for clarity after conversation with Ivan on Discord. 
```dvc
$ mkdir data
$ dvc get https://github.com/iterative/dataset-registry get-started/data.xml -o data/data.xml
```

**IMPORTANT NOTES** (please read, then delete):

- Have you followed the guidelines in
  [Contributing Documentation](https://dvc.org/doc/user-guide/contributing/docs)?

- Please use the title to provide a clear one-line present-tense summary of the
  changes introduced in the PR. For example: "Introduce the first version of the
  collection editor.".

- Please make sure to mention "Fix #bugnum" (if applicable) in the description
  of the PR. This enables GitHub to link the PR to the corresponding bug.

### DIFF
--- a/static/docs/get-started/add-files.md
+++ b/static/docs/get-started/add-files.md
@@ -6,8 +6,7 @@ Let's get a dataset example to play with:
 
 ```dvc
 $ mkdir data
-$ dvc get https://github.com/iterative/dataset-registry \
-        get-started/data.xml -o data/data.xml
+$ dvc get https://github.com/iterative/dataset-registry get-started/data.xml -o data/data.xml
 ```
 
 > `dvc get` can use any <abbr>DVC project</abbr> hosted on a Git repository to


jorgeorpinel:
Thanks @leehanchung! Are you VitaminC on Discord? Haha

The reason we have `\` after around 70 chars in many of our terminal samples is for better readability though. Otherwise the code blocks would scroll horizontally.

Fortunately we have #759 open that would solve this possible confusion for Windows users. Or if you have any other suggestions, feel free to reach out on Discord or GitHub issues.

shcheklein:
@jorgeorpinel can we do this split before `-o` to reduce chances for mistake? It's not even about Windows users vs Linux - it's also about Python habits it looks like :)

jorgeorpinel:
That would make the code not fit in the block, which would scroll:

image

shcheklein:
:( yeah, that URL is just too large

not sure though that Windows tabs is the solution for this problem

may be do `dvc get -o data/data.xml URL PATH`?

jorgeorpinel:
That still needs `\`. Probably 2:

```console
$ dvc get -o data/data.xml \
          https://github.com/iterative/dataset-registry \ 
          get-started/data.xml
```

And we have a bunch of `\` break lines all around the docs.

jorgeorpinel:
> not sure though that Windows tabs is the solution for this problem

That would definitely solve it as `\` is not supported in cmd. So for Windows cmd code samples we would have to use long lines and a special code block that wraps lines I guess. Or a horizontal scroll bar that is always shown.

leehanchung:
How about
```
dvc get https://github.com/\
iterative/dataset-registry  get-started/data.xml -o data/data.xml

Be mindful when copying the above command. Make sure there's a space between repo-url `https://github.com/iterative/dataset-registry` and `get-started/data.xml`. The usage is `dvc get` <repo-url> <repo-folder> -o <data to be pulled>
```

jorgeorpinel:
Interesting. That looks super weird though.

I'm not sure adding the note just for Windows users is a good idea. We would have to do this on every command sample with `\` throughout the docs.

Sorry to sound so negative! I think the best solution is #759, which BTW is still on `research` phase, so it could well end up being an automatic platform detection mechanism that just adds these notes automatically. I think it's best to move this discussion to that issue.

Issue #463, opened 2019-06-27
api: create documentation


jorgeorpinel:
> See src @ https://github.com/iterative/dvc/blob/master/dvc/api.py

`dvc.api.get_url`
`dvc.api.open`
`dvc.api.read`

~~Please more insights here~~ More details in https://github.com/iterative/dvc.org/issues/463#issuecomment-514139792.

Also, please update the one mention to API in the data registry (which will me merged with #818) per https://github.com/iterative/dvc.org/pull/818#issuecomment-565083934.

UPDATE:

- [ ] Also (at least basic docs for) `dvc.api.summon` and open another issue to complete it later.
- [ ] And try to do app: automatic links for API function names in md #910 once #908 is merged.

shcheklein:
@Suor can you describe it briefly, please :)

Naba7:
@jorgeorpinel @shcheklein can you please explain this a bit more. I would like to take this up.

shcheklein:
@Suor ? can you give a summary or is there a link? we should probably put docstrings around APIs before we release it.

Naba7:
sorry, I am unable to search ``dvc.api.get_content`` . Do you mean ``get_url`` or ``__getattr__`` ?

shcheklein:
@Naba7 it's `read` now, I believe. There are only three public methods in `dvc.api`. Those should be described. And let's create a ticket on iterative/dvc to update docstrings for those APIs.

Suor:
They have some short docstrings, I will update them based on future docs or discussion here if we decide to do that.

So, what is this about? There are three public things in `dvc.api` now:
1. `read(path, repo=None, rev=None, remote=None, mode="r", encoding=None)` - returns the contents of an artifact as a bytes object or a string.
2. `get_url(path, repo=None, rev=None, remote=None)` -  returns an url of an artifact.
3. `open(path, repo=None, rev=None, remote=None, mode="r", encoding=None)` - opens an artifact as a file, may only be used as context manager:
    ```python
    with dvc.api.open("path/to/data.csv", remote="my-s3", encoding="utf-8") as f:
        for line in f:
            process(line)
    ```

Arguments always mean the same:

`path` - a path to an artifact, relative to repo root,
`repo` - a path or git url of a repo,
`rev` - revision, i.e. a branch, a tag, a sha. This only works with an url in repo,
`remote` - a name of a remote to fetch artifact from/give url to
[...]

PR #818, opened 2019-11-25
use-cases: second iteration of Data Registry case

### LINKED ISSUE #779: use-cases: add API section to data registry case
From https://github.com/iterative/dvc.org/pull/679#pullrequestreview-313622925

- Once #679 is merged,
- [x] Add "Programatic access" or "Access from code" or "Access from Python" h2 section in /doc/use-cases/data-registry

### LINKED ISSUE #795: use-cases: improve Data Registry case

jorgeorpinel:
> Continuation of #805

Fix #795

- [x] Put a diagram near the top.
- [x] Further simplify intro
- [x] Introduce some section between intro and example that provides some high level abstract overview of the Git and DVC commands you would use to organize the registry (from https://github.com/iterative/dvc.org/pull/805#pullrequestreview-321999122).
- ~~We go from using regular imports/gets to setting up a dedicated data registry, while we should be comparing no DVC at all (ad-hoc conventions and total mess on S3) vs. the DVC Data Registry – which effectively provides some "meta" information for the same data on S3.~~

Also closes #779


jorgeorpinel:
Answers to your https://github.com/iterative/dvc.org/pull/805#pullrequestreview-321999122 @shcheklein:

> Further simplify intro and put an image (WIP on my end).

Feel free to commit to this PR for this (`use-cases/data-registry` branch), when ready.

> ...example turned out to be too complicated and specific (e.g. we focus on one specific data set that we don't even reuse...

What do you mean we don't reuse the data set? It's used in the get started, in the versioning tutorial, and in a bunch of cmd ref examples.

> ...and we don't mention that registry contains all the data for all example, tutorials).

I didn't understand this phrase.

shcheklein:
> Feel free to commit to this PR for this (use-cases/data-registry branch), when ready.

👍 :)

> What do you mean we don't reuse the data set? It's used in the get started, in the versioning tutorial, and in a bunch of cmd ref examples.

Where do we use in the get started? 

This is good if it's reused 👍. Probably what I wanted to say is that by going deep into explaining specifics around one particular data set we create some disconnect with the first part - that is high level about data lifecycle, data management, etc.

Managing data for _all_ data for all examples, tutorials, docs **vs** managing cats & dogs - those are two different examples. Showing benefits of being able now to reuse data.xml everywhere, see some evolution for cats & dogs and managing data on S3 properly **vs** focusing on the evolution of the cats and dogs.

Sorry, I still probably sound confusing. Only because I don't have a precise answer to this.

jorgeorpinel:
> Where do we use in the get started?

Ah, you're right, it a different directory in the same repo. Used in https://dvc.org/doc/get-started/add-files

> by going deep into explaining specifics around one particular data set we create some disconnect with the first part - that is high level about data lifecycle, data management, etc.

Got it. Same as the "Introduce some section between intro and example..." requirement for this 2nd iteration, basically 🙂 

> - [ ] Managing data for all data for all examples, tutorials, docs vs managing cats & dogs...
> Showing benefits of being able now to reuse data.xml everywhere, see some evolution for cats & dogs and managing data on S3 properly...

Got it, related to the answer about the get started.

Working on all this. Thanks for review approval but this is still WIP 😋 I'll request your review again when ready @shcheklein

static/docs/use-cases/data-registry.md
```diff
@@ -1,36 +1,28 @@
...
+[versioning of data and model files](/doc/use-cases/data-and-model-files-versioning),
+with commands such as `dvc add`. With the aim to enable reusability of these
+<abbr>data artifacts</abbr> between different projects, DVC also provides the
+`dvc get`, `dvc import`, and `dvc update` commands. This means that a project
+can depend on data from an external <abbr>DVC project</abbr>, **similar to
+package management systems, but for data**.
+
+<!-- Insert diagram image here. -->
 
 Keeping this in mind, we could build a <abbr>DVC project</abbr> dedicated to
```
Suor: I still stumble on this `could` every time.

static/docs/use-cases/data-registry.md
```diff
@@ -44,36 +36,75 @@ The advantages of using a DVC **data registry** project are:
...
+### Adding datasets to a registry
+
+<!-- Probably we will need to touch cats & dogs example and answer (again, high level perspective) to questions like - how do we get data there, how do we update it, how do we consume it from other repos.  -->
+<!-- we should be comparing no DVC at all (ad-hoc conventions and total mess on S3) vs. the DVC Data Registry – which effectively provides some "meta" information for the same data on S3. -->
+<!-- something like - we had S3 like everyone - few buckets already (for such a small case & docs), few versions of the files in different locations, directories with images repeatings, etc - explain high level problems -->
+<!-- let’s just assume that we have a directory and we keep adding files -->
+
+Imagine a training dataset with 1000 images of cats and dogs that will be used
+to build an ML model. Without DVC, in order for a team to collaborate on this
+project, we could just uploading it to cloud storage (e.g. Amazon S3) and
```
Suor: uploading -> upload ... cloud storage -> a cloud storage

static/docs/use-cases/data-registry.md
```diff
@@ -44,36 +36,75 @@ The advantages of using a DVC **data registry** project are:
...
+provide everyone with access.
+
+At some point though, we need to add another 1000 images to the dataset, but the
```
Suor: We had everything in past tense (could) and now we switch to present.

static/docs/use-cases/data-registry.md
jorgeorpinel: I don't know what this comment means. Seems like an observation. Resolving.

static/docs/use-cases/data-registry.md
jorgeorpinel: Oh this part is not to be reviewed yet, sorry for confusion. Will keep that in mind though.

static/docs/use-cases/data-registry.md
jorgeorpinel: uploading -> upload ✔️ 
cloud storage -> a cloud storage ❌ 
Updated. Thanks!

jorgeorpinel:
Hey @Suor we hadn't seen you in a while. Lots have changed since you started this discussion. I forgot to remove my request for your review which is a couple days old, sorry. This PR is marked as WIP and shouldn't be reviewed yet. Hopefully that's why you still find is super confusing. So far it's only ready to review up to before the **Adding datasets to a registry** section starts (line 7 rn) – Before cats and dogs. Do you have any comments on any of that?

jorgeorpinel:
@Suor @shcheklein feel fee to check out the current WIP version in https://dvc-org-pr-818.herokuapp.com/doc/use-cases/data-registry

static/docs/use-cases/data-registry.md
Suor: I feel present tense reads more smooth if that matters :)

jorgeorpinel: The way we're using "could" is also present tense actually

image

But I know what you mean. It's just that some of these paragraphs use hypotetical/conditional forms and that requires these auxiliary/past verbs. Would need to rewrite the whole paragraph differently, and we definitely **could**, but I don't think this is a problem.

jorgeorpinel:
Please see latest version which is the first complete draft of the entire second iteration of this doc. 

Also, @shcheklein is this still relevant? Below:

> We go from using regular imports/gets to setting up a dedicated data registry, while we should be comparing no DVC at all (ad-hoc conventions and total mess on S3) vs. the DVC Data Registry...

shcheklein:
> Also, @shcheklein is this still relevant?

It's relevant as a general comment :) Not relevant in an actionable way. Just useful though to keep in mind. Feel free to resolve it though.

Suor:
**Summary: way easier to read now.**

Title: Data Registry (match menu, easier)

> any other DVC repositories

any other DVC repository


> The actual data is stored in the project's cache and can be pushed 

For imports to work it should be pushed and you need a deafult upstream configured, this is not optional. The only exception is if we work within a single machine and import locally.

> as well as dvc.api.open()

There are also `dvc.api.read()` and `dvc.api.get_url()`. The `.read()` is the most high-level. Also, why it is not linked to anything? We don't have api documented?

> has a --revision option

It is `--rev` not `--revision`


> with dvc.api.open(data_path, repo_url) as dataset:

dataset -> file_descriptor or fd
```python
# Note repo=, this is the preferred way, more future-proof
with dvc.api.open(data_path, repo=repo_url, rev=...) as fd:
    full_text = fd.read()
    # or
    df = pandas.read_csv(fd)
    # or read line by line
    for line in fd:
        process(line)
    # or any other code consuming fd
    # ...
```

Updating registries is unfinished? you need:
```bash
dvc add music/songs
git commit music/songs.dvc -m "Update songs dataset"
git push
dvc push  # this might be done with a hook
```

jorgeorpinel:
@Suor I applied most of your suggestions.

> There are also `dvc.api.read()` and `dvc.api.get_url()`. The `.read()` is the most high-level. Also, why it is not linked to anything? We don't have api documented?

No API docs at all yet. This is the very fist mention. Let's review this paragraph as part of #463.

> dataset -> file_descriptor or fd

The example opens a model file actually. How would you consume it? I'm just leaving a commend inside to avoid getting into technical details but open to suggestions for a model file. Should we use a different function instead of open like you suggested earlier? Not sure about all this.

> Updating registries is unfinished

Just the git commit/push commands are "missing", which is not DVC related.

jorgeorpinel:
> Updating registries is unfinished

I see, we're not showing `dvc push` at all in the doc in fact. I think we want to keep it lean and that's why. What do you think @shcheklein ? I guess it's pretty critical to push the data to a remote, should I add both in **Building** and **Updating**?

static/docs/use-cases/data-registries.md
```diff
@@ -0,0 +1,195 @@
...
+import dvc.api.open
+
+model_path = 'path/to/model'
+repo_url = 'git@git-server.url:path/to/repository.git'
+
+with dvc.api.open(model_path, repo_url) as model:
+    # Make some predictions...
+```
+
+This opens `path/to/dataset` as a file descriptor. Such a method could be used
```
shcheklein: `path/to/model`? .. better use `model.pkl` ... if we use our example-get-started we can even provide a real example

static/docs/use-cases/data-registries.md
```diff
@@ -0,0 +1,195 @@
...
+
+> Note that this command (as well as `dvc import`) has a `--rev` option to
+> download specific versions of the data.
+
+### Import workflow
+
+`dvc import` uses the same syntax as `dvc get`:
+
+```dvc
+$ dvc import git@git-server.url:path/to/repository.git \
```
shcheklein: if the intention is to keep it generic - let's use just <git-repo-url>?

shcheklein: the whole notation `git@git-server.url:path/to/repository.git` feel too detailed. Also, you can do https:// which is simple, you don't need `git` protocol in a lot of cases

static/docs/use-cases/data-registries.md
```diff
@@ -0,0 +1,195 @@
+# Data Registries
```
shcheklein: I tend to agree with @Suor - still not sure why do we use Registries here instead of Registry

Suor:
> The example opens a model file actually. How would you consume it?

With models there is no universal code, however, if it's `*.pkl` then you can simply unpickle it:

```python
with dvc.api.open('model.pkl', repo='https://...') as fd:
    model = pickle.load(fd)
# or 
model = pickle.loads(dvc.api.read('model.pkl', repo='https://...'))
```

You can use any one if these examples. They are simple, common enough and useful.

> Just the git commit/push commands are "missing", which is not DVC related.

`dvc push` is dvc related, which actually sends new data to remote making it accessible for gets/imports. And if you don't do `git push` you won't update repo either, so a consumer won't see anything in `dvc status` and won't be able to `dvc update` that one.

static/docs/use-cases/data-registries.md
jorgeorpinel: This is something we decided privately [in Slack](https://iterativeai.slack.com/archives/DLW4G0F7W/p1576028094019200?thread_ts=1576025894.008300&cid=DLW4G0F7W). Although your idea originally was **Data and model registries**, but we ended only making it plural for data, and adding more mentions of models in the body. 

BTW I have changed the nav title to plural also already: https://dvc-landing-use-cases-d-b6xlg5.herokuapp.com/doc/use-cases/data-registries. Should we roll back both to singular?

shcheklein: @jorgeorpinel I will let you decide on this :) to me "data registry" sounds better than "data registries" as I'm reading it now ... model - you decided (and that was probably right) to not complicate it

static/docs/use-cases/data-registries.md
jorgeorpinel: Changing to `nlp/model.pkl` (in hypothetical `https://github.com/iterative/model-registry` repo).

Avoiding real examples so we don't have to make sure the commands are reproducible.

jorgeorpinel: Changing to `https://github.com/example/registry` per Slack conversation. Applying similar changes in the other URLS.

static/docs/use-cases/data-registries.md
jorgeorpinel: OK. Going for `https://github.com/example/registry`.

jorgeorpinel:
> `dvc push` is dvc related, which actually sends new data to remote...

@Suor yep, I noticed and mentioned that in https://github.com/iterative/dvc.org/pull/818#issuecomment-565248741 as well. I just wasn't sure we wanted to add that step or whether that would complicate the text. But I think you're right, it's a critical thing to show both in the **Building** and **Updating** sections, will add.

static/docs/use-cases/data-registries.md
jorgeorpinel: I actually like the plural better. I think otherwise it may sound like we only recommend concentrating everything in a single registry, whereas one could have several ones, some for datasets, some for models, etc.

jorgeorpinel:
OK, addressed everything and resolved conflicts with `master` branch. Merge? 😀

PR #819, opened 2019-11-25
api: Discourse API "proxy" endpoint at /api/comments

iAdramelk:
API handle to return comments count to use in blog.dvc.org (see iterative/blog#8).

### DIFF
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -14,7 +14,7 @@ const getCommentCount = (req, res) => {
   } = req
 
   if (!url.startsWith(FORUM_URL)) {
-    res.status(404).json({ error: `URL should starts with '${FORUM_URL}'` })
+    res.status(400)
 
     return
   }
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -21,7 +21,7 @@ const getCommentCount = (req, res) => {
 
   request(`${url}.json`, (error, response, body) => {
     if (error || response.statusCode !== 200) {
-      res.status(502).json({ error: "Unexpected response from Forum" })
+      res.status(502).json({ error: 'Unexpected response from Forum' })
 
       return
     }
@@ -29,7 +29,7 @@ const getCommentCount = (req, res) => {
     const json = JSON.parse(body)
 
     if (!json.posts_count) {
-      res.status(502).json({ error: "Unexpected payload from Forum" })
+      res.status(502).json({ error: 'Unexpected payload from Forum' })
 
       return
     }
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -1,3 +1,14 @@
+/*
+ * This API endpoint is used by https://blog.dvc.org
+ * to get comments count for the post, it gets
+ * discuss.dvc.org topic url as a param and returns
+ * comments count or error.
+ *
+ * It made this way to configure CORS, reduce user's payload
+ * and to add potential ability to cache comments count
+ * in the future.
+ */
+
 import Cors from 'micro-cors'
 import request from 'request'
 
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -29,7 +29,7 @@ const getCommentCount = (req, res) => {
     const json = JSON.parse(body)
 
     if (!json.posts_count) {
-      res.status(404).json({ error: "Forum's don't have 'posts_count' field" })
+      res.status(502).json({ error: "Unexpected payload from Forum" })
 
       return
     }
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -21,7 +21,7 @@ const getCommentCount = (req, res) => {
 
   request(`${url}.json`, (error, response, body) => {
     if (error || response.statusCode !== 200) {
-      res.status(404).json({ error: 'Forum returned incorrect response' })
+      res.status(502).json({ error: "Unexpected response from Forum" })
 
       return
     }
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -1,0 +1,25 @@
+import Cors from 'micro-cors'
+import request from 'request'
+
+const cors = Cors({
+  allowedMethods: ['GET', 'HEAD']
+})
+
+const getCommentCount = (req, res) => {
+  const {
+    query: { url }
+  } = req
+
+  request(`${url}.json`, (error, response, body) => {
+    if (error || response.statusCode !== 200) {
+      res.status(404).json({ error })
+    } else {
+      // post_count return all posts including topic itself
+      const count = JSON.parse(body).posts_count - 1
+
+      res.status(200).json({ count })
+    }
+  })
+}
+
+export default cors(getCommentCount)
--- a/pages/api/comments.js
+++ b/pages/api/comments.js
@@ -1,8 +1,11 @@
 import Cors from 'micro-cors'
 import request from 'request'
 
+import { BLOG_URL, FORUM_URL } from '../../src/consts'
+
 const cors = Cors({
-  allowedMethods: ['GET', 'HEAD']
+  allowedMethods: ['GET', 'HEAD'],
+  origin: BLOG_URL
 })
 
 const getCommentCount = (req, res) => {
@@ -10,15 +13,31 @@ const getCommentCount = (req, res) => {
     query: { url }
   } = req
 
+  if (!url.startsWith(FORUM_URL)) {
+    res.status(404).json({ error: `URL should starts with '${FORUM_URL}'` })
+
+    return
+  }
+
   request(`${url}.json`, (error, response, body) => {
     if (error || response.statusCode !== 200) {
-      res.status(404).json({ error })
-    } else {
-      // post_count return all posts including topic itself
-      const count = JSON.parse(body).posts_count - 1
+      res.status(404).json({ error: 'Forum returned incorrect response' })
 
-      res.status(200).json({ count })
+      return
     }
+
+    const json = JSON.parse(body)
+
+    if (!json.posts_count) {
+      res.status(404).json({ error: "Forum's don't have 'posts_count' field" })
+
+      return
+    }
+
+    // post_count return all posts including topic itself
+    const count = json.posts_count - 1
+
+    res.status(200).json({ count })
   })
 }
 
--- a/server.js
+++ b/server.js
@@ -10,10 +10,9 @@ const { createServer } = require('http')
 const { parse } = require('url')
 const next = require('next')
 const querystring = require('querystring')
-const request = require('request')
 
 const dev = process.env.NODE_ENV !== 'production'
-const app = next({ dev })
+const app = next({ dev, crossOrigin: 'anonymous' })
 const port = process.env.PORT || 3000
 const handle = app.getRequestHandler()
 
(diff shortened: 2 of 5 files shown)


iAdramelk:
Forum part https://github.com/iterative/blog/pull/8

server.js
```diff
@@ -6,12 +8,11 @@
 
 const { createServer } = require('http')
 const { parse } = require('url')
-const _ = require('force-ssl-heroku')
```
shcheklein: why removing this?

shcheklein: should we remove it from dependencies?

pages/api/comments.js
```diff
@@ -0,0 +1,25 @@
+import Cors from 'micro-cors'
+import request from 'request'
+
+const cors = Cors({
```
shcheklein: look a bit too permissive? can it be set to allow only from certain domains?

probably not a big deal right now ...

jorgeorpinel:
> Forum part iterative/blog#8

I don't understand how that PR implements a forum. What do you mean? I don't even see the word forum in that PR's changes or comments. Can you explain @iAdramelk? Just curious. (Still pretty unfamiliar with the blog engine.)

pages/api/comments.js
jorgeorpinel: Agree. The blog is at blog.dvc.org right? Is CORS needed for different subdomains?

pages/api/comments.js
```diff
@@ -0,0 +1,25 @@
...
+  allowedMethods: ['GET', 'HEAD']
+})
+
+const getCommentCount = (req, res) => {
+  const {
+    query: { url }
+  } = req
+
+  request(`${url}.json`, (error, response, body) => {
```
jorgeorpinel: What `url` would this typically be? Thinking about possible security vulnerabilities since this is input passed on to `request` without validation.

jorgeorpinel: Also (for @shcheklein) do we rate limit API requests somehow?

server.js
```diff
@@ -6,12 +8,11 @@
...
 const next = require('next')
 const querystring = require('querystring')
 
 const dev = process.env.NODE_ENV !== 'production'
-const app = next({ dev })
+const app = next({ dev, crossOrigin: 'anonymous' })
```
jorgeorpinel: Is this the same as setting a config option for that?

```js
// next.config.js
module.exports = {
  crossOrigin: 'anonymous',
}
```

which seems to be the official rec, per https://github.com/zeit/next.js#cdn-support-with-asset-prefix

server.js
iAdramelk: It was not used, yes we should probably remove it from dependencies as well.

pages/api/comments.js
iAdramelk: Yep, we can configure origins, but this CORS rule is only used for the `/api/comments` route so it is not a big problem in this case.

server.js
shcheklein: yep, I see it's a leftover before an explicit 301 was introduced

pages/api/comments.js
iAdramelk: It will be a duscourse forum urls like: https://discuss.dvc.org/t/how-do-i-remove-files-which-are-no-longer-tracked-by-dvc-from-dvc-remote-ssh-storage/113

Basically if you add `.json` to it, it will return you json representation of the topic.

server.js
iAdramelk: Ooops, my bad. Forgot to remove it from `server.js`, it is not needed here anymore.

iAdramelk:
> I don't understand how that PR implements a forum. What do you mean? I don't even see the word forum in that PR's changes or comments. Can you explain @iAdramelk? Just curious. (Still pretty unfamiliar with the blog engine.)

It is not implementing the forum, it is basically just getting comments count for existing Discourse topic using Discourse API.

pages/api/comments.js
shcheklein: can someone start using us as a proxy for their purposes? 

btw, since this API is not authenticated in any way - what is the conceptual difference between this proxy and using it directly (with CORS enabled in admin for the blog.dvc.org) from the client? You mentioned that API was not designed to be used externally, but from this implementation I don't quite understand the difference to be honest.

shcheklein:
@jorgeorpinel this is to be used along with a PR on iterative/blog to show number of comments for each post.

pages/api/comments.js
shcheklein: @iAdramelk question is still valid - the current setup means that anyone can pass any url and use us a proxy, right?

iAdramelk: @shcheklein Well. This route returns json in the form of `{ count: NUMBER }` or `{ error: STRING }`. If someone wants to get this json from us by passing their URL to this route, then yes, they can do it.

pages/api/comments.js
iAdramelk: This is not a proxy. We will not return original content, we will return our own JSON with `{ count: NUMBER }` or `{ error: STRING }`.

iAdramelk: Difference with direct usage is that original payload is 10kb of json, but we only return one field from it in our API. We can even cache this one number on our side in the future and return it without the need for downloading the original 10kb json every time.

shcheklein: kk, got it. 👍 I didn't mean that this is the exact "proxy". I didn't get from the initial conversation that it's only about optimizing 10kb workload, I though that API is not "designed" to be used from the client. Do we actually need to optimize 10kb? Especially considering that we are creating a dependency on the server and should think about rate limits for example and scaling it eventually.

pages/api/comments.js
shcheklein: it means for example that someone can hit some malware URL using our server and get us blocked for example?

jorgeorpinel: Maybe just validate agains a simple regex `/^https:\/\/discuss.dvc.org\/[a-z\/-0-9]$/` and return HTTP 403 if failed.

server.js
jorgeorpinel: OK 🙂

pages/api/comments.js
iAdramelk: Added CORS origin and URL validation to prevent it.

jorgeorpinel: Great, thanks.

pages/api/comments.js
jorgeorpinel: I think this is resolved by lines 7 and 8 below which changed after this review comment was made.

src/consts.js
```diff
@@ -8,3 +8,6 @@ export const META_DESCRIPTION =
 export const META_KEYWORDS =
   'data version control machine learning models management'
 export const META_SOCIAL_IMAGE = 'https://dvc.org/static/social-share.png'
+
+export const FORUM_URL = 'https://discuss.dvc.org'
+export const BLOG_URL = 'https://blog.dvc.org'
```
jorgeorpinel: How about something like

```suggestion
import hostname from 'os'

export const BLOG_URL = 'https://blog.' + hostname()
```

jorgeorpinel: How about something like

```js
import hostname from 'os'

export const FORUM_URL = 'https://discuss.' + hostname()
export const BLOG_URL = 'https://blog.' + hostname()
```

pages/api/comments.js
```diff
@@ -0,0 +1,44 @@
...
+  origin: BLOG_URL
+})
+
+const getCommentCount = (req, res) => {
+  const {
+    query: { url }
+  } = req
+
+  if (!url.startsWith(FORUM_URL)) {
+    res.status(404).json({ error: `URL should starts with '${FORUM_URL}'` })
```
jorgeorpinel: Not sure we need to give details as to how to form a correct response. Maybe just return 400 Bad Request.

pages/api/comments.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return
+  }
+
+  request(`${url}.json`, (error, response, body) => {
+    if (error || response.statusCode !== 200) {
+      res.status(404).json({ error: 'Forum returned incorrect response' })
```
jorgeorpinel:
```suggestion
      res.status(502).json({ error: "Unexpected response status from Forum" })
```

Notice 502 Bad Gateway response :)

pages/api/comments.js
```diff
@@ -0,0 +1,44 @@
...
+
+      return
+    }
+
+    const json = JSON.parse(body)
+
+    if (!json.posts_count) {
+      res.status(404).json({ error: "Forum's don't have 'posts_count' field" })
```
jorgeorpinel:
```suggestion
      res.status(502).json({ error: "Unexpected response payload from Forum" })
```

(Removed implementation detail)

src/consts.js
iAdramelk: We can do it, but then it will not work on Netlify and on localhost. So I think we should hardcode dvc.org here.

pages/api/comments.js
iAdramelk: Make sense, will change.

iAdramelk:
Made some changes in error responses

src/consts.js
jorgeorpinel: OK how about

```js
const WEBSITE_HOST = 'dvc.org'
export const FORUM_URL = 'https://discuss.' + WEBSITE_HOST
export const BLOG_URL = 'https://blog.' + WEBSITE_HOST
```

in case `WEBSITE_HOST` so its hardcoded in a single place?

Also, could `META_SOCIAL_IMAGE` below be just

```js
export const META_SOCIAL_IMAGE = '/static/social-share.png'
```

?

pages/api/comments.js
jorgeorpinel: Thanks

jorgeorpinel:
I guess we could also add a small section in REAMDE describing this endpoint's spec? Just so in the future we remember what request/response (payloads) are expected.

src/consts.js
iAdramelk: We can add `WEBSITE_HOST`, yes, but `META_SOCIAL_IMAGE` has to be full url – it is not used on our site, it is used by facebook/twitter to display an image preview.

iAdramelk:
@jorgeorpinel 

Added `WEBSITE_HOST` and comment in the begining of the `pages/api/comments.js`. I think it's better place for it then `readme.md` – readme is about dvc.org site, and this endpoint is not used by site, but is used by separate blog.

src/consts.js
shcheklein: since @jorgeorpinel is traveling today ... taking aa liberty to resolve and merge to move this thingy forward

shcheklein: @jorgeorpinel feel free to reopen review and we will get to it again

shcheklein:
@iAdramelk please resolve the conflicts ^^

iAdramelk:
@shcheklein Rebased it.

shcheklein:
thanks @iAdramelk !

src/consts.js
jorgeorpinel: Looks perfect. Thanks!

pages/api/comments.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * This API endpoint is used by https://blog.dvc.org
+ * to get comments count for the post, it gets
+ * discuss.dvc.org topic url as a param and returns
+ * comments count or error.
+ *
+ * It made this way to configure CORS, reduce user's payload
+ * and to add potential ability to cache comments count
+ * in the future.
+ */
```
jorgeorpinel: I'll rewrap this.

PR #823, opened 2019-11-28
get started: update index page

### LINKED ISSUE #739: update get-started index page
https://dvc.org/doc - it mentions tutorials that have been moved to another section and a deep dive tutorial - I would replace it with a link to the Tutorials section and fix the text around

shcheklein:
Fixes #739

### DIFF
--- a/static/docs/get-started/index.md
+++ b/static/docs/get-started/index.md
@@ -15,6 +15,9 @@ if you have any questions or need any help. We are very responsive ⚡.
 ✅ Check out our [GitHub repository](https://github.com/iterative/dvc) and give
 us a ⭐ if you like the project!
 
-When you are done with _Get Started_ or if you'd like to try other alternatives
-to learn DVC, check the available examples, interactive and community tutorials
-on this [page](/doc/tutorials).
+✅ Contribute to DVC [on GitHub](https://github.com/iterative/dvc) or help us
+improve this [documentation](https://github.com/iterative/dvc.org) 🙏.
+
+When you are done with _Get Started_ and ready to cover other topics, or if
+you'd just like to try other ways to learn DVC, check the available examples,
+interactive and community tutorials on this [page](/doc/tutorials).
--- a/static/docs/get-started/index.md
+++ b/static/docs/get-started/index.md
@@ -15,6 +15,6 @@ if you have any questions or need any help. We are very responsive ⚡.
 ✅ Check out our [GitHub repository](https://github.com/iterative/dvc) and give
 us a ⭐ if you like the project!
 
-When you are done with Get Started or if you'd like to try other alternatives to
-learn DVC, check the available examples, interactive and community
-[tutorials](/doc/tutorials).
+When you are done with _Get Started_ or if you'd like to try other alternatives
+to learn DVC, check the available examples, interactive and community tutorials
+on this [page](/doc/tutorials).
--- a/static/docs/get-started/index.md
+++ b/static/docs/get-started/index.md
@@ -1,12 +1,13 @@
 # Get Started
 
-_Get Started_ is a step-by-step introduction into basic DVC concepts. It doesn't
+_Get Started_ is a step-by-step introduction into basic DVC features. It doesn't
 go into details much, but provides links and expandable sections to learn more.
 
-At the very end there are a few complete examples to give you more hands-on
-experience with real life scenarios. The first one is about model and dataset
-[versioning](/doc/tutorials/versioning), and the second one is focused on
-[pipelines and reproducibility](/doc/tutorials/pipelines).
+You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
+Also, if DVC is not installed, please follow these [instructions](/doc/install)
+to do so.
+
+Before you start ...
 
 ✅ Please, join our [community](/chat) or see these [support](/support) options
 if you have any questions or need any help. We are very responsive ⚡.
@@ -14,13 +15,6 @@ if you have any questions or need any help. We are very responsive ⚡.
 ✅ Check out our [GitHub repository](https://github.com/iterative/dvc) and give
 us a ⭐ if you like the project!
 
-✅ Contribute either [on GitHub](https://github.com/iterative/dvc) or
-[on Patreon](https://www.patreon.com/DVCorg/overview) to support the project.
-
-Separate to this section, the longer [Tutorial](/doc/tutorials/deep) also
-introduces DVC step-by-step, while additionally explaining in great detail the
-motivation and what's happening internally.
-
-You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
-Also, if DVC is not installed, please follow these [instructions](/doc/install)
-to do so.
+When you are done with Get Started or if you'd like to try other alternatives to
+learn DVC, check the available examples, interactive and community
+[tutorials](/doc/tutorials).
--- a/static/docs/get-started/index.md
+++ b/static/docs/get-started/index.md
@@ -3,9 +3,9 @@
 _Get Started_ is a step-by-step introduction into basic DVC features. It doesn't
 go into details much, but provides links and expandable sections to learn more.
 
-You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
-Also, if DVC is not installed, please follow these [instructions](/doc/install)
-to do so.
+You'll need [Git](https://git-scm.com) to run the commands in this guide. Also,
+if DVC is not installed, please follow these [instructions](/doc/install) to do
+so.
 
 Before you start ...
 
@@ -18,6 +18,7 @@ us a ⭐ if you like the project!
 ✅ Contribute to DVC [on GitHub](https://github.com/iterative/dvc) or help us
 improve this [documentation](https://github.com/iterative/dvc.org) 🙏.
 
-When you are done with _Get Started_ and ready to cover other topics, or if
-you'd just like to try other ways to learn DVC, check the available examples,
-interactive and community tutorials on this [page](/doc/tutorials).
+When you're done, feel free to check out official and community
+[tutorials](/doc/tutorials). They provide in-depth explanations,
+[interactive](doc/tutorials/interactive) scenarios, and various examples on how
+DVC can be used.


static/docs/get-started/index.md
```diff
@@ -1,26 +1,20 @@
...
 
-_Get Started_ is a step-by-step introduction into basic DVC concepts. It doesn't
+_Get Started_ is a step-by-step introduction into basic DVC features. It doesn't
 go into details much, but provides links and expandable sections to learn more.
 
-At the very end there are a few complete examples to give you more hands-on
-experience with real life scenarios. The first one is about model and dataset
-[versioning](/doc/tutorials/versioning), and the second one is focused on
-[pipelines and reproducibility](/doc/tutorials/pipelines).
+You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
```
jorgeorpinel: `in this tutorial` -> `in this section` or `in these chapters`

static/docs/get-started/index.md
```diff
@@ -1,26 +1,20 @@
...
-Separate to this section, the longer [Tutorial](/doc/tutorials/deep) also
-introduces DVC step-by-step, while additionally explaining in great detail the
-motivation and what's happening internally.
-
-You'll need [Git](https://git-scm.com) to run the commands in this tutorial.
-Also, if DVC is not installed, please follow these [instructions](/doc/install)
-to do so.
+When you are done with _Get Started_ or if you'd like to try other alternatives
+to learn DVC, check the available examples, interactive and community tutorials
+on this [page](/doc/tutorials).
```
jorgeorpinel: `To cover more advanced topics step-by-step and with more in-depth explanations, please check out our various [Tutorials](/doc/tutorials).`

static/docs/get-started/index.md
```diff
@@ -1,26 +1,20 @@
...
+Before you start ...
 
 ✅ Please, join our [community](/chat) or see these [support](/support) options
 if you have any questions or need any help. We are very responsive ⚡.
 
 ✅ Check out our [GitHub repository](https://github.com/iterative/dvc) and give
 us a ⭐ if you like the project!
 
-✅ Contribute either [on GitHub](https://github.com/iterative/dvc) or
-[on Patreon](https://www.patreon.com/DVCorg/overview) to support the project.
```
jorgeorpinel: No GH repo link?

static/docs/get-started/index.md
shcheklein: I would not say they are about advanced topics only

static/docs/get-started/index.md
shcheklein: yes, need to get GH back .. good catch!

static/docs/get-started/index.md
shcheklein: don't like section or chapters - sounds too formal? Will try to come up with some alternative

static/docs/get-started/index.md
jorgeorpinel: OK just `cover other topics` or `specific topics` 🙂

static/docs/get-started/index.md
jorgeorpinel: We call Get Started a "section" (of the docs) in a bunch of other places already. It's better than calling it a tutorial. Open to suggestions, lmk in order to review the term "section" throughout. Thanks

shcheklein: How about guide?

I think the usage may vary depending on the context. And it's totally fine. The primary goal is the best experience for the reader. And, specifically, for the first time reader here. No point in limiting us to the rules for the sake of rules - it can create some bottlenecks down the road. 

So, in this specific case I would try to answer - what is easier to read and understand when you open the docs for the first time - section or tutorial or even guide? Or may be something else even?

What do you think?

static/docs/get-started/index.md
shcheklein: `When you are done with _Get Started_ and want to cover other topics or if you'd like to try other ways to learn DVC`?

static/docs/get-started/index.md
jorgeorpinel: Def. not tutorial. I do like "guide". Would be good to review the existing instances of "section" though (migrate them to "guide" where appropriate).

I also don't dislike `run the commands in the following chapters`.

static/docs/get-started/index.md
jorgeorpinel: Hmm how about

`To cover other topics step-by-step, more in-depth explanations, and even interactive scenarios, please check out our various [Tutorials](/doc/tutorials).`

It's more compact. And why should we expect Get Started to be the first guide people follow before jumping to other stuff? Tutorials are also good to "get started" with DVC.

shcheklein: >  And why should we expect Get Started

probably because this is _the_ _Get started_, right? :)

> To cover other topics step-by-step, more in-depth explanations, and even interactive scenarios, please check out our various [Tutorials](/doc/tutorials)

sounds too complicated :(

jorgeorpinel: Haha yeah it's called Get Started and it's the top section of the docs but that doesn't mean all users will want to follow it. Maybe they prefer to read the cmd refs directly or explore the docs and find about interactive or ML-focused tutorials and use those to "get started" with DVC. That's my point. Your sentence even acknowledges this fact in "try other ways to learn DVC".

`When you are done with _Get Started_ and want to cover other topics, or if you'd like to try other ways to learn DVC, check out the available examples, interactive and community tutorials on this [page](/doc/tutorials).`

^ Sentence is too long. Maybe just break it into 2 sentences, and don't assume Get Started is the first thing all people will cover from our docs? Also notice comma after "topics" and "check out". I also don't like the phrase "available examples, interactive"; I suggest "different examples, as well as interactive and community tutorials available..." But also I don't like that the word linked to tutorials is "page", and not "tutorials". Haha that's why I just rewrote the whole paragraph as a suggestion instead of trying to make small suggestions here :P

shcheklein: > Your sentence even acknowledges this fact

yep, so, don't feel the problem with `and don't assume Get Started is the first thing all people will cover from our docs`

All other changes suggestions sound minor to me.

Sentence is getting complicated as we keep adding stuff to it :)

shcheklein: The initial version was simpler.

shcheklein: it's not right to link "tutorials" in this sentence - examples are not linked this way.

jorgeorpinel: I think the simplest one is my suggestion:

> To cover other topics step-by-step, more in-depth explanations, and even interactive scenarios, please check out our various [tutorials](/doc/tutorials).

Which you think sounds complicated? I don't get it. I also don't know what's wrong with linking the word tutorials to /doc/tutorials. Seems also like the simplest solution to me. But OK, as you prefer 👍

shcheklein: ok, how about:

After you are done check this [section](link) for other examples, in-depth explanations, and interactive tutorials.

shcheklein: I think I don't like two things.

Since tutorial is taken (by the section name) you had to use second order nouns like scenarios, explanations, etc. Instead actual ones - tutorial, example.

Also, `step-by-step, more in-depth` - not sure it gives much value to be honest.

jorgeorpinel: > After you are done check this section for other examples, in-depth explanations, and interactive tutorials.

Hmmm sorry. We're really not on the same wavelengths on this particular sentence 🙁 I do like that this latest version is compact though! But:
- I'd start it `After you are done check out...`
- I thought we didn't want to use the term "section"? Why is OK for /docs/tutorials? I still don't get why linking the word "tutorials" isn't the obvious choice.
- The Get Started is not really an example, how can you say "other examples"?
- "step-by-step" tried to to the fact that you can actually reproduce the chapters/guide. But yes, maybe "end-to-end" or another similar term is better?
- "in-depth explanations" refers to the fact that tutorials may have more DVC internal/implementation details. Not sure this is always true though, but as long as some do, the usage of this phrase seems valid to me.

jorgeorpinel: But I feel this discussion is too consuming over a single paragraph. How about you decide based on everything above ^ (just try to use comma and "check out") and we merge this? We just ask everyone on Slack what they think about the sentence.

static/docs/get-started/index.md
shcheklein: Updated! My 2cs though - I think we indeed spend too much time debating this. I don't feel the difference or importance to be strict about the tutorial vs guide for the Get started. I'm pretty sure it's perceived as the same thing more or less.

In general I would say I care about consistency in the command reference most when we talk about DVC features - DVC-file, pipeline, etc. It's really not important when we talk about the documentation in the one of the supposed to reader friendly piece - get started index page.

static/docs/get-started/index.md
shcheklein: Thanks! Made some final updates and merged.

> I'd start it When you're done, feel free to check out.

👍 Thanks

> I thought we didn't want to use the term "section"? Why is OK for /docs/tutorials? 

👍 don't use it anymore. Though, see some points below ...

> I still don't get why linking the word "tutorials" isn't the obvious choice.

Because in the sentence like this (that I had initially?): 

`check the available examples, interactive and community [tutorials](http://link.com)`

linking like this means that examples do not have a link at all. The provided link is obviously related to interactive and community tutorials, not examples. So, it's just wrong. At least that's how I would read it.

> The Get Started is not really an example, how can you say "other examples"

It can be considered as an example from a regular reader perspective. It's an examples on how one can use DVC to manage NLP pipeline and data artifacts involved. Again, it feels that we go too strict in a place where it's not that needed to my mind.

> "step-by-step" tried to to the fact that you can actually reproduce the chapters/guide. But yes, maybe "end-to-end" or another similar term is better?

Because I don't fee it's important to specify that they are end-to-end. We have a full spectrum of them. And probably users would care more about feature coverage than about end-to-end-ness of the tutorials. So, it feels it can be omitted.

> in-depth explanations

👍 I have added this. We had this before and indeed it makes sense to specify because at the very top we mention Get Started is not in-depth.

> We just ask everyone on Slack what they think about the sentence.

let's not do this, too much attention to such a minor thing already :)

> But I feel this discussion is too consuming over a single paragraph.

true! the only point I'm arguing and trying to clarify my position about, is that the documentation consists of different parts and it's fine to be very (mathematically) strict in some (command reference) and less so in others (get started, use cases, etc) for which I would prefer simplicity and readability over strictness.

static/docs/get-started/index.md
jorgeorpinel: > I care about consistency in the command reference most when we talk about DVC features - DVC-file, pipeline, etc.

I like this approach, I can definitely be less strict, it will make maintaining the docs easier so I welcome this.

Still I think using terminology too loosely could result in confusing or unprofessionally looking text. Let's just use common sense? (Without trying to cross check every other doc for consistency) **As long as the terminology is correct then it should be fine.**

For example calling Get Started a "tutorial" is correct if you look at the definition of "tutorial". BUT I would still try to avoid it if possible to avoid confusion, since we have a separate section called Tutorials and Get Started isn't inside. If it's hard to avoid, then OK no problem.

static/docs/get-started/index.md
jorgeorpinel: > Because in the sentence like this

I see. My suggestion was to not use the term "examples" there, so that's why for me linking "tutorials" seemed perfect.

> It's an examples on how one can use DVC to manage NLP pipeline and data artifacts involved. Again, it feels that we go too strict...

Right, I guess an "example" (`an instance serving to illustrate a rule or precept or to act as an exercise`, [def.](https://www.merriam-webster.com/dictionary/example)) can be very broad or very narrow depending on the context. The question is, an example of what?

I'm used to thinking about examples in our docs for more narrow things: specific command uses in cmd refs. But if we're talking broadly about ML project examples, then yes, Get Started and all tutorials are examples of that. Maybe I would just specify "data processing examples", or something like that.

> I don't fee it's important to specify that they are end-to-end...

Noted.

> it's fine to be very (mathematically) strict in some (command reference) and less so in others (get started, use cases, etc) for which I would prefer simplicity and readability

Understood. Again, my only note on this is in https://github.com/iterative/dvc.org/pull/823#discussion_r352320882.

static/docs/get-started/index.md
shcheklein: 🤝

PR #827, opened 2019-11-29
Fix #729: Open external domain links in a new tab

### LINKED ISSUE #729: engine: open external domain links in a new tab
Use a special symbol :after: the link that go to an external resource.

Also, I think, we should use a new tab to open them.

iAdramelk:
Added custom renderer for links that will add target="_blank" to external links and show icon after them

### DIFF
--- a/src/Documentation/Markdown/Markdown.js
+++ b/src/Documentation/Markdown/Markdown.js
@@ -390,16 +390,14 @@ export const GithubLink = styled(LightButton)`
 `
 
 const ExternalLink = styled.a`
-  position: relative;
-  padding-right: 21px;
-
   &:after {
-    position: absolute;
-    top: 2px;
+    position: relative;
+    top: 1px;
     right: 0;
-    width: 16px;
-    height: 16px;
-    /* Icon source https://www.iconfinder.com/icons/2561428/external_link_icon */
+    width: 12px;
+    height: 12px;
+    margin-left: 1px;
+    /* Icon source https://en.wikipedia.org/w/skins/Vector/images/external-link-ltr-icon.svg */
     content: url(/static/img/external-link.svg);
   }
 `
--- a/src/Documentation/Markdown/Markdown.js
+++ b/src/Documentation/Markdown/Markdown.js
@@ -104,6 +104,27 @@ CodeBlock.propTypes = {
   value: PropTypes.node.isRequired
 }
 
+const Link = ({ children, ...props }) => {
+  const externalLink = props.href.match(/^(\/\/|http(s)?:\/\/)/)
+  const showIcon =
+    externalLink && children && typeof children[0].props.children === 'string'
+
+  const modifiedProps = externalLink
+    ? { ...props, target: '_blank', rel: 'noopener nofollow' }
+    : props
+
+  if (showIcon) {
+    return <ExternalLink {...modifiedProps}>{children}</ExternalLink>
+  }
+
+  return <a {...modifiedProps}>{children}</a>
+}
+
+Link.propTypes = {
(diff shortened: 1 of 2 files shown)


shcheklein:
Looks great! 

A few comments:

Screen Shot 2019-11-29 at 9 19 08 AM


+ can we make it slightly smaller and closer to the text? similar to Wikipedia's one:

Screen Shot 2019-11-29 at 11 08 15 AM


Otherwise this is perfect 👌

src/Documentation/Markdown/Markdown.js
```diff
@@ -104,6 +104,27 @@ CodeBlock.propTypes = {
   value: PropTypes.node.isRequired
 }
 
+const Link = ({ children, ...props }) => {
+  const externalLink = props.href.match(/^(\/\/|http(s)?:\/\/)/)
+  const showIcon =
+    externalLink && children && typeof children[0].props.children === 'string'
+
+  const modifiedProps = externalLink
+    ? { ...props, target: '_blank', rel: 'noopener nofollow' }
```
jorgeorpinel: What if props already has a `target`?

jorgeorpinel:
p.s. Here's Wikipedia's link: https://en.wikipedia.org/w/skins/Vector/images/external-link-ltr-icon.svg

src/Documentation/Markdown/Markdown.js
iAdramelk: Right now we will rewrite it with `_blank` value and it's probably ok.

Other possible values are: `_self`, `_parent`,  `_top` and `framename`. `_parent` and `_top` will not work anyway. `framename` can work in some situations, but we don't have any frames right now (and it didn't sound secure to me to open external links in the iframes on our site). `_self` is somehow useful if we want to allow to manually prevent some external links from opening in the new window, but I'd want for real usecase examples before enabling it.

iAdramelk:
Updated icon to the wikipedia's one and placed in closer to text.

src/Documentation/Markdown/Markdown.js
jorgeorpinel: So if I write <a href="..." target="_self">link</a> it will overwrite with `_target`? I think it would be god to at least note this in the README to prevent someone in the future spending hours trying to figure out this one line of code :P

shcheklein: > So if I write <a href="..." target="_self">link</a> in Markdown

you probably should not be doing this though in Markdown (does it work with the Remark at all?). But I agree, may be we should check and do not overwrite it if it's set explicitly.

jorgeorpinel: Markdown supports all HTML AFAIK. It's necessary for specific tag params like `<a target`

jorgeorpinel: > does it work with the Remark at all?

Hm.. I tried and actually `<a>` tags don't get rendered as links for some reason. We use `<abbr>` and `<details>` a lot so I don't understand this behavior.

shcheklein: there are custom handlers for <abbr> and <details>. In general it's better to avoid using any HTML in Markdown

jorgeorpinel: So if our Markdown doesn't even support the `<a>` tag, probably this is OK and there's no need to check for `target` values before overwriting? If you agree please just resolve this.

shcheklein: resolved! if @iAdramelk decide that we need to check it we'll add it later. I think we are good for now.

shcheklein:
@jorgeorpinel is it good to merge?

PR #828, opened 2019-11-29
User Guide / Manage Experiments

### LINKED ISSUE #159: user-guide: add "folders" way of experimentation
Currently, we only mention branches as a way of experimentation (in Get Started), but it is not ideal and we've seen users use folders as an alternative. Need to mention it in our docs.

### LINKED ISSUE #816: guide: Managing Experiments section(s)
UPDATE: Skip to https://github.com/iterative/dvc.org/issues/816#issuecomment-731386585

Add a ~~User Guide~~ page that explains how to manage experiments:
 - by tags
 - by branches
 - by folders (check some details and ideas in #159)
 - hybrid (mixing tags/branches/folders)

Documenting first the existing methods may also help the discussions about adding new methods for experiment management (https://github.com/iterative/dvc/issues/2799), because my impression is that most of what is being discussed there can already be accomplished with the existing tools.

Another related issue: https://github.com/iterative/dvc.org/issues/159

dashohoxha:
Close #816, Close #159


mroutis:
@dashohoxha, the "manage experiments by directories" image could improve a lot by using a more significant naming.

What I've seen is that users create directories to try different hypothesis, not exactly different versions :thinking:

dashohoxha:
> the "manage experiments by directories" image could improve a lot by using a more significant naming
> 
> What I've seen is that users create directories to try different hypothesis, not exactly different versions

@mroutis I have used dummy names just to give an idea of how it should look like (and also to show that sub-experiments can be used as well, that is experiments that are based on other experiments).
I have done the same thing (using dummy names) for tags and branches as well, just to give an idea.

However I completely agree with you. If you (or someone else) could provide some more realistic example names for directories and subdirectories, I can fix that image.

dashohoxha:
I am aware that it is not complete or perfect, especially the pages of directory management and combined management. However this is the best that I could do. If someone has any ideas for improving it further this would be great.

shcheklein:
It looks great, @dashohoxha ! 

A few comments from the first pass I've done:

* can we make an image with directories the same style as tags and commits? they look better, may be since they have at least some color to them?
* Manage -> Managing - otherwise we have to rename all other UG sections: Updating ...., Managing ..., Versioning, etc
* I would say commits - is another way to do experiments. May be we can mention this somewhere?

I'll read it more carefully today.

dashohoxha:
> - can we make an image with directories the same style as tags and commits? they look better, may be since they have at least some color to them?

Unfortunately this is not easy with the tool that I am using (umlet).

> - Manage -> Managing - otherwise we have to rename all other UG sections: Updating ...., Managing ..., Versioning, etc

I used "Manage Experiments" as a shorter version of "How To Manage Experiments". It seems smother to read than "Managing Experiments". However I agree with your consistency argument.

> - I would say commits - is another way to do experiments. May be we can mention this somewhere?

This seems like the case of tags. If you do a single commit for each experiment, then why not put a tag for each commit? It is true that you don't need the tags to switch back to a previous commit (you can see the logs and find out the commit ID) but it is more easy with tags. Besides, the command `dvc metrics show` works with tags and branches, but does not work with commits, does it?

shcheklein:
> Unfortunately this is not easy with the tool that I am using (umlet).

nothing fancy, similar to what you have in the tags/branches would do. If it's not possible it's fine either. Not a deal breaker. I can help with that.

> However I agree with your consistency argument.

Thanks 🙏 

> Besides, the command dvc metrics show works with tags and branches, but does not work with commits, does it?

yep, that why there is a ticket for this https://github.com/iterative/dvc/issues/1691 - it's an important direction. Not sure we need a separate section though - it can easily be part of Commits and tags?

dashohoxha:
>  I can help with that.

Please do if you can. You can use [umletino](https://www.umletino.com/umletino.html) which is a fabulous tool. To be honest, I am not even sure what improvement you have in mind, so it it is better to give it a try yourself.

>  that why there is a ticket for this iterative/dvc#1691 - it's an important direction. Not sure we need a separate section though - it can easily be part of Commits and tags?

Somehow I have missed that discussion. But I see no resolution there and no consensus/agreement on how to proceed. So, it is not sure whether it will be implemented some day. My opinion is that we should not document something that is not implemented yet, we can update the docs later.

Regarding that discussion, it seems to me that what the user wanted to accomplish (making small parameter changes and checking the metrics for these parameter values) can be implemented more easily and cleaner with directories for each experiment.

Let's say that the user has a table with parameters and their values. He can write a script that for each parameter values creates a new experiment directory and (re)produces the results. Then he stores on the table all the results (metrics), removes all the experiment directories (cleanup), and commits on Git this table that contains the parameter values and the corresponding results. This is much cleaner than making a small commit for each parameter value and considering each commit as an experiment.

Regarding the other idea of limiting the output of `dvc metrics show -a -T` with a range of commits, this might be useful is some cases.

static/docs/user-guide/experiments/branches.md
```diff
@@ -0,0 +1,119 @@
...
+
+What we usually want is to completely replace the master branch with the
+experiment branch. Using `git merge` is not the best option in such a situation
+since it will usually result into a mixture between the two branches (the master
+branch and the experiment branch). Instead we should copy the branches, like
+this:
+
+```dvc
+$ git checkout bigrams
+$ git branch -c master old-master
```
shcheklein: add `branch`, `diff` and other commands that are no highlighted to the list here: https://github.com/iterative/dvc.org/blob/bd774bf39525bf4d33e143e7f7b00e1f6c744204/src/Documentation/Markdown/lang/dvc.js#L59

static/docs/user-guide/experiments/tags.md
```diff
@@ -0,0 +1,115 @@
...
+### Commit and tag
+
+At the end of each experiment we commit everything to Git and add a tag, like
+this:
+
+```dvc
+$ git add .
+$ git commit -m 'Baseline experiment'
+$ dvc commit
+$ git tag baseline-experiment
```
shcheklein: it is a lightweight tag. Why do we use it here as opposed to the regular one below? Are you sure that metrics show works well with it?

static/docs/user-guide/experiments/tags.md
```diff
@@ -0,0 +1,115 @@
...
+$ git tag
+$ git log --oneline
+```
+
+To go to a previous experiment we can do:
+
+```dvc
+$ git checkout baseline-experiment
+$ dvc checkout
+$ dvc repro evaluate.dvc
```
shcheklein: why do we need to run `repro` here?

static/docs/user-guide/experiments/tags.md
```diff
@@ -0,0 +1,115 @@
...
+move this experiment on top of the Git history. We can do it like this:
+
+```dvc
+# make master identical to the tag by reverting all the changes
+$ git revert -n baseline-experiment..HEAD
+$ git status
+$ git diff --cached
+$ git revert --continue
+
+# delete the old tag and add it to the current version
```
shcheklein: please split it into two blocks

move comments out and put them as some human readable explanation to the blocks

static/docs/user-guide/experiments/tags.md
```diff
@@ -0,0 +1,115 @@
...
+
+```dvc
+$ git checkout master
+$ dvc checkout
+$ dvc repro evaluate.dvc
+```
+
+### Move the best experiment on top
+
+Let's say that `baseline-experiment` has the best performance and we want to
```
shcheklein: Those are destructive, right? Let's put a note with ! emoji.

static/docs/user-guide/experiments/tags.md
```diff
@@ -0,0 +1,115 @@
...
+    auc.metric: 0.602818
+```
+
+### Check out a previous experiment
+
+Let's list first all the tags:
+
+```dvc
+$ git tag
+$ git log --oneline
```
shcheklein: show output here?

src/Documentation/sidebar.json
```diff
@@ -120,6 +120,29 @@
         "label": "Managing External Data",
         "slug": "managing-external-data"
       },
+      {
+        "label": "Managing Experiments",
```
shcheklein: thanks !

static/docs/user-guide/experiments/dirs.md
```diff
@@ -0,0 +1,97 @@
...
+[this interactive tutorial](https://katacoda.com/dvc/courses/basics/pipelines).
+
+## How it works
+
+If we have a directory named `experiment1/` which contains the pipeline of the
+first experiment, and we want to create another experiment on `experiment2/`,
+which is based on the first one, often it is as easy as:
+
+```dvc
+$ cp --reflink -R experiment1/ experiment2/
```
shcheklein: reflinks are not supported on majority of systems. So, it's not that easy to copy it w/o hitting some performance problems by copying data. We should think a bit more on what should we do here.

static/docs/user-guide/experiments/index.md
```diff
@@ -0,0 +1,22 @@
...
+months from now) can check out and inspect all the details of an experiment.
+
+You can use several ways to manage experiments, which are described on this
+section. Which one is more suitable for you depends on your preferences and also
+on the kind and complexity of your project.
+
+- [How to Manage Experiments by Tags](/doc/user-guide/experiments/tags)
+- [How to Manage Experiments by Branches](/doc/user-guide/experiments/branches)
+- [How to Manage Experiments by Directories](/doc/user-guide/experiments/dirs)
+- [How to Manage Experiments by Several Methods](/doc/user-guide/experiments/mixed)
```
shcheklein: by a Mix of Methods? Several methods sounds strange a bit

static/docs/user-guide/experiments/mixed.md
```diff
@@ -0,0 +1,26 @@
+# How to Manage Experiments by Several Methods
```
shcheklein: Mix Methods to Manage Experiments?

static/docs/user-guide/experiments/dirs.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Examples
+
+There is a very basic example of using directories for each experiment at the
+end of
+[this interactive tutorial](https://katacoda.com/dvc/courses/basics/pipelines).
+
+## How it works
+
+If we have a directory named `experiment1/` which contains the pipeline of the
```
shcheklein: except the pipeline - what else does this directory contain? what about code? do we copy the whole project?

static/docs/user-guide/experiments/tags.md
dashohoxha: This one is a "normal" tag, the other one is an annotated tag. It is just an example.
As far as I know `metrics show` works with all the tags.

static/docs/user-guide/experiments/tags.md
dashohoxha: Maybe to make sure that `dvc checkout` has retrieved all the correct data? I don't know.

static/docs/user-guide/experiments/tags.md
shcheklein: it's not even about what can be called normal. It's more about - why do we use a lightweight here and an annotated below? If there are some reasons then explanation is required when should I use what type.

static/docs/user-guide/experiments/tags.md
shcheklein: dvc status is better then and makes more sense?

static/docs/user-guide/experiments/tags.md
dashohoxha: What do you mean by "destructive"?

shcheklein: Yep, right. I probably wanted to put this comment to the branches way of doing experiments.

static/docs/user-guide/experiments/tags.md
dashohoxha: I don't think it will help to improve the explanation in this case.

shcheklein: I would improve for me as I read it for the first time. Reading it for the second time - I don't even understand why do we need `git log --oneline` everywhere. You are not saying a word. I meant btw, an output of the `git tag`:

```dvc
$ git tag
baseline
bnigrams
```

it would be easier to read and it is what the previous paragraph is about.

static/docs/user-guide/experiments/dirs.md
dashohoxha: > reflinks are not supported on majority of systems

"reflink" is a filesystem feature, and all the systems support filesystems with reflink. For example in Linux there are XFS, Btrfs, ZFS etc. The option "--reflink" here is a hint or a reminder that they need to use a filesystem that supports reflinks (like XFS), which they should be doing anyway since the start of the project.

static/docs/user-guide/experiments/index.md
dashohoxha: I don't actually mind, choose the wording that seems right to you.

static/docs/user-guide/experiments/dirs.md
shcheklein: see my bigger comment in the review. We can't rely on XFS, etc - for now 80-90% of systems are not configured to use them by default. And it won't change anytime soon. So, if we want to provide a meaningful solution to experiments management we should at least mention something about data, code and other things.

static/docs/user-guide/experiments/mixed.md
dashohoxha: That's fine too.

static/docs/user-guide/experiments/index.md
shcheklein: @jorgeorpinel what's your take on this and on the one https://github.com/iterative/dvc.org/pull/828#discussion_r354483893 here? what the best way to word it?

static/docs/user-guide/experiments/dirs.md
dashohoxha: I think this is too project specific and it is impossible to explain such details on a section like this.
The best that can be done is to have a concrete example somewhere else and to link it from the section of the examples.

shcheklein: there are a few recommendations we can make - clean data before copying? use a simple JSON conf? use DVC-files that are setup properly to be relative.

Those ^^ are general problems. Without giving some hints at least it's not very actionable. People who can come with a solution to those problems don't even probably need a section like this (and I can send you a few names).

static/docs/user-guide/experiments/mixed.md
shcheklein: 👍

dashohoxha:
Regarding the "dirs" page, I agree that in practice there are certain "tricks" that are needed to make it work, but I don't see how they can be explained in a meaningful way without a concrete example.

dashohoxha:
By the way, the new command `dvc experiment` that is being discussed might be a good way to encapsulate or hide most of the tricks and details that are needed for the case of experiment-by-directory to work. This way the user will not have to worry about them and we will not have to explain them :)

static/docs/user-guide/experiments/tags.md
dashohoxha: I don't think there is any reason. They are both tags and both of them can be used.

static/docs/user-guide/experiments/tags.md
dashohoxha: You are right `dvc status` would make more sense.

static/docs/user-guide/experiments/tags.md
dashohoxha: Both `git tag` and `git log --oneline` are ways to check what tags are available. Different from `git tag`, `git log --oneline` also shows the position of the tags in the history of commits and their order, which one is first and which one is last.

static/docs/user-guide/experiments/dirs.md
dashohoxha: > 80-90% of systems are not configured to use them by default

That's right, but they are not configured to use DVC by default either. Installing and using XFS is actually much easier than installing and using DVC.
What I am trying to say is that if someone is working on a data project, he can and should install and use XFS (or some other deduplicating filesystem). This is a basic requirement.

static/docs/user-guide/experiments/dirs.md
dashohoxha: > there are a few recommendations we can make - clean data before copying? use a simple JSON conf? use DVC-files that are setup properly to be relative.

All these seem common sense to me. Anyway I don't see how they can be explained without having an example as a reference.
Maybe someone can give it a try and let's see whether they make sense.

shcheklein: Since there is no understanding from our end on how to give a guidance let's remove this page then for now and wait until the ticker on DVC core is resolved and wait for someone to take the #159.

shcheklein: > All these seem common sense to me.

If they are common sense why would it be so hard to explain/mention them?

static/docs/user-guide/experiments/dirs.md
shcheklein: It's very very far from the reality. There are a lot (majority?) of cases when you don't have a choice. Your argument applies only to freelancers and students.

static/docs/user-guide/experiments/tags.md
shcheklein: This explanation makes it 10x better! You see, you can do it :) And I'm not kidding - I didn't know the difference between those related to the tags.

I still have some questions though - would log include all commits? If you have a lot of commits (very regular thing to have in real project) tags be lost.

Is there an option for the `git tag` command to sort output?

I would still prefer to simplify it tbh. And put some output - it will make way easier to understand and read. At least you have one reader who is saying that it would be easier to read. Don't rely on your opinion here.

static/docs/user-guide/experiments/tags.md
shcheklein: So, how about branches part? In that place where we do force push and stuff - let's put a comment there?

static/docs/user-guide/experiments/tags.md
shcheklein: 🤝

static/docs/user-guide/experiments/tags.md
shcheklein: so, why use both? let's simplify. Though, there are some guidelines behind those. I think people tend to use lightweight tags for things that they won't be sharing/saving. May be we can use that difference as a guideline? Not sure that we need put so many details - I would probably just keep simple version in both cases if it works fine.

static/docs/user-guide/experiments/mixed.md
jorgeorpinel: Maybe ** How to Manage Experiments with a Mix of Methods**
or use the word "Combination", as in the intro paragraph.

static/docs/user-guide/experiments/index.md
jorgeorpinel: Maybe ** How to Manage Experiments with a Mix of Methods**
or use the word "Combination", as in that doc's intro.

static/docs/user-guide/experiments/branches.md
```diff
@@ -0,0 +1,119 @@
...
+### Commit and branch
+
+Let's say that we are working on the branch `master` and at the end of the
+experiment we want to save it on a branch named `unigrams`. We can do it like
+this:
+
+```dvc
+$ git commit -am 'Evaluate'
+$ dvc commit   # just to make sure all the data is committed
+$ git checkout -b unigrams
```
mroutis: why not `git branch unigrams`?

static/docs/user-guide/experiments/branches.md
```diff
@@ -0,0 +1,119 @@
...
+
+### New experiment based on another one
+
+Suppose that we want to start a new experiment based on another one, instead of
+starting from `master`. We can switch first to that branch and then start a new
+experiment on top of it:
+
+```dvc
+$ git checkout unigrams
+$ dvc checkout
```
mroutis: why not `git checkout --branch bigrams unigrams` (go to experiment of brigams based on the unigrams)

static/docs/user-guide/experiments/branches.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+Switching back to `master`:
+
+```dvc
+$ git checkout master
+$ dvc checkout
+```
+
+In any case the command `dvc repro` should not have to re-run anything and
```
mroutis:
```suggestion
In any case, the command `dvc repro` should not have to re-run anything and
```

static/docs/user-guide/experiments/branches.md
mroutis: what's the `-c` and `-C` for?

I couldn't follow up with a lot of branching going on :sweat_smile: would be great if you could add some comments

static/docs/user-guide/experiments/branches.md
jorgeorpinel: Actually, `git branch` is pretty new and we don't use it in the docs. But we do use `git checkout` a lot. I would open a separate terminology-focused issue to decide whether to change all instances or not.

jorgeorpinel: Sorry I was confused about `git branch` being new lol. I was thinking of some new commands. But still we use `git checkout -b` a lot already so this seems consistent with existing docs, and is correct right?

static/docs/user-guide/experiments/branches.md
jorgeorpinel: This would depend on your previous suggestion, https://github.com/iterative/dvc.org/pull/828#discussion_r359003964 ? Perhaps you can replace these 2 comments for a single one for the full thing? 🙂

shcheklein:
closing this as stale

Issue #757, opened 2019-10-29
server: better way to handle HTTP redirects 


shcheklein:
Every time we move a page to a new URL we need to create a proper redirect.

The only way to this now is to create a 4-5 lines entry in our custom `server.js`. This is def. not scalable. We need to find an easier way to handle this. 

The first solution that comes to my mind is a separate file with a list of simple expressions, may be converter functions.

jorgeorpinel:
> separate file with a list of simple expressions...

And maybe add a note with the date the redirect was added (can also be checked with GitHub Blame but it's not always straightforward) for periodic cleaning up of this list.

Another option is to use DNS for this, not the app.

Suor:
@jorgeorpinel I don't think those should be cleaned up unless they clash with something.

Another thing - do not forget to use HTTP code 301 for moved things, not 302.

jorgeorpinel:
Actually @Suor we are having some trouble with 301 "permanent" redirects that we now changed our minds on... They're cached by the browser and can't be changed again by us until that cache expires a long time later. **302 Found** may be better for us since we are not so sure they are so permanent. See https://github.com/iterative/dvc.org/issues/768#issuecomment-549456704

shcheklein:
I think they are still 301s @jorgeorpinel . Even though we change the end location we don't change the fact of the redirect. So, unless we don't find a good solution for caching, I would keep it 301. But I haven't googled best practices around dealing with cache yet.

shcheklein:
Ok, reading the link now in that comment :)

shcheklein:
If `no-cache` policy does not work probably we should indeed use 302s, but I would google about the consequences of this for search engines and whatnot.

jorgeorpinel:
Well, they both have impact. 301 is better but we need to be sure the redirect is actually permanent and since we're in an active development phase... I think we should not use 301s as much. Many of our redirects are temporary, which is what 302 is for. As for SEO effects:

> 302 redirect informs search engines to continue indexing old pages. It may impact importantly your SEO results. When you are optimizing the new page URL, make certain to be vigilant about not having duplicate content on both old and new pages at a similar time. The temporary redirect is very useful when you bring your website visitors back to the old page after some time.

https://www.firstrankseoservices.com/blog/effects-of-301-and-302-redirect-on-seo/

shcheklein:
[...]

PR #834, opened 2019-12-04
Docs SSR

### LINKED ISSUE #642: sidebar blinks on next/prev buttons click
Sep-24-2019 16-57-09

iAdramelk:
- Make docs page indexable by search engines.
- Add proper 404 errors for non-existing pages (partial fix for #757).
- Fixes #642.

ToDo:

~- [ ] Add smooth scroll for anchors~
~- [ ] Add preloaders on page change (not sure if we want them now)~


shcheklein:
@iAdramelk I think we are fine w/o either of those for now. In case of network failure it will show an error, right?

if smooth scroll is not something hard to do - let's do this

please, resolve the conflicts - merge/rebase

iAdramelk:
@shcheklein 

Smooth scrolling should have been easy, but wasn't for some reason. System method to do it using `window.hashchange` didn't work because Next Router considering it as navigation event and adds it's logic to it. And Next Router's special event for it `hashChangeComplete` for some reason didn't fire. Tried to debug it yesterday for couple of hours, but was not successful.

Will text for network failure, but it should work.

shcheklein:
> but wasn't for some reason.

yep, sounds like a strange problem to have ... let's not waste time on this - I think it's totally ok w/o smooth anchors

if it works fine across different browsers let's merge

src/consts.js
```diff
@@ -13,3 +13,5 @@ export const META_SOCIAL_IMAGE = `https://${WEBSITE_HOST}/static/social-share.pn
 
 export const FORUM_URL = `https://discuss.${WEBSITE_HOST}`
 export const BLOG_URL = `https://blog.${WEBSITE_HOST}`
+
+export const PAGE_DOC = '/doc'
```
jorgeorpinel: Maybe `DOC_PATH`?

pages/doc.js
```diff
@@ -25,216 +24,146 @@ import { media } from '../src/styles'
...
 
-    if (element) {
-      const headerHeight = document.getElementById(HEADER).offsetHeight
-      const elementBoundary = element.getBoundingClientRect()
-      const rootElement = document.getElementById(ROOT_ELEMENT)
-      rootElement.scroll({ top: elementBoundary.top - headerHeight })
+    if (res.status !== 200) {
+      return {
+        errorCode: res.status
+      }
```
jorgeorpinel: I'd just return a 504 here.

pages/doc.js
```diff
@@ -25,216 +24,146 @@ import { media } from '../src/styles'
...
+    return {
+      item: item,
+      headings: parseHeadings(text),
+      markdown: text
+    }
+  } catch (e) {
+    return {
+      errorCode: 404
+    }
   }
```
jorgeorpinel: Hmmm. what kind of errors can `fetch` throw that would get caught here? Can't easily tell from https://github.com/matthew-andrews/isomorphic-fetch

pages/doc.js
jorgeorpinel: Maybe with an error message including the `res.status`.

src/consts.js
jorgeorpinel:
```suggestion
export const DOC_PATH = '/doc'
```

jorgeorpinel:
> Make docs page indexable by search engines.

Sorry, can you briefly explained how this is achieved by this PR? Sounds great but I'm not easily getting it from browsing through the file changes.

src/Documentation/SidebarMenu/SidebarMenu.js
```diff
@@ -198,7 +196,6 @@ const SectionLink = styled.a`
   min-height: 26px;
   padding-bottom: 5px;
   padding-left: 15px;
-  cursor: pointer;
```
shcheklein: do you know why was it here before?

src/Layout/index.js
```diff
@@ -16,13 +17,24 @@ class Layout extends Component {
...
+    Router.events.off('routeChangeComplete', this.handleRouteChange)
   }
 
   UNSAFE_componentWillReceiveProps(nextProps) {
     const { router } = nextProps
     this.isDocPage = router.pathname.split('/')[1] === 'doc'
   }
 
+  handleRouteChange = () => {
+    // Without timeout GA will send old page title
```
shcheklein: is it a recommended way to prevent this? looks fragile a bit.

shcheklein: can we just provide an argument to the logPageView?

src/PromoSection/index.js
```diff
@@ -10,7 +11,7 @@ function goToDocGetStarted() {
 
 function goToFeatures() {
   logEvent('promo', 'features')
-  window.location = '/features'
+  Router.push('/features')
```
shcheklein: minor: put an empty line like in get-started .. or remove it everywhere :)

shcheklein:
Breaks for me in a peculiar way:

Screen Shot 2019-12-05 at 3 29 02 PM

Open docs -> navigate to features -> use back button to go to the previous page - looks like a way to reproduce this.

pages/doc.js
```diff
@@ -1,21 +1,20 @@
...
+import PropTypes from 'prop-types'
+import Error from 'next/error'
+import Router from 'next/router'
 // components
 import Page from '../src/Page'
 import { HeadInjector } from '../src/Documentation/HeadInjector'
 import Hamburger from '../src/Hamburger'
 import SearchForm from '../src/SearchForm'
 import SidebarMenu from '../src/Documentation/SidebarMenu/SidebarMenu'
-import Loader from '../src/Loader/Loader'
```
shcheklein: if we remove them here, let's remove corresponding components?

pages/doc.js
```diff
@@ -1,21 +1,20 @@
...
-import Page404 from '../src/Page404'
 import Markdown from '../src/Documentation/Markdown/Markdown'
 import RightPanel from '../src/Documentation/RightPanel/RightPanel'
 // utils
 import fetch from 'isomorphic-fetch'
 import kebabCase from 'lodash.kebabcase'
-// constants
-import { HEADER } from '../src/consts'
```
shcheklein: is it used anywhere else? remove?

iAdramelk:
@jorgeorpinel 

> Sorry, can you briefly explained how this is achieved by this PR? Sounds great but I'm not easily getting it from browsing through the file changes and clicking around the nav inside https://dvc-org-pr-834.herokuapp.com/doc seems completely dynamic to me.

Basically how it worked before:

1. All `/docs` URLs were redirected to `pages/doc.js`.
2. `doc.js` was rendered on server with blank content (because `componentDidMount` didn't fire on the server) and was sent to browser this way.
3. In browser, after page was loaded, `componentDidMount` did fire. It was checking current url, and fetching the corresponding md-file from the server then.
4. After that, we were adding a listener to the window.location changes and were implementing navigation by hand.

What changed:

1. `componentDidMount` was replaced with `getInitialProps`. Because Next.js can run `getInitialProps` on the server, it can now populate content zone and menu with actual data before sending the page to the client (try to disable JavaScript in the browser and load some page from docs).
2. Our custom navigation and routing logic was replaced with already existing Next.js components `Link` and `Router`. They allow us to navigate between pages on the client side without the need to download everything from the server again (like our custom logic did before). Basically, they change window.location, and then fire `getInitialProps` if needed. They also give us hooks to attach to page transition events and errors if we need to.
3. Also, I implemented some checks for docs pages in `getInitialProps`, and if md-file didn't exist, it returns standard Next.js error page as all other routes except `/docs` are doing now.

pages/doc.js
iAdramelk: `fetch` is a promise. It can return either `Promise.resolve` or `Promise.reject`. Resolve is the thing that we will receive in the most cases. It have `response` object and we can check it's `status` field. But if, for example, your internet connection is lost, the browser will not resolve it, but  will instead throw error which we are catching there.

Try to set connection to offline in Network tab of Developers Tools and then try to click on the internal link, it will throw error.

pages/doc.js
iAdramelk: Why not leave actual code there? It can help us debug things. Do you see some case there showing real code can be bad?

src/Documentation/SidebarMenu/SidebarMenu.js
iAdramelk: I think that probably for some reason this link didn't have `href` attribute in some of the previous versions and because of that it's cursor displayed incorrectly. But now all links have `href`'s, so it should be ok to remove it.

src/Layout/index.js
iAdramelk: I refactored layout a little, and replaced timeout with `requestAnimationFrame`, but I afraid we still need it. Basically Next's `Head` component uses promise inside (https://github.com/zeit/next.js/blob/18a9c7e371efc4c487f9c3599c3211ce30009d6c/packages/next/client/head-manager.js#L14), and, because of that, page content renders and `componentDidUpdate` fires before the `<head>` content is updated. Adding rAF there will add GA task to the next loop of Event Loop after the head update.

src/PromoSection/index.js
iAdramelk: Done

pages/doc.js
iAdramelk: Removed them for now

pages/doc.js
iAdramelk: From that I see it is used in the `TopMenu` and in the `Tooltip` components.

iAdramelk:
> Open docs -> navigate to features -> use back button to go to the previous page - looks like a way to reproduce this.

Found the bug's source after a couple of hours. Basically we this code in HeadInjector file:

```
<link
  rel="stylesheet"
  href="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.2/dist/cdn/docsearch.min.css"
/>
<link
  rel="stylesheet"
  href="https://cdn.jsdelivr.net/npm/perfect-scrollbar@1.4.0/css/perfect-scrollbar.min.css"
/>
<script
  type="text/javascript"
  src="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.2/dist/cdn/docsearch.min.js"
/>
```

And before is was not a problem because all navigation in the top and bottom menu required page reload and this code was available from the start. But now we navigate between pages without reloading the pages and this code didn't have time to load before we start calling it and this cause different types of problems.

For now I moved them to the `_document.js` and it fixed the bug.

pages/doc.js
shcheklein: 👍

pages/doc.js
shcheklein: 👍

src/Documentation/SidebarMenu/SidebarMenu.js
shcheklein: 👍

src/PromoSection/index.js
shcheklein: 👍

shcheklein:
@iAdramelk LGTM! Anything else to be fixed or ready to merge? Have you checked it on different devices/browsers?

iAdramelk:
@shcheklein checked on mobile right now and there is couple small things that we need to fix. (For one menu is not closed automatically after navigation as it should)

iAdramelk:
Fixed error on mobile. Tested in Edge. We can merge it.

jorgeorpinel:
Thanks for the [explanation](https://github.com/iterative/dvc.org/pull/834#issuecomment-562527528) Alex. Very interesting. And sorry took me a long time to get back to this haha. I see it's merged. I may have a couple questions left...

pages/doc.js
jorgeorpinel: Ah, this fetch is for the internal document? Not some sort of gateway? In that case I think returning actual code makes sense indeed. Although a generic code with a message explaining what happened (and the source error code) would make ever more sense I think. This is merged and works though, so whatevs.

pages/doc.js
jorgeorpinel: Hmmm OK but why 404? I see this is changed though... Will re-ask.

pages/doc.js
```diff
@@ -25,216 +24,149 @@ import { media } from '../src/styles'
...
-    this.setState(prevState => ({
-      isMenuOpen: !prevState.isMenuOpen
-    }))
+    return {
+      item: item,
+      headings: parseHeadings(text),
+      markdown: text
+    }
+  } catch {
+    window.location.reload()
```
jorgeorpinel: Can't this go into an infinite loop? (Blocked by the browser probably.)

> Continuation of https://github.com/iterative/dvc.org/pull/834#pullrequestreview-327869118

src/Diagram/index.js
```diff
@@ -13,16 +16,18 @@ import { Element } from 'react-scroll'
...
-      <span>Learn&nbsp;more</span>
-      <img
-        src="/static/img/learn_more_arrow.svg"
-        width={18}
-        height={18}
-        alt=""
-      />
-    </a>
+  <LearnMoreArea>
+    <NextLink href={href}>
```
jorgeorpinel: Does NextLink work without js?

jorgeorpinel: I see it does but without js the nav bar doesn't expand. Unrelated to this PR I know, but should we care about this? Cc @shcheklein

src/Documentation/Markdown/Markdown.js
```diff
@@ -209,14 +218,18 @@ export default class Markdown extends React.PureComponent {
...
-          <Button onClick={() => onNavigate(next)} disabled={!next}>
-            <span>Next</span>
-            <i className="next" />
-          </Button>
+          <NextLink href={PAGE_DOC} as={prev} passHref>
+            <Button disabled={!prev}>
+              <i className="prev" />
+              <span>Prev</span>
+            </Button>
+          </NextLink>
```
jorgeorpinel: ...or maybe the collapsed navigation without js IS related to this PR?

src/Diagram/index.js
shcheklein: Working w/o JS is out of scope for this. The good question is if we need to generate a sitemap or something for the case when we don't have a link to some page inside and search engines won't be able to find it?

src/Documentation/Markdown/Markdown.js
shcheklein: I think the previous version was not working w/o JS at all, btw

src/Diagram/index.js
jorgeorpinel: Right, engined won't be able to crawl if the nav element doesn't SSR. A sitemap sounds like a good solution too but maybe it's easy to pre-render nav? Could open a separate `engine` issue for this.

shcheklein: Yep, let's open and see what @iAdramelk has to say about this.

jorgeorpinel: OK, moving to #856.

pages/doc.js
iAdramelk: If page loads again then yes, but this error usually means that the internet is down and we will just show browser error page there. Maybe there are cases then this error is thrown in other situations, but I don't know about them.

PR #839, opened 2019-12-05
run: add concept of well-behaved commands

mroutis:
Fix https://github.com/iterative/dvc/issues/2542

Shout-out to @Suor for the initial draft: https://github.com/iterative/dvc/issues/2542#issuecomment-545010319

Any feedback on how to improve it is appreciated :slightly_smiling_face:


static/docs/user-guide/index.md
```diff
@@ -27,6 +27,9 @@ guides related to contributing to this
   [DVC-file](/doc/user-guide/dvc-files-and-directories) dependencies.
 - [Managing External Data](/doc/user-guide/managing-external-data) explains both
   external outputs and external cache.
+- [Writing well-behaved commands](/doc/user-guide/well-behaved-command) goes
+  through the few technicalities that you need to take into account when using
+  DVC to run your commands.
```
jorgeorpinel: Agree with Ivan. No need for a whole new guide for this note, let's just add it in the **Description** section of the `dvc run` cmd ref (It can be a paragraph or even an H3 mini section. But maybe also a shorter note in other relevant places where `dvc run `is introduced? E.g. Get Started or some Tutorials

static/docs/user-guide/well-behaved-command.md
```diff
@@ -0,0 +1,12 @@
...
+DVC is simple to use, you only need to wrap your commands with `dvc run`, and
+define your dependencies and outputs.
+
+To prevent unexpected behaviors, ideally, your commands should follow some
+principles:
+
+- Read exclusively from specified dependencies
+- Write exclusively to specified outputs
+- Completely rewrite the outputs (do not append)
+- Stop reading and writing when the command exits
```
jorgeorpinel: End each full sentence with period `.` please.

dashohoxha:
Maybe this should be part of a more general section about Stages and Pipelines (since it seems too short to be a standalone page).
However I am not sure, maybe such a section would be superfluous since stages and pipelines are already explained on Get Started and on the tutorials.

static/docs/command-reference/run.md
```diff
@@ -217,3 +217,18 @@ $ tree .
 │   └── test.dvc
 └── test.txt
 ```
+
+### Well-behaved commands
```
shcheklein: let's move it to the description section

mroutis:
@shcheklein , this is unexpected, do I need to do something else to include it under the section?

mroutis:
@shcheklein , I'm using an H3, maybe those are not rendered correctly?

shcheklein:
@mroutis it's just physically located under `## Options` not `## Description`. Just move it up?

mroutis:
@shcheklein , my bad!!! I was looking at the official (production) docs and couldn't see the mistake :see_no_evil: 

it should be fixed by now

static/docs/command-reference/run.md
```diff
@@ -52,6 +52,21 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
+
+DVC is simple to use, you only need to wrap your commands with `dvc run`, and
+define your dependencies and outputs.
+
+However, to prevent unexpected behaviors, ideally, your commands should follow
+some principles:
+
+- Read exclusively from specified dependencies.
+- Write exclusively to specified outputs.
+- Completely rewrite the outputs (do not append).
```
Suor:
```suggestion
- Completely rewrite the outputs (i.e. do not append or edit).
```

shcheklein:
@Suor 

> i.e. produce same results when run with the same deps later.

We should keep in mind though that _exactly_ same result is hard to produce in a lot of cases, even if seed is predefined. It might depend on hardware, for example. But even on the same hardware as far as I remember there are no guarantees.

The point is that pipeline that captures data properly and dependencies and does not define seed (i.e. depends on random generator) can be quite useful as well.

Suor:
@shcheklein so it needs to be "approximately reproducible"? As far as I see it it is only the issue for model training, data split, transformation and preparation steps might be done pure.

mroutis:
@Suor  might taking another look? I added a note that ideally the commands should be deterministic

shcheklein:
@Suor yes, I would consider an "approximately" reproducible script a well-behaved script in our sense. Not sure what is the better way to put a note about it.

shcheklein:
no new comments, previous comments are resolved - merging this. Other improvements and changes - please create a new ticket :)

jorgeorpinel:
Sorry, late to this discussion but I'm planning to revise the text anyway so here are my comments:

> Should we say about being a pure function or in other words controlling entropy sources?

> pipeline that captures data properly and dependencies and does not define seed can be quite useful as well... I would consider an "approximately" reproducible script a well-behaved script in our sense

I agree with Ivan here. Why should we encourage people to limit their data processing? What if they need to introduce some randomness into it? The pipeline itself is still reproducible, even if the results are not always the same. The graph is what matters most, no? Users can chose whether they want things to be completely deterministic.

Also I think the language we're using here is too scientific whereas our docs try to be public-friendly. Anyway, will make some changes in a new PR and tag y'all.

static/docs/command-reference/run.md
```diff
@@ -52,6 +52,26 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
+
+However, to prevent unexpected behaviors, ideally, your commands should follow
+some principles:
+
+- Read exclusively from specified dependencies.
+- Write exclusively to specified outputs.
+- Completely rewrite the outputs (i.e. do not append or edit).
+- Stop reading and writing when the command exits.
+
+To guarantee reproducibilty, your command should be
```
jorgeorpinel: reproducibility

static/docs/command-reference/run.md
```diff
@@ -52,6 +52,26 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
+[deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm) (i.e. it
+must produce the same results given the same inputs/dependencies).
+
+Have in mind what brings entropy to your command (e.g. random generators, time,
+hardware, etc.) and try to minize it (e.g. fix seeds).
```
jorgeorpinel: minimize

static/docs/command-reference/run.md
mroutis: @jorgeorpinel , thanks, will keep it in mind.

PR #840, opened 2019-12-06
Update config.md for more clarity

maykulkarni:
Added the definition of what a `section` is as it was a little confusing.

### DIFF
--- a/static/docs/command-reference/config.md
+++ b/static/docs/command-reference/config.md
@@ -55,9 +55,8 @@ corresponding config file.
 
 ## Configuration sections
 
-Sections are essentially a collection of options that define a particular 
-setting like `core`, or `remote`. These are written to the project config file 
-(in `.dvc/config` by default) by this command, and support the options below:
+The following config sections are written by this command to the project config 
+file (in `.dvc/config` by default), and they support the options below:
 
 ### core
 
--- a/static/docs/command-reference/config.md
+++ b/static/docs/command-reference/config.md
@@ -55,9 +55,9 @@ corresponding config file.
 
 ## Configuration sections
 
-Sections are essentially a collection of options that define a particular setting. 
-These `name` parameters can be used with `dvc config`, or in the <abbr>project</abbr> 
-config file (`.dvc/config`).
+Sections are essentially a collection of options that define a particular 
+setting like `core`, or `remote`. These are written to the project config file 
+(in `.dvc/config` by default) by this command, and support the options below:
 
 ### core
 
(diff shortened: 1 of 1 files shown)


static/docs/command-reference/config.md
```diff
@@ -55,8 +55,9 @@ corresponding config file.
 
 ## Configuration sections
 
-These are the `name` parameters that can be used with `dvc config`, or the
-sections in the <abbr>project</abbr> config file (`.dvc/config`).
+Sections are essentially a collection of options that define a particular setting. 
+These `name` parameters can be used with `dvc config`, or in the <abbr>project</abbr> 
+config file (`.dvc/config`).
```
jorgeorpinel: I like the intention of this PR. The reason we call them "sections" here, and not "options" is that this doc has a previous **Options** H2 header for the flags that the `dvc config` command accepts. This is a standard header we have in all our cmd refs. So in the description I just wanted to causally explain why "sections", and the reason is that DVC config files will also have sections like `[remote ...]` or `[core]` when this command is used.

> Sections are essentially a collection of options...

I agree this paragraph could be improved but I find this particular sentence also confusing. How about something like "The following config _sections_ are written to  the <abbr>project</abbr> config file (in `.dvc/config` by default) by this command, and they contain the following options:" ?

I would also add the word section to the **cache** description below:

> **cache**
> A DVC project cache is ... for more details.) This section contains the following options:

and remove the first sentence in the **state** desc:

> **state**
> ~~State config options.~~ See

jorgeorpinel:
Sorry for the delay @maykulkarni. Please lmk if you can still work on this. No worries either way (:

static/docs/command-reference/config.md
maykulkarni: @jorgeorpinel Hmm, I agree, but my primary intention was to *define* what section is, what you're suggesting is *what we can do* with it. We can maybe merge both? WDYT?

jorgeorpinel: I don't think we need to define what a configuration section is much 🙂 Maybe just mention where they exist (in the config file) like in my suggested sentence:

> The following config sections are written by this command to the project config file (in `.dvc/config` by default), and they support the options below.

jorgeorpinel: p.s. thanks for the updates, I can finish and merge this PR tomorrow if you'd like.

maykulkarni: @jorgeorpinel makes sense

Issue #849, opened 2019-12-09
links to subsections don't work on contributing page


Suor:
Links to subsections here https://dvc.org/doc/user-guide/contributing give 404

shcheklein:
Also, I'm getting this:

<img width="945" alt="Screen Shot 2019-12-09 at 3 55 36 PM" src="https://user-images.githubusercontent.com/3659196/70483107-5ca8ad00-1a9c-11ea-9ec5-3d325678cbe8.png">

if click back from the 404

shcheklein:
Closed by #853 I'll create a separate issue for the height calculation if needed.

PR #853, opened 2019-12-10
Fix client navigation for links with hashes

iAdramelk:
Fix for 404 errors in #849. Error was with our custom 404 page logic - we was checking if the page exists by matching it's url with the sidebar.json content without removing the hash.


iAdramelk:
Can't reproduce the bug with the wrong height on back button.

PR #864, opened 2019-12-14
cmd ref: add checkout --relink option, improve checkout desc.

Suor:


static/docs/command-reference/checkout.md
```diff
@@ -95,6 +95,10 @@ be pulled from remote storage using `dvc pull`.
   remove files that don't match those DVC-file references or are missing from
   cache. (They are not "committed", in DVC terms.)
 
+- `--relink` - recreates links or copies for all checked out files even ones
```
shcheklein: move it after `-R` in this section. At least after common the `h`, `q`, `v`

Suor: I don't understand the logic. We have checkout specific options now, generic later. You want to mix it? I.e. specific, generic, specific again.

Suor:
> I wonder if it would be valuable to add a sentence or two into the description that clarifies the semantics (that we don't relink by default and it's user's responsibility)

Higher on same page: "For any data files where the checksum doesn't match their DVC-file entry, the
  data file is restored from the cache. The link strategy ..."

So `--relink` description builds on top of that: "recreates links ... even ones with unchanged checksums"

static/docs/command-reference/checkout.md
shcheklein: my bad, sorry :) but it makes sense to move it after `-R` to keep them in the same order as we have them in CLI.

Suor: What do you mean by CLI? If it's the usage section at the top then it has completely different order anyway, we'll need to reorder all of them either here or in usage section.

If you mean `dvc checkout --help` then it also has completely different order )

shcheklein: usage section is in the same exact order as CLI, right? CLI itself is consistent.

In docs we don't put `h, q v` section on the top since they do not provide value (at some point we didn't have them at all described). Besides those, order should be the same as in CLI. And it's a bug that `-f` is out of order now. Most likely we have a ticket to fix them all. So, would really appreciate if we reoder `--relink` (as well as `-f` if it's not too hard :) ).

shcheklein: @Suor are you going to address this or I can merge and fix it after?

Suor: I was going to fix it in all three places, but got caught up with other stuff. So let's make it wait until tomorrow.

shcheklein: 👍 sure, np! thanks.

Suor:
Reordered it everywhere to be from common to rare options: `-d`, `-R`, `-f`, `--relink`.

shcheklein:
@jorgeorpinel @Suor do you think it makes sense to include a link to the Large Files Optimization - a section about link and different cache types?

static/docs/command-reference/checkout.md
```diff
@@ -6,7 +6,7 @@ DVC-files.
 ## Synopsis
 
 ```usage
-usage: dvc checkout [-h] [-q | -v] [-d] [-f] [-R]
+usage: dvc checkout [-h] [-q | -v] [-d] [-R] [-f] [--relink]
```
jorgeorpinel: It's

```suggestion
dvc checkout [-h] [-q | -v] [-f] [--relink] [-d] [-R]
```

when you run `dvc checkout -h`.

static/docs/command-reference/checkout.md
```diff
@@ -95,6 +95,10 @@ be pulled from remote storage using `dvc pull`.
   remove files that don't match those DVC-file references or are missing from
   cache. (They are not "committed", in DVC terms.)
 
+- `--relink` - recreates links or copies for all checked out files even ones
+  with unchanged checksums. This ensures that link types of all the files in a
+  workspace match configured `cache.type`.
```
jorgeorpinel: - `--relink` - recreates links or copies for all checked out files even unchanged ones. This ensures the link types of all the files in the workspace match the project's [`cache.type`] (/doc/command-reference/config#cache).

jorgeorpinel:
> do you think it makes sense to include a link to the Large Files Optimization

@shcheklein I think it would make more sense to mention this option **in** and link **from** https://dvc.org/doc/user-guide/large-dataset-optimization to here.

shcheklein:
@jorgeorpinel 

>  it would make more sense to mention this option somewhere in and link from https://dvc.org/doc/user-guide/large-dataset-optimization to here.

makes sense, agreed

static/docs/command-reference/checkout.md
jorgeorpinel: See also iterative/dvc/issues/2983.

static/docs/command-reference/checkout.md
Suor: I've already changed it in dvc, so it's:
```
usage: dvc checkout [-h] [-q | -v] [-d] [-R] [-f] [--relink]
                    [targets [targets ...]]
```
there noe.

static/docs/command-reference/checkout.md
Suor: Autowrapped by precommit hook, I can't change it. 

The text is tailored to whatever we have above. There are two mentions of checksums before, here we refer to them again. So not "unchanged", but "unchanged checksums".

static/docs/command-reference/checkout.md
jorgeorpinel: Ah OK, cool. Thanks.

static/docs/command-reference/checkout.md
jorgeorpinel: I meant that the paragraph I suggested above may not be wrapped properly.

I see no need to talk about checksums in this paragraph. The point is that unchanged files also get relinked, no?

Suor: Unchanged and unchanged checksums are different, e.g. a copy and a symlink may have the same checksum, but are different things so might be considered changed. We never check whether a file/dir is changed, only if checksum is changed, this is mentioned a few times before in the text. 

Talking about relinking unchanged files will be both incorrect and confusing, i.e. why do I ever need to relink whatever is unchanged? So we don't talk about unchanged files anywhere on this page, we say that we checkout outputs that have checksums not matching their stage desc, i.e. changed, and `checkout --relink` checkouts outputs with matching/unchanged checksums.

jorgeorpinel: Hmm I see what you mean @Suor. I think the sentence you're mainly referring to is:

> For any data files where the checksum doesn't match their DVC-file entry, the data file is restored from the cache. 

So maybe the text in question would be best as `...for all data files in the workspace, including the ones normally checked out by this command, as well as existing ones that already have matching checksums in current DVC-files`

```
- `--relink` - recreates links or copies from cache to workspace
  for all data files in the workspace, including the ones normally
  checked out by this command, as well as existing ones that
  already have matching checksums in current DVC-files. This
  ensures the link types of all the files in the workspace match
  the project's [`cache.type`](/doc/command-reference/config#cache).
```
(Please check wrapping.)

This has prompted me to review the language in the description above, BTW, but I will do that in a separate PR. Question: Does `dvc checkout` remove data files no longer referenced in workspace DVC-files? Thanks

jorgeorpinel: I applied my suggestion in 0d7e4f8d. Sorry @Suor, hijacking your branch to apply some other changes (including https://github.com/iterative/dvc.org/pull/864#issuecomment-567345548) and move things along 😬

jorgeorpinel: p.s. please review the new text and mark this resolved if you agree.

jorgeorpinel:
> it would make more sense to mention this option somewhere in and link from https://dvc.org/doc/user-guide/large-dataset-optimization to here.

Done in 73bbb3af.

jorgeorpinel:
and 8e1fb170

jorgeorpinel:
Guys, I took over this PR, sorry. I just had to basically rewrite much of the description to make things clearer. Thanks @Suor for explaining your points in https://github.com/iterative/dvc.org/pull/864#pullrequestreview-334419515, marked resolved now. But please review the changes here and lmk or feel free to make further changes for us to make a final review. Thanks!

static/docs/command-reference/checkout.md
jorgeorpinel: Resolving in favor of https://github.com/iterative/dvc.org/pull/864#issuecomment-569373439

Suor:
> recreates (overwrites) links or copies from cache to workspace
  for all <abbr>outputs</abbr> in the workspace, including the ones normally
  checked out by this command, **as well as existing ones** with matching
  checksums in current DVC-files.

"including the ones normally
  checked out by this command" makes this really hard to get, "this command does what it normally does, including something ...". And you are making bold "existing", which is not the point, `dvc checkout` without relink also overwrites existing files.

jorgeorpinel:
@Suor I'm signing off for the day, can you push a suggested change meanwhile? Thanks

jorgeorpinel:
OK @Suor I rephrased it again. Please review guys.

static/docs/command-reference/checkout.md
```diff
@@ -21,34 +21,31 @@ which instance of each data file or directory should be used, with the checksums
 saved in the `outs` field. The `dvc checkout` command updates the workspace data
 to match with the <abbr>cached</abbr> files that correspond to those checksums.
 
-Using an SCM like Git, the DVC-files are kept under version control. At a given
-branch or tag of the SCM repository, the DVC-files will contain checksums for
-the corresponding data files kept in the cache. After an SCM command like
-`git checkout` is run, the DVC-files will change to the state at the specified
-branch or commit or tag. Afterwards, the `dvc checkout` command is required in
-order to synchronize the data files with the currently checked out DVC-files.
+When using Git, the DVC-files can be kept under version control. DVC-files in
```
shcheklein: can be -> are

static/docs/command-reference/checkout.md
```diff
@@ -21,34 +21,31 @@ which instance of each data file or directory should be used, with the checksums
...
+different branches or tags may contain checksums for different data files, saved
```
shcheklein: TBH I prefer the previous version ... why do we need to change it?

static/docs/command-reference/checkout.md
```diff
@@ -21,34 +21,31 @@ which instance of each data file or directory should be used, with the checksums
...
+by DVC in the project's cache. The DVC-files in the workspace, or the value of
+their `outs` fields can change when using `git checkout`. The `dvc checkout`
+command is required in this situation, in order to synchronize the data files
+(tracked by DVC) with the checked out DVC-files.
 
-This command must be executed after `git checkout` since Git doesn't track files
```
shcheklein: Again, not sure why are we removing this ...

static/docs/command-reference/checkout.md
```diff
@@ -21,34 +21,31 @@ which instance of each data file or directory should be used, with the checksums
...
+`git checkout`. To install it, use `dvc install`.
 
 The execution of `dvc checkout` does:
 
-- Scan the `outs` entries in DVC-files to compare with the currently checked out
-  data files. The scanned DVC-files is limited by the listed `targets` (if any)
-  on the command line. And if the `--with-deps` option is specified, it scans
-  backward from the given `targets` in the corresponding
-  [pipeline](/doc/command-reference/pipeline).
+- Scan the `outs` entries in DVC-files to compare with the <abbr>outputs</abbr>
```
shcheklein: And again, not sure why are we changing this.

static/docs/command-reference/checkout.md
```diff
@@ -21,34 +21,31 @@ which instance of each data file or directory should be used, with the checksums
...
+  currently in the <abbr>workspace</abbr>. Scanning is limited to the given
+  `targets` (if any).
 
-- For any data files where the checksum doesn't match their DVC-file entry, the
-  data file is restored from the cache. The link strategy used (`reflink`,
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
+- Missing data files or directories, or those with checksums that don't match
+  any DVC-file, are restored from the cache. If the `--relink` option is used,
+  all <abbr>outputs</abbr> in the workspace are recreated (overwritten). The
```
shcheklein: why can't we simplify and use a simple term? or rephrase it somehow ... it's too complicated 

The essence of it - "if --relink is specified we check and update link types (as specified in ...) for all outputs".

static/docs/command-reference/checkout.md
shcheklein: does it fit into a single line?

static/docs/command-reference/config.md
```diff
@@ -139,6 +139,10 @@ for more details.) This section contains the following options:
   [File link types](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
   for a full explanation of each one.
 
+  To re-create the file links in the workspace after this option, please use
```
shcheklein: after changing this option?

static/docs/command-reference/checkout.md
```diff
@@ -94,6 +88,12 @@ be pulled from remote storage using `dvc pull`.
   remove files that don't match those DVC-file references or are missing from
   cache. (They are not "committed", in DVC terms.)
 
+- `--relink` - recreates (overwrites) file links or copies, from cache to
```
shcheklein: `recreates (overwrites) file links or copies` - too complicated. If we'd like to be 100% precise (I'm not sure it's needed to such extent), let's write two sentences.

static/docs/command-reference/checkout.md
```diff
@@ -94,6 +88,12 @@ be pulled from remote storage using `dvc pull`.
...
+  workspace, of **all <abbr>outputs</abbr>** referenced in current DVC-files
+  (regardless of whether the checksums match a DVC-file). This ensures the link
```
shcheklein: I would even prefer to just for the second sentence go first. Then some explanation if needed.

static/docs/command-reference/checkout.md
jorgeorpinel: Hmm, you don't HAVE to version them. But I rephrased this to avoid the can vs. is question.

shcheklein: In 99.999% cases they are versioned, so no reason to complicate language here, I believe.

static/docs/command-reference/checkout.md
jorgeorpinel: Yep

image
> https://dvc-landing-relink-ubdbmixgnby.herokuapp.com/doc/command-reference/checkout

static/docs/command-reference/config.md
jorgeorpinel: Yes, thanks. Improved this sentence.

static/docs/command-reference/checkout.md
jorgeorpinel: That's high confidence! 😋 yes, I've already changed this sentence completely. It now starts:

> When using Git, the different DVC-files versioned in separate branches or tags...

Will continue conversation about this paragraph in https://github.com/iterative/dvc.org/pull/864#pullrequestreview-337893478.

static/docs/command-reference/checkout.md
jorgeorpinel: It wasn't great. No need for SCM term, strange phrase "kept under version control" (vs. just "versioned"), "will contain" isn't necessarily true (only a few DVC-files may change among revisions), "command like `git checkout`" instead of just saying "after `git checkout`", confusing phrase "change to the state at the specified branch", "Afterwards" also sounds weird in that place... etc. Most sentences had a small problem so I just rewrote it completely. Maybe I just didn't do a great job. Let me work it some more, trying to recuperate parts of the version you prefer... I'll be back 🕶

static/docs/command-reference/checkout.md
```diff
@@ -17,59 +17,54 @@ positional arguments:
...
+[revisions](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+probably specify different data files from the cache. When switching to those
+versions (with Git commands such as `git checkout`), the current DVC-files will
+no longer match with all of the data in the <abbr>workspace</abbr>.
+
+The `dvc checkout` command synchronizes the workspace data to match with the
+current DVC-files, using a mechanism described below.
+
+💡 For convenience, a Git hook is available to automate running `dvc checkout`
+after `git checkout`. Use `dvc install` to install it.
```
jorgeorpinel: Here's my newest version of this first half of the cmd ref description. I think it's much clearer, what do you think?

static/docs/command-reference/checkout.md
jorgeorpinel: Ready, can we continue this cat over in https://github.com/iterative/dvc.org/pull/864#pullrequestreview-337940215 below?

static/docs/command-reference/checkout.md
jorgeorpinel: It's redundant, we've already said what situations require `dvc checkout` and why in the previous paragraph (in the new version of the p, that is). Please see https://github.com/iterative/dvc.org/pull/864#pullrequestreview-337940215 now.

static/docs/command-reference/checkout.md
jorgeorpinel: The only changes in this bullet were to introduce the "outputs" tooltip, and remove the note about `--with-deps` which I think is unnecessary (since this rare option is explained later in the **Options** section). I do realize now maybe "outputs" should be introduced earlier in the description so it makes more sense to use the term here. Just did that in 56f0780a.

static/docs/command-reference/checkout.md
jorgeorpinel: I'm not sure which term you're referring to that we could simplify. My sentence is:

> If the `--relink` option is used, all outputs in the workspace are recreated (overwritten).

Yours is:

> if `--relink` is specified, we check and update link types (as specified in ...) for all outputs.

I'm not seeing a big difference. Also what is the "as specified in ..." part? A link to the Options section?

Sorry, maybe I just didn't get what you mean here.

static/docs/command-reference/checkout.md
jorgeorpinel: OK. I rewrote it in 83757462 to be both simpler and fully correct (I think):

```
- `--relink` - recreates **all <abbr>outputs</abbr>** referenced in current
  DVC-files (regardless of whether the checksums match a DVC-file). This means
  overwriting the file links or copies from cache to workspace. This ensures the
  link types of all the data files in the workspace are consistent with the
  project's [`cache.type`](/doc/command-reference/config#cache).
```

jorgeorpinel: > I would even prefer to just for the second sentence go first. Then some explanation if needed.

Good idea, just changed to:

```
- `--relink` - ensures the link types of all the data files in the workspace are
  consistent with the project's
  [`cache.type`](/doc/command-reference/config#cache). This is achieved by
  recreating **all <abbr>outputs</abbr>** referenced in current DVC-files
  (regardless of whether the checksums match a DVC-file). This means overwriting
  the file links or copies from cache to workspace.
```

static/docs/command-reference/checkout.md
jorgeorpinel: Moved to https://github.com/iterative/dvc.org/pull/864#discussion_r362704837.

static/docs/command-reference/checkout.md
```diff
@@ -17,59 +17,54 @@ positional arguments:
...
+call these files <abbr>outputs</abbr>, and their checksums are saved in the
+`outs` fields inside DVC-files to achieve this.
+
+When using Git, different DVC-files versioned in separate
+[revisions](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+probably specify different data files from the cache. When switching to those
+versions (with Git commands such as `git checkout`), the current DVC-files will
+no longer match with all of the data in the <abbr>workspace</abbr>.
+
+The `dvc checkout` command synchronizes the workspace data to match with the
```
shcheklein: introduction is way too long before we actually say anything about `dvc checkout` and what it does. Also, it's too tight with `git checkout` this way. It's usual case for `dvc checkout` but not a single one (e.g. `git clone` + `dvc checkout` is a legit one). The point is - the first paragraph should describe the essence of the operation - read DVC-files + restore the workspace accordingly. Then we can start talking about details - when it's useful, why, the logic behind it (mention different modes, etc).

shcheklein: I would also try to use "files and directories" vs outputs terminology as much as possible.

static/docs/command-reference/checkout.md
```diff
@@ -17,59 +17,54 @@ positional arguments:
...
+after `git checkout`. Use `dvc install` to install it.
+
+The execution of `dvc checkout` does the following:
+
+- Scans the `outs` field values in DVC-files to compare with the outputs
+  currently in the <abbr>workspace</abbr>. Scanning is limited to the given
+  `targets` (if any).
+
+- Missing data files or directories, or those with checksums that don't match
+  any DVC-file, are restored from the cache. If the `--relink` option is used,
```
shcheklein: if we decided to remove `--with-deps` why do we mention `--relink`? Especially here in the execution procedure. It makes sense to mention it if it's important in a paragraph above? or as a separate paragraph?

jorgeorpinel: You're right. I'm removing this sentence about `--relink` here, just moving part of it to the **Options** section. I'll mention all the options in these bullets just briefly instead, so the user knows to scroll down to options to learn about them. See 7deb5e29

static/docs/command-reference/checkout.md
jorgeorpinel: OK I simplified the description intro into 2 paragraphs which start by mentioning what `dvc checkout` does.

Not sure why we want to avoid the "output" term. I see it as a good thing when we get to use our terminology in general (without forcing it). But OK, I can rewrite it.

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,53 @@ positional arguments:
...
+that point.
+
+💡 For convenience, a Git hook is available to automate running `dvc checkout`
+after `git checkout`. Use `dvc install` to install it.
+
+The execution of `dvc checkout` does the following:
+
+- Scans the `outs` field values in DVC-files to compare with the outputs
+  currently in the <abbr>workspace</abbr>. Scanning is limited to the given
+  `targets` (if any). See also the `--recursive` option below.
```
shcheklein: why only recursive? there are other options that change the scope

shcheklein: I think --with-deps chnages the scope?

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,53 @@ positional arguments:
...
+such as `git checkout`), the current DVC-files will no longer match with all of
+the data in the <abbr>workspace</abbr>, and so `dvc checkout` will be needed at
+that point.
+
+💡 For convenience, a Git hook is available to automate running `dvc checkout`
+after `git checkout`. Use `dvc install` to install it.
+
+The execution of `dvc checkout` does the following:
+
+- Scans the `outs` field values in DVC-files to compare with the outputs
```
shcheklein: I really don't like the usage of outputs. I don't see any need for this. We deal with files and directories for the user perspective. It's fine to point once that they can come with a different name (aka outputs because they are "outputs" in the DVC-files), but that should be enough. Unless we have a really strong reason let's prefer using human-readable terminology vs DVC internal one. Especially in cases like this. `Output` is a very very controversial name for the data file or directory that was `dvc add`-ed for example.

static/docs/command-reference/checkout.md
shcheklein: Still a bit too complicated :)

"DVC-files can be considered as a pointer or a placeholder for the actual data file or a directory under DVC control. This command reads DVC-files and updates DVC-controlled data files and directories in the workspace to be in sync with the versions specified in the DVC-files." - something along those lines?

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,53 @@ positional arguments:
...
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+The `dvc checkout` command synchronizes the workspace data to match with the
+current [DVC-files](/doc/user-guide/dvc-file-format) in the
+<abbr>project</abbr>. DVC knows which data files (a.k.a. <abbr>outputs</abbr>)
+to use because their checksums are saved in the `outs` fields inside the
+DVC-files.
+
+This is useful when the project is a <abbr>DVC repository</abbr>, since
```
shcheklein: "You usually use it after you cloned (`git clone`) the project or switched to a different branch/commit in your git history with `git checkout`." - something like this?

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,53 @@ positional arguments:
...
+the data in the <abbr>workspace</abbr>, and so `dvc checkout` will be needed at
+that point.
+
+💡 For convenience, a Git hook is available to automate running `dvc checkout`
+after `git checkout`. Use `dvc install` to install it.
+
+The execution of `dvc checkout` does the following:
+
+- Scans the `outs` field values in DVC-files to compare with the outputs
+  currently in the <abbr>workspace</abbr>. Scanning is limited to the given
```
shcheklein: it's not clear this way if targets limit outs or files? Let's be explicit and keep it simple.

static/docs/command-reference/checkout.md
shcheklein: I think it it also make sense to mention that files to checkout come from cache (not remote, the difference between checkout and pull, for example). It's also important since a few paragraphs below we talk about cache and it's not clear why.

static/docs/command-reference/checkout.md
```diff
@@ -94,6 +81,19 @@ be pulled from remote storage using `dvc pull`.
...
+- `-d`, `--with-deps` - determine files to update by tracking dependencies to
+  the target DVC-files (stages). This option only has effect when one or more
+  `targets` are specified. By traversing all stage dependencies, DVC searches
+  backward from the target stages in the corresponding pipelines. This means DVC
+  will not checkout files referenced in later stages than the `targets`.
+
+- `--relink` - ensures the file linking strategy (`reflink`, `hardlink`,
+  `symlink`, or `copy`) for all data files in the workspace is consistent with
+  the project's [`cache.type`](/doc/command-reference/config#cache). This is
+  achieved by recreating **all <abbr>outputs</abbr>** referenced in current
```
shcheklein: recreating - sounds like we run `repro`

outputs - again, not sure we need it here

static/docs/command-reference/checkout.md
jorgeorpinel: OK, I'll remove the term. Do you also mean removing mentions to the "`outs` field" ?

static/docs/command-reference/checkout.md
jorgeorpinel: I simplified it again along the lines of your suggestion, thanks. There's still a short 2nd paragraph emphasizing that this is useful in the context of Git repos.

> it also make sense to mention that files to checkout come from cache

It is already mentioned in the 2nd bullet about the execution of the command, before the other mentions of the cache you saw. I did wrap it with `<abbr>` now though, so it's more obvious and has a tooltip.

static/docs/command-reference/checkout.md
jorgeorpinel: Right, I didn't assign the options to the bullets correctly. Moved `--with-deps` to the first bullet now.

static/docs/command-reference/checkout.md
jorgeorpinel: > It's fine to point once that they can come with a different name

OK, left one mention only now.

> Unless we have a really strong reason let's prefer using human-readable terminology vs DVC internal one.

OK. I thought command references are our most technical documents, but will keep this in mind.

> Especially in cases like this. Output is a very very controversial name

Good point.

jorgeorpinel: Left mentions to `outs` field though, please lmk if you meant to remove that too. Thanks

static/docs/command-reference/checkout.md
jorgeorpinel: OK I see you did mean to remove that mention of `outs` field, at least. OK, simplified now.

jorgeorpinel: I guess this implies no need to mention checksums in the 2nd bullet either. Simplified that now too.

Wow, this brings us full circle back to my very first review that started this whole takeover! -> https://github.com/iterative/dvc.org/pull/864#discussion_r360175115

jorgeorpinel: p.s. updated --relink option desc. accordingly.

static/docs/command-reference/checkout.md
jorgeorpinel: Replaced "recreating" for "restoring" in 1813f942. "outputs" I removed previously from here.

static/docs/command-reference/checkout.md
jorgeorpinel: Yep, you read my mind. I made this change before noticing your comment. Great minds think alike 💀 

> ...after `git clone`, `git checkout`, or any other repository operations that change the
currently present DVC-files.

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-- For any data files where the checksum doesn't match their DVC-file entry, the
-  data file is restored from the cache. The link strategy used (`reflink`,
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
-
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are essentially placeholders that
+point to the actual data files or a directories under DVC control. This command
```
shcheklein: `or a directories` -> `or directories`

static/docs/command-reference/checkout.md
```diff
@@ -94,6 +83,13 @@ be pulled from remote storage using `dvc pull`.
   remove files that don't match those DVC-file references or are missing from
   cache. (They are not "committed", in DVC terms.)
 
+- `--relink` - ensures the file linking strategy (`reflink`, `hardlink`,
+  `symlink`, or `copy`) for all data files in the workspace is consistent with
+  the project's [`cache.type`](/doc/command-reference/config#cache). This is
+  achieved by restoring **all data files or a directories** referenced in
```
shcheklein: probably restoring happens only if link type does not match, right?

static/docs/command-reference/checkout.md
```diff
@@ -94,6 +83,13 @@ be pulled from remote storage using `dvc pull`.
...
+  current DVC-files (regardless of whether they match a current DVC-file). This
+  means overwriting the file links or copies from cache to workspace.
```
shcheklein: sentence is cryptic and complicated ... `overwriting ... copies from cache to workspace` - very hard to understand what does it mean

static/docs/user-guide/large-dataset-optimization.md
```diff
@@ -120,6 +120,13 @@ file link types. Please refer to the
 [Update a Tracked File](/doc/user-guide/updating-tracked-files) on how to manage
 tracked files under these cache configurations.
 
+### Re-linking data in the workspace
+
+To re-create the file links in the workspace, for example after changing the
```
shcheklein: re-create ----> make sure that link are of the right type and recreate if needed?

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are essentially placeholders that
+point to the actual data files or a directories under DVC control. This command
+synchronizes the workspace data with the versions specified in the current
+DVC-files. DVC knows which data files (<abbr>outputs</abbr>) to use because
+their checksums are saved in the `outs` fields inside the DVC-files.
+
+`dvc checkout` is useful when using Git in the <abbr>project</abbr>, after
```
shcheklein: is, for example, useful

static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-  data file is restored from the cache. The link strategy used (`reflink`,
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
-
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are essentially placeholders that
+point to the actual data files or a directories under DVC control. This command
+synchronizes the workspace data with the versions specified in the current
```
shcheklein: we use different terms everywhere - `data`, `data files` `data files and directories` - can we use one (or two at least)?

static/docs/command-reference/checkout.md
jorgeorpinel: Good question. @Suor can you confirm please?

static/docs/command-reference/checkout.md
jorgeorpinel: How about just "Note that this overwrites the data files or directories in the workspace." I've changed it to that now but may have to reword all of this depending on Alex reply in https://github.com/iterative/dvc.org/pull/864#pullrequestreview-338379551 above.

static/docs/user-guide/large-dataset-optimization.md
jorgeorpinel: I don't know if the "if needed" part is accurate. Waiting for Alex' reply in https://github.com/iterative/dvc.org/pull/864#pullrequestreview-338379551 above. But I simplified this sentence along the lines you suggested. Thanks.

static/docs/command-reference/checkout.md
jorgeorpinel: OK I'm using mostly "data files and directories" now, and "data" some times to avoid repetitive wording (then the full phrase is already mentioned in the same paragraph).

static/docs/command-reference/checkout.md
jorgeorpinel: Done too 👍  Thanks

jorgeorpinel:
Just pending to figure out https://github.com/iterative/dvc.org/pull/864#pullrequestreview-338315859 ^

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
+`dvc checkout` is useful, for example, when using Git in the
+<abbr>project</abbr>, after `git clone`, `git checkout`, or any other repository
+operation that changes the currently present DVC-files.
+
+💡 For convenience, a Git hook is available to automate running `dvc checkout`
+after `git checkout`. Use `dvc install` to install it.
+
+The execution of `dvc checkout` does the following:
+
+- Scans the DVC-files to compare vs. the data files or directories currently in
```
shcheklein: not sure I understand this sentence .. I mean I understand but it looks like it's not a correct sentence from the language perspective?

public/static/docs/command-reference/checkout.md
```diff
@@ -94,6 +83,13 @@ be pulled from remote storage using `dvc pull`.
   remove files that don't match those DVC-file references or are missing from
   cache. (They are not "committed", in DVC terms.)
 
+- `--relink` - ensures the file linking strategy (`reflink`, `hardlink`,
+  `symlink`, or `copy`) for all data in the workspace is consistent with the
+  project's [`cache.type`](/doc/command-reference/config#cache). This is
+  achieved by restoring **all data files or a directories** referenced in
+  current DVC-files (regardless of whether they match a current DVC-file). Note
+  that this overwrites the data in the workspace.
```
shcheklein: overwrites probably not a good term when we deal with links .. it sounds too scary

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
-
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are essentially placeholders that
+point to the actual data files or directories under DVC control. This command
+synchronizes the workspace data with the versions specified in the current
+DVC-files. DVC knows which data (<abbr>outputs</abbr>) to use because their
```
shcheklein: again, don't see any reason to introduce the term outputs whatsoever here. The only reasonable place is below - when we talk about implementation details. `To be precise, DVC knows which file to put into workspace by reading the checksums saved in the `outs` field in the DVC-files`. Not sure that we need this sentence at all here. Move it to the implementation details?

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
+checksums are saved in the `outs` fields inside the DVC-files.
+
+`dvc checkout` is useful, for example, when using Git in the
+<abbr>project</abbr>, after `git clone`, `git checkout`, or any other repository
```
shcheklein: repository operation -> operation

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
+operation that changes the currently present DVC-files.
```
shcheklein: `the currently present DVC-files` -> `the DVC-files` or `the DVC-files in the workspace`

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-
-- For any data files where the checksum doesn't match their DVC-file entry, the
-  data file is restored from the cache. The link strategy used (`reflink`,
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
-
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are essentially placeholders that
```
shcheklein: can you actually use "placeholders that point ..." ... or should it be just "placeholders for smth."?

shcheklein: I would also remove `essentially`

public/static/docs/command-reference/checkout.md
jorgeorpinel: Do you mean "Scans the DVC-files to compare vs. the data files or directories currently in the workspace" ? I would maybe change "vs."->"against" and remove "currently" but either way I think it's a proper sentence. Doubled checked with grammarcheck.me, gingersoftware.com/grammarcheck, and grammarchecker.online/#grammar-check 🙂

public/static/docs/command-reference/checkout.md
shcheklein: also, it better to say something like "specific version of data ..."

jorgeorpinel: I think both are correct. They are placeholders for something, and they point to that something. Do you think it's redundant? I'm just trying to be extra clear about how DVC-files work.

public/static/docs/command-reference/checkout.md
jorgeorpinel: I did mean it as a "scary" warning, but I guess it's a pretty safe operation so perhaps this whole sentence is not needed. Removed!

public/static/docs/command-reference/checkout.md
jorgeorpinel: This is the only instance left of the term "output" (other than a pre-existing one by the end of the description which has a different context). I thought we agreed it was OK to at least mention it once so the terminology is used. See https://github.com/iterative/dvc.org/pull/864#discussion_r362992294

I don't see why it's only reasonable to use this term in deeper implementation details. It's just a quick parenthesis to establish the term. But OK, I moved the sentence below, into the 2nd bullet of the implementation details. You're right that it makes more sense to have that sentence there.

> `To be precise, DVC knows which file to put into workspace by reading the checksums saved in the outs field in the DVC-files.`

Is this version of the sentence your suggestion? Or is it coming from an older version? The current version is "DVC knows which data (<abbr>outputs</abbr>) to use because their checksums are saved in the `outs` fields inside the DVC-files."

static/docs/command-reference/checkout.md
jorgeorpinel: I think this is resolved.

static/docs/command-reference/checkout.md
jorgeorpinel: I think this is also resolved (or outdated).

public/static/docs/command-reference/checkout.md
jorgeorpinel: p.s. removed "essentially" and changed "the actual data"->"specific version of data"

static/docs/command-reference/checkout.md
jorgeorpinel: (Seems like this PR has gone off Alex' radar so I just pinged him on Slack.)

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
-
-- For any data files where the checksum doesn't match their DVC-file entry, the
-  data file is restored from the cache. The link strategy used (`reflink`,
-  `hardlink`, `symlink`, or `copy`) depends on the OS and the configured value
-  for `cache.type` – See `dvc config cache`.
-
-Note that this command by default tries NOT to copy files between the cache and
-the workspace, using reflinks instead when supported by the file system. (Refer
-to
+[DVC-files](/doc/user-guide/dvc-file-format) are placeholders that point to
```
shcheklein: sorry, still feels strange to me :)

it's not right to mix placeholder and pointer, probably

`DVC-file can be considered as pointers to ... under DVC control (namely,. `outs` field values in those files)`?

public/static/docs/command-reference/checkout.md
```diff
@@ -16,60 +16,49 @@ positional arguments:
...
+
+The execution of `dvc checkout` does the following:
+
+- Scans the DVC-files to compare against the data files or directories in the
+  <abbr>workspace</abbr>. Scanning is limited to the given `targets` (if any).
+  See also options `--with-deps` and `--recursive` below.
+
+- Missing data files or directories, or those that don't match with any
+  DVC-file, are restored from the <abbr>cache</abbr>. DVC knows which data
+  (<abbr>outputs</abbr>) to use because their checksums are saved in the `outs`
```
shcheklein: it makes sense to mention about `outs` here, but it should be before this point (restoring files from cache) - when we scan to compare we already read those

public/static/docs/command-reference/checkout.md
jorgeorpinel: OK I assume this one is resolved, lmk otherwise. Thanks 🙂

public/static/docs/command-reference/checkout.md
jorgeorpinel: > sorry, still feels strange to me :)
> it's not right to mix placeholder and pointer, probably

OK, changed per your suggestion to:

`DVC-files act as pointers to specific version of data files or directories under DVC control.`

But omitted the part about `outs` field because it's explain further down in the implementation details, per https://github.com/iterative/dvc.org/pull/864#pullrequestreview-338470052.

Good?

public/static/docs/command-reference/checkout.md
jorgeorpinel: Answered in https://github.com/iterative/dvc.org/pull/864#discussion_r363517180

public/static/docs/command-reference/checkout.md
shcheklein: 👍

public/static/docs/command-reference/checkout.md
jorgeorpinel: You're right. Again. Thanks! Changed 🙂

static/docs/command-reference/checkout.md
Suor: if `--relink` is specified then it always relinks, i.e. removes and links. There is no way in general case to check whether link type is correct. For example, we can't distinguish reflink from a copy. And even checking for hardlink is not completely reliable.

jorgeorpinel: Good to know! Leaving unresolved so it's easy to find for future reference.

PR #872, opened 2019-12-17
Fix #697: Make unexisting docs pages return 404 status

### LINKED ISSUE #697: engine: return HTTP 404 when Page404 component is rendered
> Extracted from https://github.com/iterative/dvc.org/pull/690#issuecomment-541958989

Example: the engine renders a huge number "404", however it already responded with HTTP 200:

image
> URL: https://dvc.org/doc/command-reference/bad

It would be ideal if the engine could check for the URL path validity first before responding by rendering the general layout (navigation, etc) with HTTP 200, and render it instead with HTTP 404 (and `Page404 `).

> This can aid in checking links by running script in #690. Once `404` is returned, we will be able to distinguish between existing pages and `Not found` ones.

iAdramelk:
Fix #697. Add check for /doc pages in `server.js`, if page does not exists, it return correct status.


utils/helper.js
```diff
@@ -1,6 +1,7 @@
-import startCase from 'lodash.startcase'
+/* eslint-env node */
```
jorgeorpinel: Why is this needed?

utils/helper.js
```diff
@@ -1,6 +1,7 @@
...
 
-import sidebar from '../sidebar'
+const startCase = require('lodash.startcase')
+const sidebar = require('./sidebar')
```
jorgeorpinel: Agree with Ivan, this file can be left in src/Documentation/ I think.

utils/helper.js
```diff
@@ -172,4 +173,8 @@ export function getParentsListFromPath(path) {
     })
 }
 
-export default normalizedSidebar
+module.exports = {
+  sidebar: normalizedSidebar,
+  getItemByPath,
+  getParentsListFromPath
+}
```
jorgeorpinel: Is this what you meant by changing file to CommonJS? Just checking 🙂

jorgeorpinel:
I tested the deployment and it works, thanks @iAdramelk! Seems like we just have minor concerns here 🙂  ^

utils/helper.js
iAdramelk: We added two global variables for CommonJS: `require` and `exports`, linter will not understand them if we didn't explicitly say that this is node code.

utils/helper.js
iAdramelk: Not sure if this is a good place for that either. It is basically docs stricture and is edited together withe docs, it is not a part of visual component.

utils/helper.js
iAdramelk: Yep. `import` changed to `require` and `export` to `module.exports`. Also replaced `default` export with the named one.

utils/helper.js
shcheklein: yep, I would say even helper should be part of the `src/Documentation`, not utils. To me `utils` is a place for library-like code that is used in a lot of places across the code - like array extensions, or string helpers, etc

iAdramelk:
@shcheklein 

> `utils/sidebar.json` is a strange location for the file? can we move it anywhere at all?

I agree that this is a strange place, but I think that storing it inside visual component is also bad. Where should we store such files?

shcheklein:
@iAdramelk let's create `src\sidebar`?

utils/helper.js
iAdramelk: But we already use helper and sidebar data in multiple places:

- `server.js` to check if page exists,
- `pages/doc.js` to load content,
- `src/Documentation/Markdown/utils/remark-linker.js` to check if page exists,
- `src/Documentation/SidebarMenu/SidebarMenu.js` to draw sidebar. 

It's basically just docs structure and routing info, not just info for sidebar. So I think that utils is a good place for helper (but we probably should rename it). But for `sidebar.json` itself I think that it should be placed somewhere near the docs, because they are often edited together and without changing js code. What to you think?

shcheklein: why don't you like idea of moving it top-level - to a separate directory like `sidebar`? both `helper` and `json`. It's just my perception of the `utils` is different. To me it's like `commons` - I would expect for stuff like I mentioned before - collections, strings, etc, not application-specific logic. I can imagine it will be confusing to others, so why don't we make it super explicit - `src\sidebar\sidebar.json` and `src\sidebar\helper.js` . It will make it super easy and clear to navigate, explain, document, etc.

iAdramelk: @shcheklein different reasons for `helper.js` and `sidebar.json` actually.

`sidebar.json` – I think that `src` folder is for application's source code and that data and content shouldn't be there. And I think that `sidebar.json` is a part of the content as it is mostly used as part of content editing tasks, and not programming tasks. So fixing something that is a part of the `src` folder to add a page to the docs didn't feel right to me.

`helper.js` – this file, on the other hand, should be placed inside the `src` folder, because it is a part of application source code, and I don't think that it should be a part of `SidebarMenu` component because it is used in a lot of other places. We can create a folder in `src` just for this file, but it looks wasteful for me to create a top-level folder just for it. (Also we already have a lot of folders there and no distinction between component folders and other folders, so for me, it is hard to navigate between them and look for specific folder right now)

P. S. Offtopic, but I think that we should move all react components to subfolder `components` to clean `src` a little.

shcheklein: So, let's move `sidebar.json` to the very top at least, `utils` just don't make any sense for it. And `helper.js` - I'm fine with `utils` since it's dev facing anyways.

shcheklein: we should update a few docs where sidebar.json is mentioned if we move it

iAdramelk: @shcheklein agreed. To the top of `src` folder or to the site's root folder?

shcheklein: probably site's root? like you mentioned it's not the source code

utils/helper.js
jorgeorpinel: How come it's not needed in all the other Node files?

utils/helper.js
shcheklein: It can be part of `static/docs` indeed, close to the content that it defines. In this sense it's as _static_ as those Markdown files we have, so should be good.

iAdramelk:
Updated names and locations:

`utils/helper.js` -> `src/utils/sidebar.js`
`utils/sidebar.json` -> `static/docs/sidebar.json`

Also updated path to `sidebar.json` in docs.

utils/helper.js
shcheklein: @iAdramelk could you please give some context? should we add it there?

iAdramelk: @jorgeorpinel But server.js has it, I copied it from there. :)

iAdramelk: Basically in `.eslitrc.json` we wrote which set of global variables to use`{ env: browser: true } }` then linting files and then we wrote `lint-staged` rule in `package.json` that says which files to lint on commit: `"lint-staged": { "*.{js,jsx}": "eslint" }`. 

This way, all `*.{js,jsx}` files that we try to commit will be linted. Files that are written using `import`/`export` will be linted automatically, but for the node files, we need to add this rule.

There is an inconsistency in our config right now: We lint all files on commit, but only lint `pages` and `src` folders on CI. It can lead to a situation when some old node files still didn't have this rule because we didn't change them in a long time. I think we should also lint all files on CI.

jorgeorpinel: Ah sorry Idk why I didn't find that in server.js before 😅 

> There is an inconsistency in our config right now: We lint all files on commit, but only lint pages and src folders on CI. I think we should also lint all files on CI.

We recently noticed some problems with the CI formatting but, only on the `master` branch. Not exactly what you're describing though. We link only staged files on commit:

```yarn
...
      "pre-commit": "yarn format-staged && lint-staged"
...
  "lint-staged": {
    "*.{js,jsx}": "eslint"
```

But anyway, yes I need to review the hook and CI setup. Thanks for the note!

PR #875, opened 2019-12-18
Add troubleshooting guide

### LINKED ISSUE #871: sync: prepare troubleshooting guide for `Too many open files` error
Related:
https://github.com/iterative/dvc/pull/2866
https://github.com/iterative/dvc/issues/2473

Need to prepare short information about why `Too many open files` happens and how to solve it, as this information is too long for exception message.

After this issue is solved, need to improve log message to include a link to the explanation

pared:

Fixes #871


shcheklein:
@jorgeorpinel @pared let's create a user guide Troubleshooting section. I don't like this duplication and complication of the command reference. We can create a separate page per error, or for now a single page with expandable blocks or h2 segments.

shcheklein:
@pared let's wait on @jorgeorpinel before doing this change above ^^

jorgeorpinel:
@shcheklein my initial https://github.com/iterative/dvc.org/issues/871#issuecomment-566755326 was

> If it will fit in a single paragraph I recommend using existing docs, probably some of the command references. It can be repeated in a few ones wherever it's appropriate.

We have several of these repeated notes, paragraphs, sentences in different docs already, especially command references. It makes sense since there's overlap in functionality or implementation details in many places.

I don't think we need a **Common problems** section though, since there's only one and can't foresee having others in the immediate future. Maybe just add the paragraph in the bottom of the **Description** section? Starting with "A known issue..."

shcheklein:
> If it will fit in a single paragraph

It looks like it does not, right? at least it'a big paragraph already :) And it would be an artificial constraint to try to keep error explanations like this very short.

also, this is an edge case, it's not that important to include it into description for all commands (3 places at least?)

> We have several of these repeated notes, paragraphs, sentences in different docs already, especially command references.

they are usually small, there are not so many of them, and it might be a problem for some of them that they are repeated. And to my point above - they are usually general enough and highlight some important workflow (like installing S3 with DVC), they do not serve as an explanation of an error code.

> I don't think we need a Common problems section though

I like Troubleshooting as the name, someone has used it already in some comment - it makes sense to me :)

> can't foresee having others in the immediate future

I can easily add a few :)One of the most commons - what should I do when I get "DVC is busy" message. There will be more and more as we move forward and improve interface.

static/docs/command-reference/pull.md
```diff
@@ -100,6 +100,19 @@ reflinks or hardlinks to put it in the workspace without copying. See
...
+- `[Errno 24] Too many open files` - most common for MacOs users who work with
+  s3 remote. Each download thread needs to open one or more file descriptor(s).
+  The more `jobs` is specified, the more file descriptors needs to be open on
+  host filesystem. If there is too many open file descriptors, host filesystem
+  will not allow opening new ones. Recommended approach to handle this error is
+  to increase open file descriptors limit. On UNIX systems that can be done
+  using `ulimit -n`. In case of Windows, please refer to
+  [official resources](https://blogs.technet.microsoft.com/markrussinovich/2009/09/29/pushing-the-limits-of-windows-handles/).
+  If changing limits is impossible, the workaround would be to use lower `jobs`
+  option value.
```
jorgeorpinel: Suggested (slightly shorter) paragraph:

A known problem some users run into with this command is `[Errno 24] Too many open files` (most common for S3 remotes on MacOS). The more `--jobs` specified, the more file descriptors need to be open on the host file system for each download thread, and the limit may be reached, causing this error. To solve this, it's often possible to increase the open file descriptors limit, for example with `ulimit -n` on UNIX-like system, or [increasing Handles limit](https://blogs.technet.microsoft.com/markrussinovich/2009/09/29/pushing-the-limits-of-windows-handles/) on Windows. Otherwise, please try using a lower `JOBS` value.

jorgeorpinel:
> let's create a user guide Troubleshooting section

To be clear I'm not opposed to this. It's a good option, and just link from these command refs. Also @pared please notice my other https://github.com/iterative/dvc.org/issues/871#issuecomment-567231227 in the issue:

> Maybe also `checkout` since it can include `pull`. Consider adding the note in the `remote` index page too.

> at least it'a big paragraph already

Indeed, it's a long one @shcheklein even my shorter version. Would be nicer to split it in 2.

> I like Troubleshooting as the name

Yes, I like it too. Or/and **Known issues**

static/docs/command-reference/pull.md
jorgeorpinel: Or it can be 2. The second one starting at "To solve this..."

shcheklein:
I would say: `Known issues` and `Troubleshooting` have different meaning to me. First is about actual existing bugs (and probably our GH issue tracker represents it really well, though it can be complimented with some docs if some _issues_ are not going to be fixed ... _ever_). Second is about solving _troubles_ - they are not necessarily _issues_ to me.

jorgeorpinel:
Agree. This file descriptor thing to me sounds like an issue that could have a code solution to me. Not sure though.

pared:
> Agree. This file descriptor thing to me sounds like an issue that could have a code solution to me. Not sure though.

@jorgeorpinel yes, it is solvable, and this change is taking some time because we were discussing it a lot with @efiop and @Suor. Some of the conclusions can be found at 
iterative/dvc#2866 and
iterative/dvc#2473

The conclusion was that we want to preserve quite high default value of `jobs`, and don't want to introduce complicated logic to adjust (only default) jobs value on a single remote (s3). In the future we will probably fix it by retrying after this fail, but that will probably need reorganizing error handling, so it will be bigger than `try: except OsError` 
iterative/dvc#2965

pared:
@jorgeorpinel @shcheklein I created `Troubleshooting` User guide section and included @jorgeorpinel suggestion.

static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
+# Troubleshooting
+
+In this section we cover some of the known issues that DVC user might stumbe
+upon.
+
+## Too many open files error
+
+A known problem some users run into with `pull`, `fetch` and `push` commands is
```
shcheklein: `dvc pull`, `dvc fetch`, `dvc push` - to make them links to the CR (command reference)

static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
+# Troubleshooting
+
+In this section we cover some of the known issues that DVC user might stumbe
```
shcheklein: `know issues` -> replace with something, they are not _issues_, they are expected behavior, right? So, it's a regular workflow you just hit bumps that require a bit more information to deal efficiently with.

jorgeorpinel:
> yes, it is solvable...

Thanks for all the context @pared! So, my point was that this may not be a "trouble" the way @shcheklein described it, but in fact a "known issue". I think you're confirming this?

static/docs/user-guide/troubleshooting.md
jorgeorpinel: I'm not sure. See https://github.com/iterative/dvc.org/pull/875#issuecomment-567396421 and https://github.com/iterative/dvc.org/pull/875#issuecomment-567710978.

shcheklein:
@jorgeorpinel @pared if agree on it to be a "known issue" then we don't need this doc - let's open ticket in the Github DVC repo with a description, workaround, ETA on solving, etc. I don't think we need document this things. To me the current solution is a first iteration, but a solution - thus I was thinking this documentation makes sense and should be part of the troubleshooting.

static/docs/user-guide/troubleshooting.md
jorgeorpinel: p.s. whatever we decide here should probably also apply to the term "known problem" in line 8 below.

static/docs/user-guide/troubleshooting.md
jorgeorpinel:
```suggestion
A known problem some users run into with the `dvc pull`, `dvc fetch` and `dvc push` commands is
```
(Needs rewrap.)

static/docs/user-guide/troubleshooting.md
shcheklein: @jorgeorpinel see my comment here - https://github.com/iterative/dvc.org/pull/875#issuecomment-567713481 . I don't think we need extra documentation besides GitHub issue tracker for "known issues" and "known problems" (in a sense that they are actual design bugs, bugs and/or are going to be fixed)

jorgeorpinel:
> if agree on it to be a "known issue" then we don't need this doc... I don't think we need extra documentation besides GitHub issue tracker for "known issues" and "known problems" (in a sense that they are actual design bugs, bugs and/or are going to be fixed)

To be or not to be! (A known issue) That is the question. I like the "regular workflow, you just hit bumps" argument but not sure it's accurate. I have to yield to you guys in the end to decide what this thing is. But I think Ivan is right that a fully described issue on GH would be enough for "known issues".

static/docs/user-guide/troubleshooting.md
jorgeorpinel: OK. If you decide this is not a "known issue" though, in fact I think Ivan is also correct in this review. The term "known issue" should definitely not be in the description of a Troubleshooting guide 🙂

shcheklein:
> regular workflow, you just hit bumps

k, to be more precise :) `regular workflow, advanced scenario`. It can be considered somewhat similar to that "slow linking" message we have.

static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
...
+## Too many open files error
+
+A known problem some users run into with `pull`, `fetch` and `push` commands is
+`[Errno 24] Too many open files` (most common for S3 remotes on MacOS). The more
+`--jobs` specified, the more file descriptors need to be open on the host file
+system for each download thread, and the limit may be reached, causing this
+error.
+
+To solve this, it's often possible to increase the open file descriptors limit,
+for example with `ulimit -n` on UNIX-like system, or
```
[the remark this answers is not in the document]
pared: Yeah, though I would not use `JOBS` here, as this call sets limit for open files, so it should be quite big, (eg for Ubuntu there is 1024 limit)

pared:
@jorgeorpinel, @shcheklein I guess its `Known issue` though I don't think we want to include solution in log, it's just too extensive. Let's just create an issue aiming to solve the problem once and for all?

shcheklein:
@pared let's then create an issue that describes this and the workaround and link that issue instead of docs. I'm closing this for now then, will reopen if we reconsider this.

efiop:
Not sure why we've closed this. The article talks about ulimit, which might bite the user in multiple ways (this is not only about fds, you might hit proc ulimit too). I think it should be documented in the dvc.org, we won't be able to resolve it in-dvc for every case.

jorgeorpinel:
I think we do want a Troubleshooting doc but this wasn't the best thing to initiate it (referring to `[Errno 24] Too many open files`). Feel free to reuse some of the changes in this PR to talk about the matters Ruslan mentioned, guys! Ivan also mentioned he can probably add some items to the doc once it exists.

pared:
@shcheklein I think we didn't understand each other - what I mean was to create an issue for automatically solving the problem if it occurs. I don't think that hanging, open issue describing the problem is a good idea. I saw somewhere your comment about creating `Errors` section in docs, maybe that should be the way? 

Also, before closing this, I would like to achieve some kind of agreement about what do we decide to do.

shcheklein:
sure, reopening it! Let's keep the discussion going. @efiop @pared before we do a change on the docs side we need a clear vision from the DVC project - is this considered a regular advanced workflow and/or is it a temporary solution (issue) that will be resolved. It's important since you can see otherwise we can't come with a right place to put it.

shcheklein:
@pared Errors section in docs is similar to the troubleshooting in mind - this is section that describes public facing errors that are not bugs, or issue but are part of the regular workflow.

efiop:
@shcheklein ulimit issues are universal, we won't fully get rid of them ever. E.g. our tests are adjusting proc and fd ulimits for that reason. I think the article proposed by @pared is useful and we should keep it in the docs. "Troubleshooting" is a useful section to have in the docs.

shcheklein:
@efiop I'm not against it. (I think it was me who suggested the "troubleshooting" section in the first place.) My concern was that we were using confusing terminology for the section (e.g. "known issue"). Troubleshooting in my mind is about, let's say "advanced regular workflow" - this is something that will stay with us. We don't have a ticket for GH to fix this. Etc. If we are all on the same page with this, @pared please address some wording concerns we had, let's review it again and merge.

pared:
@shcheklein addressed the reviews, please check

static/docs/sidebar.json
```diff
@@ -159,6 +159,10 @@
       {
         "label": "Anonymized Usage Analytics",
         "slug": "analytics"
+      },
+      {
+        "label": "Troubleshooting guide",
+        "slug": "troubleshooting"
```
jorgeorpinel: Should probably place above analytics?

static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
+# Troubleshooting
+
+In this section we provide help for some of the problems that DVC user might
+stumbe upon.
```
jorgeorpinel: some of the problems -> some special situations ?

static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Too many open files error
+
+A known problem some users run into with the `dvc pull`, `dvc fetch` and
+`dvc push` commands is `[Errno 24] Too many open files` (most common for S3
```
jorgeorpinel: We're still using this `[Errno 24] Too many open files` known issue as the first "trouble". All for including this section but this is not a good example as discussed previously :/ Sorry to go in circles!

jorgeorpinel: Maybe just rephrase along the lines of Ruslan's https://github.com/iterative/dvc.org/pull/875#issuecomment-568251704. Or perhaps @shcheklein has an idea of a good first, actual trouble to put here.

shcheklein: @jorgeorpinel I think Ruslan's comment was specifically about this error, though. What is the GH ticket you are referring too?

My 2cs on this specific case. I agree with @efiop . It's impossible to solve `Too many open files` in general. This error in most cases means that some configuration is needed. It's very common for databases, for example, or other system level software. So, it can be part of the troubleshooting section.

I would say though, that I don't quite understand why do we use Errno 24 code - we should probably introduce our own codes for these situations or normal human readable messages with a link to this guide. But this is an UI/UX issue.

jorgeorpinel: > What is the GH ticket you are referring too?

https://github.com/iterative/dvc/issues/2473

I also agree with Ruslan but I though we had concluded that the `Too many open files` thing could possibly be fixed in the core code. If it has no solution, then this is OK. Up to you guys, feel free to mark this one resolved.

shcheklein: Yep, that ticket is closed. Resolving this. @efiop @pared please, create an UI ticket/checkbox to improve the DVC core message (at lease include a link to this guide, may be introduce error.dvc.org the same way we have man.dvc.org to make it look short and nice and have a better level of control).

static/docs/user-guide/troubleshooting.md
pared: I don't think we should avoid using strong wording, this is a problem from a point of the user view. I would say that a `special situation` does not exactly express the severeness of this. I do understand that might change if later we discover some less severe troubles to shoot, but we can then adjust. What do you think?

pared:
@shcheklein I would leave this `Errno 24`. When this error is thrown, `Too many open files` and this error code are quite specific messages that can be easily found on troubleshooting guide, once it starts growing in size. Also, user will see this error, because we do not hide it behind some dvc-specific error.

pared:
Related: https://github.com/iterative/dvc/pull/3017

static/docs/user-guide/troubleshooting.md
shcheklein: @pared so, how does exactly the message look like now with the latest PR applied?

static/docs/user-guide/troubleshooting.md
jorgeorpinel: Yes, as you prefer. I'm not about this suggestion.

jorgeorpinel:
Letting Ivan merge this one as he still has a Q in https://github.com/iterative/dvc.org/pull/875#discussion_r362073177

static/docs/user-guide/troubleshooting.md
shcheklein: @efiop @pared how about we do this properly? introduce our error codes? proper link to the section? hide details into Errno 24 into details (or even when you run it with `-v`)?

pared: @shcheklein 
This is how it looks like on current master:
https://asciinema.org/a/291591
and this is how it will look like after https://github.com/iterative/dvc/pull/3017:
https://asciinema.org/a/291592

shcheklein: So, I still don't understand why do we need this part: 

`[Errno 24] Too many open files: '/home/prd/development/projects/playground/2473/repo/.dvc/cache/10/788f0c3b6f08ef6f183f49490183d0'`

we can keep it as a `-v` detail. We already have `Too many files message before`, why duplicate it? 

That path - looks ugly. I have no idea what does it mean, what operation it failed with? How can it help me?

Second problem - as I suggested we need to come up with our own links similar to `man.dvc.org`. `https://dvc.org/doc/user-guide/troubleshooting#too-many-open-files-error` looks ugly and a burden for us to support (let's imagine we decide to move/rename it?).

Let's discuss and create our codes and URLs like `https://error.dvc.org/dvc0001` or `https://error.dvc.org/too-many-open-files`?

shcheklein:
@pared please, check my comment here https://github.com/iterative/dvc.org/pull/875/files#r362920851

static/docs/user-guide/troubleshooting.md
efiop: > we can keep it as a -v detail. We already have Too many files message before, why duplicate it?

That is how error chaining is implemented right now. `logger.exception` will automatically add the last exception so that you get some additional context. If we use `logger.error` instead, it will not show the traceback in the verbose mode, which will be harmful. To adjust that we will need to introduce a special extra flag for the log record, that would tell our logger to not append the error message but still append the stack. Or we could use `logger.error("...", stack_info=True)`, which is actually kinda meant for that :slightly_smiling_face: Created https://github.com/iterative/dvc/pull/3058 for that.

> Second problem - as I suggested we need to come up with our own links similar to man.dvc.org. https://dvc.org/doc/user-guide/troubleshooting#too-many-open-files-error looks ugly and a burden for us to support (let's imagine we decide to move/rename it?).
> Let's discuss and create our codes and URLs like https://error.dvc.org/dvc0001 or https://error.dvc.org/too-many-open-files?

Total overkill :smile:

shcheklein: > Total overkill 😄

it's not, it's hard to manage direct links and it would take 10 minutes to setup it right.

efiop: @shcheklein Got it. My thinking was that there is not enough links in use to bother with it, but I guess it is better to do sooner than later.

pared: @shcheklein @efiop 
So I should prepare some helper method, similar to `append_doc_link`?

shcheklein: yep, that would be great ... and I can help to setup the `error.dvc.org` domain and redirects. The only question left - how do we determine the suffix/"error codes"?

efiop: @shcheklein error codes are future-proof and easy to manage, but sometimes look uglier than text-based.Thinking about it, codes do seem pretty nice overall.

pared: Ok, so we do agree to set up the codes right? How do we assign them? Just as they occur? so this one will be 1? @efiop? @shcheklein I will definitely need help with setting up `error.dvc.org`

shcheklein: @pared yep, you just agree with the team on the names (I don't have any strong preferences - just to come up with a nice and simple and stable URL). I'll setup `error.dvc.org` and redirects - np with that.

pared:
Ok, lets sum up, and continue discussion here, instead of overloading one comment thread with information.
@shcheklein @jorgeorpinel @efiop :
- we set up `error.dvc.org`
- we assign custom error codes for problems described in troubleshooting guide
- we set up errors logging in that way that they do point to newly created links. 
Do I get it right?

Also, I have a big question: what is the advantage of assigning custom error codes to our troubleshooting use cases? That redirection part will be always on side of `dvc.org`, and we will not have to update `dvc` if the link for particular guide changes? In that case, we can just assign a custom redirection, like `errors.dvc.org/too-many-open-files` -> `https://dvc.org/doc/user-guide/troubleshooting#too-many-open-files`. In that case, we do not have to create an artificial concept of error code.

static/docs/user-guide/troubleshooting.md
jorgeorpinel: Commented about this on https://github.com/iterative/dvc/pull/3017#pullrequestreview-339988084. Resolving

jorgeorpinel:
> we can just assign a custom redirection, like errors.dvc.org/too-many-open-files -> https://dvc.org/doc/user-guide/troubleshooting#too-many-open-files

Here's a summary on that discussion from https://github.com/iterative/dvc.org/pull/875#discussion_r363310065 +:

> I: how do we determine the suffix/"error codes"?
> R: error codes are future-proof and easy to manage, but sometimes look uglier than text-based.Thinking about it, codes do seem pretty nice overall.
> P: Ok, so we do agree to set up the codes right?
> I: yep, you just agree with the team on the names ... just to come up with a nice and simple and stable URL

So Ruslan prefers numeric codes, while Pawel and Ivan seem neutral.

I recommend text though, but it should be very short, think of tiny URLS. In this case maybe `error.dvc.org/many-files`. It's equally future proof I think, because even if it's renamed in the future (unlikely), a new redirect to the same target can always be setup. The numeric codes have the problem of zero padding and maybe we would need to group them by hundreds or thousands similar to (HTTP statuses).

jorgeorpinel:
p.s. will the error link thing be implemented along with https://github.com/iterative/dvc/pull/3017#pullrequestreview-339988084 ?

pared:
@jorgeorpinel Yes, I was thinking that it will be a part of that change. But to be able to do it there, we need to have the troubleshooting guide in place altogether with links. I don't think that error codes are more "future-proof" than custom links redirecting to proper subsection and they introduce another "entity" to be managed by us. @efiop @shcheklein what do you think?

shcheklein:
@pared I'm totally fine with some human readable suffixes!

pared:
@shcheklein ok, so lets set up `errors.dvc.org`?

jorgeorpinel:
Done @pared. See d2293b58 ^

server.js
```diff
@@ -45,6 +45,13 @@ app.prepare().then(() => {
             : pathname)
       })
       res.end()
+    } else if (req.headers.host === 'err.dvc.org') {
```
shcheklein: let's make a regular one - just `error`? not sure we need to be cryptic here

jorgeorpinel:
p.s. I went for `err.dvc.org/{hdr}` -> `dvc.org/doc/user-guide/troubleshooting#{hdr}` BTW but feel fee to change it as needed.

server.js
```diff
@@ -45,6 +45,13 @@ app.prepare().then(() => {
...
+      // err.dvc.org/{hdr} -> dvc.org/doc/user-guide/troubleshooting#{hdr},
+      res.writeHead(301, {
+        'Cache-Control': 'no-cache',
+        Location: `https://dvc.org/doc/user-guide/troubleshooting#${pathname}`
```
shcheklein: we will have to keep one->one mapping between title in the Markdown and the error code, right? The "code" will look like "too-many-files" at least

server.js
```diff
@@ -45,6 +45,13 @@ app.prepare().then(() => {
             : pathname)
       })
       res.end()
+    } else if (req.headers.host === 'err.dvc.org') {
+      // err.dvc.org/{hdr} -> dvc.org/doc/user-guide/troubleshooting#{hdr},
+      res.writeHead(301, {
```
shcheklein: 302 is better here

server.js
jorgeorpinel: Sure? I like that it's super short kind of line `man.dvc.org` helps with tiny URL, maybe easier to print on command line. It's not like "err" could be any other word :P but as you guys prefer. Very easy change obvi

server.js
jorgeorpinel: Yes, the anchor is `too-many-open-files-error` actually. I think we **could** force it to `many-errors` with HTML in the Markdown though, what do you think? This way we can avoid having to set a redirect for each error path->header.

server.js
jorgeorpinel: I'm so confused by redirect codes at this point haha. I thought we decided having 301 everywhere was best. See https://github.com/iterative/dvc.org/issues/768#issuecomment-549486994

I have nothing agains 302 here but what about consistency vs. the`man.` redirect and others?

Sorry, I know it's a completely separate discussion to the troubleshooting guide but... Just saying.

server.js
shcheklein: 👍 if we can force it, otherwise change title to be way smaller

jorgeorpinel: Hmmm. OK I forced it in ad1b786f and it works with the desired URL hash, see https://dvc-landing-2473-amnjganfpaic8.herokuapp.com/doc/user-guide/troubleshooting#many-files

> p.s. @pared notice I moved the file to its new correct location (public/static/doc/...)

BUT the special 🔗  icon and link stopped working. I guess those are automatically rendered by `react-markdown`.

Is if OK without the clickable header links and that anchor symbol?

public/static/docs/user-guide/troubleshooting.md
```diff
@@ -1,9 +1,9 @@
 # Troubleshooting
 
 In this section we provide help for some of the problems that DVC user might
-stumbe upon.
+stumble upon.
```
jorgeorpinel: @pared consider using a spell checker plugin for your IDE 🙂

server.js
shcheklein: (let me know what name you decide to keep err, error, or something else, I'll setup the DNS record for it)

server.js
jorgeorpinel: Please vote @pared @efiop 😬 (and feel free to change the code of course), and please let Ivan know so he sets up the DNS record for the chosen subdomain.

pared: I agree with @jorgeorpinel, that `err` is hard to misinterpret. Thats not that big of an issue, can be either way, if @shcheklein thinks `error` will be more explicit, lets go this way.
@efiop ?

efiop: I agree with @shcheklein , `error` looks much more visually pleasing. I don't run into `err` very often (almost never, except in some internal code pieces), so it looks incomplete and weird to me.

pared:
@shcheklein @jorgeorpinel updated the address in docs, also the `dvc` PR, now it looks like that:
[asciicast](https://asciinema.org/a/293426)

server.js
jorgeorpinel: Ok sure as you guys prefer. I have no strong opinion against `error` 🙂

jorgeorpinel: I see this is addressed. Resolving.

jorgeorpinel: p.s. @shcheklein please setup the error.dvc.org DNS record when you can

server.js
```diff
@@ -45,6 +45,13 @@ app.prepare().then(() => {
             : pathname)
       })
       res.end()
+    } else if (req.headers.host === 'error.dvc.org') {
```
jorgeorpinel: p.s. @shcheklein please setup the error.dvc.org DNS record when you can.

jorgeorpinel:
> updated the address in docs...

Thanks @pared there's just this one matter to decide definitely on, I think: https://github.com/iterative/dvc.org/pull/875#pullrequestreview-341328486 please weight in if you can 🙂

jorgeorpinel:
And this one about 301 vs 302 response codes actually: https://github.com/iterative/dvc.org/pull/875#pullrequestreview-341328648 ...

server.js
shcheklein: ok, so in the discussion and snippet there are two things happening simultaneously. First is a backward compatible URLs like `man.dvc.org/remote-add` - and I was talking about them mostly - we don't care if they are 301 or 302 - we can even completely disable those now. Second is the actual layer of indirection for URLs like `man.dvc.org/remote/add` for those (ideally) we want to preserve the ability to being able change the location they redirect two. Considering that 301 are so aggressively cached (I didn't know about that - thus we probably had 301s everywhere), we should pick a code for `man.dvc.org` and `error.dvc.org` that is more appropriate. Like we discussed 302 is not ideal, either - so, it can be 303 or something else?

server.js
shcheklein: @jorgeorpinel @pared done!

server.js
jorgeorpinel: > (we can even completely disable those now)

(Off-topic) We only have it for `/get-url` and `/import-url`. - Will remove that exception now. See 44fe4263

jorgeorpinel: > indirection for URLs like man.dvc.org/remote/add for those (ideally) we want to preserve the ability to change the location they redirect two. Considering that 301 are so aggressively cached, we should pick a code for man.dvc.org and error.dvc.org that is more appropriate?

302 Found [has been superseded](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection) by 303 See Other, and 307 Temporary Redirect. This is a conversation pertaining more to the HTTP redirect issue actually (https://github.com/iterative/dvc.org/issues/757#issuecomment-549494538). But anyway, let's decide?

I vote for [303 See Other](https://en.wikipedia.org/wiki/HTTP_303).

shcheklein: k, let's try 303

jorgeorpinel: OK, done for existing ones in 29f3b686 of #915, and in 6630b464 for this PR.

server.js
jorgeorpinel: OK can I just get a confirmation that this ^ is OK with others before merging the PR? Thanks

shcheklein: could you summarize the `^`? :)

jorgeorpinel: Sure. So the redirect is error.dvc.org/{title} -> dvc.org/doc/user-guide/troubleshooting#{title}. The question here is about `title`.

In the website, it would naturally by a slugified version of the md header text, so in this case **Too many open files error** -> `too-many-open-files-error`. We could simply use that in the redirect BUT a) it's kind of long and b) what if we change the title later? We would need to create a new redirect each time.

We also previously agreed to use some textual codes for this, such as `many-files` in this case. But in our engine, to enforce that `id` prop of the actual `<h3>` element, we need to hardcode HTML code instead of markdown (current version). This works but the rendering is not as nice or functional as with MD – That's what I was explaining in https://github.com/iterative/dvc.org/pull/875#discussion_r365380109 above.

jorgeorpinel: > But in our engine, to enforce that id prop of the actual <h3> element, we need to hardcode HTML code

Wait... I just thought about another solution! I guess we could add hidden HTMl anchors before regular markdown headers and have the best of both worlds! (Except that clicking on the 🔗 link will produce a different hash than the tiny version from the `error...` redirect)

shcheklein: @jorgeorpinel we can create a ticket to improve the engine. It's totally fine. For now hidden HTML and/or hardcoding it totally fine.

jorgeorpinel: OK I hardcoded the whole thing by just copying the code that `react-markdown` would render, and changing the `id` and `href` HTML attributes. Please check, and I guess merge this if OK?

jorgeorpinel: p.s. now deployed: https://dvc-landing-2473-amnjganfpaic8.herokuapp.com/doc/user-guide/troubleshooting#many-files looks and works perfectly (it's just hardcoded so future updates of `react-markdown` could possibly cause problems with styles or something).

public/static/docs/user-guide/troubleshooting.md
```diff
@@ -0,0 +1,18 @@
+# Troubleshooting
+
+In this section we provide help for some of the problems that DVC user might
+stumble upon.
+
+<!-- Hardcoded link for error.dvc.org/many-files -->
+<h2 id="many-files"><a class="anchor" aria-hidden="true" href="#many-files"><svg class="octicon octicon-link" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path></svg></a>Too many open files error</h2>
```
jorgeorpinel: > can we avoid embedding SVG for now? :) let's just create an "engine" ticket please :)
From https://github.com/iterative/dvc.org/commit/d5340e6b52434a8e7488e2bdf2fc5dbf2cce5e86#r36802142

OK @shcheklein, done in 5199aa01

shcheklein: thanks : 🙏

jorgeorpinel:
Shortened nav title to just "Troubleshooting"

See https://dvc-landing-2473-amnjganfpaic8.herokuapp.com/doc/user-guide/troubleshooting

Opened _web: enable custom anchor id/href for troubleshooting guide?_ #922

PR #877, opened 2019-12-19
Regular doc updates (mid Dec)

jorgeorpinel:


package.json
```diff
@@ -9,7 +9,7 @@
     "test": "jest",
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
-    "format-check": "pretty-quick --check --verbose",
+    "format-check": "prettier --check '{.,pages/**,src/**,static/**}/*.{js,jsx,md}'",
```
shcheklein: is it better to omit extensions? and let it figure it out automatically?

jorgeorpinel: No, it tries to check literally all files (including Git-ignored ones and `.gitignore` itself) and throws a bunch of `No parser could be inferred for file` errors. You can try it yourself:

```console
$ yarn prettier --check '{.,pages/**,src/**,static/**}/*'
```

Funily, it checks `.prettierrc` (with YAML parser I guess?) and finds errors in it.

jorgeorpinel:
@shcheklein here's the glorious return of `yarn format`, please check: 9a36400

jorgeorpinel:
Wait. I did a wrong commit, had to force push it again. See 43b67ef instead please!

static/docs/command-reference/run.md
```diff
@@ -54,23 +54,20 @@ try creating a pipeline.
 
 ### Well-behaved commands
 
-DVC is simple to use, you only need to wrap your commands with `dvc run`, and
-define your dependencies and outputs.
+In a sense, `dvc run` works as a special wrapper for your commands. To prevent
+unexpected behavior when DVC executes them, your commands should ideally follow
+these rules:
```
jorgeorpinel: What "unexpected behavior" are we talking about? At `dvc run` itself, or later at `dvc repro`? Will they crash or something?

static/docs/command-reference/run.md
```diff
@@ -54,23 +54,20 @@ try creating a pipeline.
...
 
-However, to prevent unexpected behaviors, ideally, your commands should follow
-some principles:
+- Read exclusively from the specified dependencies.
+- Write exclusively to the specified outputs.
+- Completely rewrite outputs (i.e. do not append or edit).
+- Stop reading and writing files when the command exits.
```
jorgeorpinel: > - Read exclusively from the specified dependencies.
> - Write exclusively to the specified outputs.

What if there are optional inputs/outputs that users don't want to mark as deps/outs?

> - Completely rewrite outputs (i.e. do not append

Why? So the next stage has the same results as before? What if the pipeline is meant to be incremental?

> - Stop reading and writing files when the command exits.

How could a command continue writing files after it exits?

static/docs/command-reference/run.md
```diff
@@ -54,23 +54,20 @@ try creating a pipeline.
...
 
-- Read exclusively from specified dependencies.
-- Write exclusively to specified outputs.
-- Completely rewrite the outputs (i.e. do not append or edit).
-- Stop reading and writing when the command exits.
-
-To guarantee reproducibilty, your command should be
+At the very least, to guarantee reproducibility, your command should be
 [deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm) (i.e. it
```
jorgeorpinel: I think this is the only part I don't have an objection to. But I would prefer to use less imperative language i.e. "**If** you want to guarantee exact reproducibility, ..." Thoughts?

static/docs/command-reference/run.md
```diff
@@ -54,23 +54,20 @@ try creating a pipeline.
...
-must produce the same results given the same inputs/dependencies).
-
-Have in mind what brings entropy to your command (e.g. random generators, time,
-hardware, etc.) and try to minize it (e.g. fix seeds).
+produce the same outputs given the same inputs/dependencies). So avoid code that
+bring entropy into your data process (e.g. random numbers, time functions,
+hardware dependency, etc.)
```
jorgeorpinel: I'm not sure the we're really talking about "entropy" here. And it sounds too scientific (not public friendly). But I'll wait for answers on the other comments before changing this...

package.json
shcheklein: 👍

static/docs/command-reference/run.md
shcheklein: yep

static/docs/command-reference/run.md
shcheklein: it's fine I think for CR, considering that we define examples

static/docs/command-reference/run.md
shcheklein: I think both - run and repro.

In case of DVC the major "unexpected behavior" that comes to my mind is inconsistency between DVC-file (checksums) and actual data. Command writes to an output of another command. Or command writes to an input to itself.

Less obvious - when files (dependencies, outputs, DVC-files) look consistent but when you run repro you get a completely different result. A missing dependency case. To some extent it can be also considered an "unexpected behavior".

static/docs/command-reference/run.md
shcheklein: > What if there are optional inputs/outputs that users don't want to mark as deps/outs?

I think the whole point of the paragraph is to describe an ideal case.

> Why? So the next stage has the same results as before? What if the pipeline is meant to be incremental?

DVC deletes outputs before executing the command again. By depending on the output of the previous run you introduce "cycles" to some extent (command depends on it's own output) to make it reproducible - you need to checkout the previous versions of the output for `dvc repro` to reproduce the current one - it's a mess. That's why we even have an option `--persistent-outs` or something like this - but they are still not documented.

> How could a command continue writing files after it exits?

cc @mroutis @Suor - could you chime in?

static/docs/command-reference/run.md
jorgeorpinel: > - Inconsistency between DVC-file (checksums) and actual data: Command writes to an output of another command, or command writes to an input to itself.

Ah, I see. Al right. So basically stage/pipeline behaviors that DVC doesn't support? Maybe we should rename the section something like that because "well behaved" sounds like we have a strong opinion about how data pipelines should work in general.

> - Files look consistent  but when you run repro you get a completely different result. A missing dependency case.

I don't understand this one. Is the files including dependencies are consistent how can the dependency be missing? And if a dependency file is missing you can just `dvc pull` to restore it, no? Maybe I'm not getting the point. But anyway, my main suggestion here is the above (not sounding opinionated).

mroutis: > In case of DVC the major "unexpected behavior" that comes to my mind is inconsistency between DVC-file (checksums) and actual data.

Yes, this is the reason we are including this section.

> I don't understand this one. Is the files including dependencies are consistent how can the dependency be missing? And if a dependency file is missing you can just dvc pull to restore it, no? Maybe I'm not getting the point. But anyway, my main suggestion here is the previous one (not sounding opinionated).

@jorgeorpinel , I think @shcheklein was talking about forgetting to specify a dependency in a run command:
```
dvc run -d data -o myscript.py data
```

(you are not tracking your code, thus, `status` will not know whether DVC needs to reproduce it again or not)

static/docs/command-reference/run.md
mroutis: > I think the whole point of the paragraph is to describe an ideal case.

This :point_up: 

> How could a command continue writing files after it exits?

This is common when you use another tool that keeps running in the background.

Here's more context about it, @jorgeorpinel -- https://github.com/iterative/dvc/issues/2542#issuecomment-546553710

static/docs/command-reference/run.md
mroutis: entropy -> chaos

static/docs/command-reference/run.md
```diff
@@ -127,7 +124,7 @@ hardware, etc.) and try to minize it (e.g. fix seeds).
   `dvc run` expects that dependencies, outputs, metric files are specified
   relative to this directory. This value is saved in the `wdir` field of the
   stage file generated (as a relative path to the location of the DVC-file) and
-  is used by `dvc repro` to change the working directory before execuring the
+  is used by `dvc repro` to change the working directory before executing the
```
mroutis: my bad, @jorgeorpinel :see_no_evil:

static/docs/command-reference/run.md
jorgeorpinel: OK. I changed the title to "Unsupported command behavior" and made other small improvements, please review.

static/docs/command-reference/run.md
jorgeorpinel: > DVC deletes outputs before executing the command again.

Ah, this is an important note we should probably clearly mention then. Added.
Should we remove the `--remove-outs` option from docs and tab completion at this point?

> That's why we even have an option --persistent-outs or something like this - but they are still not documented.

Should we create a separate ticket to document this also? Sounds important.

static/docs/command-reference/run.md
jorgeorpinel: OK, changed.

static/docs/command-reference/run.md
jorgeorpinel: Haha I really doubt we're talking about thermodynamic entropy. I'll link to https://en.wikipedia.org/wiki/Software_entropy 🙂

static/docs/command-reference/run.md
jorgeorpinel: All good 😋

src/Tooltip/index.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import includes from 'lodash.includes'
 
-import glossary from '../Documentation/glossary'
+import glossary from '../../static/docs/glossary'
```
jorgeorpinel: @shcheklein @iAdramelk notice that similarly to sidebar.js I moved glossary.js into static/docs/

Should we also move pages/ into static/ BTW?

shcheklein: I would say `pages` looks like `src` to me. Probably we keep them in `pages` only because it's a Next convention or something? @iAdramelk should have a better idea.

`glossary` - makes sense to me.

static/docs/command-reference/run.md
shcheklein: > Should we create a separate ticket to document this also? Sounds important.

We have a ticket but we don't want to execute on it. No one likes this option :) And we've been thinking if there is a better way to solve the problem.

> Should we also remove the `--remove-outs` option from docs and tab completion at this point? 

I hope it's hidden in DVC CLI help. Yes we def should remove it from autocomplete scripts. I would keep in docs for now (with a deprecation note) since users might hit a script where it is used and might wonder how does it behave. (btw, I'm lost a bit - how did we come to `--remove-outs`?)

> Ah, this is an important note we should probably clearly mention then.

Again, I hope it's already mentioned in `dvc repro`\`dvc run`? If not it worth mentioning indeed.

jorgeorpinel: > I hope it's hidden in DVC CLI help. Yes we def should remove it from autocomplete scripts.

It's not 🙁 OK I opened iterative/dvc/issues/2988 for this.

> I would keep in docs for now (with a deprecation note) since users might hit a script where it is used and might wonder how does it behave.

OK I reworded it a little bit (in 4778205f) to make the deprecation more obvious.

> btw, I'm lost a bit...

Sorry, I use  "regular update" PRs for misc. small stuff.

> I hope it's already mentioned in dvc repro`dvc run`?

Its currently not, but this PR does introduce the note (in ced036f).

static/docs/command-reference/repro.md
```diff
@@ -39,6 +39,8 @@ By default, this command recursively searches in pipeline stages, starting from
 the `targets`, to determine which ones have changed. Then it executes the
 corresponding commands.
 
+⚠️ DVC deletes <abbr>output files</abbr> before executing commands.
+
```
jorgeorpinel: > It sounds and looks a bit too dramatic. I think it deletes only cached files. Sometimes it means just removing links to cache. If there is a possibility of a data loss it will probably prompt (which might change with DVC deprecating prompts soon and will be failing in that case).

@shcheklein OK so I guess I'll ask maybe @efiop to confirm the behavior? Meanwhile, I've softened the note as you describe above in 99a59b0a.

> Also, it's probably better to say deletes outputs since they can be directories.

Good point! I did a quick review of the "output file" term throughout the docs in e74a207e.

static/docs/command-reference/run.md
```diff
@@ -52,25 +52,23 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
-[deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm) (i.e. it
-must produce the same results given the same inputs/dependencies).
-
-Have in mind what brings entropy to your command (e.g. random generators, time,
-hardware, etc.) and try to minize it (e.g. fix seeds).
+At the very least, if you want to guarantee exact reproducibility, your code
+should be [deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm)
+(i.e. always produce the same output for a given input). So avoid code that
+bring [entropy](https://en.wikipedia.org/wiki/Software_entropy) into your data
+pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
```
jorgeorpinel: > `At the very least` -> `Ideally`. Completely deterministic code is extremely hard to achieve in ML

@shcheklein updated.

> `to guarantee reproducibility` - there are different levels of reproducibility - thus I would not tie it directly with deterministic algorithms

That's why I wrote "exact reproducibility". But can you look at the new text and suggest a change? I'm not sure how to improve it. This link between reproducibility and determinism basically comes from Alex and Ramon's version previously merged.

shcheklein: I would still omit "exact" ... it sounds like that you should read and care this only if you need the "exact" one. But usually you just have an approximate reproducibility and it can be enough.

static/docs/command-reference/repro.md
```diff
@@ -37,7 +37,8 @@ omitted, `Dvcfile` will be assumed.
 
 By default, this command recursively searches in pipeline stages, starting from
 the `targets`, to determine which ones have changed. Then it executes the
-corresponding commands.
+corresponding commands.<br /> Note that DVC removes cached <abbr>outputs</abbr>
```
shcheklein: hmm, why do we need html here?

static/docs/command-reference/run.md
```diff
@@ -52,25 +52,25 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
-- Completely rewrite the outputs (i.e. do not append or edit).
-- Stop reading and writing when the command exits.
-
-To guarantee reproducibilty, your command should be
-[deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm) (i.e. it
-must produce the same results given the same inputs/dependencies).
-
-Have in mind what brings entropy to your command (e.g. random generators, time,
-hardware, etc.) and try to minize it (e.g. fix seeds).
+### Unsupported command behavior
```
shcheklein: this title sounds very strange to me ... may be because "unsupported" word is too strict?

static/docs/command-reference/run.md
```diff
@@ -152,9 +152,9 @@ hardware, etc.) and try to minize it (e.g. fix seeds).
   command is non-deterministic (meaning it produces different outputs from the
   same list of inputs).
 
-- `--remove-outs` - it removes stage outputs before executing the command. If
-  `--no-exec` specified outputs are removed anyway. This option is enabled by
-  default and deprecated. See `dvc remove` as well for more details.
+- `--remove-outs` - DEPRECATED - remove stage outputs before executing the
```
shcheklein: I don't like using all upper case - can we for example adopt ~notation~? or (_deprecated_) should be enough

static/docs/command-reference/repro.md
jorgeorpinel: New line between the first sentence and the note:

image

Idk. Stylistic decision. There's a few `<br />`s here and there in docs. Should I review and try to remove them?

static/docs/command-reference/run.md
jorgeorpinel: How about **Reproducibility notes**?

static/docs/command-reference/run.md
jorgeorpinel: How about "Keep in mind that if the pipeline's reproducibility goals include consistent
output data, ..." ?

jorgeorpinel: Updated.

static/docs/command-reference/run.md
jorgeorpinel: Updated.

static/docs/command-reference/run.md
jorgeorpinel: OK, changed to (_deprecated_).

package.json
```diff
@@ -9,8 +9,9 @@
     "test": "jest",
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
-    "format-check": "pretty-quick --check --verbose",
-    "lint-check": "eslint src pages"
+    "format-check": "prettier --check '{.,pages/**,src/**,static/**}/*.{js,jsx,md}'",
+    "lint-check": "eslint src pages",
+    "format": "prettier --write '{.,pages/**,src/**,static/**}/*.{js,jsx,md}'"
```
shcheklein: let's make it `format-all` as we discussed?

static/docs/command-reference/repro.md
shcheklein: 👍 I think it's fine

static/docs/glossary.js
```diff
@@ -67,6 +67,14 @@ extracted features or a ML model file) that is under DVC control. Refer to
 **output*.
       `
     },
+    {
+      name: 'Dependency',
+      match: ['dependency', 'dependencies'],
+      desc: `
+A file or directory (typically under DVC control) recorded in the \`deps\`
```
shcheklein: `typically under DVC control` is not correct - I would say it's just any file or directory or even external deps.

shcheklein: e.g. using code file as a dependency is a very regular option

static/docs/command-reference/run.md
```diff
@@ -52,25 +52,25 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
+We don't want to tell you how to write your code! However, please be aware that
+in order to prevent unexpected results when DVC executes or reproduces your
+commands, they should ideally follow these rules:
+
+- Read/write exclusively from/to the specified <abbr>dependencies</abbr> and
+  <abbr>outputs</abbr>.
+- Completely rewrite outputs (i.e. do not append or edit).<br/> Note that DVC
+  removes cached outputs before running the stages that produce them (including
+  at `dvc repro`).
+- Stop reading and writing files when the `command` exits.
```
shcheklein: minor: probably remove ticks `command` -> command since we don't emphasize it before?

static/docs/command-reference/run.md
```diff
@@ -52,25 +52,25 @@ captures data and <abbr>caches</abbr> relevant <abbr>data artifacts</abbr> along
...
-- Completely rewrite the outputs (i.e. do not append or edit).
-- Stop reading and writing when the command exits.
-
-To guarantee reproducibilty, your command should be
-[deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm) (i.e. it
-must produce the same results given the same inputs/dependencies).
-
-Have in mind what brings entropy to your command (e.g. random generators, time,
-hardware, etc.) and try to minize it (e.g. fix seeds).
+### Reproducibility notes
```
shcheklein: so, this is not only about reproducibility ... this is about avoiding unexpected behavior, right? like in the case when command keeps writing files it can even corrupt the cache

src/Tooltip/index.js
iAdramelk: I do also think that correct place the for pages are inside the `src` folder and `gatsby` place theme there, but default place for `next` pages is root, and I can't find a way to change it in the `next`'s docs. But there is a lot of things that I can't find there, so it is possible that it is configurable is some way.

P. S. Maybe we should convert glossary for `.js` to `.json` if it is only data?

jorgeorpinel: I can't remember why but there was a reason we had to use a JS file for the glossary. I think having to do with formatting the md to 80 cols. Anyway, leaving pages/ in root since it's a Next thing and outside the docs engine.

package.json
jorgeorpinel: Yep. Done in 7aaa8aa0.

static/docs/glossary.js
jorgeorpinel: True, thanks! Fixed.

static/docs/command-reference/run.md
jorgeorpinel: `command` is a `dvc run` argument and it's surrounded with back ticks several times above in the description. But "command" is also just a word that I use without back ticks in some other sentences.

jorgeorpinel: But reviewing this I noticed some inconsistencies and errors in sentences around this term. Updated in 884bdef4.

static/docs/command-reference/run.md
jorgeorpinel: True. Changed to **Avoiding unexpected behavior** 😁

Issue #724, opened 2019-10-22
nav: unnecessary gaps in sidebar


algomaster99:
I think there is unnecessary space between 'Shared Development Server' and 'User Guide'.
![Screenshot from 2019-10-22 00-05-06](https://user-images.githubusercontent.com/35191225/67328867-b595a380-f537-11e9-9696-5f50de12fb18.png)

UPDATE: Note that a PR for this could also solve this related issue: #856

shcheklein:
@algomaster99 I can's reproduce this. What browser do you use? Is it production?

jorgeorpinel:
I also don't see this. Try ford reloading the site on your browser Aman. Shall we close as `invalid`?

algomaster99:
@shcheklein @jorgeorpinel I sometimes see this and sometimes not. I can't figure out what is happening.

shcheklein:
@algomaster99 could you check the console? could you check the inspector in this place?

algomaster99:
@shcheklein 
![Screenshot from 2019-10-23 02-20-14](https://user-images.githubusercontent.com/35191225/67331668-aca6d100-f53b-11e9-9a51-39c775d2f5b3.png)

The harcoded `height: 119px` might be a problem. I will have too see.

jorgeorpinel:
I get 94px correctly. But I was able to reproduce from https://www.webpagetest.org/result/191022_SR_d9e1db4a5e961674a3f0fd9d512cd22a/ so it's definitely not just Aman's browser.

shcheklein:
@algomaster99 @algomaster99 does it happen with other sections?

algomaster99:
@shcheklein No, I don't think so.

shcheklein:
Got the same problem with this configuration:

<img width="409" alt="Screen Shot 2019-10-23 at 11 56 40 AM" src="https://user-images.githubusercontent.com/3659196/67425131-4f873b80-f58c-11e9-939f-12985e9c5900.png">

Probably related to the way item height is calculated when it's close to the very edge and close to taking two lines

iAdramelk:
I think I found out why this bug happens:

1. We use custom font Brandon on our site, but before font file is loaded other font is shown (system default sans-serif font).
2. And sans-serif is wider then the Brandon. Because of that some menu items are wrapped to the second line.
3. So if the script that calculates block height for animations is fired before font was loaded it calculates incorrect height which results in the bug.

There is a few possible solutions for this problem:

[...]

PR #879, opened 2019-12-23
[WIP] engine: move static/ contents into public/ and update links

jorgeorpinel:
Per node 9.0.6+ update
See https://github.com/zeit/next.js/blob/master/errors/static-dir-deprecated.md
And https://nextjs.org/docs#static-file-serving-eg-images

\+ close #804 (adopt restyled.io) using this PR to test it.

PR #882, opened 2019-12-24
engine: move static/ contents into public/ and update links, et al.

### LINKED ISSUE #789: app: sort es6 imports
Currently there's not much order in js file imports. Some files group them with comments or spaces in between e.g.

https://github.com/iterative/dvc.org/blob/9c8cc540cfbd004db9ac3747bd56b6ddc328ed24/pages/doc.js#L3-L24

or

https://github.com/iterative/dvc.org/blob/9c8cc540cfbd004db9ac3747bd56b6ddc328ed24/src/Documentation/SidebarMenu/SidebarMenu.js#L1-L14

I suggest we remove all these comments, which don't really add any value, and apply this ESLint rule: https://eslint.org/docs/rules/sort-imports

### LINKED ISSUE #804: lint: adopt restyled.io
Check the [supported languages](https://github.com/restyled-io/restyled.io/wiki/Available-Restylers).

(This may not require a PR.)

This should also not replace the pre-commit hook that runs `yarn format-check` and `yarn lint`.

### LINKED ISSUE #854: Js console errors/warnings
When running the server:

```console
The static directory has been deprecated in favor of the public directory. https://err.sh/zeit/next.js/static-dir-deprecated
```

When going to /doc:

```console
Warning: `NaN` is an invalid value for the `height` css style property.
    in div
    in Context.Consumer
    in Context.Consumer
    in StyledComponent
    in styled.div
    in Fragment
    in SidebarMenuItem
    in div ...
```

jorgeorpinel:
Per Node 9.0.6+ update
See https://github.com/zeit/next.js/blob/master/errors/static-dir-deprecated.md
And https://nextjs.org/docs#static-file-serving-eg-images

\+ Close #804 (adopt restyled.io) using this PR to test it.
\+ Fix #854 console errors
\+ Close #789


jorgeorpinel:
> format: add three dashes `---` at the beginning of .restyled.yaml

That seems to have killed **Restyled**...

jorgeorpinel:
OK I'll leave Restyled alone for some time... See if it fixes itself. Back to the /doc 404 page error:

From https://github.com/iterative/dvc.org/pull/879#issuecomment-568533563:
> /doc shows the 404 page (with a 200 status code BTW). Other /doc/{valid} routes also show 404s (but with actual 404 code).
> I think it has to do both with my move from static to public, and with ... https://github.com/iterative/dvc.org/pull/872/files#diff-78c12f5adc1848d13b1c6f07055d996eR110-R113 maybe.

jorgeorpinel:
Hey @iAdramelk and @shcheklein I have to ask you for help on this one. And granted, it's not some sort of p1 but it's been useful for me to get my hands on the Node/Next code so your guidance is appreciated.

As you can see here, I moved static/ -> public/ and updated all the imports and links accordingly (see PR description for more context). However now /doc isn't working, showing a 404 page (with 200 status) and all the valid paths inside as well (but with actual 404 status).

I setup a debugger script in 1fcb0a6 and followed the code in server.js step by step so I know how we get to these errors but I'm not sure how exactly. I know it has to do with SSR because the server.js code runs twice, once for `pathname` "/doc" and another one with `pathname` "/docs/get-started/index.md". The latter value is not found in the sidebar JSON struct by `getItemByPath` (in src/utils/sidebar.js).

My questions are

1. What happens between server.js runs? I don't see a 301 redirection from /doc to /docs/get-started so I guess this is the SSR "middleware" Alexey implemented.
1. Why is it looking for that string? ("/docs/get-started/index.md") I think it should be looking for just "/docs/get-started" – Does my change from static/ to public/ cause this?

Thanks!

jorgeorpinel:
> I think it has to do with SSR because server.js is executed 2 (or 3) times, once for pathname "/doc", and again for pathname "/doc/get-started/index.md"

p.s. actually it does run 3 times: /doc -> /docs/get-started/index.md -> /doc/get-started/index.md FYI

iAdramelk:
@jorgeorpinel my wild guess would be that we have problems because of this line: https://github.com/iterative/dvc.org/blob/master/server.js#L98

Basically, I think that at some point in time in the past, the docs URL's were starting with `/docs/*`, and not `/doc/*`, it was changed later, and this redirect was added.

And the folder with our docs is called exactly `/docs/`. In the past, it wasn't a problem because we were using the path `/static/docs/`, but after moving to `/public/` folder it became just `/docs/`, and now redirects are catching it and forward paths that starting with that to `/doc/` folder.

jorgeorpinel:
> our docs is called exactly docs/ ... after moving into public/ its path became just /docs/

🤦‍♂ Yep. I was making complex hypothesis over a routing problem... It's still happening inside the SSR mechanism though, from what I can see in:

https://github.com/iterative/dvc.org/blob/9ee6eac5baa0bf952ff4a6a9d216041c56e63299/pages/doc.js#L144-L145

Anyway, thanks for the educated guess, @iAdramelk! Fixed in 377ab90 🙂

server.js
```diff
@@ -95,7 +95,7 @@ app.prepare().then(() => {
         )
       })
       res.end()
-    } else if (/^\/doc.*/i.test(pathname)) {
+    } else if (/^\/doc(?!s\/).*/i.test(pathname)) {
       // path /doc*/... -> /doc/...
```
shcheklein: this is not true anymore, right? 

what is the current semantics of this redirect?

shcheklein:
@jorgeorpinel what is the current status of this? could you summarize how do redirects work now?

jorgeorpinel:
> OK I'll leave Restyled alone for some time... See if it fixes itself.

So I pushed .restyled.yaml back on here and it finally works. I configured it so it behaves the same as the pre-commit hook script (`yarn format-check`). As you can see, I merged the Restyled PR in 492fca0 which makes this PR pass the formatting CI tests.

server.js
jorgeorpinel: It is. Except for /docs/* paths. Updated note in 361849e.

Another option is to revert this change and move the .md files from public/docs/ to somewhere else that doesn't overlap with this redirect, maybe public/md/ ?

jorgeorpinel: I mean in 43303c0

jorgeorpinel:
> could you summarize how do redirects work now?

@shcheklein please see https://github.com/iterative/dvc.org/pull/882#discussion_r361496960 on this. Not much changed, really.

jorgeorpinel:
From #854:
> `Warning: `NaN` is an invalid value for the `height` css style property.`

I found this happens in line 29 below:

https://github.com/iterative/dvc.org/blob/9ee6eac5baa0bf952ff4a6a9d216041c56e63299/src/Documentation/SidebarMenu/SidebarMenu.js#L26-L29

which makes sense because `heightMap ` is defined as an empty object in line 19, and never populated, so the expression is evaluated to `height += undefined = NaN`. I'm addressing this for now by changing line 29 to  `height += heightMap[current] ? heightMap[current] : 0` (in 514b3fa0), but I don't think that fixes the original intention.

I don't know what the original intention was though, do you guys know @iAdramelk @shcheklein? So I can properly implement it.

iAdramelk:
> I don't know what the original intention was though, do you guys know @iAdramelk @shcheklein? So I can properly implement it.

It runs on client after the component finishes rendering to calculate menu block height to properly do animations. We need to rewrite this code anyway to make server side renders side menu correctly and fix #724.

server.js
shcheklein: so, we have an extremely strange combination now, let's cleanup it somehow.

docs redirects to doc
docs/ does not redirect
documentation redirects
docsddfsersfasdsaer redirects

let's just kill this rule for simplicity? I think it's outdated (when we had the previous domain - dataversioncontrol.com) anyway and can be handler by the cloudflare.

Also it's a bit strange to have `/docs/something.md` and /doc/something` I liked the previous location - it was clear that it's static - `static/docs/...`

jorgeorpinel: /doctor-pepper redirects 😋 

Ok, I killed the redirect now but had to leave the rule for the part that enforces 404 response codes. See 5f9deb25.

> I liked the previous location - it was clear that it's static

Yeah but it's now [deprecated](https://github.com/zeit/next.js/blob/master/errors/static-dir-deprecated.md) from Node. public/ is pretty obvious too, I think.

shcheklein: location itself is fine, it's confusing to have `dvc.org/docs/get-started/index.md` and `dvc.org/doc/get-started` ... and the only differentiator is `s` between valid path prefixes ... it was way better to have exactly `dvc.org/static/docs/get-started/index.md`

jorgeorpinel: I mean. I can undo this whole change but I'm just following best practices here. Also, your example `/docs/get-started/index.md` is a static asset, not a routed path in our docs engine, like `/doc/get-started`. We never link to static assets directly anyway (with or without static/), so it won't be confusing to users, just maybe to us as developers (or possibly to other contributors). That's why I suggested moving the public/docs/ folder to something like public/md/ , but up to you 🙂

shcheklein: I think it should be `public/static/*`:

https://nextjs.org/blog/next-9-1

```
Because public also covers the static directory use case we have decided to deprecate the static directory in favor of creating a public/static folder with the same functionality.
```

it means that public is not a replacement for static.

jorgeorpinel: OK, will do. (In fact that was an intermediate step in this PR but I didn't see need for the additional directory back then.)

And restore the /doc*/{subpath} -> /doc/{subpath} redirect? Or leave that out?

shcheklein: It makes some sense to me to have `docs -> doc`, `documentation -> doc` and normalization `doc/` -> `doc` (actually we probably need this for any valid doc/* URL).

jorgeorpinel: OK I reinstated the redirect and moved public/docs/ to public/static/docs/, fixing all references accordingly (including some previously broken ones).

> normalization doc/ -> doc (actually we probably need this for any valid doc/* URL)

Good idea. I also implemented this via 301 redirects. Is that OK? I remember we've had problems with 301 redirects before... (They're very hard to undo.)

Note that trailing slashes in regular Next pages like /features/ result in a 404. Should we care about this inconsistency? Or perhaps implement this in our docs engine instead of 301 redirects to remove the trailing slash?

jorgeorpinel:
OK all my goals in this PR are achieved. Should be ready to merge. Moving onto the next Next thing 😋

server.js
jorgeorpinel: p.s. re 30x redirects: refer to the conversation in https://github.com/iterative/dvc.org/issues/757#issuecomment-549494538.

jorgeorpinel:
Also, please see #888 which is kind of a continuation of this.

package.json
```diff
@@ -5,13 +5,14 @@
...
     "build": "next build",
     "test": "jest",
     "start": "NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
-    "format-check": "prettier --check '{.,pages/**,src/**,static/**}/*.{js,jsx,md}'",
+    "format-check": "prettier --check '{.,pages/**,public/static/docs/**,src/**}/*.{js,md}'",
     "lint-check": "eslint src pages",
-    "format-all": "prettier --write '{.,pages/**,src/**,static/**}/*.{js,jsx,md}'",
+    "format-all": "prettier --write '{.,pages/**,public/static/docs/**,src/**}/*.{js,md}'",
     "format": "prettier --write"
```
shcheklein: so, how does `yarn format` work? is it documented somewhere?

pages/_document.js
```diff
@@ -58,33 +59,25 @@ export default class Page extends Document {
...
               rel="icon"
               type="image/png"
-              href="/static/favicon-32x32.png"
+              href="/favicon-32x32.png"
               sizes="32x32"
             />
             <link
               rel="icon"
               type="image/png"
-              href="/static/favicon-16x16.png"
```
shcheklein: they all should be part of the `/public/static` ... may be introduce even `/public/static/img` or something similar

pages/_document.js
```diff
@@ -58,33 +59,25 @@ export default class Page extends Document {
...
+              href="/favicon-16x16.png"
               sizes="16x16"
             />
-            <link
-              rel="stylesheet"
-              type="text/css"
-              href="/static/fonts/fonts.css"
-            />
+            <link rel="stylesheet" type="text/css" href="/fonts/fonts.css" />
```
shcheklein: same here - it's all part of `static` probably

pages/_document.js
shcheklein: probably the way favicons are set can be reviewed, but it looks it's good enough to be part of `static`, not `root`. May be it worth making them `static/img/favicon/...`

pages/features.js
```diff
@@ -30,7 +27,7 @@ export default function FeaturesPage() {
           <Feature>
             <Icon>
               <img
-                src="/static/img/features/icons/git-icon.svg"
+                src="/img/features/icons/git-icon.svg"
```
shcheklein: same, don't see a good reason to change URLs here

server.js
```diff
@@ -96,8 +96,11 @@ app.prepare().then(() => {
       })
       res.end()
     } else if (/^\/doc.*/i.test(pathname)) {
-      // path /doc*/... -> /doc/...
-      let normalized_pathname = pathname.replace(/^\/doc[^?/]*/i, '/doc')
+      // path /doc*/{item} -> /doc/{item} and trim trailing / if present
```
shcheklein: why would we want to keep `doc-something`? can we do exclusively `docs` and `documentation`? - that should be enough.

shcheklein: and let's create a ticker for a proper normalization for all URLs? there should be a simple general way to achieve this.

package.json
jorgeorpinel: It just formats all the relevant js and md files. It's mentioned in https://dvc.org/doc/user-guide/contributing/docs#doc-style-guidelines-java-script-and-markdown:
image

pages/_document.js
jorgeorpinel: Woops forgot to remove img/ and the favicons to static/

server.js
jorgeorpinel: Idk. This is the way the redirect was before. I thought whoever wrote this meant to catch all "document"-like words. I can limit it to docs and documentation.

pages/_document.js
jorgeorpinel: Fixed

pages/_document.js
jorgeorpinel: OK, moved the loose image files (favicons, social-share.png) into img/, and will put /img/ back into the the/static/img/ path.

pages/features.js
jorgeorpinel: All this stuff back in /static/... now! Thanks

server.js
jorgeorpinel: Fixed but may need double checking... Will open a review.

server.js
```diff
@@ -95,9 +95,12 @@ app.prepare().then(() => {
...
       res.end()
-    } else if (/^\/doc.*/i.test(pathname)) {
-      // path /doc*/... -> /doc/...
-      let normalized_pathname = pathname.replace(/^\/doc[^?/]*/i, '/doc')
+    } else if (/^\/(doc|docs|documentation)\//i.test(pathname)) {
+      // path /doc*/{item} -> /doc/{item} and trim trailing / if present
+      let normalized_pathname = pathname.replace(
+        /^(\/[^/?]*)(\/.*[^/])?[/?]*/i,
+        '/doc$2'
+      )
```
jorgeorpinel: Can you guys help me double check my regexes here?
The first one is simple, `/^\/(doc|docs|documentation)\//`
But the second one, `/^(\/[^/?]*)(\/.*[^/])?[/?]*/`
is a bit confusing. I've tested it extensively by hand but I just don't want to generate bad 301 redirects, which are very hard to undo later. Thanks @iAdramelk @shcheklein

public/static/docs/user-guide/contributing/docs.md
```diff
@@ -134,4 +152,4 @@ pre-commit hook that is integrated when `yarn` installs the project dependencies
 
 > Check out the `.md` source code of any command reference to get a better idea,
 > for example in
-> [this very file](https://raw.githubusercontent.com/iterative/dvc.org/master/static/docs/user-guide/contributing/docs.md).
+> [this very file](https://raw.githubusercontent.com/iterative/dvc.org/master/docs/user-guide/contributing/docs.md).
```
shcheklein: this is wrong URL, right? Also, this same file have other links to `sidebar.json` (and other files?). I would check README as well.

public/static/docs/user-guide/contributing/docs.md
```diff
@@ -9,8 +9,7 @@ run the website.
 
 To contribute documentation you need to know these locations:
 
-- [Content](https://github.com/iterative/dvc.org/tree/master/static/docs)
-  (`/static/docs`):
+- [Content](https://github.com/iterative/dvc.org/tree/master/docs) (`/docs`):
   [Markdown](https://guides.github.com/features/mastering-markdown/) files of
   the different pages to render dynamically in the browser.
 - [Images](https://github.com/iterative/dvc.org/tree/master/static/img)
```
shcheklein: wrong now?

public/static/docs/user-guide/contributing/docs.md
```diff
@@ -9,8 +9,7 @@ run the website.
 
 To contribute documentation you need to know these locations:
 
-- [Content](https://github.com/iterative/dvc.org/tree/master/static/docs)
-  (`/static/docs`):
+- [Content](https://github.com/iterative/dvc.org/tree/master/docs) (`/docs`):
```
shcheklein: wrong now?

server.js
shcheklein: there was not specific reason behind this decision. I think it's outdated and we can make it more strict and aligned with what we actually need on the dvc.org . As I mentioned - I can understand `docs` or `documentation` being aliases for the `doc`. But we should stop supporting some random URLs - I don't see any reasons for this.

server.js
```diff
@@ -73,7 +73,7 @@ app.prepare().then(() => {
     } else if (/^\/doc\/tutorial\/(.*)?/.test(pathname)) {
       // path /doc/tutorial/... -> /doc/tutorials/deep/...
       res.writeHead(301, {
-        Location: req.url.replace('/doc/tutorial/', '/doc/tutorials/deep/')
+        Location: req.url.replace('/doc/tutorial/', '/doc/tutorials/deep')
```
shcheklein: It does not work properly for the case like `/doc/tutorial/something`

server.js
shcheklein: see my comment to the this redirect in general. We are solving a non existent problem by supporting this. No reason for this. It should be simplified significantly.

server.js
jorgeorpinel: OK. I'll follow up in https://github.com/iterative/dvc.org/pull/882#pullrequestreview-337006182

server.js
jorgeorpinel: > I think it's outdated and we can make it more strict and aligned with what we actually need on the dvc.org . As I mentioned - I can understand docs or documentation being aliases for the doc. But we should stop supporting some random URLs

Yes, I have already simplified it. The first regex is the one that matches the URL path exactly to `/^\/(doc|docs|documentation)\//`. The 2nd one is to normalize it into `/doc/{item}` (without trailing slashes.

But actually I just realized the first one won't always have `/` at the end... And the 2nd one doesn't have to be so complicated since the first one has matched. They can probably be the same one or very similar. I'm looking into this now.

server.js
jorgeorpinel: Yeah there was no reason to change this, I was confused. I reverted this, thanks.
But I updated the other one that matched exact path strings `/doc/tutorial` or `/doc/tutorial/` to match regex `/^\/doc\/tutorial\/*$/i` now (redirects to /doc/tutorials).

server.js
```diff
@@ -64,10 +63,25 @@ app.prepare().then(() => {
       // path /(help|chat) -> Discord chat invite
       res.writeHead(301, { Location: 'https://discordapp.com/invite/dvwXA2N' })
       res.end()
+    } else if (/^\/(docs|documentation)(\/.*)?$/i.test(pathname)) {
+      // path /docs... or /documentation... -> /doc...
+      res.writeHead(301, {
+        Location: req.url.replace(/\/(docs|documentation)/i, '/doc')
+      })
+      res.end()
     } else if (/^\/doc\/commands-reference(\/.*)?/.test(pathname)) {
```
shcheklein: `$` is probably missing

server.js
```diff
@@ -64,10 +63,25 @@ app.prepare().then(() => {
...
       // path /doc/commands-reference... -> /doc/command-reference...
       res.writeHead(301, {
-        Location: req.url.replace('commands-reference', 'command-reference')
+        Location: req.url.replace(
```
shcheklein: why do we need this change?

server.js
```diff
@@ -64,10 +63,25 @@ app.prepare().then(() => {
...
+          '/doc/commands-reference',
+          '/doc/command-reference'
+        )
+      })
+      res.end()
+    } else if (/^\/doc\/tutorial\/*$/.test(pathname)) {
```
shcheklein: can it be joined with the next one?

server.js
jorgeorpinel: As in `/^\/doc\/commands-reference(\/.*)?$/`? I don't think it makes a difference. Both would work, yes.

server.js
jorgeorpinel: 1. A bit of consistency: the other `req.url.replace` calls use the full path from root `/`.
2. It makes it less probable that the string to replace is repeated in the URL (already unlikely with just `commands-reference`, but much less with the whole `/doc/commands-reference`.

server.js
jorgeorpinel: Actually you're right, better to include those `$` in regexes to avoid catching longer strings if the `(\/.*)?` part is not present (e.g. `/doc/commands-references`). Done in e76f33bc

server.js
jorgeorpinel: p.s. I just tested #2 here and it's not the case. So just consistency with all the other `req.url.replace` in the file.

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: Yeah, thanks for catching this. Will fix.

jorgeorpinel: *these

server.js
jorgeorpinel: I guess the trailing slash behavior inconsistency is not a big deal. Resolving

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: Same as https://github.com/iterative/dvc.org/pull/882#pullrequestreview-337007217

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: Fixed this one.

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: Fixed

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: > this same file have other links to sidebar.json (and other files?). I would check README as well

Double checked (and fixed) all `iterative/dvc.org` links throughout repo.

server.js
jorgeorpinel: p.s. did a similar change in https://github.com/iterative/dvc.org/pull/882/files#diff-78c12f5adc1848d13b1c6f07055d996eL93-R104

server.js
```diff
@@ -64,62 +66,64 @@ app.prepare().then(() => {
...
-      pathname === '/doc/use-cases/data-and-model-files-versioning/'
+      /^\/doc\/use-cases\/data-and-model-files-versioning\/?$/.test(pathname)
     ) {
       // path /doc/use-cases/data-and-model-files-versioning
       //  ->  /doc/use-cases/versioning-data-and-model-files
       res.writeHead(301, {
         Location: req.url.replace(
-          'data-and-model-files-versioning',
-          'versioning-data-and-model-files'
+          '/doc/use-cases/data-and-model-files-versioning',
```
shcheklein: I don't think this prefix is needed. It should be doing exactly the same to my mind.

shcheklein: k, I see this is consistency thingy

jorgeorpinel:
Yay! Merging this will probably create a bunch of merge conflicts in other open PRs though (because of the move from /static/ to /public/static/). Just a heads up guys @ilgooz @pared @iAdramelk @efiop @Suor

Issue #888, opened 2019-12-29
app: use Next.js dynamic routing?


jorgeorpinel:
We currently implement our own custom routing mechanism based on pages/doc.js, public/static/docs/sidebar.json, src/utils/sidebar.js, etc. to find and render markdown files.

We could instead try using the built-in [dynamic routing](https://nextjs.org/docs/routing/dynamic-routes) form Next which would mean having a pages/doc/ directory with an `[item].js` file in it. I think we could also use `await import` instead of `await fetch`.

Idea based on https://dev.to/tinacms/creating-a-markdown-blog-with-next-js-52hk#pages-amp-dynamic-routing

UPDATE: Or migrate to Gatsby like https://github.com/iterative/blog. See https://github.com/iterative/dvc.org/issues/888#issuecomment-571165003

iAdramelk:
@jorgeorpinel I tried dynamic routing then we were migrating to the current system, but wasn't able to make them work with the nested directories. E. g. `/docs/page` is ok, but `/docs/page/page` is not redirected to the `/docs/[item].js`. If you will be able to make it work, then yes, it will be better to migrate.

Same thing with the imports, I tried them, but they didn't work with for me with the `static` dir, so I decided to leave them out for now, but if you can make them work (and maybe they will work if we place `md` files not inside of `static`, but inside of some other folder) it will be better this way.

jorgeorpinel:
Interesting, thanks. I'll research this issue.

jorgeorpinel:
> Working on #738 I stumbled upon some more Next.js info. that brings me back to this issue...

> I tried dynamic routing but /docs/page/page is not redirected to the /docs/[item].js

@iAdramelk We have a custom route in our server.js for this. Can't we use the custom server to enforce dynamic routing of a single level `pages/doc/[item].js` file?

_Note that having this [custom server](https://nextjs.org/docs/advanced-features/custom-server) apparently prevents Next from enabling **Automatic Static Optimization**._

> Same thing with the imports, I tried them, but they didn't work with for me with the static dir...

We would probably need to import a node module that loads the static md files somehow. Maybe with `fs.open` instead of `fetch`, but I'm not sure `fs` is always available for pages/doc.js... Would need to try.

Another option is [dynamic importing](https://nextjs.org/docs/advanced-features/dynamic-import) instead of `await import` or `await fetch`.

jorgeorpinel:
[...]

PR #891, opened 2020-01-04
web: make all indices optional for nav struct (sidebar.json)

### LINKED ISSUE #738: docs: should not need index files in every parent sidebar element
> From https://github.com/iterative/dvc.org/issues/727#issuecomment-545756811 with e.g. /doc/user-guide/contributing
> Also related to #305

Using ` "source": false,` on `sidebar.json` doesn't work unless it is the top level (e.g. `get-started` or `understanding-dvc`). In other levels, this breaks the engine.

Consider also automatically generating a list of H2s in child documents when no `source` field is used for a parent `slug`.

- And finish #727 after this.

jorgeorpinel:
Fix #738

- [ ] Finish #903 after this is merged.


src/utils/sidebar.js
```diff
@@ -22,11 +18,18 @@ const sidebar = require('../../public/static/docs/sidebar.json')
   }
 */
 
+const startCase = require('lodash.startcase')
+
+/* Base sidebar structure */
```
shcheklein: I think comments like this are not informative and don't provide much value

jorgeorpinel: OK. We've discussed comments before and I'm not trying to add complete file documentation. But we have some ad hoc patterns in the docs engine and I comment things that took me some time to figure out. In this case, this signals that the structure in the long file comment above is NOT the same as the one in sidebar.json, but rather that it's just based on the JSON file.

jorgeorpinel: I also separated it from the previous `require` because this one is less of an "import" and more of a constant definition.

shcheklein: I still think that code is pretty much self-descriptive and comments like this do not provide much value )

jorgeorpinel: No worries, I will strip down as many comments as I can when this PR is no longer draft.

jorgeorpinel:
So this is now working. The change was rather small but hard to find due to the complexity of sidebar.js and + me learning Next.js + my inability to properly debug this for a while (figured it out now). Still draft because I need to clean up a bunch of the comments I added, and maybe migrate some to other issues.

src/utils/sidebar.js
```diff
@@ -45,36 +58,16 @@ function findItem(data, targetPath) {
...
-    const prevItem = findItem(data, item.prev)
+function findPrevItemWithSource(data, ref) {
+  if (ref && ref.source) {
+    return ref
+  } else if (ref && ref.prev) {
+    const prevItem = findItem(data, ref.prev)
 
     return findPrevItemWithSource(data, prevItem)
   }
 }
```
jorgeorpinel: Other than deleting `public/static/docs/user-guide/contributing/index.md` and updating `sidebar.json`, This is the only actual change that fixes the issue in question @shcheklein. FYI

jorgeorpinel:
p.s. You can see it working in https://dvc-landing-engine-opti-dkvual.herokuapp.com/doc/user-guide/contributing

p.p.s I guess I could also include **docs: remove unnecessary index pages that serve as a table of contents** #727 in this PR.

src/utils/sidebar.js
jorgeorpinel: It's not even Next.js specific BTW, just some logic that node runs at load time. (One of the custom patterns I was mentioning.)

src/utils/sidebar.js
```diff
@@ -142,21 +135,36 @@ function normalizeSidebar({
...
   const normalizedPath = path.replace(/\/$/, '')
   const isRoot = normalizedPath === PATH_ROOT.slice(0, -1)
   const item = isRoot
     ? normalizedSidebar[0]
     : findItem(normalizedSidebar, normalizedPath)
 
-  return item && findChildWithSource(item)
+  if (!item) return false
+
+  // TODO: Refactor this recursive fn into a loop inside `getItemByPath`
```
shcheklein: no TODOs in the code, if you feel this should be fixed create an issue/ticket with a context, explanation, priority, etc.

src/utils/sidebar.js
```diff
@@ -142,21 +135,36 @@ function normalizeSidebar({
...
+// Runs at module load time
 const normalizedSidebar = normalizeSidebar({
   data: sidebar,
   parentPath: ''
 })
 
-// Exports
-
+/**
+ * Finds `path` in sidebar struct
```
shcheklein: again, not sure, how valuable this is ... this comment does not give more information to my mind and function itself is more or less self-descritptive

src/utils/sidebar.js
shcheklein: not sure about `custom patterns` ... but yes, it's not Next-specific 👍

src/utils/sidebar.js
```diff
@@ -45,36 +58,16 @@ function findItem(data, targetPath) {
...
-function findChildWithSource(item) {
-  return item.source ? item : findChildWithSource(item.children[0])
-}
-
-function findPrevItemWithSource(data, item) {
-  if (item.source) {
-    return item
-  } else if (item.prev) {
-    const prevItem = findItem(data, item.prev)
+function findPrevItemWithSource(data, ref) {
```
shcheklein: why this rename - item to ref? is it consistent with other code around?

src/utils/sidebar.js
jorgeorpinel: Sure! Don't worry, I know. See my previous https://github.com/iterative/dvc.org/pull/891#issuecomment-570833631:

> Still draft because I need to clean up a bunch of the comments I added, and maybe migrate some to other issues.

src/utils/sidebar.js
jorgeorpinel: Hey, sorry. This is still draft. Honestly I didn't expect you to review this until it became a real PR. Please wait, I will request your review soon 🙂

jorgeorpinel:
> should we add a sidebar helper test for this? (also, check that we run them on CI)

Good idea. And I don't think they run on CI.

I also noticed Restyled isn't working on this PR :/ I hope it's because it's a draft but I have a feeling their system is failing again or something. I have the impression that it's unreliable...

shcheklein:
@jorgeorpinel I would not include all index files refactoring as part of this PR - let's try to keep a single change per PR, otherwise it's hard to review and take longer time to release.

src/utils/sidebar.js
jorgeorpinel: The name `item` is used in other parts of this module for a different data structure. It was confusing when debugging, as these other areas of code call this function. When the argument is used, it's called `prevRef` in the caller, but the name of the function already has `Prev`, so I thought just `ref` would be appropriate here.

shcheklein: could you give a bit more details? what is the difference among structures?

src/utils/sidebar.js
jorgeorpinel: I mean that we implement our own "static content optimization" by using a custom server.js that imports a "helper" (sidebar.js) that puts the nav sidebar structure in memory **at server load time**. I call this a custom pattern because Next has a feature called [Automatic Static Optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization) for similar purposes, but we can't use it because we need a custom server script for all the redirects and for the custom route to our docs engine (page). Related to #888

src/utils/sidebar.js
jorgeorpinel: Sure. basically `item` from the previous scope is a chunk of the sidebar.json struct, while `item` (now `ref` in this fn is a fully build "target structure" as spec'd on the top of the file.

<details>

`normalizeItem` calls this function sending `prevRef` to the argument in question, which is a previously "normalized" sidebar item:

https://github.com/iterative/dvc.org/blob/0b607e942be52160bb3429316663f360acec1be7/src/utils/sidebar.js#L78-L85

`normalizeItem` itself receives an `item` argument as well as `prevRef`. This scope's `item` comes from looping `data` in `normalizeSidebar`:

https://github.com/iterative/dvc.org/blob/0b607e942be52160bb3429316663f360acec1be7/src/utils/sidebar.js#L112-L120

which in turn is the original "raw" sidebar.json struct loaded when the server starts:

https://github.com/iterative/dvc.org/blob/0b607e942be52160bb3429316663f360acec1be7/src/utils/sidebar.js#L145-L148

</details>

Maybe `ref` isn't a great name either. In fact maybe this is the true `item` and some other `items` should be called `rawItem` instead. I can make that change, or I can just revert this.

shcheklein: right. thanks! I think it's valuable indeed to distinct item and rawItem by name. It should be consistent around the file, variable names and function names. (like `validateRawItem` vs `findItem` for example)

jorgeorpinel: Done.

Let me clean up all the comments and work on the sidebar test. Then I'll make this a normal PR and ping you. Thanks!

jorgeorpinel:
> let's try to keep a single change per PR 

So should we also separate the task about adding a test for this feature? I'm actually not familiar with how sidebar.test.js works. I'm looking into it now, but feel fee to review this as-is now @shcheklein.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "dev": "node server.js",
-    "dev:debug": "node --inspect server.js",
+    "dev:debug": "node --inspect-brk server.js",
```
jorgeorpinel: Should probably just call this `debug` and maybe mention it in the README or docs contrib guide...

jorgeorpinel:
> And I don't think they run on CI now.

p.s. never mind, they are run on CI with a `test` node script.

package.json
jorgeorpinel: Name updated. And it's already mentioned in https://dvc.org/doc/user-guide/contributing/docs#debugging.

shcheklein:
> So should we also separate the task about adding a test for this feature?

no, usually tests come at the same moment as code

jorgeorpinel:
> add a sidebar helper test for this

@shcheklein turns out we already kind of had a test for this, now called `Returns first child with source for all parents with source:false` which I just had to update to include a 3rd level in the mock nav structure, and so it tests that all 3 levels return the first child with source. See a09dcc17.

BTW maybe other tests are also outdated with not enough levels in their mock nav structures now that we support up to 3? Would need a separate issue to check the validity of the tests.

src/utils/sidebar.js
```diff
@@ -45,42 +56,22 @@ function findItem(data, targetPath) {
   }
 }
 
-function findChildWithSource(item) {
-  return item.source ? item : findChildWithSource(item.children[0])
-}
-
 function findPrevItemWithSource(data, item) {
-  if (item.source) {
+  if (item && item.source) {
```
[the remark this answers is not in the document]
jorgeorpinel: When `source:false` in nested items, it hits this function which used to crash when trying to access `item.source`

src/utils/sidebar.js
```diff
@@ -111,9 +102,9 @@ function normalizeSidebar({
 
   data.forEach(rawItem => {
     const isShortcut = typeof rawItem === 'string'
-    const item = isShortcut ? { slug: rawItem } : rawItem
+    rawItem = isShortcut ? { slug: rawItem } : rawItem
```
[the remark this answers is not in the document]
jorgeorpinel: Agree in general but this is just an anonymous arrow fn for the `forEach` block. Should I revert anyway?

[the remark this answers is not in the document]
jorgeorpinel: p.s. similar to how it's done here by necessity: 
https://github.com/iterative/dvc.org/blob/dd6723cc01b7fb88de8a31bd3b0b55c3067d4a76/src/Documentation/RightPanel/RightPanel.js#L19-L22
https://github.com/iterative/dvc.org/blob/dd6723cc01b7fb88de8a31bd3b0b55c3067d4a76/src/Documentation/RightPanel/RightPanel.js#L34-L37
https://github.com/iterative/dvc.org/blob/dd6723cc01b7fb88de8a31bd3b0b55c3067d4a76/src/styles.js#L54-L58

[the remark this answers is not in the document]
jorgeorpinel: I did find a place where it's probably not justified though:
https://github.com/iterative/dvc.org/blob/dd6723cc01b7fb88de8a31bd3b0b55c3067d4a76/src/Documentation/Markdown/Markdown.js#L158-L164

Which ones should we rewrite?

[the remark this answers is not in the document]
iAdramelk: I'd say we should try to write immutable code everywhere we can by default and only mutate things if there is no other choice.

I'm ok with the images example because we didn't mutate images themselves and just call methods on them.

I'm more or less ok with accumulator example because we only use this arg in reduce and exactly to do this things.

But I'd rewrite `isInsideCodeBlock` code if there is a time for that later.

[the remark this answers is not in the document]
jorgeorpinel: Agreed. OK, let me try to rewrite the other 2 examples and then I'll merge this 🙂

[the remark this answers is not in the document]
jorgeorpinel: Done!
See https://github.com/iterative/dvc.org/pull/891/files?file-filters%5B%5D=.js#diff-b1d807de8934c24309015d3c1ea18882
and https://github.com/iterative/dvc.org/pull/891/files?file-filters%5B%5D=.js#diff-65c1b1a621d382085d912a379ad71c9b
.

src/Documentation/Markdown/Markdown.js
```diff
@@ -156,9 +156,9 @@ export default class Markdown extends React.PureComponent {
...
   isInsideCodeBlock = elem => {
-    for (; elem && elem !== document; elem = elem.parentNode) {
-      if (elem.tagName === 'PRE') return true
-      if (elem.tagName === 'ARTICLE') return false
+    for (let el = elem; el && el !== document; el = el.parentNode) {
+      if (el.tagName === 'PRE') return true
+      if (el.tagName === 'ARTICLE') return false
     }
     return false
   }
```
jorgeorpinel: 1

src/styles.js
```diff
@@ -51,13 +51,14 @@ export const sizes = {
...
-  return accumulator
+export const media = Object.keys(sizes).reduce((acc, s) => {
+  return Object.assign(acc, {
+    [s]: (...args) => css`
+      @media (max-width: ${sizes[s]}px) {
+        ${css(...args)};
+      }
+    `
+  })
 }, {})
```
jorgeorpinel: 2

src/Documentation/Markdown/Markdown.js
jorgeorpinel: Not sure what this does but all Mardown stuff looks good (also with mobile touch screen).

src/styles.js
jorgeorpinel: Yep, all the media width breaks seem the same as in production. BTW one of them seems to narrow to me... Opened #905.

PR #896, opened 2020-01-06
docs: regular updates (early Jan 2020)

jorgeorpinel:


public/static/docs/command-reference/config.md
```diff
@@ -119,7 +119,7 @@ for more details.) This section contains the following options:
   this config option on forces you to run `dvc unprotect` before updating a
   file, providing an additional layer of security to your data.
 
-  It's highly recommended to enable this mod when `cache.type` is set to
+  It's highly recommendable to enable this option when `cache.type` is set to
```
shcheklein: recommended vs recommendable - 900K vs 12K in google.

jorgeorpinel: "It's recommended" by who? 🙂 
I think we should either say "It's recommendable" or "We recommend".

src/Documentation/Tutorials.js
```diff
@@ -10,8 +10,8 @@ export default function Tutorials({ compact, tutorials }) {
 
   return (
     <Wrapper compact={compact}>
-      {tutorialsData.map(([key, value]) => {
-        const ButtonComponent = icons[key] || ExternalButton
+      {tutorialsData.map(([k, value]) => {
```
shcheklein: why do we change names aggressively? human readable names (like key) are good enough when they stay short and to the point :)

public/static/docs/command-reference/config.md
shcheklein: I thinks it's just a common phrase more or less - https://www.google.com/search?q=%22It%27s+highly+recommended+to+use%22&oq=%22It%27s+highly+recommended+to+use%22&aqs=chrome..69i57.4247j0j7&sourceid=chrome&ie=UTF-8 . I does feel natural even w/o specifying "by who" part.

On the other hand - `It's recommendable` is extremely rare.

I'm not agains `We recommend` if we can get a prove that `It's recommended` is wrong.

jorgeorpinel: 900k results is not rare :P but fine, no reason to change something if it's not wrong. It's just not literally correct (again, recommended by who?) but if it's a common phrase then no problem. Will revert.

jorgeorpinel: I mean 12k

src/Documentation/Tutorials.js
jorgeorpinel: Oh, this one is because the JSX inside also has a `key` prop, which is assigned to `{value}` so I thought that was confusing and renamed the key argument or Array.map. I thought single-character variables for things like loops and array keys were a common (ok) practice but I can change it to something or revert this if you prefer, please lmk. Thanks

public/static/docs/install/index.md
```diff
@@ -7,6 +7,10 @@ Please double check that you don't already have DVC (for example running
 - [Install on Windows](/doc/install/windows)
 - [Install on Linux](/doc/install/linux)
 
+⚠️ Please try to use the same/latest DVC version on all the team member
+machines. DVC is in active development and compatibility issues may arise
+between significantly different versions.
```
jorgeorpinel: What do you think about adding this note @shcheklein @efiop? Per https://groups.google.com/a/iterative.ai/forum/#!msg/support/ZBNM2dydyxU

shcheklein: It might be good, but it probably take Ruslan's exact language about backward and forward compatibility and also let's not scary users with `is in active development` phrase :) First of all, only projects are of not any interest are not being actively developed and any mature project keeps in mind compatibility policies. Second, our version number is already enough to scary users :)

Also, it can be a larger topic a bit - we can clarify the policy around the versioning - major, minor, etc. It's important to understand those.

public/static/docs/command-reference/config.md
jorgeorpinel: OK I reviewed the usage of this term throughout all docs now. Feel fee to review again. Thanks

src/Documentation/Tutorials.js
jorgeorpinel: I thought single-character variables for things like loops and array keys were a common (ok) practice but I can change it to something else or revert this if you prefer, please lmk. Thanks

shcheklein: Sure! np. was just curious ... it's totally fine it the name overlaps and might be confusing 👍

public/static/docs/install/index.md
shcheklein: let's remove this and create a separate ticket to document this properly?

jorgeorpinel: OK, removed. Created 906

PR #902, opened 2020-01-06
remote: use `-d` in s3 example

efiop:
User complained that local one has `-d` but s3 doesn't.

### DIFF
--- a/public/static/docs/command-reference/remote/index.md
+++ b/public/static/docs/command-reference/remote/index.md
@@ -101,7 +101,7 @@ remote = myremote
 > [Create a Bucket](https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html).
 
 ```dvc
-$ dvc remote add mynewremote s3://mybucket/myproject
+$ dvc remote add -d mynewremote s3://mybucket/myproject
 $ dvc remote modify mynewremote region us-east-2
 ```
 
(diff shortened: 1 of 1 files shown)


public/static/docs/command-reference/remote/index.md
```diff
@@ -101,7 +101,7 @@ remote = myremote
 > [Create a Bucket](https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html).
 
 ```dvc
-$ dvc remote add mynewremote s3://mybucket/myproject
+$ dvc remote add -d mynewremote s3://mybucket/myproject
```
shcheklein: let's change title above and mention explicitly at least that we are adding a default remote?

efiop: Good point, haven't noticed that. Adjusted. Thanks!

jorgeorpinel:
Going back to this. I'm not sure why we wanted to change this example over a single user's feedback. I think having significantly different examples is a good thing. Maybe we should revert?

jorgeorpinel:
> Maybe we should revert?

I'm improving the remote index examples (including reverting this `-d`) in 00f47519. Feel free to review (in #915).

efiop:
@jorgeorpinel Not worth reverting in my opinion. This PR just made it consistent within one doc. I think that complaint was a reasonable one.

jorgeorpinel:
Understood, but I've already rewritten the examples of that doc anyway, please see https://github.com/iterative/dvc.org/pull/915#pullrequestreview-341595711. Thanks!

Issue #550, opened 2019-08-11
guide: new DVC Concepts section


jorgeorpinel:
Per https://github.com/iterative/dvc.org/pull/431#pullrequestreview-253571711

There should be a section or a few sections called **Basic Concepts** in the user guide that is/are properly written on a user-level language explaining some concepts. The glossary tooltips could point to these pages too.

> we need to explain concepts like Cache, Remote, Wokspace, etc, and their relationships (with diagrams if needed).
From https://github.com/iterative/dvc.org/issues/550#issuecomment-544162594 below

~~Figure out a connection to glossary tooltips (currently special md files with Frontmatter in `content/docs/user-guide/basic-concepts/` — see #1395): Just link to the Concepts? Use a similar engine (with Frontmatter)? Merge completely?~~

UPDATE: Skip to https://github.com/iterative/dvc.org/issues/550#issuecomment-673585050 for summary

### UPDATE: Specific plan

Use contents from #1655 as needed.

- Do we need the changes in package.json, src/components/Documentation/Markdown/Tooltip/index.tsx, src/gatsby/models/glossary/index.js, and src/utils/front/glossary.ts from iterative/dvc.org/pull/1944 ?
- [x] _Workspace_ (#2197)
- [x] _DVC Project_ (vs. Git/DVC Repository) [#2754]
- [ ] (from cmd ref) (DVC) Metafiles
- [ ] (from cmd ref) DVC/Project Cache
- [ ] (from cmd ref) Data Pipeline(s)
- [ ] (from cmd ref) Metrics & Plots
- [ ] (from cmd ref) Remote (Storage)
- [ ] (from cmd ref) Stage
  & update some of the links around this term that currently go to `run/stage`.
  (Some should link to the future pipelining guide, see #2883)
- [ ] (from cmd ref) (Stage) Output
- [ ] (from cmd ref) (Stage) Dependency
- [ ] Run-Cache?
- [ ] Garbage Collection clarifications (in cache or remote pages, see #2169)
- [ ] Workspace clarifications? Discussed in #1127
- Add Further Reading sections to all Basic Concepts documents? (#2127) 
---

- [x] Look into #1395 after or along with this one.
- Consider also a related DVC Principles section (more philosophical, maybe in the Concepts section index)? See https://github.com/iterative/dvc.org/issues/550#issuecomment-544106096 below.
  UPDATE: Maybe use the section index for this or/and **What is DVC?**.

jorgeorpinel:
[...]

PR #903, opened 2020-01-07
cmd ref: remove unnecessary indices?

### LINKED ISSUE #727: docs: remove unnecessary index pages that serve as a table of contents

jorgeorpinel:
~~Blocked by #891~~

Closes #727


jorgeorpinel:
**Blocked by #891**

Closes #727

jorgeorpinel:
(Please do not review yet.)

public/static/docs/tutorials/index.md
```diff
@@ -1,18 +0,0 @@
-# Tutorials
```
shcheklein: tbh: some kind of an index is useful in this case (and other cases?) - when give a direct link to tutorial it feels strange that it opens the same page as the first section? what do you think, @jorgeorpinel ?

jorgeorpinel: This is the only index I could find that definitely seemed unnecessary to me at this point, actually. Please see discussion in https://github.com/iterative/dvc.org/issues/727#issuecomment-571407120 (and the next comment too).

jorgeorpinel: To summarize, I agree that

> it feels strange that it opens the same page as the first section

But I also don't want to maintain these bullet lists.

Maybe just strip down these indices to only a single paragraph and a sentence like "Please chose a tutorial in the left navigation side bar or click Next to jump into the first category"?

shcheklein: Yep, we can strip the description to the essence of it. A paragraph or two. It does not hurt to mention that there are different kinds of tutorials though. It def tedious to maintain an exact description.

jorgeorpinel: OK I've stripped down the indices for tutorials (restored), use-cases, and user-guide 👯

So... #891 was only really for the contribution guide 😋

jorgeorpinel:
> So... #891 was only really for the contribution guide 😋 (from https://github.com/iterative/dvc.org/pull/903#discussion_r364350355)

Actually, on this @shcheklein what about my https://github.com/iterative/dvc.org/issues/727#issuecomment-571407120 about cmd ref indices?

> - https://dvc.org/doc/command-reference/cache,
> - https://dvc.org/doc/command-reference/metrics,
> - https://dvc.org/doc/command-reference/pipeline, and
> - https://dvc.org/doc/command-reference/remote
>
> have useful, unique content but I'm not sure the cmd ref is the best place for those explanations. Should we take the opportunity to move them into a new **user-guide: "Basic Concepts"** (#550)?

jorgeorpinel:
This would mean that the first subcommand opens by default. For cache -> cache/dir (good), for metrics and remote -> */add, and for pipeline -> pipeline/list

shcheklein:
@jorgeorpinel hmm ... I think they are valuable. Not matter what, but `dvc list -h` is a legit command by itself and should be described. Or in the future we can decide to make `dvc remote` an alias of `dvc remote list`?

public/static/docs/tutorials/index.md
shcheklein: Cool! 😎

jorgeorpinel:
Yes, they have valuable content explaining what the concepts are (cache, metrics, pipeline, remote). My point is that such content probably doesn't belong in the command reference and we can move it to a new **user-guide: "Basic Concepts"** (per #550) – and once that is done, it won't make sense to keep these indices.

Should they just be a short list of links to their subcommands? Or remove completely and let the first one load. I think at least for `dvc cache` since it only has `dir` inside, we should just remove the index.

`dvc list`? That's not released yet. Nor related to the commands I mentioned above ^ :suspect:

shcheklein:
@jorgeorpinel `dvc remote`, `dvc cache`, etc are valid commands. So I think they belong to the command reference and should have more or less the same structure.

shcheklein:
> dvc list? That's not released yet. Nor related to the commands I mentioned above ^ :suspect:

:) sorry, that was a mistake - I've updated the comment

jorgeorpinel:
> dvc remote, dvc cache, etc are valid commands

Ah yes good point. Didn't think about that. So we definitely need the indices. OK! This PR is now officially reviewable!

jorgeorpinel:
p.s. but please note that I've extracted the discussion about moving concepts explanations out of the cmd ref to https://github.com/iterative/dvc.org/issues/550#issuecomment-572223386.

PR #914, opened 2020-01-11
gdrive: Shared drives support

MaxRis:
Related to https://github.com/iterative/dvc/pull/3016

### DIFF
--- a/public/static/docs/command-reference/remote/add.md
+++ b/public/static/docs/command-reference/remote/add.md
@@ -226,6 +226,24 @@ On successful access token generation, token data will be cached in git ignored
 directory with path `.dvc/tmp/gdrive-user-credentials.json`. Do not share token
 data with anyone else to prevent unauthorized access to your Google Drive.
 
+#### Shared drives support
+
+You need to obtain `id` of your shared drive directory to use it as part of URL
+passed to DVC. `id` can be found in your web browser address bar when shared
+drive is opened. For example, for the URL
+
+```dvc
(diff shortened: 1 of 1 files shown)


public/static/docs/command-reference/remote/add.md
```diff
@@ -226,6 +226,24 @@ On successful access token generation, token data will be cached in git ignored
 directory with path `.dvc/tmp/gdrive-user-credentials.json`. Do not share token
 data with anyone else to prevent unauthorized access to your Google Drive.
 
+#### Shared drives support
```
jorgeorpinel: Not sure about adding subheaders in hidden sections. We don't even have H4s anywyere yet, this would be the first one in all the docs actually (except for 2 wrong ones I just noticed in `dvc version` which I'll fix soon). The anchor icon is not displayed properly:
image
and if you try to use the direct link (https://dvc.org/doc/command-reference/remote/add#shared-drives-support) the page won't open the section and scroll to that header.

jorgeorpinel: Hm I just found another one in `dvc remote add` actually, inside the **Aliyun OSS** section.

Anyway, since this is merged I'll just come up with a better solution in a near-future "regular updates" PR.

shcheklein: @jorgeorpinel yep, we can change it to a regular **bold**

jorgeorpinel: OK addressed. Feel free to review: 26d9928b Thanks

PR #915, opened 2020-01-12
doc: regular updates (mid Jan)

### LINKED ISSUE #698: user-guide: improve Windows POSIX/terminal tips

jorgeorpinel:
Misc. stuff +
- Closes #698 (improve Windows terminal tips)
- Addresses https://github.com/iterative/dvc.org/issues/757#issuecomment-574317692 (303 server redirects)
- Addresses #789


public/static/docs/command-reference/remote/add.md
```diff
@@ -217,28 +217,22 @@ $ dvc remote modify myremote gdrive_client_id my_gdrive_client_id
...
+associated with the Google Drive you want to use as remote. The login process
+will guide you through the granting of the required access permissions.
 
-On successful access token generation, token data will be cached in git ignored
-directory with path `.dvc/tmp/gdrive-user-credentials.json`. Do not share token
-data with anyone else to prevent unauthorized access to your Google Drive.
+On successful access token generation, the token data will be cached in a
+Git-ignored directory (located in `.dvc/tmp/gdrive-user-credentials.json`).
 
-#### Shared drives support
```
shcheklein: I think it makes to have some kind of title, even if it's a regular bold paragraph

public/static/docs/command-reference/remote/add.md
```diff
@@ -287,19 +281,13 @@ $ export OSS_ACCESS_KEY_SECRET="my-key-secret"
 $ export OSS_ENDPOINT="endpoint"
 ```
 
-#### Test your OSS storage using docker
```
shcheklein: same here, it's better to have some structure

public/static/docs/command-reference/remote/index.md
```diff
@@ -95,17 +95,37 @@ url = /path/to/remote
...
 
-## Example: Add a default Amazon S3 remote and modify its region
+## Example: Add an Amazon S3 remote and modify its region
 
 > 💡 Before adding an S3 remote, be sure to
 > [Create a Bucket](https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html).
 
 ```dvc
-$ dvc remote add -d mynewremote s3://mybucket/myproject
+$ dvc remote add mynewremote s3://mybucket/myproject
```
jorgeorpinel: @ivan thinks that

> no need to revert that -d for now

but I think given the other improvements to this doc it's best to remove it from here. Basically each example shows the most typical uses of all the subcommands 🙂

jorgeorpinel: Plus it seems @efiop is thinking about changing the behavior of `dvc remote add` so the first remote is set as `--default` by default, at which point we'll have to revisit these examples again anyway.

public/static/docs/command-reference/remote/add.md
jorgeorpinel: OK, reinstated in 371400b5.

public/static/docs/command-reference/remote/add.md
jorgeorpinel: Reinstated as well.

public/static/docs/command-reference/remote/index.md
jorgeorpinel: Hmmm. Ruslan also likes having `-d` here "for consistency" (see https://github.com/iterative/dvc.org/pull/902#issuecomment-573534572). I'm not sure about this consistency argument, will have to review the other remote subcommand docs and check where `-d` is used in examples and where it isn't. But it's not clear to me what the purpose of having `-d` consistently everywhere is. Again, I think having variety in the examples is more valuable 🙂

shcheklein: Probably the argument is that in 95% of the cases people setting up it for the first time and in 99% of those cases they need `-d`. So, we are optimizing the first time user experience with this.

jorgeorpinel: OK, it's no problem for me to make sure all (or most) of the `remote add` examples have `-d`.

But my counter-argument is that the goal of having multiple examples is precisely to not just address the use case of 95% (presumably) of users but to have a variety of interesting combinations (to show how flexible the tool is). The first example would usually be the simplest one, geared for the 99%, and they get more interesting from there.

server.js
```diff
@@ -42,11 +42,7 @@ app.prepare().then(() => {
       res.writeHead(301, {
         'Cache-Control': 'no-cache',
         Location:
-          'https://dvc.org/doc/command-reference' +
-          // replace - for / in {cmd} except for get-url, import-url
-          (['/get-url', '/import-url'].indexOf(pathname) < 0
-            ? pathname.replace('-', '/')
-            : pathname)
+          'https://dvc.org/doc/command-reference' + pathname.replace('-', '/')
```
jorgeorpinel: > so, no need to replace anything at all.
From https://github.com/iterative/dvc.org/commit/44fe4263e4f9eb0568118fd82d584e1604b5fad0#r36784752

🤦‍♂ Good point @shcheklein! Fixed.

jorgeorpinel:
OK I guess this is starting to grow too big. Let's officially review it?

public/static/docs/command-reference/remote/index.md
jorgeorpinel: OK update on this! I reviewed all the 54  instances of `remote add` in the docs (some are just links) and added `-d` into some of them  (the ones with a context where it seems like this would be the first remote added to that example project).

**Note that this includes in the home page `/`:**

image

Please note that most of the instances don't have `-d` because it wouldn't make sense (like in the list of remote types of the `dvc remote add/modify` cmd refs, et al.) In the future, if setting the first remote as default is the default behavior or `dvc add remote`, we should probably remove all of these `-d`, except if talking specifically about setting an extra remote as default.

Please see 4950d252.

public/static/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -10,10 +10,17 @@ workarounds that can help avoid these potential problems:
...
 
-We recommend [Git Bash](https://gitforwindows.org/#bash) or
-[Anaconda Prompt](https://docs.anaconda.com/anaconda/user-guide/getting-started/#open-prompt-win).
-Its even possible to enjoy a full Linux terminal experience with the
+Common Windows terminal alternatives are
+[Git Bash](https://gitforwindows.org/#bash) or
+[Anaconda Prompt](https://docs.anaconda.com/anaconda/user-guide/getting-started/#open-prompt-win)
+– but they may not support all the POSIX features (e.g. `\` line continuation).
+
+**We recommend the full [Cmder](https://cmder.net/) console emulator** (which
```
shcheklein: probably no need to emphasize it? or put an emoji?

public/static/docs/command-reference/remote/modify.md
```diff
@@ -382,26 +382,28 @@ For more information on configuring Azure Storage connection strings, visit
...
-Modify its endpoint URL:
+Modify its access profile:
 
 ```dvc
-$ dvc remote modify myremote endpointurl https://object-storage.example.com
+$ dvc remote modify myremote profile myusername
 ```
 
-Now the config file should look like (run `cat .dvc/config`):
+Now the project config file should look like this:
```
shcheklein: I would keep the place where it is located

pages/doc.js
```diff
@@ -1,23 +1,22 @@
...
 import { HeadInjector } from '../src/Documentation/HeadInjector'
 import Hamburger from '../src/Hamburger'
 import SearchForm from '../src/SearchForm'
 import SidebarMenu from '../src/Documentation/SidebarMenu/SidebarMenu'
 import Markdown from '../src/Documentation/Markdown/Markdown'
 import RightPanel from '../src/Documentation/RightPanel/RightPanel'
-// sidebar data and helpers
-import { getItemByPath, structure } from '../src/utils/sidebar'
-// styles
 import { media } from '../src/styles'
```
jorgeorpinel: @iAdramelk and Ivan: Here's the sample import sort order I came up with, as explained in https://github.com/iterative/dvc.org/issues/789#issuecomment-575267234

jorgeorpinel: There's an external import group, sorted by first-used. Then there's an internal import group (only has one line, internally sorted by first-used), and an internal component import group which includes the styled components import – also sorted internally by first-used. Notice that `Error` is a component but since it's an external import it's in the first group...

jorgeorpinel: Anyway, that issue is now closes so I guess this isn't very important anymore but just wanted to follow up with you guys in case there's any further comments or ideas. Thanks

public/static/docs/command-reference/config.md
```diff
@@ -110,8 +110,8 @@ for more details.) This section contains the following options:
   > directory into paths relative to the config file location.
 
 - `cache.protected` - make files under DVC control read-only. Possible values
-  are `true` or `false` (default). Run `dvc checkout` for the change to go into
-  effect.
+  are `true` or `false` (default). Run `dvc checkout` after setting changing the
```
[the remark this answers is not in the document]
jorgeorpinel: changing. Fixing

public/static/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: Good idea 💡 <-- using emoji

public/static/docs/command-reference/remote/modify.md
jorgeorpinel: It's not always located there and this example doesn't start by initializing the project so it's not guaranteed to be in that location. There's several other examples of config file in other remote commands and none mention the config file location.

public/static/docs/command-reference/import.md
```diff
@@ -75,13 +75,16 @@ data artifact from the source project.
...
-  import the data from. The tip of the repository's default branch is used by
-  default when this option is not specified. Note that this adds a `rev` field
-  in the import stage that fixes it to this revision. This can impact the
-  behavior of `dvc update`. (See **re-importing** example below.)
+  (such as a branch name, a tag, or a commit hash) of the repository to download
+  the file or directory from. The tip of the default branch is used by default
+  when this option is not specified.
+
+  > Note that this adds a `rev` field in the import stage that fixes it to this
+  > revision. This can impact the behavior of `dvc update`. (See
```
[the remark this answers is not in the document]
jorgeorpinel: Can. Depends on the type of revision. Keep in mind this version doesn't reflect the new --rev option in `dvc update`. That update is still in process: iterative/dvc.org/pull/890 - we'll have to review all this then anyway...

pages/doc.js
shcheklein: I'm fine with this sorting. But let's wait for @iAdramelk and may be it's not a good time to make cosmetic changes to JS when we are transferring stuff - it will create conflicts and complicate merging.

src/Documentation/Markdown/Markdown.js
```diff
@@ -2,13 +2,13 @@ import React from 'react'
...
 import NextLink from 'next/link'
 import Router from 'next/router'
-// components
+
 import ReactMarkdown from 'react-markdown'
 import { LightButton } from '../LightButton'
 import Tutorials from '../Tutorials'
 import Tooltip from '../../Tooltip'
 import Collapsible from 'react-collapsible'
-// syntax highlighter
```
shcheklein: same here - let's not do this right now. Let's postpone cosmetic JS changes please and wait for @iAdramelk to wrap up the two major branches.

server.js
```diff
@@ -70,19 +64,19 @@ app.prepare().then(() => {
...
       res.end()
     } else if (/^\/(help|chat)\/?$/i.test(pathname)) {
       // path /(help|chat) -> Discord chat invite
-      res.writeHead(301, { Location: 'https://discordapp.com/invite/dvwXA2N' })
+      res.writeHead(303, { Location: 'https://discordapp.com/invite/dvwXA2N' })
       res.end()
     } else if (/^\/(docs|documentation)(\/.*)?$/i.test(pathname)) {
       // path /docs... or /documentation... -> /doc...
-      res.writeHead(301, {
+      res.writeHead(303, {
```
shcheklein: should be 301 here?

pages/doc.js
jorgeorpinel: OK. I only changed this file as an example. Not planning to change any other ones since it's a manual task and there's like 40 files. This shouldn't create a conflict unless the transfer branch also modifies the imports in this file (unlikely).

src/Documentation/Markdown/Markdown.js
jorgeorpinel: You mean not to remove those unnecessary comments? I thought we agreed to remove them. Should I bring them back? Up to you, but I think a simple removal like this is very unlikely to produce merge conflicts.

shcheklein: yes, I mean not do a lot of cosmetic changes to JS at this moment. We can do it later. It might create conflicts for @iAdramelk for no good reason.

server.js
jorgeorpinel: Hmmm probably not? docs.and documentation. subdomains are "structural" redirects, not previously existing pages we would need to transfer "link juice" (SEO stuff) over to a new address. I think this is the criteria we came up with to decide which response status code to use. I may be confused so up to you, please lmk.

shcheklein: but they are not subdomains in this case, right?

jorgeorpinel: Ah yeah sorry I meant /docs and /documentation **paths**.

shcheklein: So, the are not subdomains. There are external links out there on other websites like dataversioncontrol.com/documentation/tutorial and we should be notifying Google explicitly that we moved it to the new domain/path.

src/Documentation/Markdown/Markdown.js
jorgeorpinel: OK... Reverted https://github.com/iterative/dvc.org/compare/4950d25..7ea2edb. This is the 2nd time these changes get reverted so I'll just forget about this until/if someone else brings it up 😋

pages/doc.js
jorgeorpinel: Also reverted in c64d5906.

server.js
jorgeorpinel: > There are external links out there on other websites like dataversioncontrol.com/documentation/tutorial

So this would be redirected to dvc.org/documentation/tutorial ? I wasn't aware. I guess in that case `301` is more appropriate. Changed in cc376793.

PR #931, opened 2020-01-19
doc: regular updates (late Jan)

### LINKED ISSUE #506: document that hooks don't act in a non-DVC branch
Document that:

* (once https://github.com/iterative/dvc/pull/2338 is merged) the hooks installed by `dvc install` only act in a “dvc-enabled” branch, meaning one that has tracked files under `.dvc`
* initializing dvc on a branch doesn't affect other branches until the change gets merged to those branches. (So you can safely try it out in a branch before merging to master, for instance.)

Prompted by https://github.com/iterative/dvc/issues/2208

### LINKED ISSUE #686: term: word-emphasizing best practice?
Now we mainly use **bold** words in some docs but not very consistently. There may also be some _italics_ and hopefully no underlines but it's possible (need to check).

Which one of these should we adopt? I think probably just one to keep things simple (so I would incline towards **bold**. What kind of words or phrases should it be applied to?

- [x] Decide (see https://github.com/iterative/dvc.org/issues/686#issuecomment-575966629);
- [x] Note to docs contrib guide (See dfcd3bec); ~~and~~
- ~~Start reviewing all existing docs (a little daunting 😓)~~.

### LINKED ISSUE #809: cmd ref: link from remote add to remote modify
[This section](https://dvc.org/doc/command-reference/remote/add#supported-storage-types) in `remote add` docs touches on some `remote modify` functionality, since you need it to set up some of the remotes, but it's incomplete.

I suggest adding something like "See full list of remote options you may modify at `remote modify` page" in the end of the section.

### LINKED ISSUE #841: cmd ref: document dir support for import-url and get-url
After https://github.com/iterative/dvc/pull/2894 we automatically support s3, gs, local, ssh directories in import-url and get-url.

NOTE: just trying to not forget about this, will try to contribute a PR myself if I can.

### LINKED ISSUE #855: docs: add emoji symbols to improve docs
Look at current notes, for example. Useful symbols:

- ⚠️ Warnings about possible problems related to DVC usage (similar to `**Note!**` and "Note that..." notes)
- 💡 Useful tips related to external tools/integrations

Some other emojis currently in use here and there: ⚡🙏🐛⭐❗

UPDATE: I just found also 📙🙂💎🏷📖 used in the change log (which may soon be removed though, see #897).

### LINKED ISSUE #946: tutorials: replace `dvc pull` example in deep tut
The [Pulling data from the cloud](https://dvc.org/doc/tutorials/deep/sharing-data#pulling-data-from-the-cloud) deep tutorial code example doesn't work because https://github.com/dmpetrov/new_tag_classifier had been deleted. It's been restored now but the S3 remote is also gone so `dvc pull` fails.

image

We should probably just use a completely different project as example here (and not directly in Dmitry's account) instead of restoring the bucket, as the example is not really tied to the larger tutorial anyway?

> `cd` command is also missing here, but being addressed in #931.

jorgeorpinel:
- Re-closes #855
- Closes #809
- Closes #686
- Closes #841
- Closes #506
- Other misc. stuff, as usual
- Closes #946


public/static/docs/command-reference/import-url.md
```diff
@@ -48,15 +47,20 @@ determine whether the local copy is out of date.
...
-| `remote` | Remote path (see explanation below)                     | `remote://myremote/path/to/file`           |
+| Type     | Description                                           | `url` format                               |
+| -------- | ----------------------------------------------------- | ------------------------------------------ |
+| `local`  | Local path                                            | `/path/to/local/data`                      |
+| `s3`     | Amazon S3                                             | `s3://mybucket/data`                       |
+| `gs`     | Google Storage                                        | `gs://mybucket/data`                       |
+| `ssh`    | SSH server                                            | `ssh://user@example.com:/path/to/data`     |
+| `hdfs`   | HDFS to file\*                                        | `hdfs://user@example.com/path/to/data.csv` |
+| `http`   | HTTP to file with _strong ETag_ (explanation below)\* | `https://example.com/path/to/data.csv`     |
+| `remote` | Remote path (see explanation below)                   | `remote://myremote/path/to/data`           |
```
jorgeorpinel: Quick question @efiop. Only `import-url` supports remote:// URL notation? Asking because the `get-url` cmd ref doesn't have this line in the same table.

public/static/docs/command-reference/get.md
```diff
@@ -12,8 +12,8 @@ into the current working directory.
 usage: dvc get [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
 
 positional arguments:
-  url         Location of DVC project or Git repository to download from
-  path        Path to a file or directory within the project or repository
+  url         URL of Git repository with DVC project to download from.
+  path        Path to a file or directory within a DVC repository.
```
jorgeorpinel: Change contingent on iterative/dvc/issues/3196 (same in import.md)

public/static/docs/command-reference/checkout.md
```diff
@@ -203,23 +204,25 @@ once to download missing data from the remote storage to the <abbr>cache</abbr>.
...
 ```dvc
 $ dvc install
 ```
 
 `dvc install` installs Git hooks to automate common operations, including
 running `dvc checkout` when needed.
 
-We can then checkout the master branch again:
+> Note that the hooks only have an effect in repository versions (branches,
+> etc.) where DVC is active.
```
jorgeorpinel: @shcheklein here's that note about the hooks (per https://github.com/iterative/dvc.org/issues/506#issuecomment-576053513).
BTW, I noticed `commit` and `push` commands don't have similar references to `dvc install`, should they? Maybe not a full blown example like in here but at least a note.

jorgeorpinel: I decided to refine the existing notes and include it also in the other commands where it was missing (commit, repro, push).

public/static/docs/command-reference/import-url.md
jorgeorpinel: Seems like will need another update per iterative/dvc/pull/3214

public/static/docs/command-reference/get.md
jorgeorpinel: Seems like will need another update per iterative/dvc/pull/3214.

public/static/docs/tutorials/deep/sharing-data.md
```diff
@@ -51,13 +51,14 @@ the master branch:
 
 ```dvc
 $ git clone https://github.com/dmpetrov/new_tag_classifier.git
+$ cd new_tag_classifier
 $ dvc pull
 ```
```
jorgeorpinel: Here's the fix to https://github.com/iterative/dvc.org/pull/938#issuecomment-576884513 (Dmitry restored the repo) `cd` was also missing.

public/static/docs/command-reference/import-url.md
jorgeorpinel: I guess this is correct. I tried `dvc get-url remote://storage/get-started/data.xml` from inside my local clone of https://github.com/iterative/dataset-registry and got:
`ERROR: unexpected error - 'NoneType' object has no attribute 'repo'`

public/static/docs/command-reference/get.md
jorgeorpinel: OK done in d641dc84.

public/static/docs/command-reference/commit.md
```diff
@@ -21,7 +21,13 @@ The `dvc commit` command is useful for several scenarios where a dataset is
 being changed: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/pipeline) is in development, when one wishes
 to run commands outside the control of DVC, or to force DVC-file updates to save
-time tying stages or a pipeline.
+time tying stages or a pipeline. These scenarios are further detailed below.
+
+💡 For convenience, a Git hook is available to remind you to `dvc commit` when
+needed after a `git commit`. See `dvc install` for more details.
+
+> Note that the hooks only have an effect in repository versions (branches,
```
shcheklein: probably too much to include this note here

public/static/docs/command-reference/checkout.md
```diff
@@ -25,7 +25,12 @@ workspace data with the versions specified in the current DVC-files.
 that changes the DVC-files in the workspace.
 
 💡 For convenience, a Git hook is available to automate running `dvc checkout`
-after `git checkout`. Use `dvc install` to install it.
+after `git checkout`. See the
+[**Automating** example](#example-automating-dvc-checkout) below or
+`dvc install` for more details.
+
+> Note that the hooks only have an effect in repository versions (branches,
```
shcheklein: `Note that hook won't activate if the branch (commit, tag, etc) you checkout does not have DVC initialized (no .dvc directory)`?

public/static/docs/command-reference/fetch.md
```diff
@@ -185,6 +185,13 @@ $ tree .dvc
 ├── ...
 ```
 
+> The cache file with `.dir` extension is a special text file that contains the
```
shcheklein: probably not the purpose of fetch to explain exact cache structure, I keep it simpler

public/static/docs/command-reference/get-url.md
```diff
@@ -36,14 +35,19 @@ single-purpose command that can be used out of the box after installing DVC.
...
-| `hdfs`  | HDFS           | `hdfs://user@example.com/path/to/data.csv` |
-| `http`  | HTTP to file   | `https://example.com/path/to/data.csv`     |
+| `local` | Local path     | `/path/to/local/data`                      |
+| `s3`    | Amazon S3      | `s3://mybucket/data`                       |
+| `gs`    | Google Storage | `gs://mybucket/data`                       |
+| `ssh`   | SSH server     | `ssh://user@example.com:/path/to/data`     |
+| `hdfs`  | HDFS to file\* | `hdfs://user@example.com/path/to/data.csv` |
+| `http`  | HTTP to file\* | `https://example.com/path/to/data.csv`     |
+
+> \* HDFS and HTTP **do not** support downloading entire directories, only
```
shcheklein: did we check re the HDFS one?

public/static/docs/command-reference/get.md
```diff
@@ -12,8 +12,8 @@ into the current working directory.
 usage: dvc get [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
 
 positional arguments:
-  url         Location of DVC project or Git repository to download from
-  path        Path to a file or directory within the project or repository
+  url   Location of DVC project or Git repository to download from.
```
shcheklein: usually more spaces here?

public/static/docs/command-reference/import-url.md
```diff
@@ -48,15 +47,20 @@ determine whether the local copy is out of date.
...
+
+> \* HDFS and HTTP **do not** support downloading entire directories, only
```
shcheklein: same re HDFS

public/static/docs/command-reference/import.md
```diff
@@ -15,8 +15,8 @@ import.
 usage: dvc import [-h] [-q | -v] [-o [OUT]] [--rev [REV]] url path
 
 positional arguments:
-  url         Location of DVC project or Git repository to download from
-  path        Path to a file or directory within the project or repository
+  url   Location of DVC project or Git repository to download from.
```
shcheklein: same as with get, more spaces?

public/static/docs/command-reference/install.md
```diff
@@ -16,6 +16,9 @@ like Git to store code and configuration files. With `dvc install`, the two are
 more tightly integrated in order to cause certain convenient actions to happen
 automatically.
 
+> Note that the hooks only have an effect in repository versions (branches,
```
shcheklein: see the suggestion above, and probably this document is the only place we should keep this note. Also not sure if we need to make it `>` ... may be keep as a regular paragraph?

public/static/docs/command-reference/push.md
```diff
@@ -244,6 +250,11 @@ $ tree ../vault/recursive
 5 directories, 5 files
 ```
 
+> Cache files with `.dir` extension is a special text file that contains the
```
shcheklein: same as above

public/static/docs/command-reference/push.md
```diff
@@ -29,6 +29,12 @@ The `dvc push` command allows one to upload data to remote storage. It doesn't
 save any changes in the code or DVC-files. Those should be saved by using
 `git commit` and `git push`.
 
+💡 For convenience, a Git hook is available to automate running `dvc push` after
+`git push`. See `dvc install` for more details.
+
+> Note that the hooks only have an effect in repository versions (branches,
```
shcheklein: same as above

public/static/docs/command-reference/repro.md
```diff
@@ -28,6 +28,12 @@ then be executed in the correct order, reproducing pipeline results.
 > Pipeline stages are typically defined using the `dvc run` command, while
 > initial data dependencies can be registered by the `dvc add` command.
 
+💡 For convenience, a Git hook is available to remind you to `dvc repro` when
+needed after a `git commit`. See `dvc install` for more details.
+
+> Note that the hooks only have an effect in repository versions (branches,
```
shcheklein: same here

public/static/docs/tutorials/deep/sharing-data.md
shcheklein: switched to example-get-started like was suggested?

public/static/docs/user-guide/dvcignore.md
```diff
@@ -79,10 +80,16 @@ $ tree .dvc/cache
     └── c3d3797971f12c7f5e1d106dd5cee2
 ```
 
-As we can see, `data1` has been ignored. Cache contains only one file entry (for
-`data2`) and one dir entry (`data`).
+Only the checksums of a directory (`data/`) and one files have been
+<abbr>cached</abbr>. This means that `dvc add` ignored one of the files
+(`data1`).
+
+> The cache file with `.dir` extension is a special text file that contains the
```
shcheklein: same as above

public/static/docs/user-guide/dvcignore.md
```diff
@@ -44,9 +44,9 @@ it raises an error. Ignoring files inside such directory should be handled from
 
 The same as for [`.gitignore`](https://git-scm.com/docs/gitignore).
 
-## Example: Modification of ignored data
+## Example: Ignoring specific files
```
shcheklein: I think we should either keep the previous title (it was correct), or remove it completely (if it's a single example)

public/static/docs/user-guide/dvcignore.md
```diff
@@ -44,9 +44,9 @@ it raises an error. Ignoring files inside such directory should be handled from
...
 
-Let's see what happens when we modify ignored file.
+Let's see what happens when we add a file to `.dvcignore`.
```
shcheklein: `add and then try to modify` ....

public/static/docs/command-reference/checkout.md
jorgeorpinel: Will use suggestion for https://github.com/iterative/dvc.org/pull/931#pullrequestreview-348280129, removing this one...

public/static/docs/command-reference/commit.md
jorgeorpinel: Removing...

public/static/docs/command-reference/install.md
jorgeorpinel: OK, using

```markdown
Note that this command requires the <abbr>DVC project</abbr> to be a Git
repository. But **the hooks won't activate** if the current branch (commit, tag,
etc.) doesn't have DVC initialized (no `.dvc/` directory present).
```

(not in block quote)

Pushed in 014e57f8.

public/static/docs/command-reference/push.md
jorgeorpinel: Removed

public/static/docs/command-reference/repro.md
jorgeorpinel: Removed

public/static/docs/command-reference/fetch.md
jorgeorpinel: You're right. I'm just going to move the "Note that the `.dvc/cache` directory was created and populated." note from the command sample to a regular paragraph and leave the link to /doc/user-guide/dvc-files-and-directories#structure-of-cache-directory for more info... ⏳

public/static/docs/command-reference/push.md
jorgeorpinel: OK just left link to /doc/user-guide/dvc-files-and-directories#structure-of-cache-directory

See bf82ff22.

public/static/docs/user-guide/dvcignore.md
jorgeorpinel: Same. Just left note with link. See bf82ff22.

public/static/docs/command-reference/fetch.md
jorgeorpinel: Done as well. I also found this note in another few docs and removed/updated accordingly.

See 13e516ec

public/static/docs/command-reference/import-url.md
jorgeorpinel: Hmmm. Actually this is probably not true. I was just assuming because of the URL given above `hdfs://user@example.com/path/to/data.csv` which ends with a file extension, same as the HTTP example. I'll double check on this though, but I don't see a reason why we can't support HDFS dirs. Probably reverting, thanks!

jorgeorpinel: I checked internally, @efiop confirmed indeed we do NOT support HDFS directories for these commands. In fact only local, s3, gs, ssh support dirs! Will update these refs.

public/static/docs/command-reference/get-url.md
jorgeorpinel: See https://github.com/iterative/dvc.org/pull/931#pullrequestreview-348279586 below. Updating...

jorgeorpinel: > only local, s3, gs, ssh support dirs 

Dah ok that's all the remaining ones. No update needed! Resolving

public/static/docs/command-reference/import-url.md
jorgeorpinel: Same here: No update needed! Resolving

public/static/docs/command-reference/get.md
jorgeorpinel: This is to keep in a single line:

image

public/static/docs/command-reference/import.md
jorgeorpinel: Same:

image

public/static/docs/user-guide/dvcignore.md
jorgeorpinel: It's 1 out of 2 examples so we can either remove the subtitle or keep it. The previous one wasn't correct though, this example doesn't show any modification of previously ignored data. It's about how it looks to only ignore some files and other not.

jorgeorpinel: But yeah its a pretty basic example so if you're not convinced with the new subtitle I can just remove it for this one, that works too.

public/static/docs/user-guide/dvcignore.md
jorgeorpinel: Oooooh I see what they meant with the previous subtitle now... That was confusing. Let me think about this ⏳

public/static/docs/tutorials/deep/sharing-data.md
jorgeorpinel: Yes. Will be addressed in #946.

public/static/docs/user-guide/dvcignore.md
jorgeorpinel: OK I'm just removing the subtitle.

public/static/docs/user-guide/dvcignore.md
jorgeorpinel: I'm not sure about this suggestion. If you keep reading below, there's this line:

> Now, let's modify file `data1` and see if it affects `dvc status`.

so that covers what you were trying to include here, I think. Especially now that the example's subtitle has been removed.

public/static/docs/command-reference/import.md
```diff
@@ -159,15 +159,10 @@ deps:
...
-However, for typically static references (e.g. tags), or for SHA commits, in
-order to actually "update" an import, it's necessary to **re-import the data**
-instead, by using `dvc import` again without or with a different `--rev`. This
-will overwrite the import stage (DVC-file), either removing or replacing the
-`rev` field, respectively. This can produce an import stage that is able to be
-updated normally with `dvc update` going forward. For example:
+If `rev` moves (e.g. a branch), you may use `dvc update` to bring the data up to
+date. If it's static however (e.g. a commit hash), it would be necessary to
+**re-import the data** instead, by using `dvc import` again with a different (or
+without) `--rev`. For example:
```
jorgeorpinel: Here's the simplification I was talking about in https://github.com/iterative/dvc.org/pull/933#discussion_r373742627 @shcheklein

public/static/docs/glossary.js
```diff
@@ -9,12 +9,9 @@ export default {
...
 Directory containing all your project files. For example raw datasets, source
-code, ML models, etc. A workspace becomes a **DVC project** when
-[\`dvc init\`](/doc/command-reference/init) is run, and
-[DVC-files](/doc/user-guide/dvc-file-format) or stage files are created in it.
-
-Note that [external outputs](/doc/user-guide/managing-external-data) also
-form part of your expanded workspace, technically.
+code, ML models, etc. Note that
+[external outputs](/doc/user-guide/managing-external-data) also form part of
+your expanded workspace.
```
jorgeorpinel: @shcheklein please also notice this change over in this PR as we recently discussed. Thanks

public/static/docs/command-reference/checkout.md
```diff
@@ -25,7 +25,9 @@ workspace data with the versions specified in the current DVC-files.
 that changes the DVC-files in the workspace.
 
 💡 For convenience, a Git hook is available to automate running `dvc checkout`
-after `git checkout`. Use `dvc install` to install it.
+after `git checkout`. See the
+[**Automating** example](#example-automating-dvc-checkout) below or
```
shcheklein: do we need to put into `**`? looks excessive a bit - emoji, link, bold

public/static/docs/user-guide/troubleshooting.md
```diff
@@ -3,6 +3,8 @@
 In this section we provide help for some of the problems that DVC user might
 stumble upon.
 
+<!-- This file uses a special engine feature for the following headers, so that a custom anchor link is used. Just add {#custom-anchor} after each title: -->
```
shcheklein: no multi-line support for the comment?

public/static/docs/user-guide/contributing/docs.md
```diff
@@ -154,3 +154,20 @@ pre-commit hook that is integrated when `yarn` installs the project dependencies
 > Check out the `.md` source code of any command reference to get a better idea,
 > for example in
 > [this very file](https://raw.githubusercontent.com/iterative/dvc.org/master/public/static/docs/user-guide/contributing/docs.md).
+
+## General language guidelines
```
shcheklein: Would it make sense to formalize and start giving examples in terms how we choose terms - being human-friendly first, consistent (to a reason), etc?

public/static/docs/tutorials/deep/sharing-data.md
shcheklein: 👍

public/static/docs/glossary.js
shcheklein: I really don't like this note about external outputs (even if I suggested i, hah - I'm not sure where does it come from:) ). It would not make any sense to 100% readers. We can even test it with our team. Could recall what was the reason?

public/static/docs/command-reference/repro.md
```diff
@@ -28,6 +28,9 @@ then be executed in the correct order, reproducing pipeline results.
 > Pipeline stages are typically defined using the `dvc run` command, while
 > initial data dependencies can be registered by the `dvc add` command.
 
+💡 For convenience, a Git hook is available to remind you to `dvc repro` when
```
shcheklein: it's not necessarily `dvc repro` though. It might be `dvc commit` or `dvc checkout -f` or manually remove files ... it varies - we just run status, that's it

shcheklein: so, not even sure it worth having this note here

public/static/docs/glossary.js
jorgeorpinel: Sure, I can remove it. That note was already there though, this change removes the 2nd sentence (about becoming a DVC project). I'll test the external outputs sentence with the team meanwhile...

public/static/docs/command-reference/import.md
jorgeorpinel: I assume this one is good 🙂

public/static/docs/command-reference/commit.md
```diff
@@ -21,7 +21,10 @@ The `dvc commit` command is useful for several scenarios where a dataset is
 being changed: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/pipeline) is in development, when one wishes
 to run commands outside the control of DVC, or to force DVC-file updates to save
-time tying stages or a pipeline.
+time tying stages or a pipeline. These scenarios are further detailed below.
+
+💡 For convenience, a Git hook is available to remind you to `dvc commit` when
```
shcheklein: I think it does not remind to this, it's just a regular status message.

public/static/docs/command-reference/get.md
shcheklein: but we already have a few commands with two lines? this format makes it somewhat harder to read I would say ... minor issue though

public/static/docs/command-reference/import-url.md
shcheklein: this one looks like a bug we should create a ticket for in case you haven't done it yet

public/static/docs/glossary.js
shcheklein: test and ask if they are able to understand this :) Do you remember why was it added?

public/static/docs/command-reference/import.md
shcheklein: `moves` is not clear enough. How about something like this:


If `rev` is a branch or tag name (any other <right git term>) it is usual that the most recent commit in the brach has updated, or tag now points to a different commit, etc. To update data and the DVC-file to point to the new commit in the branch or tag (<mention rev-lock?>) use ...

If `rev` is static `dvc update` does not have effect, you would need to re-import a new commit (or something like this).

public/static/docs/command-reference/fetch.md
shcheklein: so, my point is more about the `.dir` explanation. We don't need to have a note about it. It's fine to mention about cache structure.

public/static/docs/install/linux.md
```diff
@@ -42,7 +42,7 @@ In this case it installs `boto3` library as well, besides DVC.
 $ conda install -c conda-forge dvc
 ```
 
-> Currently, it supports only Python versions 2.7, 3.6, and 3.7.
+> Currently, it supports only Python versions 3.6, and 3.7.
```
jorgeorpinel: Does anyone know if this note is still current? Cc @efiop thanks

jorgeorpinel: OK Ruslan confirmed that 3.5 3.6 and 3.7 are supported now (3.8 not yet)

jorgeorpinel: Updated.

public/static/docs/command-reference/fetch.md
jorgeorpinel: Yep, I got it. I did remove the notes about .dir cache file and just left a note/link to cache structure 🙂

public/static/docs/command-reference/import.md
jorgeorpinel: Will address along with https://github.com/iterative/dvc.org/pull/931#pullrequestreview-348279331 above.

public/static/docs/command-reference/get.md
jorgeorpinel: Yeah there's no exact rule about how many spaces we leave for these. Probably it's not very consistent. Since you feel there's not enough spaces I'll just add a few more haha and break the lines... ⌛️

jorgeorpinel: OK done in 6f4de01f. I also updated a few more cmd refs for more consistent spacing (just up to index.md for now)

public/static/docs/glossary.js
jorgeorpinel: Just Ruslan answered, he said it makes sense to him. I think they will understand it but anyone who hasn't read about external data will not so you're right about that. But at the same time, this whole paragraph even including the note is pretty short now so I don't think we need to remove it. Anyone who reads this will get the idea from the previous sentence and if they don't understand this question they'll probably just ignore it (or click on the "external outputs" link). Up to you though, please lmk.

jorgeorpinel: I don't remember why we added it. Probably had to do with reviewing or writing about the external outputs, cache, etc.

shcheklein: Yep, that's expected from Ruslan :)

The problem with this that we mix so different things in the paragraph that anyone would think that this advanced one is of the same importance (while it's not). If we decide to keep, make it a paragraph, put `Advanced:`  or `in an advanced case ....`. So that it's clear for 99% of the readers it's super clear that they don't care about this.

public/static/docs/command-reference/import-url.md
jorgeorpinel: I think it makes sense that `get-url` doesn't support `remote://` URLs since it's meant to be used from anywhere, not just from inside DVC repos. Maybe just the error message could be better? I did open an issue about that: iterative/dvc/issues/3236 I will ask a Q there to see if the behavior is correct in the first place.

public/static/docs/tutorials/deep/sharing-data.md
jorgeorpinel: OK it was a small change so I did it here as well (54bcf5e5).

public/static/docs/command-reference/import.md
jorgeorpinel: OK. How about the newest version? (b0de4099)

```markdown
If `rev` is a branch, it will probably have newer commits periodically. To bring
the data up to date (and update `rev_lock` in the DVC-file), simply use
`dvc update`. If `rev` is more static (e.g. a tag or a commit hash) though, you
would need to **re-import** a different commit instead by using `dvc import`
again with a different (or without) `--rev`. For example:
```

public/static/docs/glossary.js
jorgeorpinel: Ok I've removed it and placed a similar but more correct note in the actual external output doc instead. See e6c244c4

> I also did small a revision of the other nearby glossary terms.

public/static/docs/glossary.js
```diff
@@ -25,14 +20,14 @@ Initialized by running \`dvc init\` in the **workspace** (typically in a Git
...
-\`dvc add\` or \`dvc run\`. It may also be a Git repository.
+\`dvc add\` or \`dvc run\`.
       `
     },
     {
       name: 'DVC Repository',
       match: ['DVC repository', 'DVC repositories'],
       desc: `
-**DVC project** initialized using \`dvc init\` in a Git repository. It will
+**DVC project** initialized (with \`dvc init\`) in a Git repository. It will
```
jorgeorpinel: @shcheklein maybe we should merge the terms "DVC project" and "DVC repository" (into just project)? They're basically the same, but one of them specifies that it's in a Git repo.

public/static/docs/command-reference/checkout.md
jorgeorpinel: Haha no we don't. Idk why I did that. Bolding removed.

public/static/docs/user-guide/troubleshooting.md
jorgeorpinel: Ah yeah sorry. I just had to add the first new line after `<!--` and then it rewraps it 😬. Fixed

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: OK good idea. I'm not totally sure what you have in mind but check out this paragraph I added in 6347a36f:

```markdown
We prefer human-friendly language than exact jargon, but we do seek preciseness,
even if using simple, general terms. Our
[command-references](/doc/command-reference) and some
[user-guides](/doc/user-guide) are the most technical documents we have, where
specialized terminology is used the most, but even there, often we use
expandable sections for complex implementation details.
```

Also not sure what you meant exactly by "start giving examples" though.

public/static/docs/command-reference/repro.md
jorgeorpinel: Hmmm well, that's what the `dvc install` ref says it does: https://github.com/iterative/dvc.org/blob/master/public/static/docs/command-reference/install.md

But it sounds like I should try this and confirm that doc first.

jorgeorpinel: > The installed Git hook automates reminding the user to run either dvc repro or dvc commit, as needed.

https://dvc.org/doc/command-reference/install

public/static/docs/command-reference/commit.md
jorgeorpinel: Will follow up on all these notes (it's in several cmd refs) in https://github.com/iterative/dvc.org/pull/931#pullrequestreview-351859026 above.

public/static/docs/command-reference/repro.md
shcheklein: that's should be easy enough to fix, it's two functions 5-10 lines of code each - https://github.com/iterative/dvc/blob/master/dvc/scm/git/__init__.py#L277 . The purpose of the initial doc was probably to clarify when it might be useful to see the status and fail (forgot to commit, repro, manually, may be even push (I would check this one), basically every situation status return non-zero exit code and what does it mean from a user perspective).

public/static/docs/user-guide/contributing/docs.md
shcheklein: `preciseness` sounds strange - correctness? not sure how to better put it


example: `prefer commit over revision whenever possible`
example: options in commands - start with the essence first, simple language, clarifications, edge cases, precision after ...

public/static/docs/glossary.js
shcheklein: yep, good idea. And we can mention that it might or might not be Git controlled.

shcheklein: or even mention there as well that some commands, options etc expect it to be Git as well as DVC and usually it's obvious from the context

public/static/docs/command-reference/import.md
shcheklein: so, `dvc update` works with tags as well, it not that rate that people use them (like `latest` to designate the last to date model file).

otherwise looks good. I'm not sure if there is "static" term used in Git docs.

jorgeorpinel: > dvc update works with tags as well... not that rare that people use them (like latest

I would argue it's pretty rare but OK, I came up with another version that is more complete, hopefully not overcomplicated again:

```markdown
If `rev` is a Git branch or tag, it may have new commits periodically. To bring
the data up to date (and update `rev_lock` in the DVC-file), simply use
`dvc update <stage>.dvc`. If `rev` is a specific commit (does not change),
`dvc update` will never have an effect on this import stage. You may
**re-import** a different commit instead, by using `dvc import` again with a
different (or without) `--rev`. For example:
```

Note that there's a feature request to add `--rev` to `dvc update` but it's not implemented yet: iterative/dvc/issues/2849 – we'll review this text again when that is implemented.

> not sure if there is "static" term used in Git docs

I don't know. I was just using general language 🙂 but that word is gone now anyway.

shcheklein: `it may have new commits periodically` - tags may point to the new commit, not have them. Otherwise looks good :)

jorgeorpinel: Changed to "it may point to a newer commit periodically" but it feels to technical or Git-specific again... ⏳

jorgeorpinel: How about

```markdown
If `rev` is a Git branch or tag (where the commit it points to changes), the
data source may have updates at a later time. To bring it up to date if so (and
update `rev_lock` in the DVC-file), simply use `dvc update <stage>.dvc`. If
`rev` is a specific commit (does not change), `dvc update` will never have an
effect on the import stage. You may **re-import** a different commit instead, by
using `dvc import` again with a different (or without) `--rev`. For example:
```

(latest version in f409d81b)

public/static/docs/install/macos.md
```diff
@@ -63,9 +60,6 @@ In this case it installs `boto3` library as well, besides DVC.
 $ conda install -c conda-forge dvc
 ```
 
-> Currently, it supports only Python versions 2.7, 3.6, and 3.7.
+> Currently, it supports Python 3.5-3.7
 
 <!-- Separate MD quote: -->
```
jorgeorpinel: > can remove it now?
From https://github.com/iterative/dvc.org/commit/d1a6c129a65eff35907d0509044844e1dd1dad96#r37081978

Yes @shcheklein, thanks. Removing

public/static/docs/user-guide/contributing/docs.md
jorgeorpinel: Updated to:

```markdown
We prefer human-friendly language than exact jargon, as long as it's correct,
even if using general terminology. Example: avoid Git jargon such as _revision_
or _reference_, preferring the more basic concept _commit_.

The [command-reference](/doc/command-reference) contains some of our most
technical documents where specialized language is used the most, but even there,
we use expandable sections for complex implementation details.

Start by writing the essence in simple terms, and complete it with
clarifications, edge cases, or other precisions in a separate iteration.
```

Better? Too long now?

public/static/docs/glossary.js
jorgeorpinel: OK! Terms merged in c55a2fe7 (updated all docs accordingly).

> we can mention that it might or might not be Git controlled

This is already mentioned in the definition of _DVC project_.

> some commands, options etc expect it to be Git as well as DVC and usually it's obvious from the context

Added this note into the definition. It's a little long now though:

image

public/static/docs/command-reference/get.md
jorgeorpinel: I'm extracting this into the next regular updates PR #966

public/static/docs/user-guide/contributing/docs.md
shcheklein: looks great! :)

public/static/docs/command-reference/import.md
shcheklein: 👍

public/static/docs/command-reference/get-url.md
```diff
@@ -13,49 +12,55 @@ system.
...
 > storage, you might need to install these optional dependencies: `[s3]`,
 > `[azure]`, `[gdrive]`, `[gs]`, `[oss]`, `[ssh]`. Alternatively, use `[all]` to
 > include them all. The command should look like this: `pip install "dvc[s3]"`.
 > (This example installs `boto3` library along with DVC to support S3 storage.)
 
-Another way to understand the `dvc get-url` command is as a tool for downloading
-data files.
+<!-- Separate MD quote: -->
+
+> See `dvc get` to download data/model files or directories from other <abbr>DVC
```
shcheklein: three notes in a raw - smells bad )

public/static/docs/command-reference/get.md
```diff
@@ -12,39 +12,37 @@ into the current working directory.
...
-can point to any file or directory in the source project, including
-<abbr>outputs </abbr> tracked by DVC as well as files tracked by Git. Note that
-for the former, data should be specified in one of the
-[DVC-files](/doc/user-guide/dvc-file-format) of the source repository. (In this
-case, a default [DVC remote](/doc/command-reference/remote) needs to be
-configured in the project, containing the actual data.)
+target to be downloaded within the source repository at `url`. It can point to
+any file or directory in there, including <abbr>outputs</abbr> tracked by DVC,
+as well as files tracked by Git. Note that for DVC repos, the target should be
+found in one of the [DVC-files](/doc/user-guide/dvc-file-format) pf the project.
```
shcheklein: pf - of?

public/static/docs/command-reference/repro.md
jorgeorpinel: OK. I'm extracting this to a new issue #968

public/static/docs/glossary.js
jorgeorpinel: Should I remove the last sentence? "Some DVC commands..." I don't feel like it belongs in the short definition of the term. It will be mentioned in the appropriate cmd refs anyway.

public/static/docs/command-reference/get-url.md
jorgeorpinel: Agree. And import-url has 4 in a row... Will think about this ⏳

public/static/docs/command-reference/get.md
jorgeorpinel: Addressed in #966

jorgeorpinel:
> few typos

Other than "pf"? (Addressed)

> unaddressed comments

Moving them to #966

> merging this anyway to prevent it from growing again

Thank you!

public/static/docs/glossary.js
jorgeorpinel: Yeah. Removed in 7ef09685 (part of #966)

public/static/docs/command-reference/get-url.md
jorgeorpinel: Addressed in 1551b4b9 (part of #966).

public/static/docs/glossary.js
```diff
@@ -8,33 +8,27 @@ export default {
...
-      match: ['DVC repository', 'DVC repositories'],
-      desc: `
-**DVC project** initialized using \`dvc init\` in a Git repository. It will
-contain \`.git/\` and [\`.dvc/\`](/doc/user-guide/dvc-files-and-directories)
-directories, as well as any DVC-files created by DVC.
+\`dvc add\` or \`dvc run\`. Some DVC commands require it to be in a Git repo –
+explained in each [command reference](/doc/command-reference) but usually
+obvious from the context.
       `
     },
```
jorgeorpinel: These terms were merged per https://github.com/iterative/dvc.org/pull/931#pullrequestreview-351903751, which was the right move. But after a short discussion in https://github.com/iterative/dvc/pull/3277#discussion_r374425536 I think maybe we should make "DVC repository" (or just repo) the main or only term?

Initially I only thought of repos as the Git part of DVC projects, but later on the term DVC repo started popping up and we even had its glossary entry, now we've merged them both but left "project" as the main term. Comparing vs the core DVC repo though, the term "project" is not used at all so it would be easier to just manage the single term "repo" even when sometimes it will be necessary to disambiguate between Git vs DVC – a problem we currently have anyway.

Thoughts @shcheklein ? I can do this in my next regular updates or in its own PR (if it affects lots of files) if so.

public/static/docs/glossary.js
jorgeorpinel: Discussion revived in https://github.com/iterative/dvc.org/pull/931#pullrequestreview-354892810.

PR #936, opened 2020-01-20
cmd ref: get: document --show-url flag

### LINKED ISSUE #930: cmd ref: document `get --show-url` option
iterative/dvc#3156 just got merged. Write docs and reference for the command.

skshetry:
Fixes #930 

### DIFF
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -65,12 +65,10 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
-- `--show-url` - Returns path or url to the location in remote for given path,
-  from where the **file** will be downloaded. `path` is used to specify the location
-  of the target artifact within the source project in `url`, relative to the project's
-  root. `url` can be an address to a DVC project or a path to a local DVC project.
-  When this option is specified, it will not download the files.
-  Git-only repository is not supported.
+- `--show-url` - `url` is expected to represent a DVC project for this option to
+  have an effect (as opposed to a Git-only, non-DVC repository). When
+  `--show-url` is used, instead of downloading the file or directory, this
+  command just prints the storage location (URL) of the target data.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
@@ -124,6 +122,18 @@ repository.
 $ dvc get https://github.com/schacon/cowsay install.sh
 $ ls
 install.sh
+```
+
+### Example: Getting the storage URL of a DVC-tracked file
+
+We can use `dvc get --show-url` to get the actual location of the resulting
+model file from our
+[get started example repo](https://github.com/iterative/example-get-started), a
+<abbr>DVC project</abbr> hosted on GitHub:
+
+```dvc
+$ dvc get https://github.com/iterative/example-get-started model.pkl --show-url
+https://remote.dvc.org/get-started/66/2eb7f64216d9c2c1088d0a5e2c6951
 ```
 
 ## Example: Compare different versions of data or model
@@ -187,17 +197,4 @@ Untracked files:
 
 	model.bigrams.pkl
 	model.monograms.pkl
-```
-
-### Example: Getting a url for a file in a DVC repository
-> Note that `dvc get` can be used from anywhere in the file system, as long as
-> DVC is [installed](/doc/install).
-
-We can use `dvc get --show-url` to get the actual location of the resulting model file from our
-[get started example repo](https://github.com/iterative/example-get-started), a
-<abbr>DVC project</abbr> hosted on GitHub:
-
-```dvc
-$ dvc get https://github.com/iterative/example-get-started model.pkl --show-url
-https://remote.dvc.org/get-started/66/2eb7f64216d9c2c1088d0a5e2c6951
 ```
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -58,8 +58,7 @@ name.
   an existing directory is specified, then the output will be placed inside of
   it.
 
-- `--rev` - `url` is expected to contain a Git repository for this option to
-  have an effect. Specific
+- `--rev` - specific
   [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
   (such as a branch name, a tag, or a commit hash) of the repository to download
   the file or directory from. The tip of the default branch is used by default
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -65,6 +65,13 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
+- `--show-url` - Returns path or url to the location in remote for given path,
+  from where the **file** will be downloaded. `path` is used to specify the location
+  of the target artifact within the source project in `url`, relative to the project's
+  root. `url` can be an address to a DVC project or a path to a local DVC project.
+  When this option is specified, it will not download the files.
+  Git-only repository is not supported.
+
 - `-h`, `--help` - prints the usage/help message, and exit.
 
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
@@ -180,4 +187,17 @@ Untracked files:
 
 	model.bigrams.pkl
 	model.monograms.pkl
+```
+
+### Example: Getting a url for a file in a DVC repository
+> Note that `dvc get` can be used from anywhere in the file system, as long as
+> DVC is [installed](/doc/install).
+
+We can use `dvc get --show-url` to get the actual location of the resulting model file from our
+[get started example repo](https://github.com/iterative/example-get-started), a
+<abbr>DVC project</abbr> hosted on GitHub:
+
+```dvc
+$ dvc get https://github.com/iterative/example-get-started model.pkl --show-url
+https://remote.dvc.org/get-started/66/2eb7f64216d9c2c1088d0a5e2c6951
 ```
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -66,7 +66,7 @@ name.
 
 - `--show-url` - instead of downloading the file or directory, just print the
   storage location (URL) of the target data. `path` is expected to represent a
-  cached, DVC-tracked file for this option to have an effect.
+  cached, DVC-tracked file for this option to have effect.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -64,9 +64,9 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
-- `--show-url` - `path` is expected to represent a cached, DVC-tracked file for
-  this option to have an effect. Instead of downloading the file or directory,
-  just print the storage location (URL) of the target data.
+- `--show-url` - instead of downloading the file or directory, just print the
+  storage location (URL) of the target data. `path` is expected to represent a
+  cached, DVC-tracked file for this option to have an effect.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -65,9 +65,9 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
-- `--show-url` - `path` is expected to represent a DVC-tracked file for this
-  option to have an effect. Instead of downloading the file or directory, just
-  print the storage location (URL) of the target data.
+- `--show-url` - `path` is expected to represent a cached, DVC-tracked file for
+  this option to have an effect. Instead of downloading the file or directory,
+  just print the storage location (URL) of the target data.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/public/static/docs/command-reference/get.md
+++ b/public/static/docs/command-reference/get.md
@@ -126,25 +126,29 @@ install.sh
 
 ### Example: Getting the storage URL of a DVC-tracked file
 
-We can use `dvc get --show-url` to get the actual location of the resulting
-model file from our
-[get started example repo](https://github.com/iterative/example-get-started), a
-<abbr>DVC project</abbr> hosted on GitHub:
+We can use `dvc get --show-url` to get the actual location where the final model
+file from our
+[get started example repo](https://github.com/iterative/example-get-started) is
+stored:
 
 ```dvc
 $ dvc get https://github.com/iterative/example-get-started model.pkl --show-url
 https://remote.dvc.org/get-started/66/2eb7f64216d9c2c1088d0a5e2c6951
 ```
 
+`remote.dvc.org/get-started` is an HTTP
+[DVC remote](/doc/command-reference/remote), whereas
+`662eb7f64216d9c2c1088d0a5e2c6951` is the file's checksum.
+
 ## Example: Compare different versions of data or model
 
-`dvc get` has the `--rev` option, to specify which version of the repository to
-download a <abbr>data artifact</abbr> from. It also has the `--out` option to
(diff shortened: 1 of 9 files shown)


public/static/docs/command-reference/get.md
```diff
@@ -65,6 +65,13 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
+- `--show-url` - Returns path or url to the location in remote for given path,
+  from where the **file** will be downloaded. `path` is used to specify the location
+  of the target artifact within the source project in `url`, relative to the project's
+  root. `url` can be an address to a DVC project or a path to a local DVC project.
+  When this option is specified, it will not download the files.
+  Git-only repository is not supported.
```
skshetry: Probably, non-DVC repository would have been better?

skshetry:
Ooo, lint is failing.

public/static/docs/command-reference/get.md
```diff
@@ -65,6 +65,13 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
+- `--show-url` - Returns path or url to the location in remote for given path,
```
shcheklein: `Returns` -> `returns`

shcheklein:
@skshetry install the pre-commit hook with `npm install -g yarn`, `yarn` to fix the style

public/static/docs/command-reference/get.md
shcheklein: let's not explain the positional args (`path` and `url` here) - they have the same meaning and are probably already described, right?

public/static/docs/command-reference/get.md
shcheklein: why is it `path or ..`?

I would simplify significantly the whole piece.

"instead of downloading the file or directory print its actual location (URL) it would be downloaded from, e.g. s3://remote/storage/0c/121243234"

public/static/docs/command-reference/get.md
shcheklein: and/or even put a link to the example below ...

public/static/docs/command-reference/get.md
```diff
@@ -181,3 +188,16 @@ Untracked files:
 	model.bigrams.pkl
 	model.monograms.pkl
 ```
+
+### Example: Getting a url for a file in a DVC repository
+> Note that `dvc get` can be used from anywhere in the file system, as long as
```
shcheklein: have some doubts ... does not feel we need to repeat it here ... why did you decide to put it?

public/static/docs/command-reference/get.md
jorgeorpinel: Agree with Ivan. This could start:
```suggestion
- `--show-url` - returns URL to the remote location of the target data
```
but the rest of the paragraph can be simplified as well...

public/static/docs/command-reference/get.md
jorgeorpinel: Yep, `url` and `path` are explained in the description. No need to do it again here, you can simply reference them. You can even use ``(`url`/`path`)`` as in the introductory paragraph.

jorgeorpinel: Oh and
> Probably, "non-DVC repository"...

Yes, I like that term better. But typically we begin with applicability/restriction sentences in option paragraphs, for example see the language used here:
image

Thanks

public/static/docs/command-reference/get.md
jorgeorpinel: You mean the note about "can be used from anywhere in the file system" right Ivan? Not the whole example

jorgeorpinel: Yaah, agree about the note. No need 🙂

public/static/docs/command-reference/get.md
```diff
@@ -181,3 +188,16 @@ Untracked files:
...
+> DVC is [installed](/doc/install).
+
+We can use `dvc get --show-url` to get the actual location of the resulting model file from our
+[get started example repo](https://github.com/iterative/example-get-started), a
+<abbr>DVC project</abbr> hosted on GitHub:
```
jorgeorpinel: Suggestion, not necessarily to use exactly:

``We can use `dvc get --show-url` to get the actual location where the final model file from our [get started example repo](https://github.com/iterative/example-get-started) is stored. Note that `remote.dvc.org/get-started` is an HTTP [DVC remote](/doc/command-reference/remote), whereas `662eb7f64216d9c2c1088d0a5e2c6951` is the file's checksum.``

public/static/docs/command-reference/get.md
skshetry: @jorgeorpinel, is the distinction not important, that, it can also be a path, not just the URL?

jorgeorpinel: It's already confusing enough that the command takes both `url` and `path`, `url` accepting URLs and paths (explained in description), and then on top of that saying this option returns a "path or URL".
But OK I guess you're right, maybe check out the paragraph about the `url` command param and use similar language to explain this returned URL could be a local file system path (for local remotes).
OR don't use either path nor URL and just say something like "storage location" with URL corresponding to the type of remote, linking to remote cmd ref, where the "local" case is explained.

shcheklein: @jorgeorpinel @skshetry what about the way I suggested it? One sentence, using terms users would understood? `instead of downloading the file or directory print its actual location (URL) it would be downloaded from, e.g. s3://remote/storage/0c/121243234`? One simple sentence + even example to put more context.

shcheklein:
@skshetry @jorgeorpinel Saugat, are you okay if we start iterating on this with you? Jorge will put some modifications and will ask you to review before merging.

skshetry:
@shcheklein, yes, if that'd be faster. I was just waiting for replies on the discussions before working on this.

shcheklein:
@skshetry @jorgeorpinel yes, I think it would be faster - your first iteration -> our changes applied -> you review and suggest fixes -> merge :) otherwise we spend too much time iterating for no reason. The most important here is that you review the last iteration to learn and teach us :)

skshetry:
@shcheklein, okay, makes sense. :slightly_smiling_face:

jorgeorpinel:
> I think it would be faster - your first iteration -> our changes applied -> you review and suggest fixes -> merge

It is faster for small changes but not necessarily when we need to rethink the phrasing, etc. (because the docs team is small and can easily get backed up). But yep, I'll take over this one ASAP.

The CI format-check is failing though (and Restyled didn't catch it, weird). Please take a look: https://circleci.com/gh/iterative/dvc.org/1890

---

Note: please use a branch directly on this repo instead of your fork in future PRs @skshetry – Ivan can give you access if needed. It makes it easier for others to contribute to the PR. Thanks

shcheklein:
@jorgeorpinel 

> when we need to rethink the phrasing

in certain cases we can merge it then, even if it's not 100% perfect. It would be taking too long to expect from everyone to write the text of the highest quality.

public/static/docs/command-reference/get.md
jorgeorpinel: Yeah I think you got it right Ivan. I added to the core repo in https://github.com/iterative/dvc/pull/3220/commits/ee0193ccd36fdc067ace131958f7960819202643 and will update here as well.

public/static/docs/command-reference/get.md
jorgeorpinel: Updated in 072defa3.

public/static/docs/command-reference/get.md
jorgeorpinel: Updated in 072defa3.

public/static/docs/command-reference/get.md
jorgeorpinel: Removed in 072defa3.

public/static/docs/command-reference/get.md
```diff
@@ -181,3 +188,16 @@ Untracked files:
 	model.bigrams.pkl
 	model.monograms.pkl
 ```
+
+### Example: Getting a url for a file in a DVC repository
```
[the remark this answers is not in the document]
jorgeorpinel: Also done in 072defa3.

jorgeorpinel:
I wasn't referring to language quality but to clarify the idea and include all the information needed, even if it will be rewritten with different wording. Some reviews include this kind of questions.

jorgeorpinel:
Some reviews include this kind of questions which is easier/faster for the person who developed the change to include.

public/static/docs/command-reference/get.md
jorgeorpinel: Updated in 663f589a.

jorgeorpinel:
> I'll take over this one ASAP.

Ok, I have now, and addressed all the feedback. Please review @shcheklein @skshetry. Thanks

public/static/docs/command-reference/get.md
```diff
@@ -65,6 +65,11 @@ name.
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
+- `--show-url` - `url` is expected to represent a DVC project for this option to
+  have an effect (as opposed to a Git-only, non-DVC repository). When
+  `--show-url` is used, instead of downloading the file or directory, this
+  command just prints the storage location (URL) of the target data.
```
shcheklein: I still think we are overcomplicating this. The fact that it does not handle Git repos is only somewhat important and at least should go after the essence of what this command does. Also, probably it's not about DVC repo vs Git repo - it's about Git-tracked files vs DVC-tracked files. Am I correct here, @skshetry ?

Also, consider including a link the example below.

skshetry: No, it's about DVC repo vs Git repo. As an example, take this *git* repo. How can you get an url for this `public/static/docs/command-reference/get.md`? Git does not work with files, so there's no URL/path to generate for (unless we do somekind of *magic*).

shcheklein: okay, to clarify it's not about Git-only vs DVC. Any Git+DVC has files that are not under DVC control and probably we don't generate URL for them, right?

shcheklein: I'm not even sure about DVC-controlled non-cached files (`dvc run -O something`), probably we don't even handle that with this new option (even though we can download it)

skshetry: Yes, that should not also work, as it checks for all DVC controlled outputs.

skshetry: So, it's also about Git vs DVC tracked files. Sorry about that.

jorgeorpinel: > The fact that it does not handle Git repos is only somewhat important and at least should go after the essence of what this command does.

@shcheklein You may be right but about this we're already using this style in most other options when there are pre-requisites. Putting the pre-req's as the first sentence. (See `--rev` just above this.) That's not a great reason to keep this but I'm also not convinced that leaving the pre-req's for last is best. Maybe it is better to leave them on top? they are "pre" requisites after all 🙂

All that said I did shorten that first sentence now, and also the main explanation following it.

> to clarify it's not about Git-only vs DVC. Any Git+DVC has files that are not under DVC control

True, I updated that as well. So the pre-req is not about `url` then, but about `path`: it should be a DVC-tracked file/dir.

See ab40ed7f.

jorgeorpinel: > I'm not even sure about DVC-controlled non-cached files (dvc run -O something), probably we don't even handle that with this new option (even though we can download it)

This sounds like something I should note here and in `dvc.api.get_url` as well?

jorgeorpinel: > I'm not even sure about DVC-controlled non-cached files (dvc run -O something), probably we don't even handle that with this new option (even though we can download it)

Actually, both this option and `get_url()` should support creating a URL for these. They don't atually check for the existence of the file in remote cache. They just interpret the project configured remote + the checksum in DVC-file from what I understand.

public/static/docs/command-reference/get.md
```diff
@@ -58,13 +58,17 @@ name.
   an existing directory is specified, then the output will be placed inside of
   it.
 
-- `--rev` - `url` is expected to represent a Git repository for this option to
+- `--rev` - `url` is expected to contain a Git repository for this option to
```
shcheklein: I would say in this case we don't need to specify these "pre-req" at all - neither before or after :) May be something like "Obviously, a Git repository must be located at provide location (`url` arg)." at the end.

public/static/docs/command-reference/get.md
```diff
@@ -58,13 +58,17 @@ name.
...
   have an effect. Specific
   [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
   (such as a branch name, a tag, or a commit hash) of the repository to download
   the file or directory from. The tip of the default branch is used by default
   when this option is not specified.
 
+- `--show-url` - `path` is expected to represent a cached, DVC-tracked file for
```
shcheklein: same here, still would prefer to see the essence of it. After that something like: Limitations: it does not print anything (or throws an exception?) for data files or directories that are not DVC-tracked (usually, saved into Git directly instead). See example ... link"

efiop:
@shcheklein @jorgeorpinel Considering that Saugat is away for ~2weeks, could we merge it and work on top later?

public/static/docs/command-reference/get.md
jorgeorpinel: Yeah I guess it's obvious. No need to add the "obviously" note though, doesn't sound friendly. Removed this prereq sentence.

shcheklein:
@efiop yep, it's on us now, no expectations from the eng team at this point.

public/static/docs/command-reference/get.md
jorgeorpinel: OK, moved the prereq to last in cab347ca.

public/static/docs/command-reference/get.md
jorgeorpinel: Applied the same for `dvc import`.

public/static/docs/command-reference/get.md
jorgeorpinel: And made standard use of this order in other commands.

jorgeorpinel: > Limitations: it does not print anything...

p.s. I think it's best to just say what the prereq is and not what the command will do if the prereq is not met. Easier text to maintain.

jorgeorpinel:
Docs team for the win 🎉

public/static/docs/command-reference/pull.md
```diff
@@ -78,9 +78,9 @@ reflinks or hardlinks to put it in the workspace without copying. See
   searches backward from the target stages in the corresponding pipelines. This
   means DVC will not pull files referenced in later stages than the `targets`.
 
-- `-R`, `--recursive` - `targets` is expected to contain one or more directories
-  for this option to have effect. Determines the files to pull by searching each
-  target directory and its subdirectories for DVC-files to inspect.
+- `-R`, `--recursive` - determines the metric files to show by searching each
+  target directory and its subdirectories for DVC-files to inspect. `targets` is
+  expected to contain one or more directories for this option to have effect.
```
skshetry: copypasta? :slightly_smiling_face:

public/static/docs/command-reference/push.md
```diff
@@ -87,9 +87,9 @@ to push.
   searches backward from the target stages in the corresponding pipelines. This
   means DVC will not push files referenced in later stages than the `targets`.
 
-- `-R`, `--recursive` - `targets` is expected to contain one or more directories
-  for this option to have effect. Determines the files to push by searching each
-  target directory and its subdirectories for DVC-files to inspect.
+- `-R`, `--recursive` - determines the metric files to show by searching each
+  target directory and its subdirectories for DVC-files to inspect. `targets` is
+  expected to contain one or more directories for this option to have effect.
```
skshetry: Same here as https://github.com/iterative/dvc.org/pull/936/files#r376252167

public/static/docs/command-reference/pull.md
skshetry: Fixed by #980

public/static/docs/command-reference/push.md
skshetry: Fixed by #980

PR #943, opened 2020-01-23
server: refactor redirects so they're more declarative

### LINKED ISSUE #757: server: better way to handle HTTP redirects

fabiosantoscode:
This places redirects in a separate file and attempts to simplify them.

I've identified two kinds of redirects, a redirect based on the host, and a redirect based on pathname. I've used classes but I'm not married to that approach.

I've added cache-control: no-cache to all redirects and they're all 303's, but this approach can be extended to include some options specific to each redirect as an options object in the constructor.

Fixes #757

----


fabiosantoscode:
Looks like an unrelated lint issue popped up in a file I didn't change.

shcheklein:
@fabiosantoscode okay, this is a good ticket, but we should handle it in a different way. Let's discuss this on Discord.

server.js
```diff
@@ -14,104 +14,44 @@ const { parse } = require('url')
...
-      // path /(deb|rpm)/... -> corresponding S3 bucket
+    let redirectLocation = getRedirectLocation(req, host, pathname)
+
+    if (redirectLocation) {
+      // should be getting the query as a string
+      const { query } = parse(req.url)
+      if (query) {
+        redirectLocation += '?' + query
+      }
       res.writeHead(303, {
```
shcheklein: the difference between 301 and 303 is very important

shcheklein:
@fabiosantoscode check the comment re the 301 vs 303, let's try to wrap it as a worker for the Cloudflare. Move redirects.js into redirects.json and try to make it way more declarative.

fabiosantoscode:
@shcheklein do you mean that you want me to make this a worker right now? I couldn't find any existing cloudflare workers on this repository.

fabiosantoscode:
@shcheklein I've addressed the part of the feedback I'm sure about, by making some redirects pure and others not. I've also changed the file to JSON and used an approach that will be familiar to anyone who knows how `''.replace` works.

 * `match` will indicate what to match against, currently only `url` for the whole URL or `pathname` for just the pathname
 * `regex` is the regexp. The groups are important here.
 * `replace` is passed verbatim to the second argument of `''.replace`. Here you can use `$1`, `$2`, etc to refer to strings captured in the regex groups.
 * `permanent` causes a permanent redirect. At first I used 308 but I figured support for this would be worse than 301, and after looking at the RFCs it looks like they don't have any semantic difference, and no behaviour difference except for POST requests, which are not a thing on this website.

Please let me know if you'd like me to create a service worker which could be used in the browser or cloudflare, or if the changes I made were only in anticipation of moving to cloudflare workers.

Cheers! :snowman:

fabiosantoscode:
Is it me or codeclimate is being extremely picky?

shcheklein:
@fabiosantoscode yes, pay reasonable attention to it, no need to fix all complaints :)

shcheklein:
Good stuff! Next steps here:

1. this json is too bloated, hard to read, manage, follow. Let's do an array of "expression redirect code". We can figure out from the expression if it's url or path, we can specify explicitly the code (and by default consider it 301).
2. we need tests. Not that we have a wrapper we can write some unit tests for this and check carefully. Also tests should be able to detect loops.

fabiosantoscode:
@shcheklein So an array of space-separated strings then?

The tests will be cool to write, I'll get on it.

fabiosantoscode:
@shcheklein I've used a bunch of spaces to align the targets and make the JSON file easier to read, even if it's wider.

I've had to create a `itRedirects` "custom it function" to try and make the tests concise.

fabiosantoscode:
I mean, I didn't have to but it seemed nicer.

.eslintrc.json
```diff
@@ -7,7 +7,8 @@
   ],
   "parser": "babel-eslint",
   "env": {
-    "browser": true
+    "browser": true,
+    "jest": true
```
shcheklein: just curious, what does it do?

server.js
```diff
@@ -14,124 +14,100 @@ const { parse } = require('url')
...
+  if (req != null && req.headers['x-forwarded-proto'] !== 'https' && !dev) {
+    return [301, `https://${host.replace(/^www\./, '')}${req.url}`]
+  }
+
+  return matchRedirectList(host, pathname)
+}
+
+exports.getRedirect = getRedirect
+
+// Do not run the server if the module is being required by something else
```
shcheklein: let's move this logic to a separate file? redirets.js?

shcheklein: so that here we keep only clean simple server runner

server.test.js
```diff
@@ -0,0 +1,135 @@
...
+    expect(regex.source).toEqual('^\\/foo')
+    expect(replace).toEqual('/bar')
+    expect(code).toEqual(418)
+  })
+  it('code defaults to 301', () => {
+    const { code } = processRedirectString('^/x /x')
+
+    expect(code).toEqual(301)
+  })
+  it('detects whether we are matching a pathname or the whole url', () => {
```
shcheklein: convention is to have a single space between test in the existing tests (one single file probably) :)

server.test.js
```diff
@@ -0,0 +1,135 @@
...
+      expect(rCode).toEqual(code)
+
+      // Detect redirect loops
+      const secondUrl = url.parse(addHost(rLocation))
+      const secondRedirect = getRedirect(secondUrl.hostname, secondUrl.pathname)
+      expect(secondRedirect).toEqual([])
+    })
+  }
+
+  describe('host redirects', () => {
```
shcheklein: @jorgeorpinel could you take a second look please, that we got all the redirects right

.eslintrc.json
fabiosantoscode: It adds `describe`, `it`, `expect`, etc. as global variables. This prevents the lint from going crazy in the tests. The tests might not be linted, but my editor certainly doesn't care :)

server.js
fabiosantoscode: Sure!

server.test.js
fabiosantoscode: Sorry, I missed that. I'll add spaces.

.eslintrc.json
shcheklein: 👍

fabiosantoscode:
@shcheklein this sounds like something pingdom can do, but I'm not sure we can check where the redirect goes to.

We can also use circleCI to run periodic tests against production. We can reuse the redirects tests I added, but add an environmental variable that tells them to hit production. We can add more tests like this later.

Another thing that might work is google cloud platform uptime checks. You can poll a certain URL (like man.dvc.org/add) and make sure the response _content_ contains or matches a certain string. However this is only feasible if we add a message explaining the redirect containing a redirect link target as the redirect response body. Additionally, another similar check can check for the uptime of the blog itself. And these uptime checks can ping people through e-mail, SMS and other media.

shcheklein:
merging this, if we hit any issue will be solving as we go. Takes too long otherwise.

server.js
```diff
@@ -14,124 +14,59 @@ const { parse } = require('url')
...
-      res.end()
-    } else if (/^\/doc(\/.*)?$/.test(pathname)) {
       /*
-       * Docs Engine handler
+       * HTTP redirects
        */
+      let [redirectCode, redirectLocation] = getRedirect(host, pathname, {
+        req,
+        dev
+      })
```
jorgeorpinel: Maybe ```js
      let [redirectCode, redirectLocation] = getRedirect(
        host,
        pathname,
        {req, dev}
      )
```

Comment (jorgeorpinel) [reply server.js]: Nope. I see this is automatically formatted this way. Never mind.

File: server.js [hunk 7]
```diff
@@ -14,124 +14,59 @@ const { parse } = require('url')
...
 const app = next({ dev })
 const handle = app.getRequestHandler()
 const port = process.env.PORT || 3000
 
-app.prepare().then(() => {
-  createServer((req, res) => {
-    const parsedUrl = parse(req.url, true)
-    const { pathname, query } = parsedUrl
+// Do not run the server if the module is being required by something else
+if (module.parent == null) {
```
jorgeorpinel: What is this condition for @fabiosantoscode ? I don't think server.js is ever required by any other source code file.

server.js
```diff
@@ -14,124 +14,59 @@ const { parse } = require('url')
 const next = require('next')
 
 const { getItemByPath } = require('./src/utils/sidebar')
+const { getRedirect } = require('./src/redirects.js')
```
jorgeorpinel: Should this live in src/utils @fabiosantoscode @shcheklein ? Thanks

server.js
jorgeorpinel: What is this condition for @fabiosantoscode ? I don't think server.js is ever required by any other source code file.

shcheklein: it's not part of the merged latest version it seems

server.js
shcheklein: yep, we can move them .. move redirects-list to the root porbably

server.js
shcheklein: @fabiosantoscode let's remove this `if`, we don't need this anymore I think

fabiosantoscode: Correct. @jorgeorpinel this was because I was importing server.js from the tests before, as the redirect logic was contained in it.

server.js
fabiosantoscode: Will move and PR

jorgeorpinel:
My last comment here is that I personally find the space separation in the JSON file strange. A normal JSON file with an array of arrays or something like that should work as well and easier to read? Let me review the new PR and will see if this comment is still valid. I may also just push some suggested changes myself if needed. Thanks

server.js
jorgeorpinel: Thanks!

server.js
jorgeorpinel: Got it, thanks.

shcheklein:
@jorgeorpinel we had an iteration like this and I didn't like it (not sure about @fabiosantoscode  :) ) - it's 1000 lines vs 10. It's way easier to read line by line. It's very regular format for redirects files - one rule a line. There should be very strong reasons to complicate this.

PR #954, opened 2020-01-28
Extend Privacy Policy with used Google scopes

MaxRis:


shcheklein:
@MaxRis @jorgeorpinel any idea why it does not run CircleCI for this branch from the top of your head?

public/static/docs/user-guide/privacy.md
```diff
@@ -42,6 +42,13 @@ section of this User Guide.
 
 ## Scopes
 
+userinfo.email scope (view your email address), userinfo.profile scope (see your
```
shcheklein: let's use `userinfo.email` (inline code) blocks for scope names to make a bit more readable

MaxRis: @shcheklein have updated markup and have added mention about `drive.appdata` scope as well.
Not sure why CicrcleCI won't start (how it might be related to my GitHub account).

shcheklein: @MaxRis don't worry about this for now. Have asked their support to check what the issue could be.

public/static/docs/user-guide/privacy.md
```diff
@@ -42,8 +42,16 @@ section of this User Guide.
 
 ## Scopes
 
+`userinfo.email` scope (view your email address), `userinfo.profile` scope (see
+your personal info, including any personal info you've made publicly available),
+and `openid` scope (authenticate using OpenID Connect) are required to be used
+by Google Auth. DVC authorizes users via
+[Google's server](https://accounts.google.com/o/oauth2/auth) and doesn't access
```
jorgeorpinel: Hi! This link doesn't work like this. Should we just display it instead like `` `https://accounts.google.com/o/oauth2/auth` `` ?

MaxRis: Hey @jorgeorpinel , the idea here was to mention explicitly that server belongs to Google and link was added just as a prove. Also link might change with time.

MaxRis: Also not sure why it doesn't work? I can click at "Google's server" and the URL is opened in new tab ( error in new tab is displayed, but it is fine )

shcheklein: I think @jorgeorpinel refers to the fact that it returns 404 and looks like broken bc of that. Also trigger our CI scripts that detect broken/dead links. I've added it to the exclusion list for now. Still, not clear if it makes sense to keep it this way, even though I understand the intention.

jorgeorpinel: Yep, we need it in the link exclusion list anyway, but it's best to display the link in the markdown like a `` `quote` `` instead of a clickable link, I think.

jorgeorpinel: And yes, I meant it's not a link to any useful website. It doesn't work like that = it's not meant to be navigated, just the base URL to begin a OAuth authentication (but needs a special token in the URL query)

MaxRis: @jorgeorpinel got it, thanks, should I propose the changes?

shcheklein: @jorgeorpinel let's fix it as part of the regular updates flow.

jorgeorpinel: Done in 45e93dc6. I reorganized and reworded some paragraphs along the way so please review the draft PR (#987).

jorgeorpinel: p.s. I ay have some more Qs, will add them there.

jorgeorpinel:
Sorry to keep kicking a dead horse but I have another follow up question to this PR guys:

Will the current guide in https://dvc.org/doc/user-guide/setup-google-drive-remote be completely rewritten soon? Since we're using a default _DVC_ Application on GCP it seems to me the guide will be much simpler, but maybe this isn't completely done yet (Google APIs verification process).

shcheklein:
@jorgeorpinel it'll be augmented most likely, not rewritten. We still recommend using your own app/project.

PR #967, opened 2020-02-04
api: add list command

JIoJIaJIu:
Relates:
- [Pull#2346](https://github.com/iterative/dvc/pull/3246)
- [Issue#2509](https://github.com/iterative/dvc/issues/2509)

<hr/>

### DIFF
--- a/public/static/docs/command-reference/list.md
+++ b/public/static/docs/command-reference/list.md
@@ -1,0 +1,130 @@
+# list
+
+List Git repository contents including files and directories under DVC control
+(<abbr>data artifacts</abbr>).
+
+## Synopsis
+
+```usage
+usage: dvc list [-h] [-q | -v] [-R] [--outs-only] [--rev [REV]] url [target]
+
+positional arguments:
+  url              Url to the repo. Can be either local fs path or Git repository URL.
+  target           Path to directory within the repository to list contents for.
+```
+
+## Description
+
+List files, dirs and <abbr>data artifacts</abbr> for the pointed url. The output
+is sorted lexicographically.
+
+The `url` argument specifies the address of the DVC or Git repository containing
+the data source. Both HTTP and SSH protocols are supported for online repos
+(e.g. `[user@]server:project.git`). `url` can also be a local file system path.
+When the url is remote git url the content is checkout into temporary directory.
+
+`--outs-only` option allows to check all <abbr>data artifacts</abbr> into the
+repo, which can be useful before `dvc get` or `dvc import` invocation.
+
+The `target` argument allows to check existense of the target into the repo.
+With recursive option `-R` it allows to filter output by prefix. The `target`
+should be relative file system path.
+
+## Options
+
+- `--outs-only` - show only <abbr>data artifacts</abbr>.
+
+- `-R`, `--recursive` - recursively prints the repo.
+
+- `--rev` - commit hash, branch or tag name, etc. (any
+  [Git revision](https://git-scm.com/docs/revisions)) of the repository to
+  download the file or directory from. The latest commit in `master` (tip of the
+  default branch) is used by default when this option is not specified.
+
+- `-h`, `--help` - prints the usage/help message, and exit.
+
+- `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
+  problems arise, otherwise 1.
+
+- `-v`, `--verbose` - displays detailed tracing information. when this option is
+  not specified.
+
+## Example: List remote git repo
+
+We can use the command for getting information about remote repository with all
+files, dirs and <abbr>data artifacts</abbr>.
+
+```dvc
+$ dvc list -R https://github.com/iterative/dataset-registry
+
+.gitignore
+README.md
+get-started/.gitignore
+get-started/data.xml
+get-started/data.xml.dvc
+images/.gitignore
+images/dvc-logo-outlines.png
+images/dvc-logo-outlines.png.dvc
+images/owl_sticker.png
+images/owl_sticker.png.dvc
+images/owl_sticker.svg
+...
+```
+
+Or
+
+```dvc
+$ dvc list https://github.com/iterative/dataset-registry
+
+.gitignore
+README.md
+get-started
+images
+tutorial
+use-cases
+```
+
+for getting flat information about the repo
+
+## Example: List the repo with the rev
+
+Another useful case is checking the files for the **specific revision**
+
+```dvc
+$ dvc list -R --rev 7476a858f6200864b5755863c729bff41d0fb045 \
+        https://github.com/iterative/dataset-registry
+
+.gitignore
+README.md
+get-started/.gitignore
+get-started/data.xml
+get-started/data.xml.dvc
+tutorial/nlp/.gitignore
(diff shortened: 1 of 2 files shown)


public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+## Synopsis
+
+```usage
+usage: dvc list [-h] [-q | -v] [-R] [--outputs-only] [--rev [REV]] url [target]
+
+positional arguments:
+  url              Url to the repo. Can be either local fs path or git+ssh/https url
+  target           Path to directory within the repository to list outputs for.
+
+optional arguments:
```
shcheklein: we don't include this section, check other commands please (kinda repetitive with the `## options` below)

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
+# list
+
+List file structure of a repo including <abbr>data artifacts</abbr>
```
shcheklein: List Git repository contents including files and directories under DVC control (data artifacts)

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Synopsis
+
+```usage
+usage: dvc list [-h] [-q | -v] [-R] [--outputs-only] [--rev [REV]] url [target]
+
+positional arguments:
+  url              Url to the repo. Can be either local fs path or git+ssh/https url
```
shcheklein: might require changes on the DVC core repo:

Url -> URL

I would simplify: `Git repository URL.`

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+  target           Path to directory within the repository to list outputs for.
```
shcheklein: `outputs` -> `to list contents for`

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+
+optional arguments:
+  --outputs-only   Show only dvc outputs.
+  --rev [REV]      Git revision (e.g. branch, tag, SHA)
+```
+
+## Description
+
+The command list files and <abbr>data artifacts</abbr> into pointed repo.
```
shcheklein: need to expand and elaborate here. No need to repeat just the same phrase we used to above. Even http://man7.org/linux/man-pages/man1/ls.1.html has something meaningful and non-repetitive here.

Would be great to expand on the fact that we show actual data behind DVC-files. That it can be used before `dvc get` to see that file name to path (which is not easy with Github), etc.

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Options
+
+- `--outputs-only` - Show only <abbr>data artifacts</abbr>.
```
shcheklein: Show -> show (and for other options as well)

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+
+- `-R`, `--recursive` - Recursively traverse the repo.
```
shcheklein: the repo -> the target? actually it's is the path (that's why I don't like the target name for the CLI argument)

shcheklein: check the aws s3 ls and ls man pages. traverse is not the right verb here ... print, list, etc are better

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+
+- `--rev` - specific
+  [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
```
shcheklein: @jorgeorpinel will need to update this when your PR is merged, or suggest the change here

public/static/docs/command-reference/list.md
```diff
@@ -0,0 +1,40 @@
...
+  (such as a branch name, a tag, or a commit hash) of the repository to download
+  the file or directory from. The tip of the default branch is used by default
+
+- `-h`, `--help` - prints the usage/help message, and exit.
+
+- `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
+  problems arise, otherwise 1.
+
+- `-v`, `--verbose` - displays detailed tracing information. when this option is
```
shcheklein: `Works only for remote repos` ?

public/static/docs/command-reference/list.md
JIoJIaJIu: `dvc list` can list simple folder (without `.git`), shouldn't it be pointed

public/static/docs/command-reference/list.md
JIoJIaJIu: If the repo is on the filesystem pointing `rev` throwing an error. `dvc list` lists files through fs walk rather then `ls-tree` so for the current implementation needs to checkout the repo
https://github.com/iterative/dvc/pull/3246#discussion_r373534682

public/static/docs/command-reference/list.md
shcheklein: I would mention it in the description - that's what it is for - to elaborate and expand

public/static/docs/command-reference/list.md
shcheklein: but this `-v` option, verbose ... It should be working for any combination of params

public/static/docs/command-reference/list.md
jorgeorpinel: It's merged. This text seems correct except download->list and the last sentence is incomplete, should be "...by default when this option is not specified."

jorgeorpinel:
Please let us know when iterative/dvc/pull/3246 is merged to check this doc vs the final behavior of the command 🙂

JIoJIaJIu:
@shcheklein @jorgeorpinel Could you take a look please on the PR

public/static/docs/command-reference/list.md
jorgeorpinel: Agree. If `target` can only be a directory path, why not call it `path` or `dir`? (Also commenting in iterative/dvc.org/pull/967)

jorgeorpinel:
@JIoJIaJIu any chance you could move this to a regular branch in this repo though? If not, it's OK.

Please check out https://github.com/iterative/dvc.org/wiki/2.-dvc.org-pull-requests-for-core-repo-changes for future ref. 🙂

JIoJIaJIu:
@jorgeorpinel @shcheklein 

> any chance you could move this to a regular branch in this repo though

I've create a new PR - https://github.com/iterative/dvc.org/pull/1021 and close this one

> Agree. If target can only be a directory path, why not call it path or dir?

I suggest to apply my current PR and in the next iteration I will rename `target` to `path`, at this moment I am blocked with the https://github.com/iterative/dvc/pull/3351 for being able to refactor the source code. As soon it is merged I will change all `target` entries to `path` and update the docs

jorgeorpinel:
Thanks Guro!

> I suggest to apply my current PR and in the next iteration I will rename target to path

Yep no rush, it's included in iterative/dvc/issues/3381 actually so it should be addressed whenever someone takes that issue.

Issue #792, opened 2019-11-14
css: migrate from styled comps to plan css or css-modules


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/776#discussion_r343969564

Our [styled components](https://www.styled-components.com/) tend to be defined after the exported component in each file (pages/, src/) so it could be confusing to find them in the JSX of previously defined components.

However we want to keep them in the bottom of each file given their strange code format. The idea then is to RENAME them so it's more obvious that they are styled components. My proposal is to simply add the name of the HTML tag they use in CamelCase. For example a `Container` component that styles a `div` tag would be renamed `ContainerDiv` and look like this in JSX.

UPDATE: Probably going for extracting and not renaming. See https://github.com/iterative/dvc.org/issues/792#issuecomment-569555264

shcheklein:
I would check if there are best practices around this. Also, would love to know @iAdramelk 's take on this.

jorgeorpinel:
I don't think styled-components.com are so used that there will be broadly established practices around it but I did look this up just now and found https://medium.com/inturn-eng/naming-styled-components-d7097950a245 where they propose using `StyledContainer` or `S.Container`, which is what Alexey suggested in https://github.com/iterative/dvc.org/pull/776#discussion_r343969564. But I still like `ContainerDiv` more. Let's just vote on it?

jorgeorpinel:
BTW if we do this and rename the `Link` styles components to something else, we should probably rename `NextLink` to just `Link` (it's natural name from [next/link](https://nextjs.org/docs/api-reference/next/link)).

jorgeorpinel:
Oh, and another option to consider for all this is to migrate to https://github.com/zeit/styled-jsx (but doesn't support nested styles I think)

jorgeorpinel:
Do you have a vote here, @iAdramelk? Thanks

iAdramelk:
@jorgeorpinel  I do think that we should probably move them to the separate files at least. I started experimenting with this approach in https://github.com/iterative/dvc.org/pull/874 (check the files in `src/Community`) and I think that this way it is a lot more convenient. We can rename them too, but if we move them to the separate files it is not that important, because it is easier to understand that they are a part of the styles. Alternatively, naming convention can be to add prefix to them. Like `SCLink` instead of `Link`. This way we will not lock ourselves in using predefined tags there and will be able to change tags without changing component names.

[...]

Issue #777, opened 2019-11-06
js: refactor repetitive React components


jorgeorpinel:
(and possibly elements, if any)

Could many of these be merged and use State/Props instead of repetition to apply contextual differences? Or made into a class hierarchy?

Example:

https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/PromoSection/index.js#L16-L32

vs.

https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/TrySection/index.js#L7-L22

vs.

https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/Subscribe/index.js#L7-L20

not to mention lots and lots of styled components – Perhaps the easiest to begin with being the ones inside pages/*.js files.

jorgeorpinel:
> not to mention lots and lots of styled components

Example (exact copy):

https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/TrySection/index.js#L29-L36
vs.
https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/PromoSection/index.js#L34-L41

Example (very similar, could use `props` for differences):

https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/TrySection/index.js#L102-L128
vs.
https://github.com/iterative/dvc.org/blob/4a849f56444a326c6d1e4255d5f556fc35b08b52/src/Subscribe/index.js#L53-L83
> Also detected `const Title` and `const Button` in src/TrySection vs. the ones in src/PromoSection.

jorgeorpinel:
cc @iAdramelk input please 🙂

iAdramelk:
i think that a lot of existing components can be refactored/merged. But can't say beforehand which exactly should, examples above was created before me, and I'm not really familiar with them. In some cases some code repetition that add readability is better code without repetition, but which is harder to read.

jorgeorpinel:
Thanks for the feedback Alexey. Mainly wanted to know whether others who have more familiarity with our React code agree with me on this (and seems like you do, in general).

> In some cases some code repetition that add readability is better code without repetition, but which is harder to read.

[...]

PR #969, opened 2020-02-04
Code cleanup

iAdramelk:
We are preparing to migrate from `nextjs` to `gatsby`, and to make migration easier, I made some code cleanup. Cleanup goals were to make folder structure more comfortable to use with gatsby, separate next-specific code from react components and remove as much `next.js` stuff that we didn't need as possible.

There is a lot of changes. To make it easier to review and read, I tried to group different tasks to different commits, so it will probably be easier to read it commit by commit then to read full diff.

What have been done:

- 7748c863 Move react components to components folder. Gatsby stored pages in the `src/pages` folder, but we are storing our components here too. To make it easier to navigate and distinguish between them, I moved components to the subfolder `src/components`.
- fa371db9 and 7bf941f0  Consistently rename components and styles files. Most of our components are named `Component/index.js`, but some older components are named `Component/Component.js`, I renamed all such components to the newer format.
- 9258f529, 57b33bfa, 81cdee89, 50beb010 - I split pages to the two separate components - one responsible for render and visual styles (moved to `src/components`) and one responsible for routing, header data and getting initial props from API (left in the `pages` folder). Because I will need to refactor pages logic from `nextjs` to `gatsby`, it will be easier to do if there is no presentational logic in them. 
- 59e399dc Move styled-components to separate files. After we migrate to `gatsby`, we wouldn't need `styled-components` anymore. To make replacing them easier in the future, I moved them all to separate files with the same logic as was used on the community page.
- 74c65978 Refactor NextLink and NextRouter. Here I create a standard link component and use it instead of the `next/link` in our files; this way, we will only need to replace for `gatsby-link` in one place. Also, here I removed some cases of `next/router` usage that we don't need.


jorgeorpinel:
Thanks @iAdramelk. Massive but greatly explained PR!

> Move styled-components to separate files

~~Nice. Should we also rename all the styles.js files to styled.js (per a conversation had some time ago).~~ (Never mind, see next point.)
Also, do you think the issues below are still valid now?
- _glossary: soft code style constants in tooltip file_ #658
- _js: refactor repetitive React components_ #777

> After we migrate to gatsby, we wouldn't need styled-components anymore

Let's try remembering to close _css: migrate from styled comps to plan css or css-modules_ #792 when that happens.

Thanks!

iAdramelk:
@jorgeorpinel Thanks! I think that #658 and #777 are still valid, but I'd place them on hold for now, before we finish with Gatsby and decide where to migrate styles.

PR #970, opened 2020-02-04
refactor: move some files around, fix loose ends and fix broken redirect

fabiosantoscode:
Fixing some loose ends with #943, addressing @jorgeorpinel and @shcheklein's feedback. Also fixed the troubleshooting redirect URL which was missing the `/doc` prefix.

-------

Please chose to
[allow us](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
to edit your branch when creating the PR.

### DIFF
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -1,7 +1,7 @@
 [
   "^https://(?:www\\.)dvc\\.org/(.+)?       https://dvc.org/$1",
   "^https://man\\.dvc\\.org/(.+)?           https://dvc.org/doc/command-reference/$1 303",
-  "^https://error\\.dvc\\.org/(.+)?         https://dvc.org/user-guide/troubleshooting#$1 303",
+  "^https://error\\.dvc\\.org/(.+)?         https://dvc.org/doc/user-guide/troubleshooting#$1 303",
   "^https://(code|data|remote)\\.dvc\\.org/(.+) https://s3-us-east-2.amazonaws.com/dvc-public/$1/$2 303",
   "^/((?:deb|rpm)/.+)                       https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                       https://discordapp.com/invite/dvwXA2N 303",
--- a/scripts/exclude-links.txt
+++ b/scripts/exclude-links.txt
@@ -15,7 +15,6 @@ https://github.com/example/registry
 https://github.com/iterative/dvc.org/blob/master/public$
 https://github.com/iterative/dvc/releases/download/$
 https://github.com/myaccount/myproject.git
-https://dvc.org/user-guide/troubleshooting
 https://myendpoint.com
 https://object-storage.example.com
 https://www.youtube.com/embed/$
--- a/scripts/exclude-links.txt
+++ b/scripts/exclude-links.txt
@@ -15,6 +15,7 @@ https://github.com/example/registry
 https://github.com/iterative/dvc.org/blob/master/public$
 https://github.com/iterative/dvc/releases/download/$
 https://github.com/myaccount/myproject.git
+https://dvc.org/user-guide/troubleshooting
 https://myendpoint.com
 https://object-storage.example.com
 https://www.youtube.com/embed/$
--- a/src/utils/redirects.js
+++ b/src/utils/redirects.js
@@ -1,6 +1,6 @@
 /* eslint-env node */
 
-let redirects = require('./redirects-list.json')
+let redirects = require('../../redirects-list.json')
 
 const processRedirectString = redirectString => {
   let [regex, replace, code = 301] = redirectString.split(/\s+/g)
--- a/scripts/link-check.sh
+++ b/scripts/link-check.sh
@@ -46,7 +46,7 @@ fails=0
 for file in "$@"; do
   echo -n "$file:"
   prev=$fails
-  checker $(finder "$file" | sort -u | comm -23 - <(echo "$exclude" | sort -u)) || fails=$(($fails + 1))
+  checker $(finder "$file" | sed 's/#.*//g' | sort -u | comm -23 - <(echo "$exclude" | sort -u)) || fails=$(($fails + 1))
   [ $prev -eq $fails ] && echo OK
 done
 [ $fails -eq 0 ] || echo -e "ERROR:$fails failures\n---" >&2
--- a/server.js
+++ b/server.js
@@ -14,59 +14,56 @@ const { parse } = require('url')
 const next = require('next')
 
 const { getItemByPath } = require('./src/utils/sidebar')
-const { getRedirect } = require('./src/redirects.js')
+const { getRedirect } = require('./src/utils/redirects')
 
 const dev = process.env.NODE_ENV !== 'production'
 const app = next({ dev })
 const handle = app.getRequestHandler()
 const port = process.env.PORT || 3000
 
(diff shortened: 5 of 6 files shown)


fabiosantoscode:
`link-check-git-diff` is failing because I've removed the broken redirects, and it thinks those were links that worked.

redirects-list.json
```diff
@@ -1,7 +1,7 @@
 [
   "^https://(?:www\\.)dvc\\.org/(.+)?       https://dvc.org/$1",
   "^https://man\\.dvc\\.org/(.+)?           https://dvc.org/doc/command-reference/$1 303",
-  "^https://error\\.dvc\\.org/(.+)?         https://dvc.org/user-guide/troubleshooting#$1 303",
+  "^https://error\\.dvc\\.org/(.+)?         https://dvc.org/doc/user-guide/troubleshooting#$1 303",
```
shcheklein: good catch!

shcheklein:
looks good!

@fabiosantoscode please add those failed links to the exclusion list please.

fabiosantoscode:
Cool! I've also used `sed` to remove the hash part, instead of adding three links which differ only by hash but are the same document.

scripts/exclude-links.txt
```diff
@@ -15,6 +15,7 @@ https://github.com/example/registry
 https://github.com/iterative/dvc.org/blob/master/public$
 https://github.com/iterative/dvc/releases/download/$
 https://github.com/myaccount/myproject.git
+https://dvc.org/user-guide/troubleshooting
```
shcheklein: so, where does this link come from? I don't quite understand

shcheklein: removing this for now and merging the PR

redirects-list.json
jorgeorpinel: Oops. Maybe we should try to thoroughly test redirects manually (sanity check) even if we have tests because manually you can come up with ways to try to break the logic 😋  This is because publishing incorrect redirects (especially 301) causes almost permanent "damage" to new users browser cache. Just a thought

jorgeorpinel: > Needless to say I'm available to do this but just need to prioritize as I've been struggling with switching from task to task recently.

src/utils/redirects.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-env node */
 
-let redirects = require('./redirects-list.json')
+let redirects = require('../../redirects-list.json')
```
jorgeorpinel: I know Ivan suggested this but I'm not getting why this doesn't go in src/util/ along with the redirect module and tests.

jorgeorpinel: Never mind, I guess it makes sense to have it near server.js

shcheklein:
@jorgeorpinel because it's website global config. It's a regular thing to have some redirects file in the root. e.g. Netlify, as far as I remember is just one of the examples.

shcheklein:
> I know this is merged but for the record I approve haha. I do have some JS code suggestions thought, may open a separate PR or include in my current regular updates one (#966)

please, if it's possible let's not do this unless they are just stylistic changes (like reoder ifs, or something) that are not directly related to this PR. Otherwise, please leave some comments here - we'll try to address this while we still have context.

jorgeorpinel:
> let's not do this unless they are just stylistic changes

They are small refactors and language/stylistic changes indeed. They don't change the behavior. I can open a sample draft PR so you guys see them... ⏳ 

Also, I do still find the space separation in the JSON file strange. Why not instead of:

```json
[
  "^https://(?:www\\.)dvc\\.org/(.+)?       https://dvc.org/$1",
  ...
```
do something like
```json
[
  ["^https://(?:www\\.)dvc\\.org/(.+)?", "https://dvc.org/$1", '301'],
  ...
```
?

jorgeorpinel:
I thin think is Ivan's logic on the redirects JSON file:

> From https://github.com/iterative/dvc.org/pull/943#issuecomment-582054859
> we had an iteration like this and I didn't like it (not sure about @fabiosantoscode :) ) - it's 1000 lines vs 10. It's way easier to read line by line. It's very regular format for redirects files - one rule a line. There should be very strong reasons to complicate this.

So why use JSON at all? Probably a regular redirect file is just plan text? Even simpler. Up to you guys, no strong opinion as long as it works well.

jorgeorpinel:
> it's 1000 lines vs 10

Not sure how though. See my suggestion in https://github.com/iterative/dvc.org/pull/970#issuecomment-582057203 above.

I won't touch this file in my small refactoring suggestion though. Let's see what Fabio prefers 🙂

shcheklein:
> They are small refactors and language/stylistic changes indeed. They don't change the behavior. I can open a sample draft PR so you guys see them... ⏳

if they are not related to this PR at all - include them to the regular updates, up to you!

> So why use JSON at all?

simpler to deal with in JS

> do something like `array of fields`

one line array instead of a string might work fine as well, don't have a strong opinion here

shcheklein:
> one line array instead of a string might work fine as well, don't have a strong opinion here 

would probably still personally prefer a simple string, just don't see a reason to write all these additional symbols in this case :)

jorgeorpinel:
> if they are not related to this PR at all - include them to the regular updates, up to you!

They're related, see https://github.com/iterative/dvc.org/pull/973/files

Thanks

fabiosantoscode:
@jorgeorpinel I've added the spaces because it was pretty hard to read the big-picture of the file without them. I'm happy to remove if you feel strongly against them.

I also think arrays would be fine. Maybe the space alignment would be more natural in that situation, or we could do something like this to make it more readable, even if it's taller:

```
[
  [
    "/from",
    "/to",
    308
  ],
[...]
```

jorgeorpinel:
In my case my IDE wraps at 80 chars so taller is easier to read in cases like this. ESLint would probably reformat it that way anyway. Again, it's not a strong opinion but I do think it's a better format and less error prone for later edits (how many spaces should I add? space typos, etc.)

shcheklein:
still prefer compact vs very tall - it's just hard to see the whole picture when it's tall, means hard to catch errors. In this case you just see clearly one block that looks exactly like:

```
string     string
string     string
string     string
```

it's super easy to read and see all at once. Especially if you start grouping them by the first string.

jorgeorpinel:
This is how it looks for me 😋 

image

13" laptop screen, and clearly I like large fonts 😬

jorgeorpinel:
BTW lines 4 and 5 don't have the same space alignment on the 2nd column:

```json
  "^https://error\\.dvc\\.org/(.+)?         https://dvc.org/doc/user-guide/troubleshooting#$1 303",
  "^https://(code|data|remote)\\.dvc\\.org/(.+) https://s3-us-east-2.amazonaws.com/dvc-public/$1/$2 303",
```

shcheklein:
I'm not sure 100% but my intuition is that you don't set auto-wrapping for code related stuff - it's extremely dangerous.

shcheklein:
also, you at least have a chance to make it wider for a moment, in case it's long - no one has a chance whatsoever to see it all at once, group it, etc

jorgeorpinel:
editor line wrapping = extremely dangerous? 😱 well I need some adrenaline in life

Didn't quite get that last comment about "see it all at once, group it, etc" but OK, those are all my arguments and still my position isn't very strong about this, plus I'm not convincing anyone, so let's just move on 🙂

PR #973, opened 2020-02-04
server: reformat/refactor around redirects

jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/970#pullrequestreview-353200254

For demo of suggested changes. (Not sure whether we'll move to merge these changes here.)


server.js
```diff
@@ -23,23 +24,21 @@ const port = process.env.PORT || 3000
...
-      // should be getting the query as a string
-      const { query } = parse(req.url)
-      if (query) {
-        redirectLocation += '?' + query
+      /*
+       * HTTP redirects
+       */
+
+      if (queryStr) {
+        redirectLocation += '?' + queryStr
```
jorgeorpinel: No need to call `url.parse` twice because we can `querystring.parse` the query string instead.
Same result

server.js
```diff
@@ -57,9 +56,12 @@ app.prepare().then(() => {
       }
 
       // Custom route for all docs
-      app.render(req, res, '/doc', query)
+      app.render(req, res, '/doc', parseQuery(queryStr))
```
shcheklein: was there a bug before? not sure I get this change

server.js
```diff
@@ -57,9 +56,12 @@ app.prepare().then(() => {
...
     } else {
-      // Regular Next.js handler
+      /*
+       * Regular Next.js handler
```
shcheklein: I would actually use single line comments everywhere when it is a single line

src/utils/redirects.test.js
```diff
@@ -57,18 +57,17 @@ describe('getRedirect', () => {
...
+      // Detect redirect loops.
       const secondUrl = url.parse(addHost(rLocation))
       const secondRedirect = getRedirect(secondUrl.hostname, secondUrl.pathname)
       expect(secondRedirect).toEqual([])
     })
   }
 
-  describe('host redirects', () => {
-    // remove the www (when already HTTPS)
+  describe('Subdomain-based redirects', () => {
```
shcheklein: minor: I think it's a convention to keep it one phrase long and start with lower case

src/utils/redirects.test.js
```diff
@@ -126,22 +125,23 @@ describe('getRedirect', () => {
...
 
-  describe('discord', () => {
+  describe('Discord', () => {
     itRedirects('/help', 'https://discordapp.com/invite/dvwXA2N', 303)
 
     itRedirects('/chat', 'https://discordapp.com/invite/dvwXA2N', 303)
   })
 
-  describe('in-site moves', () => {
+  describe('Path redirects', () => {
```
shcheklein: same here

shcheklein:
minor questions, otherwise looks good

server.js
jorgeorpinel: minor refactor: it was calling `url.parse` twice because the first one sent `true` as the 2nd param to make the `query` part of the returned value into an object, but we could simply reuse that one call and parse just the query from string to query using `querystring.parse` (used internally in `url.parse` anyway). Same result mentioned in https://github.com/iterative/dvc.org/pull/973#pullrequestreview-353231202 above

shcheklein: don't get it, sorry :(

so, `queryStr` has the same value as `query`? not sure still why do we call `parseQuery` on top of it?

shcheklein:

server.js
jorgeorpinel: > queryStr has the same value as query

No, `queryStr` is a string, `query` is an object. `query`  = `parseQuery(queryStr)`

> See also my previous https://github.com/iterative/dvc.org/pull/973#pullrequestreview-353231202 above.

server.js
jorgeorpinel: OK, updated.

src/utils/redirects.test.js
jorgeorpinel: Not sure but OK, using cammelCase convention now for `describe()` calls. Please check.

src/utils/redirects.test.js
jorgeorpinel: Same.

jorgeorpinel:
Sorry for the delay. Addressed your feedback now, @shcheklein.

server.js
shcheklein: k, I see, thanks @jorgeorpinel !

jorgeorpinel:
This was reverted in 822adad because dvc.org was crashing after merge/deploy.

Trying again in #1009...

PR #981, opened 2020-02-07
Add sentry

iAdramelk:
Test integration with Sentry. To make it work we need to add env variable `SENTRY_DSN` to Heroku.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,0 +1,14 @@
+/* eslint-env node */
+
+const withSourceMaps = require('@zeit/next-source-maps')
+
+const settings = withSourceMaps({
+  webpack(config) {
+    return config
+  },
+  env: {
+    SENTRY_DSN: process.env.SENTRY_DSN
+  }
+})
+
+module.exports = settings
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -1,0 +1,41 @@
+/* global process */
+
+import React from 'react'
+import App from 'next/app'
+import * as Sentry from '@sentry/browser'
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN
+})
+
+class MyApp extends App {
+  static async getInitialProps({ Component, ctx }) {
+    let pageProps = {}
+
+    if (Component.getInitialProps) {
+      pageProps = await Component.getInitialProps(ctx)
+    }
+
+    return { pageProps }
+  }
+
+  componentDidCatch(error, errorInfo) {
+    Sentry.withScope(scope => {
+      Object.keys(errorInfo).forEach(key => {
+        scope.setExtra(key, errorInfo[key])
+      })
+
+      Sentry.captureException(error)
+    })
+
+    super.componentDidCatch(error, errorInfo)
+  }
+
+  render() {
+    const { Component, pageProps } = this.props
+
+    return <Component {...pageProps} />
+  }
+}
+
+export default MyApp
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -1,6 +1,17 @@
+/* eslint-env node */
+
 import React from 'react'
 import Document, { Head, Main, NextScript } from 'next/document'
 import { ServerStyleSheet } from 'styled-components'
+import * as Sentry from '@sentry/browser'
+
+process.on('unhandledRejection', err => {
+  Sentry.captureException(err)
+})
+
+process.on('uncaughtException', err => {
+  Sentry.captureException(err)
+})
 
 import {
   META_BASE_TITLE,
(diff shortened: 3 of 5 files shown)


jorgeorpinel:
Interesting. So Sentry is a SRE/APM dashboard? I quickly read https://blog.sentry.io/2018/03/06/the-sentry-workflow but would be great to know what this is all about 🙂

iAdramelk:
@jorgeorpinel , yep. Pretty much it. We can collect info about user's javascirpt errors with it, see information about environment, stack traces, etc.

pages/_app.js
```diff
@@ -0,0 +1,41 @@
...
+import * as Sentry from '@sentry/browser'
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN
+})
+
+class MyApp extends App {
+  static async getInitialProps({ Component, ctx }) {
+    let pageProps = {}
+
```
jorgeorpinel: Where is this app used?

next.config.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-env node */
+
+const withSourceMaps = require('@zeit/next-source-maps')
```
jorgeorpinel: I see `@zeit/next-source-maps` was also added here. Is this specifically just a nice addition for Sentry logs/traces?

pages/_app.js
```diff
@@ -0,0 +1,41 @@
...
+import React from 'react'
+import App from 'next/app'
+import * as Sentry from '@sentry/browser'
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN
+})
+
+class MyApp extends App {
```
jorgeorpinel: Where is this app used?

pages/_document.js
```diff
@@ -1,6 +1,17 @@
...
 import { ServerStyleSheet } from 'styled-components'
+import * as Sentry from '@sentry/browser'
+
+process.on('unhandledRejection', err => {
+  Sentry.captureException(err)
+})
+
+process.on('uncaughtException', err => {
+  Sentry.captureException(err)
+})
```
jorgeorpinel: Why are these needed if Sentry supposedly has automated handling of uncaught exceptions and unhandled rejections? See https://docs.sentry.io/platforms/javascript/#automatically-capturing-errors

iAdramelk: It's for server side errors in node.js, Sentry will automatically log errors in browser, but not unhandled errors in SSR.

pages/_app.js
iAdramelk: It's replacement of the standard Next.js component https://nextjs.org/docs/advanced-features/custom-app . Basically it's a wrapper around the pages.

next.config.js
iAdramelk: Source maps is a mechanism to map bundled and minified code with the original source files. If we just run sentry as it is, it will show errors like: `You nave an error on line 1, symbol 3454560 in the gX method.` Source maps will ideally show us original file, line and method name. (But right now it's looks like it didn't work correctly)

It's not strictly related to Sentry and is used for example by browser Dev Tools, but it make using Sentry more convenient.

pages/_document.js
jorgeorpinel: Got it, thanks.

Would these also be needed separately for server.js or other build-time code? e.g. sidebar.js

next.config.js
jorgeorpinel: Right, I did read about source maps and it seems like a great feature. I was just wondering if this was added specifically for Sentry traces or whether it can also be used on the Browser now.

iAdramelk: Well, we should be, because I leave them enabled (there is also an option to send them directly to sentry through API and not publish on site). But in practice from that I see, I misconfigured them somehow and didn't see them anywhere.

pages/_document.js
iAdramelk: This catches all code in the actual pages including `sidebar.js`, but I think we are now missing errors from server.js because they happens before we call `_document.js`.

next.config.js
jorgeorpinel: OK cool please let us know when you figure it out 🙂

pages/_document.js
jorgeorpinel: Yes some server code runs before the Next app and partially includes sidebar.js code I think. That's what I was wondering about. But it's probably not a big deal? Only runs when the server is being initialized, probably only when new versions are deployed to Heroku.

iAdramelk: That's pretty much my thoughts as well. And also the fact that we are planning to migrate to Gatsby and it didn't have a server.  So I decided to leave it as is for now.

PR #985, opened 2020-02-10
Revert "Revert "Add sentry""

iAdramelk:
First attempt was reverted because of the bug, so this a new one for testing and debug.


iAdramelk:
@shcheklein Ok,  I think I fixed it https://dvc-landing-sentry-inte-ymkwh6.herokuapp.com/doc

Problem was that I put `@zeit/next-source-maps` inside  `devDependencies` instead of `dependencies`. Moving it fixed error.

Issue #731, opened 2019-10-24
docs: support items without nav sidebar entry


dashohoxha:
In some cases it might be useful to hide (not-show) a page on the left-side menu.
For example in the "Installation" page, the pages about "completion" and "plugins" are referenced from the index page of installation, but I wish they were not shown on the sidebar menu.

Currently this is not possible (unless I have missed something).
This could be specified for example with a field like `"label": false,` or `"label": null,` or `"label": "",`

jorgeorpinel:
Seems to me like a nice-to-have feature: Having the files there but with no entry in sidebar.json

However I don't agree that neither `/doc/user-guide/install/completion` nor `...plugins` should be hidden from the sidebar. They help users discover those topics by navigating the side bar. In fact can't find any element in the navigation at the moment that shouldn't be listed.

jorgeorpinel:
p.s. The problem with those 2 entries is that they are repeated as stand-alone User Guides since #695 was merged.

jorgeorpinel:
p.s. The problem with those 2 entries is that they are repeated as stand-alone User Guides since #695 was merged. This is reported in #701.

shcheklein:
Not sure I see any valuable use case for this for now. Closing this. Will revisit if have a real use case for this.

jorgeorpinel:
BTW this is now possible on Gatsby I think. Are there any pages we want to hide from the docs @shcheklein? Maybe the Google API Privacy Policy (per https://github.com/iterative/dvc.org/issues/144#issuecomment-584441916)

rogermparent:
You're correct, this has been possible since the Models PR. All MD source files will become pages, and if they aren't added to the sidebar they won't show up automatically, fulfilling this behavior.
Should the docs implementation change (and it probably will eventually), I'll keep this feature in mind when changing anything as to not break it so you should be safe to use it.

PR #987, opened 2020-02-10
more GAPI privacy policy updates

jorgeorpinel:


public/static/docs/user-guide/privacy.md
```diff
@@ -16,13 +15,16 @@ service. See
...
 
-DVC access Google resources from your machine you run DVC on. Your machine
+DVC accesses Google resources from the machine where you use DVC. Your machine
 communicates directly with the Google APIs.
 
-DVC project never receives your data or the permission to access your data. The
-owners of the project can only see anonymous, aggregated information about usage
-of tokens obtained through its OAuth client, such as which APIs and endpoints
-are being used.
+The _DVC_ application never receives your data or the permission to access your
```
jorgeorpinel: @MaxRis I changed "DVC project" for "_DVC_ application" but I'm not sure which one is which. Is there a GCP project called DVC? Or is it just a Gogle application that's called that? Can you share some links to all and any of these entities so I can express it here clearly? Thanks

jorgeorpinel: > Can you share some links to all and any of these entities

Maybe read access to my Google account if you think it could be useful (jorge@orpinel.com)

jorgeorpinel:
Another question @shcheklein. Shouldn't the privacy policy be a root section of its own, instead of buried inside User Guide?

shcheklein:
@jorgeorpinel don't see a big reason for this, it's specific to GDrive for now.

public/static/docs/user-guide/privacy.md
MaxRis: @jorgeorpinel I afraid you should be invited by @shcheklein to have access to the GCP project.
So, yes, we are talking here about GCP project and `DVC` is a title for this project given through OAuth Consent Screen. `DVC` name is visible to the users, but we are explaining that GCP project behind doesn't receive access to any users' data.

jorgeorpinel:
I agree. But it's also not a user guide. It wold be nice to be able to not have it in the sidebar I guess, just linked from some place... Maybe work on #731 ?

BTW, should re remove the current link from the footer then? And put it in the remote add and remote modify cmd refs.

public/static/docs/user-guide/privacy.md
jorgeorpinel: So the right term is "DVC GCP Project" right? There's no _DVC_ application? Thanks

MaxRis: Just "DVC Application" would mean dvc command line application itself, I believe. Yes, "DVC GCP Project" should be fine to go with.

shcheklein:
@jorgeorpinel don't see it as a big deal right now, let's get back to this discussion when we start the User Guide restructuring.

shcheklein:
@jorgeorpinel footer link was by the Google verification team.

public/static/docs/user-guide/privacy.md
shcheklein: DVC GCP Project sounds too complicated. The right term is DVC application I believe. This is the one we get secret/token for and the one used to authenticate (generate credentials) for users.

shcheklein: in this specific case DVC project was the right phrase though. We refer to the DVC team, etc.

shcheklein: So, I would keep it the way it was before probably.

jorgeorpinel: > Just "DVC Application" would mean dvc command line application itself 

Exactly. I'm trying to avoid confusions. Alright then, I changed all the instances of _DVC_ application for DVC GCP Project in . Thanks Max!

jorgeorpinel: Oops I just noticed your comments here @shcheklein 

> in this specific case DVC project was the right phrase though. We refer to the DVC team

I don't think that's the case, this refers to the dvc project or app on GCP which we own, right? See, this is the confusion I'm trying to clear up. If you could five me read access to the GCP I could figure it out on my own.

public/static/docs/user-guide/privacy.md
```diff
@@ -1,13 +1,12 @@
...
-details on how to setup it check the respective sections in the `dvc remote add`
-and `dvc remote modify` commands as well as the relevant Google Drive section of
-this User Guide.
+make it easier to store files on Google Drive. For more details on how to set it
+up, refer to
+[Setup a Google Drive DVC Remote](/doc/user-guide/setup-google-drive-remote)
 
-DVC uses the _DVC_ application on Google Cloud Platform. That is the name you
-will see in a consent screen. DVC integration is governed by common policies
+DVC uses the _DVC_ Project on Google Cloud Platform. (That is the name you will
```
shcheklein: why is it upper case, btw?

public/static/docs/user-guide/privacy.md
shcheklein: kk. still looks like an overkill DVC GCP Project sounds way too complicated. DVC project is enough in either case. You can make _DVC project_ everywhere (but we already use _DVC_ to refer to the GCP app/project).

(Also, most likely the right term is application, not project from the Google console perspection:

Screen Shot 2020-02-10 at 4 24 59 PM
)

jorgeorpinel: > still looks like an overkill DVC GCP Project... DVC project is enough in either case

Yes, I don't disagree with that. But DVC project is a term we have in the glossary also so Idk... Unless we move to stop using that term in favor of DVC repo always – please see https://github.com/iterative/dvc.org/pull/931#discussion_r376181149 BTW, been trying to have your attention to that question haha.

> most likely the right term is application

Perfect, I'm using "_DVC_ Application" then! Done in 9e14575a

public/static/docs/user-guide/privacy.md
jorgeorpinel: It's Application now BTW. Upper case because it's a special kind of app (a GCP app, used to setup GAPI integrations) vs the general concept of software applications. The distinction is useful, for example, in the **Usage in other packages or applications** section.

shcheklein: 👍

public/static/docs/user-guide/privacy.md
shcheklein: 👍

jorgeorpinel:
> don't see it as a big deal right now, let's get back to this discussion when we start the User Guide restructuring

OK, noted in https://github.com/iterative/dvc.org/issues/144#issuecomment-584441916

> footer link was (required) by the Google verification team

OK, can I rename it "Google APIs Privacy Policy" just for correctness? Typically a link like this on the footer is a general privacy policy for data collection of DVC itself, not specific to the integration in this case. But we don't really collect any personal data, so its a bit misleading.

shcheklein:
> OK, can I rename it "Google APIs Privacy Policy" just for correctness?

Let's not mess with this right now. We can think how to generalize, improve this, etc later, way later.

public/static/docs/sidebar.json
```diff
@@ -162,7 +162,7 @@
         "slug": "analytics"
       },
       {
-        "label": "Privacy Policy",
+        "label": "Privacy Policy (Google APIs)",
```
jorgeorpinel: > it's probably too long to fit into one line?

@shcheklein nope, it fits 🙂 but that's the reason I shortened the docs title actually.

image

jorgeorpinel:
OK, I renamed the PR, merging!

PR #991, opened 2020-02-13
Replace custom collapse in sidemenu

### LINKED ISSUE #724: nav: unnecessary gaps in sidebar

### LINKED ISSUE #856: nav: left sidebar isn't visible without js
~~Disclaimer: I'm not sure the title is completely accurate.~~ See https://github.com/iterative/dvc.org/pull/834#pullrequestreview-330780455 for context Summary:

> J: without js the nav bar doesn't expand
> I: The good question is if we need to generate a sitemap or something for the case when we don't have a link to some page inside and search engines won't be able to find it?
> J: maybe it's easy to pre-render nav?

So, possible solutions:

1. Pre-render all nav elements
2. Introduce a (automatic) site map.

iAdramelk:
As part of community page I updated `react-collapse` to version 5.0 and it so happens this version can be also used in side menu and is fixing some of our existing issues.

Fix #856 
Fix #724

### DIFF
--- a/src/components/Documentation/SidebarMenu/index.js
+++ b/src/components/Documentation/SidebarMenu/index.js
@@ -1,4 +1,5 @@
 import React, { useEffect, useRef, useState } from 'react'
+import { Collapse } from 'react-collapse'
 import PerfectScrollbar from 'perfect-scrollbar'
 import scrollIntoView from 'dom-scroll-into-view'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


shcheklein:
This looks awesome!

shcheklein:
@fabiosantoscode could you review this please?

PR #993, opened 2020-02-14
Regular updates (Feb 13)

### LINKED ISSUE #719: term: don't use "take X under control" phrase
Starting with https://dvc.org/doc/command-reference/add, per https://github.com/iterative/dvc/pull/2546#discussion_r337122908 (look at resolution there for final wording).

And look for any other instances.

jorgeorpinel:
- Misc. stuff
- Closes #719 (core repo match: iterative/dvc/pull/3322)
- Reflect iterative/dvc/pull/3312

### DIFF
--- a/public/static/docs/command-reference/config.md
+++ b/public/static/docs/command-reference/config.md
@@ -22,9 +22,9 @@ takes a config option `name` (a section and a key, separated by a dot) and its
 This command reads and updates the DVC configuration files. By default (if none
 of `--local`, `--global`, or `--system` is provided) a project's config
 (`.dvc/config`) file is read or modified. This file is by default meant to be
-under Git control and should not contain sensitive and/or user-specific
-information (passwords, SSH keys, etc). Use `--local`, `--global`, or `--system`
-options instead to override project's settings, for sensitive, or user-specific
+tracked by Git and should not contain sensitive and/or user-specific information
+(passwords, SSH keys, etc). Use `--local`, `--global`, or `--system` options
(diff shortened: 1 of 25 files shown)


public/static/docs/command-reference/config.md
```diff
@@ -22,9 +22,9 @@ takes a config option `name` (a section and a key, separated by a dot) and its
 This command reads and updates the DVC configuration files. By default (if none
 of `--local`, `--global`, or `--system` is provided) a project's config
 (`.dvc/config`) file is read or modified. This file is by default meant to be
-under Git control and should not contain sensitive and/or user-specific
-information (passwords, SSH keys, etc). Use `--local`, `--global`, or `--system`
-options instead to override project's settings, for sensitive, or user-specific
+tracked by Git and should not contain sensitive and/or user-specific information
```
jorgeorpinel: Not sure about "tracked by Git" but at least has ~20k results on Google, vs ~4k of "under Git control"

jorgeorpinel: I guess it's fine? "track"-like words are common in https://git-scm.com/docs/git-add (whereas no instances of "control" are found).

shcheklein: both fine with, I would not try to be very formal about this ... every formality like this should have a strong reason/ground ... in this case it's quite easy to understand both I would say

jorgeorpinel: Its not about formality in this case. "Under X control" sounds strange, as discussed in #719

PR #995, opened 2020-02-14
Move API calls on community page to client side

iAdramelk:
Right now on community we fetch data from APIs inside `getInitialProps` it locks page render until we get all responses and also prevents us from caching renders results for this page as static html.

This PR moves these requests to the client side.

### DIFF
--- a/src/components/Community/Learn/index.js
+++ b/src/components/Community/Learn/index.js
@@ -10,6 +10,7 @@ import CommunityBlock from '../Block'
 import CommunityButton from '../Button'
 import CommunitySection from '../Section'
 
+import { getLatestPosts, getCommentsCount } from '../../../utils/api'
 import { pluralizeComments } from '../../../utils/i18n'
 
 import {
@@ -53,9 +54,7 @@ function CommunityBlogPost({
 
   useEffect(() => {
     if (commentsUrl) {
-      fetch(`/api/comments?url=${commentsUrl}`)
-        .then(result => result.json())
-        .then(data => setCount(data.count))
+      getCommentsCount(commentsUrl).then(result => setCount(result))
     }
   }, [])
 
@@ -185,7 +184,17 @@ CommunityDocumentation.propTypes = {
   url: PropTypes.string
 }
 
-export default function CommunityLearn({ posts, theme }) {
+export default function CommunityLearn({ theme }) {
+  const [isPostsLoaded, setIsPostsLoaded] = useState(false)
+  const [posts, setPosts] = useState([])
+
+  useEffect(() => {
+    getLatestPosts().then(result => {
+      setPosts(result)
+      setIsPostsLoaded(true)
+    })
+  }, [])
+
   return (
     <Wrapper>
       <CommunitySection
@@ -255,15 +264,17 @@ export default function CommunityLearn({ posts, theme }) {
                 )
               }
             >
-              {posts.length ? (
+              {!isPostsLoaded && <Placeholder>Loading...</Placeholder>}
+              {isPostsLoaded &&
+                !!posts.length &&
                 posts.map(post => (
                   <CommunityBlogPost
                     {...post}
                     key={post.url}
                     color={theme.color}
                   />
-                ))
-              ) : (
+                ))}
+              {isPostsLoaded && !posts.length && (
                 <Placeholder>Blog is unavailable right now</Placeholder>
               )}
             </CommunityBlock>
@@ -286,7 +297,6 @@ export default function CommunityLearn({ posts, theme }) {
 }
 
 CommunityLearn.propTypes = {
-  posts: PropTypes.array,
   theme: PropTypes.shape({
     backgroundColor: PropTypes.string,
     color: PropTypes.string
--- a/src/components/Community/Learn/index.js
+++ b/src/components/Community/Learn/index.js
@@ -251,7 +251,7 @@ export default function CommunityLearn({ theme }) {
             >
               {!ready && <Placeholder>Loading...</Placeholder>}
               {error && (
-                <Placeholder>Blog is unavailable right now</Placeholder>
+                <Placeholder>Blog unavailable right now</Placeholder>
               )}
               {posts &&
                 posts.map(post => (
--- a/src/components/Community/Learn/index.js
+++ b/src/components/Community/Learn/index.js
@@ -250,9 +250,7 @@ export default function CommunityLearn({ theme }) {
               }
             >
               {!ready && <Placeholder>Loading...</Placeholder>}
-              {error && (
-                <Placeholder>Blog unavailable right now</Placeholder>
-              )}
+              {error && <Placeholder>Blog unavailable right now</Placeholder>}
               {posts &&
                 posts.map(post => (
                   <CommunityBlogPost
--- a/src/components/Community/Learn/index.js
+++ b/src/components/Community/Learn/index.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback } from 'react'
 import PropTypes from 'prop-types'
 import format from 'date-fns/format'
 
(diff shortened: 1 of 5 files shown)


shcheklein:
@fabiosantoscode PTAL :)

src/utils/api.js
```diff
@@ -54,3 +53,16 @@ export async function getLatestTopics(req) {
...
+export async function getCommentsCount(commentsUrl) {
+  try {
+    const res = await fetch(`/api/comments?url=${commentsUrl}`)
+
+    if (res.status === 200) {
+      const { count } = await res.json()
+      return count
+    }
+  } catch (e) {
+    console.error(e)
```
fabiosantoscode: Out of scope: maybe we should reject the promise when there's a non-200 response?

src/components/Community/Meet/index.js
```diff
@@ -231,15 +251,17 @@ export default function CommunityMeet({ issues, theme, topics }) {
...
                 issues.map(issue => (
                   <CommunityIssue
                     {...issue}
                     key={issue.url}
                     color={theme.color}
                   />
-                ))
-              ) : (
+                ))}
+              {isIssuesLoaded && !issues.lengts && (
```
fabiosantoscode: Typo here

src/components/Community/Meet/index.js
```diff
@@ -110,7 +111,24 @@ CommunityIssue.propTypes = {
...
+
+  useEffect(() => {
+    getLatestIssues().then(result => {
+      setIssues(result)
+      setIsIssuesLoaded(true)
+    })
+    getLatestTopics().then(result => {
+      setTopics(result)
+      setIsTopicsLoaded(true)
+    })
```
fabiosantoscode: Nit: this doesn't stop the set calls when the component has already unmounted. It's not a big deal but it generates a warning in the console.

src/components/Community/Meet/index.js
```diff
@@ -110,7 +111,24 @@ CommunityIssue.propTypes = {
   color: PropTypes.string
 }
 
-export default function CommunityMeet({ issues, theme, topics }) {
+export default function CommunityMeet({ theme }) {
+  const [isIssuesLoaded, setIsIssuesLoaded] = useState(false)
+  const [issues, setIssues] = useState([])
+
+  const [isTopicsLoaded, setIsTopicsLoaded] = useState(false)
+  const [topics, setTopics] = useState([])
```
fabiosantoscode: nit: perhaps we could use a custom hook for these API calls, to keep things consistent:

```js
const useAPICall = (apiFn, ...args) => {
  const [loading, setLoading] = useState(false)
  const [result, setResult] = useState(null)

  useEffect(() => {
    setLoading(true)
    setResult(null)
    apiFn(...args).then(result => {
      setLoading(false)
      setResult(result)
    })
  }, [apiFn, ...args])

  return { loading, result }
}

// ...
const topics = useAPICall(getLatestIssues, /* args for API call can go here */)

topics.loading // bool
topics.result // any
```

src/utils/api.js
iAdramelk: It's mostly a question of there to handle rejections. We can do it either here or in components themselves. For now I decided that it will probably be easier to do there them increase components size.

P. S. I like your idea to transforms API calls to custom hooks, it will solve this problem and they can also have error state in them that components can use. I will try to convert them to custom hooks now.

src/components/Community/Meet/index.js
iAdramelk: Good catch, thank you!

src/components/Community/Meet/index.js
iAdramelk: Good idea, I'd probably would add error state here too.

src/components/Community/Meet/index.js
iAdramelk: Good observation. I will think about how to solve it, but maybe will leave it to the future.

fabiosantoscode: Usually I do something like this:

```
useEffect(() => {
  let canceled = false
  asyncStuff().then(() => {
    if (canceled) return
  })
  return () => {
    canceled = true
  }
}, [])
```

It works for mount/unmount as well as one of the useEffects dependencies changing.

src/utils/api.js
fabiosantoscode: I see what you mean.

But maybe a non-200 response should have the same treatment as an error. This decreases the amount of outcomes the function may have. Right now, it's [error, undefined, response]. If we treat non-200 as an error it's only [error, response].

src/components/Community/Meet/index.js
iAdramelk: Good idea, I will do it this way!

iAdramelk:
@fabiosantoscode Rewrote with hooks, please check.

src/components/Community/Learn/index.js
```diff
@@ -255,17 +249,18 @@ export default function CommunityLearn({ posts, theme }) {
                 )
               }
             >
-              {posts.length ? (
+              {!ready && <Placeholder>Loading...</Placeholder>}
+              {error && (
+                <Placeholder>Blog is unavailable right now</Placeholder>
```
jorgeorpinel:
```suggestion
                <Placeholder>Blog unavailable right now</Placeholder>
```

src/components/Community/Meet/index.js
```diff
@@ -191,17 +195,18 @@ export default function CommunityMeet({ issues, theme, topics }) {
               }
               icon="/static/img/community/discourse.svg"
             >
-              {topics.length ? (
+              {!topicsReady && <Placeholder>Loading...</Placeholder>}
+              {topicsError && (
+                <Placeholder>Forum is unavailable right now</Placeholder>
```
jorgeorpinel:
```suggestion
                <Placeholder>Forum unavailable right now</Placeholder>
```

src/components/Community/Meet/index.js
```diff
@@ -231,17 +236,18 @@ export default function CommunityMeet({ issues, theme, topics }) {
               }
               icon="/static/img/community/github.svg"
             >
-              {issues.length ? (
+              {!issuesReady && <Placeholder>Loading...</Placeholder>}
+              {issuesError && (
+                <Placeholder>Github is unavailable right now</Placeholder>
```
jorgeorpinel:
```suggestion
                <Placeholder>Github unavailable right now</Placeholder>
```

src/utils/api.js
```diff
@@ -7,50 +8,64 @@ export function makeAbsoluteURL(req, uri) {
...
+  useEffect(() => {
+    let cancelled = false
+
+    const fetchData = async () => {
+      try {
+        const res = await fetch(url)
+
+        if (!cancelled) {
+          if (res.status !== 200) {
+            setError('Wrong response types')
```
jorgeorpinel:
```suggestion
            setError('Bad response status')
```

jorgeorpinel:
p.s. sanity check on https://dvc-landing-async-commu-rmgs1p.herokuapp.com/community also looks good. I can see the blog, discourse, and github calls in Network tab.

fabiosantoscode:
Looks good.

The fact that we have to write so much code for this makes me sad we can't inspect promises from the outside. We would be able to read the result, error and whether it's resolved.

src/components/Community/Meet/index.js
```diff
@@ -110,7 +111,10 @@ CommunityIssue.propTypes = {
   color: PropTypes.string
 }
 
-export default function CommunityMeet({ issues, theme, topics }) {
+export default function CommunityMeet({ theme }) {
+  const { erorr: issuesError, ready: issuesReady, result: issues } = useIssues()
+  const { erorr: topicsError, ready: topicsReady, result: topics } = useTopics()
```
fabiosantoscode: Oh, typo here!

jorgeorpinel: I don't see it.

shcheklein: I don't see it either, resolving ... please send a separate update for this if we missed something

shcheklein:
@fabiosantoscode 

> The fact that we have to write so much code for this makes me sad we can't inspect promises from the outside. We would be able to read the result, error and whether it's resolved.

could you elaborate please?

fabiosantoscode:
@shcheklein

Just an aside. But you can't inspect a promise's completion state from the outside, like promise.isFullfilled or promise.isRejected, or get its value. If that was the case, we could use that as a standard object and pass it around.

Issue #992, opened 2020-02-13
blog/website: choose the new hosting


iAdramelk:
UPDATE: See summary of options in https://github.com/iterative/dvc.org/issues/992#issuecomment-587552490

This is a ticket to discuss and compare possible solutions, based on the criteria listed below. 

We plan to convert dvc.org to `gatsby` and merge it with the blog. Right now, dvc.org is hosted on Heroku, and the blog is hosted on Netlify. We need to choose one hosting that will work for both of those services' needs.

What we want:

**API endpoints**

dvc.org and blog both use some custom API endpoints to fetch and transform data from github and discourse. We also need to be able to cache the results of these requests because they are not very fast and shouldn't be updated more often than once in 15 minutes. 

Right now, this is implemented as Node.js server on Heroku with in-memory cache, but we wouldn't necessarily have a server after we migrate to `gatsby`. Also, our current implementation has a problem - for the first user that will try to access the page until the cache is created, load time can be quite significant (~10s). Ideally, we should perform cache update ourselves with something like cron and always sent cached results for all of our users.

To solve this, we can use Netlify functions, Cloudflare workers, or something else.

**Redirects**

We have a large list of redirects that we need to support https://github.com/iterative/dvc.org/blob/master/redirects-list.json

New hosting should allow it.

**Build time**

Our current build time for the blog on Netlify is long, 7m now, and will only get longer after we merge it with the main site. We can speed it up by preserving `.cache` dir, yarn module cache between builds. A new hosting/build option should allow us to preserve them.

**Demo stands**

Right now, both Netlify and Hero allow us to automatically created preview stands from github PRs. We want to have this functionality in the future too.

--

Our new hosting solution may be not one server, but a combination of the few different ones, e. g. CircleCI + Cloudflare Workers + Netlify/Now.sh, but it should be able to all of the things listed above.

shcheklein:
@fabiosantoscode please, take a look. Let's discuss it here.

fabiosantoscode:
I agree with you, in that the holy grail is to have our API be just a bunch of static data updated by a cron. It's similar to the architecture I left behind at a paper, where events would wake up a lambda function which rendered a react page, and then push the result to S3. This was called the [reactive manifesto](https://www.reactivemanifesto.org/) back in the day.

[...]

PR #1004, opened 2020-02-17
Feature/992 now hosting

fabiosantoscode:
Zeit now didn't pan out very well.

They have deprecated the custom node server (all hail cloud lock-in!), so we lose some flexibility there.

We can do redirects using now.json (except subdomain redirects, which need a separate branch hack to be achieved)

I have deployed this:

https://dvc-org-16jswpee8.now.sh

The API functions are working as normal.

Let me know if we can move forward from this point, as I don't want to spend time trying to get something to work that's going to be thrown away :snowman: .

Things should be better with gatsby + serverless functions, that was super easy to integrate earlier. However without a custom server it seems we lose some flexibility, and whether we make this tradeoff is not a decision I'm comfortable making.

My opinion is that we should go ahead with zeit now, because we will probably never need any custom server.


shcheklein:
@fabiosantoscode we are migrating dvc.org to Gatsby ... can we start picking the new hosting taking this into account? probably it means that we need to do some experiments with the blog first? Unless I'm missing something.

shcheklein:
@fabiosantoscode looks good, let's keep discussion going in the ticket.

fabiosantoscode:
Dropping the reference, I forgot

#992

iAdramelk:
@fabiosantoscode quick question about new cache util: Does it cache response only for current user or for all users? It looks for me now that it is only cache response for the user who made it and any new user will make server response again. If I'm correct, then this is not a good option for us, because these requests for API are slow and will decrease out rate limit, while the content that we serve for different users are basically the same.

fabiosantoscode:
@iAdramelk the cache is public.

If you access https://blog-fihp2x2rk.now.sh/api/example-function with 2 different IPs (like with your phone on mobile data and your computer, or using a VPN) you will see the same timestamp.

iAdramelk:
@fabiosantoscode tried it 3 browsers on the same IP and it seems that it works (but the request have only 60s maxAge). Where do you store example-function btw? I don't see it in changed files.

fabiosantoscode:
(replied in https://github.com/iterative/dvc.org/issues/992#issuecomment-587482446)

pages/doc.js
```diff
@@ -35,25 +35,31 @@ DocumentationPage.propTypes = {
   errorCode: PropTypes.number
 }
 
-DocumentationPage.getInitialProps = async ({ asPath, req }) => {
+DocumentationPage.getInitialProps = async ({ asPath, req, res }) => {
+  if (res) {
+    res.setHeader('cache-control', 'public, max-age=3600')
+  }
```
jorgeorpinel: Why no `cache from './utils/cache'`?

pages/doc.js
```diff
@@ -35,25 +35,31 @@ DocumentationPage.propTypes = {
...
   if (!item) {
+    if (res) res.statusCode = 404
     return {
       errorCode: 404
     }
   }
 
   try {
-    const res = await fetch(makeAbsoluteURL(req, item.source))
+    const fetchRes = await fetch(makeAbsoluteURL(req, item.source))
```
jorgeorpinel: Maybe use `docRes` name

jorgeorpinel:
I like the cache code simplification.

> We can do redirects using now.json (except subdomain redirects, which need a separate branch hack to be achieved)

I see the redirects work, but I don't see them in now.json in this PR?

fabiosantoscode:
Closing since we're going for Heroku.

PR #1009, opened 2020-02-20
server: reformat/refactor around redirects (again)

jorgeorpinel:
Second attempt at #973.

### DIFF
--- a/server.js
+++ b/server.js
@@ -24,7 +24,7 @@ const port = process.env.PORT || 3000
 
 app.prepare().then(() => {
   createServer((req, res) => {
-    const parsedUrl = parseURL(req.url)
+    let parsedUrl = parseURL(req.url)
     const { pathname, query: queryStr } = parsedUrl
     const host = req.headers.host
 
@@ -56,6 +56,7 @@ app.prepare().then(() => {
     } else {
       // Regular Next.js handler
 
+      if (null === parsedUrl.query) parsedUrl.query = {}
       handle(req, res, parsedUrl)
     }
   }).listen(port, err => {
--- a/server.js
+++ b/server.js
@@ -11,6 +11,7 @@
 
 const { createServer } = require('http')
 const { parse } = require('url')
+const { stringify } = require('querystring')
 const next = require('next')
 
 const { getItemByPath } = require('./src/utils/sidebar')
@@ -34,13 +35,10 @@ app.prepare().then(() => {
     if (redirectLocation) {
       // HTTP redirects
 
-      console.log(redirectLocation)
-      const { query: queryStr } = parse(req.url)
-      console.log(queryStr)
+      const queryStr = stringify(query)
       if (queryStr) {
         redirectLocation += '?' + queryStr
       }
-      console.log(redirectLocation)
       res.writeHead(redirectCode, {
         'Cache-control': 'no-cache',
         Location: redirectLocation
--- a/server.js
+++ b/server.js
@@ -33,9 +33,7 @@ app.prepare().then(() => {
       dev
     })
     if (redirectLocation) {
-      /*
-       * HTTP redirects
-       */
+      // HTTP redirects
 
       if (queryStr) {
         redirectLocation += '?' + queryStr
@@ -46,9 +44,7 @@ app.prepare().then(() => {
       })
       res.end()
     } else if (/^\/doc(\/.*)?$/.test(pathname)) {
-      /*
-       * Docs Engine handler
-       */
+      // Docs Engine handler
 
       // Force 404 response code for any inexistent /doc item.
       if (!getItemByPath(pathname)) {
@@ -58,9 +54,7 @@ app.prepare().then(() => {
       // Custom route for all docs
       app.render(req, res, '/doc', parseQuery(queryStr))
     } else {
-      /*
-       * Regular Next.js handler
-       */
+      // Regular Next.js handler
 
       handle(req, res, parsedUrl)
     }
--- a/server.js
+++ b/server.js
@@ -25,7 +25,7 @@ const port = process.env.PORT || 3000
 app.prepare().then(() => {
   createServer((req, res) => {
     const parsedUrl = parseURL(req.url)
-    const { pathname, queryStr } = parsedUrl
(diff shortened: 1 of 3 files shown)


jorgeorpinel:
Weird. This works for me locally (`yarn dev`) but crashes in the (built) review app (https://dvc-landing-server-redi-btalyj.herokuapp.com/)... Investigating ⏳

jorgeorpinel:
@iAdramelk @shcheklein related: do you guys know why when you `yarn build` and `yarn start` this locally, it tries to enforce HTTPS on localhost:3000? Makes it impossible for me to debug the built app locally.

jorgeorpinel:
Anyway, looking at https://dashboard.heroku.com/apps/dvc-landing-server-redi-btalyj/logs I see the review app error log is showing this a lot:

```
2020-02-20T07:43:44.228837+00:00 app[web.1]: TypeError: Cannot read property 'amp' of null
2020-02-20T07:43:44.228839+00:00 app[web.1]:     at Server.findPageComponents (/app/node_modules/next/dist/next-server/server/next-server.js:465:19)
...
```

which unfortunately is not a very informative stack trace... But per https://github.com/zeit/next.js/issues/7303#issuecomment-511053550 this probably has to do with my changes around `url.parse`... Will try one of the fixes in that report ⏳

iAdramelk:
@jorgeorpinel 

> related: do you guys know why when you yarn build and yarn start this locally, it tries to enforce HTTPS on localhost:3000? Makes it impossible for me to debug the built app locally.

You can temporarily disable this redirect by commenting these lines: https://github.com/iterative/dvc.org/blob/master/src/utils/redirects.js#L37-L39

shcheklein:
@iAdramelk @fabiosantoscode @jorgeorpinel this is a bug . It should not be redirecting to https locally.

fabiosantoscode:
@jorgeorpinel NODE_ENV is set to production, making `dev` false, which makes the redirect stuff issue a redirect to HTTPS. I'll send a PR to avoid such a redirect on localhost.

shcheklein:
@fabiosantoscode I would check if the redirects engine itself also matches on https only (thus preventing us from testing at least some redirects on localhost)

fabiosantoscode:
@shcheklein path redirects are relative, so they go to localhost. Hostname redirects (man.dvc.org, error.dvc.org) take you to the real site, where HTTPS works.

server.js
```diff
@@ -23,43 +24,38 @@ const port = process.env.PORT || 3000
 
 app.prepare().then(() => {
   createServer((req, res) => {
-    const parsedUrl = parse(req.url, true)
-    const { pathname, query } = parsedUrl
+    const parsedUrl = parseURL(req.url)
+    const { pathname, queryStr } = parsedUrl
```
fabiosantoscode: There's no `queryStr` when parsing a URL. The querystring is called "query" just the same. The difference is that it's an object if `url.parse`'s second argument is true.

src/utils/redirects.test.js
```diff
@@ -3,41 +3,41 @@ const { processRedirectString, getRedirect } = require('./redirects')
...
 describe('processRedirectString', () => {
   it('reads the regex, replacement and code', () => {
-    const { regex, replace, code } = processRedirectString('^/foo /bar 418')
+    const { regex, replace, code } = processRedirectString('^/foo /bar 308')
 
     expect(regex).toBeInstanceOf(RegExp)
     expect(regex.source).toEqual('^\\/foo')
     expect(replace).toEqual('/bar')
-    expect(code).toEqual(418)
+    expect(code).toEqual(308)
```
fabiosantoscode: You removed my teapot joke :( it's old anyway.

server.js
jorgeorpinel: Thanks! Yes, this is part of the problem (my bad, fixed in b85eb64). But the actual issue is a bug in Next, it seems. See https://github.com/zeit/next.js/issues/7303#issuecomment-511053550

Addressing... ⏳

src/utils/redirects.test.js
jorgeorpinel: Wow I didn't even remember that one. I thought it was just a random number haha. Reverted now.

jorgeorpinel:
Thanks Fabio, that was part of it.

Fixed this now in e51bc64 ! It's actually a kinda bug from Next... It expects that you always call `url.parse` sending `true` to its 2nd param (`parseQueryString`)... See https://github.com/zeit/next.js/issues/7303#issuecomment-589267045

jorgeorpinel:
In any case... This review app works well now. https://dvc-landing-server-redi-btalyj.herokuapp.com/

Thanks everyone for all the feedback.

server.js
```diff
@@ -23,43 +24,39 @@ const port = process.env.PORT || 3000
...
         res.statusCode = 404
       }
 
       // Custom route for all docs
-      app.render(req, res, '/doc', query)
+      app.render(req, res, '/doc', parseQuery(queryStr))
     } else {
       // Regular Next.js handler
+
+      if (null === parsedUrl.query) parsedUrl.query = {}
```
shcheklein: does it worth doing this minor optimization considering that we have to this hack?

jorgeorpinel: Probably not.

I don't see it as a hack though, just part of the necessary custom Next server logic.

But anyway should I just revert this whole file and leave the redirect file changes (totally cosmetic)? If this whole PR is pointless at this point, please just close it 🙂

shcheklein: up to you @jorgeorpinel  :)

jorgeorpinel: OK I reverted the optimization and hack. But then I found a better one! Implemented in 08522d88, please check.

shcheklein:
👍 LGTM

jorgeorpinel:
Deployed and double sanity checked on prod. All good! 🎉

fabiosantoscode:
@jorgeorpinel let me know if you had issues testing redirects locally. I tested them with:

`curl -I --header 'host: man.dvc.org` localhost:3000/something`

And checked the status code and location header for each redirect. This is also possible with HTTP GUIs like postman.

jorgeorpinel:
Seems to work properly after #1010. Thanks!

PR #1010, opened 2020-02-20
redirects: do not redirect to HTTPS on localhost

fabiosantoscode:
This prevents an HTTPS redirect while testing with `yarn build && yarn start` on localhost.

### DIFF
--- a/src/utils/redirects.js
+++ b/src/utils/redirects.js
@@ -34,7 +34,8 @@ const matchRedirectList = (host, pathname) => {
 }
 
 const getRedirect = (host, pathname, { req, dev } = {}) => {
-  if (req != null && req.headers['x-forwarded-proto'] !== 'https' && !dev) {
+  const httpsRedirect = req != null && !dev && !/^localhost(:\d+)?$/.test(host)
+  if (httpsRedirect && req.headers['x-forwarded-proto'] !== 'https') {
     return [301, `https://${host.replace(/^www\./, '')}${req.url}`]
   }
 
--- a/scripts/exclude-links.txt
+++ b/scripts/exclude-links.txt
@@ -18,6 +18,7 @@ https://dvc.org/static/img/<filename>.gif
 https://example.com/data.txt
 https://example.com/foo
 https://example.com/foo/bar?baz
+https://dvc.org/foo/bar?baz
 https://example.com/path/to/data
 https://example.com/path/to/data.csv
 https://example.com/path/to/dir
--- a/src/utils/redirects.test.js
+++ b/src/utils/redirects.test.js
@@ -37,8 +37,8 @@ describe('getRedirect', () => {
       url: '/foo/bar?baz'
     }
     expect(
-      getRedirect('www.example.com', '/not-used', { req: fakeReq, dev: false })
-    ).toEqual([301, 'https://example.com/foo/bar?baz'])
+      getRedirect('www.dvc.org', '/not-used', { req: fakeReq, dev: false })
+    ).toEqual([301, 'https://dvc.org/foo/bar?baz'])
   })
 
   const itRedirects = (source, target, code = 301) => {


src/utils/redirects.js
```diff
@@ -34,7 +34,8 @@ const matchRedirectList = (host, pathname) => {
 }
 
 const getRedirect = (host, pathname, { req, dev } = {}) => {
-  if (req != null && req.headers['x-forwarded-proto'] !== 'https' && !dev) {
+  const httpsRedirect = req != null && !dev && /^(www)?\.dvc\.org$/.test(host)
```
shcheklein: don't quite understand why `!dev` is not enough like it was before

fabiosantoscode: The reason for this, is that `yarn serve` sets NODE_ENV to production, making `dev` here false and forcing a redirect from localhost to https://localhost.

I could've removed the NODE_ENV change in the package.json script, but it affects a lot of other things.

shcheklein: do we need to force it production? I think Heroku set's this env variable on its own. 

mainly my question - why this check was enough before? trying to understand what has changed.

fabiosantoscode: Nothing's changed. It's just that while testing this I didn't think to run `yarn build && yarn serve` and didn't know of this env set.

fabiosantoscode: Heroku probably does force it to production, but when locally testing something we want it to be close to production.

shcheklein: got it. don't understand then why do we need to be able to run production locally ... we have `yarn dev` for this, no? cc @jorgeorpinel ?

fabiosantoscode: I can see the benefits of testing locally. Any difference between dev and prod becomes apparent.

shcheklein: kk, review environments should work fine with this?

also, is there a global var already by any chance that defines `dvc.org` hostname?

shcheklein: may be we should be just testing `localhost` name?

fabiosantoscode: That sounds good.

iAdramelk: We need to set `NODE_ENV` to `production` in yarn build because it do a lot of additional optimizations for the build:

1. Remove unused code (half of the react library for example, debugging and console spamming parts).
2. Minimize it.
3. Create static js/html files.
4. Etc.

Without correct `NODE_ENV` it will not work.

Dev server is run with `NODE_ENV` `development`:

1. A lot of additional debug stuff.
2. No minification.
3. Hot reload.
4. Different parts of js- and html-files in memory.
5. Etc.

Most of the time dev server is enough. But in some cases we want to test exactly the code that we ship to users. And in this case `yarn build` & `yarn serve` can be used.

fabiosantoscode: I see where you're coming from. However minimization is really slow, doesn't change the flow of the program and makes stack traces harder to read. Tree shaking should be pretty speedy but doesn't change program flow either.

Even with webpack's cache, since it sends whole chunks at once to Terser, we can't really benefit from its cache for the slowest part of the build. Rollup doesn't do this, but I'm not arguing for a bundler change here.

I would argue that react spamming the console is a positive. We want to squish these bugs as hard as we can.

Barring bugs in either of these, which are a bit unlikely, and bundle sizes, we're running exactly the same thing.

There will be a small performance change between local and production, but most of the time it's a negligible difference, and we want to make sure things are snappy in dev mode on our snazzy laptop, so they're closer to snappy on mobile.

If we do need to test production, we can set the environment variable and wait some more time to build.

shcheklein: @iAdramelk I think NODE_ENV is set to prod on Heroku, no?

iAdramelk: @fabiosantoscode my point is that we should still use dev mode most of the time, but be able to test prod builds is important too. So we need to preserve it too. I think we are speaking about the same thing here.

@shcheklein It can be set on Heroku too, but I'm pretty sure that it is also hardcoded inside the `next build` command. And disabling it will require some hacks. My point here is that to be able to test some esoteric and hard-to-reproduce bugs locally (and also to be able to do bundle size profiling) we should leave it as is and not to try to remove `NODE_ENV=production` from `build` step.

shcheklein: @iAdramelk got it! So, no need to change anything. It looks like we all agreed to keep it this way.

PR #1014, opened 2020-02-22
Regular updates (late Feb)

### LINKED ISSUE #1026: Does DVC support rsync remotes?
The homepage says yes. The rest of the documentation doesn't mention it. The code doesn't seem to mention it.

jorgeorpinel:
- Misc. stuff
- Update paragraph about .dvc/ dir created by `dvc init` per https://github.com/iterative/dvc.org/pull/1022#pullrequestreview-371480713
- Closes #1026
- Uses infinitive ("Does xyz") in the short intros of cmd refs, and simple present ("Do this and that") or other forms of speech in the Description sections.

### DIFF
--- a/public/static/docs/command-reference/cache/index.md
+++ b/public/static/docs/command-reference/cache/index.md
@@ -15,10 +15,10 @@ positional arguments:
 
 ## Description
 
-At DVC initialization, a new `.dvc/` directory will be created with the
-`config`, `.gitignore`, among
-[other files and directories](/doc/user-guide/dvc-files-and-directories),
-including the default cache directory.
+At DVC initialization, a new `.dvc/` directory will be created for internal
+configuration and cache
+[files and directories](/doc/user-guide/dvc-files-and-directories) that are
+hidden from the user.
 
 The cache is where your data files, models, etc (anything you want to version
 with DVC) are actually stored. The corresponding files you see in the
--- a/public/static/docs/command-reference/cache/index.md
+++ b/public/static/docs/command-reference/cache/index.md
@@ -15,8 +15,9 @@ positional arguments:
 
 ## Description
 
-After DVC initialization, a hidden directory `.dvc/` is created to contain the
-[DVC files and directories](/doc/user-guide/dvc-files-and-directories),
+At DVC initialization, a new `.dvc/` directory will be created with the
+`config`, `.gitignore`, among
+[other files and directories](/doc/user-guide/dvc-files-and-directories),
 including the default cache directory.
 
 The cache is where your data files, models, etc (anything you want to version
--- a/public/static/docs/command-reference/metrics/index.md
+++ b/public/static/docs/command-reference/metrics/index.md
@@ -1,6 +1,6 @@
 # metrics
 
-A set of commands to collect and display project metrics:
+A set of commands to add, manage, collect, and display project metrics:
 [add](/doc/command-reference/metrics/add),
 [show](/doc/command-reference/metrics/show),
 [diff](/doc/command-reference/metrics/diff),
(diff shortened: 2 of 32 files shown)


public/static/docs/command-reference/cache/index.md
```diff
@@ -15,8 +15,9 @@ positional arguments:
 
 ## Description
 
-After DVC initialization, a hidden directory `.dvc/` is created to contain the
-[DVC files and directories](/doc/user-guide/dvc-files-and-directories),
+At DVC initialization, a new `.dvc/` directory will be created with the
+`config`, `.gitignore`, among
```
shcheklein: may be not for this PR - what do you think about the choice of files to mention? I feel that `.dvc/cache` is even more important to mention? or may be we should not be mentioning them at all - just put a link? 

This choice `config`, `.gitignore` looks very artificial to me, tbh. Especially when we don't explain it.

public/static/docs/command-reference/metrics/index.md
```diff
@@ -16,7 +16,7 @@ positional arguments:
   COMMAND
     show                Output metric values.
     add                 Tag file as a metric file.
-    modify              Modify metric file options.
+    modify              Modify metric file values.
```
jorgeorpinel: Opened iterative/dvc/pull/3471 to match this (and other output string updates).

public/static/docs/command-reference/metrics/modify.md
```diff
@@ -1,7 +1,7 @@
 # metrics modify
 
-Modify [project metric](/doc/command-reference/metrics) values such as type,
-path expression that is used to parse it, etc. (See full options below.)
+Modify [project metric](/doc/command-reference/metrics) default formatting with
```
shcheklein: hmm ... I kinda like it, but we might need to change it later ... it's not about formatting only eventually

public/static/docs/command-reference/cache/index.md
jorgeorpinel: `.dvc/cache` was mentioned originally I think, but it's not created at `dvc init` now. Only until some data is actually put in cache with add, fetch, etc.

> may be we should not be mentioning them at all - just put a link?

We could do this. I don't have a strong opinion. `config` and `.gitignore` seem like the more important ones because a) they're always created (the other ones are optional or get created later than init) and b) they're the ones a user is more likely to look at or even edit manually (along with cache/ dir of course).

public/static/docs/command-reference/metrics/modify.md
jorgeorpinel: I understand. I see that the original intention was probably to add more options than type and xpath but for now its only about formatting so it makes sense I think, and yes of course we'll have to update later if more options are added that change other kind of definitions.

jorgeorpinel: But if you prefer I can try to use a different term like "definition" or "specification".

public/static/docs/command-reference/cache/index.md
shcheklein: yeah, probably better remove them ... or rephrase - to mention that this directory is about configs (I at least understand the importance of this) and eventually cache

public/static/docs/command-reference/metrics/modify.md
shcheklein: not a strong opinion, let's see what other guys on the DVC core say about this ... I like actually how does it sound now - it's very user friendly

public/static/docs/command-reference/cache/index.md
jorgeorpinel: OK, updated!

jorgeorpinel: Gonna merge this since you previously approved 🙂

PR #1017, opened 2020-02-24
Fix for 'update' of null error in Sentry

iAdramelk:
We were trying to update Perfect Scrollbar after 400ms timeout. If the was user already left the page it caused an error. Now we only update it if the ref to perfect scrollbar is sill present on page after timeout.

See:

https://sentry.io/organizations/iterative/issues/1529236771/
https://sentry.io/organizations/iterative/issues/1525523962/

And others like that.

### DIFF
--- a/src/components/Documentation/SidebarMenu/index.js
+++ b/src/components/Documentation/SidebarMenu/index.js
@@ -76,9 +76,11 @@ export default function SidebarMenu({ id, sidebar, currentPath, onClick }) {
     setIsScrollHidden(true)
 
     setTimeout(() => {
-      psRef.current.update()
-      scrollIntoView(node, parent, { onlyScrollIfNeeded: true })
-      setIsScrollHidden(false)
+      if (psRef.current) {
+        psRef.current.update()
+        scrollIntoView(node, parent, { onlyScrollIfNeeded: true })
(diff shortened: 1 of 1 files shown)


fabiosantoscode:
I'm not sure about this approach. I realise it's not entirely a part of this PR (feel free to ask me to push a separate one), but I recommend the use of `clearTimeout` for this reason:

The `currentPath` might change without an unmount while traversing between two pages. This will make useEffect's returned "undo" function be called, and `psRef.current` be reassigned to something new. When this happens, there will be two setTimeout callbacks scheduled, and both of them will see that `psRef.current` is not null, since refs don't change unless the component unmounts.

An alternative to `clearTimeout` is to create a `cancel` variable, and check for that instead of the ref.

iAdramelk:
@fabiosantoscode you are totally right. Replaced it with `clearTimeout`.

iAdramelk:
@fabiosantoscode you are totally right. Replaced it with `clearTimeout`.

shcheklein:
@iAdramelk could you please merge/rebase?

iAdramelk:
@shcheklein ooops, I merged master yesterday, but forgot to write about it.

PR #1018, opened 2020-02-24
Fix ie11 fill

iAdramelk:
Another Sentry error https://sentry.io/organizations/iterative/issues/1521458935/


fabiosantoscode:
Looks good! Good to know IE11 doesn't support Array#fill.

PR #1022, opened 2020-02-25
init: add subdir description

pared:

Please chose to [allow us](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to edit your branch when creating the PR.

Related to https://github.com/iterative/dvc/pull/3257

### DIFF
--- a/public/static/docs/command-reference/init.md
+++ b/public/static/docs/command-reference/init.md
@@ -3,12 +3,12 @@
 This command initializes a <abbr>DVC project</abbr> on a directory.
 
 Note that by default the current working directory is expected to contain a Git
-repository, unless the `--no-scm` option is used.
+repository, unless the `--no-scm` or `--subdir` option is used.
 
 ## Synopsis
 
 ```usage
-usage: dvc init [-h] [-q | -v] [--no-scm] [-f]
+usage: dvc init [-h] [-q | -v] [--no-scm] [-f] [--subdir]
 ```
 
 ## Description
@@ -29,6 +29,9 @@ local cache and you cannot `git push` it.
 
 - `--no-scm` - skip Git specific initialization, `.dvc/.gitignore` will not be
   written.
+
+- `--subdir` - initialize <abbr>DVC repository</abbr> in current directory and
+  allow to search for Git repository in parent directories
 
 - `-f`, `--force` - remove `.dvc/` if it exists before initialization. Will
   remove any existing local cache. Useful when a previous `dvc init` has been
@@ -66,4 +69,29 @@ $ cat .dvc/.gitignore
 /lock
 ...
 /cache
+```
+
+Create a new <abbr>DVC repository</abbr> in a subdirectory of Git repository:
+
+```dvc
+$ mkdir repo && cd repo
+
+$ git init
+$ mkdir subrepo && cd subrepo
+
+$ dvc init --subdir
+```
+
+In this case, Git repository is inside `repo` directory, while <abbr>DVC
+repository</abbr> is inside `repo/subrepo`.
+
+```dvc
+$ tree repo -a
+repo
+├── .git
+.
+.
+.
+└── subrepo
+    └── .dvc
 ```
--- a/public/static/docs/command-reference/init.md
+++ b/public/static/docs/command-reference/init.md
@@ -1,9 +1,6 @@
 # init
 
-This command initializes a <abbr>DVC project</abbr> on a directory.
-
-Note that by default the current working directory is expected to contain a Git
-repository, unless the `--no-scm` or `--subdir` option is used.
+This command initializes a <abbr>DVC project</abbr> in a directory.
 
 ## Synopsis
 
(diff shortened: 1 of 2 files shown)


public/static/docs/command-reference/init.md
```diff
@@ -67,3 +70,28 @@ $ cat .dvc/.gitignore
...
 /cache
 ```
+
+Create a new <abbr>DVC repository</abbr> in a subdirectory of Git repository:
+
+```dvc
+$ mkdir repo && cd repo
+
+$ git init
+$ mkdir subrepo && cd subrepo
```
jorgeorpinel:
```suggestion
$ mkdir subdir && cd subdir
```

public/static/docs/command-reference/init.md
```diff
@@ -67,3 +70,28 @@ $ cat .dvc/.gitignore
...
+
+$ dvc init --subdir
+```
+
+In this case, Git repository is inside `repo` directory, while <abbr>DVC
+repository</abbr> is inside `repo/subrepo`.
```
jorgeorpinel: The Git repo is in the `repo/` directory, while the _subdir_ <abbr>DVC repository</abbr> lives in `repo/subdir`.

public/static/docs/command-reference/init.md
```diff
@@ -67,3 +70,28 @@ $ cat .dvc/.gitignore
...
+
+```dvc
+$ tree repo -a
+repo
+├── .git
+.
+.
+.
+└── subrepo
```
jorgeorpinel:
```suggestion
└── subdir
```

public/static/docs/command-reference/init.md
```diff
@@ -30,6 +30,9 @@ local cache and you cannot `git push` it.
 - `--no-scm` - skip Git specific initialization, `.dvc/.gitignore` will not be
   written.
 
+- `--subdir` - initialize <abbr>DVC repository</abbr> in current directory and
+  allow to search for Git repository in parent directories
```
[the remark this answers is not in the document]
pared: @jorgeorpinel
- parent can be a DVC repo
- yes, if there is no `no-scm` inside `config`, we will search the whole tree

pared:
@jorgeorpinel introduced your suggestions

public/static/docs/command-reference/init.md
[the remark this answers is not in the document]
jorgeorpinel: Thanks, I'll add some notes about these details 🙂

jorgeorpinel:
@shcheklein there's a few possible improvements to this but I think the PR is mergeable, I can address in regular updates.

shcheklein:
@pared @jorgeorpinel 

I was not following the DVC core PR, could you please clarify the following things. No need to add them right now in the PR, but the will help in taking this PR over. Would be great to have some answers here in any way your prefer :)

1. Did we change anything about --no-scm (do we write configs, flags now)?
2. Same question - do we write anything when we use `subdir`.
3. Can I do nesting - one subdir, inside another, etc
4. Can I do subsid in the root. Does it makes sense? What will happen.
5. How having subdirs affect the flow (basically, why do we need them in the first place). Flow of the regular commands like `pull`, `push`, etc.
6. What happens if I do no-scm with subdir?


Let's start with this. I think we need more clarify for this before we merge this. The way we've done it so far is too low level and technical, not clear the motivation and not clear how does it affect behavior.

public/static/docs/command-reference/init.md
```diff
@@ -30,6 +30,9 @@ local cache and you cannot `git push` it.
 - `--no-scm` - skip Git specific initialization, `.dvc/.gitignore` will not be
   written.
```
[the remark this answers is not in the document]
jorgeorpinel: Actually, not sure about this per https://github.com/iterative/dvc/pull/3257#discussion_r383759165

jorgeorpinel:
> 6. What happens if I do no-scm with subdir?

@shcheklein I had this same question in https://github.com/iterative/dvc/pull/3257#pullrequestreview-362870906, and Pawel's answer is

> it is technically possible to do that. Still, I would leave this check (both options can't be used together) as using both flags at the same time indicates a misunderstanding of how they work.

public/static/docs/command-reference/init.md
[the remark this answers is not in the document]
jorgeorpinel: Never mind, this is fine in this PR, as it reflects the current --subdir behavior.

pared:
@shcheklein 
Answering your questions:
1. Yes, now `no-scm` is a config flag now, and that change comes with behavior change:
now if `no-scm` is found in the config, we will use NoSCM class, always. Before it was possible to init with this flag and later initialize git, which was acknowledged
2. No, `subdir` is the default behavior (unless `no-scm` is in config)
3. Yes
4. Yes, nothing will change, we will look for git root starting from root, git will be found. So we can say that we are taking "closest" scm in tree. In a way, a normal use case (Git + DVC in one dir) is just a specific `subdir` case.
5. Subdir does not influence normal use case, data sync operations will not be performed in subrepos, which are ignored during tree traversing. 
6. Init error. Without this error, we would end up with NoSCM case, because it just happens that check for `no-scm` is earlier in code than searching for Git

shcheklein:
@pared thanks! please, check the next round of questions to clarify the PR! Any details would be helpful. You can edit the same comment https://github.com/iterative/dvc.org/pull/1022#issuecomment-591960882

pared:
@shcheklein Tried to provide some more broad answer.
You found a bug in the current implementation. I believe that we can proceed in 2 ways:
1. Try to implement logic to support calling particular stage files and be aware that it can be in subdir.
2. Forbid initializing `subdir` if dvc repo is found while searching for git root. 

I believe that second option is appropriate, as original issue was about having multiple dvc repos inside one git repo.

jorgeorpinel:
> (1) The config flag is no_scm. It can be changed using config:
dvc config core.no_scm True

- [ ] Should be added to config ref

> (2) subdir is the default behavior (unless no-scm is in config)... if we do not find no_scm flag inside the config, we assume that "There must be git somewhere"

Hmmm. I tried `dvc init --subdir` in an empty dir. Got `ERROR: failed to initiate DVC - /.../test is not tracked by any supported SCM tool (e.g. Git)`. Also you later say "if we don't specify --subdir, we don't allow dvc to look for git up the tree" so I still don't understand what you mean by "subdir is the default behavior" 🙁 

> (4) ... subdir in the root?
Yes, nothing will change

👍 
- [ ]  I'd just add this in the option description: "If run in a project root, this option is ignored."

> (5) data sync commands without specific target. Stages for subdirs will not be gathered

Kind of like Git subrepos?
- [ ] Important info to put in cmd refs as well 🙂 

> (5) run dvc pull subdir/some.dvc from the root... will try to use current repo to perform the operation.
That is a flaw... You found a bug

Should we open a separate dvc core issue to discuss https://github.com/iterative/dvc.org/pull/1022#issuecomment-592463575 ?

> (2) why do they contradict each other?
() I guess the confusion comes from the flag naming, initially, I named it --subrepo

--subrepo may also confusing, it sounds like its a full repo inside a full repo, but its not a full repo per se: its a DVC project inside a subdirectory of a Git/DVC repo. Maybe --sub-scm ? (I'd consider renaming --no-git and --sub-git)

Could making both --no-scm and this option automatic/implicit help?

pared:
@jorgeorpinel 
>  so I still don't understand what you mean by "subdir is the default behavior"

It is the default behavior when handling SCM inside our code. `init` command is a different thing. 
Yours did fail, because you tried initializing inside empty directory, and no git was up the tree right?

> I'd just add this in the option description: "If run in a project root, this option is ignored."

Agree

> Kind of like Git subrepos?
Yes

> Should we open a separate dvc core issue to discuss #1022 (comment) ?
yes, i will do that

>Could making both --no-scm and this option automatic/implicit help?

Probably yest, though I would discuss it, as for now `--subdir` guards user from unintentionally creating repo down the tree.

pared:
@jorgeorpinel 
> Important info to put in cmd refs as wel

I would wait with adding information about that, as its not decided yet whether we will support repo inside repo scenario after detecting the "targeted sync" bug.

shcheklein:
@jorgeorpinel @pared I've update the doc and added more information about this options, why they are needed, etc. Please review and let me know WYT.

@jorgeorpinel feel free to put small fixes on top/merge/etc.

shcheklein:
May be we should consider moving some part to the User Guide.

pared:
@shcheklein I think those changes are desirable, they point "main" use case and provide info on additional possible workflows.

shcheklein:
@jorgeorpinel ping!

jorgeorpinel:
I'm back guys, sorry for the delay.

> > I still don't understand what you mean by "subdir is the default behavior"
> It is the default behavior when handling SCM inside our code. init command is a different thing.
> ... --subdir guards user from unintentionally creating repo down the tree

@pared OK so it doesn't affect the docs in this PR but I'm still trying to understand what "default behavior when handling SCM inside our code" means or what impacts it has. Can you elaborate a little? Maybe it impacts other docs...

Other than that we've taken over this PR and I'm trying to merge it ASAP. Thanks

public/static/docs/command-reference/init.md
```diff
@@ -1,49 +1,178 @@
...
 After DVC initialization, a new directory `.dvc/` will be created with the
-`config` and `.gitignore` files. These and other files and directories are
-hidden from user, as typically there's no need to interact with them directly.
-See [DVC Files and Directories](/doc/user-guide/dvc-files-and-directories) to
-learn more.
+`.dvc/config`, `.dvc/.gitignore` and other files. These files and directories
+are hidden from the user (but committed with Git), as typically there's no need
+to interact with them directly. See
+[DVC Files and Directories](/doc/user-guide/dvc-files-and-directories) to learn
+more about the structure of the `.dvc/` directory.
```
jorgeorpinel: I'm going to revert this because there's no significant difference and it's a paragraph that's repeated in several docs.

shcheklein: yeah, not a big deal ... I just didn't like `... and other` and other directories - not cleat what other?  too complicated - I just tried to make it very simple

jorgeorpinel:
Merged 🎉 thanks again @pared.

Let's continue the discussion in #1039, and address depending on priority.

Feel free to review my latest changes here and leave feedback in that continuation issue.

Thanks

public/static/docs/command-reference/init.md
jorgeorpinel: I'll revise it in all the docs where this exists.

shcheklein: 👍

jorgeorpinel: Done in bc43303c for #1014

PR #1025, opened 2020-02-26
[WIP] Gatsby migration

iAdramelk:
ToDo:

- [x] Install Gatsby.
- [x] Remove Next.js.
- [x] Move `_document.js` content to `html.js` and `SEO` files.
- [x] Rewrite `LocalLink` to use `gatsby-link`.
- [x] Update GA to work with new setup.
- [x] Move sentry from `_app.js` to gatsby plugin.
- [x] Set up `styled-components` SSR in gatsby.
- [x] Move static pages to gatsby (index, features, support, community).
- [x] Replace CDN urls for css and js assets in head with normal imports.
- [x] Docs page
    - [x] Basic render
    - [x] Sidebar
        - [x] Make standard pages work
        - [x] Make `source: false` menu items work (update logic for `normalizedSidebar` generation)
    - [x] Right menu 
        - [x] Headers with ids
        - [x] Custom ids in headers
        - [x] Move `headings` extraction to backend
    - [x] External link icons
    - [x] Prism syntaxes
        - [x] Port custom syntax for `dvc` from blog
        - [x] Convert custom syntax for `usage`
        - [x] Find existing syntax for `vim`
    - [x] Convert dvc commands to links
    - [x] Custom React components for `details` and `abbr`
- [x] Set up redirects.
- [x] Set up API endpoints.


shcheklein:
@fabiosantoscode could start reviewing this? )

src/components/LocalLink/index.js
```diff
@@ -1,28 +1,20 @@
...
-  as: Component,
-  href,
+  as: SC,
+  href: to,
   ...restProps
 }) {
-  const nextProps = href.match(/^\/doc/)
-    ? { href: PAGE_DOC, as: href }
-    : { href }
+  const Component = SC ? SC.withComponent(GatsbyLink) : GatsbyLink
```
fabiosantoscode: I'm sorry to say this API is [being deprecated](https://styled-components.com/docs/api#withcomponent)!

The new way to do this is to pass an `as` attribute to the StyledComponent itself as described [here](https://styled-components.com/docs/api#as-polymorphic-prop).

src/components/TrySection/index.js
```diff
@@ -9,14 +9,14 @@ export default function TrySection({ title, buttonText = 'Get Started' }) {
   return (
     <Wrapper>
       <Container>
-        <Glyph src="/static/img/glyph-3.svg" gid={'topleft'} />
+        <Glyph src="/img/glyph-3.svg" gid={'topleft'} />
         <Title>{title}</Title>
         <Buttons>
-          <LocalLink href="/doc/get-started" as={Button} first>
+          <LocalLink href="/doc/get-started" as={Button} first="true">
```
fabiosantoscode: Could you make sure this ends up being rendered as an `<a>`? You probably did, but just making sure for SEO/A11y purposes.

src/html.js
```diff
@@ -0,0 +1,58 @@
...
+          key={`body`}
+          id="___gatsby"
+          dangerouslySetInnerHTML={{ __html: props.body }}
+        />
+        {props.postBodyComponents}
+        <script
+          type="text/javascript"
+          src="//downloads.mailchimp.com/js/signup-forms/popup/embed.js"
+          data-dojo-config="usePlainJson: true, isDebug: false"
+        ></script>
```
fabiosantoscode: Maybe we could use Helmet to only render this on the page which needs it? (IIRC the home page)

src/utils/ga.js
```diff
@@ -1,24 +1,22 @@
...
 
-export const logException = (description = '', fatal = false) => {
-  if (description) {
-    ReactGA.exception({ description, fatal })
+  if (exDescription) {
+    window.ga('send', {
+      hitType: 'exception',
+      exDescription,
+      exFatal
+    })
```
fabiosantoscode: Aside: We can probably use something like sentry or crashlytics here. If we end up implementing it in another iterative property we can plop it here as well at no additional cost.

src/components/Documentation/NewMarkdown/index.js
```diff
@@ -0,0 +1,113 @@
...
+            <span>Prev</span>
+          </LocalLink>
+        ) : (
+          <Button disabled={true}>
+            <i className="prev" />
+            <span>Prev</span>
+          </Button>
+        )}
+        {next ? (
+          <LocalLink href={next} as={Button} disabled={!next}>
```
fabiosantoscode: `next` here will always be truthy, so there should be no need to assign the disabled attribute.

src/html.js
```diff
@@ -13,6 +13,14 @@ export default function HTML(props) {
...
           content="width=device-width, initial-scale=1, shrink-to-fit=no"
         />
+        <link
+          rel="stylesheet"
+          href="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.2/dist/cdn/docsearch.min.css"
+        />
+        <script
+          type="text/javascript"
+          src="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.2/dist/cdn/docsearch.min.js"
+        />
```
fabiosantoscode: Should we not import these files from our JS entry point to make them available in a single request?

fabiosantoscode:
This is a gargantuan amount of work, good luck @iAdramelk :)

src/components/LocalLink/index.js
iAdramelk: Good catch! I missed this part on the docs somehow. Will change it.

src/components/TrySection/index.js
iAdramelk: I probably should rename this component, all `Button`s on this page are actually `styled.a`, but yes, the all are rendered to `a`.

src/html.js
iAdramelk: I'd like too, but we need to test it, to find out if it injects and executes code correctly, and that there is no problems if the code executes before html-element for it is rendered.

For now I'd like to leave it as it was in nextjs version, but maybe we should create separate task for that.

src/utils/ga.js
iAdramelk: This, or maybe switch to GTM and send it from there.

src/components/Documentation/NewMarkdown/index.js
iAdramelk: You are right, will try to remove and check if there is some PropTypes errors. (I have vague memory that I left it there because of them, but I need to recheck)

src/html.js
iAdramelk: I did it like that at first, but I started getting build errors with `window is undefined`, so for now I moved it back there. I'd like to research it a little later, probably in separate PR.

But I think that I can return css without problem.

fabiosantoscode: I see! This is probably due to serverside rendering.

In that case, there's 2 options:

1. Use the client entrypoint (maybe import it from gatsby-browser.js if you can't find it)
2. Use a conditional require. `if (typeof window === 'object') require('...')`.

fabiosantoscode: Either way, sure, it's not a big deal.

iAdramelk: Yep, it's definitely SSR. I thought about using `gatsby-browser.js` to fix it, but decided to leave it as is for now, until I make other stuff work.

iAdramelk:
@fabiosantoscode I added AIP and reedirects middlewares + simple express server, can you please review them and advice on additional headers and other stuff that I maybe should add here?

shcheklein:
@iAdramelk please rebase/merge, and let's check why CI is failing

src/components/Community/Section/index.js
```diff
@@ -47,12 +45,6 @@ export default function CommunitySection({
     }
 
     updateVisibility()
-
-    Router.events.on('hashChangeComplete', updateVisibility)
-
-    return () => {
-      Router.events.off('hashChangeComplete', updateVisibility)
-    }
   }, [])
```
fabiosantoscode: This needs `anchor` as a dependency so it updates after the first render.

src/components/Documentation/Markdown/index.js
```diff
@@ -1,279 +1,145 @@
...
-  return typeof title === 'string' && SLUG_REGEXP.test(title)
-}
-
-export function extractSlugFromTitle(title) {
-  // extracts expressions like {#too-many-files} from the end of a title
-  const meta = title.match(SLUG_REGEXP)
-
-  if (meta) {
-    return [title.substring(0, meta.index), meta[1]]
+const isInsideCodeBlock = elem => {
```
fabiosantoscode: You might want to have a look at [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest).

src/components/Documentation/index.js
```diff
@@ -1,50 +1,25 @@
...
-export default function Documentation({
-  source,
-  path,
-  next,
-  prev,
-  tutorials,
-  markdown
-}) {
-  const headings = useMemo(() => parseHeadings(markdown))
+  // const headings = useMemo(() => parseHeadings(markdown))
```
fabiosantoscode: Forgotten comment?

src/html.js
fabiosantoscode: Maybe in another PR. No big deal I think.

src/components/Documentation/NewMarkdown/index.js
iAdramelk: Removed it

server.js
```diff
@@ -1,66 +1,16 @@
...
-        res.statusCode = 404
-      }
+app.use(redirectsMiddleware)
+app.use('/api', apiMiddleware)
 
-      // Custom route for all docs
-      app.render(req, res, '/doc', query)
-    } else {
-      // Regular Next.js handler
+app.use(express.static('public'))
```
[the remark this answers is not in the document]
iAdramelk: Added, thanks

gatsby-node.js
```diff
@@ -0,0 +1,106 @@
...
+      matches.push({
+        text: title,
+        slug: slug
+      })
+    }
+  } while (match)
+
+  slugger.reset()
+  return matches
+}
```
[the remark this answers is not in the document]
iAdramelk: Yep, I just copied current implementation for the most part now. We should rework it in separate PR.

jest.config.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
   testEnvironment: 'node',
   transform: {
     '^.+\\.js?$': 'babel-jest'
-  }
+  },
+  testPathIgnorePatterns: ['/node_modules/', '/.cache']
```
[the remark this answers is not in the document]
iAdramelk: Good catch, added it to ignore list.

middlewares/api/index.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-env node */
```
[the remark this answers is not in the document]
iAdramelk: Fixed, thanks!

middlewares/redirects/index.js
```diff
@@ -0,0 +1,32 @@
...
+  if (redirectLocation) {
+    // HTTP redirects
+
+    const queryStr = stringify(query)
+    if (queryStr) {
+      redirectLocation += '?' + queryStr
+    }
+    res.writeHead(redirectCode, {
+      Location: redirectLocation
+    })
```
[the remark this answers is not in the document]
iAdramelk: We should add `res.end()` here, but we still need to run `next()` after that block for the cases that are not redirected and should be handled by static middleware.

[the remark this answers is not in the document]
fabiosantoscode: You still need to return early here though :snowman: . Otherwise, the next middleware in line will run, which we don't want since `res.end()` has already been called.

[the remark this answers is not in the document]
iAdramelk: Yes, you are right, I forgot about it.

.gitignore
```diff
@@ -56,8 +56,11 @@ typings/
...
 .idea
-.next/
 *.log
 
 # Mac finder artifacts
 .DS_Store
+
+# Gatsby cache
+.cache
+public
```
shcheklein: minor: newline

config/prismjs/dvc-commands.js
```diff
@@ -0,0 +1,50 @@
...
+  'remote add',
+  'remote',
+  'push',
+  'pull',
+  'pkg',
+  'pipeline show',
+  'pipeline list',
+  'pipeline',
+  'move',
+  'metrics show',
```
shcheklein: + metrics diff

config/prismjs/dvc-commands.js
```diff
@@ -0,0 +1,50 @@
...
+  'metrics remove',
+  'metrics modify',
+  'metrics add',
+  'metrics',
+  'lock',
+  'install',
+  'init',
+  'import-url',
```
shcheklein: + list

README.md
```diff
@@ -1,4 +1,4 @@
-![owl logo](https://dvc.org/static/img/logo-github-readme.png)
+![owl logo](https://dvc.org/img/logo-github-readme.png)
```
shcheklein: why removing static prefix?

content/docs/install/completion.md
```diff
@@ -134,7 +134,7 @@ $ exec $SHELL -l
 This step is optional but will make the DVC output look much nicer, by adding
 more colors to it. Add the following to your `~/.zshrc`:
 
-```zsh
+```bash
```
shcheklein: why this change? it is zsh, right?

iAdramelk: Basically because both `hljs` and `prsim` actually didn't have syntaxes for `zsh` or `sh` and they fallback for default mode then you use it.

See: https://github.com/conorhastings/react-syntax-highlighter/tree/master/src/languages/hljs
And https://github.com/conorhastings/react-syntax-highlighter/tree/master/src/languages/prism

And we also are not importing them it https://github.com/iterative/dvc.org/blob/master/src/components/Documentation/Markdown/index.js either.

Default mode for `hljs` here is something c-like and is highlight some things as a result. But `prismjs` that `gatsby` uses just leave code unhighlighted.  So here I have two options – either write `zsh` syntax from scratch (but we only use it in this one place) or try to find something close enough to the default `hljs` syntax. I choose later for now.

jorgeorpinel:
Hello! Been trying to catch up with the review here. I have one intro Q. When I click on BLOG in the review app (https://dvc-landing-gatsby-migr-abefpq.herokuapp.com/features) it goes straight to the blog on production. Shouldn't it render the blog in the review app so we can check it since it's all integrated now? Thanks

content/docs/install/completion.md
shcheklein: sounds good, @iAdramelk ! thanks.

gatsby-config.js
```diff
@@ -0,0 +1,100 @@
...
+          options: {
+            enableCustomId: true,
+            isIconAfterHeader: true
+          }
+        }
+      ]
+    }
+  },
+  'gatsby-plugin-catch-links',
+  /*{
```
shcheklein: remove commented code

jorgeorpinel:
> edit on Github still does not work (from https://github.com/iterative/dvc.org/pull/1025#pullrequestreview-370574242)

It goes to master branch so that will be fixed when this is merged, right?

- The DOC link seems to be hard coded to /doc/get-started ? It's /doc currently
- /doc path should open the first doc (currently the Get-Started) without redirecting. It returns a 404 now. This is important for redirects (not just for the DOC link above).

**Redirects**
- How can we test the subdomain redirects? (www, man, error, code/data/remote)
  > Also, we should consider that redirects.test.js is always testing dvc.org, not the review app.
- URL path redirects e.g. /documentation/command-reference add a trailing slash e.g. -> /doc/command-reference/, which causes a 2nd redirect.

> NEXT on https://dvc-landing-gatsby-migr-abefpq.herokuapp.com/doc/command-reference/version does not work

This is still happening. Refers to the image button 🙂

gatsby-node.js
```diff
@@ -0,0 +1,106 @@
+/* eslint-env node */
+
+const path = require('path')
+const GithubSlugger = require('github-slugger')
+
+const { getItemBySource } = require('./src/utils/sidebar')
+
+const slugger = new GithubSlugger()
+
+// Generate hedinga data from markdown
```
shcheklein: typo: headings

package.json
```diff
@@ -43,22 +47,25 @@
...
     "react-collapse": "^5.0.1",
     "react-collapsible": "^2.6.2",
     "react-dom": "^16.12.0",
     "react-ga": "^2.7.0",
+    "react-helmet": "^5.2.1",
     "react-markdown": "^4.2.2",
     "react-popover": "^0.5.10",
     "react-scroll": "^1.7.13",
     "react-slick": "^0.25.2",
     "react-syntax-highlighter": "^11.0.2",
```
shcheklein: do we need this? is there a way to cleanup the dependencies?

iAdramelk: We can safely remove `react-syntax-highlighter` here, yes, as well as `react-markdown` and some other libraries.

src/components/Documentation/Markdown/index.js
```diff
@@ -1,279 +1,145 @@
...
+    <Content id="markdown-root">
+      {tutorials && (
+        <TutorialsWrapper>
+          <Tutorials tutorials={tutorials} compact={true} />
+        </TutorialsWrapper>
+      )}
+      <GithubLink href={githubLink} target="_blank" rel="noreferrer noopener">
+        <i /> Edit on GitHub
+      </GithubLink>
+      <div className="markdown-body">{renderAst(htmlAst)}</div>
```
shcheklein: so, what is the plugin that holding us from the danger html?

iAdramelk: Basically those two blocks: https://github.com/iterative/dvc.org/blob/5f589deee8fa32aca0f8c498947649fa02a87099/src/components/Documentation/Markdown/index.js#L31-L57

shcheklein: it should be straightforward similar to what we did already on the blog side a few times?

src/components/Documentation/Markdown/styles.js
```diff
@@ -31,6 +31,131 @@ export const Content = styled.article`
...
+      position: relative;
+      top: 1px;
+      right: 0;
+      width: 12px;
+      height: 12px;
+      margin-left: 1px;
+      content: url(/img/external-link.svg);
+    }
+
+    pre[class*='language-'] {
```
shcheklein: we will need to share it anyway? move it to a separate file?

src/components/Documentation/Markdown/index.js
iAdramelk: It's a little harder here. We just replaced one html with another in the blog. But for example `<Tooltip>` is a custom react component with a lot of custom logic for positioning. And `<Collapsible>` is using amitations and react state to show and hide content. We can't just replace them with static html.

shcheklein: that's sad ... mostly means JS bundle will be way bigger? did you compare it already? Also means probably that we spent quite a lot of time on the blog side for no reason - they will be using the same JS, right?

src/components/Documentation/SidebarMenu/index.js
```diff
@@ -5,24 +5,37 @@ import scrollIntoView from 'dom-scroll-into-view'
...
+  source,
+  path,
+  activePaths,
+  onClick
+}) {
   const isActive = activePaths && includes(activePaths, path)
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
 
+  // ToDo: Dirty hack to make links with source=false work
```
shcheklein: create a ticket or put it to some other visible place

src/components/Documentation/index.js
```diff
@@ -1,50 +1,24 @@
 /* global docsearch:readonly */
```
shcheklein: why do we have it in some files?

src/components/Footer/index.js
```diff
@@ -16,6 +18,8 @@ import {
   Wrapper
 } from './styles'
 
+const docsPage = getFirstPage()
```
shcheklein: means that `https://dvc-landing-gatsby-migr-abefpq.herokuapp.com/doc/` does not work

src/html.js
shcheklein: please, create a ticket for this

README.md
iAdramelk: Gatsby use /public/static/` folder to store autogenerated images and other assets. I decided that using the same name can cause possible new bugs with name confilcts and can make checking the build results harder, so I removed `/static/` part.

iAdramelk: Oops, my error. Forgot that `gatsby` actually automatically remove it. We still have folder `/static/img` in repo, but `gatsby` copies it's content to the root of the `/public/` folder. I could add nested folder `/static/static/img`, but I decide that it will be strange.

src/components/Documentation/Markdown/styles.js
iAdramelk: We will. but we will first convert `styled-components` to something else. If we move it right now we will still need to rewrite it later so I leave it this way for now.

src/components/Documentation/index.js
iAdramelk: In this case, because we connect `docsearch` here https://github.com/iterative/dvc.org/pull/1025/files#diff-58826c9c508b82875d8f9a9872d19e2dR18 and it register global variable `window.docsearch`. Eslint didn't know about this global variable so we tell him about it here.

src/components/Footer/index.js
iAdramelk: We need to add it to the redirects list, yes. Having two identical pages on different urls probably not good for SEO too.

iAdramelk: Added redirects for all pages with `source: false`.

src/components/Documentation/SidebarMenu/index.js
iAdramelk: Done https://github.com/iterative/dvc.org/blob/5f589deee8fa32aca0f8c498947649fa02a87099/src/components/Documentation/SidebarMenu/index.js#L31

src/components/Documentation/Markdown/index.js
iAdramelk: Not exaclty. We remove around 90% of code and libraries that we have before here: `react-markdown` (21kb minified), `react-syntax-hightlighter` (hard to say exactly, all themes and code is 397kb minified, but we probably used only 50-150kb in our case), `unist-util-visit` and our custom plugins for links and custom syntaxes. It's a lot of code.

We added new library `rehype-react` that have weight of `7.7kb` gzipped which is not perfect, but bearable and increased size of page data json because we are now sending not html, but parsed html ast. But these changes again are partially mitigated by gzip because ast is easy to compress. So in the end we have more code than we could if we just use static html, but significantly less then before migration.

Also blog pages will not be affected by it. Gatsby and next both create separate js-chunks for different page types, so `rehype-react` will only be inside doc page chunk.

package.json
iAdramelk: Ok. Removed `react-syntax-highlighter` but didn't remove `react-markdown`. I forgot that it is used in Tooltip. Created ticket for removal https://github.com/iterative/dvc.org/issues/1045

gatsby-node.js
iAdramelk: Fixed

iAdramelk:
> Hello! Been trying to catch up with the review here. I have one intro Q: When I click on BLOG in the review app (https://dvc-landing-gatsby-migr-abefpq.herokuapp.com/features) it goes straight to the blog on production. Shouldn't it render the blog in the review app so we can check it since it's all integrated now? Thanks

We only converted site itself right now. Blog is still in separate repo. :(



> edit on Github still does not work
> It goes to master branch so that will be fixed when this is merged, right?

Yes, it should work after we merge this PR to master.

gatsby-config.js
iAdramelk: Uncommented it actually, and added favicon, totally forgot about it.

config/prismjs/dvc-commands.js
iAdramelk: Done

config/prismjs/dvc-commands.js
iAdramelk: Done

.gitignore
iAdramelk: Done

src/components/Documentation/index.js
iAdramelk: Yep, removed it

middlewares/redirects/index.js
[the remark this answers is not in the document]
iAdramelk: Done

src/components/Documentation/Markdown/index.js
iAdramelk: Just copied block for now too. Probably we need to update it separately.

src/html.js
iAdramelk: Done https://github.com/iterative/dvc.org/issues/1046

README.md
shcheklein: yeah, my concern is to keep as less URLs "locked" as possible. If we can get it into static it would be better (better w/o static/static of course). Also, I was thinking that we apply some image processor like in blog - to unify it, unify plugins, etc?

src/components/Documentation/Markdown/styles.js
shcheklein: kk, resolving 👍

README.md
iAdramelk: Yep. Perfectly we should move images to corresponding components' folders and import them as other assets, this way will webpack automatically optimize images for us and rename them, move them correctly.

shcheklein: to corresponding "md"-files' folders you mean, right?

iAdramelk: Some to the md files, some to the components (like the icons in `/static/img/community/`).

shcheklein: 👍

package.json
shcheklein: what about other libraries?

src/components/Documentation/Markdown/index.js
shcheklein: yeah, I'm just wondering would have it bern then easier on the blog end to the same with plugins - we spent quite a lot of time writing them ... or am I missing something? at least I know this for the future decisions ...

package.json
iAdramelk: `unist-util-visit` is still used, but now on server instead of client. Same with `isomorphic-fetch` (we still using it for API). Need to check others in more details, but at the first glance we still need them.

src/components/Footer/index.js
shcheklein: > probably not good for SEO too.

yes, it's pretty bad

> We need to add it to the redirects list

make sure that it is 302 or 307 or whatever we use for temporary ones - we don't want them to be cached. we need to have a separate index page for dvc.org/doc

src/components/Documentation/index.js
shcheklein: 👍

src/components/Documentation/Markdown/index.js
iAdramelk: Hard to say for sure. Simple tag replacement can probably be made simpler in browser but would increase bundle size. But stuff like that we did for image sizes are better implemented on the server.

src/components/Footer/index.js
iAdramelk: Added 307 status code.

fabiosantoscode:
I've noticed the preview deployment doesn't include the latest cache headers. Is this just a lack of rebasing, or was something lost in a merge?

The `cache-control` header needs a `s-maxage=999999` bit, so that the browser can validate against cloudflare, which is faster and doesn't hit our supple express server.

After each deployment, cloudflare is cleared (`npm start` runs `scripts/clear-cloudflare-cache.js`), so there should be no problem with this.

fabiosantoscode:
I also noticed the 404 page is `text/plain`. Gatsby should output a `public/404.html` file, which can be used to handle 404's.

I think this can be done with an express error handler at the end of the chain. As far as I can see `express.static` doesn't provide an option for a 404 page.

pavelgrinchenko:
> I've noticed the preview deployment doesn't include the latest cache headers. Is this just a lack of rebasing, or was something lost in a merge?

Added `Cache-Control` header

> I also noticed the 404 page is text/plain. Gatsby should output a public/404.html file, which can be used to handle 404's.

Added 404 page which is similar to the one in the blog

src/components/Nav/index.js
```diff
@@ -4,10 +4,14 @@ import PropTypes from 'prop-types'
 import LocalLink from '../LocalLink'
 
 import { logEvent } from '../../utils/ga'
+import { getFirstPage } from '../../utils/sidebar'
+
+const docsPage = getFirstPage()
```
shcheklein: do we still need this? it should be always `dvc.org/doc`

src/pages/404.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
```
shcheklein: do we need both 404 pages? I thought I saw the second one ..

shcheklein: @pavelgrinchenko ? ^^

src/components/Nav/index.js
shcheklein: @pavelgrinchenko ? ^^

README.md
shcheklein: @iAdramelk please, create a ticket for this as well

src/pages/404.js
pavelgrinchenko: Do you mean `src/pages/404.js` and `src/components/404/`?
In first case it file for Gatsby which tells to it to generate page `404.html`. In the second case it's React component which will be rendered on this page, with all styles that it has. Other pages in the `src/pages' folder have the same structure.

src/components/Nav/index.js
pavelgrinchenko: I tried to remove this helper and rely on server.js redirects, but they work just for the initial load. For example, we load `/doc` page in the browser - in this case, server.js redirects us to the correct page. But if we are navigating from another page of dvc.org then Gatsby's router will intercept this request and show the 404 page, because in this case, it doesn't know about any redirects on the server side.
@iAdramelk told me that you are working on an index page for `/doc`. So we can wait for it and then remove this helper.
Another solution is to use something like [this plugin](https://www.gatsbyjs.org/packages/gatsby-plugin-client-side-redirect/), which will create all redirects on the client-side too, to have identical behavior on both sides: client and server.

shcheklein: @pavelgrinchenko makes sense. it might be that we have a problem for other similar pages (we have 3 or 4 pages like this) if we have links to them in the docs?

> Another solution is to use something like this plugin, which will create all redirects on the client-side too, to have identical behavior on both sides: client and server.

I don't like the, but they might be good as a fallback for those sections we won't have index pages for some reason.

src/components/DocLayout/SidebarMenu/index.js
```diff
@@ -76,14 +76,18 @@ export default function SidebarMenu({ id, sidebar, currentPath, onClick }) {
       })
     }
 
-    const node = document.getElementById(currentPath)
+    const node = document.getElementById(currentPath.replace(/\/$/, ''))
```
fabiosantoscode: This should be handled at the serverside I believe. It should act as the gatekeeper which prevents us from having to deal with the trailing-slash-case on the client.

src/pages/404.js
shcheklein: 👍

src/components/DocLayout/SidebarMenu/index.js
fabiosantoscode: We can add a redirect to redirects-list.json which redirects from `^(.+)/$` to `$1` and that should do it. We never want trailing slashes.

@iAdramelk told me this breaks the static server even if its [redirects](https://github.com/expressjs/serve-static#redirect) option is disabled, but we can always use [serve-handler](https://npmjs.com/serve-handler). The choice doesn't matter much. This is all going in S3 eventually.

README.md
iAdramelk: Done https://github.com/iterative/dvc.org/issues/1063

shcheklein: thanks!

shcheklein:
> /doc path redirects to /doc/get-started ... The DOC link still seems to be hardcoded to /doc/get-started

@jorgeorpinel there is a ticket to handle all of this after the merge - https://github.com/iterative/dvc.org/issues/1060 - please chime in here. 

> How can we test the subdomain redirects? (www, man, error, code/data/remote)

we will deploy and test them

> which causes a 2nd redirect to the final location: /doc/command-reference/

will be handled when we merge dvc.org with blog the way I understand this cc @fabiosantoscode @iAdramelk @pavelgrinchenko

src/components/DocLayout/SidebarMenu/index.js
pavelgrinchenko: I'll take a look a `serve-handler` tomorrow.
But we also discussed this issue [yesterday at Slack](https://evilmartians.slack.com/archives/CLN9J8Q8K/p1584377665016000?thread_ts=1584348448.007700&cid=CLN9J8Q8K) and decided that it can be resolved after merging with blog's codebase.

iAdramelk:
@jorgeorpinel 

> /doc path redirects to /doc/get-started. Again, what if we want to change the first doc to something else? Would we have to change the redirects file every time?

It's actually not hardcoded now. It resolves to the first item with the source in the `sidebar.json`.

jorgeorpinel:
> It's actually not hardcoded now

I'm seeing it hardcoded in redirects-list.json @iAdramelk

iAdramelk:
> I'm seeing it hardcoded in redirects-list.json

It's hardcoded in redirects for now (to redirect users who hit urls from search engines or bookmarks), but actual links in html and site header/footer are not hardcoded. They are rosolved using sidebar.json. 

But you are right, it's not an ideal situation and it can break if we change root sidebar.json right now without changing redirects.

PR #908, opened 2020-01-08
api: create docs

jorgeorpinel:
Closes #463 

> See https://github.com/iterative/dvc.org/issues/463#issuecomment-514139792 for initial spec.

- [x] Matching core update(s): iterative/dvc/pull/3352, iterative/dvc/pull/3426

PR #1027, opened 2020-02-27
cmd ref: review option defaults and sentence priority per API params

jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/908#pullrequestreview-356423629
> and https://github.com/iterative/dvc.org/pull/908#pullrequestreview-356423672

### DIFF
--- a/public/static/docs/command-reference/run.md
+++ b/public/static/docs/command-reference/run.md
@@ -35,10 +35,10 @@ instead of `"` to wrap the `command` if there are environment variables in it,
 that you want to be evaluated dynamically. E.g.
 `dvc run -d script.sh './myscript.sh $MYENVVAR'`
 
-Unless the `-f` options is used, by default the DVC-file name generated is
-`<file>.dvc`, where `<file>` is file name of the first output (`-o`, `-O`, `-m`,
-or `-M` option). If neither `-f`, nor outputs are specified, the stage name
-defaults to `Dvcfile`.
+Unless the `-f` options is used, the stage file (DVC-file) is generated in the
+current working directory and named `<file>.dvc`, where `<file>` is file name of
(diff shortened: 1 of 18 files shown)


public/static/docs/command-reference/run.md
```diff
@@ -141,18 +144,14 @@ data pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
...
   confirmation.
 
-- `--ignore-build-cache` - has an effect if an equivalent stage file exists
-  (same dependencies, outputs, and `command` to execute), that has been already
-  executed, and is up to date. In this case, `dvc run` won't normally execute
-  the `command` again. The exception is when the existing stage is considered
-  always changed (see `--always-changed` option). This option gives a way to
-  forcefully execute the `command` anyway. Useful if the command's code is
+- `--ignore-build-cache` - forcefully execute the `command` again, even if an
+  equivalent stage file exists. Useful if the command's code is
```
shcheklein: I would clarify what do we mean by equivalent - same list of same versions of deps, same list of outs, same command. Even if it's part of the different pipeline.

jorgeorpinel: "equivalent stage file" is defined exactly in the description already, so I didn't also add it here, to keep the paragraph not-so-long. Should we repeat it anyway?

jorgeorpinel: > I have an "equivalent" (hehe) question for the api docs PR. Should we repeat default behavior in each option paragraph, even when they are explained in the definition?

jorgeorpinel: p.s. in this case I'm changing the explanation a little to "even if the
  same `dvc run` command has already been run in this workspace"... See 46eb0551

PR #1029, opened 2020-02-28
cmd ref: command options impros and consistency

jorgeorpinel:
> Kind of continues #1027
- Addresses https://github.com/iterative/dvc.org/pull/1027#pullrequestreview-366058890
- gc goes before get (in nav sidebar)
- Review mentioning option values e.g. `--file FILE`
- Remove unnecessary default value info.


public/static/docs/command-reference/add.md
```diff
@@ -99,19 +99,19 @@ reproducible.
   analogous to using `git add` before `git commit`. Use `dvc commit` when ready
   to commit the results to cache.
 
+- `-f FILE`, `--file FILE` - specify name of the DVC-file it generates. This
```
shcheklein: better to use use `path`

public/static/docs/command-reference/fetch.md
```diff
@@ -74,8 +73,7 @@ by `dvc fetch` (unless the `-a` or `-T` options are used).
 
 - `-r REMOTE`, `--remote REMOTE` - name of the
```
shcheklein: better to use NAME?

public/static/docs/command-reference/add.md
jorgeorpinel: Hmmm. But this comes from the core repo (command help output):

```console
$ dvc add -h
usage: dvc add [-h] [-q | -v] [-R] [--no-commit] [-f FILE]
               targets [targets ...]
```

public/static/docs/command-reference/fetch.md
jorgeorpinel: Same:
```console
$ dvc fetch -h
usage: dvc fetch [-h] [-q | -v] [-j JOBS] [-r REMOTE] [-a] [-T] [-d] [-R]
                 [targets [targets ...]]
```

jorgeorpinel:
> name these params to match some "user space terminology". Especially, --outs-no-cache OUTS_NO_CAHCE looks strange and not very meanigful

Wile I agree with you, these are the argument value names in the help output:

```console
$ dvc run -h
usage: dvc run [-h] [-q | -v] [-d DEPS] [-o OUTS] [-O OUTS_NO_CACHE]
               [-m METRICS] [-M METRICS_NO_CACHE] [-f FILE] [-c CWD] [-w WDIR]
               ...
```
Should I change all these both here and in a corresponding PR to the core repo?

shcheklein:
@jorgeorpinel in that snippet it makes a little bit more sense, since it does not include the longer version of the option I suppose. So, probably some improvements can be made but I'm not sure how would the look like.

jorgeorpinel:
OK. If this is your only concern here, can we consider merging this and open a separate issue in the core repo about these UPPER_CASE_ARG_VALUE literals printed by some command -h help outputs?

shcheklein:
Why would merge something that we want to change even in the core repo? It's better to just open the issue in the core repo first, decide on how (if we need any at all?) do we show the accepted values.

Alternatively, we can analyze the best practices (e.g. Git itself, Heroku CLI, etc) - how do they name, show accepted values and start driving the change from the docs side.

jorgeorpinel:
> Why would merge something that we want to change even in the core repo?

Because merging it is an improvement from the current published version, that matches the DVC version in prod.

I checked the Git cmd ref and they do the same as this PR, except the value placeholders have nicer names (descriptive and not in ALL_CAPS) i.e. `git checkout -b <new_branch>`. The same full text used when explaining the option:
image
> https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnewbranchgtltstartpointgt

Heroku's case is quite different. They use the exact same value name as the option name, but assign with `=`. They don't have a nice cmd ref, it's just a dump of all of their help outputs (which are pretty good):
image
> https://devcenter.heroku.com/articles/heroku-cli-commands

I think we should decide one Ill send it as a suggestion PR to core, so the engineering team can give feedback, I vote for Git's way e.g. `dvc run ... [-o <outputs>]` + keep the same naming in the Options section bullets, as this PR does now.

WDYT @shcheklein ?

shcheklein:
> Because merging it is an improvement from the current published version, that matches the DVC version in prod.

so, this is my concern. Matching DVC != docs/readability improvement.

The idea itself to show values is very valuable, it's just really really painful to see things like:

`-M METRICS_NO_CACHE`, `--metrics-no-cache METRICS_NO_CACHE`

it's a lot of duplication, upper case makes it hard to read, etc

If want to do this, I would really come up with a better (Heroku or Git - both are fine to me) approach, implement it and show this to the core team.

public/static/docs/command-reference/add.md
jorgeorpinel: Discussing in https://github.com/iterative/dvc.org/pull/1029#issuecomment-594911780 and below.

jorgeorpinel: Resolving here.

public/static/docs/command-reference/fetch.md
jorgeorpinel: Resolving as well. See Discussing in https://github.com/iterative/dvc.org/pull/1029#issuecomment-594911780 and below.

jorgeorpinel:
OK I can fix it here and send a simultaneous PR to the core repo. I'm a bit concerned about conflicts with regular updates and this PR so it would be ideal to merge ASAP but let me fix this first... ⏳ 

> The idea itself to show values is very valuable...
> If want to do this

We agree we want to do this right?

shcheklein:
> We agree we want to do this right?

yes. in this `I would really come up with a better (Heroku or Git - both are fine to me) approach` sense. Just having `-M METRICS_NO_CACHE, --metrics-no-cache METRICS_NO_CACHE` is not a good approach.

also, would not consider this p1.

public/static/docs/command-reference/run.md
```diff
@@ -79,10 +79,10 @@ data pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
 
 ## Options
 
-- `-d`, `--deps` - specify a file or a directory the stage depends on. Multiple
-  dependencies can be specified like this: `-d data.csv -d process.py`. Usually,
-  each dependency is a file or a directory with data, or a code file, or a
-  configuration file. DVC also supports certain
+- `-d <path>`, `--deps <path>` - specify a file or a directory the stage depends
```
shcheklein: just to check - can it be a comma-separated list?

jorgeorpinel: Not for `--deps`, but it can be in other options. That is shown in the Synopsis though, not here. I think the only one is `dvc gc -p.`

jorgeorpinel: Please merge this if you're satisfied with the answer 🙂

shcheklein: yeah, that's right  ... it's more about options that could be specified once vs those that could be specified multiple time ... not sure if we need a special syntax to highlight that

shcheklein:
I approved this already, so feel free to merge! It's a good improvement.

public/static/docs/command-reference/run.md
jorgeorpinel: Ah good question. I didn't even think about that but it seems like a relevant thing to note somewhere. Opened #1041

PR #1032, opened 2020-03-03
suggest some minor things to API

shcheklein:

Please chose to [allow us](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to edit your branch when creating the PR.

### DIFF
--- a/public/static/docs/api-reference/get_url.md
+++ b/public/static/docs/api-reference/get_url.md
@@ -74,7 +74,7 @@ or
 
 - `dvc.exceptions.NoRemoteError` - no `remote` is found.
 
-## Example: Getting URL to a DVC-tracked file
+## Example: Getting the URL to a DVC-tracked file
 
 ```py
 import dvc.api
--- a/public/static/docs/api-reference/get_url.md
+++ b/public/static/docs/api-reference/get_url.md
@@ -4,33 +4,42 @@ Returns the URL to the storage location of a data file or directory tracked in a
 <abbr>DVC project</abbr>.
 
 ```py
-dvc.api.get_url( path: str, repo: str = None,
-    rev: str = None, remote: str = None ) -> str
+def get_url(path: str,
+            repo: str = None,
+            rev: str = None,
+            remote: str = None) -> str
 ```
 
-**Usage:**
+#### Usage:
 
 ```py
 import dvc.api
 
-resource_url = dvc.api.get_url(
+dvc.api.get_url(
     'get-started/data.xml',
     repo='https://github.com/iterative/dataset-registry')
+
+# https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e10355
 ```
 
 ## Description
 
 Returns the URL string of the storage location (in a
 [DVC remote](/doc/command-reference/remote)) where a target file or directory,
-specified by its `path` in a `repo` (<abbr>DVC project</abbr>), is stored. If
-`repo` is not a DVC project, a `dvc.api.UrlNotDvcRepoError` is raised.
+specified by its `path` in a `repo` (<abbr>DVC project</abbr>), is stored.
 
 The URL is formed by reading the the project's
 [remote configuration](/doc/command-reference/config#remote) and the
-[DVC-file](/doc/user-guide/dvc-file-format) where the given `path` is an output
-(see the **Examples** below). The URL schema returned depends on the
+[DVC-file](/doc/user-guide/dvc-file-format) where the given `path` is an
+<abbr>output</abbr> The URL schema returned depends on the
 [type](/doc/command-reference/remote/add#supported-storage-types) of the
 `remote` used (see the [Parameters](#parameters) section).
+
+If the target is a directory, the returned URL will point to a file that
+contains the mapping of files in the directory (as a JSON array), along with
+their hash values. Refer to
+[Structure of cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
+and `dvc add` to learn more about how DVC handles data directories.
 
 ⚠️ This function does not check for the actual existence of the file or
 directory in the remote storage.
@@ -41,39 +50,31 @@ an appropriate library, such as
 or
 [`paramiko`](https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.get).
 
-Note that if the target is a directory, the URL will end in `.dir`, as DVC
-stores a special JSON file with `.dir` extension that contains the mapping of
-files in the directory (as a JSON array), along with their hash values. Refer to
-[Structure of cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
-and `dvc add` to learn more about how DVC handles data directories.
-
 ## Parameters
 
 - **`path`** - location and file name of the file or directory in `repo`,
   relative to the project's root.
 
 - `repo` - specifies the location of the DVC project. It can be a URL or a file
-  system path. Both HTTP and SSH protocols are supported for online Git repos
-  (e.g. `[user@]server:project.git`).
-
-  The current project is used by default if a `repo` argument is not given (the
-  current working directory tree is walked up to find it).
+  system path. Both HTTP and SSH protocols are supported for online Git
+  repositories (e.g. `[user@]server:project.git`). _Default_: The current
+  project is used (the current working directory tree is walked up to find it).
 
 - `rev` - Git commit (any [revision](https://git-scm.com/docs/revisions) such as
   a branch or tag name, or a commit hash). If `repo` is not a Git repo, this
-  option is ignored.
-
-  `HEAD` is used by default if a `rev` argument is not given.
+  option is ignored. _Default_: `HEAD`.
 
 - `remote` - name of the [DVC remote](/doc/command-reference/remote) to use to
-  form the returned URL string.
+  form the returned URL string. _Default_: The
+  [default remote](/doc/command-reference/remote/default) of `repo` is used.
 
-  The [default remote](/doc/command-reference/remote/default) of `repo` is used
-  if a `remote` argument is not given.
+## Exceptions
 
-  A `dvc.exceptions.NoRemoteError` is raised if no `remote` is found.
+- `dvc.api.UrlNotDvcRepoError` - `repo` is not a DVC project.
 
-## Example: Get the remote storage URL to a data file
+- `dvc.exceptions.NoRemoteError` - no `remote` is found.
+
+## Example: Getting URL to a DVC-tracked file
 
 ```py
 import dvc.api
@@ -82,10 +83,13 @@ resource_url = dvc.api.get_url(
     'get-started/data.xml',
     repo='https://github.com/iterative/dataset-registry'
     )
+
 print(resource_url)
 ```
 
----
+It prints
+`https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e10355` if
+we run it:
 
 ```dvc
 $ python script.py
@@ -107,7 +111,7 @@ and the project configuration
 ([`.dvc/config`](https://github.com/iterative/dataset-registry/blob/master/.dvc/config))
 where the remote URL is saved:
 
-```dvc
+```ini
 ['remote "storage"']
 url = https://remote.dvc.org/dataset-registry
 ```


public/static/docs/api-reference/get_url.md
```diff
@@ -4,34 +4,43 @@ Returns the URL to the storage location of a data file or directory tracked in a
...
 
 ```py
 import dvc.api
 
-resource_url = dvc.api.get_url(
+dvc.api.get_url(
     'get-started/data.xml',
     repo='https://github.com/iterative/dataset-registry')
+
+# https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e10355
```
jorgeorpinel: Agee about removing the var name but its not very obvious what the comment below is (and probably creates a horizontal scroll). Not sure it's needed for basic usage.

public/static/docs/api-reference/get_url.md
```diff
@@ -4,34 +4,43 @@ Returns the URL to the storage location of a data file or directory tracked in a
...
 `remote` used (see the [Parameters](#parameters) section).
 
+If the target is a directory, the returned URL will point to a file that
+contains the mapping of files in the directory (as a JSON array), along with
+their hash values. Refer to
+[Structure of cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
+and `dvc add` to learn more about how DVC handles data directories.
+
 ⚠️ This function does not check for the actual existence of the file or
 directory in the remote storage.
```
jorgeorpinel: I like this note but I think I had a similar one originally and we agreed it wasn't really about this function so we removed it. It's a little confusing given the note coming after (that actually who knows whether the file exists). Advanced DVC users should know this anyway if they need to know it? No strong opinion.

public/static/docs/api-reference/get_url.md
shcheklein: I think it's intuitively clear ... we can even bring var name back in the comment ... we need to show something than nothing

we can just add 

`# returns 'http://....'

public/static/docs/api-reference/get_url.md
jorgeorpinel: Wait I see this was just moved from further down haha. Ok let me think... ⏳

shcheklein: you didn't remove it actually )

jorgeorpinel: OK to move it but the original text is a little better I think, because it doesn't imply that the .dir file will exist (and contain "the mapping of files...").

shcheklein: not sure I got the point ...

shcheklein: I just wanted to avoid to many details (like .dir, etc - it's internal stuff)

public/static/docs/api-reference/get_url.md
```diff
@@ -41,39 +50,31 @@ an appropriate library, such as
...
 
 - `repo` - specifies the location of the DVC project. It can be a URL or a file
-  system path. Both HTTP and SSH protocols are supported for online Git repos
-  (e.g. `[user@]server:project.git`).
-
-  The current project is used by default if a `repo` argument is not given (the
-  current working directory tree is walked up to find it).
+  system path. Both HTTP and SSH protocols are supported for online Git
+  repositories (e.g. `[user@]server:project.git`). _Default_: The current
+  project is used (the current working directory tree is walked up to find it).
```
jorgeorpinel: Question: Is "Deault:" and "...if argument is not given" the same? I guess so but perhaps it's confusing to call this _default_ when the signature shows `None` is the default value? And is "Default:" in the same p better or clearer than a separate paragraph? I'm not sure.

Again no strong opinion here. Up to you but we should update the params in all the API refs.

public/static/docs/api-reference/get_url.md
```diff
@@ -41,39 +50,31 @@ an appropriate library, such as
...
 
 - `rev` - Git commit (any [revision](https://git-scm.com/docs/revisions) such as
   a branch or tag name, or a commit hash). If `repo` is not a Git repo, this
-  option is ignored.
-
-  `HEAD` is used by default if a `rev` argument is not given.
+  option is ignored. _Default_: `HEAD`.
```
jorgeorpinel: I do like how much shorter this one is.

public/static/docs/api-reference/get_url.md
jorgeorpinel: p.s. https://github.com/iterative/dvc.org/pull/1032/files#r386754413

jorgeorpinel: p.s. https://github.com/iterative/dvc.org/pull/1032#pullrequestreview-367651464

public/static/docs/api-reference/get_url.md
```diff
@@ -41,39 +50,31 @@ an appropriate library, such as
...
 
-  The [default remote](/doc/command-reference/remote/default) of `repo` is used
-  if a `remote` argument is not given.
+## Exceptions
 
-  A `dvc.exceptions.NoRemoteError` is raised if no `remote` is found.
+- `dvc.api.UrlNotDvcRepoError` - `repo` is not a DVC project.
 
-## Example: Get the remote storage URL to a data file
+- `dvc.exceptions.NoRemoteError` - no `remote` is found.
```
jorgeorpinel: What's the motivation behind this new section? I like that it's structured and easy to find (via the right hand nav) but it's also good to have the exceptions in their context (where they were before, in the description).

public/static/docs/api-reference/get_url.md
```diff
@@ -41,39 +50,31 @@ an appropriate library, such as
...
+
+## Example: Getting URL to a DVC-tracked file
```
jorgeorpinel:
```suggestion
## Example: Getting the URL to a DVC-tracked file
```

public/static/docs/api-reference/get_url.md
```diff
@@ -82,10 +83,13 @@ resource_url = dvc.api.get_url(
     'get-started/data.xml',
     repo='https://github.com/iterative/dataset-registry'
     )
+
 print(resource_url)
```
jorgeorpinel: Kind of like it better without the extra new line 😋

public/static/docs/api-reference/get_url.md
```diff
@@ -82,10 +83,13 @@ resource_url = dvc.api.get_url(
...
 ```
 
----
+It prints
+`https://remote.dvc.org/dataset-registry/a3/04afb96060aad90176268345e10355` if
+we run it:
 
 ```dvc
 $ python script.py
```
jorgeorpinel: This is repetitive. Maybe just something like "Prints the URL string value if we run it from terminal:" ?

public/static/docs/api-reference/get_url.md
jorgeorpinel: I don't think the basic usage block needs to show the variable value. That's what examples are for I think.

But if you want to keep this I would not reintroduce var name, maybe just make the comment shorter? `# https://remote.dvc.org/dataset-registry/a3/04a...` (maybe in the example BTW, so that no scrolls are shown) but not a huge deal about the scrolls for these cases.

shcheklein: I think it makes this example significantly better if we show the value, I would even say that this the most important part  - you see right away what it does

I would keep at as simple as possible, would add `returns` at most

public/static/docs/api-reference/get_url.md
shcheklein: I think it's clear .. also I think you used the phrase by default. I didn't change the meaning - just made it shorter. I'm fine to keep them each on a separate line, but with some common structure and highlighting _Default_: or something else

public/static/docs/api-reference/get_url.md
jorgeorpinel: I think mentioning the .dir ending is appropriate as it may be something the user notices and wonders about when using this fn.

```suggestion
If the target is a directory, the returned URL will end in `.dir`: a file that
contains the mapping of files in the directory. Refer to
[Structure of cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
and `dvc add` to learn more about how DVC handles data directories.

⚠️ This function does not check for the actual existence of the file or
directory in the remote storage.
```

public/static/docs/api-reference/get_url.md
shcheklein: Yep, you can find them when you need them ... and you don't need them often ... it makes it easier to read what's important

shcheklein: also, we need to check - there should be other relevant exception here

public/static/docs/api-reference/get_url.md
shcheklein: it just makes it easier to read this way ... probably no need in saving lines :) up to you though

public/static/docs/api-reference/get_url.md
shcheklein: Again, it's easier to read. Repetition not a big problem. Code snippet just illustrates the code (and gives a command to run). I would keep it extra explicit. Up yo you again. Don't have a strong opinion

public/static/docs/api-reference/get_url.md
shcheklein: they should not care about such specifics, we might change schema for those URL ... we can mention something like `(ends with `.dir`)` - but that should be enough

public/static/docs/api-reference/get_url.md
jorgeorpinel: OK. Actually I would reintroduce the var name now that I think about it. You wouldn't call a string returning function without assigning it to something. So my full suggestion is:

```py
resource_url = dvc.api.get_url(
    'get-started/data.xml',
    repo='https://github.com/iterative/dataset-registry')
    
# resource_url = https://remote.dvc.org/dataset-registry/a3/04a...
```

jorgeorpinel: Or wrap in `print()`

public/static/docs/api-reference/get_url.md
jorgeorpinel: > we can mention something like (ends with .dir) - but that should be enough 

Good idea. Keeping the last sentence too (link to dvc dir struct), I assume.

public/static/docs/api-reference/get_url.md
jorgeorpinel: No, its OK. I'm convinced here now. Resolving.

public/static/docs/api-reference/get_url.md
jorgeorpinel: Yep, please check this comment: https://github.com/iterative/dvc.org/pull/908#discussion_r386204040

public/static/docs/api-reference/get_url.md
jorgeorpinel: Too minor. The option would be to wrap in `print()` (esp. if we do the same in the basic usage section on top). But resolving this...

public/static/docs/api-reference/get_url.md
jorgeorpinel: > Repetition not a big problem.

Yeah but why do it if it's not needed? Just keeping the top paragraph and end it in "if we run it.", and removing the console block code should be enough I think.

> The original one I did with the `---` hr was also kind of obvious, like in the basic usage comment 🙂

shcheklein: > The original one I did with the --- hr was also kind of obvious, like in the basic usage comment 🙂 and I like how it looked rendered

hr looks like an overkill ... it's a split of the whole document, not very suitable to split example and it's output

public/static/docs/api-reference/get_url.md
shcheklein: don't use `=` though, write it in a plain language - resource_url is now 'http:///'

shcheklein: otherwise it looks like a commented code

public/static/docs/api-reference/get_url.md
jorgeorpinel: OK, true.

jorgeorpinel:
I'm going to merge this and address the remaining stuff in the `api` branch (#908 PR).

jorgeorpinel:
Done! In 8089423d.

PR #1036, opened 2020-03-05
deployment: set cache headers

fabiosantoscode:
Set some cache headers in dvc.org, such that we can set cloudflare to respect the headers it gets.

These headers are for instant invalidation, making the client validate against cloudflare, but preventing cloudflare from validating against the server. After each deploy, the cache is cleared.


shcheklein:
@fabiosantoscode CI is failing

PR #1037, opened 2020-03-08
api ref: better explanation on disc and memory usage for read/open

jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/908#pullrequestreview-369254697

> super relevant example would to show SAX or StAX parser instead of a DOM one - that's where it shines. Or we can make CSV example the main one and show how we process it in steam fashion (e.g. calculating sum or avg) - it would show the "streaming" aspect of the open() way better.

### DIFF
--- a/public/static/docs/api-reference/open.md
+++ b/public/static/docs/api-reference/open.md
@@ -91,7 +91,7 @@ streamed, which optimizes memory usage.
 ## Example: Use data or models from DVC repositories
 
 Any <abbr>data artifact</abbr> hosted online can be processed directly in your
-Python app with this API. For example, an XML file tracked in a public DVC repo
+Python code with this API. For example, an XML file tracked in a public DVC repo
 on Github can be processed like this:
 
 ```py
--- a/public/static/docs/api-reference/open.md
+++ b/public/static/docs/api-reference/open.md
@@ -45,9 +45,8 @@ file can be tracked by DVC or by Git.
 This function makes a direct connection to the
 [remote storage](/doc/command-reference/remote/add#supported-storage-types)
 (except for Google Drive), so the file contents can be streamed as they are
-read. This means it does not require disc space to save the file before making
-it accessible. The only exception is when using Google Drive as
-[remote type](/doc/command-reference/remote/add#supported-storage-types).
+downloaded. No disc space and very little memory are needed to save the file
+before making it accessible.
 
 ## Parameters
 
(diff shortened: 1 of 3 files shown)


jorgeorpinel:
For now I've implemented a SAX XML example in the doc, but let's continue the discussion on what we want to show in examples.

What's the advantage of streaming files in open/read? Probably just making a big file available quickly so you can start processing it before it's all downloaded.

> make CSV example the main one and show how we process it in steam fashion (e.g. calculating sum or avg

I don't think you'd want to show the progress of a real-time processing, or is that a major use case you guys see?

public/static/docs/api-reference/open.md
```diff
@@ -45,8 +45,8 @@ file can be tracked by DVC or by Git.
 This function makes a direct connection to the
 [remote storage](/doc/command-reference/remote/add#supported-storage-types)
 (except for Google Drive), so the file contents can be streamed as they are
-read. This means it does not require space on the disc to save the file before
-making it accessible. The only exception is when using Google Drive as
+read. This means it does not require disc space to save the file before making
```
shcheklein: are read -> is read? (similar to data, it's strange to see `contents are`)

jorgeorpinel: "file contents are" seems correct to me. I can change it to "file content is" though. Same meaning

public/static/docs/api-reference/open.md
```diff
@@ -39,15 +39,14 @@ file can be tracked by DVC or by Git.
...
-
 This function makes a direct connection to the
 [remote storage](/doc/command-reference/remote/add#supported-storage-types)
-(except for Google Drive), so the file contents can be streamed as they are
-read. This means it does not require space on the disc to save the file before
-making it accessible. The only exception is when using Google Drive as
-[remote type](/doc/command-reference/remote/add#supported-storage-types).
+(except for Google Drive), so the file contents can be streamed. Your code can
+process the data [buffer](https://docs.python.org/3/c-api/buffer.html) as it's
+streamed, which optimizes memory usage.
```
shcheklein: 👍

public/static/docs/api-reference/open.md
```diff
@@ -91,46 +90,55 @@ making it accessible. The only exception is when using Google Drive as
...
+    parse(fd, mySAXHandler)
 ```
 
-> Notice that if you just need to load the complete file contents to memory, you
-> can use `dvc.api.read()` instead:
+Notice that we use a [SAX](http://www.saxproject.org/) XML parser here because
+`dvc.api.open()` is able to stream the data download. (The `mySAXHandler` object
+should handle the event-driven parsing of the document in this case.) This
+increases the performance of the code (minimizing memory usage), and is
+typically faster than loading the whole data into memory.
```
shcheklein: 👍

jorgeorpinel:
Woot woot 🎉 

Cc @Suor feel fee to review this port-merge. Thanks!

PR #1043, opened 2020-03-13
Typo fix: remove apostrophe from possessive "its".

mfickett:
Used the **Edit on GitHub** button from dvc.org to improve a doc in place.


shcheklein:
thanks @mfickett ! 🎉

PR #1049, opened 2020-03-15
Regular updates (mid Mar)

jorgeorpinel:
- missing instances of "option" term
- and option value names
- polish `dvc list` ref per https://github.com/iterative/dvc.org/pull/1021?email_token=AALIXH555APWCCHURFFRFBTRGGTB7A5CNFSM4K2Z3RC2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEODLWOA#issuecomment-596032312

PR #1050, opened 2020-03-15
use-cases: improve mention of `dvc list` in data registry case

jorgeorpinel:
Address feedback from #1049


shcheklein:
there is a conflict, please resolve and feel free to merge! LGTM

Issue #599, opened 2019-08-30
get-started: rewrite it (reach "next level") for user interaction


jorgeorpinel:
~~After #487~~ ✔️ 

- [x] Review validity of open [get-started issues](https://github.com/iterative/dvc.org/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3Adoc-content+get-started) when the above is addressed.
- [x] Consider these comments about the current Get Started: https://discuss.dvc.org/t/about-get-started/194
- [x] Include #634 as well (replace the orphan stage from add-files for an import stage)

shcheklein:
@dashohoxha and @dmpetrov and our users brought good points that it's still not clear from the get started that basic data versioning can be used independently from pipelines. I think it's a good time to try to split into segments - Versioning, Accessing (import/get), Pipelines (aka Reproducibility?), Experiments (metrics, etc, etc).

The challenge here is to make them simple, easily reproducible in a sense that we need to provide commands to run. Ideally it still should be base on one project so that you can all these new section one by one.

The things that should change for example, is that Get Old Files should become a part of the first part where basic data management is described. So, we would need to have two versions of the data file at this point.

What options do we have? One singe project, multiple different project that server different needs?

shcheklein:
@jorgeorpinel please, feel free to update and change the ticket (checkboxes) accordingly :)

jorgeorpinel:
> I think it's a good time to try to split into segments - Versioning, Accessing (import/get), Pipelines...
> The challenge here is to make them simple...

I don't know if I agree with this. That's what we have tutorials and other docs for. What is a Get Started for? I think it should be a basic introduction to some key features so a user can quickly get familiar with DVC. If its getting to complicated or confusing, we should probably just remove some of the chapters and simplify others (like ** Import Data** perhaps).

> - [ ] **Get Older Files** (currently the last chapter) should become a part of (the first part where basic data management is described)

Maybe the name or description could be improved but in the current project at least, the only file that has more than one version is `model.pkl`, the last output in the whole pipeline, so that's why this chapter has to be at the end.

> update and change the ticket (checkboxes) accordingly

Looking at them now.

shcheklein:
> I think it should be a basic introduction to some key features so a user can quickly get familiar with DVC.

[...]

PR #1051, opened 2020-03-15
get-started: initial refactoring

### LINKED ISSUE #747: docs: move /get-started index to root and /install above /get-started
(Probably after #738)

I like that **Get Started** is the first section in our docs, but ~~its 2nd page links to **Installation** for the user to go and figure out how to best install the tool. This may kind of break the continuity of get-started from the very beginning.~~ Perhaps if we have **Installation** on top in the left side bar, most users will already have DVC installed by the time they get to the Get Started -> Initialize page.

jorgeorpinel:
- Closes #747 (dedicated landing page for docs)
- Addresses part of #1035 (merge get-started and tutorials sections)


jorgeorpinel:
@shcheklein so far this only implements #747 but probably a little hacky – see `getItemByPath` in src/utils/sidebar.js and `SidebarMenu.timeout` in src/components/Documentation/SidebarMenu/index.js. Can you guys look at it and lmk what a better way to implement this could be? Cc @fabiosantoscode 

Also if this should be a PR of it's own I can separate it from the remaining changes (Get Started related).

shcheklein:
Looks good to me! We should spend a bit of time to think about the landing page content (check other projects). Also, see this discussion that is relevant to your changes in the sidebar.json -  https://github.com/iterative/dvc.org/pull/1025#commitcomment-37819121 ... cc @iAdramelk - if we can merge this one before Gatsby migration we might just disallow having any entry w/o index page? But to be honest I would still do redirects if file is missing, and we should be doing this on the middleware level automatically (not relying on sidebar.json).

shcheklein:
another point - we need to introduce some special "Home" entry to the sidebar?

jorgeorpinel:
> - [ ] We should spend a bit of time to think about the landing page content (check other projects)

Working on that ⏳ 

> we need to introduce some special "Home" entry to the sidebar?

I don't think so, we have the DOC link in the top menu. But if you prefer that then sure, that wouldn't require the js code changes I did.

> if we can merge this one before Gatsby migration

If you want to merge this PR soon, please approve it and I'll limit it to this first step for now.

shcheklein:
> I don't think so, we have the DOC link in the top menu. But if you prefer that then sure, that wouldn't require the js code changes I did.

yeah, I think it's good to clearly see the structure and being able to return to the page - here is an example https://www.algolia.com/doc/ .... we might even want to use an icon and separate it a bit from the other menu items, etc - need to think a bit about UX

> if we can merge this one before Gatsby migration

I would confirm with @iAdramelk first.

jorgeorpinel:
> think about the landing page content

Updated

> I think it's good to clearly see the structure and being able to return to the page

OK, done.

jorgeorpinel:
> I would confirm with @iAdramelk first.

Well, too late! Haha. I've merged get-started and tutorials into [/doc/learn](https://dvc-landing-refactor-ge-nouj86.herokuapp.com/doc/learn) now (so this PR has become big because of all the link renaming). What do you think?

shcheklein:
I think it can be very hard to merge both now. I would focus on the content for now, not significant structural changes.

shcheklein:
Or at least, we'll need to merge Gatsby migration first and then deal with this here.

shcheklein:
Screen Shot 2020-03-15 at 8 39 13 PM


we should remove triangle, try to use icon and separate it from other items I think

shcheklein:
https://dvc-landing-refactor-ge-nouj86.herokuapp.com/doc/Home - url should be lower case

shcheklein:
content for the home page - probably should provide some good entry point and navigation for other sections

shcheklein:
also, `learn` - name is too general now, what is the plan for this section?

jorgeorpinel:
> we should remove triangle

Extracted to #1053

> learn - name is too general now, what is the plan for this section?

"Learn" is a common name for this kind of section with all the guides, tutorials, etc e.g. https://nextjs.org/learn 

---

I addressed your other couple comments @shcheklein.

jorgeorpinel:
I'm gonna limit the scope of this PR to what's done now. It's too big already. Fortunately its mostly due to renaming /doc/tutorials -> /doc/learn

jorgeorpinel:
> I would focus on the content for now, not significant structural changes

I thought structural changes for Get Started was the priority 😅 (both this one and then #599 which I've been working on as well). Should I postpone this work for now and focus on another high priority that is more content-related e.g. redesign the older use cases?

shcheklein:
@jorgeorpinel 

>  I thought structural changes for Get Started was the priority

that's true, bu to my mind the biggest potion of this is content - examples, preparing repos, writing text, etc - more than moving sections around.

jorgeorpinel:
> From https://github.com/iterative/dvc.org/issues/599#issuecomment-599883773
> ... Learn is not even part of Docs. Mostly because it's too generic. Second, it looks like the whole Learn section for Next is actually one single interactive tutorial

OK, I reverted back to /doc/tutorials but I still need to finish merging all the conflicts from the Gatsby migration... ⏳

jorgeorpinel:
Merged. https://dvc-landing-refactor-ge-nouj86.herokuapp.com/doc is running on Gatsby now.

redirects-list.json
```diff
@@ -6,12 +6,13 @@
   "^/((?:deb|rpm)/.+)                       https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                       https://discordapp.com/invite/dvwXA2N 303",
   "^/(?:docs|documentation)(/.*)?$          /doc$1",
-  "^/doc/commands-reference(/.*)?$          /doc/command-reference$1",
+  "^/doc/?$                                 /doc/tutorials/get-started 307",
```
shcheklein: should it be home now?

redirects-list.json
```diff
@@ -6,12 +6,13 @@
...
+  "^/doc/get-started                        /doc/tutorials/get-started",
   "^/doc/tutorial/?$                        /doc/tutorials",
   "^/doc/tutorial/(.*)?                     /doc/tutorials/deep/$1",
+  "^/doc/commands-reference(/.*)?$          /doc/command-reference$1",
   "^/doc/use-cases/data-and-model-files-versioning/?$ /doc/use-cases/versioning-data-and-model-files",
-  "^/doc/?$                                 /doc/get-started 307",
+  "^/doc/user-guide/contributing/?$         /doc/user-guide/contributing/core 307",
   "^/doc/understanding-dvc/?$               /doc/understanding-dvc/collaboration-issues 307",
-  "^/doc/changelog/?$                       /doc/changelog/0.18 307",
```
shcheklein: I really want to get rid of changelog, make it an external link ... again a special case like a Home button ... we spend time on it again and again

src/components/Community/data.json
```diff
@@ -28,13 +28,13 @@
...
     {
-      "url": "/doc/get-started",
+      "url": "/doc/tutorials/get-started",
       "title": "Get started",
       "description": "Reproducible walkthroughs on advanced topics"
     },
     {
       "url": "/doc/tutorials",
-      "title": "Tutorials",
+      "title": "DVC Tutorials",
```
shcheklein: why DVC Tutorials?

redirects-list.json
jorgeorpinel: Yes, I also noticed this earlier, it's fixed. Thanks

redirects-list.json
jorgeorpinel: I thought it was an external link already before? I didn't introduce this in the Gatsby migration, just moved the order of some redirects.

src/components/Community/data.json
jorgeorpinel: I was changing just back from "Learn DVC", didn't remember it was just "Tutorials" before. Reverted.

jorgeorpinel:
Let's try to address the remaining refactoring in #1074 (or further PRs)? Any problems specific to these nav order changes I can still fix here, of course.

PR #1055, opened 2020-03-16
deployment: require CF tokens

### LINKED ISSUE #1048: we need to purge CF cache on deployment
Changes introduced to the caching layer for DVC blog affected DVC.org - we need to purge cache in the same as we do in blog.

fabiosantoscode:
This ensures the CF tokens are available in production, otherwise the deployment fails.

closes #1048

### DIFF
--- a/scripts/clear-cloudflare-cache.js
+++ b/scripts/clear-cloudflare-cache.js
@@ -15,9 +15,9 @@
 // - CLOUDFLARE_ZONE_ID: The zone ID to purge. You can find it in the
 // sidebar of the "overview" tab for dvc.org
 
-const fetch = require('isomorphic-fetch');
+const fetch = require('isomorphic-fetch')
 
-const { CLOUDFLARE_TOKEN, CLOUDFLARE_ZONE_ID } = process.env;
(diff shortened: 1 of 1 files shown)


scripts/clear-cloudflare-cache.js
```diff
@@ -15,9 +15,9 @@
 // - CLOUDFLARE_ZONE_ID: The zone ID to purge. You can find it in the
 // sidebar of the "overview" tab for dvc.org
 
-const fetch = require('isomorphic-fetch');
```
shcheklein: why this style change?

fabiosantoscode: It was automatic, I assume something was changed.

shcheklein: nothing I am aware of at least ... I think we use `;` in the blog

fabiosantoscode: I don't see any semicolons in this repository.

shcheklein: my bad ... really sorry ... still though we are doing this with blog 🤦‍♂

PR #1068, opened 2020-03-18
[WIP] Gatsby migration fixes

pavelgrinchenko:
Fixes related to issues that we got after migrating to Gatsby.js in [this PR](https://github.com/iterative/dvc.org/pull/1025)

- [x]  404 code is not returned on 404
- [x] if I open https://dvc.org/doc/tutorials/versioning#automating-capturing Automatic Capturing is not highlighted on the right sidebar (and it’s not even active)
- [x]  go to the last anchor https://dvc.org/doc/tutorials/versioning#whats-next and then click Interactive Tutorial - behavior between previous version and the current one should be the same (no long scrolling animation)
- [x] styles are broken (different) in the new version - go carefully through all of them - make sure that docs are identical, inline code blocks, line spacing, list items spacing, notes, notes, cases like titles with links, etc, etc
- [x] in a Phablet mode Edit on Github button creates a second line (it might be related to styles issue - we already have two lines in the new version, compared to the previous one)
- [x] SEO problems - run Lighthouse new and old version
- [x] image subtitles are broken - check this page https://dvc.org/doc/tutorials/versioning , (Who doesn’t love ASCII directory art?)
- [x] make sure that GA runs the same way as before, with all the events
- [x] check Sentry - we are getting some new errors, hard to tell what page they come from, from what action

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,18 +2,43 @@
 
 const express = require('express')
 const app = express()
+const serveHandler = require('serve-handler')
 
 const apiMiddleware = require('./middleware/api')
 const redirectsMiddleware = require('./middleware/redirects')
-const notFoundMiddleware = require('./middleware/notFound')
 
 const port = process.env.PORT || 3000
 
 app.use(redirectsMiddleware)
 app.use('/api', apiMiddleware)
 
-app.use(express.static('public', { cacheControl: true, maxAge: 0 }))
-
-app.use(notFoundMiddleware)
+app.use((req, res) => {
+  serveHandler(req, res, {
+    public: 'public',
+    cleanUrls: true,
+    trailingSlash: false,
+    directoryListing: false,
+    headers: [
+      {
+        source: '**/*.@(jpg|jpeg|gif|png)',
+        headers: [
+          {
+            key: 'Cache-Control',
+            value: 'max-age=86400'
+          }
+        ]
+      },
+      {
+        source: '!**/*.@(jpg|jpeg|gif|png)',
+        headers: [
+          {
+            key: 'Cache-Control',
+            value: 'max-age=0'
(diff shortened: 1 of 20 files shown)


server.js
```diff
@@ -1,19 +1,47 @@
...
-app.use(notFoundMiddleware)
+app.use((req, res) => {
+  serveHandler(req, res, {
+    public: 'public',
+    cleanUrls: true,
+    trailingSlash: false,
+    directoryListing: false,
+    headers: [
+      {
+        source: '**/*.@(jpg|jpeg|gif|png)',
```
shcheklein: @fabiosantoscode could you please review this?

src/utils/resources.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const loadResource = url => {
+  if (!resourcesCache[url]) {
+    resourcesCache[url] = new Promise((resolve, reject) => {
+      const howToHandle = Object.keys(resourceNodeCreators).find(regExp =>
+        new RegExp(regExp).test(url)
+      )
+
+      console.log(url, howToHandle)
```
shcheklein: is it a debug?

pavelgrinchenko: Yes. Removed it.

pavelgrinchenko:
> make sure that GA runs the same way as before, with all the events

I've checked few events and they was sent. So all seems good.

pavelgrinchenko:
> check Sentry - we are getting some new errors, hard to tell what page they come from, from what action

Most critical issues were fixed. The rest needs to be watched.

server.js
shcheklein: @fabiosantoscode ??

PR #1072, opened 2020-03-19
Regular updates (March 18)

jorgeorpinel:
- Address https://github.com/iterative/dvc/pull/3471#pullrequestreview-377170844
- Misc stuff

### DIFF
--- a/content/docs/command-reference/install.md
+++ b/content/docs/command-reference/install.md
@@ -75,11 +75,11 @@ directory:
 To disable them, you need to **remove** or **edit** those files (i.e.
 `rm .git/hooks/post-checkout`, `vim .git/hooks/pre-commit`).
 
-## Using Pre-Commit tool
+## Using the **pre-commit** tool
 
-DVC provides support for [pre-commit](https://pre-commit.com/) users. To adjust
-`.pre-commit-config.yaml` you could either use
-`dvc install --use-pre-commit-tool`, or add these entries by hand:
+DVC provides support to manage Git hooks with
+[pre-commit](https://pre-commit.com/). To adjust `.pre-commit-config.yaml`, you
+can either use `dvc install --use-pre-commit-tool` or add these entries by hand:
 
 ```yaml
 repos:
@@ -103,8 +103,9 @@ repos:
 
 ## Options
 
-- `--use-pre-commit-tool` - install pre-commit/pre-push/post-checkout hooks into
-  the [pre-commit](https://pre-commit.com/) config (`.pre-commit-config.yaml`).
+- `--use-pre-commit-tool` - installs pre-commit, pre-push, post-checkout Git
+  hooks into the [pre-commit](https://pre-commit.com/) config file
+  (`.pre-commit-config.yaml`).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/install.md
+++ b/content/docs/command-reference/install.md
@@ -75,7 +75,7 @@ directory:
 To disable them, you need to **remove** or **edit** those files (i.e.
 `rm .git/hooks/post-checkout`, `vim .git/hooks/pre-commit`).
 
-## Using the **pre-commit** tool
+## Using the Pre-commit tool
 
 DVC provides support to manage Git hooks with
 [pre-commit](https://pre-commit.com/). To adjust `.pre-commit-config.yaml`, you
--- a/content/docs/install/linux.md
+++ b/content/docs/install/linux.md
@@ -29,7 +29,7 @@ plan to use, you might need to install optional dependencies: `[s3]`, `[azure]`,
 ### Example: How to install DVC with support for Amazon S3 storage
 
 ```dvc
-$ pip install 'dvc[s3]'
+$ pip install "dvc[s3]"
 ```
 
 In this case it installs `boto3` library as well, besides DVC.
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -62,8 +62,8 @@ $ git clone git@github.com:<username>/dvc.org.git
 $ cd dvc.org
 ```
 
-Make sure you have the latest version of [Node.js](https://nodejs.org/en/), and
-install [Yarn](https://yarnpkg.com/):
+Make sure you have a **recent** version of [Node.js](https://nodejs.org/en/),
+and install [Yarn](https://yarnpkg.com/):
 
 ```dvc
 $ npm install -g yarn
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -62,8 +62,8 @@ $ git clone git@github.com:<username>/dvc.org.git
 $ cd dvc.org
 ```
 
-Make sure you have a **recent** version of [Node.js](https://nodejs.org/en/),
-and install [Yarn](https://yarnpkg.com/):
+Make sure you have a recent version of [Node.js](https://nodejs.org/en/)
+(`^12.0.0`), and install [Yarn](https://yarnpkg.com/):
 
 ```dvc
 $ npm install -g yarn
@@ -118,7 +118,7 @@ The first few of the rules below should be enforced automatically by a Git
 pre-commit hook that is integrated when `yarn` installs the project dependencies
 (explained above).
 
-- No trailing whitespaces are allowed.
+- No trailing white spaces are allowed.
 
 - Content must be properly formatted at 80 symbols width.
 
--- a/content/docs/user-guide/running-dvc-on-windows.md
+++ b/content/docs/user-guide/running-dvc-on-windows.md
@@ -25,12 +25,6 @@ perfect solution, bu here are some ideas:
 - Install an actual Linux distro (e.g. Ubuntu) on a virtual machine, or in a HD
   partition (dual boot).
 
-> \* Please consult this [solution](https://stackoverflow.com/a/1257613/761963)
-> if you have strange file permission issues with Git for Windows. Also, Running
-> Node apps on WSL may produce
-> [`EACCESS`](https://github.com/microsoft/WSL/issues/3395) file permission
-> issues.
-
 ## Disable short-file name generation
 
 With NTFS, users may want to disable `8dot3` as per
--- a/content/docs/user-guide/running-dvc-on-windows.md
+++ b/content/docs/user-guide/running-dvc-on-windows.md
@@ -21,7 +21,10 @@ perfect solution, bu here are some ideas:
 - Consider enabling and using
   [WSL](https://blogs.windows.com/windowsdeveloper/2016/03/30/run-bash-on-ubuntu-on-windows/)
   ([Windows Terminal](https://devblogs.microsoft.com/commandline/) also
-  recommended). But it may be slow and unable to access GPUs, et al.\*
+  recommended). But it has major
+  [I/O performance issues](https://www.phoronix.com/scan.php?page=article&item=windows10-okt-wsl&num=2)
+  and is [unable to access GPUs](https://github.com/Microsoft/WSL/issues/829),
+  et al.\*
 - Install an actual Linux distro (e.g. Ubuntu) on a virtual machine, or in a HD
   partition (dual boot).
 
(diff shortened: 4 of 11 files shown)


jorgeorpinel:
> cmder is an integrated environment ... we also recommend Git Bash above and Anaconda prompt above
> From https://github.com/iterative/dvc.org/commit/635d6e992cfb6c135367d517649af037079794a2#r37945264
> image

Hi, sorry for the delay here @shcheklein. I tried Cmder and didn't love it. It's kind of complicated and not easy to configure. I'm not sure what you mean by an integrated environment or why that would be desirable. To me, the closest to a simple terminal app that can serve as a true Linux console is ideal, and that's what Windows Terminal does (using WSL). It also supports regular cmd and PowerShell.

> looks a bit messy now

I agree it's not super clear but is it because of this change? We were already recommending the other 2 before. We're not really recommending them, just mentioning them because they're well known. Maybe we should rec WT first, and then just mention other interesting options (I can re-include Cmder in that list)?

shcheklein:
> To me, the closest to a simple terminal app that can serve as a true Linux console is ideal, and that's what Windows Terminal does (using WSL).

WSL could be slow. sometimes _very_ slow. When we recommend an environment we need to ask data scientists/ML folks what they prefer. All of them have very specific needs, e.g. they should be able to install conda, pandas, numpy, etc .... they need access to GPU ... they crunch a lot of files. I'm just not sure if WSL is a good option. Cmder was recommended in the DS community since conda can be installed (or even comes with it) and other features.

content/docs/command-reference/install.md
```diff
@@ -75,11 +75,11 @@ directory:
 To disable them, you need to **remove** or **edit** those files (i.e.
 `rm .git/hooks/post-checkout`, `vim .git/hooks/pre-commit`).
 
-## Using Pre-Commit tool
+## Using the **pre-commit** tool
```
shcheklein: make it inline code `pre-commit`?

content/docs/install/linux.md
```diff
@@ -29,7 +29,7 @@ plan to use, you might need to install optional dependencies: `[s3]`, `[azure]`,
 ### Example: How to install DVC with support for Amazon S3 storage
 
 ```dvc
-$ pip install 'dvc[s3]'
+$ pip install "dvc[s3]"
```
shcheklein: they didn't work in some environment? did you check fish, zsh, bash?

shcheklein: most likely you are right and we need double quotes ... just to make sure

content/docs/user-guide/contributing/docs.md
```diff
@@ -62,8 +62,8 @@ $ git clone git@github.com:<username>/dvc.org.git
 $ cd dvc.org
 ```
 
-Make sure you have the latest version of [Node.js](https://nodejs.org/en/), and
-install [Yarn](https://yarnpkg.com/):
+Make sure you have a **recent** version of [Node.js](https://nodejs.org/en/),
```
shcheklein: what does "recent" mean?

shcheklein:
Check this link for example - https://www.phoronix.com/scan.php?page=article&item=windows10-okt-wsl&num=2

CmdEr has an option to install it with Git (so you don't need GitBash, Git, etc) - you have shell and git with it - that's why it can be considered a bit more than just terminal.

With Windows Terminal you either have to go with WSL (might be slow for DVC) or you have to install Git Bash and at this point it's not clear what experience is easier - WT or Cmder ... Also I would check how easy it is to install conda, if tqdm is compatible with win terminal (running git bash) ... etc

jorgeorpinel:
> WSL could be slow

Agreed @shcheklein. We're not recommending WSL here, just [Windows Terminal](https://www.microsoft.com/en-us/p/windows-terminal-preview/9n0dx20hk701?activetab=pivot:overviewtab). It can run regular cmd, PowerShell, or WSL (same as Cmder but nicer and easier to use).

content/docs/command-reference/install.md
jorgeorpinel: Since it's the official name of the tool I just wanted to bold it but I see that it has no effect in the rendering since the header is already bold:

image

Will change...

content/docs/install/linux.md
jorgeorpinel: Yes I remember it doesn't work with single quote in some env, Windows I think. We have notes about installing these optional reqs in other docs and they use `"` for that reason.

jorgeorpinel: Seems to work on Windows cmd. Sorry, I don't remember what env but its out there. I'll ask @efiop, he probably remembers.

content/docs/user-guide/contributing/docs.md
jorgeorpinel: Good question. I'm not sure what's the oldest version this works on, probably its safe to recommend 10+ cc @iAdramelk

jorgeorpinel: Its easier to revert to "latest" but will always work? We can't guarantee it, maybe we should get specific. But up to you, no strong opinion from my part.

content/docs/command-reference/install.md
jorgeorpinel: Done in 9aea2690

jorgeorpinel:
> Cmder was recommended in the DS community since conda can be installed
> CmdEr has an option to install it with Git

True, the full version comes with Git and SSH built-in but only when running some of the 9 tasks it comes with, I think. Again, not a very easy config:

image

> With Windows Terminal you either have to go with WSL (might be slow for DVC) or you have to install Git Bash

I think Git Bash is a stand alone terminal-like app as well, not something you can run from within other terminals. But it's a good argument: no point using WT with cmd or PowerShell if what we're looking for is a POSIX-like experience... You'd have to install cygwin or something like that.

OK I'm convinced. I'll add Cmder back. Should I mention WT at all?

jorgeorpinel:
> With Windows Terminal you either have to go with WSL (might be slow for DVC) or you have to install Git Bash

I think Git Bash is a stand alone terminal-like app as well, not something you can run from within other terminals. But it's a good argument: no point using WT with cmd or PowerShell if what we're looking for is a POSIX-like experience... You'd have to install cygwin or something like that.

OK I'm convinced. I'll add Cmder back. Should I mention WT at all?

shcheklein:
@jorgeorpinel so, WT with WSL is a reasonable approach, we can and we should recommend both of them ... I more worried about use recommending git bash as well above ... and even more that we do it in two different paragraphs ... can we combine all of this somehow - and recommend all three options for example?

shcheklein:
Git bash should be the last one

shcheklein:
and for WSL we should put a link and note about potential performance problems

shcheklein:
also, should we split terminal we recommend from shell - the way I understand this - both WT and CmdEr could be used with WSL for example

content/docs/install/linux.md
jorgeorpinel: He thought also Windows. Maybe they fixed it? `"` works everywhere anyway so let's keep?

shcheklein: 👍

content/docs/user-guide/contributing/docs.md
shcheklein: yes, I would just specify explicitly the versions we support

jorgeorpinel:
> Git bash should be the last one

Moved it under Anaconda Prompt

> WT with WSL is a reasonable approach, we can and we should recommend both of them

Yes, left this recommendation as a note under all the other ones.

> also, should we split terminal we recommend from shell

Hmmm it's not that straightfoward in Windows. Cmder default is "bash" but it's not a real shell, it uses Git Bash under the hood. I did rename the section to a more general term: **POSIX-like command line experience**.

content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -10,17 +10,19 @@ workarounds that can help avoid these potential problems:
...
-[Anaconda Prompt](https://docs.anaconda.com/anaconda/user-guide/getting-started/#open-prompt-win)
-– but they may not support all the POSIX features (e.g. `\` line continuation).
-
 💡 We recommend the full [Cmder](https://cmder.net/) console emulator (which
 already includes _Git for Windows_).
 
-Its also possible to enjoy a full Linux terminal experience with the
-[WSL](https://blogs.windows.com/windowsdeveloper/2016/03/30/run-bash-on-ubuntu-on-windows/)
-– but it may not be possible to access GPUs from this subsystem.
+Other common terminal alternatives for Windows are
```
shcheklein: still quite complicated, we have three paragraphs now cmd, git bash and anaconda, then a note with WSL ... it should be as simple as - cmd is bad, try to use environments like CmdEr, Windows Terminal + WSL, etc.

jorgeorpinel: OK Ivan I pushed a new version. It's clearer but not shorter. What do you think? I've actually been trying these options (except Anaconda Prompt) all this time which is why I keep adding caveats.

content/docs/user-guide/contributing/docs.md
jorgeorpinel: This is info. I don't currently know exactly. I know the std. node provided by apt-get (Ubuntu) is really old so I had to install [like this](https://github.com/nodesource/distributions/blob/master/README.md#debinstall). @iAdramelk could you chime in here please?

content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -5,22 +5,31 @@ involving system performance. Some, for example, have to do with NTFS file
...
+  is another recommendation, but it may not support all the desired CLI features
+  (e.g. `\` line continuation).
+- Consider enabling and using
+  [WSL](https://blogs.windows.com/windowsdeveloper/2016/03/30/run-bash-on-ubuntu-on-windows/)
+  ([Windows Terminal](https://devblogs.microsoft.com/commandline/) also
+  recommended). But it may be slow and unable to access GPUs, et al.\*
+- Install an actual Linux distro (e.g. Ubuntu) on a virtual machine, or in a HD
+  partition (dual boot).
+
+> \* Please consult this [solution](https://stackoverflow.com/a/1257613/761963)
```
shcheklein: def would not put it here - this is not DVC related in any way

content/docs/user-guide/running-dvc-on-windows.md
shcheklein: looks good to me, except the last note about git problems on win

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: I know but it's an issue I struggled with from the go with Cmder just trying to use `git status`... Can't even commit changes because of this.

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: OK I removed the whole note.

content/docs/user-guide/running-dvc-on-windows.md
shcheklein: make it single link close to Cmder ... but I would avoid this at all ... we can't cover all edge case - that's what SO for

content/docs/user-guide/running-dvc-on-windows.md
shcheklein: 👍

content/docs/user-guide/contributing/docs.md
shcheklein: we have engines specified as

```
  "engines": {
    "node": "^12.0.0"
  },
``` - in blog (which will be part of the dvc.org soon). @fabiosantoscode and @iAdramelk should have more context - ping them in Slack if needed

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: Yeah I just removed it but pretty sure it's something most people will battle with. Oh well, not our fault that Windows sucks

content/docs/user-guide/contributing/docs.md
jorgeorpinel: Pinged Fabian but he's probably offline rn. I'm using 12 for now so we can merge this and I can update with whatever they say in another PR later. Going to focus on other doc PRs for now.

jorgeorpinel:
> and for WSL we should put a link and note about potential performance problems

Added links to https://www.phoronix.com/scan.php?page=article&item=windows10-okt-wsl&num=2 and https://github.com/Microsoft/WSL/issues/829.

Imma merge this.

jorgeorpinel:
- [ ] Just pending a definitive answer in https://github.com/iterative/dvc.org/pull/1072#discussion_r398235246 which can be addressed in a later PR. Thanks!

PR #1086, opened 2020-03-22
Fix ChunkLoadError

pavelgrinchenko:
- Update Gatsby.js to the latest version for getting [this fix](https://github.com/gatsbyjs/gatsby/pull/18051). It allows Gatsby.js ignore ChunkLoadError and load static pages when it happens.
- Ignore ChunkLoadError in Sentry
- Optimize images loading checker a little bit


shcheklein:
@pavelgrinchenko thanks!

1. Let's resolve conflicts
2. See some comments I left

gatsby-config.js
```diff
@@ -70,7 +70,13 @@ const plugins = [
     options: {
       dsn: process.env.SENTRY_DSN,
       environment: process.env.NODE_ENV,
-      enabled: process.env.NODE_ENV === 'production'
+      release: process.env.CIRCLE_SHA1,
+      enabled: process.env.NODE_ENV === 'production',
+      ignoreErrors: [
+        // Ignore errors on missing chunks when new version was deployed
```
shcheklein: Please, let's put a better comment here - it's not clear why do we have this.

gatsby-config.js
```diff
@@ -70,7 +70,13 @@ const plugins = [
...
+        'ChunkLoadError'
+      ],
+      ignoreUrls: [/extensions\//i, /^chrome:\/\//i]
```
shcheklein: why do we need this?

gatsby-config.js
pavelgrinchenko: Updated comment. Now it should me more clear. Wdyt?

gatsby-config.js
pavelgrinchenko: Added comment to the code

PR #1089, opened 2020-03-24
Open footer social links in the new tab

pavelgrinchenko:
Now discord link in the footer doesn't work. It happens because Gatsby.js don't know anything about `/chat` redirect in the server.js. I suggest open these links in the new tab as it is for the same links in the header.
But also I think that we should duplicate redirects on front side and I have plan to do this in the merging PR.


shcheklein:
> But also I think that we should duplicate redirects on front side and I have plan to do this in the merging PR.

@pavelgrinchenko could you elaborate on this?

PR #1091, opened 2020-03-24
[WIP] Blog merging into main website codebase

pavelgrinchenko:
- [x] Setup TS
- [x] Migrate blog's content and components to the main repo


.eslintrc.json
```diff
@@ -1,12 +1,18 @@
...
     "plugin:jsx-a11y/recommended",
     "plugin:json/recommended"
   ],
-  "parser": "babel-eslint",
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true // Allows for the parsing of JSX
+    }
+  },
```
fabiosantoscode: Do we need to start using typescript?

I feel that it isn't really necessary here.

.tmp-empty-package/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "empty-package",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT"
+}
```
fabiosantoscode: Was this mistakenly committed?

config/postcss/media.js
```diff
@@ -0,0 +1,20 @@
+/* tslint:disable object-literal-sort-keys */
```
fabiosantoscode: tslint is deprecated and we shouldn't be using it right here. Or typescript for that matter :)

.eslintrc.json
```diff
@@ -17,10 +23,14 @@
...
+    "@typescript-eslint/interface-name-prefix": [
       "error",
-      { "code": 80, "ignoreTemplateLiterals": true, "ignoreUrls": true }
+      { "prefixWithI": "always" }
     ],
-    "no-tabs": "error"
+    "no-tabs": "error",
+    "react/prop-types": "off", // Props should be described as TS types
+    "@typescript-eslint/explicit-function-return-type": "off",
+    // TODO: remove after rewriting code to TS
```
shcheklein: is this PR still WIP? we don't merge TODOs usually

.eslintrc.json
shcheklein: agreed to use it, since blog is on TS

.restyled.yaml
```diff
@@ -4,5 +4,8 @@ restylers:
     include:
       - './*.{js,md}'
       - 'pages/**/*.js'
-      - 'content/docs/**/*.{js,md}'
+      - 'content/**/*.{js,md}'
```
shcheklein: do we have js inside content?

src/api/index.ts
```diff
@@ -0,0 +1,2 @@
+export const getCommentsCount = (url: string) =>
```
shcheklein: so, do we need other APIs? should they all be in the same place?

package.json
```diff
@@ -29,17 +29,24 @@
   "homepage": "https://github.com/iterative/dvc.org#readme",
   "dependencies": {
     "@octokit/graphql": "^4.3.1",
```
shcheklein: do we need to specify engines like we had in blog ? cc @fabiosantoscode

shcheklein: cc @iAdramelk

package.json
```diff
@@ -89,29 +103,56 @@
...
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn format-staged && yarn lint-staged"
+      "pre-commit": "yarn format-staged && tsc --noEmit && yarn lint-staged"
     }
   },
   "lint-staged": {
-    "*.{js,json}": "eslint"
+    "*.{js,json,ts,tsx}": "eslint --quiet"
```
shcheklein: is it temporary?

shcheklein: I mean `--quite`?

package.json
```diff
@@ -11,7 +11,7 @@
     "start": "./scripts/clear-cloudflare-cache.js; NODE_ENV=production node server.js",
     "format-staged": "pretty-quick --staged --no-restage --bail",
     "format-check": "prettier --check '{.,pages/**,content/docs/**,src/**}/*.{js,md,json}'",
-    "lint-check": "eslint --ext .json,.js src middleware",
+    "lint-check": "tsc --noEmit && eslint --ext .json,.js,.ts,.tsx src middleware --quiet",
```
shcheklein: do we need `--skipLibCheck`? like we had in blog?

.tmp-empty-package/package.json
pavelgrinchenko: Unfortunately not. Styled components have bug in [their typings.](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311). And it's workaround to fix it. It'll be removed after full migration to CSS modules

config/postcss/media.js
pavelgrinchenko: Removed all such comments

.eslintrc.json
pavelgrinchenko: Yes, it's still WIP and will be few more days at least. As comment says this TODO will be removed after migrating all components to TS

.restyled.yaml
pavelgrinchenko: I don't think so. Removed `.js` extensions there

src/api/index.ts
pavelgrinchenko: Good point! Refactored it to the similar one from the `utils/api`

package.json
pavelgrinchenko: It was temporary too, yes. Refactored remaining files with warnings and removed this option

package.json
pavelgrinchenko: Yep. Good options, it can increase typechecking performance. Added.

package.json
pavelgrinchenko: Specified it.

pavelgrinchenko:
@shcheklein What do you mean by "deduplicate fonts"? We have original otf fonts and I added woff/woff2 versions of these fonts. I don't see any duplication.

README.md
```diff
@@ -32,6 +52,64 @@ If you have any questions, please join the [community](https://dvc.org/chat) and
 use the `#dev-docs` channel to discuss any issues in our website or docs. We are
 very responsive and happy to help.
 
+# Writing posts
```
shcheklein: minor: Writing blog posts

shcheklein:
@pavelgrinchenko CI failed

shcheklein:
> What do you mean by "deduplicate fonts"? We have original otf fonts and I added woff/woff2 versions of these fonts

yeah, probably it's not clear - do we need both kinds - otf and woff?

.tmp-empty-package/package.json
shcheklein: let's ccreate a ticket with checkboxes to fix after merging this, please

gatsby-node.js
```diff
@@ -1,10 +1,19 @@
 /* eslint-env node */
 
+const fs = require('fs')
```
shcheklein: not for this PR: this file is clearly getting out of hand - it's close to impossible to review it. We should split into docs and blog is possible at all, move common helpers outside?

src/components/IconSet/icons/index.ts
```diff
@@ -0,0 +1,5 @@
+export { default as chat } from './chat.png'
+export { default as discord } from './discord.png'
```
shcheklein: did we remove them from dvc.org static? do we need those still?

src/components/LandingHero/styles.js
```diff
@@ -28,7 +28,7 @@ export const Title = styled.h1`
   font-size: 40px;
   font-weight: 500;
   color: #40364d;
-  font-family: BrandonGrotesqueMed;
+  font-family: BrandonGrotesque;
```
shcheklein: Are you sure about this change? Styles look different now in some dvc.org places

src/components/LayoutFooter/logo.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="36px" height="22px" viewBox="0 0 36 22" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
shcheklein: same here, do we remove it from other places?

src/components/Page/fonts/fonts.css
```diff
@@ -0,0 +1,119 @@
+/* Prevent mobile browsers from changing font-size */
+* {
+  -webkit-text-size-adjust: none;
+  text-size-adjust: none;
+}
+
+@font-face {
+  font-family: 'BrandonGrotesque';
+  src: local('Brandon Grotesque Light'), local('BrandonGrotesque-Light'),
+    url('./BrandonGrotesque-Light.woff2') format('woff2'),
```
shcheklein: just curious - do we need both woff2 and woff?

src/components/SEO/index.tsx
```diff
@@ -0,0 +1,98 @@
+import React from 'react'
```
shcheklein: should we start splitting blog, common, docs, etc? most likely SEO will be different across those .. and number of components is getting out or hands as well

src/components/Video/index.js
```diff
@@ -52,11 +52,8 @@ export default class Video extends Component {
   watch = () => {
     logEvent('button', 'video')
     this.setState({ watching: true })
-    this.play()
   }
 
-  play = () => {}
```
shcheklein: do you know why was it there in the first place? )

src/queries/siteMeta.ts
```diff
@@ -0,0 +1,22 @@
+import { useStaticQuery, graphql } from 'gatsby'
```
shcheklein: just to understand - where is it used?

src/templates/blog-home.tsx
```diff
@@ -0,0 +1,45 @@
+import { graphql } from 'gatsby'
```
shcheklein: is there analog for doc page? should it be unified - the way we process pages?

src/utils/api.ts
```diff
@@ -0,0 +1,84 @@
+import { useEffect, useState } from 'react'
+import fetch from 'isomorphic-fetch'
+
+const useAPICall = <R>(url: string) => {
+  const [ready, setReady] = useState(false)
```
shcheklein: so, do we cache results? for how long?

src/utils/paginatablePageGenerator.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-env node */
```
shcheklein: can we keep utils split by their meaning - it was in the blog - we should have blog component, paginator and utils there instead of very long names ... and it's def not a global util

src/utils/tagToSlug.js
```diff
@@ -0,0 +1,6 @@
+module.exports = tag =>
```
shcheklein: feels like a wrong place for these utils ... in general feels like components, utils, etc should have structure now - blog, docs ... but again, not sure about best practices

src/utils/redirects.js
```diff
@@ -42,4 +42,21 @@ const getRedirect = (host, pathname, { req, dev } = {}) => {
   return matchRedirectList(host, pathname)
 }
 
+const handleFrontRedirect = (host, pathname, clickEvent) => {
```
shcheklein: do we run it for every link of the website? just to make sure that we run it for link inside docs (especially those that we generate with plugins)

src/components/Community/Events/index.js
```diff
@@ -63,9 +59,9 @@ function CommunityEvent({
 
         <Link
           color={theme.color}
+          as={LinkSC}
           href={url}
           target="_blank"
-          rel="noreferrer noopener"
```
shcheklein: why remove it here? (and we added it in some other places?)

src/components/DocLayout/index.js
```diff
@@ -1,8 +1,8 @@
 import React, { useCallback, useState } from 'react'
 import PropTypes from 'prop-types'
+import MainLayout, { LayoutModifiers } from '../MainLayout'
```
shcheklein: should we move DocLayout into Document/Layout?

src/components/Documentation/Markdown/lang/dvc.js
```diff
@@ -1,17 +1,16 @@
 /* global exports:readonly */
-
 'use strict'
```
shcheklein: do we reuse this in blog? we need to move it outside then

shcheklein: actually this is highlight js? it should not be used anymore at all

src/components/Documentation/Markdown/lang/usage.js
```diff
@@ -6,8 +6,8 @@ Object.defineProperty(exports, '__esModule', {
   value: true
 })
 
-var _javascript = function(hljs) {
```
shcheklein: same here -  we don't use this file anymore I hope

src/components/LandingHero/styles.js
pavelgrinchenko: Yes, I'm sure. For `font-weight: 500` will be used `Brandon Grotesque Medium`, i.e. the same font as was before.
https://github.com/iterative/dvc.org/blob/blog-merging/src/components/Page/fonts/fonts.css#L87-L89

src/components/IconSet/icons/index.ts
pavelgrinchenko: Good catch! Removed these icons from `static/` folder

README.md
pavelgrinchenko: Fixed

src/components/LayoutFooter/logo.svg
pavelgrinchenko: Deduplicated it too

src/components/Page/fonts/fonts.css
pavelgrinchenko: Unfortunately yes. IE 11 doesn't support woff2... https://caniuse.com/#search=woff2

src/components/SEO/index.tsx
pavelgrinchenko: It depends on what you want. Could you describe it more precisely and we will decide then?
Now SEO data should be same as it was before.

src/components/Video/index.js
pavelgrinchenko: I think once upon a time here was another logic (or it was planned =)). But for now it looks like unnecessary code, so I just removed it.

src/queries/siteMeta.ts
pavelgrinchenko: It's used [here](https://github.com/iterative/dvc.org/blob/blog-merging/src/components/Page/DefaultSEO/index.tsx#L6) and [here](https://github.com/iterative/dvc.org/blob/blog-merging/src/components/SEO/index.tsx#L5).

src/templates/blog-home.tsx
pavelgrinchenko: I refactored `doc.js` to TS, and renamed it to `doc-home.ts` to keep consistensy with other templates. Is that what you meant?

src/utils/api.ts
pavelgrinchenko: It's cached on our proxy server side, as it was before.
Here it is: 
https://github.com/iterative/dvc.org/blob/blog-merging/middleware/api/comments.js#L18
https://github.com/iterative/dvc.org/blob/blog-merging/middleware/api/discourse.js#L8
https://github.com/iterative/dvc.org/blob/blog-merging/middleware/api/github.js#L6

src/components/Community/Events/index.js
pavelgrinchenko: It will be added automatically in Link component for all not relative links
https://github.com/iterative/dvc.org/blob/db97402599db07da2c9421ce094658944a1d0bfd/src/components/Link/index.tsx#L22-L24
It much more safety, than setting it manually every time.

src/components/DocLayout/index.js
pavelgrinchenko: Do you want to add some namespaces for docs (and maybe blog) components?

src/utils/redirects.js
pavelgrinchenko: It handles it two places: in the Link component when user clicks on link (before changing route) and as fallback after changing url (404 can be showed in this case, but then user will be redirected to correct url). I replaced all links in the generated pages from markdown to the Link component, so there it will be handled as first variant.

src/utils/paginatablePageGenerator.js
pavelgrinchenko: There are few reasons why I moved it here:
- It uses on global level (on page generation process in the `gatsby-node.js`) and not in the Paginator itself
- By putting it in the `utils` folder override rules for eslint applied automatically 
https://github.com/iterative/dvc.org/blob/233e9334a5b1c1a0a375e77d8eba8302ec18c465/.eslintrc.json#L36-L42
I can rename it to something else, but I prefer clarity over shorteness

src/utils/tagToSlug.js
pavelgrinchenko: I agree in this case. Inlined this helper right into the component where it uses

pavelgrinchenko: About utils structure we can think in the next PRs. Maybe we should separate them at least on usage sphere: gatsby node api/server/react side for example.

pavelgrinchenko: It's too uses in `gatsby-node` (ie pages generation side) and in react component...

src/components/Documentation/Markdown/lang/dvc.js
pavelgrinchenko: Seems it is hightligth.js configs. Removed it.

pavelgrinchenko:
>  features, support - header height id different

What header do you mean? I see the same height for both pages
image

pavelgrinchenko:
> fonts look different old/new (e.g. top menu, Chrome)

There is small difference in hintint, but seems it depends on font format. But font the same. I believe that we should leave woff/woff2 with this uncritical differences as it is optimal for the web

iAdramelk:
> There is small difference in hintint, but seems it depends on font format. But font the same. I believe that we should leave woff/woff2 with this uncritical differences as it is optimal for the web

@pavelgrinchenko did you take woff/woff2 fonts form the blog, or generated them yourself? I remember that I have some hinting problems then I tried to convert them first time and only make correct hinting version after few tries.

shcheklein:
@pavelgrinchenko 
> What header do you mean? I see the same height for both pages

height of the space between menu and the first text on the page

shcheklein:
> with this uncritical differences as it is optimal for the web

let's try to keep the fonts as they were, not sure what is the reason to break the design - both websites handle this fine. Let's keep it as it was before? unless I'm missing something

src/components/DocLayout/index.js
shcheklein: we already have a directory `Documentation` - we can move Layout for it there, right?

if do so we don't need Doc prefix

src/components/Community/Events/index.js
shcheklein: got it!

src/utils/redirects.js
shcheklein: > 404 can be showed in this case, but then user will be redirected to correct url)

yep, I saw this - does not look nice. Could you clarify this part `fallback after changing url`?

src/templates/blog-home.tsx
shcheklein: 👍 yes, that was a comment about some consistency how we build and process stuff

src/components/SEO/index.tsx
shcheklein: we have Documentation already for components related to `/doc` - that where I would start. Move blog to Blog/, for example. keep common in the `components` root

SEO - the ways we process them is different - for blog we rely on some additional meta-fields in Mardown, for docs we don't have that, right? Thus we (or will) have different SEO components for docs/blog

src/components/Page/fonts/fonts.css
shcheklein: got it ... but do we need the third type - otf? or what was that name? can we just keep otf?

src/components/Video/index.js
shcheklein: probably some GA event that was remove along these refactoring. It's fine for now. 👍

src/components/LandingHero/styles.js
shcheklein: so, fonts look different. We need to fix this.

src/components/IconSet/icons/index.ts
shcheklein: please, check other places assets we move as well (to the extent it is possible)

README.md
shcheklein: thanks!

src/utils/paginatablePageGenerator.js
shcheklein: * linters should never dictate code structure, so I would not consider a good argument
* global usage is a good point - we had a debate when implementing this and at that point decided to keep under some kind of `blog/utils` since it's related to blog ... it's totally fine for gatsby-node to use utils/code from different levels

pavelgrinchenko: Totally agree with first point. It's just conveniently for this case

pavelgrinchenko:
Pushed fonts files from the blog

pavelgrinchenko:
@shcheklein Also I made some changes to blog styles, primary for code blocks. Could you check it?

shcheklein:
@pavelgrinchenko it failed to deploy looks like?

pavelgrinchenko:
> https://dvc-landing-blog-mergin-a6sosq.herokuapp.com/blog/march-20-community-gems - "mins read" differs with prod 5 vs 7

It depends on updated Gatsby.js version (to be more precise on internal `gatsby-transformer-remark` plugin) and not our own changes

pavelgrinchenko:
> it failed to deploy looks like?

I added `public` folder to Heroku cache. Seems it resolved problem.

src/utils/redirects.js
pavelgrinchenko: We can intercept url change only with `Link` component, or handle it after location change. I tried to replace all direct `a` usages to `Link` component. But if somewhere still there is any `a` - it will be handled after front location change

gatsby-node.js
pavelgrinchenko: Agree, I will try to do it in the next PR.

.tmp-empty-package/package.json
pavelgrinchenko: I save all such comments to my list and will add as description in the next PR

src/components/DocLayout/index.js
pavelgrinchenko: Moved it to `Documentation` directory

pavelgrinchenko:
> features, support - header height id different

But it's the same. Could you clarify this item?
image
image

pavelgrinchenko:
> https://dvc-landing-blog-mergin-a6sosq.herokuapp.com/blog/march-20-community-gems - "mins read" differs with prod 5 vs 7

What about this item? I described reason [a little bit above](https://github.com/iterative/dvc.org/pull/1091#issuecomment-605984577)? Could you check it?

src/components/Documentation/DocLayout/SidebarMenu/index.js
```diff
@@ -0,0 +1,131 @@
+import React, { useEffect, useRef, useState } from 'react'
```
shcheklein: you don't need `Doc` prefix in this case?

fabiosantoscode:
There's a font flash at least on firefox:

Screenshot from 2020-03-30 19-37-30

fabiosantoscode:
I also discovered that `blog.dvc.org` does not redirect to `/blog`:

```
curl --header 'host: blog.dvc.org' https://dvc-landing-blog-mergin-hdexao.herokuapp.com -I
```

An entry needs to be created in redirects-list.json so that we can deploy this to both domains and handle the redirect.

The entry can look like this:

`^https://blog.dvc.org(/.*)$       /blog$1       301`

pavelgrinchenko:
@fabiosantoscode There is separate PR in the blog repo with redirects https://github.com/iterative/blog/pull/149
Regarding font flash - does current master have it too? (I mean dvc.org) I think yes - so we can handle it in separate PR

src/components/Documentation/DocLayout/SidebarMenu/index.js
pavelgrinchenko: Fixed

shcheklein:
> Regarding font flash - does current master have it too? (I mean dvc.org) I think yes - so we can handle it in separate PR

current master does not have it.

pavelgrinchenko:
> current master does not have it.

But it has https://take.ms/sRHMm
And honestly we can't do any special here. On the production it not so obvious because we have CDN there and fonts load faster.

fabiosantoscode:
> This behaviour much better, because on the slow connections (mobile networks, etc.) user can't see text for a long time. This flash is showed only one time on first loading.

:clap: :clap: and here I was thinking that was a mistake!

Maybe later we can do something where default fonts are only shown when the user (mobile or not) has waited for at least 1 or 2 seconds. That way users on fast wifi don't see the default fonts.

fabiosantoscode:
> There is separate PR in the blog repo with redirects iterative/blog#149

Ok, sounds good! Eventually we will want to bring that site down, archive the repository and serve everything from here.

shcheklein:
> This PR has font-display: swap which allows to browsers render local fonts first and replace to custom when it will be loaded https://take.ms/R7rOn This behaviour much better, because on the slow connections (mobile networks, etc.) user can't see text for a long time. This flash is showed only one time on first loading.

Got it @pavelgrinchenko ! thanks for the explanations 🙏

src/components/Page/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+const Page: React.SFC<IPageProps> = props => {
+  let LayoutComponent = MainLayout as LayoutComponent // TODO: remove type cast
+
+  useRedirects()
+  useAnchorNavigation()
+  useSmoothScroll(props.enableSmoothScroll)
+
+  if (!props.pageContext.is404) {
+    if (props.pageContext.isDocs) {
+      LayoutComponent = DocumentationLayout as LayoutComponent // TODO: remove type cast
```
shcheklein: add a checkbox or create a ticket, let's not merge with TODOs in code

pavelgrinchenko: Removed it and added to my list

pavelgrinchenko:
> Ok, sounds good! Eventually we will want to bring that dyno down, archive the repository and serve everything from here.

It's fine to do this in the PR with migrating deployment to S3 from blog.

PR #1094, opened 2020-03-27
cmd ref: review commit explanations

jorgeorpinel:
Addresses docs part of iterative/dvc/issues/3446

### DIFF
--- a/content/docs/command-reference/add.md
+++ b/content/docs/command-reference/add.md
@@ -96,8 +96,8 @@ reproducible.
 - `--no-commit` - do not save outputs to cache. A DVC-file is created and an
   entry is added to `.dvc/state`, while nothing is added to the cache.
   (`dvc status` will report that the file is `not in cache`.) Use `dvc commit`
-  when ready to commit the results to cache. This is analogous to using
-  `git add` before `git commit`.
+  when ready to commit outputs with DVC. This is analogous to using `git add`
+  before `git commit`.
 
 - `-f <filename>`, `--file <filename>` - specify name of the DVC-file it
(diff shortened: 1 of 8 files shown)


content/docs/command-reference/add.md
```diff
@@ -93,11 +93,11 @@ reproducible.
...
 
-- `--no-commit` - do not save outputs to cache. A DVC-file is created, and an
-  entry is added to `.dvc/state`, while nothing is added to the cache. (The
-  `dvc status` command will report that the file is `not in cache`.) This is
-  analogous to using `git add` before `git commit`. Use `dvc commit` when ready
-  to commit the results to cache.
+- `--no-commit` - do not save outputs to cache. A DVC-file is created and an
+  entry is added to `.dvc/state`, while nothing is added to the cache.
+  (`dvc status` will report that the file is `not in cache`.) Use `dvc commit`
+  when ready to commit the results to cache. This is analogous to using
```
shcheklein: don't see any improvements to be honest ... commit smth to cache -not sure this is a good idea .. also it's like explaining `commit is commit`

jorgeorpinel: There's no change, I only reordered the sentences for consistency with the desc. of this same option in run and repro refs.

shcheklein: yep, thus PR is approved :)

jorgeorpinel: > commit smth to cache -not sure this is a good idea

Changing to "when ready to commit outputs with DVC" in all refs.

> it's like explaining commit is commit

No because this is the add ref, not the commit ref. It's just a link to the commit ref 🙂

PR #1096, opened 2020-03-27
gdrive: Service account support; Default credentials; Trash only conf…

### LINKED ISSUE #1095: user-guide: update gdrive setup guide
- New optional param to turn on Trash_Only option for gdrive remote;
- DVC starts to provide default client id and secret to access GDrive API;
- Service account support.

maxhora:
Fixes #1095 

- Service account setup and support;
- Default OAuth credentials;
- Trash only config param.


content/docs/command-reference/remote/modify.md
```diff
@@ -271,6 +271,43 @@ including obtaining the necessary credentials, and how to form `gdrive://` URLs.
...
+
+- `gdrive_service_account_p12_file_path` - Google Project's
+  [service account](/doc/user-guide/setup-google-drive-remote#configure-the-gdrive-remote-with-a-service-account)
+  `.p12` file path.
+
+  ```dvc
+  $ dvc remote modify myremote gdrive_service_account_p12_file_path <file_path.p12>
+  ```
+
+- `gdrive_service_account_user_email` - user email to be impersonated by
```
jorgeorpinel:
```suggestion
- `gdrive_service_account_user_email` - user email corresponding to the
```

content/docs/command-reference/remote/modify.md
```diff
@@ -271,6 +271,43 @@ including obtaining the necessary credentials, and how to form `gdrive://` URLs.
...
+  ```dvc
+  $ dvc remote modify myremote gdrive_service_account_user_email <user account email>
+  ```
+
+- `gdrive_trash_only` - configures DVC to move a remote file to the trash or to
+  delete it permanently. By default, is `False`.
+
+  ```dvc
+  $ dvc remote modify myremote gdrive_trash_only <True|False>
+  ```
```
jorgeorpinel: I don't understand this. Can you explain please, so I can think of a suggestion? Thanks

Also, the values should probably be lower case? i.e. `false` in the bullet, and `<true|false>` in the example.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -3,7 +3,11 @@
 Follow this guide to setup Google Drive as your <abbr>DVC project</abbr>'s
 [remote storage](/doc/command-reference/remote).
 
-## Configure a Google Cloud project
+## Configure a Google Cloud project (Highly recommended optional step)
```
jorgeorpinel:
```suggestion
## Configure a Google Cloud project (highly recommended)
```

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -3,7 +3,11 @@
...
+
+We highly recommend to follow this guide and complete setup of your own Google
+Cloud project and
```
jorgeorpinel:
```suggestion
Optionally but highly recommended, follow this guide to setup your own Google Cloud project and
```

jorgeorpinel: We should also probably expand on why this is beneficial to the user in another sentence in this same paragraph. E.g. he will not share rate limits with others, maybe more control/visibility for debugging, Idk, etc.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -98,6 +102,30 @@ $ dvc remote modify mygdfolder gdrive_client_id <client ID>
 $ dvc remote modify mygdfolder gdrive_client_secret <client secret>
 ```
 
+## Configure the GDrive remote with a service account
```
jorgeorpinel:
```suggestion
## Configure the Google Drive remote with a service account
```
Same change for the previous header.

jorgeorpinel: Also, what's the difference in the previous header? Should it be something like **Configure the Google Drive remote with user credentials** to make the difference clear at first glance?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -98,6 +102,30 @@ $ dvc remote modify mygdfolder gdrive_client_id <client ID>
...
+
+A
+[service account](https://cloud.google.com/docs/authentication/getting-started#creating_a_service_account)
+can be used to configure the GDrive remote.
```
jorgeorpinel: What's a service account? What are the benefits?

content/docs/command-reference/remote/modify.md
maxhora: `impersonate some user account` is wide used term for Google Service Accounts mentioned in official docs as well

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: yes, we def need to add motivation, mention that sharing potentially could lead for slower performance, give hints how to check if users hit limits, etc

content/docs/command-reference/remote/modify.md
maxhora: Idea was to specify possible values of the param - True or False. Value from uppercase is a Python style, but I'm not sure that lowercase will be parsed correctly into boolean.

content/docs/user-guide/setup-google-drive-remote.md
maxhora: I'm not sure what phrase fits the desired balance here - it seems we highly recommend to configure their own GCP project, but we should also highlight that the step is optional and user can save a lot time by skipping it.

content/docs/user-guide/setup-google-drive-remote.md
maxhora: `dvc remote modify` already mentions `service account` for Google Cloud. It is the same thing for Google Drive as well.
GCP project can have list of service accounts. Each service account has its own email address and some granted access rights, etc.
Benefit to use service account is to remove dependency on some user's account. Also it is recommended way to do some automated actions, like running tests on Travis.

content/docs/user-guide/setup-google-drive-remote.md
maxhora: DVC should be configured with different params in config to auth user account (client id and secret should be in the config) and to auth service account (service account email address and .p12 local key file path params in the config). Renaming previous section header sounds great!

content/docs/command-reference/remote/modify.md
jorgeorpinel: Interesting. OK, how about

```suggestion
- `gdrive_service_account_user_email` - email of the service account to impersonate
```

I'm a little confused about the term "user" here and not sure how/whether to include it. Is there a user account AND a service account (owned by the user) AND a 3rd account that will impersonate?

content/docs/command-reference/remote/modify.md
jorgeorpinel: I meant please explain what this option does. I don't get it.

As for the letter case, all the other remote params use lower case. The example can probably just be:

  ```dvc
  $ dvc remote modify myremote gdrive_trash_only true
  ```

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Recommended implies optional 🙂

jorgeorpinel: Also I added the term "optionally" in my next suggestion. Thanks

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Thanks for the info.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Got it. I also found https://cloud.google.com/iam/docs/understanding-service-accounts. I'll review the GC section and maybe copy some info into this one.

jorgeorpinel:
Thanks @maxhora I think we'll take this over now 👍

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: > give hints how to check if users hit limits

I'm not sure how to do this.

Could you guys help me summarize the benefits of crating your own GCP project?

I added some general info for now but it's probably not enough, please check it out.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Updated. Moved Q about terminology to https://github.com/iterative/dvc/commit/3ff804b3b28abe67ffa206b92c8345acb61935dd#r38129539.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: p.s. you know what I should just try all the gdrive config options again to help in QA which is probably needed anyway, I think that will allow me to review this and the whole doc anyway ⏳

content/docs/command-reference/remote/modify.md
jorgeorpinel: OK found the explanation in iterative/dvc/issues/3477, also Ivan shared some references. Updated!

content/docs/command-reference/remote/modify.md
```diff
@@ -271,6 +271,45 @@ including obtaining the necessary credentials, and how to form `gdrive://` URLs.
...
+
+- `gdrive_service_account_p12_file_path` - Google Project's service account
+  `.p12` file path.
+
+  ```dvc
+  $ dvc remote modify myremote gdrive_service_account_p12_file_path path/to/file.p12
+  ```
+
+- `gdrive_service_account_user_email` - email of the service account to
+  impersonate.
```
jorgeorpinel: Hmmm actually I still don't understand the difference between `gdrive_service_account_email` and `gdrive_service_account_user_email`. Will read about this [impersonation](https://medium.com/google-cloud/impersonating-users-with-google-cloud-platform-service-accounts-ba762db09092) thing...

jorgeorpinel: @maxhora I assume this is just an optional gdrive remote config param?

jorgeorpinel: the article reads "enables another user or service account to impersonate a service account". So who is impersonating who? `gdrive_service_account_email` impersonates `gdrive_service_account_user_email`? Also could the param names be less similar and more explicit e.g. change the latter for `gdrive_impersonated_service_account_email`?

content/docs/command-reference/remote/modify.md
```diff
@@ -271,6 +271,45 @@ including obtaining the necessary credentials, and how to form `gdrive://` URLs.
...
+  ```dvc
+  $ dvc remote modify myremote gdrive_service_account_user_email <user account email>
+  ```
+
+- `gdrive_trash_only` - configures DVC to move remote files to
+  [trash](https://developers.google.com/drive/api/v2/reference/files/trash)
+  instead of
+  [deleting](https://developers.google.com/drive/api/v2/reference/files/delete)
+  them permanently. `false` by default, meaning "delete". Useful for shared
+  drives, where delete permissions may not be given.
```
jorgeorpinel: Do we have a DVC Google org?

- [ ] Let's create a [shared drive](https://developers.google.com/drive/api/v2/about-shareddrives) to test all this with that kind of drive.

jorgeorpinel: Cc @shcheklein 🙂

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -3,7 +3,13 @@
 Follow this guide to setup Google Drive as your <abbr>DVC project</abbr>'s
 [remote storage](/doc/command-reference/remote).
 
-## Configure a Google Cloud project
+## Configure a Google Cloud project (highly recommended)
+
+Optionally, follow this guide to setup your own Google Cloud project and
+[OAuth](https://developers.google.com/identity/protocols/OAuth2) credentials.
```
jorgeorpinel: OK. I'm not seeing how this is optional. You need a GC project to use either client creds (id/secret) or a service account. How can this be skipped @maxhora ? I.e. how do you use the default DVC project? Thanks

maxhora: @jorgeorpinel the change to use client id and secret from the default DVC project was made in the last my commit there https://github.com/iterative/dvc/commit/8a4827017ea171ef0dab49cb79785c997075bab6#diff-21bfd10d01c6715fc8634644730fa720R102
If user haven't specified client_id and client_secret in DVC config the default pair will be used by DVC (user will not need to create their own GCP project in this case).

content/docs/command-reference/remote/modify.md
maxhora: @jorgeorpinel from the referenced article this option allows to do following:

"But what if you want to do the inverse, ie. impersonate a user with a service account? This is a common use case in a service-based architecture: services will perform actions (eg. store a record, retrieve a blob) on behalf of users."

It might be also possible to impersonate service account with user account, but I haven't seen such option before at all. I would suggest just reference official Google Docs about impersonating. Mainly, `gdrive_service_account_user_email` was included into DVC only because PyDrive2 supports such feature and such feature might be usable to advanced GDrive users who already know why they need users impersonating.

content/docs/command-reference/remote/modify.md
shcheklein: we do have DVC app, we do have Iterative org, you can share your personal drive with your work account to test stuff

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: Ping me on Slack, if it's still not clear, @jorgeorpinel

jorgeorpinel: Got it. It's just not clear that you don't have to provide anything other than the gdrive URL, and then the first `dvc push` will give you a link to authorize the DVC app:

image

But I got it now I think. I'll clarify all this, thanks.

shcheklein: Yes! Google drive is in fact the most simple of all remotes to start - and that's what we should clearly communicate. With GD you can start with just an URL. Then if you hit limits (you can check logs and see the message), you should create your own project. Service account and stuff - advanced features.

Also, might potentially check rclone's docs - how they document their shared credentials and how they motivate user on creating their own projects.

jorgeorpinel:
> the whole doc (user guide one) is a mix of 4 sections which follow each other in some random order

I think I've mostly fixed this now. Some changes pending, about to push up. Will re-request your review soon, Ivan.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: OK, added some precisions, including about this now.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Got it, updated now.

But in the Google docs I see no refernces of service accounts impersonating users, only the other way around. Seems like a bit of a hack? E.g. see https://cloud.google.com/iam/docs/understanding-service-accounts

jorgeorpinel: p.s. I guess this one doc kind of hints at this feature: https://developers.google.com/admin-sdk/directory/v1/guides/delegation

jorgeorpinel:
> Probably optional configure Google Project section should go last, introduction should be extremely clear that you usually need to specify the path (the same in the remote add/modify)

Since we're highly recommending it, I didn't move it last. But I did make it extra clear how easy it is to start using GDrive remotes now both in the guide and in the remote add cmd ref.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: OK I've mostly addressed this except didn't mention this progression about first just use URL, if you hit limits create a GC project, etc. Didn't check rclone docs yet either but will do ⏳

content/docs/command-reference/remote/modify.md
jorgeorpinel: OMG I forgot about iterative.ai 🤦‍♂ The thing is though, if I create a folder on the org it will still be owned by me, and the Shared drives doc says:

> Shared drives support files owned by an organization rather than an individual user.

So maybe an org owner user has to create it? Or maybe it doesn't matter and its just all confusing wording...

jorgeorpinel: Yeah I created a folder in my iterative.ai GDrive but I'm pretty sure it's owned by me, not the org. I will share it with my personal Google acct and test around with it but that's a separate thing. Let's call it "shared folder".

If you can try to create an organization-owned shared drive and some folders and subfolders in it, that would be great @shcheklein. See https://support.google.com/a/users/answer/9310249?hl=en&ref_topic=9298590

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,61 @@
 # Setup a Google Drive DVC Remote
 
-Follow this guide to setup Google Drive as your <abbr>DVC project</abbr>'s
-[remote storage](/doc/command-reference/remote).
+In this guide we explain the existing alternatives to setup Google Drive
```
shcheklein: minor: alternatives -> ways (a bit easier to read, since other wise it's easy to misread as an alternative to Google Drive)

shcheklein: we don't explain Basic setup ... let's do a quick start even before URL, or even overview that includes the basic setup (and even authorization) - like very simple a quick? May be very similar to what we have in `dvc remote add` now. And a few links to URLs and "Using your won project" sections?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,61 @@
...
+[remote storage](/doc/command-reference/remote) for your <abbr>DVC
+project</abbr>, along with the different benefits each one brings.
 
-## Configure a Google Cloud project
+Note that to start using a GDrive remote, you only need to add it with a valid
```
shcheklein, line 7: see above, I think we need a quick setup or basic setup or something

this whole doc should be self-contained, so should include `dvc remote add` in the very beginning.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,61 @@
...
+```
+
+## Configure custom Google Cloud project & app (recommended)
+
+Optionally, follow this guide to setup your own Google Cloud project and
+generate OAuth credentials for your <abbr>DVC projects<abbr> to connect to
+Google Drive. Doing so avoids sharing our default GCP project/app with every
+other casual user of GDrive remotes out there. This gives you full control over
+Google API usage and rate limits, and ensures optimal performance to access your
+Google Drive with DVC. This is highly recommended for heavy and enterprise use.
```
shcheklein, line 56: > This is highly recommended for heavy and enterprise use.

Not sure about enterprise. Let's just keep heavy, frequent, or something? Should be enough?

The whole paragraph is too technical and complicated. Can we explain it in some simple way?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,61 @@
...
+
+> Please jump to [Authorization](#authorization) to skip this setup.
```
shcheklein, line 58: why skip? why jump? not clear

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,67 +97,55 @@ team. These credentials are only used to generate the
...
+$ dvc remote modify mygdfolder gdrive_client_secret <client secret>
 ```
 
-[Shared drives](https://developers.google.com/drive/api/v2/about-shareddrives)
-should be referenced by _Folder ID_ in the same way.
-
-3. `appDataFolder` - special
-   [application-specific data](https://developers.google.com/drive/api/v2/appdata)
-   folder only accessible by your application and hidden from the user.
+### Access Google Drive remote with a service account
```
shcheklein, line 117: Simplify title - `Using service accounts`

Make it a separate section, put a link to it in the `Creating you app` and back to `Creating your app` as a prerequisite.

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: Simple cases:

* limits (Retying something ... message should be visible in logs), possibility to ask Google to extend limits independently - they answer quickly
* use service accounts (e.g. some automation, CI/CD, etc)

*

content/docs/command-reference/remote/add.md
```diff
@@ -201,13 +201,25 @@ $ dvc remote add myremote "azure://"
...
+a full guide on using Google Drive as DVC remote storage, including how to form
+`gdrive://` URLs, and high-performance/enterprise setups.
 
 ```dvc
 $ dvc remote add -d myremote gdrive://root/path/to/folder
-$ dvc remote modify myremote gdrive_client_id <client ID>
-$ dvc remote modify myremote gdrive_client_secret <client secret>
+```
+
+To start using a GDrive remote, just use another command that needs it and
```
shcheklein, line 211: another -> any

content/docs/command-reference/remote/modify.md
```diff
@@ -250,27 +250,71 @@ For more information on configuring Azure Storage connection strings, visit
 
 Please check out
 [Setup a Google Drive DVC Remote](/doc/user-guide/setup-google-drive-remote) for
-a full guide on configuring Google Drives for use as DVC remote storage,
-including obtaining the necessary credentials, and how to form `gdrive://` URLs.
+a full guide on using Google Drive as DVC remote storage, including how to form
+`gdrive://` URLs, and high-performance/enterprise setups.
 
 - `url` - remote location URL.
```
shcheklein, line 256: put link to how URL is formed right here

content/docs/command-reference/remote/modify.md
```diff
@@ -250,27 +250,71 @@ For more information on configuring Azure Storage connection strings, visit
 
 Please check out
 [Setup a Google Drive DVC Remote](/doc/user-guide/setup-google-drive-remote) for
-a full guide on configuring Google Drives for use as DVC remote storage,
-including obtaining the necessary credentials, and how to form `gdrive://` URLs.
+a full guide on using Google Drive as DVC remote storage, including how to form
+`gdrive://` URLs, and high-performance/enterprise setups.
```
shcheklein, line 254: same question, it's not about enterprise ... not sure about term high-perofrmance

shcheklein: may be just omit ` including how to form `gdrive://` URLs, and high-performance/enterprise setups.`

content/docs/command-reference/remote/modify.md
```diff
@@ -250,27 +250,71 @@ For more information on configuring Azure Storage connection strings, visit
...
 
 - `url` - remote location URL.
 
   ```dvc
   $ dvc remote modify myremote url gdrive://root/path/to/folder
   ```
 
-- `gdrive_client_id` - Google Project's OAuth 2.0 **client ID**.
+- `gdrive_client_id` - Google Project's **client ID** for authentication with
```
shcheklein, line 262: "Required along with `gdrive_client_secret ` if you use custom app ..." - with a link to the section in the UG

content/docs/command-reference/remote/modify.md
```diff
@@ -250,27 +250,71 @@ For more information on configuring Azure Storage connection strings, visit
...
 
-- `gdrive_client_secret` - Google Project's OAuth 2.0 **client secret**.
+- `gdrive_client_secret` - Google Project's **client secret** for authentication
+  with OAuth 2.0.
 
   ```dvc
   $ dvc remote modify myremote gdrive_client_secret <client secret>
   ```
 
+- `gdrive_use_service_account` - instructs DVC to authenticate using a
```
shcheklein, line 276: check Google Storage add/modify with some structure/suggestions on these service account options

content/docs/command-reference/remote/modify.md
```diff
@@ -250,27 +250,71 @@ For more information on configuring Azure Storage connection strings, visit
...
+
+- `gdrive_service_account_user_email` - email of a user account to
+  [impersonate](https://developers.google.com/admin-sdk/directory/v1/guides/delegation)
+  with the service account.
+
+  ```dvc
+  $ dvc remote modify myremote gdrive_service_account_user_email <user account email>
+  ```
+
+- `gdrive_trash_only` - configures DVC to move remote files to
```
shcheklein, line 307: on GC

content/docs/command-reference/remote/modify.md
```diff
@@ -292,38 +336,22 @@ including obtaining the necessary credentials, and how to form `gdrive://` URLs.
...
+  Make sure that the service account has read/write access (as needed) to the
+  file structure in the remote `url`.
 
   ```dvc
-  $ dvc remote modify myremote credentailpath "/home/.../project-XXXXXXX.json"
+  $ dvc remote modify \
+        myremote credentailpath "/home/.../project-XXXXXXX.json"
   ```
 
-**Accessing data with a service account:**
```
shcheklein, line 303: why removing this?

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Using "ways".
And I did already include a paragraph starting in "Note that to start using a GDrive remote, you only need..." but OK, copied the basic example from `remote add` now.
64e85917

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I just added a basic example under this paragraph now, would that be enough? Our user guides aren't typically super self-contained or fully reproducible like tutorials though, we do link to the cmd refs.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Updated with a bullet list, please check.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Because it's optional but a pretty long section, nice to have a link to the next step?

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: OK, done.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Hmmm, GDrive URL formation is mentioned and linked in in the paragraph above, seems overkill to repeat it here?

content/docs/command-reference/remote/modify.md
jorgeorpinel: Hmmm... At least the note about GDrive URL formation is pretty important to have in both remote add and modify, I think.

content/docs/command-reference/remote/modify.md
shcheklein: that's the whole point ) no one reads long descriptions (unless you really need to clarify some behavior) :) It's better to remove it from that paragraph (at least mention that the link explain URLs as well) ... or repeat it. The way I would read is a I would start with the list of params and it's really better to be self-descriptive more or less.

content/docs/command-reference/remote/add.md
```diff
@@ -201,13 +201,25 @@ $ dvc remote add myremote "azure://"
 
 Please check out
 [Setup a Google Drive DVC Remote](/doc/user-guide/setup-google-drive-remote) for
-a full guide on configuring Google Drives for use as DVC remote storage,
-including obtaining the necessary credentials, and how to form `gdrive://` URLs.
+a full guide on using Google Drive as DVC remote storage, including how to form
+`gdrive://` URLs, and high-performance/enterprise setups.
```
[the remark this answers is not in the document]
jorgeorpinel: K. How about "and setups for advanced use cases" ? Updated.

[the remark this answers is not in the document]
shcheklein: `how to form URLs in different case like shared drives and directories, how to setup your own project if you hit limits, using service account and other advanced configurations`

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,79 @@
...
+## URL format for Google Drive remotes
+
+A GDrive remote URLs is needed for the `dvc remote add` command. It can be
+constructed with a _base_, and an **optional** _path_ to a folder i.e.
+`gdrive://<base>/path/to/folder`. The base can be one of:
+
+1. `root` - indicates your topmost Google Drive directory.
+
+```dvc
+$ dvc remote add -d mygdroot gdrive://root
```
shcheklein, line 34: let's not use empty `//root` as an example - we should probably raise an exception if it's used

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: yes, it's enough .. I was thinking about being more explicit and making a title like "Quick start" or "Basic setup"

content/docs/command-reference/remote/add.md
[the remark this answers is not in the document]
jorgeorpinel: OK that's good but maybe at this point its better to remove the sentence like you suggested in https://github.com/iterative/dvc.org/pull/1096#discussion_r401259115 and https://github.com/iterative/dvc.org/pull/1096#discussion_r401951516 below, and add more links to that doc from each param.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,79 @@
...
 
-Follow this guide to setup Google Drive as your <abbr>DVC project</abbr>'s
-[remote storage](/doc/command-reference/remote).
+In this guide we explain the existing ways to setup Google Drive
+[remote storage](/doc/command-reference/remote) for your <abbr>DVC
+projects</abbr>, along with the different benefits each one brings.
 
-## Configure a Google Cloud project
+Note that to start using a GDrive remote, you only need to add it with a valid
+URL (see next section), and use any DVC command that requires it (see
```
shcheklein, line 8: see next section -> see below how to specify it for shared drive, shared folders, etc ? be a bit more explicit why would I need to take a look there

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,79 @@
...
+Go to the following link in your browser:
+
+    https://accounts.google.com/o/oauth2/auth # ... copy this link
+
+Enter verification code: # <- enter resulting token
+```
+
+See the [Authorization](#authorization) section for more details.
+
+## URL format for Google Drive remotes
```
shcheklein, line 25: do we need to mention that it's Google Drive remotes? We are already in the doc ..

would be better to be explicit here as well - Specifying location ?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,79 @@
...
+```
+
+## Configure custom Google Cloud project & app (recommended)
+
+Optionally, follow this guide to setup your own Google Cloud project and
+generate OAuth credentials for your <abbr>DVC projects<abbr> to connect to
+Google Drive. We highly recommend this for heavy and advanced use because:
+
+- you control your Google API usage and rate limits, being able to request an
+  increase to Google if needed.
```
shcheklein, line 71: to request .. to? do we need a second to? should it be just "request Google to increase"?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,79 @@
...
+   [special hidden folder](https://developers.google.com/drive/api/v2/appdata)
+   (unique per user) meant to store application-specific data. This is a good
+   choice to prevent accidentally deleting remote storage data from the Google
+   Drive web UI.
+
+```dvc
+$ dvc remote add mygdappata gdrive://appDataFolder
+```
+
+## Configure custom Google Cloud project & app (recommended)
```
shcheklein, line 64: we should mention it briefly at least in the intro? 

also, I still don't like the title.

Ideally it should go from the task people would probably trying to solve with this ... not sure how to phrase this in a simple way

How about:

Using custom project
Use custom credentials
Connect with custom project
etc

(recommended) - better to highlight it here, or in the note on the top that mentions this section and provides and why would I need this

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: so, let's put Authorization before all optional sections?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,67 +115,60 @@ team. These credentials are only used to generate the
...
 > [exceed the limits](https://developers.google.com/drive/api/v2/handle-errors?hl=ro#resolve_a_403_error_usage_limit_exceeded).
 
-## URL format for Google Drive remotes
-
-A GDrive DVC remote URLs is needed for the `dvc remote add` command. It can be
-constructed with a _base_, and an **optional** _path_ to a folder i.e.
-`gdrive://<base>/path/to/folder`. The base can be one of:
-
-1. `root` - indicates your topmost Google Drive directory.
+### Enabling your GC project with OAuth credentials
```
shcheklein, line 124: I don't quite understand this title ... why do we enable GC project?

content/docs/command-reference/remote/modify.md
jorgeorpinel: OK, refactored with shorter p about full guide, and additional links to guide sections in params.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Updated

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Yeah you're right. Moved under **URL format** section.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Just refactored so all the info. needed is in the param descriptions.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Shouldn't it be up to the user? If it's not a good practice then I'll use a different example ⏳ but not sure about the exception. Maybe a warning. Anyway, that's a core repo issue 🙂

content/docs/command-reference/remote/modify.md
jorgeorpinel: This param only has an effect when using a custom CG project? I thought it was more related to shared drive/folder access permissions.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Not sure I understand this comment. But remote add doesn't even mention this directly, just links to the full guide. I did already add a link to the service acct section of the full guide under this line. I also added info about which params require which other params. I feel like that may be enough? Otherwise may have to rewrite these cmd ref sections significantly (maybe open a separate issue?)

shcheklein: Sure, up to you.

I meant something like this:

Adding structure like this? Just to combine them and include a link to the full guide and service account google docs? 

Screen Shot 2020-04-01 at 6 41 54 PM

shcheklein: Again, up to you - looks good as it is now.

jorgeorpinel: You may be right about having a separate section but I have doubts about how GCS remotes work to make sure the section is useful in both types. Leaving this as is for now. Sent you some Qs on slack

content/docs/command-reference/remote/modify.md
jorgeorpinel: Discussing in https://github.com/iterative/dvc.org/pull/1096#discussion_r402006323. Resolving here.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Removed line, and opened dvc/issues/3586

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I agree with giving more motivation on URL formation, but I added it in the URL format section and just linked from here instead. See 9d9ebec4. Good?

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Good point, updated in 34f8b410.

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: thanks! a good one. Made me think once again about basic scenario

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: kk, it's fine, but see other comment about the basic case ... it feels that it's wrong that we use `root` and if we switch to something like `gdrive://FOLDER_ID/path` section that explains URLs becomes even more important. And explaining limitations for different URLs becomes very important (e.g. for `grdive://root` it won't be easy to share, and can be even dangerous since `gdrive://root` will be resolved fine and multiple people will push data to two different GD accounts)

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: make it - `Configuring DVC to use you custom credentials`?

content/docs/command-reference/remote/add.md
```diff
@@ -199,17 +199,27 @@ $ dvc remote add myremote "azure://"
...
+[valid URL format](/doc/user-guide/setup-google-drive-remote#url-format-to-specify-folder-location).
+Then simply use any DVC command that needs it (e.g. `dvc pull`, `dvc fetch`,
+`dvc push`), and follow the instructions to connect your Google Drive with DVC.
+For example:
 
 ```dvc
-$ dvc remote add -d myremote gdrive://root/path/to/folder
-$ dvc remote modify myremote gdrive_client_id <client ID>
-$ dvc remote modify myremote gdrive_client_secret <client secret>
+$ dvc remote add -d myremote gdrive://root/dvcstore
```
jorgeorpinel, line 209: > Now thinking a bit more, have some doubt about how we mention root as a good basic start.
From https://github.com/iterative/dvc.org/commit/ff380fc5c3e82faad03a85b6c70c841014a26a50#r38242840

Agree. Will review and also address https://github.com/iterative/dvc/issues/3586#issuecomment-608117910 ⏳

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I think this is addressed now. I did also check out https://rclone.org/drive/ but didn't find much inspiration anyway.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Yes, good points. Will address along with what I summarized in https://github.com/iterative/dvc.org/pull/1096#pullrequestreview-386926510.

content/docs/command-reference/remote/add.md
jorgeorpinel: Also addressing
> ... if we switch to something like gdrive://FOLDER_ID/path section that explains URLs becomes even more important.
> And explaining limitations for different URLs becomes very important (e.g. for grdive://root it won't be easy to share, and can be even dangerous multiple people will push data to two different GD accounts)
⏳

content/docs/command-reference/remote/modify.md
jorgeorpinel: OK I added a little heading and paragraph about service accounts to link to the guide and separate "regular" params from service acct params.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Alright. Title updated and note added in intro to this section, as well as to the service account section.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Yeah. How about we just remove the header actually? I feel like it's simply part of the custom gc project section anyway. See f6b1aebc.

jorgeorpinel: p.s. the intro is kinda long now...

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: p.s. the intro is kinda long now...

content/docs/command-reference/remote/add.md
jorgeorpinel: OK I addressed all these concerns in 452924a7. Please check latest version 🙂

content/docs/command-reference/remote/modify.md
shcheklein: created the drive! resolving

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: yep, will check it again

shcheklein:
CI failed ^^

content/docs/command-reference/remote/modify.md
```diff
@@ -250,29 +250,89 @@ For more information on configuring Azure Storage connection strings, visit
```
shcheklein, line 250: this one and `remote add` look perfect!

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
 # Setup a Google Drive DVC Remote
 
-Follow this guide to setup Google Drive as your <abbr>DVC project</abbr>'s
-[remote storage](/doc/command-reference/remote).
+In this guide we explain the existing ways to setup Google Drive
+[remote storage](/doc/command-reference/remote) for your <abbr>DVC
+projects</abbr>, along with the different benefits each one brings.
 
-## Configure a Google Cloud project
+Note that to start using a GDrive remote, you only need to add it with a
```
shcheklein, line 7: Let's do a "Quick start", "Basic config" or whatever 

It should start with:

`To start using Gdrive remote ...`

Authorization section should be just part of it

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+$ dvc push
+...
+Go to the following link in your browser:
+
+    https://accounts.google.com/o/oauth2/auth # ... copy this link
+
+Enter verification code: # <- enter resulting code
+```
+
+See the [Authorization](#authorization) section for more details on connecting
```
shcheklein, line 22: this separate authorization still feels strange and not needed

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+DVC with Google Drive.
+
+DVC uses the Google Drive API to synchronize your <abbr>DVC project</abbr> data
```
shcheklein, line 25: these two paragraphs can stay at the very beginning, before Quick start

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+quotas, which by default are shared with other GDrive remote storage users. For
+heavy use, it's highly recommended to
+[connect using a custom Google Cloud project](#connect-using-a-custom-google-cloud-project),
+which puts you in control of these limits.
+
+Having your own GC project, it's also possible to
+[use a service account](#using-service-accounts) for automating tasks that need
+to establish GDrive remote connections (e.g. CI/CD).
+
+## URL formats to specify folder location
```
shcheklein, line 36: ... to specify remote (or storage?) location?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+
+1. _Folder ID_ - unique identifier for each Google Drive folder
+
+   To obtain it, navigate to that folder in your web browser, and find it in the
+   address bar. For example, for
+   `https://drive.google.com/drive/folders/0AIac4JZqHhKmUk9PDA`:
+
+   ```dvc
+   $ dvc remote add gdfolder gdrive://0AIac4JZqHhKmUk9PDA
+   $ dvc remote add gdsubfolder \
```
shcheklein, line 51: these two commands should be split with `or`

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+to establish GDrive remote connections (e.g. CI/CD).
+
+## URL formats to specify folder location
+
+There's a few alternatives to construct a GDrive remote URL for different uses,
+such as a folder or subfolder in root, shared folders not owned by your account,
+etc. The URL is formed with a _base_, and an optional _path_ to an **existing**
+folder i.e. `gdrive://<base>/path/to/folder`. The base can be one of:
+
+1. _Folder ID_ - unique identifier for each Google Drive folder
```
shcheklein, line 43: probably mention shared drive, shared folder here

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+[remote storage](/doc/command-reference/remote) for your <abbr>DVC
+projects</abbr>, along with the different benefits each one brings.
 
-## Configure a Google Cloud project
+Note that to start using a GDrive remote, you only need to add it with a
+[valid URL format](#url-format-to-specify-folder-location). Then use any DVC
+command that needs it (e.g. `dvc pull`, `dvc fetch`, `dvc push`). For example:
+
+```dvc
+$ dvc remote add -d gdremote gdrive://0AIac4JZqHhKmUk9PDA/dvcstore
```
shcheklein, line 12: we can explain below that `0AIac4JZqHhKmUk9PDA` is a folder id and check it out how to get it

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+   ```dvc
+   $ dvc remote add mygdfolder gdrive://root/dvcstore
+   ```
+
+   > Although valid, we don't recommend using just `gdrive://root`, as it's not
+   > possible for more than one Google account to access a GDrive root ("My
+   > Drive").
+
+3. `appDataFolder` -
+   [special hidden folder](https://developers.google.com/drive/api/v2/appdata)
```
shcheklein, line 74: also suitable for personal case only

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+
+   ⚠️ Only suitable for personal use, as sharing a remote configured this way
+   would cause DVC to try synchronizing data to/from different Google Drives for
+   every user.
+
+   ```dvc
+   $ dvc remote add mygdfolder gdrive://root/dvcstore
+   ```
+
+   > Although valid, we don't recommend using just `gdrive://root`, as it's not
```
shcheklein, line 69: it will be very hard to understand I believe

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+with a [valid URL](#url-format-to-specify-folder-location), DVC will prompt you
+to visit a special Google authorization web page. There you'll need to sign into
+your Google account. The
+[auth process](https://developers.google.com/drive/api/v2/about-auth) will ask
+you to grant DVC the necessary permissions, and produce a verification code
+needed for DVC to complete the connection. On success, this code will be cached
+in a Git-ignored directory located in `.dvc/tmp/gdrive-user-credentials.json`.
+
+⚠️ In order to prevent unauthorized access to your Google Drive, **do not share
+your verification code with others**. Each team member should go through this
```
shcheklein, line 96: ... don't share produced credentials file (not sure sharing verification code is a big deal)

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,120 @@
...
+## Authorization
+
+On the first usage of a GDrive [remote](/doc/command-reference/remote), for
+example when trying to `dvc push` for the first time after adding the remote
+with a [valid URL](#url-format-to-specify-folder-location), DVC will prompt you
+to visit a special Google authorization web page. There you'll need to sign into
+your Google account. The
+[auth process](https://developers.google.com/drive/api/v2/about-auth) will ask
+you to grant DVC the necessary permissions, and produce a verification code
+needed for DVC to complete the connection. On success, this code will be cached
```
shcheklein, line 92: I don't think it save the code in the file

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,71 +156,46 @@ team. These credentials are only used to generate the
...
+   navigate to **IAM & Admin** in the left sidebar, and select **Service
+   Accounts**. Click **+ CREATE SERVICE ACCOUNT**, on the next screen, enter
+   **Service account name** e.g. "My DVC project", and click **Create**. Select
+   **Continue** at the next **Service account permissions** page, click at **+
+   CREATE KEY**, select **P12** and **Create**. Save generated `.p12` key file
+   at your local disk.
 
-Use the `dvc remote modify` command to set the credentials for each `gdrive://`
-remote, for example:
+1. Copy a downloaded `.p12` file to your DVC project root directory.
```
shcheklein, line 195: this is a bad idea I think, this file must never be put into Git ... so probably should stay in some safe local location

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,71 +156,46 @@ team. These credentials are only used to generate the
...
 
 ```dvc
-$ dvc remote add -d mygdfolder gdrive://root/path/to/folder
```
shcheklein, line 96: it looks like we are missing `3. Setup DVC to use ...`

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,71 +156,46 @@ team. These credentials are only used to generate the
...
-$ dvc remote modify mygdfolder gdrive_client_id <client ID>
-$ dvc remote modify mygdfolder gdrive_client_secret <client secret>
+$ dvc remote modify gdremote gdrive_use_service_account True
```
shcheklein, line 198: True -> true?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -45,71 +156,46 @@ team. These credentials are only used to generate the
...
+Google account associated with your GCP project, and not a specific user. They
+are intended for scenarios where your code needs to access data on its own, e.g.
+running inside a Compute Engine, automatic CI/CD, etc. No interactive user OAuth
+authentication is needed.
 
-```dvc
-$ dvc remote add mygdappata gdrive://appDataFolder
-```
+> This requires having your own
+> [GC project](/doc/user-guide/setup-google-drive-remote#connect-using-a-custom-google-cloud-project)
```
shcheklein, line 183: link is probably broken - please check other places

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: it's still here?

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Good catch. Updated.

jorgeorpinel:
> Ci failed

Hmmm the CI passed actually, @shcheklein: https://circleci.com/gh/iterative/dvc.org/2861

The build failed on Heroku though: https://dashboard.heroku.com/apps/dvc-landing-1095-gxpubiooiy6vx/activity/builds/49718832-9612-4445-87c8-74cc6c26e1db — with

`MissingRequiredParameter: Missing required key 'Bucket' in params`

Also, it takes a very long time during `Generating image thumbnails`. Over 10 minutes.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,122 @@
...
+3. `appDataFolder` -
+   [special hidden folder](https://developers.google.com/drive/api/v2/appdata)
+   (unique per user) meant to store application-specific data. This is a good
+   choice to prevent accidentally deleting remote storage data from the Google
+   Drive web UI.
+
+   ⚠️ Only suitable for personal use.
+
+   ```dvc
+   $ dvc remote add gdappata gdrive://appDataFolder
```
jorgeorpinel, line 82: Should it be lower case, something like `gdrive://appdata` instead, maybe? I don't think we use cammelCase much in these kind of literals inside DVC in general. Not "pythonic", I think. Obviously not a huge deal.

@shcheklein @maxhora

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Well, the Authorization section is pretty long for the "Quick" start, and applies to any URL format and even with a custom GC project. I think that's why it's separate.

jorgeorpinel: In fact there's a link to it from the Custom GC project section.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Created header and other small adjustments in 3692cf09. But let's decide about the Auth section in https://github.com/iterative/dvc.org/pull/1096#pullrequestreview-387659095 below.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Moved

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: p.s. Also moved the Auth section back to the bottom since custom gc project and service acct sections are mentioned before quick start now, per https://github.com/iterative/dvc.org/pull/1096#pullrequestreview-387659149.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: TBH I think this should just be "URL format". We're already in the "Setup a Google Drive DVC Remote" doc (title). So it's the URL format for Google Drive DVC remotes.

I would maybe change the title to **Setup a Google Drive Remote Storage** though.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: It is mentioned in this point, after the console samples. Not on the top because every GD folder has an ID, not just shared ones.

jorgeorpinel: OK I actually went ahead and moved the note up to this paragraph.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Done as well.

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: `appDataFolder` is a predefined Google Drive specific name, it's not something we control. Same as `root`.

shcheklein:
@jorgeorpinel 

> The build failed on Heroku though

fixed!

> Also, it takes a very long time during Generating image thumbnails. Over 10 minutes.

that's what S3 cache is for! @fabiosantoscode and @pavelgrinchenko are still looking into these issues with cache. It should be way faster - similar to blog 2-3 minutes.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,103 @@
...
+etc. The URL is formed with a _base_, and an optional _path_ to an **existing**
+folder i.e. `gdrive://<base>/path/to/folder`. The base can be one of:
+
+1. _Folder ID_ - unique identifier for every Google Drive folder, including
+   [shared folders](https://support.google.com/drive/answer/7166529) and
+   [shared drives](https://support.google.com/a/users/answer/9310351) (these two
+   can only be referenced by ID).
+
+   ```dvc
+   $ dvc remote add gdfolder gdrive://0AIac4JZqHhKmUk9PDA
```
shcheklein, line 50: should we add `-d` here and in other places?

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,103 @@
...
+   ```
+
+   or
+
+   ```dvc
+   $ dvc remote add gdsubfolder \
```
shcheklein, line 56: I'mt not sure about using these names like gdsubfolder, etc - It might look like it is some convention. Probably it's better to keep it simple everywhere - `myremote`

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -42,74 +136,70 @@ API connections, and its
...
-`gdrive://<base>/path/to/folder`. The base can be one of:
-
-1. `root` - indicates your topmost Google Drive directory.
+Finally, use the `dvc remote modify` command to set the credentials (for each
+GDrive remote), for example:
 
 ```dvc
-$ dvc remote add -d mygdroot gdrive://root
-$ dvc remote add mygdfolder gdrive://root/path/to/folder
+$ dvc remote add gdfolder gdrive://0AIac4JZqHhKmUk9PDA
```
shcheklein, line 146: probably we should not mention this command here - better to mention basic workflow and specify additional commands only

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Changed to
   > We don't recommend using `gdrive://root` itself, as it's likely used for
   > many other reasons, and pushing data with DVC here can make it seem messy.

WDYT?

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Good catch. Fixed!

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I had just changed "token" to "code" here, but left Max' wording. I see the code is not saved indeed. It's just a one-time thing. Updated text now, per https://github.com/iterative/dvc.org/pull/1096#pullrequestreview-387659574 above.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Ok, updated. Also added this note:

⚠️ Be careful about sharing the key file with others.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I'm not sure on the location of this note since it's left in a removed line and the line count has changed a lot. But if this was about the Service Acct section, I think it's no longer needed.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I've been keeping them current.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Yes, fixed.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: The note has been removed. The sections were moved around a few times to match the order in the intro and **Quick start**.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Yes, makes sense. But the remote URL is an internal thing, they could be mapped in the code to what Google expects.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: I intentionally only added it in the Quick start section where it's follower by `dvc push`. I don't think that it adds value in these other examples.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: OK, changed.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Removed.

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: this is a good point. If we support root and appData I would still keep names as is - to be more explicit. It might be though for simplicity we can just decide to support only URLs like: 

`gdrive://<folder-id>` in the future to make less error prone.

at least it feels like we should recommend this approach as much as possible.

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: It's not obvious to be honest, if we use `-d` in all other places, let's use it everywhere?

Also, still don't see any value in these special names (one place is left).

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: 👍

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: 👍

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -1,9 +1,105 @@
...
+etc. The URL is formed with a _base_, and an optional _path_ to an **existing**
+folder i.e. `gdrive://<base>/path/to/folder`. The base can be one of:
+
+1. _Folder ID_ - unique identifier for every Google Drive folder, including
+   [shared folders](https://support.google.com/drive/answer/7166529) and
+   [shared drives](https://support.google.com/a/users/answer/9310351)\* (these
+   two can only be referenced by ID).
+
+   ```dvc
+   $ dvc remote myremote gdfolder gdrive://0AIac4JZqHhKmUk9PDA
```
shcheklein, line 50: `dvc remote _add_`

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Names updated now. I accidentally missed those, thanks.

I would rather remove -d from everywhere? The topic of default remotes seems totally separate fom gdrive to me. But then if I remove it from the Quick start example, I then need to add it to `dvc push -r myremote`... Will do that for now... ⏳

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Agree.

> we should recommend this approach

Added "recommended" in folder id, and both other ways have warnings about being only good for personal use, so I think it's clear now.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Fixed, thanks. Also a couple lines down.

jorgeorpinel:
Merging, but lmk your final decision about -d and I'll open a quick separate PR. (See https://github.com/iterative/dvc.org/pull/1096#discussion_r404406708). Thanks!

content/docs/user-guide/setup-google-drive-remote.md
shcheklein: let's def not do `dvc push -r`, please ... better to include `-d` everywhere ... or even keep as is for now

jorgeorpinel: I changed it to dvc remote add --default just in Quick start for now. That way it's more obvious and no need to explain -d or add it everywhere in this doc?

I noticed the dvc push cmd ref examples need an update and I already started another regular updates PR so I can address it there 🙂

jorgeorpinel: (See #1110)

PR #1103, opened 2020-03-31
Regular updates (late March)

### LINKED ISSUE #1102: remote: document new add behavior (sets fist added remote as default)
Seems we missed to document this behavior change in `dvc remote add` but I do remember having that conversation some time ago.

jorgeorpinel:
- More Windows related stuff. (per me actually trying to use and contribute to DVC and docs from Win 10)
- ~~Close #1102~~
- Misc stuff
- Complementing #1107


content/docs/user-guide/contributing/core.md
```diff
@@ -44,7 +44,10 @@ $ git clone git@github.com:<username>/dvc.git
 
 Make sure that you have Python 3 installed. Version 3.6 or higher is required to
 run style checkers on pre-commit. On MacOS, we recommend using `brew` to install
-Python.
+Python. For Windows, we recommend an official
```
shcheklein, line 47: seems too specific to include this tbh

jorgeorpinel: Hmmm OK. But it was really annoying having to figure all this out haha. Windows...

I'm just leaving "For Windows, we recommend an official python.org release." here, since we already had a similar rec for MacOS. Or should we remove both?

shcheklein: Yeah, we do for Mac because it has a pre-installed Python 2 - which is very old and incompatible and should not be used at all

content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -54,6 +54,13 @@ characters. If required, the user can explicitly enable long paths by following
 [this](https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/)
 guide.
 
+## Fix or disable Search Indexing
+
+Search Indexing can also slow down file I/O operations slow on Windows. Try
```
shcheklein, line 59: slow ... slow - type

content/docs/user-guide/contributing/core.md
jorgeorpinel: Good point. But that's the thing though, Window also comes with built in `python` command, but it's an alias that go to the Microsoft Store Python installer. But if you install it that way, it doesn't work on Git Bash or other special shells.

jorgeorpinel: I saw we keep both (see latest version) or remove both, because the Python version that comes on MacOS is 2 anyway, from what I remember, and we already specify that we require py 3.

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: Oops, thanks. Fixed

content/docs/user-guide/contributing/core.md
jorgeorpinel: Merged this as is now. Lmk if you prefer that I remove the notes and I'll send a quick PR for that. Thanks

PR #1106, opened 2020-04-01
Migrate codebase to Typescript and CSS Modules

pavelgrinchenko:
- [x] refactor gatsby-node.js (We should split into docs and blog if it possible at all, move common helpers outside?)
- [x] review `.eslintrc.json` after migration
- [x] remove temporary`.tmp-empty-package` after getting rid of `styled-components`
- [x] utils refactoring 
    - [x] https://github.com/iterative/dvc.org/pull/1091#pullrequestreview-382510226
    - [x] https://github.com/iterative/dvc.org/pull/1091#pullrequestreview-382512151
- [x] review blogpost markdown styles after getting rid of styles components (.layoutBlog, etc) at the end of migration
- [x] update deps


package.json
```diff
@@ -90,12 +90,17 @@
...
+    "@types/lodash.throttle": "^4.1.6",
+    "@types/lodash.topairs": "^4.3.6",
     "@types/promise-polyfill": "^6.0.3",
     "@types/react": "^16.9.27",
     "@types/react-collapse": "^5.0.0",
+    "@types/react-dom": "^16.9.6",
     "@types/react-helmet": "^5.0.15",
     "@types/react-popover": "^0.5.3",
     "@types/react-slick": "^0.23.4",
+    "@types/rehype-react": "^4.0.0",
```
shcheklein, line 103: we already have it in regular `deps`, do we need it in both?

shcheklein: I see, it's a @types deps

src/components/ShowOnly/styles.module.css
```diff
@@ -1,15 +1,15 @@
 .mobile {
-  display: none;
+  display: none !important;
```
shcheklein, line 2: looks like a hack?

pavelgrinchenko: Remake it in another way

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,15 @@
-> You may disregard these recommendations if you used the **Edit on GitHub** button from dvc.org to improve a doc in place.
+> You may disregard these recommendations if you used the **Edit on GitHub**
```
shcheklein, line 1: don't include .github - there is a reason for this file to be formatted this way (long single line) dues to the way Github handles them

.eslintrc.json
```diff
@@ -29,15 +29,15 @@
     ],
     "no-tabs": "error",
     "react/prop-types": "off", // Props should be described as TS types
-    "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/no-var-requires": "warn"
+    "@typescript-eslint/explicit-function-return-type": "off"
```
shcheklein, line 32: any particular reason for this to be disabled?

scripts/deploy-with-s3.js
```diff
@@ -124,7 +124,7 @@ async function main() {
   await remove(publicDir)
 }
 
-main().catch(e => {
+main().catch((e) => {
```
shcheklein, line 127: should we just simplify this rule? looks like consistency breaks readability in simple cases cc @iAdramelk @fabiosantoscode ?

src/components/Blog/Post/index.tsx
```diff
@@ -111,7 +111,7 @@ const BlogPost: React.SFC<IBlogPostData> = ({
               ))}
             </div>
           )}
-          {commentsUrl && ready && !error && (
+          {commentsUrl && ready && !error && typeof result === 'number' && (
```
shcheklein, line 114: just curious, should we raise na exception/log if it's not a number?

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -51,37 +67,34 @@ function SidebarMenuItem({ children, label, path, activePaths, onClick }) {
...
-      psRef.current = new PerfectScrollbar(`#${id}`, {
+    if (!psRef.current && rootRef.current) {
+      psRef.current = new PerfectScrollbar(rootRef.current, {
         wheelPropagation: true
       })
     }
 
-    const node = document.getElementById(currentPath)
-    const parent = document.getElementById(id)
+    const node = document.getElementById(currentPath.replace(/\/$/, ''))
```
shcheklein, line 91: looks like a hack ... why is it needed? is there way to do this properly?

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -27,17 +41,19 @@ function SidebarMenuItem({ children, label, path, activePaths, onClick }) {
...
+        )}
         onClick={onClick}
-        className={isRootParent ? 'docSearch-lvl0' : ''}
       >
         {label}
       </Link>
       {children && (
-        <Collapse isOpened={isActive}>
-          {children.map(item => (
+        <Collapse isOpened={!!isActive}>
```
shcheklein, line 55: ? is there bool(something) in JS ?

src/utils/shared/redirects.js
```diff
@@ -52,7 +52,7 @@ const handleFrontRedirect = (host, pathname, clickEvent) => {
     }
 
     if (redirectUrl.startsWith('/')) {
-      redirectUrl = redirectUrl + location.search
+      redirectUrl = redirectUrl + location.search + location.hash
```
shcheklein, line 55: why this change?

src/utils/node/blog.js
```diff
@@ -0,0 +1,151 @@
+const path = require('path')
```
shcheklein, line 1: like the split, but looks like createPages are not utils?

src/utils/front/scroll.ts
```diff
@@ -0,0 +1,61 @@
+import scroll from 'scroll'
```
shcheklein, line 1: it feels that we just put everything into utils that are not components, not sure this is right

shcheklein:
List of issue I've found so far (Chrome, Mac OS for now)

- Docs: page animation is broken
- right sidebar button are different - style, spacing
- blog, mobile , scroll to the very bottom sharing buttons overlap
- Open CR version -> scroll everything to the very bottom -> open Changelog
- Tooltip (abbr) in docs are not styled properly
- Click -> mobile version -> menu is not hidden
- Main page - fonts, styles are different
- Main page - mobile version Get Started button is broken
- Main page - For data scientists, by data scientists section - align buttons
- We don't render sag anymore (the same in prod) - https://dvc.org/blog/a-public-reddit-dataset
- minor: Aloglia search box is not properly connected with the input box

src/server/middleware/redirects/index.js
```diff
@@ -20,6 +20,17 @@ module.exports = (req, res, next) => {
     dev: !isProduction
   })
 
+  // Disable trailing slash redirect for development mode.
+  // Because it leads to infinite loops as Gatsby in dev mode redirects to urls with trailing slashes
+  if (
+    !isProduction &&
+    location &&
+    location.startsWith('/') &&
+    parsedUrl.pathname === `${location}/`
```
shcheklein, line 29: pathname === `${location}/` ?

so does it mean that in dev mode we do have `/`s?

.eslintrc.json
shcheklein: if it complicates React code unreasonably and hurts readability - should we just omit returns types everywhere for React? what are the best practices here?

pavelgrinchenko: Best practices is to make TS strict as possible to use all advantages of it, so you are right! I removed this rule and refactored all warnings. Furthermore I made tsconfig more stricter.

.github/PULL_REQUEST_TEMPLATE.md
pavelgrinchenko: Added `.github` folder to `.prettierignore` and returned back old formatting

scripts/deploy-with-s3.js
pavelgrinchenko: Prettier's team changed `arrowParens` to `always` in v2. I manually set it to the previous one `avoid` and returned back formatting

src/components/Documentation/Layout/SidebarMenu/index.tsx
pavelgrinchenko: Yes, my bad. It was necessary for testing it in the development mode. Removed.

src/components/Documentation/Layout/SidebarMenu/index.tsx
pavelgrinchenko: `Boolean(something)` or its shortcut `!!something`. They are equivalent

src/server/middleware/redirects/index.js
pavelgrinchenko: Yes. In dev mode we have trailing slashes and I'm not sure we can get rid of them at all... That's why we need extra handling in some another places.

src/utils/front/scroll.ts
pavelgrinchenko: What do you suggest? Where we should store such util functions in another case?

src/utils/shared/redirects.js
pavelgrinchenko: It's necessary in dev mode when we have redirect which removes trailing slash. In this case we miss hash. But you are right - we shouldn't do it always. I added necessary condition.

src/utils/node/blog.js
pavelgrinchenko: I agree, but I could not come up with a better place for these helpers. I can move them to root `helpers` for example. Will it be better?

src/components/Blog/Post/index.tsx
pavelgrinchenko: Changed it to simpler check, because I added return types to all utils. I'm not sure why we should want log it. Such checks necessary for TS typings system. But if you want to be sure that API returns correct data - I think we should validate it in api helpers, not in UI components.

pavelgrinchenko:
> Main page - fonts, styles are different

Could you clarify it, point by point? In this case I can review it step by step too. I already fixed some of them, and some styles was changed after getting rid of duplication.
For example Download and Watch Video buttons now have more balanced paddings
image

> We don't render sag anymore (the same in prod) - https://dvc.org/blog/a-public-reddit-dataset

I’m not sure what do you mean when saying `sag`. Could you clarify?

> Tooltip (abbr) in docs are not styled properly

Previous version was heavily dependent on interpolations from styles-components so I fully refactored it. Styles a little bit another, right. But here we can only improve them. What should I change?

src/utils/node/blog.js
pavelgrinchenko: Moved these helpers to `src/gatsby`

package.json
```diff
@@ -34,124 +34,126 @@
...
+    "gatsby-link": "^2.3.2",
+    "github-markdown-css": "^4.0.0",
+    "iso-url": "^0.4.7",
     "isomorphic-fetch": "^2.2.1",
-    "lodash.fill": "^3.4.0",
     "lodash.includes": "^4.3.0",
     "lodash.kebabcase": "^4.1.1",
     "lodash.startcase": "^4.4.0",
     "lodash.throttle": "^4.1.1",
     "lodash.topairs": "^4.3.0",
```
fabiosantoscode, line 58: aside: we have tree-shaking so we can simply do:

```
import { get } from 'lodash'
```

Knowing only `get` is included in the bundle. So we don't have to install lodash bit by bit.

src/components/Community/Contribute/index.tsx
```diff
@@ -1,25 +1,27 @@
...
 const { description, mobileDescription, title } = data.section.contribute
 
-const logPR = () => logEvent('community', 'contribute-pr')
-const logBlogpost = () => logEvent('community', 'contribute-blogpost')
-const logTalk = () => logEvent('community', 'contribute-talk')
-const logAmbassador = () => logEvent('community', 'contribute-ambassador')
+const logPR = (): void => logEvent('community', 'contribute-pr')
+const logBlogpost = (): void => logEvent('community', 'contribute-blogpost')
+const logTalk = (): void => logEvent('community', 'contribute-talk')
+const logAmbassador = (): void => logEvent('community', 'contribute-ambassador')
```
fabiosantoscode, line 18: IIRC there's an option to not have to mark up every return type, if you get bored of doing that.

src/components/Community/Contribute/index.tsx
```diff
@@ -29,93 +31,87 @@ export default function CommunityContribute({ theme }) {
...
               title="Give a talk"
               action={
-                <Button
-                  theme={theme}
+                <Link
+                  className={sharedStyles.button}
+                  style={theme}
                   href="mailto:info@dvc.org?subject=I want to give a talk!"
                   target="_blank"
-                  rel="noreferrer noopener"
```
fabiosantoscode, line 79: Why remove this?

src/utils/front/ga.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  interface IGaException {
+    hitType: 'exception'
+    exDescription: string
+    exFatal: boolean
+  }
+
+  // eslint-disable-next-line @typescript-eslint/interface-name-prefix
+  interface Window {
+    ga: (command: 'send', data: IGaEvent | IGaException) => void
```
fabiosantoscode, line 17: Should be `ga?`. Ad blockers and privacy extensions cause `ga` to not load.

pavelgrinchenko: `window.ga` is set by `gatsby-plugin-google-analytics`. I only described its interface

src/components/Community/Contribute/index.tsx
pavelgrinchenko: It's now automatically handled by Link component

src/components/Community/Contribute/index.tsx
pavelgrinchenko: Yep, but we intentionally enabled this rule to have more strict type checking

package.json
pavelgrinchenko: Unfortunately not =( Tree shaking usually works like a 💩 
There are a lot articles which describe this problem, like this one https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/
But we can replace it via `import get from lodash/get` and it should work, but maybe it can be done in another PR (or tomorrow if it still won't be merged)

Issue #986, opened 2020-02-10
use-cases: use 'dvc cache dir' command in shared dev server


dmpetrov:
In the shared server use case https://dvc.org/doc/use-cases/shared-development-server we set up the shared cache by changing config `dvc config cache.dir /path/to/dvc-cache` while a separate command was introduced `dvc cache dir /path/to/dvc-cache`.

I'm trying to understand the motivation behind the command `dvc cache dir`... It feels like we should deprecate the command `dvc cache dir` or change the docs.

@efiop could you please explain the motivation behind the command?

UPDATE: also probably we should update dvc cache dir doc as well - it's not that clear from the doc (from https://github.com/iterative/dvc.org/issues/986#issuecomment-586541617)

shcheklein:
The doc is just outdated. 

There are two things behind this command:

https://github.com/iterative/dvc/issues/1284
https://github.com/iterative/dvc/issues/1421#issue-388849219
https://github.com/iterative/dvc/pull/1467

(didn't find specific discussion re the `dvc cache dir` itself, but the point is - we want to keep `dvc config` w/o any magic, w/o any transformations, while higher level commands like `dvc cache dir` provide some use case specific, higher level interface)

dmpetrov:
@shcheklein thanks for the clarification.

> we want to keep `dvc config` w/o any magic, w/o any transformations, while higher level commands like

It makes sense. The current `dvc cache dir` still seems to provide very little value for a dedicated command but it can be extended in the future. If we follow this logic `dvc config cache.shared group` should be also replaced to something higher level.

So, we can keep this issue to update the docs or close it if duplication. @jorgeorpinel you should know better.

jorgeorpinel:
Sure, I'll look into updating the docs and close this then. Thanks

jorgeorpinel:
> The doc is just outdated.

Why is https://dvc.org/doc/use-cases/shared-development-server outdated @shcheklein ? For an external cache you would probably prefer to use `dvc config cache.dir` instead of `dvc cache dir` because its easier to use the absolute path than a path relative to the project.

shcheklein:
https://dvc.org/doc/command-reference/cache/dir#example-using-absolute-path - `Absolute path /path/to/dir saved as is.` (Probably we should update `dvc cache dir` doc as well - it's not that clear from the doc).

If we have two APIs and one is made to be user friendly why not advertise it? 

[...]

PR #1110, opened 2020-04-06
Regular updates (Apr 6)

jorgeorpinel:
- Address https://github.com/iterative/dvc/pull/3566#pullrequestreview-388346381
- Address https://github.com/iterative/dvc.org/pull/1096#pullrequestreview-387865301 about `remote add -d`
- Review external data user guide and shared dev server use case to clarify shared/external cache concept
- Complement https://github.com/iterative/dvc.org/pull/1090#pullrequestreview-386203081 (import/update --rev changes)
- Misc. stuff

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -84,9 +84,9 @@ data artifact from the source repo.
   download the file or directory from. The latest commit in `master` (tip of the
   default branch) is used by default when this option is not specified.
 
-  > Note that this adds a `rev` field in the import stage that fixes it to this
-  > revision. This can impact the behavior of `dvc update`. (See
-  > **re-importing** example below.)
+  > Note that this adds a `rev` field in the import stage that fixes it to the
+  > revision. This can impact the behavior of `dvc update`. (See the **Importing
+  > and updating fixed revisions** example below.)
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
@@ -172,6 +172,9 @@ different commit with `dvc update --rev`:
 ```dvc
 $ dvc update --rev cats-dogs-v2
 ```
+
+> In the above example, the value for `rev` in the new import stage will be
+> `master` (a branch) so it will be able update normally going forward.
 
 ## Example: Data registry
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -85,8 +85,9 @@ data artifact from the source repo.
   default branch) is used by default when this option is not specified.
 
   > Note that this adds a `rev` field in the import stage that fixes it to the
-  > revision. This can impact the behavior of `dvc update`. (See the **Importing
-  > and updating fixed revisions** example below.)
+  > revision. This can impact the behavior of `dvc update`. (See the
+  > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
+  > example below.)
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/update.md
+++ b/content/docs/command-reference/update.md
@@ -81,7 +81,7 @@ stable.
 > [DVC-file](/doc/user-guide/dvc-file-format) (when there are changes to bring
 > in).
 
-## Example: Updating fixed-revision imports to a different version
+## Example: Updating fixed revisions to a different version
 
 > See also
 > [Importing and updating fixed revisions](/doc/command-reference/import#example-importing-and-updating-fixed-revisions).
--- a/content/docs/command-reference/update.md
+++ b/content/docs/command-reference/update.md
@@ -36,14 +36,13 @@ dvc update --rev master
 
 ## Options
 
-- `--rev` - specific
-  [Git revision](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
-  (such as a branch name, a tag, or a commit hash) of the repository to update
-  the file or directory from (also starts tracking the given revision).
+- `--rev` - commit hash, branch or tag name, etc. (any
+  [Git revision](https://git-scm.com/docs/revisions)) of the repository to
+  update the file or directory from. The latest commit in `master` (tip of the
+  default branch) is used by default when this option is not specified.
 
-  > Note that this adds or updates a `rev` field in the DVC-file that fixes it
-  > to this revision (and updates `rev_lock` in the DVC-file). This can have an
-  > impact on the behavior of `dvc update` later.
+  > Note that this changes the `rev` field in the import stage, fixing it to the
+  > revision.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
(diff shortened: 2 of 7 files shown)


content/docs/use-cases/shared-development-server.md
```diff
@@ -3,102 +3,110 @@
...
+so that they can be accessed by your colleagues (assuming their users are
+members of the same group):
 
 ```dvc
-$ sudo find /path/to/dvc-cache -type f -exec chmod 0664 {} \;
-$ sudo find /path/to/dvc-cache -type d -exec chmod 0775 {} \;
-$ sudo chown -R myuser:ourgroup /path/to/dvc-cache/
+$ sudo find /home/shared/dvc-cache -type d -exec chmod 0775 {} \;
+$ sudo find /home/shared/dvc-cache -type f -exec chmod 0664 {} \;
+$ sudo chown -R myuser:ourgroup /home/shared/dvc-cache/
```
jorgeorpinel, line 49: I'm not sure these commands will always work. What if the shared cache dir has some files your user can't access (if someone didn't do this process before, for example)? Wouldn't the command fail?

content/docs/use-cases/shared-development-server.md
```diff
@@ -3,102 +3,110 @@
...
 
-Tell DVC to use the directory we've set up above as an shared cache location by
-running:
+Tell DVC to use the directory we've set up above as the <abbr>cache</abbr> for
+your <abbr>project</abbr>:
 
 ```dvc
-$ dvc config cache.dir /path/to/dvc-cache
+$ dvc config cache.dir /home/shared/dvc-cache
 ```
```
jorgeorpinel, line 59: Why do we have all these exact and reproducible command guides in a use case? This section seems more like a user guide to me.

content/docs/user-guide/managing-external-data.md
```diff
@@ -29,22 +29,22 @@ supported:
...
 
-> Avoid using the same remote location that you are using for `dvc push`,
-> `dvc pull`, `dvc fetch` as external cache for your external outputs, because
-> it may cause possible file hash overlaps: The hash value of a data file in
-> external storage could collide with that generated locally for another file.
+> Avoid using the same location of the
+> [remote storage](/doc/command-reference/remote) that you have for `dvc push`
+> and `dvc pull` for external outputs or as external cache, because it may cause
+> file hash overlaps: The hash value of a data file in external storage could
+> collide with the one generated locally for another file.
```
jorgeorpinel, line 41: Improved this note but it's still pretty strange. Difficult to understand what it is about. The ambiguity of the term "remote" is giving us problems here. Not sure how to address, maybe just remove the note?

content/docs/command-reference/import.md
```diff
@@ -84,9 +84,9 @@ data artifact from the source repo.
   download the file or directory from. The latest commit in `master` (tip of the
   default branch) is used by default when this option is not specified.
 
-  > Note that this adds a `rev` field in the import stage that fixes it to this
-  > revision. This can impact the behavior of `dvc update`. (See
-  > **re-importing** example below.)
+  > Note that this adds a `rev` field in the import stage that fixes it to the
+  > revision. This can impact the behavior of `dvc update`. (See the **Importing
+  > and updating fixed revisions** example below.)
```
shcheklein, line 89: minor - make it a link?

content/docs/command-reference/update.md
```diff
@@ -82,15 +81,19 @@ stable.
 > [DVC-file](/doc/user-guide/dvc-file-format) (when there are changes to bring
 > in).
 
-## Example: Updating imported artifacts to a specified revision
+## Example: Updating fixed-revision imports to a different version
```
shcheklein, line 84: `fixed-revision imports` sounds very technical :)

content/docs/use-cases/shared-development-server.md
```diff
@@ -3,102 +3,110 @@
...
-$ mkdir -p /path/to/dvc-cache
+$ mkdir -p /home/shared/dvc-cache
 ```
 
-You will have to make sure that the directory has proper permissions setup, so
-that all your colleagues can read and write to it, and can access cache files
-written by others. The most straightforward way to do this is to make sure that
-everyone's users are members of the same group, and that your shared cache
-directory is owned by this group, with the aforementioned permissions.
+> The `/home/shared` directory used as example above is typical in Linux
```
shcheklein, line 21: no, let's not include notes like this :)

content/docs/use-cases/shared-development-server.md
shcheklein: yes, I agree ... most of it should be part of the `shared cache` article in the UG. And use case should be probably more generic "Optimizing data management" which can cover this one (multiple people working on a single box), k8s scenario - multiple machines + NAS, etc. We are long overdue on this. There were contributors that made an attempt, but didn't get it right.

content/docs/user-guide/managing-external-data.md
shcheklein: yep, it's a very complicated to simplify it :) the whole topic is very advanced and very technical. Let's keep it for now with your revisions and get back to it when we have more time to spend on all this data management problems.

shcheklein:
@jorgeorpinel just minor comments, overall looks good to merge.

content/docs/command-reference/update.md
jorgeorpinel: Changing to "Updating fixed revisions to a different version".

content/docs/use-cases/shared-development-server.md
jorgeorpinel: Ok, reverting. But I just copied it from https://dvc.org/doc/user-guide/managing-external-data#examples where it's already being used. Should I remove that one too?

jorgeorpinel: I'll just remove it from there too.

shcheklein: 👍

content/docs/use-cases/shared-development-server.md
jorgeorpinel: OK. I see we have #429 for something like this already. And #986 is also related.  Should we try to consolidate them into a single issue and include the comments from this review?

shcheklein: so, these two mostly about improving the existing case - but they are relevant, up to you to consolidate or not ... there should be tickets related to this from other angles - shared cache on NAS/NFS for example

shcheklein: Some other related links:

https://discuss.dvc.org/t/setup-dvc-to-work-with-shared-data-on-nas-server/180/4
https://github.com/iterative/dvc.org/pull/732
https://github.com/iterative/dvc.org/pull/455
https://github.com/iterative/dvc.org/issues/563

and others ... someone should a first step systematize and suggest the way to structure this properly

Issue #1064, opened 2020-03-17
ci: restyled seems broken


jorgeorpinel:
~~per https://github.com/iterative/dvc.org/pull/1058#pullrequestreview-376234425~~

Let's double check it's still working and that it can fix as much as possible from our format checks...

jorgeorpinel:
Now Restyled seems completely removed from the repo. Maybe a recent merge guys? @iAdramelk @pavelgrinchenko @fabiosantoscode 

Thanks

shcheklein:
@jorgeorpinel it's not, we just need to update and make sure it works - https://github.com/iterative/dvc.org/blob/master/.restyled.yaml

jorgeorpinel:
Yeah turns out someone added `stylelint` but it's not supported. I just removed it in 20c17baf5ce5c5e83096878a86c9da5260264299.

Should we add any others? Availeble ones: https://github.com/restyled-io/restyled.io/wiki/Available-Restylers

shcheklein:
sounds good, I would not worry about even JS files to be honest ... mostly we should provide simple workflow for md files.

can we test it?

shcheklein:
Fixed, see #1159

PR #1131, opened 2020-04-10
Regular updates (Apr 9)

jorgeorpinel:
- Updates to metrics diff to match language in params diff (#1152)
- Misc. copy edits
- Enabled Restyled again (see #1064)


content/docs/command-reference/run.md
```diff
@@ -10,8 +10,6 @@ usage: dvc run [-h] [-q | -v] [-d <path>] [-o <path>] [-O <path>]
                [-p <params>] [-m <path>] [-M <path>] [-f <filename>]
                [-w <path>] [--no-exec] [-y] [--overwrite-dvcfile]
                [--ignore-build-cache] [--remove-outs] [--no-commit]
-               [--outs-persist OUTS_PERSIST]
-               [--outs-persist-no-cache OUTS_PERSIST_NO_CACHE]
```
jorgeorpinel, line 14: This 2 options appeared after merging the params PR. These are old and no longer existent right? @shcheklein @efiop because they're not documented in the cmd ref. Just wondering what's going on here...

jorgeorpinel: Hmmm, I see they come out in `dvc run -h` so are these new? Can someone give me their description or link to the original PR so I reinstate them here and add them to the cmd ref?

jorgeorpinel: I've reinstated them and added basic descriptions to the cmd ref in afeba4e9 for now.

shcheklein: We were deliberately not documenting this :) But it's fine, I think we can keep that simple explanation.

jorgeorpinel: Kk

PR #1135, opened 2020-04-10
Removed repeating 'myremote' remote name

v-goncharenko:
There was double remote name, so it didn't work.


shcheklein:
thanks @v-goncharenko !

PR #1136, opened 2020-04-10
add meta tags for SEO

fabiosantoscode:
Port of https://github.com/iterative/blog/pull/144

On monday I'll see if all the which should be in here, are.

I've removed some stuff from the blog SEO which was repeated from the DefaultSEO component.

### DIFF
--- a/src/components/Page/DefaultSEO/index.tsx
+++ b/src/components/Page/DefaultSEO/index.tsx
@@ -1,23 +1,25 @@
 import React from 'react'
-import ReactHelmet from 'react-helmet'
+import Helmet from 'react-helmet'
 
 import { MetaProps } from '../../SEO'
-import { IPaginatorPageInfo } from '../../Paginator'
 import getSiteMeta from '../../../queries/siteMeta'
 
 interface IDefaultSEOProps {
-  pageInfo?: IPaginatorPageInfo
+  pathname: string
 }
 
-const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
+const metaImage = {
+  src: '/social-share.png',
+  width: '1200',
+  height: '630'
+}
+
+const DefaultSEO: React.SFC<IDefaultSEOProps> = ({ pathname }) => {
   const siteMeta = getSiteMeta()
-  const metaTitle =
-    pageInfo && pageInfo.currentPage > 1
-      ? `${siteMeta.title} page ${pageInfo.currentPage}`
-      : siteMeta.title
+  const metaTitle = siteMeta.title
   const metaDescription = siteMeta.description
   const metaKeywords = siteMeta.keywords
-  const metaImage = '/social-share.png'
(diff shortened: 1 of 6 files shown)


src/components/SEO/index.tsx
```diff
@@ -72,18 +76,28 @@ const SEO: React.FC<ISEOProps> = ({
...
+        property: 'og:image:width',
+        content: String(image.presentationWidth)
+      },
+      {
+        property: 'og:image:height',
+        content: String(image.presentationHeight)
       },
       {
         name: 'twitter:image',
-        content: encodeURI(`${siteMeta.siteUrl}${image}`)
```
shcheklein, line 86: was `encodeURI` actually needed? cc @iAdramelk ?

src/templates/blog-post.tsx
```diff
@@ -67,24 +68,17 @@ interface IBlogPostPageProps {
...
     markdownRemark: IBlogPostData
   }
-  pageContext: {
-    next: IBlogPostData
-    previous: IBlogPostData
-  }
 }
 
-const BlogPostPage: React.FC<IBlogPostPageProps> = ({ data }) => {
+const BlogPostPage: React.SFC<IBlogPostPageProps> = ({ data }) => {
```
shcheklein, line 73: SFC is deprecated?

src/components/Page/DefaultSEO/index.tsx
```diff
@@ -76,21 +94,13 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
       meta={meta}
       link={[
         {
-          rel: 'shortcut icon',
+          rel: 'icon',
           type: 'image/vnd.microsoft.icon',
```
shcheklein, line 98: is this type required? legacy?

src/components/Page/DefaultSEO/index.tsx
```diff
@@ -76,21 +94,13 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
...
           href: '/favicon.ico'
         },
         {
-          rel: 'icon',
```
shcheklein, line 84: why are we removing these?

src/components/SEO/index.tsx
iAdramelk: To be hones I'm not sure. The one who originally added it to the blog was you https://github.com/iterative/blog/commit/3b5c975363fd4f86ef66dd102a2a130759df8fcb . I think then that you where fixing some kind of bug that we have without escaping, but if not, I'm not sure that we need it.

shcheklein: Good question. I see that I was solving a different problem in the first place - absolute URL vs relative one. Don't remember at all why did I put `encodeURI` to be honest.

shcheklein:
Closing this in favor of #1183

Issue #1073, opened 2020-03-19
create a dedicated Home page for docs at dvc.org/doc (design part left)


jorgeorpinel:
In a reorganization of docs (#1051) we're creating a Home page that /doc redirects to.

We'd like it to BE in /doc. But keeping the **Home** entry in the nav sidebar. Is this easy with the current Gatsby engine?

---------

- [x] It should have dedicated icon. Related https://github.com/iterative/dvc.org/issues/1053.
- [x] Putting it in `dvc.org/doc` most likely means it's a separate page, rather than a regular Markdown file.
- [ ] Design. It should have a table with existing sections, with graphics, etc. Again, probably means it's a separate page rather than a regular one.
- [x] On the sidebar it should stay separate under the search bar.
- [x] Main topics should have a special CSS design (after #1074 is merged):
  ![image](https://user-images.githubusercontent.com/1477535/79699242-da06da00-8253-11ea-9dde-5359bfe5cd7b.png)
  > UPDATE: Probably 4 tiles instead, jump to https://github.com/iterative/dvc.org/issues/1073#issuecomment-630552134

  Good example (at least the middle section) in https://cloud.google.com/sql/docs/mysql:
  ![image](https://user-images.githubusercontent.com/1477535/80674319-2d103680-8a77-11ea-9cfb-9b50c02d82af.png)

jorgeorpinel:
But keeping the **Home** entry in the nav sidebar...

iAdramelk:
@pavelgrinchenko can you check this issue please?

pavelgrinchenko:
@jorgeorpinel In the current codebase you can't do it without changes in the `src/utils/sidebar.js`. If you want you can show your branch with changes and I'll do it for you.

pavelgrinchenko:
I prepared patch. You can get it [here](https://github.com/iterative/dvc.org/commit/69ec1e82), and then just set something like this in sidebar's config
```
{
    "label": "Home",
    "slug": "",
    "source": "home/index.md"
}
```

jorgeorpinel:
Thanks @pavelgrinchenko that works! Cherry-picked your patch into my working branch now :)
See https://github.com/iterative/dvc.org/compare/9ddaf88..ef4ece6

shcheklein:
Updated the title and description. I think Home page is a separate sub-project and a special page. For the SEO purpose we should do it quick

jorgeorpinel:
[...]

Issue #1120, opened 2020-01-03
blog: make author name clickable, show description and posts by the author


shcheklein:

utkarshsingh99:
Does this mean a new template would have to be added in the src/templates folder additionally?  
Since there is no page that has a profile like page.  
Also, how should be the design?  
I think even a graphql query will be required to fill in the specific posts based on the authors.

jorgeorpinel:
@utkarshsingh99 as for the technical implementation, that would be the role of whoever takes this issue. Ivan or maybe @rogermparent can probalby give you some quick tips if needed. The design is probably not a big deal initially, once the page exists in a PR, we can worry about that part.

rogermparent:
> Does this mean a new template would have to be added in the src/templates folder additionally?
> Since there is no page that has a profile like page.

Yes, I believe so

> Also, how should be the design?

I'm not very good at design, so I can't help you here.

> I think even a graphql query will be required to fill in the specific posts based on the authors.

Yes, we have separate author nodes so we just need to use a GraphQL query in the `createPages` logic to make Author pages, then make links wherever Authors are already present.

The current `gatsby-node` logic for Authors is at `src/gatsby/models/authors` where you'd make an `onCreatePages` file. The GraphIQL tool in `gatsby develop` can help a lot here in showing you what data is already exposed.

Let me know if you need any help! Most of the schema logic handling stuff related to this was written by me, so I can help you find your way around.

utkarshsingh99:
Sure. If it's okay, I can start working on this.

rogermparent:
> Sure. If it's okay, I can start working on this.

I don't see any problems with that! I'll focus on other issues and won't hop in unless you need me to.

utkarshsingh99:
My idea is to re-use the **Feed** component used in the template `blog-home.tsx`, and pass the Author Name as _header_, and Description as _leadParagraph_ props.
Please tell me if my approach is correct:
1. I create a `blog-authors.tsx`, and in `src/gatsby/models/authors/createPages.js` make a graphql query listing all authors with their posts.
2. I call the `blog-authors.tsx` file in createPages subsequently.
3. The `blog-authors.tsx` will be almost exactly like the blog Home template. (maybe some styling to make it look a little different, but the good thing is we can have bigFirst post and ordered pairs for the rest of the posts)
[...]

Issue #1084, opened 2020-03-21
dev mode improvements


jorgeorpinel:
- [x] Easy: Can we name the yarn script just `dev` like before? faster to type

- [x] Reasonable: Can it have watch/autoreload? ~~Or perhaps allow reloading the current page by clicking on the current nav item (used to work).~~

- [x] Changing sudebar.json requires stopping `yarn develop` and running it again (this was automatic before with Next.js)

- [ ] Maybe hard: Can dev mode somehow purge the Gatsby cache automatically?
  There are times when I make a change to the code or contents, I simply don't get the changes reflected. Even stopping `yarn develop` and running it isn't enough. I have to stop it, `rm -rf .cache/`, and run it again.
  E.g. this happens when changing redirects, even if the browser cache is disabled.

jorgeorpinel:
p.s. can we name the yarn script dev like before? faster to type 😬

jorgeorpinel:
Also, changing sidebar.json doesn't reflect at reload until you stop and rerun `yarn develop` (not cache related).

jorgeorpinel:
Updated the description above to summarize several possible improvements to the local dev mode on Gatsby.

iAdramelk:
CC @pavelgrinchenko 

Some quick answers:

> 2. Can it have watch/autoreload? Or perhaps allow reloading the current page by clicking on the current nav item (used to work).
> E.g. changing sudebar.json requires stopping yarn develop and running it again (this was automatic before with Next.js)

It watch/autoreload for the most part, but right now `sidebar.json` is an exception to some degree. Changing titles and katacode links should still autoreload, but changing slug part will not automatically create page on the new url because this change didn't start page rebuild process (it will probably update menu, but actual page will not be created).

We can either fix it by implementing this issue https://github.com/iterative/dvc.org/issues/1044 (renaming md file or creating new md file will automatically trigger this page build). Or we can look for a way to say to `gatsby` the it should rebuild all of the doc pages on the `sidebar.json` changes. Not sure that it possible, but there may be a hook for that.

> 3. Maybe hard: Can dev mode somehow purge the Gatsby cache automatically?
> There are times when I make a change to the code or contents, I simply don't get the changes reflected. Even stopping yarn develop and running it isn't enough. I have to stop it, rm -rf .cache/, and run it again.
> E.g. this happens when changing redirects, even if the browser cache is disabled.

[...]

PR #1145, opened 2020-04-14
Custom Node Types and big Gatsby refactor

rogermparent:
Hi everyone! I just discovered this project from an advertisement on the Gatsby jobs board on Discord. I have a lot of time with Gatsby under my belt, especially the schema, and this PR is an application of some best practices I've picked up along the way.

- Blog Posts, Docs Pages, and Authors now have their own custom node types. Fields that used to be accessed through `MarkdownRemark` are all copied and accessible through the custom nodes with old nested fields like `frontmatter` and `childImageSharp` flattened to the Node's root level.
  While not as initially simple as directly accessing remark nodes, these custom types pave the way for more complex logic and a more concise API in the future!

- The old `blog.js` and `docs.js` helpers for `createPage` have now evolved into a dedicated pattern that spans the whole Gatsby lifecycle: "Models".
  Models are basically tiny Gatsby plugins to the point where they respond to the same hooks, but Models are local to this repository and can respond to custom events like `onCreateMarkdownContentNode`.
  Models are simply to organize code specific to this repo, and can easily be broken out into dedicated plugins later if need be.

- I put a few little flags in page creation that skip most pages while keeping at least one of each kind when `NODE_ENV` is `development`. This means `gatsby develop` only has to remake a bit over 200 images after cache clear instead of 1000+. Developers can still build the full site with `gatsby build`, and production isn't affected. Still, this is a tad hacky so I'm open to removing it if necessary.

- The new `Author` node has a `posts` field with a `limit` argument, which paves the way to fixing #1120 after a few design choices. My vote is for linked separate Author pages, but I'm not a designer.

I understand this is a massive PR, but there aren't many points outside of the existing commits that can be removed from the final product without breaking the build. Feel free to ask me anything, and I'm open to any changes that need to be made!


rogermparent:
I should mention there's a few less-than-dry parts of the code on node generation, but I left them in there for easier adaptation to node-specific changes.

gatsby-config.js
```diff
@@ -1,6 +1,6 @@
-/* eslint-env node */
+//redirector.googlevideo.com/videoplayback?expire=1586853957&ei=JVyVXv3xGJfjkgbV0qOQBQ&ip=104.223.33.41&id=a260613bb3f6d9b2&itag=37&source=picasa&begin=0&requiressl=yes&mh=Du&mm=30&mn=sn-a5mlrnel&ms=nxu&mv=u&mvi=2&pl=17&sc=yes&ttl=transient&susc=ph&app=fife&mime=video/mp4&cnr=14&dur=1434.273&lmt=1586782374367924&mt=1586846464&sparams=expire,ei,ip,id,itag,source,requiressl,ttl,susc,app,mime,cnr,dur,lmt&sig=AJpPlLswRQIgfdQ5lyTw4MaUyYzL6sU5p9kKwtYJ1jocgr87QjZ7RwQCIQDZ2qd3BdfsqDO1itbNBlV0x8XtO2khf3Hc1YEqOcqhpA==&lsparams=mh,mm,mn,ms,mv,mvi,pl,sc&lsig=ALrAebAwRQIgIrELVT4JW3z2TWN4nOW9rJ9Ck1ZH5jTyuCvf7xLeBhICIQDFt3xVEWMsw8kVHx3SuDlhkRg3cyL5UL2hmcA9mYziig==&jparams=MTA0LjE5Mi4xMzAuMTkz&upx=TW96aWxsYS81LjAgKFgxMTsgTGludXggeDg2XzY0OyBydjo3NS4wKSBHZWNrby8yMDEwMDEwMSBGaXJlZm94Lzc1LjA=#betaHost&tr=2/* eslint-env node */
```
shcheklein, line 1: could you clarify, what is this link for?

rogermparent: Oh my goodness, that was an accidental paste and went over my head. I'm so sorry!

rogermparent:
This PR can be adapted to fix #1073 by dropping a ternary [in the docs model's path resolution](https://github.com/iterative/dvc.org/blob/92b8efbe75939feebc45dc4762229d643f379b13/src/gatsby/models/docs/onCreateNode.js#L13)

src/gatsby/utils/models/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+    1. In gatsby-node, outside of any api calls, store the result of this function called with
+       an array of all models.
+    2. Call the function returned from step 1 with the name of the function you want to run on
+       all models.
+    3. Use the value from step 2 as a gatsby call, or manually call it within another function using
+       the first two positional parameters from Gatsby.
+    4. Return or await the promise received from step 3. Don't worry about the promises' return values.
+ */
+function buildModelApi(models, childAPI = { models }) {
```
shcheklein, line 14: To be honest this looks like a premature over-generalization. We have only one set of models right, so far, right? Why don't we just make a simple map wrapper that takes a string and run on all of them.

src/gatsby/utils/models/index.js
```diff
@@ -0,0 +1,28 @@
+/** Model API builder
```
shcheklein, line 1: why do we use `js` everywhere, not TS?

gatsby-node.js
```diff
@@ -1,27 +1,15 @@
...
-  getNodeSlug,
   setPageContext,
   removePageTrailingSlash
 } = require('./src/gatsby/common')
-const { createPages: createDocPages } = require('./src/gatsby/doc')
-const { createPages: createBlogPages } = require('./src/gatsby/blog')
 
-exports.onCreateNode = ({ node, actions, getNode }) => {
-  const { createNodeField } = actions
+const buildModelApi = require('./src/gatsby/utils/models')
```
shcheklein, line 6: see my comment below in the `buildModelApi` definition, I think we can simplify and not create multi-layered builder to this simple task

src/gatsby/models/blog/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  createPages: require('./createPages.js'),
+  onCreateMarkdownContentNode: require('./onCreateNode.js'),
+  createSchemaCustomization: require('./createSchemaCustomization')
```
shcheklein, line 4: should it end with `.js` like everywhere else?

src/gatsby/models/blog/createPages.js
```diff
@@ -91,13 +75,19 @@ const createPages = async ({ graphql, actions }) => {
...
+  } = blogResponse.data.allBlogPost
+
   // Create home blog pages (with pagination)
   const blogHomeTemplate = path.resolve('./src/templates/blog-home.tsx')
 
   for (const page of pagesGenerator({
     basePath: '/blog',
     hasHeroItem: true,
-    itemCount: blogResponse.data.home.pageInfo.itemCount
+    itemCount: isDevMode ? 1 : itemCount
```
shcheklein, line 90: it is useful! we should though make some file that engineers can change locally to set this, or at least read ENV var.

shcheklein: why do we need to limit itemCount here btw?

src/gatsby/utils/resolvers/index.js
```diff
@@ -0,0 +1,79 @@
...
+          }
+        }
+      },
+      type,
+      firstOnly: true
+    })
+  }
+}
+
+function resolveRelativeImage() {
```
shcheklein, line 45: does it repeat `gatsby-remark-relative-images` that we had before? (I know that the newer version of `gatsby-remark-relative-images` failed, but should we at least remove it from the dependcies?)

src/gatsby/utils/resolvers/index.js
```diff
@@ -0,0 +1,79 @@
...
+  return async function (source, args, context, info) {
+    const relativeChildPath = context.defaultFieldResolver(
```
shcheklein, line 47: need to ask @jorgeorpinel eventually to test this on Windows

src/gatsby/models/docs/onCreateNode.js
```diff
@@ -0,0 +1,39 @@
...
+  const splitDir = parentNode.relativeDirectory.split('/')
+  if (splitDir[0] !== 'docs') return
+  // Make a special exemption for the root doc.
+
+  const { node, actions, createNodeId, createContentDigest } = api
+  const { createNode, createParentChildLink } = actions
+  const { name, relativePath } = parentNode
+  splitDir[0] = 'doc'
+
+  const pagePath =
```
shcheklein, line 13: let's call it `pageSlug`

src/gatsby/utils/models/index.js
rogermparent: Part of it is me not being familiar with TS, but here specifically TS has to be transpiled before running in `gatsby-node` since the build APIs are run in node and not webpacked. If we were to start using TS in the build process, we'd need add an extra transpilation step for the build code to support it.

src/gatsby/models/blog/index.js
rogermparent: Good catch! seems like neither linters nor I picked this up.

src/gatsby/models/blog/createPages.js
rogermparent: I was definitely thinking env var for a quick fix!
limiting itemCount is so only one blog index page is made. The blog indexes happen to make a lot of their own images with the thumbnails.

src/gatsby/utils/resolvers/index.js
rogermparent: This bit does perform the same functionality for custom nodes as `remark-relative-images` does with frontmatter, but that plugin is still useful for any images in the markdown body content.

shcheklein: frontmatter image URLs are already relative ... so still not sure I understand what's happening here ... can we put some explanation? how was it handled before? can we change .md files to avoid path transformations at least?

src/gatsby/utils/models/index.js
shcheklein: yes, makes sense. No need to transform this. I thought for a moment that I saw some logic (like templates) was moved from ts/tsx to js. Resolving this.

src/gatsby/models/blog/createPages.js
```diff
@@ -47,41 +47,25 @@ function* pagesGenerator({ itemCount, hasHeroItem = false, basePath }) {
 }
 
 const createPages = async ({ graphql, actions }) => {
+  const isDevMode = process.env.NODE_ENV === 'development'
```
shcheklein, line 50: there should be already a check like this around the code, should be a helper function

src/gatsby/models/blog/createPages.js
shcheklein: I just don't quite understand (don't have time to dig) why do we limit number of posts in to two places (here and below) ... intuitively it should be one - at the very top of the processing

src/gatsby/models/blog/index.js
shcheklein: agreed - strange that linters allow both ways

src/gatsby/models/docs/onCreateNode.js
```diff
@@ -0,0 +1,39 @@
+const path = require('path')
+
+function createMarkdownDocsNode(api, options, { parentNode }) {
+  const splitDir = parentNode.relativeDirectory.split('/')
+  if (splitDir[0] !== 'docs') return
+  // Make a special exemption for the root doc.
```
shcheklein, line 6: move this comment below?

src/gatsby/models/markdown-content/index.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/** ContentTransformer
+   (api: Object<GatsbyAPI>, options: Object<GatsbyOptions>, markdownContentApi: Object<MarkdownContentApi>)
+ */
+
+const api = {
+  async onCreateNode(api, options, { models }) {
+    const { node, getNode } = api
+    // Only operate on nodes from the file instance named "content".
```
shcheklein, line 25: move this below

src/gatsby/models/markdown-content/index.js
```diff
@@ -0,0 +1,40 @@
+const buildApi = require('../../utils/models')
+
+/*
```
shcheklein, line 3: joint comments into one blog, try to explain it a little bit better what's happening here

src/gatsby/models/markdown-content/index.js
```diff
@@ -0,0 +1,40 @@
...
+    if (node.internal.type === 'MarkdownRemark') {
+      const parentNode = getNode(node.parent)
+      const { sourceInstanceName } = parentNode
+      if (sourceInstanceName === 'content') {
+        return buildApi(models, { models, parentNode })(
+          'onCreateMarkdownContentNode'
+        )(api, options)
+      }
+    }
+    // Return null in any other situation.
```
shcheklein, line 35: it's more or less obvious :)

src/gatsby/models/blog/onCreateNode.js
```diff
@@ -0,0 +1,52 @@
...
+    author,
+    description,
+    descriptionLong,
+    commentsUrl,
+    picture,
+    pictureComment
+  } = frontmatter
+  const { name, relativePath } = parentNode
+
+  const slug = name === 'index' ? '' : /[-\d]*(.*)/.exec(name)[1]
```
shcheklein, line 21: curious, where does `index` come from here?

src/gatsby/utils/models/index.js
```diff
@@ -0,0 +1,36 @@
...
+async function asyncCallOnAll(models, name, api, modelOptions = { models }) {
+  return Promise.all(
+    models.map(
+      model =>
+        typeof model[name] === 'function' && model[name](api, modelOptions)
+    )
+  )
+}
+
+module.exports = asyncCallOnAll
```
shcheklein, line 36: may be call it `runOnModels` here?

gatsby-node.js
```diff
@@ -1,27 +1,17 @@
...
-}
-
-exports.createPages = async ({ graphql, actions }) => {
-  createDocPages({ graphql, actions })
-  createBlogPages({ graphql, actions })
-}
+exports.createSchemaCustomization = api =>
+  runOnModels(models, 'createSchemaCustomization', api)
+exports.onCreateNode = api =>
+  runOnModels(models, 'onCreateNode', api, { models })
```
shcheklein, line 13: do we need the last `{ models }`?

src/gatsby/models/blog/onCreateMarkdownContentNode.js
```diff
@@ -0,0 +1,52 @@
...
+    author,
+    description,
+    descriptionLong,
+    commentsUrl,
+    picture,
+    pictureComment
+  } = frontmatter
+  const { name, relativePath } = parentNode
+
+  const slug = name === 'index' ? '' : /[-\d]*(.*)/.exec(name)[1]
```
shcheklein, line 21: if we don't need it - let's remove?

src/gatsby/models/docs/onCreateMarkdownContentNode.js
```diff
@@ -0,0 +1,39 @@
...
+  const splitDir = parentNode.relativeDirectory.split('/')
+  if (splitDir[0] !== 'docs') return
+
+  const { node, actions, createNodeId, createContentDigest } = api
+  const { createNode, createParentChildLink } = actions
+  const { name, relativePath } = parentNode
+  splitDir[0] = 'doc'
+
+  // Make a special exemption for the root doc.
+  const slug =
```
shcheklein, line 13: I'm not sure this is correct - slug comes from the sidebard.json for docs, not from the file system

src/gatsby/utils/resolvers/index.js
shcheklein: sounds good, let's get rid of it 👍

shcheklein:
@jorgeorpinel could please review this on your dev environment?

gatsby-node.js
rogermparent: We do need it, because Models that run other Models cannot import the site's Models from the module and run them. I tried switching the dependency injection with a require for aesthetics, but it ends up leading to a circular dependency and breaking the build. It took me a while to figure out what was going on.

src/gatsby/models/blog/onCreateMarkdownContentNode.js
rogermparent: Ah yes, I meant to remove this. I put it down out of habit before I realized no blog entries are named `index`.

gatsby-node.js
rogermparent: We don't, I had it like this before I added the bit that passes it by default. Good catch!

src/gatsby/models/docs/onCreateMarkdownContentNode.js
shcheklein: it is correct, slug is the same as on the file system

src/gatsby/utils/models/index.js
```diff
@@ -0,0 +1,36 @@
...
+      If this isn't provided, it defaults to an object including all the models
+      currently being run as an injected dependency: this allows for sharing
+      Gatsby build code via models that call custom hooks on other models.
+  )
+
+  Standard usage:
+    In a Gatsby API, this mimics the original Gatsby call on all Models:
+
+    exports.onCreateNode = (api) =>
+      asyncCallOnAll(models, "onCreateNode", api)
```
shcheklein, line 21: update the name, also probably update it in the code runOnModels -> callOnModels ?

shcheklein: may be I'm missing some semantics behind the `runOnModels` name though ...

src/gatsby/utils/resolvers/index.js
jorgeorpinel: Assuming this is about the thumbnail generation on dev env?

src/gatsby/utils/models/index.js
rogermparent: You aren't missing anything, this is just me be scatterbrained with a default import that got its name changed a few times. I've got a fix coming in hot.

jorgeorpinel:
> I put a few little flags in page creation that skip most pages while keeping at least one of each kind when NODE_ENV is development. This means gatsby develop only has to remake a bit over 200 images after cache clear instead of 1000+

Yup! This dev build time is much faster than before. Under a minute for me (vs 10+).

It also solves parts of #1084! Specifically, renaming md files in the content dir no longer procude 404s (per https://github.com/iterative/dvc.org/issues/1084#issuecomment-616226306) (not redirects though).

And it also allows for "hidden" docs )with no sidebar entry like we once wanted (see #731).

So pretty positive it seems 👍

shcheklein:
Awesome stuff, @rogermparent !

PR #1150, opened 2020-04-15
S3 Rebuild Caching

rogermparent:
This is another pretty fat commit, and with less of an excuse than the previous one. However, this one trades theoretical messiness for functionality!

The previous solution would upload public to the prefix `${s3Prefix}/public/`, but I believe the website is served from `${s3Prefix}`. This solution fixes that and changes `.cache` to go into the prefix `${s3Prefix}-cache` so it doesn't get picked up in "folder" queries for the root prefix.

Also, the pruning functionality that used to happen in the deploy script by parsing the XML sitemap is now moved to a more "native" solution in Gatsby's `onPostBuild` hook.

I've tested this locally and successfully got rebuilds with no image processing after running `gatsby clean` and then downloaded a previous cache with the deploy script. (It even handled a partial cache after I unthrottled the page count!)


scripts/deploy-with-s3.js
```diff
@@ -22,16 +22,18 @@ const path = require('path')
 const { execSync } = require('child_process')
 const { remove, move, ensureDir } = require('fs-extra')
 const { s3Prefix, s3Bucket, s3Client } = require('./s3-utils')
-const { cleanUpPageData } = require('./page-data-utils')
 
-const rootDir = path.join(__dirname, '..')
-const cacheDirName = '.cache'
 const publicDirName = 'public'
+const publicDirEntry = [publicDirName, '/']
+const cacheDirEntry = ['.cache', '-cache/']
```
rogermparent, line 28: This tuple syntax can be changed to something like objects if we're not all into it.

rogermparent:
I rebased my fork's branch onto master earlier today without thinking of the ramifications on the PR. I think my changes will have to be opened up on a new PR to master.

Issue #1149, opened 2020-04-15
[accessibility] keyboard navigation is broken and some interactive elements are missing focus styles.


nisarhassan12:
According to WCAG and WAI websites and web apps must be fully operable via the keyboard. 

There is a short video that tells why it is important: 

[Web Accessibility Perspectives: Keyboard Compatibility](https://www.youtube.com/watch?v=93UgG72os8M)

Here are a few things that should be done:

- [x] Focus styles must be added to all interactive elements(most of the elements already have them but some of them are missing focus styles). Without the focus keyboard, users can get lost i.e they won't be able to easily know where the current focus is or what interactive element they are on. https://slack.com is a good example of an accessible site. they use an outlined border around links when they receive focus:

![image](https://user-images.githubusercontent.com/46004116/79295379-f58b8100-7ef1-11ea-8c15-333f198f54bc.png)

The elements that are missing focus styles are anchors, and images that are wrapped by anchors, docs navigation, docs side panel and a few buttons.

- [ ] Apart from the focus state styles there must be a tabbing order for keyboard users and dropdowns should reveal their content when they receive focus.

- [x] the hamburger menu is missing keyboard support i.e users should be able to open the menu via tabbing to the hamburger and then pressing enter which will make the menu appear.

- [x] On Desktop when the hamburger menu is hidden visually it should also be inaccessible to the keyboard users as well. Currently, On the Desktop to get to the contents on any page(below the nav) I have to first tab through 2 navigations with the exact same links.

nisarhassan12:
Sometimes what happens is designer don't like the outlines or dotted borders around elements but their value for accessibility can't be ignored.

![image](https://user-images.githubusercontent.com/46004116/79295617-882c2000-7ef2-11ea-992c-37ece4623b5c.png)

There is a way via which we can only make the outlines or specific focus styles appear when the user is navigating the content using a keyboard. For instance, https://gitpod.io is a good example which does that.

nisarhassan12:
I volunteer to make small Prs related to accessibility :slightly_smiling_face:  I'll start off by making a PR for focus styles for Keyboard users only.

nisarhassan12:
It would be nice if someone else can work on it.

julieg18:
> dropdowns should reveal their content when they receive focus.

[...]

PR #1153, opened 2020-04-15
accessibility: add focus state styles to interactive elements

nisarhassan12:
This Pr addresses the following from #1149 

image

This is how it looks with the changes in place.

image

The dotted outline does not look pretty and provides no benefits to users that are using a mouse. So I have made sure that: 
- it only appears when the user is navigating via a keyboard.
- and if in the middle of Navigation let say the user swaps from using a keyboard to a mouse or vice versa for navigation then the handlers will be injected or ejected accordingly.

### DIFF
--- a/src/components/Page/base.css
+++ b/src/components/Page/base.css
@@ -38,14 +38,59 @@ body {
   color: var(--color-black);
 }
 
+/* Focus state styles for interactive elements */
+
+button:focus,
+input:focus,
+select:focus,
+textarea:focus,
+a:focus {
+  outline: 2px dotted var(--color-gray-hover);
+  outline-offset: 5px;
+}
+
+*::-moz-focus-inner {
+  border: 0;
+}
+
+body:not(.user-is-tabbing) button:focus,
+body:not(.user-is-tabbing) input:focus,
+body:not(.user-is-tabbing) select:focus,
+body:not(.user-is-tabbing) textarea:focus,
+body:not(.user-is-tabbing) a:focus {
+  outline: none;
+}
+
+body.user-is-tabbing .btn-with-focus {
+  &:focus {
+    color: #fff;
+    background-color: var(--color-orange);
+    border-color: var(--color-orange);
+    outline: none;
+  }
+
+  &--white:focus {
+    color: var(--color-black);
+    background: #fff;
+    border-color: #fff;
+  }
+}
+
+.link-with-focus {
+  border: 2px dotted transparent;
+}
+
+body.user-is-tabbing .link-with-focus {
+  &:focus {
+    outline: none;
+    border: 2px dotted;
+  }
+}
+
 /* Prevent mobile browsers from changing font-size */
 * {
   -webkit-text-size-adjust: none;
   text-size-adjust: none;
-
-  &:focus {
-    outline: 0;
-  }
 }
 
 /* TODO: check it later */
--- a/config/postcss/mixins.js
+++ b/config/postcss/mixins.js
@@ -2,8 +2,7 @@ const { customMedia } = require('./media')
 
 const focus = {
   '&:focus': {
-    color: 'var(--color-orange)',
-    outline: 'none'
+    color: 'var(--color-orange)'
   }
 }
 
--- a/src/components/Blog/Feed/Item/styles.module.css
+++ b/src/components/Blog/Feed/Item/styles.module.css
@@ -41,7 +41,6 @@
 
   &:hover,
   &:focus {
-    outline: none;
     opacity: 0.7;
   }
 }
--- a/src/components/Blog/Post/Markdown/styles.module.css
+++ b/src/components/Blog/Post/Markdown/styles.module.css
@@ -70,8 +70,8 @@
 
     &:hover,
     &:focus {
-      outline: none;
       opacity: 0.7;
+      color: var(--color-blue);
     }
   }
 
--- a/src/components/Blog/Post/Share/styles.module.css
+++ b/src/components/Blog/Post/Share/styles.module.css
@@ -12,7 +12,6 @@
   }
 
   &:focus {
-    outline: none;
     opacity: 0.7;
   }
 }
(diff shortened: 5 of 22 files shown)


shcheklein:
@nisarhassan12 CI failed, please check when you have time

shcheklein:
@nisarhassan12 can we highlight buttons with some "glow" effect around them? border might look find around links (sometimes), but buttons should have some nice effect. Check the Slack website that you shared - they have some nice effects for buttons.

nisarhassan12:
Thanks! @shcheklein hopefully I will try to fix the CI and adding the glow effect to the buttons  tomorrow.

jorgeorpinel:
Other buttons are missing that indeed
image
image

jorgeorpinel:
Also, and not withing the scope of this PR, but navigating with tab seems to be using hidden elements (probably responsive stuff?) because from [Get Started] to [Download] there's like 15 jumps.

nisarhassan12:
> Also, and not withing the scope of this PR, but navigating with tab seems to be using hidden elements (probably responsive stuff?) because from [Get Started] to [Download] there's like 15 jumps.

Thanks @jorgeorpinel this is already mentioned in #1153 

image

I will make a separate PR for that.

nisarhassan12:
About the CI failure maybe it's just me but I think code climate is maybe too strict here depending on what we are trying to do here. indeed the code blocks are similar but they are doing the opposite of each other. @shcheklein do you have any ideas on how one can go about refactoring this. I can split the 2 lines that are responsible for adding/removing event listeners by splitting them into a separate module exported from a different file that both functions can use but that is too much in my opinion i.e they already function calls with different arguments. What do you say :slightly_smiling_face:?
  

image

shcheklein:
@nisarhassan12 yeah, no need to solve all code climate warning! I was talking about CircleCi.

shcheklein:
It also does not highlight docs sidebar properly.

nisarhassan12:
> let's put all this into `src/utils/accessibility.ts` looks like one file per function is too much

I tried this I was using a similar approach but it would cause Circle CI to fail because of the way the event handlers work splitting them was king of a workaround:

image

Should I do this anyway even if Circle CI fails?

shcheklein:
@nisarhassan12 I see. splitting into two files probably not a good solution - I would explore if there is a better way to do this

src/components/Page/base.css
```diff
@@ -38,14 +38,29 @@ body {
...
 * {
   -webkit-text-size-adjust: none;
   text-size-adjust: none;
-
-  &:focus {
-    outline: 0;
-  }
 }
 
 /* TODO: check it later */
```
shcheklein, line 66: @pavelgrinchenko do you remember what is this? do we have a ticket for this?

nisarhassan12:
> It also does not highlight docs sidebar properly.

Thanks, I also noticed a few other places where it does not highlight properly I will update the Pr soon and also add some other effect to the buttons other then the outline.

nisarhassan12:
> Thanks, I also noticed a few other places where it does not highlight properly I will update the Pr soon and also add some other effect to the buttons other than the outline.
I have made the changes. @shcheklein  please have a look whenever you'll have time.

src/utils/accessibility/handleFirstTab.ts
```diff
@@ -0,0 +1,11 @@
+import handleMouseDownOnce from './handleMouseDownOnce'
```
shcheklein, line 1: this is still not fixed @nisarhassan12 ... I hope, there should be some other way to deal with this than just splitting files

jorgeorpinel:
This particular button doesn't feel focused going from transparent to white backgroud:

image
image

The other ones in the landing page at least look good but the dashed border is gone from all buttons. Should we keep both things?

shcheklein:
I asked to remove dashed board from buttons, it think it looks nice that button just have different glow/highlight - looks like enough visually.

src/components/Page/base.css
pavelgrinchenko: It was in [the old styles](https://github.com/iterative/dvc.org/blob/03c2355df50b10a1bedec2f442da4b97e5d6252c/src/styles.js#L31-L33) and I only transferred it carefully to the new version to not break anything.

shcheklein: @pavelgrinchenko I'm specifically curious about "/* TODO: check it later */" part. It's not clear what to check.

pavelgrinchenko: Ah, got it. I thought about migration this code to `src/components/Documentation/Markdown/styles.module.css`.

shcheklein: @pavelgrinchenko so, is it still relevant? could you create a ticket with some explanation please?

shcheklein: (and remove this TODO from code)

pavelgrinchenko: Created PR for this one https://github.com/iterative/dvc.org/pull/1171

shcheklein: thanks @pavelgrinchenko ! 🙏

src/utils/accessibility/handleFirstTab.ts
shcheklein: @nisarhassan12 ping, no other solutions here? :) @rogermparent hey, do you know if there is a better way to structure this?

nisarhassan12: sorry! @shcheklein I have not found other ways than either splitting the file or disabling eslint for the line about which it was complaining. Having these functions the way they are won't cause a problem in the browser at all i.e because of the time when they are executed
both of them will be defined for sure.

image

shcheklein: @nisarhassan12 I think eslint disable is better if there are no other options for functions, let's go with it.

shcheklein:
thanks @nisarhassan12 ! cool stuff.

PR #1134, opened 2020-04-10
[WIP] Improve rebuild time

pavelgrinchenko:
- [x] Save all necessary folders upon build
- [x] Remove outdated page-data folders

PR #1156, opened 2020-04-16
Rebuilds Caching (Rebased)

rogermparent:
This is a continuation #1150 opened up against master instead of the other rebuilds-caching branch. I wasn't thinking about how rebasing my fork's branch onto master would cause havoc with PRs that already exist.
I could probably rebase what I have now back to the point where #1134 starts, but a new PR onto master seems like an easier option.


jorgeorpinel:
Hello. I noticed that when you rename a md file in contents and update the sodebar.json accordingly, the dev. The sidebar reflects the change but trying to load it yields a 404. I have to stop the dev server, run `gatsby clean`, and `yarn develop` again which takes several minutes to regenerate the thumbnails...

jorgeorpinel:
Oops never mind, posted in the wrong PR.

rogermparent:
Just for anyone else coming into this PR, it was meant mostly to generate the deploy preview. It currently doesn't work with our build system, and I'm trying to get it to do so via other means. I'll probably close this and open another one later when I have something more functional.

PR #1097, opened 2020-03-28
Get rid of CSV/TSV metrics

dmpetrov:
Getting rid of TSV/CSV metrics formats. Part of https://github.com/iterative/dvc/issues/3409#issuecomment-605444643:

> 1. CSV/TSV types are specific to continuous metrics. There is no need to keep them in current, scalar ones. Let's get rid of CSV/TSV/HCSV/HTSV metrics format.

PR #1160, opened 2020-04-16
Regular updates (Apr 16)

jorgeorpinel:
- Address https://github.com/iterative/dvc.org/issues/1123#issuecomment-614929585
- Complete #1097 by removing mentions of `raw` metrics type
- Add --summary option to checkout cmd ref.

### DIFF
--- a/content/docs/command-reference/checkout.md
+++ b/content/docs/command-reference/checkout.md
@@ -6,8 +6,8 @@ DVC-files.
 ## Synopsis
 
 ```usage
-usage: dvc checkout [-h] [-q | -v] [-d] [-R] [-f] [--relink]
-                    [targets [targets ...]]
+usage: dvc checkout [-h] [-q | -v] [--summary] [-d] [-R] [-f]
+                    [--relink] [targets [targets ...]]
 
 positional arguments:
   targets          DVC-files to checkout. Optional. (Finds all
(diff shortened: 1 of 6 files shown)


content/docs/command-reference/metrics/add.md
```diff
@@ -106,7 +104,7 @@ outs:
     path: metrics.json
     cache: true
     metric:
-      type: raw
+      type: json
```
jorgeorpinel, line 107: This change is correct right @dmpetrov? See https://dvc.org/doc/command-reference/metrics/add#examples

I also removed all mentions of `raw` type metrics in this PR.

content/docs/command-reference/checkout.md
```diff
@@ -69,16 +69,18 @@ be pulled from remote storage using `dvc pull`.
 
 ## Options
 
+- `--summary` - show summary of the changes.
```
jorgeorpinel, line 72: I noticed checkout now has this option so I'm adding it. But this description isn't great, where can I find more details? I'm guessing @efiop may know. I.e. summary of what changes? Thanks!

shcheklein: --summary is covered in the Saugat's PR that we have been trying to merge for while

jorgeorpinel: Ah yes I see there's a note about it in https://github.com/iterative/dvc.org/pull/1054#pullrequestreview-391199946. OK.

But in this PR I also changed the order of the options so I say let's keep it here for now and we'll have to solve a little conflict later.

shcheklein:
@jorgeorpinel merging this?

jorgeorpinel:
Yes, thanks.

Issue #1164, opened 2020-04-17
[accessibility] Subscribe section form the input is missing a label.


nisarhassan12:
Info about why it is important can be found here: https://dequeuniversity.com/rules/axe/3.4/label

jorgeorpinel:
I don't understand the title, sorry.

Could you also give some examples or specifically where in dvc.org this is an issue in the description? Thanks

jorgeorpinel:
OK got it from the PR #1165 so never mind.

Feel free to skip opening issues if able to submit a PR directly 🙂

PR #1165, opened 2020-04-17
[accessibility] add label to the subscribe from and hide it visually.

nisarhassan12:
Fixes iterative/dvc.org#1164

### DIFF
--- a/src/components/SubscribeSection/Form/index.tsx
+++ b/src/components/SubscribeSection/Form/index.tsx
@@ -36,6 +36,7 @@ const Form: React.FC = () => {
         id="mce-EMAIL"
         placeholder="email address"
         required={true}
+        aria-label="Enter your email"
       />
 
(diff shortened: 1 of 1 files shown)


shcheklein:
@nisarhassan12 


> If the input is already labeled visually some other way, such as through a magnifying glass icon on a search submit button, it may be acceptable to use `aria-label` to create an invisible text label for screen readers to read.

would it be eiasier?

nisarhassan12:
> @nisarhassan12
> 
> > If the input is already labeled visually some other way, such as through a magnifying glass icon on a search submit button, it may be acceptable to use `aria-label` to create an invisible text label for screen readers to read.
> 
> would it be eiasier? 

@shcheklein Thanks for pointing this out I have updated the Pr accordingly. Indeed this is orders of magnitude easier.

PR #1166, opened 2020-04-18
[docs]: fix the RightPanel to work correctly with browser history.

### LINKED ISSUE #1140: navigation: browser history navigation is broken
Go https://dvc.org/doc/command-reference/checkout

Click one the links from the right side content bar.

Click back in your browser.

The same should work and return to the exact link in the document you got the section.

nisarhassan12:
Fixes #1140

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -1,5 +1,23 @@
 /* eslint-env node */
 
 const PageWrapper = require('./src/components/PageWrapper').default
+const { getDetails, setDetails } = require('./src/utils/shared/localStorage')
 
 exports.wrapPageElement = PageWrapper
+
+exports.onPreRouteUpdate = ({ location, prevLocation }) => {
+  const detailsList = document.querySelectorAll('.Collapsible')
+  let details = getDetails()
+  detailsList.forEach(detail => {
+    const titleElement = detail.childNodes[0]
+    const textElement = detail.childNodes[1]
+    const detailData = { title: '', text: '' }
+    detailData.title = titleElement.textContent
+    detailData.text = textElement.textContent
+
+    if (titleElement.classList[1] == 'is-open') {
+      details = [detailData, ...details]
+    }
+  })
+  setDetails(details)
+}
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -12,6 +12,8 @@ import { getPathWithSource } from '../../../utils/shared/sidebar'
 import 'github-markdown-css/github-markdown.css'
 import sharedStyles from '../styles.module.css'
 import styles from './styles.module.css'
+import { getDetails, setDetails } from '../../../utils/shared/localStorage'
+import { getText } from '../../../utils/front/dom'
 
 const isInsideCodeBlock = (node: Element): boolean => {
   while (node?.parentNode) {
@@ -37,10 +39,24 @@ const Details: React.FC<{
   if (!filteredChildren.length) return null
   if (typeof filteredChildren[0] === 'string') return null
 
-  const text = filteredChildren[0].props.children[0]
+  let details = getDetails()
+  const detailTitle = filteredChildren[0].props.children[0]
+  const detailText = getText(filteredChildren.slice(1))
+  let isOpen = false
+
+  details.forEach(({ text }: { text: string }) => {
+    if (text === detailText) {
+      isOpen = true
+    }
+  })
+
+  details = details.filter(
+    ({ text }: { text: string }) => !(text === detailText)
+  )
+  setDetails(details)
 
   return (
-    <Collapsible trigger={text} transitionTime={200}>
+    <Collapsible trigger={detailTitle} open={isOpen} transitionTime={200}>
       {filteredChildren.slice(1)}
     </Collapsible>
   )
--- a/src/components/Documentation/RightPanel/index.tsx
+++ b/src/components/Documentation/RightPanel/index.tsx
@@ -4,6 +4,7 @@ import throttle from 'lodash/throttle'
 
 import { IHeading } from '../'
 import Link from '../../Link'
+import { Link as GatsbyLink } from 'gatsby'
 import Tutorials from '../TutorialsLinks'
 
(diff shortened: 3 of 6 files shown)


src/components/Documentation/RightPanel/index.tsx
```diff
@@ -98,16 +107,16 @@ const RightPanel: React.FC<IRightPanelProps> = ({
...
         {headings.map(({ slug, text }) => (
-          <Link
+          <GatsbyLink
             className={cn(
               styles.headingLink,
               current === slug && styles.current
             )}
             key={`link-${slug}`}
-            href={`#${slug}`}
+            to={`${window.location.pathname}#${slug}`}
```
jorgeorpinel, line 116: This isn't building:
```
...
       failed Building static HTML for pages - 18.238s
error "window" is not available during server side rendering.
       
         114 |             )}
         115 |             key={`link-${slug}`}
       > 116 |             to={`${window.location.pathname}#${slug}`}
             |                    ^
         117 |           >
         118 |             {text}
         119 |           </GatsbyLink>
       
       
         WebpackError: ReferenceError: window is not defined
         
         - index.tsx:116 
           src/components/Documentation/RightPanel/index.tsx:116:20
         
         - index.tsx:109 RightPanel
           src/components/Documentation/RightPanel/index.tsx:109:19
         
       
       not finished Generating image thumbnails - 719.726s
error Command failed with exit code 1.
       info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
------------------------
Error: Command failed: yarn build
    at checkExecSyncError (child_process.js:611:11)
    at execSync (child_process.js:647:15)
    at run (/tmp/build_654cf463aa916749b29d4c6b1456d24e/scripts/deploy-with-s3.js:31:3)
    at main (/tmp/build_654cf463aa916749b29d4c6b1456d24e/scripts/deploy-with-s3.js:106:5) {
  status: 1,
  signal: null,
...
```
Please test your changes locally. See https://dvc.org/doc/user-guide/contributing/docs

nisarhassan12: Thanks! fixed.

jorgeorpinel:
Tested on https://dvc-landing-right-panel-bsvd42.herokuapp.com/doc/command-reference/checkout and this works, thanks! I'd approve it but it sounds like Ivan has some other feedback.

nisarhassan12:
> Shared feedback in private

Thanks! @shcheklein I have updated the Pr accordingly. Please have a look whenever you'll have time.

shcheklein:
@nisarhassan12 I still do have some strange behavior for the right panel after going back - try to open https://dvc-landing-right-panel-lul3q9.herokuapp.com/doc/command-reference/diff#example-checking-workspace-changes , open a few collapsible sections there, click to some link, go back and use the right content panel.

shcheklein:
Also, what will we be getting in IE 11?

nisarhassan12:
> Also, what will we be getting in IE 11?

@shcheklein According to MDN `window.localStorage` is availabe in IE if the version is 8 or above.
image

I'm sorry I can't test in an actual IE 11 because I use Arch as my OS. 

but let say if a browser is very old and does not support localStorage then state of the collapsed panel won't be persisted although the back button would take you to the exact location.

A bit un-related: the site currently uses CSS custom properties which are not supported at all in all versions of IE and old versions Edge( source: https://caniuse.com/#feat=css-variables ) I would recommend not using them if you want to support IE and other older browsers.

image

shcheklein:
@nisarhassan12 thanks! IE is no a top-top priority - the only thing we want to avoid is a complete crash on it and Sentry complaining.

Re this PR - please take a look here https://github.com/iterative/dvc.org/pull/1166#issuecomment-619396858 .

nisarhassan12:
> @nisarhassan12 thanks! IE is no a top-top priority - the only thing we want to avoid is a complete crash on it and Sentry complaining.
> 
> Re this PR - please take a look here [#1166 (comment)](https://github.com/iterative/dvc.org/pull/1166#issuecomment-619396858) .

Thanks! I'll try to fix it soon.

gatsby-browser.js
```diff
@@ -1,5 +1,23 @@
 /* eslint-env node */
 
 const PageWrapper = require('./src/components/PageWrapper').default
+const { getDetails, setDetails } = require('./src/utils/shared/localStorage')
 
 exports.wrapPageElement = PageWrapper
+
+exports.onPreRouteUpdate = ({ location, prevLocation }) => {
```
shcheklein, line 8: can we make collapsible itself saving this on the expnasion? don't like that we do this global search on every navigate

src/utils/shared/localStorage.js
```diff
@@ -0,0 +1,13 @@
+export const getDetails = () => {
+  let details = []
+  try {
+    details = JSON.parse(localStorage.getItem('details') || '[]')
+  } catch {}
```
shcheklein, line 5: silent `catch` is bad? what kind of errors do we expect here?

src/components/Documentation/Markdown/index.tsx
```diff
@@ -37,10 +39,24 @@ const Details: React.FC<{
   if (!filteredChildren.length) return null
   if (typeof filteredChildren[0] === 'string') return null
 
-  const text = filteredChildren[0].props.children[0]
+  let details = getDetails()
+  const detailTitle = filteredChildren[0].props.children[0]
+  const detailText = getText(filteredChildren.slice(1))
```
shcheklein, line 44: we need a better way to determine the specific collapsible. collecting the whole text feels too much, expensive.

src/utils/shared/localStorage.js
nisarhassan12: an error can occur when localStorage is not supported or it fails to parse what is coming back from localStorage which can happen if someone goes into localStorage and set the value for the key to something like `{`, ']' and so on i.e when there is invalid JSON. 

Since we are returning the empty array no matter what the app won't break.

I think adding a console.log would be helpful.

nisarhassan12:
I'm closing this PR as it does not fixes the issue and would add complexities to the codebase.

PR #1174, opened 2020-04-21
Regular updates (Apr 21)

jorgeorpinel:
- Import `dvc lock` cmd ref about unlocking or updating import stages.
- `dvc list` and `dvc root` cmd refs rewrite
- Other misc. stuff


content/docs/command-reference/get.md
```diff
@@ -36,14 +36,13 @@ the data source. Both HTTP and SSH protocols are supported for online repos
...
-target to be downloaded within the source repository at `url`. `path` can
-specify any file or directory in the source repo, including <abbr>outputs</abbr>
-tracked by DVC, as well as files tracked by Git. Note that for DVC repos, the
-target should be found in one of the
-[DVC-files](/doc/user-guide/dvc-file-format) of the project. The project should
-also have a default [DVC remote](/doc/command-reference/remote), containing the
-actual data.
+The `path` argument is used to specify the location of the target to be
+downloaded within the source repository at `url`. `path` can specify any file or
+directory in the source repo, including <abbr>outputs</abbr> tracked by DVC, as
```
shcheklein, line 41: can we avoid the outputs term here? I would say in general, we should be using outputs only when we deal with pipelines. I would avoid using this term when we apply it to a `dvc added` file

content/docs/command-reference/import.md
```diff
@@ -40,14 +40,13 @@ the data source. Both HTTP and SSH protocols are supported for online repos
...
-target to be downloaded within the source repository at `url`. `path` can
-specify any file or directory in the source repo, including <abbr>outputs</abbr>
-tracked by DVC, as well as files tracked by Git. Note that for DVC repos, the
-target should be found in one of the
-[DVC-files](/doc/user-guide/dvc-file-format) of the project. The project should
-also have a default [DVC remote](/doc/command-reference/remote), containing the
-actual data.
+The `path` argument is used to specify the location of the target to be
+downloaded within the source repository at `url`. `path` can specify any file or
+directory in the source repo, including <abbr>outputs</abbr> tracked by DVC, as
```
shcheklein, line 45: same here, with outputs

content/docs/command-reference/list.md
```diff
@@ -68,10 +70,12 @@ files recursively.
 - `-v`, `--verbose` - displays detailed tracing information. when this option is
   not specified.
 
-## Example: List files and directories in a DVC repository
+## Example: Find files to download from a repository
```
shcheklein, line 73: may be we want import them?

content/docs/command-reference/get.md
jorgeorpinel: OK. Yes, it makes sense. I need to review all the instances of this glossary term then ⏳

content/docs/command-reference/import.md
jorgeorpinel: Will address as part of https://github.com/iterative/dvc.org/pull/1174#pullrequestreview-400440864

content/docs/command-reference/list.md
jorgeorpinel: Yes, dvc import is mentioned in the paragraph below. I don't understand the implication or suggestion in terms of this header change.

content/docs/command-reference/get.md
jorgeorpinel: Is "data artifact" a better term here? Or should that also be avoided? Because in the glossary it's basically a synonym of "outputs" currently and I've been thinking about merging them or getting rid of "data artifact"... Maybe this is my chance!

jorgeorpinel: (Maybe this is ~~my~~ the chance) to redefine it a little so it's useful for non stage/pipeline files/dirs traced by DVC (found in `outs` fields of DVC-files)

shcheklein: data files, directories, models 

`outs` field is a technical impl detail

content/docs/command-reference/list.md
shcheklein: `to download` a bit to narrow in this case ... I can list them for different purposes

content/docs/command-reference/get.md
jorgeorpinel: Hmmm but we are often saying "found insid a DVC-file" so specifying "in the `outs` field" isn't a bad idea, plus the command reference is supposed to have technical details, no?

content/docs/command-reference/list.md
jorgeorpinel: This example uses list and then get, that's why the header says "download". It doesn't mean to describe what list is for, that's done earlier in the description.

shcheklein: No a big deal, @jorgeorpinel 👍 .. up to you :)

content/docs/command-reference/get.md
shcheklein: It's fine when we specifically describe _how_ we do something. I would avoid mentioning this when we describe _what_ we do, the essence of something from the user perspective.

jorgeorpinel: OK well I removed it form here and other related files in 90befbd1 but I still need to review the rest of the references. Resolving here. I'll request your review ASAP.

content/docs/tutorials/deep/preparation.md
```diff
@@ -58,19 +58,18 @@ $ pip install -r code/requirements.txt
...
-hidden from the user. We describe some DVC internals below for a better
-understanding of how it works.
+DVC works best inside Git repositories like the one we're in. Initialize DVC
+with:
 
 ```dvc
 $ dvc init
 ...
 
+At DVC initialization, a new `.dvc/` directory is created for internal
```
shcheklein, line 68: something funny is happening here with formatting?

jorgeorpinel: Oops my bad. Fixed in 26db8de6

Issue #1119, opened 2020-01-03
review meta tags, keywords, favicons, etc


shcheklein:
https://docs.google.com/document/d/1lOhYiIxE02X27mRNXezJCNZEAS2nObjShZ1wnPBpxpY/edit

also check this PR https://github.com/iterative/blog/pull/144 for some fixes

MerlynMShelley:
Hi Team,
I have found these issues in improving the SEO of the site,
![IMG_20200607_164351](https://user-images.githubusercontent.com/46276713/83967387-6f326080-a8de-11ea-946b-0d696124bd29.jpg)
![IMG_20200607_164334](https://user-images.githubusercontent.com/46276713/83967392-735e7e00-a8de-11ea-99de-2621cd6e35d8.jpg)
![IMG_20200607_164311](https://user-images.githubusercontent.com/46276713/83967395-778a9b80-a8de-11ea-88c9-bc8f4c13b956.jpg)

You can find the complete report here,
https://www.seoptimer.com/www.dvc.org

Thank you @shcheklein @ jorgeorpinel

jorgeorpinel:
@MerlynMShelley I collapsed those long vertical images. Hard to read and unnecessary. We can click the link. Please just copy/paste test or small horizontal screenshots of key elements you'd like to bring to our attention.

Meaning: what specific suggestions about meta tags, etc. (this issue's topic) would you make based on that report? Thanks

shcheklein:
@julieg18 One of the existing problems:

https://dvc.org/favicon.ico - is the old low res icon.

According to this https://realfavicongenerator.net/favicon_checker - a lot of formats are not set/available.

Issue #1118, opened 2020-04-06
SEO: review best practices, run analyzers, fix issues


shcheklein:

jorgeorpinel:
From #1785:

> There are a large number of /docs pages on the dvc.org domain... it's a best practice to have unique SEO descriptions/metadata

PR #1183, opened 2020-04-25
SEO Improvements

rogermparent:
This is my continuation of the SEO PR by @fabiosantoscode.

I've merged master, removed a deprecated SFC typedef, and added `aria-label` props to a few buttons and links on the homepage that boost its Lighthouse accessibility score by around 10 points on my local tests.

We can squeeze the last few points of accesibility out by addressing the last flag it throws, which is some elements not having enough contrast between text and background. The first two I found were the mobile "Get Started" button and the gray-on-white GitHub text. 

I can solve this by simply darkening colors and increasing contrast while sticking with the style, but this would easily be the biggest SEO change so I'm going to leave it out of this PR. The stuff here is invisible and pure SEO benefit.


shcheklein:
@rogermparent please, check also #1119 and #1118 . Especially for the meta fields one - it has some points to check.

src/components/Page/DefaultSEO/index.tsx
```diff
@@ -42,11 +48,23 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
...
+      content: metaImage.height
     },
     {
       property: 'og:image:secure_url',
-      content: metaImage
+      content: metaImage.src
+    },
+    {
+      name: 'twitter:site',
+      content: '@dvcORG'
```
shcheklein, line 67: should it be actually just dvc.org URL?

src/components/Page/DefaultSEO/index.tsx
```diff
@@ -76,21 +94,13 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
       meta={meta}
       link={[
         {
-          rel: 'shortcut icon',
+          rel: 'icon',
           type: 'image/vnd.microsoft.icon',
           href: '/favicon.ico'
```
shcheklein, line 99: does it have all necessary dimensions? 

also, I wonder if png with certain sizes are needed in some other cases (besides browsers, check the list in the document in the #1119 )

src/components/Page/DefaultSEO/index.tsx
rogermparent: I didn't add this, but I verified it-
Despite what the field's name implies, this is how `twitter:site` supposed to be used according to Twitter.
https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started

shcheklein: yep, make sense! 👍

src/components/Page/DefaultSEO/index.tsx
rogermparent: I didn't remove these, and I decided to re-add them since their inclusion has minimal impact. They can always be removed again later if the true reason arises.

rogermparent:
Looks like I need to clean up some TS issues that got past the linter before a test deploy will work. I'll get on that.

src/components/Page/DefaultSEO/index.tsx
```diff
@@ -76,21 +95,23 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pageInfo }) => {
...
-          href: '/favicon-32x32.png',
-          sizes: '32x32'
+          href: siteUrl + '/favicon-32x32.png'
         },
         {
           rel: 'icon',
           type: 'image/png',
-          href: '/favicon-16x16.png',
-          sizes: '16x16'
+          href: siteUrl + '/favicon-32x32.png'
```
shcheklein, line 110: 16x16? do we need more sizes? we need to do this if we understand why do we have them.

rogermparent: Oops! Looks like I kinda messed up this manual re-add without a quick `git diff` to lean on.
I just found this:
https://support.google.com/webmasters/answer/9290858?hl=en
Seems like the original change that removed them was made based on this info, which provides guidelines to have one favicon at 48x48 (or a larger multiple of that) and explicitly says not to provide a 16x16 icon.
It also says to use the `rel='shortcut icon` bit removed from before, so I'll re-add that.
I'll adapt the PR to this new info.

shcheklein: yep, that link makes sense, but I would not do SEO in isolation from other needs. meta tags exist not only for Google search page, also link sharing, browsers, embeds, etc, etc - that's why reviewing meta tags ticket ticket is relevant to this one.

rogermparent: You're right, and the article itself says that "penalized" sites weren't downranked, simply not allowed to use their custom 16x16 icon.

This should mean that despite Google ignoring it and downscaling a larger image, having the 16x16 (and any other sizes) shouldn't cause any harm. I'll keep the old entries exactly how they are and re-add `shortcut` to the `.ico` entry as it seems like the most proper one.

Making these favicons absolute is probably premature as well, since favicons work now. I now think these favicons should just be reverted to their original state, which I'll do next commit.

shcheklein: @rogermparent my only point is that this ticket + meta tags one requires a bit of research :) what are the best practices now? what images and for what should we include, etc, etc.

rogermparent: I get what you mean. I generally use Google Lighthouse as the source of truth when if comes to SEO because SEO is more or less for Google. Since Lighthouse doesn't flag having all our original icon sizes as a problem, I think it's best for this PR to keep them to remove later.
Since they were removed in the first commit from @fabiosantoscode, I'll tag him to ask the reason.
I think not removing any is the best for this PR so it can go through easier and fix the share card image issue, which this PR appears to do now from my tests of a new build on it with both the Facebook debugger and another at http://debug.iframely.com for good measure.

[The opengraph spec](https://ogp.me/#metadata) shows that they expect urls to be absolute instead of root-relative, which is where I believe the issue with share card images comes from.

rogermparent:
I think this PR is currently in a good place to be considered for merging (after a review of course) to fix the OpenGraph issue in prod, and any further SEO-focused changes can continue this or open another.

shcheklein:
@rogermparent sounds good! merging. ...

PR #1195, opened 2020-04-27
Regular updates (Apr 27)

jorgeorpinel:
- Update Windows installation instructions (thanks for the reminder, @fabiosantoscode)
- Misc. updates


content/docs/install/index.md
```diff
@@ -16,12 +16,12 @@ if needed. The [Python API](/doc/api-reference) module is `dvc.api`.
 
 ## Advanced options
 
+- You can [install a stable pre-release](/doc/install/pre-release) of DVC to
```
shcheklein, line 19: we should probably update that doc to include an instruction to install pip pre-release (command is in the last blog post)

jorgeorpinel: Done in a3a70895 I think

content/docs/install/pre-release.md
```diff
@@ -10,16 +10,9 @@ releases, you can install it from our code repository GitHub.
...
-$ pip install git+https://github.com/iterative/dvc
+$ pip install --upgrade --pre dvc
 ```
 
-> `gitpython` allows the installation process to generate a DVC version using
-> the current Git commit hash. This lets us to distinguish official DVC releases
-> (e.g. `0.64.3`) from a development version (e.g. `0.64.3-9c7381`). For more
-> information on our versioning convention, refer to
-> [Components of DVC version](/doc/command-reference/version#components-of-dvc-version).
-
```
jorgeorpinel, line 22: Is this change correct? Not sure who to check with exactly so cc @efiop, @skshetry, and @dmpetrov 

I tried it just like this with and without gitpython and dvc version printed `DVC version: 1.0.0a0`.

Thanks

shcheklein: No, this is not exactly correct. Since we now have dev and actual pre-release we need to describe how to install both. Or keep only the way it was before since I'm not sure we are going to do that many official pre-releases.

jorgeorpinel: Reverted in eee573fc. I think no need to describe both since it seems the pip option will be rarely used and we can always announce it on the blog.

content/docs/command-reference/remote/add.md
```diff
@@ -159,16 +159,17 @@ For more information about the variables DVC supports, please visit
 ### Click for Microsoft Azure Blob Storage
 
 ```dvc
-$ dvc remote add myremote azure://my-container-name/path
-$ dvc remote modify --local myremote connection_string "my-connection-string"
+$ dvc remote add --local myremote azure://my-container-name/path
+$ dvc remote modify --local myremote connection_string \
+                            "my-connection-string"
 ```
```
jorgeorpinel, line 165: Did this per #1258 but should we start the example with the env var method, which doesn't need --local instead?

shcheklein: it's fine I think, let's keep it this way for now

PR #1200, opened 2020-04-29
Remove reference to core.loglevel in docs

### LINKED ISSUE #1199: core.loglevel doesn't exist but remains in docs
When setting the config option for `loglevel` (either programmatically or by editing the config file) my command returns the following error:
```bash
ERROR: configuration error - config file error: extra keys not allowed @ data['core']['loglevel']
```

As far as I can tell, core.loglevel was removed in https://github.com/iterative/dvc/pull/3321. However the current docs for `dvc config` continue to list this as a valid configuration option.

woodshop:
Fixes #1199. `core.loglevel` was removed from the available config options in https://github.com/iterative/dvc/pull/3321. This PR updates the documentation to remove any references to the deprecated configuration setting.

### DIFF
--- a/content/docs/command-reference/config.md
+++ b/content/docs/command-reference/config.md
@@ -62,9 +62,6 @@ file (in `.dvc/config` by default), and they support the options below:
 
 This is the main section with the general config options:
 
-- `core.loglevel` - log level that the `dvc` command should use. Accepts values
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Thanks @woodshop !

PR #1203, opened 2020-05-01
Explicitly add node_modules to cacheDirectories for Heroku

rogermparent:
Adding node_modules to cacheDirectories seems to download a more complete cache than the default behavior, despite the node buildpack docs saying node_modules are cached by default.

While out of the scope of a git PR, another part of reducing rebuild time on Heroku is setting the environment variable YARN_PRODUCTION to false so Heroku doesn't waste over a minute pruning devDependencies.
I tried just moving the Gatsby packages from devDependencies to dependencies, but it takes nearly as long to prune a minimal set of devDependencies as it does with all of them.
Even without pruning devDependencies we're way below the slug max, and I don't see it growing a few hundred MB any time soon. This will be even less of a problem once images are removed from the source repo, though if the images do pose a problem we should be able to remove them like public and .cache after the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,5 +162,8 @@
     "*.css": [
       "yarn stylelint --fix"
     ]
-  }
+  },
+  "cacheDirectories": [
+    "node_modules"
+  ]
 }


shcheklein:
so, are we good to merge this? or we run more experiments still?

have you enabled YARN_PRODUCTION on prod?

rogermparent:
Ah, I'm sorry! We're able to merge this. I didn't think I had permissions to change the environment variables on prod, but I'll try that now.

rogermparent:
Just added the variable- we're good to merge!

shcheklein:
@rogermparent done!

PR #1205, opened 2020-05-01
Fix Windows development

### LINKED ISSUE #1204: blog: can't develop/build locally on Win
I'm on Windows (Git Bash on Cmder). I cleaned up mode modules, buid, cache, etc. to build from scratch. I ran `yarn` from scratch, which went well. I ran `yarn develop` and it built in a few minutes, no errors. But I get this when I go to http://localhost:8000/blog:

image

(The rest of the site seems to work.)

Then I tried `yarn build` but it throws an error:

image

rogermparent:
Ever since Models dropped, we use `path.join` to resolve links between files at build time (e.g. Blog Posts and Authors). I believe this is what's breaking Windows development, so I used a small module called `upath` that wraps the path API with cross-platform safe versions of functions like `join`.

`upath` is already installed by another module as you can see in `yarn.lock`, but the other package specifies an earlier version. The minimal size difference at build time and completely unchanged bundle size make it not worth pinning our usage to the same version, in my opinion.

Fixes #1204, or at least I think it does. It will before merging, at least.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "scroll": "^3.0.1",
     "serve-handler": "^6.1.2",
     "slick-carousel": "^1.8.1",
-    "unist-util-visit": "2.0.2"
+    "unist-util-visit": "2.0.2",
+    "upath": "^1.2.0"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
--- a/src/gatsby/utils/resolvers/index.js
+++ b/src/gatsby/utils/resolvers/index.js
@@ -1,4 +1,4 @@
-const path = require('path')
+const path = require('upath')
 
 /** Source Path Resolver Builder
    This function makes Gatsby resolvers that connect a node to another node via
--- a/yarn.lock
+++ b/yarn.lock
@@ -16982,7 +16982,7 @@ unset-value@^1.0.0:
     has-value "^0.3.1"
     isobject "^3.0.0"
 
-upath@^1.1.1:
+upath@^1.1.1, upath@^1.2.0:
   version "1.2.0"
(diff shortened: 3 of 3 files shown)


shcheklein:
@jorgeorpinel ping?

rogermparent:
This PR's commits are included in #1210, but if we happen to merge this before that lands the histories should still be compatible. If not, I can pull 1210 and do the rebase.

package.json
```diff
@@ -78,7 +78,8 @@
     "scroll": "^3.0.1",
     "serve-handler": "^6.1.2",
     "slick-carousel": "^1.8.1",
-    "unist-util-visit": "2.0.2"
+    "unist-util-visit": "2.0.2",
+    "upath": "^1.2.0"
```
shcheklein, line 82: not sure ... but should it be part of the devDeps section?

jorgeorpinel:
No need to rebase Roger. I properly merged your branch to mine so Git will know what needs to be merged where/when.

package.json
rogermparent: I suppose it would fit more in the current setup, but technically this is the right place for it and most of those devDependencies should be moved up to regular dependencies. It makes no difference in the practical sense as the switch to `YARN_PRODUCTION=false` means we keep all deps and devDeps in the same way on deploy.

shcheklein: yep, you are right. For the website there are no difference at all, right? 

may be we can keep linters, tests, and other stuff there? may be should move them to optionalDeps? to avoid installing them on Heroku?

shcheklein: anyway, merging this ... since this one can stay in deps.

PR #1207, opened 2020-05-01
Add a script in package.json enabling "yarn run dev"

### LINKED ISSUE #1084: dev mode improvements

rogermparent:
Fulfilling a simple request from #1084, up to the team whether to accept it or not.
I think this also finally fixes #1084, as Models seems to have enabled the sidebar to hot reload.

The last bit about clearing the cache comes from an understandable misconception, I believe. Clearing Gatsby's cache all the time would mean we need to spend ~10m transforming images each build- any HMR issues within Gatsby should be able to be solved without blowing away cache all the time when properly implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "develop": "gatsby develop",
+    "dev": "gatsby develop",
     "build": "gatsby build",
     "start": "./scripts/clear-cloudflare-cache.js; node ./src/server/index.js",
     "heroku-postbuild": "./scripts/deploy-with-s3.js",


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "develop": "gatsby develop",
+    "dev": "gatsby develop",
```
shcheklein, line 8: should we keep only one of those?

also, need to update contributing guide probably (in content/docs/user-guide...)

rogermparent: Removing the other one would throw off devs used to using `yarn run develop` for no real benefit, but it doesn't matter much to me so I could go either way.

Good point about the contributing guide, but if we keep the prior invocation it'd be fine as is with this just being some free syntactic sugar for devs who like to minimize their keystrokes.

What the docs are changed to, if at all, depends on what syntax we keep; because of this, I'll wait for some feedback.
I'd personally recommend to keep both commands and add a small aside in the docs mentioning that this is possible: this implementation lets people who want the feature know it's there, it won't upset anyone using the prior behavior, and it confers no performance drawback.

shcheklein: Sounds good, @rogermparent . Let's keep both.

jorgeorpinel: I'd go for keeping only dev and updating the contrib guide. But I see this is merged.

PR #1208, opened 2020-05-01
Make Link component recognize mailto links

### LINKED ISSUE #1201: support: email link is broken
in https://dvc.org/support

image

It renders as `href="https://dvc.org/mailto:support@dvc.org"` (that's a 404 of course) but the code is correct so this is engine-related.

Let's check any that no other special links have a similar problem too?

rogermparent:
Basically, Gatsby Link is only for regular links within the site. We currently have a wrapper that chooses whether to use it or a raw `a` tag depending on if the link is within the site or not. This change adds a step to that check that filters out links that start with `mailto:`, making them work properly wherever this wrapper component is used, like the support page.

Fixes #1201

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -15,13 +15,14 @@ export type ILinkProps = {
 
 const PROTOCOL_REGEXP = /^https?:\/\//
 const isRelative = (url: string): boolean => !PROTOCOL_REGEXP.test(url)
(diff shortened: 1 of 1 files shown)


rogermparent:
[It's working on preview deploy's support page!](https://dvc-landing-fix-mailto-njpmbyn.herokuapp.com/support)

PR #1209, opened 2020-05-02
Fix and improve cache pruning

rogermparent:
I had some testing behavior in the cache pruner that logged file removals without
actually doing them using `&&` (`console.log` returns falsy). This would be fine if 
I had marked it in a TODO or any comment whatsoever.

Anyway, this commit fixes that.

I also changed the logic to use regular `fs` and `Promise.new` instead of
`fs.promises` to avoid the experimental warning. `upath` is used because
I believe that without it cache pruning won't work on Windows.


rogermparent:
After a few simulated runs with git checkout, I can say this fix will solve the phantom doc page issue.

PR #1226, opened 2020-05-04
simplify code contributions + reviews by fully automating the dev setup.

nisarhassan12:
Hi :slightly_smiling_face:

This Pr simplifies code contributions by fully automating the dev setup with Gitpod(a full-featured online cloud IDE) which is free for OSS. With a single click, it'll launch a workspace and automatically:

- clone the dvc.org repo.
- install the dependencies.
- run `yarn build`.
- start the webserver via `yarn start`.

so that anyone interested in contributing can start straight away without any friction.

this is how it looks:

image

You can give it a try on my fork of the repo via the following link:

https://gitpod.io/#https://github.com/nisarhassan12/dvc.org

A lot of popular OSS projects out there are already using Gitpod to streamline the experience for their contributors and maintainers, some of them are: 

- https://github.com/freeCodeCamp/freeCodeCamp
-  https://github.com/facebook/docusaurus
-  https://github.com/mozilla/pdf.js/

You can find a brief list of them at https://contribute.dev

image

Prs can also be reviewed from within Gitpod.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,7 @@
+tasks:
+  - init: yarn install && yarn build
+    command: yarn develop
+
+ports:
+  - port: 8000
+    onOpen: open-preview
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -53,9 +53,9 @@ We will review your PR as soon as possible. Thank you for contributing!
 
 ## Development environment
 
-We highly recommend running this web app locally to check documentation or blog
-changes before submitting them, and it's quite necessary when making changes to
-the website engine itself. Source code and content files need to be properly
+We highly recommend running this web app to check documentation or blog changes
+before submitting them, and it's quite necessary when making changes to the
+website engine itself. Source code and content files need to be properly
 formatted and linted as well, which is also ensured by the full setup below.
 
 Make sure you have a recent LTS version of [Node.js](https://nodejs.org/en/)
@@ -81,6 +81,13 @@ This will start the server on the default port, `8000`. Visit
 `http://localhost:8000/` and navigate to the page in question. This will also
 enable the Git pre-commit hook that will be formatting and linting your code and
 documentation files automatically.
+
+Alternatively you can contribute to dvc.org using Gitpod(a fully featured online
+development environment), with a single click it will launch a workspace and
+automatically: clone the repo, install the dependencies, run `yarn build` and
+start the webserver so that you can start straight away.
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/from-referrer/)
 
 ### All commands
 
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Maintainability](https://api.codeclimate.com/v1/badges/5872e0a572ec8b74bd8d/maintainability)](https://codeclimate.com/github/iterative/dvc.org/maintainability)
 [![CircleCI](https://circleci.com/gh/iterative/dvc.org.svg?style=svg)](https://circleci.com/gh/iterative/dvc.org)
+[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/iterative/dvc.org)
 
 [DVC](https://github.com/iterative/dvc) project website's source code.
 [Documentation](https://dvc.org/doc) and [blog](https://dvc.org/blog) content.


shcheklein:
@nisarhassan12 looks cool!

@casperdcl link check fails again, looks like a false positive.

casperdcl:
much sneaky such wow. This pattern:

```
[](https://gitpod.io/#https://github.com/iterative/dvc.org)
```

is pretty darn hard to fix. I'd recommend just adding it to the excludes rather than coming up with a solution. Currently the above line results in:

```bash
https://github.com/iterative/dvc.org)
https://gitpod.io/#https://github.com/iterative/dvc.org
```

so `https://github.com/iterative/dvc.org)` would have to be added to `scripts/exclude-links.txt`

shcheklein:
why does it result in https://github.com/iterative/dvc.org) ... don't we ignore `)`?

casperdcl:
nope because parentheses are valid chars in URLs. There are many in the docs, e.g. https://en.wikipedia.org/wiki/Hyperparameter_(machine_learning)

casperdcl:
I'll suggest a minor fix in a PR but would be best to double-check...

shcheklein:
@nisarhassan12 could please the latest master? to fix CI check

.gitpod.yml
```diff
@@ -0,0 +1,7 @@
+tasks:
+  - init: yarn install && yarn build
+    command: yarn develop
+
+ports:
+  - port: 8000
+    onOpen: open-preview
```
casperdcl, line 7: does this file have to be in the repo root? Couldn't find anything in https://www.gitpod.io/docs/config-gitpod-file/ about location

nisarhassan12: Yeah, it has to be in the repo root. Sorry! that It's not documented.

content/docs/user-guide/contributing/docs.md
```diff
@@ -51,6 +51,14 @@ Otherwise, please refer to the following procedure:
 
 We will review your PR as soon as possible. Thank you for contributing!
 
+## Automated development environment in a single click
```
shcheklein, line 54: let's make it a paragraph in the Dev environment. Like a second paragraph for example should be a good fit. May remove "locally" from the first paragraph.

.gitpod.yml
shcheklein: 👍

nisarhassan12:
> @nisarhassan12 could you please rebase against or merge the latest master? to fix CI check

Thanks! @shcheklein I have rebased it againist the latest master.

shcheklein:
@nisarhassan12 awesome, one more ask - please check the comment. I think we can make it part of the dev env section in the docs.

content/docs/user-guide/contributing/docs.md
nisarhassan12: Thanks! I have made the change.

shcheklein:
done, merged! and I moved a bit in the master ... thanks 🙏 Hopefully we'll be getting even more contributors now. Will you make a PR to the list of sites, btw?

nisarhassan12:
@shcheklein  sure! please open an issue here https://github.com/gitpod-io/contribute.dev/ :slightly_smiling_face:

jorgeorpinel:
Sorry for the delay. This is really cool (:

The button looks a little strange in https://dvc.org/doc/user-guide/contributing/docs#development-environment though: 

image

☝️ I can address this in a future PR, but it's not working for me:

image

Can you maybe check what the problem is, @nisarhassan12? It works from https://contribute.dev/

shcheklein:
Yep, does not work for me either (worked yesterday).

Also, we need to adjust CSS a bit to remove the "external link" icon. @jorgeorpinel mind creating a ticket for this?

nisarhassan12:
> Can you maybe check what the problem is, @nisarhassan12? It works from https://contribute.dev/

Sure. I will shortly make a Pr with the fix.

jorgeorpinel:
Thanks again Nisar!

jorgeorpinel:
Hmmm I tried using Gitpod to format the note I mentioned above. It was all working nicely, I even installed a VS Code plugin to the IDE. But after committing with Git, I'm unable to push because I didn't grant Gitpod permission to push directly to dvc.org

image

Open source contributors will never have this permission. Can it be made so a fork is created in their accounts first, and then Gitpod launched from there? Or maybe if I was using a different user, that would've been the behavior... Not sure

jorgeorpinel:
> Or maybe if I was using a different user, that would've been the behavior... Not sure

Nope. I tried with @AyokVoluntarios whici I also own and it tries to use the upstream directly:

image

So even if I tell Github to let Gitpod push to my repos, I simply can't push to iterative/dvc.org from that account.

image

The user would need to configure the Git remote and branch manually from the IDE. Do you know whether that's the only solution, @nisarhassan12? Thanks

shcheklein:
@jorgeorpinel good catch, let's create a separate ticket and ping @nisarhassan12 there?

jorgeorpinel:
Sure, done: #1253

PR #1236, opened 2020-05-05
Update user-guide/contributing/blog

rogermparent:
The frontmatter format for blog posts and authors is quite different now, and I left these quite outdated.

### DIFF
--- a/content/docs/user-guide/contributing/blog.md
+++ b/content/docs/user-guide/contributing/blog.md
@@ -29,9 +29,9 @@ descriptionLong: |
   Some long
   multiline
   text (supports _basic_ [Markdown](link))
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/contributing/blog.md
```diff
@@ -29,9 +29,9 @@ descriptionLong: |
   Some long
   multiline
   text (supports _basic_ [Markdown](link))
-picture: /uploads/image.jpeg
```
shcheklein, line 32: let's put something similar to what we have in the current posts - the are organized by date?

rogermparent: Great point!

PR #1245, opened 2020-05-05
Fix images in the RSS feed

### LINKED ISSUE #1177: rss: no title, no image in the feed for new posts
Screen Shot 2020-04-22 at 1 05 11 PM

rogermparent:
It seems many Gatsby sites don't really have images within their RSS feeds' content, and this turned out to be slightly more involved than I thought it would.

Basically, this branch changes the way the feed content is parsed. Instead of taking pre-rendered HTML from GraphQL, it uses the [Rehype](https://github.com/rehypejs/rehype) htmlAst that `gatsby-transformer-remark` already makes for us and transforms links and images into a simpler form of HTML that works with RSS.

To get images working, I had to do two things to the feed's AST before rendering to HTML:
1. Change root-relative `src` attributes (`/example`) to absolute links (`https://dvc.org/example`)
2. "Unwrap" responsive image wrappers by replacing them with the `img` tag fallback within.

Fixes the images part of #1177, and since titles are already fixed it fixes #1177 entirely.


shcheklein:
Looks too complicated and affects the website (optimization for smaller screen sizes, etc?). Also, two notes:

1. I see that other images are rendered in RSS even in the current version.
2. We should be applying these changes to the RSS render, w/o affecting and complicating even further the website renderer.

shcheklein:
3. If we can't do these two points ^^ - let's just ignore RSS issue for now - it's not the biggest priority.

rogermparent:
This change _is_ only on the RSS renderer.

shcheklein:
Oh, sorry, @rogermparent - function is already big enough that I missed the context.

How is it possible though that other images do work in RSS w/o applying this transformation?

can it be somehow related to some custom image processing plugins/logic that we already have?

rogermparent:
> Oh, sorry, @rogermparent - function is already big enough that I missed the context.
> 
> How is it possible though that other images do work in RSS w/o applying this transformation?
> 
> can it be somehow related to some custom image processing plugins/logic that we already have?

I'm not sure which images you're talking about that do work, as none of the images in the body work for me. Embeds work, but those are just iframes. Any images that aren't wrapped with a `gatsby-image` component, like external ones, also should work currently although I haven't seen them.

The problem is twofold- the HTML from the remark plugin has `gatsby-image` responsive images, which are actually a bunch of different elements wrapped together, embedded in its HTML. This works fantastically for real web browsers, but for more simple embedded browsers like RSS readers the trick actually breaks things.

The other half is root-relative links, which are a browser-specific feature that falls apart outside of that context. Once you have some way to reliably alter HTML properties (rehype, in this case, as it's the highly compatible sister project to remark), it's incredibly easy to make all root-relative links into absolute by simply prepending the site URL to them, much like how `gatsby-plugin-feed` uses links outside of the content and requires the same `site_url` variable to do so.

shcheklein:
> I'm not sure which images you're talking about that do work

hmm : 🤔 This is the most recent post in feedly for me - 

Screen Shot 2020-05-05 at 3 28 13 PM

rogermparent:
I just looked at that image, and it's broken for me on a linux-based local reader called FeedReader.
That image in particular is an SVG, which can't be processed by Sharp and doesn't become a Gatsby Image like raster formats.

This means those particular SVG images bypass the Gatsby-image problem but still suffer from the root-relative problem. The latter must be an issue that only certain readers suffer from.

gatsby-config.js
```diff
@@ -131,22 +135,54 @@ const plugins = [
...
+           */
+            const processor = rehype()
+              .use(urls, url =>
+                url.href.startsWith('/')
+                  ? site.siteMetadata.site_url + url.href
+                  : undefined
+              )
+              .use(() => tree => {
+                // Unwrap all gatsby-resp-image-wrapper elements into plain images.
+                selectAll('.gatsby-resp-image-wrapper', tree).forEach(node => {
```
shcheklein, line 166: should probably do something like:

```
const {
  imageClass,
  imageWrapperClass
} = require('gatsby-remark-images/constants')
```

similar to our `plugins/resize-image-plugin/index.js`

gatsby-config.js
```diff
@@ -131,22 +135,54 @@ const plugins = [
...
+                  const fallbackImg = select('img', node)
+                  delete node.children
+                  Object.assign(node, fallbackImg)
+                  node.properties = {
+                    ...node.properties,
+                    style: 'max-width: 100%'
```
shcheklein, line 172: should/can we respect our "resize" syntax from the "resize" plugin?

yarn.lock
```diff
@@ -8510,6 +8510,22 @@ hast-util-to-html@^4.0.1:
...
+    hast-util-is-element "^1.0.0"
+    hast-util-whitespace "^1.0.0"
+    html-void-elements "^1.0.0"
+    property-information "^5.2.0"
+    space-separated-tokens "^1.0.0"
+    stringify-entities "^2.0.0"
+    unist-util-is "^3.0.0"
+    xtend "^4.0.1"
+
 hast-util-to-html@^7.0.0:
```
shcheklein, line 8529: why do we have both versions now?

gatsby-config.js
```diff
@@ -131,22 +135,54 @@ const plugins = [
...
                     title
                     date
                     description
                   }
                 }
               }
-            `,
+        `,
           serialize: ({ query: { site, allBlogPost } }) => {
+            /* This processor works on Rehype ASTs, doing two things:
```
shcheklein, line 148: still feels we should extract it into a separate plugin/function?

gatsby-config.js
```diff
@@ -131,22 +135,54 @@ const plugins = [
...
+                    ...node.properties,
+                    style: 'max-width: 100%'
+                  }
+                })
+                return tree
+              })
+              .freeze()
+
             return allBlogPost.nodes.map(node => {
+              const html = processor.stringify(processor.runSync(node.htmlAst))
```
shcheklein, line 180: check the `const { convertHtmlToHast, convertHastToHtml } = require('../utils/convertHast')` ?

rogermparent: That looks like a good place to extract this function. My parser uses different plugins so I can't just use that one as it is nor change it to mine, but it can sit next to the other rehype plugins.

gatsby-config.js
rogermparent: It took me a little bit to get what you meant- good idea though! I'll do this next push.

gatsby-config.js
rogermparent: I started using the convertHastToHtml for the latter half of the feed processor, after the AST is transformed.

yarn.lock
rogermparent: `rehype-urls` depended on an older version of `rehype` and its ecosystem. I've replaced it with a version I wrote, and now no packages are changed.

gatsby-config.js
```diff
@@ -170,7 +172,7 @@ const plugins = [
               }
             }
           }
-        `
+    `
     },
     resolve: `gatsby-plugin-feed`
```
shcheklein, line 177: not related directly to this PR ask: let's unify the order - plugin name -> options if possible

rogermparent: I'm totally for that, just didn't want to go out of scope without a reason. This shouldn't take long.

plugins/utils/makeFeedHtml.js
```diff
@@ -0,0 +1,53 @@
+const { imageWrapperClass } = require('gatsby-remark-images/constants')
+const unified = require('unified')
+const { convertHastToHtml } = require('./convertHast.js')
+const { select, selectAll } = require('hast-util-select')
+
+const rootToAbsolute = siteUrl => tree => {
+  selectAll('a', tree).forEach(node => {
```
shcheklein, line 7: let's put a link here https://github.com/gatsbyjs/gatsby/issues/14133 to remove this when it's fixed upstream?

plugins/utils/makeFeedHtml.js
```diff
@@ -0,0 +1,53 @@
...
+          ...fallbackImg.properties,
+          style: 'max-width: 100%; margin: auto;'
+        }
+      }
+    ]
+  })
+  return tree
+}
+
+function makeFeedHtml(htmlAst, siteUrl) {
```
shcheklein, line 38: really annoying that gastby plugin feed does not handle this ... kinda feels like all of this is a requirement for that plugin to be useful

plugins/utils/makeFeedHtml.js
```diff
@@ -0,0 +1,53 @@
...
+  selectAll(`.${imageWrapperClass}`, tree).forEach(node => {
+    // Set the fallback image as the wrapper's only child, and then
+    // give that image the wrapper's original style.
+    const fallbackImg = select('img', node)
+    node.children = [
+      {
+        ...fallbackImg,
+        properties: {
+          ...fallbackImg.properties,
+          style: 'max-width: 100%; margin: auto;'
```
shcheklein, line 30: qq-  does it makes/do we respect our custom image sizing logic?

rogermparent: This just changes the wrapper's internal image to have `max-width: 100%` so larger images don't extend offscreen on narrower displays and `margin: auto` so smaller ones are centered. 
The original `max-width` on the wrapper that the existing plugin operates on is untouched, as well as any other properties on the wrapper.

plugins/utils/makeFeedHtml.js
rogermparent: Sounds good to me!

plugins/utils/makeFeedHtml.js
rogermparent: I agree to a certain extent, especially for our use case, but RSS feeds are very often used for something like podcasts where the encoded content is more of a summary than the whole blog (ads are a lot harder in an RSS reader, after all), and the encoded content is kept simple.

This isn't to say the default and suggested queries of `gatsby-plugin-feed` are good, I think naively passing through the content's raw `html` string is kind of poor way to go about things, but it's not as if the plugin is useless without this kind of content reprocessing.

PR #1257, opened 2020-05-07
Change "docs/home" to "docs"

### LINKED ISSUE #1073: create a dedicated Home page for docs at dvc.org/doc (design part left)

rogermparent:
Address #1073 partially
With some minor changes that allow a blank string slug in sidebar.json, this can be done without making the docs homepage a special exemption.
The existing redirect from /docs to /docs/home also had to be reversed.

### DIFF
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -19,7 +19,7 @@
   "^/((?:deb|rpm)/.+)                                                                     https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                                                                     https://discordapp.com/invite/dvwXA2N 303",
   "^/(?:docs|documentation)(/.*)?$                                                        /doc$1",
-  "^/doc/?$                                                                               /doc/home 307",
+  "^/doc/home/?$                                                                          /doc",
   "^/doc/get-started(/.*)?$                                                               /doc/tutorials/get-started$1",
   "^/doc/tutorial/?$                                                                      /doc/tutorials",
   "^/doc/tutorial/(.*)?                                                                   /doc/tutorials/deep/$1",
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -19,7 +19,7 @@
   "^/((?:deb|rpm)/.+)                                                                     https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                                                                     https://discordapp.com/invite/dvwXA2N 303",
   "^/(?:docs|documentation)(/.*)?$                                                        /doc$1",
-  "^/doc/home/?$                                                                          /doc",
+  "^/doc/home/?$                                                                          /doc 307",
   "^/doc/get-started(/.*)?$                                                               /doc/tutorials/get-started$1",
   "^/doc/tutorial/?$                                                                      /doc/tutorials",
   "^/doc/tutorial/(.*)?                                                                   /doc/tutorials/deep/$1",
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -18,13 +18,13 @@
 
   "^/((?:deb|rpm)/.+)                                                                     https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                                                                     https://discordapp.com/invite/dvwXA2N 303",
-  "^/(?:docs|documentation)(/.*)?$                                                        /doc$1",
+  "^/(?:documentation)/?$                                                                 /doc",
+  "^/(?:documentation)/(.*)?/?$                                                           /doc/$1",
   "^/doc/home/?$                                                                          /doc 307",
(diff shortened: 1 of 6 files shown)


src/gatsby/models/docs/onCreateMarkdownContentNode.js
```diff
@@ -10,10 +10,7 @@ function createMarkdownDocsNode(api, { parentNode }) {
   splitDir[0] = 'doc'
 
   // Make a special exemption for the root doc.
```
shcheklein, line 12: is it still a special "exemption"? or a general rule now?

shcheklein:
@rogermparent please also update description a little bit - it only partially fixes the ticket.

shcheklein:
and one minor thing I noticed - nav link (and it other places) ends with `/` for DOC

redirects-list.json
```diff
@@ -19,7 +19,7 @@
   "^/((?:deb|rpm)/.+)                                                                     https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                                                                     https://discordapp.com/invite/dvwXA2N 303",
   "^/(?:docs|documentation)(/.*)?$                                                        /doc$1",
-  "^/doc/?$                                                                               /doc/home 307",
+  "^/doc/home/?$                                                                          /doc 307",
```
jorgeorpinel, line 22: Do we need this redirect, since the previous one was 307? Hopefully it's not cached in browsers/ Not sure

shcheklein: Only if someone put a link somewhere in the wild. I would probably not worry too much and remove it.

redirects-list.json
```diff
@@ -18,13 +18,13 @@
 
   "^/((?:deb|rpm)/.+)                                                                     https://s3-us-east-2.amazonaws.com/dvc-s3-repo/$1 303",
   "^/(?:help|chat)/?$                                                                     https://discordapp.com/invite/dvwXA2N 303",
-  "^/(?:docs|documentation)(/.*)?$                                                        /doc$1",
-  "^/doc/?$                                                                               /doc/home 307",
+  "^/(?:documentation)/?$                                                                 /doc",
+  "^/(?:documentation)/(.*)?/?$                                                           /doc/$1",
+  "^/doc/home/?$                                                                          /doc 307",
```
jorgeorpinel, line 23:
```suggestion
```

jorgeorpinel: per https://github.com/iterative/dvc.org/pull/1257#pullrequestreview-407639243

src/gatsby/models/docs/onCreateMarkdownContentNode.js
rogermparent: That's just a comment I forgot, good catch.

rogermparent:
> check the code base for instances of `doc/home`. content\blog\2020-04-30-gsod-ideas-2020.md and src\gatsby\models\docs\onCreateMarkdownContentNode.js have it, at least.

Part of this branch was removing `doc/home` from the docs model node creator, and it turns out after a ripgrep that that blog post was the only other instance! Thanks for pointing that out.

redirects-list.json
jorgeorpinel: I still see this one in the code?

shcheklein:
Looks good to me, all concerns are fixed as far as I can tell.

jorgeorpinel:
Why does the link on the top menu end with slash `/` ? I mean, where in the code is that? I don't see it in the source code changes.

jorgeorpinel:
It's weird because it only happens whe you load the site. However if you load any area outside of docs, after clicking on docs the slash goes away once you're in the docs. Not really a problem... but funny

shcheklein:
@rogermparent yes, this is not good (mean prefetch does not work and most likely we hit a redirect). I actually checked this when this branch was deployed, so not sure wha't going on here. Could you please create a ticket to investigate and fix this?

rogermparent:
Got it, I'm not sure what's up but I'll get started investigating it.

rogermparent:
Looks like some leftover Sidebar link behavior was the culprit (even the top nav bar still sources from it). I've got a quick fix at #1265 now, but it's worth further decoupling the sidebar from all the links outside of itself in the future.

PR #1258, opened 2020-05-07
Add missing --local option in Azure instructions

nilushancosta:
When `dvc remote modify --local` is executed after `dvc remote add`, it shows the error `ERROR: configuration error - config file error: remote 'myremote' doesn't exists.`

This seems to be caused because the given `add` command writes to `.dvc/config` while the given `modify` command attempts to modify the remote in `.dvc/config.local`. 
This PR adds the missing `--local` option to the `add` command


jorgeorpinel:
See also https://github.com/iterative/dvc.org/pull/1195#pullrequestreview-407597477

PR #1260, opened 2020-05-07
Change how relative fragments are handled.

### LINKED ISSUE #1140: navigation: browser history navigation is broken

rogermparent:
Fixes #1140 as I understand it, but somebody please do check for possible regressions.

I can successfully do the following behavior:

> Click one the links from the right side content bar.
> 
> Click back in your browser.
> 
> The same should work and return to the exact link in the document you got the section.

I can't think of any behavior this change would break, as both the previous and current behavior only apply to fragment links within the same page (with the `href`'s first character being `#`), and other fragment links are left alone.

One thing to check, though, is the previous behavior of combining existing `rel` props with `noopener noreferrer` on all external links. The behavior I replaced it with only does so if the link is external and doesn't already specify a rel, as I figure covering basic external links is important but if the writer is going out of their way to add a rel then they should assume full control.  
If we depend on the prior behavior and can't do without it, I can revert this branch to have that functionality while keeping the fix.

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -20,13 +20,29 @@ const isMailto = (url: string): boolean => url.startsWith('mailto:')
 const ResultLinkComponent: React.FC<ILinkProps> = ({
   href,
   children,
+  rel,
   ...restProps
 }) => {
-  if (!isRelative(href) || isMailto(href) || restProps.target) {
-    let rel = 'noopener noreferrer'
+  // Handle all situations where a basic `a` must be used over Gatsby Link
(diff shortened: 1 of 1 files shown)


rogermparent:
Restyled had some sort of auth error that doesn't look related to the PR itself:
https://restyled.io/gh/iterative/repos/dvc.org/jobs/270328/log

shcheklein:
👍 cool stuff @rogermparent .

PR #1261, opened 2020-05-07
Change tag page paths to be under /blog

### LINKED ISSUE #1222: blog: tags URL should be under blog prefix
https://dvc.org/tags/heartbeat -> https://dvc.org/blog/tags/heartbeat

- [ ] Don't forget to update existing links
- [ ] Check RSS (do we have it/need for tags?)

rogermparent:
Also change existing links to tag pages to use this new format

Fixes #1222, disregarding the question about RSS because we don't currently use tags in the RSS feed. We could add them somehow, but that's another issue.

### DIFF
--- a/content/blog/2020-04-06-april-20-dvc-heartbeat.md
+++ b/content/blog/2020-04-06-april-20-dvc-heartbeat.md
@@ -30,8 +30,8 @@ tags:
 ---
 
 Welcome to the April Heartbeat, our
-[monthly roundup of cool happenings](https://dvc.org/tags/heartbeat), good reads
-and other bright spots in our community.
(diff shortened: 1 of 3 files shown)


rogermparent:
```
yarn run v1.22.4
$ ./scripts/link-check-git-diff.sh
~/repo ~/repo
content/blog/2020-04-06-april-20-dvc-heartbeat.md:/dev/fd/63:
 ERROR:404:https://dvc.org/blog/tags/heartbeat
ERROR:1 failures
---
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Exited with code exit status 1
```
Looks like the CI test is checking the link in the content which points to the prod site, so it fails on this branch because it's looking for the new tag page that won't exist until this hits prod.

PR #1265, opened 2020-05-08
Fix sidebar support for empty links (/doc/ trailing slash issue)

### LINKED ISSUE #1264: Problems with trailing slashes on the new docs path
For some reason, at some point after #1257, trailing slashes appear in the `hrefs` of links pointing to the new homepage at "/doc", but inconsistently as the effect is removed upon navigation.

This issue is to investigate and fix that.

rogermparent:
There was a hard-coded `/doc/` prepended to every path before, but now we start
with `/doc` and add a second slash only if something comes after.

This is a quick fix, but this problem stems from the `sidebar.json` setup still having influence in many places of the site. Ideally it should be _just_ be the canonical ToC and have nothing to do with link generation, but that's a problem for another time.

fixes #1264

### DIFF
--- a/src/utils/shared/sidebar.js
+++ b/src/utils/shared/sidebar.js
@@ -154,7 +154,7 @@ function getFirstPage() {
 
 function getItemByPath(path) {
   const normalizedPath = path.replace(/\/$/, '')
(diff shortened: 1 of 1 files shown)


shcheklein:
thanks @rogermparent 🙏

Issue #1045, opened 2020-03-13
remove `react-markdown` from the Tooltip component


iAdramelk:
Right now Tooltip uses `react-markdown` to render descriptions from `content/docs/glossary.js` to HTML. It was ok then we have to use `react-markdown` to render docs pages, but now we should simplify it.

We can either convert md to html on the server side with `gatsby`, or just replace it with plain html, because right now we only have couple of links and code samples there.

- [ ] Address #1137 after this one.

shcheklein:
We should probably do `We can either convert md to html on the server side with gatsby` . But it feels it won't save us much. Would be great first to do a comparison on how much JS and load time we can shave off with this effort.

rogermparent:
I can do this one in a way similar to the way I did Event expiration: loading the data from `glossary.js` into a GraphQL and processing it at build time will allow us to freely parse the MD into HTML during build time with any method- probably Remark, since we use it in other features similar to this like `descriptionLong`.

PR #1268, opened 2020-05-08
Add templates to doc pages and give docs home one

rogermparent:
This branch is now at the point where docs have configurable templates and the docs home has a unique template at `src/templates/doc-home.tsx` as opposed to regular docs which use `src/templates/doc.tsx` (before, both used `doc-home`).

Now that the scaffolding is there, we just have to actually make the page.

- Default docs page template moved from `src/templates/doc-home.tsx` to
  `src/templates/doc.tsx`. A new homepage-specific (but currently identical)
  template has been added in its place.

- Added frontmatter to docs home source to define the template

- Docs now have a "template" field, which when defined changes the component
  used to make its page. This is a file relative to `src/templates` with the
  extension left off. This defaults to `doc`, or the file
  `src/templates/doc.tsx`. This is not currently on blog posts.


src/templates/doc.tsx
```diff
@@ -0,0 +1,48 @@
...
+  data: {
+    page: {
+      htmlAst: Node
+    }
+  }
+  pageContext: {
+    slug: string
+    headings: []
+  }
+}
```
rogermparent, line 20: Obviously there's some massive code duplication right now which codeclimate picks up on, but it's likely different templates' data fetching will require different interfaces and queries.

I would recommend we only worry about this kind of refactoring somewhere between the page being finished and merging to master.

shcheklein: 👍 let's not worry about CC at this stage, agreed

shcheklein:
@rogermparent what is the plan moving forward? I'm not sure how will we connect MD with this custom template. What kind of customization it will allow us?

rogermparent:
The path forward from here is very open and subjective, as we have full freedom on how to connect the docs home's metadata with the new copied component.
There are a few implementations we could go with, depending on editors' preference.

Given a simple "blocks and some text content" layout, there's three broad directions we can take: 

1. Define the homepage "blocks" in frontmatter, rendering them programmatically in the component.

Pros:  
- All content can be edited in the frontmatter page.
- Once finished, the template doesn't have to change for content.

Cons:
- Frontmatter-heavy source file looks a little clunky.

2. Lean on the template component for everything that isn't the body text.

Pros:
- Very simple implementation
Cons:
- Page content is split between Markdown and Template

3. Use custom `rehype-react` components for blocks (we already use `rehype-react` so there's no worries about adding it). Since this branch has a unique home template we can choose to add the custom components for either just this page or all pages.

Pros:
- All content in one clean, understandable source file
- Most flexible
- Components are just React, no extra rehype/mark plugins required!
- Useful for unique pages in the future
Cons:
- Writing React in MD can feel a little weird.

I suppose it's a little obvious that I'm most for using the `rehype-react` approach to make the block menu out of custom components. Frontmatter-based is viable if editors prefer that way of making content. Leaning on the template and splitting the content defeats the purpose of using MD in the first place, so it's my least favorite option.

shcheklein:
@rogermparent if we go all components, will we need the md file at all? will it have empty body?

Also, if we go all components - will we still (on a component level) use Markdown component to render block of texts?

Probably, would be useful at this point to see an example?

rogermparent:
> @rogermparent if we go all components, will we need the md file at all? will it have empty body?
> Also, if we go all components - will we still (on a component level) use Markdown component to render block of texts?

Components is one of the only options that keeps content in the body. The text content with this method is written in the markdown file's body as mostly markdown while components are used to wrap special constructs like the home menu blocks.

> Probably, would be useful at this point to see an example?

Absolutely! Making the example is my next top priority right now, I've just been taking it light on the weekend. I'll do one with the component syntax first.

src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,62 @@ const Abbr: React.FC<{ children: [string] }> = ({ children }) => {
...
+    </div>
+  )
+}
+
 const renderAst = new rehypeReact({
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   createElement: React.createElement as any,
   Fragment: React.Fragment,
-  components: { details: Details, abbr: Abbr, a: Link }
+  components: {
```
shcheklein, line 102: transformation happens build time, right? I wonder if we can "simplify" the external-links plugin this way?

rogermparent: Sort of, but not quite. the HAST is precompiled JSON, and this just map components to json types when rehydrating.
If you're talking about the one that puts a little box next to external links, I do believe we can do that by replacing the `a` here with a `Link` component that has that behavior, and it wouldn't incur any runtime cost beyond normal React.

src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,62 @@ const Abbr: React.FC<{ children: [string] }> = ({ children }) => {
   return <Tooltip text={children[0]} />
 }
 
+const Cards: React.FC<{
+  children: Array<object | string>
```
shcheklein, line 54: out of curiosity - why is it different from the above type declaration? Why is `| string`, btw?

src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,62 @@ const Abbr: React.FC<{ children: [string] }> = ({ children }) => {
...
+}> = ({ children }) => {
+  return (
+    <div className={styles.cards}>
+      {Array.isArray(children)
+        ? children.reduce(
+            (acc: Array<object>, child, i) =>
+              typeof child === 'object'
```
shcheklein, line 61: what else can it be ? do we ignore it?

src/components/Documentation/Markdown/index.tsx
shcheklein: nno, I mean `plugins/external-link` the one that generates embed in blog. Try to grep `external-link` elements in the blog content.

content/docs/index.md
```diff
@@ -1,3 +1,7 @@
+---
+template: 'doc-home'
```
shcheklein, line 2: it looks like we made `cards` general enough, do we need a separate template now?

rogermparent: You're right, I expected to need to do something in another component but everything went fine with custom component. 
I'll remove this template frontmatter and the existing home template. We may need a unique template in the future, but we'll add it when that time comes.

src/components/Documentation/Markdown/index.tsx
shcheklein: okay, I see .. `renderAst` happens client side ... to be honest I don't particularly like this approach, I think `Details` and `Abbr` exist here just because @pavelgrinchenko and @iAdramelk (they can chime in) didn't have enough time to do this properly (e.g. similar to the external-link-plugin) when they were migrating this to Gatsby. Ideally we should have something similar to what we have in the blog where we don't `renderAst` on the client side and just insert html "as-is".

To be honest, I'm not sure how much it would make experience faster and reduce the bundle size - would love to see some experiments. Also, would love to hear from @iAdramelk on this :)

src/components/Documentation/Markdown/index.tsx
rogermparent: Newlines in markdown manifest as `"\n"` strings in `children`, so `string` needs to be considered as a possible input with custom components that consume Markdown.
`children` is a non-array in component with <= 1 children, so that case has to be considered for when manipulating `children` with array functions and TS complains unless you're very explicit about things.

Unlike `card`, there isn't really a use-case for a `cards` component with one element, so I omitted non-array input from the type signature to keep `ts-lint` from complaining that I always treat it as an array.

src/components/Documentation/Markdown/index.tsx
rogermparent: This filters out the "\n" strings that occur in remark custom component children, as well as anything else that isn't a `card`, as `cards` components should only contain `card` components.

src/components/Documentation/Markdown/index.tsx
rogermparent: My description may have been a bit confusing- the Rehype approach is still turned into HTML by Gatsby like any other page and there isn't a noticeable performance hit. When I say "not build time", I mean these components don't transform or fetch external data like remark/rehype plugins can so it's harder to do things like images (hence the MD image style used for icon boxes)

Compared to `rehype-react`, MDX does move that last little bit of client-side rehydration into the build step, as it completely turns the MD+components into ES5 at the build step.

shcheklein: Yeah, you are right that Gatsby renders it statically. And may be I'm wrong, but the question here is on what is happening on navigation/pre-fetch - what is being returned to the client and how is it processed. I remember that in case of blog we have html and use dangerousHtml. In this case does it return JSON (or some "raw" html) that is then being parsed on the client? It means that client needs to pull some libs to do that?

(again, don't have enough time to dig too deep into this - so, it might be I'm completely wrong here)

rogermparent: No need to worry, you have a legitimate concern!

Using `rehype-react` at all, which still applies even before the homepage with custom components, does lead to parsing JSON into React for hydration and when running in JS-space. 
However, all the HAST is still transformed, parsed, and ready to be interpreted straight from GraphQL. The `page-data` cache will hold JSON HAST object instead of a raw HTML string.

Even if somehow React doesn't parse the `dangerouslySetInnerHTML` string into components at all and simply inserts it into the DOM, translating fully preprocessed HAST to React is so fast that both methods can be considered equally performant in practical terms.

If I were to make an educated hypothesis, using `html` may be some milliseconds faster to parse and `htmlAst` likely has a bit of a smaller data payload, but the degree is to the point where these are micro-optimizations. I'll try running lighthouse tests on both to see how they actually stack up.

shcheklein: Yeah, I'm not worried even about additional cycles we spend on processing, more about JS bundle size. Again, w/o looking into specific data points it's hard to tell exactly, but I remember we decided to keep it all server side for blog (it was built with Gatsby from the very beginning).

At this point, I think it's fine to keep it this way, but eventually I would try to get rid of all libs on the client side (e.g. we have to send Markdown because of tooltips as far as I remember).

iAdramelk: @shcheklein @rogermparent  Sorry for the long answer. Basically we originally wanted to render all markdown to html on the backend. This would save us some bundle size and calculation time because we don't need to bundle `rehype` and it's plugins to the bundle to process it on the client.

The plan didn't work through. `Abbr` and `Details` components above are not just static html, they require a lot of custom JS logic to work for positioning, interactive stuff, etc. So to make them work we had to use `rehype-react` to replace standard tags with React components. I personally didn't like it very much, but without replacing this components with something a lot simpler, the is no way to make it work on backend.

Card and Cards on the other hand can be implemented using small remark plugin like [external-link-plugin](https://github.com/iterative/dvc.org/blob/master/plugins/external-link-plugin/index.js). It will not reduce bundle size that much, but it will be consistent with the other parts of the app and still a little smaller.

rogermparent: @iAdramelk Thanks for the detailed answer! That sheds some light on a few aspects of the project for me that I was unsure of the purpose of.

My personal style and opinion leads me to believe using a Remark plugin for MD custom components when we're already loading and importing `rehype-react` is making a simple feature more complex and fragile for very minimal benefit. I can understand why we do it in the blog's `external-link` since that means the blog pages don't have to bundle `rehype-react`.

I also posted a comment in the related issue- I think the best way forward given the short timeframe until DVC 1.0 drops is to shelve making this page with custom components for a bit and just make an alternate React template for the docs home to present exactly what we want to interested users while requiring the least amount of accommodating changes in our data pipeline.

Once that's handled, we have less of a time crunch to hash out how exactly we want special pages like this to be handled in the future.

shcheklein: Resolving this. We have a ticket https://github.com/iterative/dvc.org/issues/1045 for this - Roger please update it to generalize it - that it's not about tooltips only.

content/docs/index.md
shcheklein: hey, any update to this? are we removing the template?

rogermparent: I totally can, and quite quickly, but considering we don't know the exact direction this page is going I haven't committed a change to it yet.

Given my comment on the home issue, I think the best course of action may be to actually go all-in on the home template and make it a standalone component. This allows us to not be limited in design and get the docs home we want up quickly.

My current plan is that once I get the design I'll make the template that's a React component implementing that design, and once that makes the Docs page match what we want, we merge to master- this first step of the plan lets us get a great docs home out before the important DVC 1.0 events around the corner because the implementation is already here and React components are the easiest to port to any other method down the line.

After we have a page we want made in a simple React component template, we can later quite easily port the page content to something more dev/editor friendly that lets editors change the page in MD.

src/components/Documentation/Markdown/index.tsx
shcheklein: may be it is safer to raise an exception?

shcheklein: at least if it something except `\n`

shcheklein: `\s*`

src/components/Documentation/Markdown/index.tsx
shcheklein: 👍

src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,62 @@ const Abbr: React.FC<{ children: [string] }> = ({ children }) => {
   return <Tooltip text={children[0]} />
 }
 
+const Cards: React.FC<{
+  children: Array<object | string>
+}> = ({ children }) => {
+  return (
+    <div className={styles.cards}>
+      {Array.isArray(children)
```
shcheklein, line 58: I'm still not sure about this - if we have this check, we should have a different type above to my mind ... it is strange that `ts-lint` complains - any idea why is it happening?

content/docs/index.md
shcheklein: @rogermparent I would be happy if we can get something simple and clean like this - https://cube.dev/docs . I don't think we need something too complicated/custom at this point. Does it help to make a decision? 

Also, we can take design ideas there (except icons of course :) ) . I'll ask designer after that to make us icons.

shcheklein:
Let's do the following as a next step (to merge it):

- [ ] Remove icons for now
- [ ] Make card clickable. It means one action (topic) per card cc @jorgeorpinel 
- [ ] Card should have a title

After that I'll be able to show this to a designer and get some actual design.

rogermparent:
> Let's do the following as a next step (to merge it):

I'm on it!

>     * [ ]  Remove icons for now

Easy enough

>     * [ ]  Make card clickable. It means one action (topic) per card cc @jorgeorpinel

Also easy, but we'll have to decide which one takes over for the cards the currently have two.

>     * [ ]  Card should have a title

Similar to the topics, the implementation is easy once we decide on what the title are.

I'll get as far as I can with my own decisions as an example and adapt it when other feedback comes in.

rogermparent:
Done with the design changes, and the components are more sanely written as well.

shcheklein:
@jorgeorpinel let's keep Get started instead on Install and Python API reference instead of Use Cases (use cases are not ready to be honest to put them here)?

shcheklein:
@rogermparent : (sorry for pixel movements again) - let's for now do the same effect as we have for buttons to the right? It might be enough and will look consistent (I'll ask for icons after we decide on the content with @jorgeorpinel )

rogermparent:
@shcheklein I changed the Card hover animation to only `background-color` transitioning from white to `var(--color-light-blue)`, which is the same as what the buttons to the right do. Unlike the small buttons, the color eases over 0.3s.

The cube.dev example has their cards start dark and brighten, while ours start bright and darken. We could invert the color, but I feel like there's too much contrast between the current two colors to do that- maybe we could use a shade of gray only barely below white, maybe with a blue hue like the buttons' hover color.

shcheklein:
Done! Last step - ping designer to make it look better. Thanks @rogermparent !

content/docs/index.md
```diff
@@ -4,7 +4,33 @@ Welcome! In here you may find all the guiding material and technical documents
...
+
+<card href="/doc/tutorials/get-started" heading="Get Started">
+
+A step-by-step introduction into basic DVC features
+
+</card>
+
+<card href="/doc/user-guide" heading="User Guide">
+
+Study the detailed inner-workings of DVC in its user guide
```
jorgeorpinel, line 17: Super minimal: Missing period. I'll add it no worries. Same in cmd ref card text

PR #1273, opened 2020-05-08
Add external doc sidebar links and replace "Changelog" with one.

### LINKED ISSUE #1272: support external links on the sidebar (quick fix)
This issue is specifically to allow for external links on the current sidebar setup. 
There are plans to renovate the sidebar code entirely, but it's worth fixing this issue in the current set-up because making a real sidebar overhaul may take a little while.

rogermparent:
Fixes #1272 

Keep in mind that this is meant as a quick fix, and the current sidebar system
is likely to be overhauled in the future in a way that will have this
functionality, but not necessarily invoked with the same "type" syntax.

This branch allows the sidebar to have external links, and changes the
"Changelog" link at the end to be a link to the DVC GitHub.

Both the sidebar utils and component had to be changed to enable this- the
sidebar utils moreso as the "external" type bypasses previous logic entirely.
Anything without "external" set on it is treated the same, however.

The component change just treats external-typed entries a little differently by
setting `target=_blank` and preventing the `href` from having anything prepended
to it. The best solution is to move this out of the component level entirely,
but the best solution isn't the quickest and I've already mentioned big sidebar
changes are coming soon.


rogermparent:
I'm sure we're going to want the "Changelog" entry in right now to be labeled and/or pointing to another link. I just have what's there as an example right now.

rogermparent:
I had to wrestle with the sidebar logic some more, but I think it's ready to go unless we want to change how it looks.

This PR now has:

1. Working external links on the sidebar with little external icons  
(We can probably put a little more work in to change the color if that's important).

2. Changelog links changed to DVC's GitHub releases everywhere (sidebar and community page)

3. Old changelog articles removed.

content/docs/sidebar.json
```diff
@@ -440,17 +440,8 @@
...
-        "slug": "0.18"
-      },
-      {
-        "label": "v0.19 - v0.35",
-        "slug": "0.35"
-      }
-    ]
+    "label": "Changelog",
+    "slug": "https://github.com/iterative/dvc/releases",
+    "type": "external"
```
jorgeorpinel, line 445: Can we change `slug` for `link` or `href`?

jorgeorpinel: Or `url` like in data.json file below.

rogermparent: Done! It's `url` now.

rogermparent:
> Can the community page link open in a new window like the links in the User Content section? Thanks

Done! I actually added functionality to our Link component that makes external links default to `target="_blank"` when not given a `target`.

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -8,6 +8,7 @@ import includes from 'lodash/includes'
 import ShowOnly from '../../../ShowOnly'
 import DownloadButton from '../../../DownloadButton'
 import Link from '../../../Link'
+import { ReactComponent as ExternalLinkIcon } from '../../../../../static/img/external-link-minimal.svg'
```
shcheklein, line 11: does it make sense to just put it nearby? like we do with css and I think other assets already

rogermparent: It looks to me like nearly all image assets, especially icons like this, live in `/static`.  
But where the new icon is only used here right now I see no problem in co-locating it with the component. It can always be moved up if we decide to use it elsewhere.

shcheklein: 👍 thanks!

There is an example already - src/components/Blog/Post/Share/icons

It makes sense to me, to be honest - easier to understand and keep track, and path is simpler, etc. I think, moving forward we should try to move all the images close to MD or components. We even have a ticket for this I believe. What is your take on this?

PR #1279, opened 2020-05-11
Fix mobile doc sidebar

### LINKED ISSUE #808: doc sidebar menu on mobile, don't close it on navigate
~On main page there is no top menu, so no way to go into docs unless you type it into address field~

Additionally, to switch to a submenu of other than currently open menu you need to open menu twice:
- open menu
- hit top level item
- it closes
- open menu
- hit next level item

iKBAHT:
Fixes #808

There is some conflict in the mobile doc sidebar. Some root sidebar items have their own articles. When the user clicks to that item two things happen:

1. article opens
2. submenu in the sidebar collapsed

On mobile this cause problem because both logic can't be visible at the same time. I think a better way to solve the issue is to separate 1 and 2 logic to different clicks (ie need special UI element to collapse submenu). To do it need some other design for a mobile sidebar.

As a quick fix, I implement the following logic: the mobile doc sidebar doesn't close on navigating until leaf item is selected.

### DIFF
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -24,7 +24,7 @@ interface ISidebarMenuItemProps {
   label: string
   path: string
   source: boolean | string
-  onClick: (e: React.MouseEvent) => void
+  onClick: (isLeafItemClicked: boolean) => void
   activePaths?: Array<string>
   type?: string
 }
@@ -40,6 +40,8 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   const isActive = activePaths && includes(activePaths, path)
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
+  const isLeafItem = children === undefined || children.length === 0
+  const currentLevelOnClick = (): void => onClick(isLeafItem)
 
   const className = cn(
     styles.sectionLink,
@@ -54,7 +56,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
         href={path}
         id={path}
         className={className}
-        onClick={onClick}
+        onClick={currentLevelOnClick}
         target="_blank"
       >
         {label} <ExternalLinkIcon />
@@ -64,7 +66,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
         href={getPathWithSource(path)}
         id={path}
         className={className}
-        onClick={onClick}
+        onClick={currentLevelOnClick}
       >
         {label}
       </Link>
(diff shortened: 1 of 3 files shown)


iKBAHT:
`yarn format-check` failed, but it also failed on `master` 59a570c3757086a6862792d4395dd63be07d2dcd

src/components/Documentation/Layout/SidebarMenu/styles.module.css
```diff
@@ -60,10 +60,6 @@
 
   &:hover {
     color: #3c3937;
-
-    @media (--xs-scr) {
```
shcheklein, line 64: @iKBAHT why do we remove this?

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -40,6 +40,8 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   const isActive = activePaths && includes(activePaths, path)
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
+  const isLeafItem = children === undefined || children.length === 0
```
shcheklein, line 43: (not sure we need to change or fix this in this PR) curious - is it undefined or zero length array after all? this check feels a bit excessive.

rogermparent:
I'll run this as well, but simply looking at the code doesn't make any issues jump out to me. I'll try to address @shcheklein's comments as well.

src/components/Documentation/Layout/SidebarMenu/index.tsx
rogermparent: This isn't currently necessary as no sidebar items return empty arrays (and probably won't as long as it's defined as JSON). On top of that, if the issue ever arises we can control for that kind of thing at build time.
Removing this extra check is a micro-optimization, but a valid one nonetheless.

src/components/Documentation/Layout/SidebarMenu/styles.module.css
iKBAHT: This fixed bug.

Steps:
- Open docs in mobile
- Open sidebar
- Click to "Install"

Exp: "Install" is highlighted
Act: "Install" remains as it was

shcheklein: Not an easy to test in the preview environment for me (your change - this one removal) to see the bug you mention. But this one removal introduces another one for me - when I move the sidebar with my finger it highlights (and keep it highlighted) any item under my finger. So, you kinda see two items highlighted at once.

iKBAHT: Sorry, it is my fail. I fix this and fix the bug that I found.

iKBAHT:
Rebased from the master

PR #1281, opened 2020-05-11
"Expiration Date" feature for Community Hero and Events

### LINKED ISSUE #1277: community page: events expiration date, do not show irrelevant events
Introduce expiration date for the main banner and per event for the events section.

If event/banner no relevant anymore - don't show it.

If no events to show - do not show the whole events section, show some (nicely done) stub - "No upcoming events"

If no banner - do not show banner.

rogermparent:
The hero and events in the community page's `data.json` now accept an 
optional `expires` field, which stops the item in question from rendering when 
the site is built past that date.

This is done through a new shared expiration helper and its child function
`isExpired`, which uses the `moment` library that Gatsby already depends on to
determine if an arbitrary content object is expired.

Any object with a `date` and/or `expires` key that can be parsed by `moment` is
usable in `isExpired`. Objects without an `expires` will be treated as if they
have an `expires` 7 days after their `date`; this is configurable in a constant.

Also, the expiration helper is in ES5 so it can be used on both during the build
and at runtime.

Made for and fixes #1277
Currently includes #1282's files, but not its commits.


rogermparent:
The basic functionality is working, but I need to fix lint issues and make sure edge cases are handled correctly. Others can also check it out to give their opinion, but know it's currently not quite final.

src/components/Community/Events/index.tsx
```diff
@@ -8,6 +8,7 @@ import Link from '../../Link'
 import Block from '../Block'
 import Section from '../Section'
 import { logEvent } from '../../../utils/front/ga'
+import { isExpired } from '../../../utils/shared/expiration.js'
```
shcheklein, line 11: how much does it affect the bundle size ? (just thinking if it makes sense to filter them built time)

rogermparent: I would *love* to do this at build time! There's certainly a cost to doing this in the component, it's just the quickest route. `isExpired` works at build time so I can pivot to using this there. It'll take some more code but be worth it in the long run IMO.

rogermparent:
Just pushed some big changes, moving Community data into GraphQL so we can transform it at build time. I still need to fix linting, but local builds work.

I added in the header from #1282 to show off the "multiple expiring heroes with a fallback" feature. I only wanted the files so I used `checkout` to get them without the commits. I don't know if I'll have to reconcile that if that PR gets merged first (and it likely will), but I'm sure it won't be too difficult.

shcheklein:
hey, Roger - just curious - is it expected :): 

Screen Shot 2020-05-12 at 4 31 26 PM

content/community.json
```diff
@@ -1,9 +1,18 @@
 {
-  "hero": {
-    "pictureDesktop": "/img/community/banner.png",
-    "pictureMobile": "/img/community/banner-mobile.png",
-    "url": "https://www.mlprague.com/#schedule-saturday"
-  },
+  "hero": [
+    {
+      "pictureDesktop": "/img/community/banner.png",
```
shcheklein, line 4: we'll need to reconcile this as well ... I think @andronovhopf removed them on my request :( so, might want to get them back or remove this entry after testing

rogermparent:
> hey, Roger - just curious - is it expected :):

Yeah, that massive line change is mostly `yarn.lock`. I ran a regular Yarn upgrade earlier in this branch because the latest version of Gatsby uses the latest version on Moment, and upgrading lets us use the same copy of Moment that Gatsby does.
Since it wasn't a `--latest` upgrade, `package.json` is mostly untouched compared to the lockfile.

content/community.json
rogermparent: Thanks for pointing it out! I think the reconciliation won't be too bad.
Having the old expired image was just for testing, I intended to remove it before merge anyway!

src/utils/front/community.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+  `)
+  return useMemo(
+    (): ICommunityData => ({
+      rest: data.rest.content,
+      hero: data.hero.nodes[0],
+      // Return null if there are no items
+      // Otherwise, pad arrays shorter than 3 items with null
+      // TODO change Events component so it doesn't require this
```
shcheklein, line 72: please, don't put TODOs into comments - create a ticket with a description, and put a link here to it

scripts/deploy-with-s3.js
```diff
@@ -61,8 +60,6 @@ const cacheDirs = [
   ['.cache', '-cache/']
 ]
 
-const fs = require('fs')
```
shcheklein, line 64: curious - how did you find that is not being used? linter?

rogermparent: our `yarn ts-lint` complains about this once all actual errors are removed- at least for me.

scripts/s3-utils.js
```diff
@@ -3,7 +3,7 @@
 const path = require('path')
 const s3 = require('s3-client')
 const { s3Prefix, s3Bucket } = require('../src/server/config')
-const { remove, move, ensureDir } = require('fs-extra')
```
shcheklein, line 6: same here - why don't we detect this automatically?

src/utils/front/community.ts
rogermparent: Fair enough! I'll take this out.

content/community.json
```diff
@@ -52,7 +55,7 @@
       "pictureUrl": "/img/community/ugc/medium_logo.png"
     },
     {
-      "url": "https://blog.codecentric.de/en/2020/01/remote-training-gitlab-ci-dvc/",
+      "url": "https://mribeirodantas.xyz/blog/index.php/2020/03/05/r-dvc-and-rmarkdown/",
```
shcheklein, line 58: 👍

scripts/s3-utils.js
rogermparent: I don't know, I thought we did and that's why these are still here. It's only after I started using the TS linter more that I noticed.

scripts/deploy-with-s3.js
shcheklein: why didn't we detect it on CircleCI then? 🤔

rogermparent: The `lint-ts` command is `tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .`, so it must be `eslint` finding this after `tsc` checks TypeScript. I guess eslint warnings don't make the test fail on CI.

content/community.json
shcheklein, line 58: let's also update the `Women in Data Science San Diego` date - let's set Friday, March 27, 2020 for now . I'm not sure if we'll attend this event at all now

shcheklein:
Looks great! Really cool refactoring. One question - do we handle the case when no events available - how do we handle it?

content/community.json
rogermparent: Do we want to do this here, or in another smaller PR that may get in quicker? I'm fine doing it either way.

scripts/deploy-with-s3.js
shcheklein: good catch ... let's create a ticket to change this - it's p1-p0

content/community.json
shcheklein: here is fine

rogermparent: Done !

src/utils/shared/expiration.js
```diff
@@ -0,0 +1,56 @@
...
+
+const DEFAULT_EXPIRATION_OFFSET = { days: 7 }
+
+/*
+   Get the expiration date for an object with date and/or expires fields
+   If expires is given, return it as a moment
+   If expires is undefined, return date plus 7 days
+   If expires is false, return false, as the item never expires.
+*/
+function getExpirationDate({ date, expires }) {
```
shcheklein, line 11: I think we can add tests for this

scripts/deploy-with-s3.js
rogermparent: I'm on it!

src/utils/shared/expiration.js
shcheklein: the function can be simplified I think ... try to check CodeClimate for cases like this

if (expires === true) return moment(expires)
if (expires === false) return null
if (date) return moment(date).add(DEFAULT_EXPIRATION_OFFSET)
return null

rogermparent:
> Looks great! Really cool refactoring. One question - do we handle the case when no events available - how do we handle it?

Yes! I made sure to test this. When there's no events available, the helper hook returns `null`. All other cases return a 3-length array with `null` padding any usused slots (this fixes a visual bug, but that bug should probably be fixed with CSS instead of data shape).

The component interprets this `null` as a signal to not render the events section at all, header included. This isn't how it has to be, however, as we can have the component render however we want to when `events` is `null`. 

Two options come at first thought for handling situations with no events:
- Remove the whole section like we do now
- Keeping the header and displaying a "There are no upcoming events, check back later" message

shcheklein:
> This isn't how it has to be, however, as we can have the component render however we want to when events is null.

yeah, the problem with not rendering it is that we still have in the Menu (and potentially other outside links we don't have control over). So, it's better to render it with some message "No upcoming events. Subscribe to be up to date 👇"

rogermparent:
> yeah, the problem with not rendering it is that we still have in the Menu (and potentially other outside links we don't have control over). So, it's better to render it with some message "No upcoming events. Subscribe to be up to date point_down"

Good point! I'll go with the text then.

scripts/deploy-with-s3.js
rogermparent: Done at #1291

rogermparent:
I canceled the deploy yesterday because it started to take forever with odd errors, and I wanted to get a good night's rest before I go in on Heroku issues.

Thankfully, it seems the only issue is that the caches were incompatible and a full rebuild succeeded- though it took 800s, the cache is fresh now and the next build should be normal. I'll report back to confirm this when I do it. I'll try a quick test now and then keep an eye on it when I make unit tests for the expiration behavior.

rogermparent:
There was an odd issue where a field not being explicitly declared broke cache re-use on Heroku alone, but it seems like adding it did the trick! I didn't even have to revert the `yarn upgrade`.

On that note, the newer version of Gatsby the branch has logs which specific queries are taking too long! 

```
warning Query takes too long:
File path: /tmp/build_c87156d0007824c1d78716f29e5d33db/src/templates/blog-home.tsx
URL path: /blog
Context: {
    "isBlog": true,
    "limit": 7,
    "pageInfo": {
        "currentPage": 1,
        "nextPage": "/blog/page/2"
    },
    "skip": 0
}
```

This will be a valuable clue on how to improve build time further, as well as keep it low in the face of future additions!

src/utils/shared/expiration.js
rogermparent: Now that I fixed builds, tests are next on my radar.

I believe this is the simplest way to write the exact desired behavior for expiration dates and it seems to work, but automated tests will make that much easier to confirm.

src/components/Community/Events/index.tsx
```diff
@@ -112,11 +104,20 @@ const Events: React.FC<{ theme: ICommunitySectionTheme }> = ({ theme }) => {
...
-            <div className={sharedStyles.item} key={key}>
-              {event && <Event {...event} theme={theme} key={event.url} />}
+          {events ? (
+            events.map((event, key) => (
+              <div className={sharedStyles.item} key={key}>
+                {event && <Event {...event} theme={theme} key={event.url} />}
+              </div>
+            ))
+          ) : (
+            <div className={styles.eventsPlaceholder}>
```
shcheklein, line 114: one quick suggestion -from the design perspective - there are two ways we show text on the page - gray small and black large - let's use one of those for the text. I would start with grey small.

rogermparent: Sounds good! I was wondering about something like that.

rogermparent: Done, but you may want to see if I made the optimal implementation. I'm not sure where the right shared CSS to use would be, and don't seem to see any from context.

src/utils/shared/expiration.js
rogermparent: Code is simplified, and now there's tests that all pass!

src/utils/shared/expiration.js
```diff
@@ -0,0 +1,48 @@
...
+
+/*
+   This is the primary logic to check if a date is expired,
+   It simply uses Moment to parse a date input and comparse that to the current
+   time.
+   Use this on the result of getExpirationDate to get both pieces of
+   information.
+*/
+function dateIsExpired(expires) {
+  return Boolean(expires && expires.isBefore(moment()))
```
shcheklein, line 26: just curious - can && return undefined? do you need to wrap it with explicit Boolean( )?

shcheklein:
Not merging at night, let's do this tomorrow - since we had some problems with cache or something (will need to invalidate it, right?)

shcheklein:
Btw, the "upcoming events text" is still black for me - is it expected?

rogermparent:
@shcheklein I'm not sure why it's still black, it was gray in development for me. I'll fix that quickly, it's supposed to be gray.

I don't think we'll have to invalidate cache, the only reason I had to before was because an inferred field behaved differently on Heroku, breaking the first build and triggering the fallback "clean and retry" mechanism. Now that the field is explicitly defined, the caches should be compatible enough not to re-trigger image generation.

src/utils/shared/expiration.js
rogermparent: JS `&&` and `||` will always return the value it ended on, even on short-circuits.

For example, `null && true` will return `null`, not `false`. You can try it on the Node command line yourself, I just did to confirm.

rogermparent:
> Btw, the "upcoming events text" is still black for me - is it expected?

I had a commit that fixed a typo in the class that made it gray for me- turns out our verify step rejects commits that contain only case changes because it sees them as unchanged, and I missed that before I stopped last night.
It went through fine with `no-verify`, but that's a quirk worth remembering.

PR #1282, opened 2020-05-11
new header

andronovhopf:
New header image about the mailing list for community page

Needs:
-> a link. currently links to ML Prague event page :)
-> any suggestions for editing banner- need bigger text?


src/components/Community/data.json
```diff
@@ -1,7 +1,7 @@
 {
   "hero": {
-    "pictureDesktop": "/img/community/banner.png",
-    "pictureMobile": "/img/community/banner-mobile.png",
+    "pictureDesktop": "/img/community/header_mailing_list.png",
```
shcheklein, line 3: can we get the same with a plain white background? (and may be w/o the DVC logo)

andronovhopf:
definitely

src/components/Community/data.json
```diff
@@ -1,7 +1,7 @@
...
+    "pictureMobile": "/img/community/mobile_header_mailing_list.png",
     "url": "https://www.mlprague.com/#schedule-saturday"
```
shcheklein, line 5: we need to update the link .. we don't have a good one now ... @rogermparent can help you setup an anchor and some smooth scroll to the subscribe section ... or we can link to the mailchimp popup

src/components/Community/data.json
andronovhopf: just pushed a white banner

rogermparent:
I just pushed a small change to this branch that links the Hero to the Subscribe section. A few tiny code changes had to be made because it assumed an external link before, and I also added CSS smooth scrolling.

andronovhopf:
A few other updates on this branch:

> updated "Become an ambassador" link to point to Marcel's blog, not send an email to DVC
> updated user content to have three more recent blogs

src/components/Community/Contribute/index.tsx
```diff
@@ -94,11 +94,11 @@ const Contribute: React.FC<{ theme: ICommunitySectionTheme }> = ({ theme }) => {
                 <Link
                   className={sharedStyles.button}
                   style={theme}
-                  href="mailto:info@dvc.org?subject=I want to become an Ambassador!"
+                  href="https://dvc.org/blog/dvc-ambassador-program-announcement"
                   target="_blank"
```
shcheklein, line 98: @andronovhopf @rogermparent we can remove `target=_blank` in this case?

src/components/Community/data.json
```diff
@@ -44,26 +44,26 @@
...
       "author": "Marcel Mikl and Bert Besser",
       "date": "2020-01-28",
       "pictureUrl": "/img/community/ugc/codecentric.png"
-    },
-    {
-      "url": "https://martinfowler.com/articles/cd4ml.html",
-      "title": "Continuous Delivery for Machine Learning",
-      "author": "Danilo Sato, Arif Wider and Christoph Windheuser",
-      "date": "2019-09-20",
-      "pictureUrl": "/img/community/ugc/fowler_icon.ico"
```
shcheklein, line 59: @andronovhopf  we can delete these images?

src/components/Page/base.css
```diff
@@ -29,6 +29,7 @@ html {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   text-rendering: optimizeLegibility;
+  scroll-behavior: smooth;
```
shcheklein, line 32: @rogermparent this affects other pages, including docs. We didn't want this behavior on docs for a good reason (it takes too long to scroll if page is long and open a link someone shared with you, or click on the right side bar).

shcheklein:
@andronovhopf since we making these updates :) Let's update:

```
1380+
registered developers
50K+
messages posted over the past month
``` 
?

src/components/Community/Hero/index.tsx
```diff
@@ -17,12 +17,7 @@ const Hero: React.FC = () => {
 
   return (
     <LayoutWidthContainer className={styles.container}>
-      <Link
-        className={styles.link}
-        href={data.hero.url}
-        target="_blank"
-        onClick={logHero}
-      >
+      <Link className={styles.link} href={data.hero.url} onClick={logHero}>
```
shcheklein, line 20: @rogermparent what will happen if put an external link in the next banner?

src/components/Community/data.json
```diff
@@ -1,8 +1,8 @@
 {
   "hero": {
-    "pictureDesktop": "/img/community/banner.png",
-    "pictureMobile": "/img/community/banner-mobile.png",
-    "url": "https://www.mlprague.com/#schedule-saturday"
+    "pictureDesktop": "/img/community/header_mailing_list_skinny.png",
```
shcheklein, line 3: @rogermparent when you start reconciling with your new PR - let's rename those to default-banner.png and default-mobile-banner.png ?

src/components/Community/data.json
```diff
@@ -1,8 +1,8 @@
 {
   "hero": {
-    "pictureDesktop": "/img/community/banner.png",
```
shcheklein, line 3: @andronovhopf if we don't use these images anymore - we can delete them

src/components/Community/Contribute/index.tsx
rogermparent: I'd say so. Our Link wrapper will actually add `target="blank"` because it detects this as an absolute link. If we want this link not to behave as `target="_blank"`, I'd suggest changing it to root-relative (i.e. `/blog/dvc-ambassador-program-announcement`) on top of removing `target="blank"` so the wrapper detects it as relative. Using root-relative for inter-site links is best practice for us as well.

shcheklein: @rogermparent sounds good! would mind making that change please?

src/components/Page/base.css
rogermparent: Ah, okay! I'll revert this line and we can go for another approach.

src/components/Community/Hero/index.tsx
rogermparent: Our `Link` wrapper component detects external links and adds proper `target="_blank"` and `rel="noopener noreferrer"` accordingly. Explicitly setting these props isn't necessary unless there's an edge case like wanting an internal link to have `target="_blank"`.

src/components/Community/data.json
rogermparent: Sounds good!

src/components/Community/data.json
```diff
@@ -1,8 +1,8 @@
...
-    "pictureMobile": "/img/community/banner-mobile.png",
-    "url": "https://www.mlprague.com/#schedule-saturday"
+    "pictureDesktop": "/img/community/header_mailing_list_skinny.png",
+    "pictureMobile": "/img/community/mobile_header_mailing_list_skinny.png",
+    "url": "#subscribe"
```
shcheklein, line 5: @rogermparent I'm also getting some strange behavior on the mobile version - it does scroll (w/o smooth effect for some reason), but does not open the right point on the page (some text is hidden) ... it might be related or looks similar to #712

rogermparent: I have to check this out more closely, thanks for the catch!

src/components/Community/data.json
andronovhopf: yep

src/components/Community/data.json
andronovhopf: got it!

andronovhopf: done

src/components/Community/data.json
andronovhopf: done

src/components/Community/data.json
shcheklein: 🙏

src/components/Community/data.json
shcheklein: 🙏

src/components/Community/data.json
```diff
@@ -44,26 +44,26 @@
...
+      "author": "Vimarsh Karbhari",
+      "date": "2020-04-23",
+      "pictureUrl": "/img/community/ugc/medium_logo.png"
+    },
+    {
+      "url": "https://blog.codecentric.de/en/2020/01/remote-training-gitlab-ci-dvc/",
+      "title": "Manage your Data Science Project in R",
+      "author": "Marcel Ribeiro-Dantas",
+      "date": "2020-03-05",
+      "pictureUrl": "/uploads/avatars/marcel.jpg"
```
shcheklein, line 59: one more small ask, @andronovhopf  :) please let's duplicate this image to the `/img/community/ugc` or something inside `/img/community` - it will be easier to manage these assets - if we don't mix them.

rogermparent:
@shcheklein I re-used some existing JS smooth-scrolling behavior in a way that doesn't break when the Hero is external. I broke some TS though, so I need to fix that.

rogermparent:
My TS is fixed! There may be a better way to smooth scroll, but what I have works. It also seems to handle the scrolling issue @shcheklein had because it's using an actual JS implementation and not just depending on default Fragment behavior.

src/components/Community/data.json
andronovhopf: great! doing it now

andronovhopf: done

shcheklein:
thanks @andronovhopf and @rogermparent !!

Issue #753, opened 2019-10-28
react: CONTENTS highlights 2nd header on top window position


jorgeorpinel:
Right-side CONTENTS element doesn't highlight the right header, at least with scroll 0 in the example below:

![image](https://user-images.githubusercontent.com/1477535/67724514-e72dc500-f9a4-11e9-8702-e75e7a3b5a2c.png)
> https://dvc.org/doc/install/macos

jorgeorpinel:
Maybe also related: #750

shcheklein:
#750 is not related. In this case the problem is that three and/or more H2 titles are visible. Some of the even above the middle of the page. In this case the heuristic we use to determine which of the titles to highlight decides to show the second one. I think this just an edge case that can be fixed by applying a very specific detection that page is at it's most top or very very close to that and in that case force the very first H2.

jorgeorpinel:
Also related/ part of the solution: #723

iKBAHT:
Some ideas:
https://github.com/iterative/dvc.org/blob/master/src/components/Documentation/RightPanel/index.tsx#L44
A peculiarity of the navigation algorithm

jorgeorpinel:
Ping @iterative/websites small bug outstanding here.

Also related: https://github.com/iterative/dvc.org/issues/723

Issue #723, opened 2019-10-22
engine: create title entry in CONTENTS nac (right side bar) 


jorgeorpinel:
Often there's a long intro in our docs before getting to the first H2 header (example below) but the CONTENTS navigation element in the right side bar does not list the document's H1 title, so it's not possible to scroll to the very top from there, which I think would be a logical feature.

![image](https://user-images.githubusercontent.com/1477535/67315649-9d3c6f00-f4cc-11e9-9869-3562716c30a6.png)
> https://dvc.org/doc/tutorials/versioning

jorgeorpinel:
I keep bumping into this issue. WDYT @shcheklein, worth prioritizing soonish?

rogermparent:
Seems simple enough, but we may want to consider a design other than simply adding the top heading to the content bar.

For example, we could add a "Return to Top" button like in many other blogs; it could be located next to the "content" heading when the list is present, and maybe show standalone when on a page without subheadings.

shcheklein:
@jorgeorpinel @rogermparent let's not prioritize this for now please. We'll do this when we have more capacity.

jorgeorpinel:
> There is an issue we must solve in how to handle pages that hide the content menu, i.e. pages with no headings besides the top h1.

Good point.

> let's not prioritize this for now

OK

PR #1284, opened 2020-05-12
doc engine: fix scroll bar for the content sidebar

### LINKED ISSUE #1198: doc engine: fix scroll bar for the content sidebar
We want for buttons to always stay on the screen and the sidebard to have its own scrollbar.

Screen Shot 2020-04-28 at 6 46 10 PM

iKBAHT:
Fixes #1198

### DIFF
--- a/src/components/Documentation/RightPanel/index.tsx
+++ b/src/components/Documentation/RightPanel/index.tsx
@@ -97,7 +97,14 @@ const RightPanel: React.FC<IRightPanelProps> = ({
     setIsScrollToCurrentHeadingHappened
   ] = useState(false)
   useEffect(() => {
-    if (currentHeadingSlug && !isScrollToCurrentHeadingHappened) {
+    if (isScrollToCurrentHeadingHappened) {
+      return
+    }
+    if (!document.location.hash) {
+      setIsScrollToCurrentHeadingHappened(true)
+      return
+    }
+    if (currentHeadingSlug) {
       setIsScrollToCurrentHeadingHappened(true)
       const currentHeadingSlugElem = document.getElementById(
         `link-${currentHeadingSlug}`
--- a/src/components/Documentation/RightPanel/index.tsx
+++ b/src/components/Documentation/RightPanel/index.tsx
@@ -27,12 +27,14 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   tutorials,
   githubLink
 }) => {
-  const [height, setHeight] = useState(0)
+  const [documentHeight, setDocumentHeight] = useState(0)
   const [headingsOffsets, setHeadingsOffsets] = useState<IHeadingsCoordinates>(
     {}
   )
-  const [current, setCurrent] = useState<string | null>(null)
-  const setCurrentHeader = (): void => {
+  const [currentHeadingSlug, setCurrentHeadingSlug] = useState<string | null>(
+    null
+  )
+  const updateCurrentHeader = (): void => {
     const currentScroll = getScrollPosition()
     const coordinateKeys = Object.keys(headingsOffsets)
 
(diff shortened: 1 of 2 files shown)


shcheklein:
@iKBAHT thanks! great.

A few things I noticed so far:

1. Katacoda button should not be part of the scroll I think

Screen Shot 2020-05-12 at 8 03 33 AM

2. CONTENT - - should not be part of the scroll probably. It'll become a back to the top of the page. (related #723 )

3. Can we make buttons go right after the sidebar as it was before. The reason here is we need to keep them as visible as possible. Especially "Katacoda" if it's visible.

iKBAHT:
I implemented noticed things. Also, I decrease distance between buttons from `25px` to `15px`. Otherwise, buttons block take too many heights with `Katacoda` block.

shcheklein:
@iKBAHT looks great overall, one more question - right now when there are some items that are hidden - sometimes it is not obvious that they exist (scroll is hidden). I kinda like that scroll is hidden, but is there a way from a user perspective to see that there is something below and you can scroll it ...? what do you think?

iKBAHT:
I think the best way, in this case, is to always show a scroll bar if it exists.

iKBAHT:
I don't even know how else to show the user that there is something below

shcheklein:
@iKBAHT yeah, I was thinking to try some fade out and show scroll when you hover over it. 

The problem with always showing it is that it looks like a border line in some cases (when it's almost full) :)

Anyways, not a big deal - let me know if it's easy to try those two things ... otherwise I think we cab merge it as is.

iKBAHT:
Fade-out is not a common practice, users are familiar with scrolls. Also, fade-out takes an already limited amount of space.
A solution with scroll is very easy to implement.

shcheklein:
> A solution with scroll is very easy to implement.

yep, works for now

One more thing we need to think and take care of:

We need to scroll into view the highlighted element when we open a link like this (and screen is not large enough).

https://dvc-landing-fix-scroll--bt87yn.herokuapp.com/doc/command-reference/fetch#example-with-dependencies

or as we scroll the screen and some items are not visible

iKBAHT:
> We need to scroll into view the highlighted element when we open a link like this
done

During implementation find the new bug #1306

jorgeorpinel:
> Also, during implementation, I found the new bug #1306

Probably the same as #753 🙂

shcheklein:
Something funny is happening - when I open https://dvc-landing-fix-scroll--bt87yn.herokuapp.com/ and click docs I'm getting:

Screen Shot 2020-05-18 at 6 14 35 PM

Note - no Home entry, Content went under the menu

iKBAHT:
Yes, it's my fail. I tested only when scroll exists.

shcheklein:
@iKBAHT still see some problems - https://dvc-landing-fix-scroll--bt87yn.herokuapp.com/doc/command-reference/fetch

shcheklein:
Looks like scroll is not always visible as expected in Firefox

Also getting this:
Screen Shot 2020-05-19 at 5 56 56 PM

(It hides the text)

In Safari - it always scrolls the page for me, even if mouse pointer located on top of the sidebar area.

Also, it does not scroll the selected item into view after navigate? Am I missing something?

iKBAHT:
- I wasn't aware of the need to maintain FF. But anyway FF doesn't support "force" render scrollbar.
- > In Safari - it always scrolls the page for me, even if mouse pointer located on top of the sidebar area.

It's standard behavior, chrome works the same way. I mean if block doesn't have it own scroll, then scroll goes to parent.

- > Also, it does not scroll the selected item into view after navigate? Am I missing something?

Can you please write steps to reproduce. Everything works for me.

shcheklein:
> I mean if block doesn't have it own scroll, then scroll goes to parent.

In this case it looks like it happens even if block has its own scroll.

> I wasn't aware of the need to maintain FF. 

yes, we support all major browsers + iOS Safari + Android

> Can you please write steps to reproduce. Everything works for me.

may be we talk about different problem. I mean that an item on the right _sidebar_ should be scrolled into view (same as we do with the left sidebar).

iKBAHT:
> In this case it looks like it happens even if block has its own scroll.

Everything works for me. Can you share the problem somehow.

> may be we talk about different problem. I mean that an item on the right sidebar should be scrolled into view (same as we do with the left sidebar).

I think it impossible. Right and left sidebars control by one scrollbar. And if scroll goes to the last article, and it short, the left bar also goes down with the whole page. It comes from main goal of this PR - "fix scroll bar for the content sidebar"

shcheklein:
Okay 👌 Merging this as this is a great improvement anyway, and I will create a separate issue to describe improvements left. It'll be easier to iterate from this.

PR #1297, opened 2020-05-16
Mentioned issue in repro, #233

### LINKED ISSUE #233: ref: `cmd` should be able to create at `repro`
From https://github.com/iterative/dvc/issues/1800#issuecomment-480059069

> another factor contributing to my problems with dvc repro has to do with particularities around listing file folders as outputs in a .dvc file.
>
> This may seem obvious, but if file folders are listed as outputs in a particular stage, users need to ensure that the command for that stage includes a means of checking for the folder's existence and creating it if it doesn't exist. Otherwise dvc repro will fail due to a FileNotFoundError when that stage's command is executed (because dvc repro will have deleted those folders before executing the stage's command).

Zenix27:
Closes #233

### DIFF
--- a/content/docs/command-reference/repro.md
+++ b/content/docs/command-reference/repro.md
@@ -38,7 +38,7 @@ command: by specifying stage file `targets`, or by using the `--single-item`,
 If specific [DVC-files](/doc/user-guide/dvc-file-format) (`targets`) are
 omitted, `Dvcfile` will be assumed.
 
->Here, user script should be able to make directories.
+Here, user script should be able to make directories.
 
 `dvc repro` does not run `dvc fetch`, `dvc pull` or `dvc checkout` to get data
 files, intermediate or final results.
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Hi. Is this WIP? (work in progress) Thanks

Also, please link to the original issue by using special words like "Closes #issue-no" in the PR description above. See https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

shcheklein:
@Zenix27 sorry, but I don't quite understand this change. It does not seem to fix the #233 the way I understand it. Could you describe in a comment how did you understand the issue, what should we put a note about?

Zenix27:
@jorgeorpinel I thought its just a Documentation issue, that had to be literally mentioned in the repro documentation

jorgeorpinel:
@Zenix27 I don't think you understood the task correctly. Granted it didn't have much of an explanation... So I just updated the desciption of #233, please check it out.

Addressing this will require you to first learn what `dvc repro` does and play with it first (see https://dvc.org/doc/command-reference/repro). Then understand the description of that issue, and then you should be able to confirm whether some details are missing from the command reference. Please lmk if you have further questions.

shcheklein:
Closing this for now. @Zenix27 thanks for helping us!! 🙏 please take your time read through the description Jorge mentioned, may be try to run the get started to better understand the issue. And feel free to ask question in the issue if it is still not clear enough.

PR #1283, opened 2020-05-12
Regular updates (May 11)

jorgeorpinel:
- A few more changes to plot cmd refs (matched to core repo) — see iterative/dvc/pull/3759
- GSoD blog post update
- Misc. stuff e.g. https://github.com/iterative/dvc.org/pull/1268#pullrequestreview-417252532
- Closes #1314
- Finish #1318 to close #1301

PR #1305, opened 2020-05-18
docs: API auto-mentioning in docs (#910)

### LINKED ISSUE #910: automatic links for API function names (to api ref)
Similar to how we automatically link `dvc <command>` to their cmd ref in the docs, the following quoted strings (can be pulled from sidebar.json) should link to the corresponding API reference page: `dvc.api.get_url()`, `dvc.api.open()`, `dvc.api.read()`, and `dvc.api.summon()`.

UPDATE: Another one could be `dvc.api` -> /doc/api-reference (also get from sidebar.json)

This way we can easily start mentioning these functions from everywhere in the docs, and have the users be able to jump to the API ref.

- [ ] And replace any manual links to these introduced with #908 or any other PR before this is implemented.

mvshmakov:
Fix #910

### DIFF
--- a/content/docs/api-reference/read.md
+++ b/content/docs/api-reference/read.md
@@ -4,11 +4,11 @@ Returns the contents of a tracked file.
 
 ```py
 def open(path: str,
-        repo: str = None,
-        rev: str = None,
-        remote: str = None,
-        mode: str = "r",
-        encoding: str = None)
+         repo: str = None,
+         rev: str = None,
+         remote: str = None,
+         mode: str = "r",
+         encoding: str = None)
 ```
 
 #### Usage:
--- a/content/docs/api-reference/read.md
+++ b/content/docs/api-reference/read.md
@@ -4,11 +4,11 @@ Returns the contents of a tracked file.
 
 ```py
 def open(path: str,
-         repo: str = None,
-         rev: str = None,
-         remote: str = None,
-         mode: str = "r",
-         encoding: str = None)
+        repo: str = None,
+        rev: str = None,
+        remote: str = None,
+        mode: str = "r",
+        encoding: str = None)
 ```
 
 #### Usage:
@@ -24,9 +24,9 @@ modelpkl = dvc.api.read(
 
 ## Description
 
-This function wraps [`dvc.api.open()`](/doc/api-reference/open), for a simple
-way to return the complete contents of a file tracked in a <abbr>DVC
-project</abbr>. The file can be tracked by DVC or by Git.
+This function wraps `dvc.api.open()`, for a simple way to return the complete
+contents of a file tracked in a <abbr>DVC project</abbr>. The file can be
+tracked by DVC or by Git.
 
 > This is similar to the `dvc get` command in our CLI.
 
--- a/content/docs/use-cases/data-registries.md
+++ b/content/docs/use-cases/data-registries.md
@@ -139,7 +139,7 @@ current working directory (anywhere in the file system with user write access).
 
 ```dvc
 $ dvc import https://github.com/example/registry \
-            images/faces/
+             images/faces/
 ```
 
 > Note that unlike `dvc get`, which can be used from any directory, `dvc import`
@@ -202,7 +202,7 @@ and this will be picked up by Git:
 $ git status
 Changes not staged for commit:
 ...
-  modified:   music/songs.dvc
+      modified:   music/songs.dvc
 $ git commit -am "Add 1,000 more songs to music/ dataset."
 ```
 
--- a/content/docs/use-cases/data-registries.md
+++ b/content/docs/use-cases/data-registries.md
@@ -164,9 +164,9 @@ the project dependency metadata in the import stage (DVC-file).
 
 ### Programmatic reusability of DVC data
 
-Our Python API, included with the `dvc` package installed with DVC, includes the
-`open` function to load/stream data directly from external <abbr>DVC
-projects</abbr>:
+Our Python API (`dvc.api`), included with the `dvc` package installed with DVC,
+includes the `open` function to load/stream data directly from external
+<abbr>DVC projects</abbr>:
 
 ```python
 import dvc.api.open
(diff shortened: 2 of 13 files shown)


mvshmakov:
There is one thing: I don't know whether I should consider the proposed by CodeClimate "Cognitive Complexity"  issue or not due to the fact that the commandLinker has the same complexity and it was in the production already.

content/docs/api-reference/read.md
```diff
@@ -4,11 +4,11 @@ Returns the contents of a tracked file.
 
 ```py
 def open(path: str,
-         repo: str = None,
```
shcheklein, line 7: that's weird a bit - was it done automatically by the pre-commit hook?

content/docs/use-cases/data-registries.md
```diff
@@ -202,7 +202,7 @@ and this will be picked up by Git:
 $ git status
 Changes not staged for commit:
 ...
-	modified:   music/songs.dvc
```
shcheklein, line 205: @jorgeorpinel what are the recommended settings for the editor? do we require to keep tabs? (it can be annoying a bit)

shcheklein: @mvshmakov please, check that if you replace tabs with spaces, make it at least 4 so that output looks more or less the same as in the DVC

content/docs/use-cases/data-registries.md
```diff
@@ -139,7 +139,7 @@ current working directory (anywhere in the file system with user write access).
 
 ```dvc
 $ dvc import https://github.com/example/registry \
-             images/faces/
```
shcheklein, line 142: same here - not sure what's going on

shcheklein:
> There is one thing: I don't know whether I should consider the proposed by CodeClimate "Cognitive Complexity" issue or not due to the fact that the commandLinker has the same complexity and it was in the production already.

CC is too aggressive and we _try_ to fix those issue, but it's not a requirement - especially complexity and number of lines of code

shcheklein:
Deployed a preview (it does not happen automatically on forks, usually we use branches) - https://dvc-landing-feature-910-23uvld.herokuapp.com/

content/docs/api-reference/read.md
rogermparent: I just tried it, and that doesn't seem to be the case. Neither `yarn format-all` nor attempting a dummy commit automatically entered this indent.

rogermparent: In fact, our formatting tools seem not to check these code blocks at all.

content/docs/use-cases/data-registries.md
jorgeorpinel: > what are the recommended settings for the editor?

We currently have a mix of spaces and tabs in command outputs throughout the docs I think, because people do whatever they prefer. I've advocated to just copy and paste the output exactly as it is from console, so that would typically include tabs.

shcheklein: yep, agreed. But it's a little bit hard to manage- a lot of editors change them from Tabs to spaces automatically. Anyways, @mvshmakov let's keep the formatting the way it was before and create a ticket if we need to update out policies.

jorgeorpinel: Editors shouldn't change tabs inside code blocks, if they understand the format is MD. Anyway, it's a separate issue, agree!

content/docs/api-reference/read.md
mvshmakov: I have `prettier` installed in my IDE. It automatically detects `.prettierrc` and formats everything according to some IDE defaults and setting specified explicitly in `.prettierrc` and `.eslintrc.json`. @rogermparent, there are no specific settings for `.md`, but prettier is still checking these files according to defaults. Anyway, I have returned the formatting.

content/docs/use-cases/data-registries.md
mvshmakov: > @mvshmakov please, check that if you replace tabs with spaces, make it at least 4 so that output looks more or less the same as in the DVC

Yes, this is my default "indent on save" even for the first commit.

> yep, agreed. But it's a little bit hard to manage- a lot of editors change them from Tabs to spaces automatically. Anyways, @mvshmakov let's keep the formatting the way it was before and create a ticket if we need to update out policies.

That's it https://github.com/iterative/dvc.org/issues/1315.

mvshmakov:
> Looks good. I just wonder if we need `dvc.api` to link to the API ref. Probably yes but maybe also remove some manual links that make it repetitive, like this one from https://dvc-landing-feature-910-23uvld.herokuapp.com/doc/install
> 
> image

I have removed what you suggested, and I can't find anything like this anymore.

content/docs/use-cases/data-registries.md
```diff
@@ -164,9 +164,9 @@ the project dependency metadata in the import stage (DVC-file).
 
 ### Programmatic reusability of DVC data
 
-Our Python API, included with the `dvc` package installed with DVC, includes the
-`open` function to load/stream data directly from external <abbr>DVC
-projects</abbr>:
+Our Python API (`dvc.api`), included with the `dvc` package installed with DVC,
```
jorgeorpinel, line 167: Here I would actually prefer to link `[Python API]` manually for readability.

content/docs/api-reference/read.md
shcheklein: 👍

shcheklein:
@jorgeorpinel I will merge this (since you approved and to move forward), let's update that text as part of the regular PRs?

jorgeorpinel:
Sure!

content/docs/use-cases/data-registries.md
jorgeorpinel: Done in https://github.com/iterative/dvc.org/pull/1283/commits/036672de4d6b2253a6462a2b152c251f6d0f03d3 along with a couple more instances I found.

PR #1312, opened 2020-05-19
Fix docs sidebar home highlight

rogermparent:
This is partway through my work adding custom icons to the sidebar-

I noticed that the docs home currently doesn't have its entry highlighted when on the page. This branch fixes that, and is what I'll be building on top of in the `docs-sidebar-bullets` branch that tackles that issue. This branch is in case we want to fix this small issue now before I finish sidebar icons. Not that I expect it to take too long, but the changes may need more review than this logical midpoint.

### DIFF
--- a/src/components/Documentation/Layout/index.tsx
+++ b/src/components/Documentation/Layout/index.tsx
@@ -14,7 +14,7 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
   const [isMenuOpen, setIsMenuOpen] = useState(false)
 
   // Remove trailing slash from location for normalized usage all throughout the app
-  const pathname: string = location.pathname.endsWith('/')
+  const pathname: string = restProps.location.pathname.endsWith('/')
     ? restProps.location.pathname.slice(
         0,
         restProps.location.pathname.length - 1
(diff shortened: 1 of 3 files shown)


jorgeorpinel:
Please link to the issue this addresses/fixes/closes 🙂

rogermparent:
> Please link to the issue this addresses/fixes/closes slightly_smiling_face

I would have, but I didn't find one. Maybe I didn't search for the right keywords? I figured it was such a small fix that there was no need to open a new issue.

src/components/Documentation/Layout/index.tsx
```diff
@@ -13,6 +13,14 @@ import styles from './styles.module.css'
 const Layout: LayoutComponent = ({ children, ...restProps }) => {
```
shcheklein, line 13: where do we get `restProps` ? can we make this trailing slash issue localized somehow? (don't like that we have to apply "slice" in a few places already

rogermparent: You're right to be suspect of this check. I was thinking that the `location` from here is the one Gatsby provides, but that wouldn't change with the URL so this must be the global `location` from the browser that's prop drilled into the docs component for some reason.

Next push will have a change that resolves this.

rogermparent: I've now changed the way the Layout sources the current path:

before, it came from the window global `location`, but now it comes from Gatsby's static `location`. Since the paths Gatsby provides are all ensured to have the trailing slash removed, we now don't have to check for it- at least for anything that sources `pathname` from the Layout, and I think I got every instance of that here.

PR #1322, opened 2020-05-19
Special bullet/style for sidebar docs home

### LINKED ISSUE #1053: nav: allow different bullet icons in sidebar
Currently the only one is a little triangle.
- We may want no icon sometimes, e.g. the new Home entry being introduced in https://github.com/iterative/dvc.org/pull/1051#issuecomment-599330238
- Sidebar entries without children should probably have a regular circle bullet point (per https://github.com/iterative/dvc.org/issues/1052#issuecomment-599318329)

rogermparent:
Builds on #1312 
Fixes #1053, which is also one of the remaining parts of #1073

This branch adds capability for pre-defined special styles and icons to be applied on a per-item basis in `sidebar.json`.
It also uses this capability to give the doc home a little house icon.

I got the icon from [here](https://www.svgrepo.com/svg/56717/home) (It's under CC0 / public domain so we don't have to worry, but it can be replaced if needed) and modified it to use `currentColor`, meaning it takes on the color of the containing link.

The default `:before` based bullet behavior is left the same, but special icons are real elements because it offers much more flexibility for no cost, as well as the `currentColor` capability and optimization components like SVGR.

Custom icons are defined as React components, so we could define anything as one- even raster images or pure CSS divs!


rogermparent:
This branch now also turns leaf item default bullets into circles via CSS. These are implemented as `:before` like the original bullets, not special ones like home.

shcheklein:
@rogermparent looks cool but a bit noisy from the design perspective.

let's try this:

1. change triangles state - it should face down when list is open
2. if there are no children - skip the icon at all. 
3. don't make Home **bold**

rogermparent:
Can do! Thanks for the feedback!

rogermparent:
>     1. change triangles state - it should face down when list is open
> 
>     2. if there are no children - skip the icon at all.
> 
>     3. don't make Home **bold**

All fixed! I changed some padding around too because the left pad looked oddly large without bullets sitting in the space.

shcheklein:
@rogermparent looks better! I still don't like the home 🏠 icon - let's try to remove it? or may be the problem is that we highlight along with the text (not the case for triangles) + it's not well aligned (should be a bit to the right and down). Let's play with it a little bit.

shcheklein:
@rogermparent I can do it on my won, give me a few minutes

rogermparent:
Sure! Get it exactly how you want it, I don't have a great eye for design. I'm sure you're already using the files changed to navigate what you need to edit.

It's worth noting that you can set `color` on the home icon to lock it to whatever you want, it's only inheriting the text color because it doesn't have one set- no special logic to undo.

shcheklein:
Configuration I like the most so far for the home icon:


Screen Shot 2020-05-20 at 3 04 07 PM
Screen Shot 2020-05-20 at 3 06 10 PM


The thing that was not easy to try for me - for leaf items - can we make a bullet, but a small one (it should fit inside a triangle to give you a reference).

Sorry for this pixel movements - I'm not a designer either, but have a good eye - just need to move things around to see myself to decide :)

rogermparent:
> The thing that was not easy to try for me - for leaf items - can we make a bullet, but a small one (it should fit inside a triangle to give you a reference).

Ah, yes, I disabled leaf bullets in logic and removed them completely, sorry about that! I can re-add them with a new SVG that fits your description.

> Sorry for this pixel movements - I'm not a designer either, but have a good eye - just need to move things around to see myself to decide :)

Don't worry at all! I'm glad that someone else is giving input, even if it's pixel movements- especially where bullets are positioned using `absolute` and not auto-centered.

shcheklein:
Cool 😎 thanks @rogermparent !

src/utils/shared/sidebar.js
```diff
@@ -77,14 +77,20 @@ function findPrevItemWithSource(data, item) {
 function normalizeItem({ rawItem, parentPath, resultRef, prevRef }) {
   validateRawItem(rawItem)
 
-  const { label, slug, source, tutorials, type, url } = rawItem
+  const { label, slug, source, tutorials, type, url, style, icon } = rawItem
```
jorgeorpinel, line 80: May want to update the inline doc example in sidebar.js

jorgeorpinel: *in the top of this file

shcheklein: 👍 yep

rogermparent: @jorgeorpinel I'm not sure what you mean by inline doc example in this file. Did you mean to mention a doc file that describes how to use the sidebar?

rogermparent: Done! #1332

PR #1326, opened 2020-05-21
Update events

andronovhopf:
I've updated the events section of the community page.


andronovhopf:
@rogermparent good catch, i shrank it to 500x500

shcheklein:
@andronovhopf @rogermparent so, it should 315 * 2 = 630 to look good on a modern screen, right?

rogermparent:
Other images range around 500x500, so I think that's a good size!
I haven't heard of screens needing double the image size, but an image this simple isn't going to have any problems: especially the difference between 500x500 and 630x630.

As long as the data of the events themselves is good, I think this one is good to go.

shcheklein:
@rogermparent it's a good practice to do 2x - https://www.danrodney.com/blog/retina-web-graphics-explained-1x-versus-2x-low-res-versus-hi-res/ (the first link that came). It's fine indeed to keep them ~2x, but 500 is already visible. Not a big deal though for this section.

PR #1327, opened 2020-05-21
change term scm to git

### LINKED ISSUE #540: term: SCM -> Git

srishti-nema:
Fixes #540 . Did not change core.no_scm under configuration sections. Also, did not alter the blogs. Let me know if any other changes are required.


jorgeorpinel:
> Did not change core.no_scm under configuration sections.

Good call. Anything that reflects DVC core we don't change in docs lightly without that change actually having happened in the iterative/dvc repo first 🙂

content/docs/command-reference/add.md
```diff
@@ -41,7 +41,7 @@ Under the hood, a few actions are taken for each file (or directory) in
    <abbr>output</abbr>), using its path and hash to identify the cached data.
    Unless the `-f` option is used, the DVC-file name generated by default is
    `<file>.dvc`, where `<file>` is the file name of the first target.
-5. Unless `dvc init --no-scm` was used when initializing the project, add the
+5. Unless `dvc init --no-git` was used when initializing the project, add the
```
jorgeorpinel, line 44: Oops please revert changes to actual DVC commands/options like this one because that has to reflect the reality of the DVC core repo/tool and the term SCM is still used there unfortunately. Thanks!

content/docs/command-reference/commit.md
```diff
@@ -279,5 +279,5 @@ $ dvc status
 Data and pipelines are up to date.
 ```
 
-Nothing special is required, we simply `commit` to both the SCM and DVC. Since
+Nothing special is required, we simply `commit` to both the Git and DVC. Since
```
jorgeorpinel, line 282: This change is correct but can you also please improve the sentence? It's not greatly written "to both the Git and DVC" doesn't sound great.

content/docs/command-reference/add.md
srishti-nema: Alright.

content/docs/command-reference/commit.md
srishti-nema: Okay. Will do that.

content/docs/command-reference/commit.md
```diff
@@ -279,5 +279,5 @@ $ dvc status
 Data and pipelines are up to date.
 ```
 
-Nothing special is required, we simply `commit` to both the SCM and DVC. Since
+Optimization of resources can be easily maintained. Use `commit` to both Git and DVC, when ready. Since
 this pipeline is up to date, `dvc repro` will not do anything.
```
jorgeorpinel, line 283: Hmmm... I'm not sure what this new sentence means either. Please try to understand the context and close the idea here. It's something along the lines of being able to force-register the changes to a stage/pipeline without reproducing it. It's best if you actually try the example to fully get the idea 🙂

jorgeorpinel: May also want to try basic dvc run and dvc repro. See the examples in their cmd refs. Thanks

jorgeorpinel:
@srishti-nema CIis failing and #1336 was open automatically because there are formatting issues in this PR. Please review the corresponding guidelines in our docs contrib guide (https://dvc.org/doc/user-guide/contributing/docs).

content/docs/command-reference/commit.md
srishti-nema: Okay.

srishti-nema:
> @srishti-nema CI is failing and #1336 was open automatically because there are formatting issues in this PR. Please review the corresponding guidelines in our docs contrib guide (https://dvc.org/doc/user-guide/contributing/docs).

Will go through and get back on this.

content/docs/command-reference/commit.md
```diff
@@ -279,5 +279,6 @@ $ dvc status
 Data and pipelines are up to date.
 ```
 
-Nothing special is required, we simply `commit` to both the SCM and DVC. Since
-this pipeline is up to date, `dvc repro` will not do anything.
+Instead of reproducing the pipeline for changes that do not produce a different
+output of pipeline stages, use `commit` to both git and DVC. Since this pipeline
+is up to date, `dvc repro` will not do anything.
```
jorgeorpinel, line 284: Let's just simplify/generalize terminology, so complex context that aren't closely related to this cmd ref or example are avoided:
```suggestion
Instead of reproducing the pipeline for changes that do not produce a different
results, just use `commit` on both Git and DVC.
```
That last sentence was still redundant vs the previous paragraph before the code block...

jorgeorpinel: But this is kind of still missing the main point I mentioned in https://github.com/iterative/dvc.org/pull/1327#discussion_r429603929

jorgeorpinel: > It's something along the lines of being able to force-register the changes to a stage/pipeline without reproducing it.

jorgeorpinel: p.s. the keyword being "force"

content/docs/command-reference/commit.md
jorgeorpinel: Moving to https://github.com/iterative/dvc.org/pull/1327#pullrequestreview-421506486

content/docs/command-reference/commit.md
srishti-nema: I thought we would use the force option when committing to both DVC and Git if force pushing the changes was a requirement but it was not used in the example given. I am sorry, I think I got confused.

jorgeorpinel: Hmmm I forgot there's a `--force` option. I guess using the term "force" here could be confusing after all. It's not really related to force-pushing to Git though.

But my point is the main idea for this last paragraph should be or should include that we are "forcing" DVC to commit the stage file changes even when the stage wasn't reproduced after it's dependencies change (because we know there's no need). Did you actually run the example commands to fully understand the example? In fact it may even need the `--force` option, I'm not sure. Please double check 🙂

srishti-nema: Yes I ran the commands. By using the -f option the prompt doesn't ask the user a yes/no question regarding the decision of committing. That is the only difference. Personally, I don't think it is necessary to put the force option but if that provides more clarity then we can do it.

content/docs/command-reference/commit.md
```diff
@@ -279,5 +279,5 @@ $ dvc status
 Data and pipelines are up to date.
 ```
 
-Nothing special is required, we simply `commit` to both the SCM and DVC. Since
-this pipeline is up to date, `dvc repro` will not do anything.
+Instead of reproducing the pipeline for changes that do not produce a different
```
shcheklein, line 282: I think it should be ` produce different results` or ` produce a different result`.

content/docs/command-reference/commit.md
shcheklein: I think it looks good already, the motivation for using commit in this example is provided a bit above - "we don't want to run repro because changes are not substantial ...".

(also, improving example itself probably out of scope for the ticket)

jorgeorpinel:
Sorry for the delay @srishti-nema this is great, thanks again!

PR #947, opened 2020-01-24
user-guide: Basic Concepts

jorgeorpinel:
Close _user-guide: new "Basic Concepts" doc(s)_ #550

Issue #542, opened 2019-08-10
glossary: create stand-alone page


jorgeorpinel:
It would be nice to have a **Glossary** menu item on the nav bar (or/and linked from tooltips) that opens a page where all these terms are listed (with `<a name='the-term'>` anchors. Extra steps:

- [ ] This is mainly to link from glossary descriptions (because we can't have tooltips inside tooltips)
  Note: currently **bold** words in tooltip texts mark other glossary entries;
- [ ] and to read the full list in case anyone is curious.

> Extracted from https://github.com/iterative/dvc.org/issues/518#issuecomment-519813441

- [ ] Check whether #550 is still necessary

jorgeorpinel:
Maybe after #461 though.

jorgeorpinel:
At this point I think we're definitely merging this idea and the basic concepts so lets close this in favor of #550?

shcheklein:
Agreed!

PR #1334, opened 2020-05-22
Build time tooltips

### LINKED ISSUE #1045: remove `react-markdown` from the Tooltip component

rogermparent:
Fixes #1045 

This PR loads Tooltip data into GraphQL at build time, processing the Markdown descriptions into HTML. The components that consumed this data now no longer depend on `react-markdown`, and instead load the build-time HTML content via `dangerouslySetInnerHTML`.


shcheklein:
A small bug:

Screen Shot 2020-05-22 at 11 38 50 AM

rogermparent:
Thanks for finding that! 
It was an indentation issue in the glossary source. Remark dislikes having things like links broken midway through and skips over offenders. Similar issues can happen with Rehype and `rehype-react` components.

shcheklein:
@rogermparent still not fixed for me, deployment problem?

rogermparent:
> @rogermparent still not fixed for me, deployment problem?

It's fixed on my local machine, so it looks like a cache issue. Maybe it's part of my `contentDigest` logic.

rogermparent:
I changed the Glossary data file to YAML, as it's the data format most friendly to line breaks (as opposed to JSON). All parsing is done at build time, and components still receive the exact same JSON object as before.

Doing this instead of using an ES5/6 module also allows for HMR on tooltips. The rest of the implementation doesn't quite allow for it yet, but this change allows us to simply refresh the page to update instead of rebuilding the application.

shcheklein:
@rogermparent good stuff! a few questions:

1. do we run prettier for yml/yaml?
2. do we use a single Node right now for all glossary entries? would it make sense to actually split the whole file into multiple? we can then probably use regular MD with frontmatter?

rogermparent:
>     1. do we run prettier for yml/yaml?
I don't think so- at least our `format-all` command doesn't include it. I can fix that.

>     2. do we use a single Node right now for all glossary entries? would it make sense to actually split the whole file into multiple? we can then probably use regular MD with frontmatter?

My custom model does that- it mimics the exact data shape we used previously. I can change it to be any data shape we'd like. Our Tooltip component is currently built with a custom state that expects to consume the all the tooltip data as a JSON blob ahead of time and show tooltips when needed. If we split the data into multiple nodes we'll either have to recombine it for the components or change the components to accept piecemeal data.

At this point, the way we store Glossary entries is 100% writer's preference- we can take anything and transform it into the JSON that the frontend components expect during build time for no runtime penalty.

For example, we could have a separate `glossary` directory in `content` that is used to hold glossary entries, and then use those entries to make the Tooltip JSON at build time. We can then also consume the data in other places if we want to.

shcheklein:
> For example, we could have a separate glossary directory in content that is used to hold glossary entries, and then use those entries to make the Tooltip JSON at build time. We can then also consume that data elsewhere.

it sound reasonable to me. It's easier for writers, we don't need to add prettier, we don't need converters. WDYT?

rogermparent:
> > For example, we could have a separate glossary directory in content that is used to hold glossary entries, and then use those entries to make the Tooltip JSON at build time. We can then also consume that data elsewhere.
> 
> it sound reasonable to me. It's easier for writers, we don't need to add prettier, we don't need converters. WDYT?

Sounds good to me! I'll adapt this branch to that data shape once it deploys with the current one. I had to sort out some merge issues with master.

rogermparent:
I got a good deploy with the original setup on Heroku manually a while ago, though it doesn't show here.
I've got the Glossary working with individual Markdown files now! There's a few decisions we can make here on how we want to handle things:

Primarily, we need to figure out **what directory to use for glossary entries** and **where we want the glossary entries to show up** (glossary page, tooltips, and optionally individual pages). I'll explain a few common routes, but we could really do anything here so feel free to make a suggestion for something that I don't describe below.

#### content/docs/glossary

It's right next to the rest of the docs. Entries will be picked up by the page maker by default, creating pages at `docs/glossary/${term}`. The pages work, and are regular doc MD pages that don't show up in the sidebar since the sidebar is manually defined.

To keep glossary entries co-located with docs here, we must either build on these glossary entry pages and turn them into real pages (which is easily done with templates) or suppress them so they only show on the `docs/glossary` page and tooltips. This is a design preference, but suppressing the pages is easier by definition and we can add glossary term pages later on if we decide we want them.

There's also the option of moving the glossary to a directory that isn't picked up by the DocsPage model.

#### content/glossary

Still pretty close to the docs, this folder is still picked up by our `gatsby-source-filesystem` instance but not the `DocsPage` model, so it's safe to keep glossary entries here without them making pages. If we don't ever want individual pages for glossary entries, this is the simplest way to go about it.

rogermparent:
I just realized that the glossary doesn't seem to have its own page in `master`, but the source file has data as if it were a page. Did we have one that broke in the past, or was that data just left over?

Either way, that's another design decision- we can either have a glossary page or only use the data for tooltips.

shcheklein:
Excellent stuff @rogermparent !

> Did we have one that broke in the past, or was that data just left over?

We never had a stand-alone page for it - see #542

> Primarily, we need to figure out what directory to use for glossary entries and where we want the glossary entries to show up

Good question. My thoughts on this:

Let's move them into `user-guide/basic-concepts/*` (keep them hidden for now).

We can make it the first strep of solving the #947 and #542 - we'll have a section under the user guide with multiple pages, each page is a MD file with frontmatter that defines tooltip + some longer description that describes the concept in details. 

Start expanding the concept descriptions as part of the #947 - btw, could be a good GSoD task /cc @jorgeorpinel

@rogermparent @jorgeorpinel WDYT?

rogermparent:
> @rogermparent @jorgeorpinel WDYT?

I think that's an even better solution than I could think of! All of what you said is absolutely possible: I can move the glossary entries to `user-guide/basic-concepts` and suppress their page creation, then solve the related issue specifically. 

Once we get to a point where this branch solves the issue and goes to master, we can tackle #947 and #542 using the new master as a base and re-enable page creation in the branches that need it.

I think the best route for the data to do this is implement some sort of system that lets us access a short and long version of each Basic Concept's description. This is usually done with a frontmatter field, a special page break in the body, or a combination of both methods so we have the option for whichever suits each entry. 
 
Essentially, the difference is that *the frontmatter method* is good if we want to re-word the short description and *the page break method* is good for easily slicing off the first paragraph or two to use as a tooltip while leaving the rest for the full page if users want more elaboration.

We can link the Basic Concept's full page as a "Read More"-type link on Tooltips as well. This is possible with pretty much any implementation, really.

I'll get started on that part that solves this issue, and we can move from there!

src/gatsby/models/json-files/index.js
```diff
@@ -1,13 +1,28 @@
 const callOnModels = require('../../utils/models')
+const yaml = require('js-yaml')
+
+async function parseContentToJson({ node, loadNodeContent }) {
+  switch (node.extension) {
+    case 'json':
+      return JSON.parse(await loadNodeContent(node))
+
+    case 'yml':
```
shcheklein, line 9: probably we don't need this for now - let's simplify the logic back?

rogermparent: Absolutely! I forgot about that bit when removing the rest.

rogermparent:
I need to bring that Gatsby child node factory I made in the Community Model into this one and the others- I didn't think of it until the codeclimate complaint.

Not necessarily related to this issue, but I think most everything's taken care of so I'll start on this branch and make a new PR if we merge this PR before I'm done with the refactor.

content/docs/user-guide/basic-concepts/import-stage.md
```diff
@@ -0,0 +1,7 @@
+---
+name: 'Import Stage'
+match: ['import stage', 'import stages']
+---
+
+Stage (DVC-file) created with the "dvc import" or "dvc import-url" commands.
```
shcheklein, line 6: commands should be in single quotes

content/docs/user-guide/basic-concepts/dependency.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Dependency
+match: [dependency, dependencies]
+---
+
+A file or directory (possibly tracked by DVC) recorded in the "deps" section of
```
shcheklein, line 6: looks like in most cases we want single quotes not double - for inline code blocks

package.json
```diff
@@ -70,11 +70,11 @@
     "react-dom": "^16.13.1",
     "react-ga": "^2.7.0",
     "react-helmet": "^5.2.1",
-    "react-markdown": "^4.3.1",
     "react-popover": "^0.5.10",
     "react-slick": "^0.25.2",
     "react-use": "^14.0.0",
     "rehype-react": "^5.0.1",
+    "remark-rehype": "^7.0.0",
```
shcheklein, line 77: do we actually use this now?

content/docs/user-guide/basic-concepts/import-stage.md
rogermparent: Got it!

package.json
rogermparent: It seems that we don't, nice catch!

content/docs/user-guide/basic-concepts/data-artifact.md
```diff
@@ -0,0 +1,8 @@
+---
+name: 'Data Artifact'
+match: ['data artifact', 'data artifacts']
+---
+
+Any data file or directory, as well as intermediate or final result that is
+tracked by DVC, for example by using 'dvc add'. See
```
shcheklein, line 7: My bad Roger - it's not even single quotes, but backtick ` ... line inline-code block in Markdown.

rogermparent: Ohhh, I see! I'll fix that up tomorrow while checking over all the changes again myself.

shcheklein:
@rogermparent it has not been redeployed for some reason?

rogermparent:
@shcheklein That's really weird, I'm not sure what happened there but I triggered a redeploy on Heroku and that fixed it for me. It deployed over the same URL at https://dvc-landing-build-time--fnpeae.herokuapp.com, but GitHub still marks the automatic deploy as inactive.

rogermparent:
It works on the latest auto-deploy! Might've been some sort of caching issue, but unless it shows up again I'm willing to write it off as a fluke.

shcheklein:
@rogermparent thanks! awesome stuff!

content/docs/glossary.js
```diff
@@ -1,89 +0,0 @@
...
-        'DVC repositories'
-      ],
-      desc: `
-Initialized by running \`dvc init\` in the **workspace** (typically in a Git
-repository). It will contain the
-[\`.dvc/\` directory](/doc/user-guide/dvc-files-and-directories) and
-[DVC-files](/doc/user-guide/dvc-file-format) created with commands such as
-\`dvc add\` or \`dvc run\`.
-      `
-    },
```
jorgeorpinel, line 32: I think we missed migrating this one.

image

image

rogermparent: Whoops! Thanks for the catch!

jorgeorpinel: Fixing it in #1358 🙂

rogermparent: Great, thanks!

PR #1341, opened 2020-05-23
Add optional link field to authors

### LINKED ISSUE #1161: blog: post authors twitter account
I've noticed that after moving out of Medium, fewer people start following me on Medium and Twitter.
Medium followers - very expected. Twitter - not.

We should drive more followers to the author's profiles and @dvcorg profile. Also, we can deploy the email subscription somewhere (it might be not enough to have it on the very bottom of the page).

It can be done by incorporation of author twitter accounts next to author avatar/image or have it on the author page.

@andronovhopf what do you think?

PS: It might be related to #1120

rogermparent:
This is a quick addition to allow for optionally linking arbitrary pages to authors, making them clickable on Blog Posts.

Fixes #1161, and is an interim fix while we make something more elaborate in #1120

Just add `link: https://mysocialmedia.com/profile` to an author's Markdown frontmatter and that Author's name will become a link in Blog posts! I haven't committed any content changes yet, so we'll have to do that to have an example on display.

We'll probably need to add docs for this somewhere as well.

### DIFF
--- a/content/authors/dmitry_petrov.md
+++ b/content/authors/dmitry_petrov.md
@@ -1,6 +1,7 @@
 ---
 name: Dmitry Petrov
 avatar: dmitry_petrov.png
+link: https://twitter.com/fullstackml
 ---
 
 Creator of [http://dvc.org](http://dvc.org) — Git for ML. Ex-Data Scientist
--- a/content/authors/elle_obrien.md
+++ b/content/authors/elle_obrien.md
@@ -1,6 +1,7 @@
 ---
 name: Elle O'Brien
 avatar: elle_obrien.jpg
+link: https://twitter.com/andronovhopf
 ---
 
 Data scientist at [http://dvc.org](http://dvc.org)
--- a/content/authors/george_vyshnya.md
+++ b/content/authors/george_vyshnya.md
@@ -1,6 +1,7 @@
 ---
 name: George Vyshnya
 avatar: george_vyshnya.jpeg
+link: https://www.linkedin.com/in/gvyshnya
 ---
 
 Seasoned Data Scientist / Software Developer with blended experience in software
--- a/content/authors/jorge_orpinel.md
+++ b/content/authors/jorge_orpinel.md
@@ -1,6 +1,7 @@
 ---
 name: Jorge Orpinel Pérez
 avatar: jorge.jpg
+https://www.linkedin.com/in/jorgeorpinel
 ---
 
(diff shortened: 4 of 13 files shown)


rogermparent:
I added @dmpetrov's Twitter as data to show off the feature!
[Here's a convenient link to the preview](https://dvc-landing-author-link-pfszqj.herokuapp.com/blog/dvc-3-years-and-1-0-release)

shcheklein:
@rogermparent looks cool! Let's add same links for other authors if it's possible to find them?

rogermparent:
> @rogermparent looks cool! Let's add same links for other authors if it's possible to find them?

I'm willing to do all of the existing ones before we merge this, but do you happen to know the best way to reach every who's interested? Some sort of meeting or group ping? That might be the most efficient way to go about things. Either way, I'll hunt down what I can. 

There aren't many authors so it probably won't be an issue even if I hunt them down manually. It's possible we may want to confirm if everyone wants their name linked to their twitter before we do so, but I'll trust the rest of the team to make the best decision on that.

shcheklein:
@rogermparent no need to spend too much time on it. Just do the best you can with simple Google Search and pick Twitter or LinkedIn whichever looks more active. Also, some of those people have Medium and on it they link to a social account usually.

Again, if you don't see it immediately just mention it in a comment.

shcheklein:
That's great @rogermparent . Not sure it completely fixes #1161 since ideally we need to do some design ui/ux for that. May be we can only single #1120 for that though. I was thinking also that it would be great to redesign and probably add some tooltip, or social network icons close to the author.

rogermparent:
> I was thinking also that it would be great to redesign and probably add some tooltip, or social network icons close to the author.

Absolutely! We could have the redesign of the blog post page and the design of the new author page be two separate issues, with the former being smaller and quicker to do.

A schema addition that comes to mind is to turn this Author `link` frontmatter field into `links`, which has the shape of a list describing external links. These links can be written in frontmatter whatever way we'd like, but I'd suggest using a mixed `Array<object|string>` solution where we use strings like `twitter.com/profile` that are recognized and parsed into objects like `{type: 'twitter', handle: 'profile'}` while optionally allowing authors to directly supply an object for cases where we want something like custom metadata or descriptions. #1120's author page could then use that data for its template, and the Blog page could use the data to make social link icons when relevant.

I'll draft up some issues for this tomorrow after I gather my thoughts, it's a way too late for me to waterfall right now.

PR #1348, opened 2020-05-24
Change jorge's Author link

rogermparent:


rogermparent:
No problem! I greatly empathize with the feeling of not wanting particular social media shared.

I was also able to make the change in the web UI so it was _really_ easy.

Issue #510, opened 2019-07-31
glossary: tooltips break details sections


jorgeorpinel:
As a result of implementing #424, `<abbr>` tags now create glossary tooltips around certain terms in the docs.

However if one of these terms is placed inside a `<details>` section, the docs engine fails to create the expandable boxes (a pre-existing feature).

---
UPDATE: The example below is outdated but you get the idea hopefully.

For example open `static/docs/user-guide/external-dependencies.md` and change

```md
### Expand to learn more about DVC internals

The resulting "import stage" DVC-file will contain something like this:
```

which looks

![image](https://user-images.githubusercontent.com/1477535/62177182-3f4f1f80-b301-11e9-9fb3-5da46710726a.png)

for

```md
### Expand to learn more about DVC internals

The resulting <abbr>import stage</abbr> DVC-file will contain something like
this:
```

resulting in:

![image](https://user-images.githubusercontent.com/1477535/62177236-6f96be00-b301-11e9-94fb-74eba6e3927c.png)

jorgeorpinel:
@algomaster99 do you want to take a look at this? 🙂

algomaster99:
@jorgeorpinel This is happening since it only detects the `abbr` tags and just ignores the `details` tag. Seems weird to me. I'll have to check the inner functioning of HtmlRenderer.

jorgeorpinel:
Any luck  with this first glossary-related issue, @algomaster99 ? Thanks 🙂

jorgeorpinel:
- [ ] When fixed, we should add `<abbr>` to term "project" in the first `<details>` section of static/docs/get-started/configure.md, for now removed in ce3d940

jorgeorpinel:
"Remember me @algomaster99" -This issue 🙂

algomaster99:
@jorgeorpinel I think this would require some changes in `react-markdown` itself. I found one issue which is relevant - https://github.com/rexxars/react-markdown/issues/302

fabiosantoscode:
I've opened this PR there which fixes this problem: https://github.com/rexxars/react-markdown/pull/386

shcheklein:
@fabiosantoscode awesome, thanks.

rogermparent:
With #1334, this seems to be fixed for me:
![dvc-abbr-desktop](https://user-images.githubusercontent.com/9111807/82840987-89158180-9ea2-11ea-8820-37e1fddc2bb8.png)
[...]

PR #1357, opened 2020-05-25
Small null check fix to abbrs inside details blocks

rogermparent:
Docs `abbr` tags inside `details` breaks on responsive view simulated touch for me because e.changedTouches[0] is null. This null check fixes that.

Related to #510

### DIFF
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -146,7 +146,9 @@ const Markdown: React.FC<IMarkdownProps> = ({
     touchstartXRef.current = e.changedTouches[0].screenX
   }, [])
   const onTouchEnd = useCallback(e => {
-    touchendXRef.current = e.changedTouches[0].screenX
+    const changedTouch = e.changedTouches[0]
+    if (!changedTouch) return
+    touchendXRef.current = changedTouch.screenX
     handleSwipeGesture()
   }, [])
 


src/components/Documentation/Markdown/index.tsx
```diff
@@ -146,7 +146,9 @@ const Markdown: React.FC<IMarkdownProps> = ({
     touchstartXRef.current = e.changedTouches[0].screenX
   }, [])
   const onTouchEnd = useCallback(e => {
-    touchendXRef.current = e.changedTouches[0].screenX
+    const changedTouch = e.changedTouches[0]
```
shcheklein, line 149: hmm .. 🤔 interesting why can it be `null`?

rogermparent: I'm not sure why exactly, but the page threw when I tried clicking a details element. I would assume this check can also be done with an early return on `e.changedTouches.length === 0`.

shcheklein: how did you emulate it? what OS/Browser?

rogermparent: `gatsby develop` and Firefox 76.0.1 Responsive Design Mode

shcheklein: Just tried Chrome and FF the master - can't reproduce it still. What is the specific action do we need for this?

rogermparent: Odd, I'll try again and get more specific details on it.

rogermparent: I keep getting the error on master anywhere on the mobile view with Firefox's touch simulation enabled- I don't get the issue with the touch simulation feature off.

For what it's worth, [every example of the touch APIs on MDN iterate over `changedTouches` property as a list](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#Handling_the_end_of_a_touch), implicitly handling the case of no `changedTouches` where our code throws. I believe we're currently using the API incorrectly by failing in the case of no `changedTouches` in the `TouchList`.

shcheklein: I still don't see why do we break the API tbh. Also, does it mean that the logic that depends on `touchendXRef` does not work on a mobile Firefox (I don't have it to try, unfortunately - actual one, not emulator).

rogermparent: The API examples indicate that an empty `changedTouches` isn't necessarily some sort of malformed input that should be thrown on. Our current code just assumes something will always be in the `changedTouches` list and starts operating on `changedTouches[0]` even if nothing's there, which is unlike any of the examples I can find.

[Looks like empty `changedTouches` is a known firefox Responsive Display Mode issue, though](https://bugzilla.mozilla.org/show_bug.cgi?id=1615824) (Fixed in 77, I have 76- ha!), so we probably don't have to worry about Firefox Mobile being broken.

I'll update my browser and see if that fixes it. What version of Firefox are you using, @shcheklein? If it's 77+, that explains why you can't reproduce the issue.

shcheklein: No, mine is 76 and I was able to reproduce it after you provided additional details. It seemed to me still that it is a FF issue, not API. The examples you brought for a reference are good but could not be considered an API spec to my mind. They make sense since in all of them you deal with multiple finger touches (not sure how would it work in our case), but they never specify explicitly that array can be empty.

So, can we close this then? :)

rogermparent: Sure enough, I install my repo's Firefox Dev Edition which is on 77 and stopped having the issue! That's pretty funny that I ran into it just before fix.

I suppose we don't necessarily have to add this null check. I can't say if FF<77 RDM was the only touch handler that has this issue, but it was probably the only notably large one.

rogermparent:
Closed for being a Firefox RDM issue more than our usage of the API, and it's even fixed in the next FF version.

PR #1359, opened 2020-05-25
Re-add abbrs within details

### LINKED ISSUE #510: glossary: tooltips break details sections

rogermparent:
Having `abbr`s in `details` used to break them, leading to #510. 

At some point since then, the issue was fixed and `abbr` works within `details` with no problem, even in quotes.

This branch re-adds the `abbr`s removed in #510.
I think this also fixes #510, unless I missed something from the list or there are other docs to re-introduce this to.

Uses the list from [this comment](https://github.com/iterative/dvc.org/issues/510#issuecomment-520898641)

* [x]  We should add `<abbr>` to term "project" in the first `<details>` section of static/docs/get-started/configure.md, for now removed in [ce3d940](https://github.com/iterative/dvc.org/commit/ce3d940031e10a366298a24870557a5194af3d98).
* [x]  Re-add `<abbr>` in tags static/docs/get-started/add-files.md removed temporarily in [ed4062e](https://github.com/iterative/dvc.org/commit/ed4062eba4d09ec2e98abc3e8f2ef3ae808cd4f4).
* [x]  Undo also [6d642ef](https://github.com/iterative/dvc.org/commit/6d642ef4b8b7b90b36266a1aad19aa573de9c05d).
* [x]  And [d28add6](https://github.com/iterative/dvc.org/commit/d28add6869e29168a1175cf941f68bf18781056a)

### DIFF
--- a/content/docs/command-reference/import-url.md
+++ b/content/docs/command-reference/import-url.md
@@ -146,7 +146,7 @@ $ git checkout 2-remote
 $ mkdir data
 ```
 
-You should now have a blank workspace, just before the
+You should now have a blank <abbr>workspace</abbr>, just before the
 [Add Files](/doc/tutorials/get-started/add-files) chapter.
 
 </details>
--- a/content/docs/command-reference/remote/add.md
+++ b/content/docs/command-reference/remote/add.md
@@ -355,9 +355,10 @@ $ dvc remote add myremote https://example.com/path/to/dir
 A "local remote" is a directory in the machine's file system.
 
 > While the term may seem contradictory, it doesn't have to be. The "local" part
-> refers to the machine where the project is stored, so it can be any directory
-> accessible to the same system. The "remote" part refers specifically to the
-> project/repository itself. Read "local, but external" storage.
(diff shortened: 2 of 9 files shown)


rogermparent:
> Woot woot thanks for fixing this old one @rogermparent! It's good not to have this limitation any more.

Thanks!

> That commit is old but if you look for that note on local remotes throughout the docs now, you may find it in more places than the ones modified in there.

Thanks for pointing that out! I'll recheck it.

shcheklein:
Thanks @rogermparent 👍

PR #1361, opened 2020-05-26
first draft of May gems

andronovhopf:
review requested- is everything accurate? Keyword suggestions for tags? And how is length?


content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+  and using DVC with teams.
+descriptionLong: |
+  A roundup of technical Q&A's from the DVC community. This month, we discuss 
+  development best practices, sharing models and data across projects, 
+  and using DVC with teams.
+picture: 2020-05-26/May_20_Gems_Header.png
+author: elle_obrien
+commentsUrl: https://discuss.dvc.org/t/april-20-community-gems/356
+tags:
+  - Discord
```
shcheklein, line 16: do we use both - Discord and Gems everywhere?

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+  A roundup of technical Q&A's from the DVC community. This month, we discuss 
+  development best practices, sharing models and data across projects, 
+  and using DVC with teams.
+descriptionLong: |
+  A roundup of technical Q&A's from the DVC community. This month, we discuss 
+  development best practices, sharing models and data across projects, 
+  and using DVC with teams.
+picture: 2020-05-26/May_20_Gems_Header.png
+author: elle_obrien
+commentsUrl: https://discuss.dvc.org/t/april-20-community-gems/356
```
shcheklein, line 14: don't forget to change the URL before merge

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+  - Cache
+  - Google Cloud Storage
+  - Import
+---
+
+## Discord gems
+
+Here are some Q&A's from our Discord channel that we think are worth sharing.
+
+### Q: [How do I completely delete a file from DVC?](https://discord.com/channels/485586884165107732/563406153334128681/710546561498873886)
```
shcheklein, line 27: it's a good one and very valuable! there are some caveats to this, though:

1. removing a file from dir - not as simple, it would be nice to mention or even cover this as well to make the A especially useful.
2. when we remove manually a file from cache there is a risk that it was referenced from another commit (possibly even under a different name). `gc --all-commits` handles this safely.

let me know if you need my help with this one - we can discuss the details.

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+to remove a single `.dvc` file without doing a cache cleanup, look into the
+`.dvc` file and note the `md5` field inside. Then use this value to identify the
+corresponding file in your DVC cache and delete it. For example: if your target
+file has `md5: 123456`, the corresponding file in your cache will be
+`.dvc/cache/12/3456`.
+
+In the future, we plan to support a function like `git rm` that will allow for
+more granular delete options.
+
+### Q: [I'm using an S3 bucket as a DVC remote; the bucket is shared with several coworkers, who have projects in other folders in the bucket. Will it disrupt DVC if I manually add a README inside the folder being used for remote storage, to let my coworkers know what the directory is used for?](https://discord.com/channels/485586884165107732/563406153334128681/707551737745244230https://discord.com/channels/485586884165107732/563406153334128681/707551737745244230)
```
shcheklein, line 41: may be we can generalize it a bit - Is it safe to add custom file to remote storage? or something like this? And put the long version at the beginning of the answer (like E.g. I want to add REAME to describe the content ...)

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+We ourselves have stored datasets containing up to 2 million files, so 100,000
+is certainly feasible. Of course, the larger your dataset, the more time data
+transfer operations will take. Luckily, we have a
+[DVC 1.0 contains several data transfer optimizations](https://dvc.org/blog/dvc-3-years-and-1-0-release#data-transfer-optimizations)
+to substantially reduce the time needed to `dvc pull / push / status -c / gc -c`
+for very large datasets.
+
+### Q: [Two developers on my team are doing `dvc push` to the same remote. Should they `dvc pull` first?](https://discord.com/channels/485586884165107732/563406153334128681/704211629075857468)
+
+We encourage users to think of a DVC project like using Git- the same workflow
```
shcheklein, line 58: I think we can keep the simplest version here- it's safe to push simultaneously, any files, no pull is needed. Also, in this particular case I would be very careful comparing DVC to Git. When it comes to data management DVC broke quite a few conventions comparing to Git - e.g.  it does not pull the whole history by default, it provides quite explicit interface to pull a specific file/directory. All of this because pulling can be quite expensive.

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+make sure a developer's local workspace is synced with the master branch of a
+project before pushing changes.
+
+With that said, because DVC remotes stores files indexed by `md5`s, there's
+usually a very low probability of a collision (if two developers have two
+different versions of a file, they'll be stored as two separate files in the DVC
+remote- so no merge conflicts). Still, being in the habit of frequent pulling is
+a good practice.
+
+### Q: [What are `*.tmp` files in my DVC remote?](https://discord.com/channels/485586884165107732/563406153334128681/698163554095857745)
```
shcheklein, line 71: Caveat - make sure that they don't belong to some actively running `dvc push`

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+
+Inside your DVC remote, you might see `.tmp` files from incomplete uploads. This
+can happen if a user killed a process like `dvc push`. You can safely remove
+them; for example, if you're using an S3 bucket, `aws s3 rm ... *.tmp` will do
+the trick.
+
+### Q: [I'm using Google Cloud Storage as a DVC remote. I'm getting an error: `ERROR: unexpected error - ('invalid_grant: Bad Request', '{\n "error": "invalid_grant",\n "error_description": "Bad Request"\n}')`. Any ideas?](https://discord.com/channels/485586884165107732/485596304961962003/705131622537756702)
```
shcheklein, line 78: can we simplify the error message a bit (potentially move json into code block at the beginning of the A):

E.g. if you are getting:

```json
---> error JSON comes here
```

it usually means a permission ...

shcheklein: I think we should clarify and expand the answer a little bit here . First I would check the `dvc remote add` and `dvc remote modify` docs - they should have some links to external docs on how to setup Google Cloud CLI and authenticate it. We can at least put those link. @efiop can potentially help with some default flow and/or some external links that describe it.

content/blog/2020-05-26-may-20-community-gems.md
```diff
@@ -0,0 +1,93 @@
...
+
+This is a permissions error for sure! This can happen if DVC can't find `.json`
+credentials file for your GCP bucket. Try authenticating using
+`gcloud beta auth application-default login`.
+
+### Q: [I'm working on several projects that all need involve the same saved model. One project trains a model and pushes it to cloud storage with `dvc push`, and another takes the model out of cloud storage for use. What's the best practice for doing this with DVC?](https://discord.com/channels/485586884165107732/485596304961962003/708318821253120040)
```
shcheklein, line 84: We can put an image from the data registry case here? just to entertain readers a bit, wdyt?

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: yep, we've been using both. want to change to just one?

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: yeah that's a great idea!

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: changed to: 
```
### Q: [Is it safe to add a custom file to my DVC remote?](https://discord.com/channels/485586884165107732/563406153334128681/707551737745244230https://discord.com/channels/485586884165107732/563406153334128681/707551737745244230)

Definitely. Some people add additional files to their DVC remote, like a README to explain to teammates what the folder is being used for. 
Having an additional file in the remote that isn't part of
DVC tracking won't pose any issues. You would only encounter problems if you
were manually modifying or deleting contents of the remote managed by DVC.
```

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: Simplified to: 
```
It's safe to push simultaneously, no `dvc pull` needed. While some teams might be in the habit of frequently pulling, like in Git flow, there are less risks of "merge conflicts" in DVC.  That's because DVC remotes store files indexed by `md5`s, so there's
a very low probability of a collision (if two developers have two
different versions of a file, they'll be stored as two separate files in the DVC
remote- so no merge conflicts). 
```

if you still think it's too much, I can simplify to just the first sentence or so.

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: i've expanded the answer, but i'm not positive it's 100% accurate. I don't think DVC tracks the `.json` containing a user's credentials, it sounds like that's managed by GCP CLI. Let me know what you think in the next commit

content/blog/2020-05-26-may-20-community-gems.md
andronovhopf: i think i need a little help understanding the first point- why is removing a `.dvc` file from a directory not so simple? the second point seems reasonable

content/blog/2020-05-26-may-20-community-gems.md
shcheklein: it's totally fine, just to confirm

content/blog/2020-05-26-may-20-community-gems.md
shcheklein: so, regarding the first point:

There is a level (file) of indirection in the way DVC stores directories. It solves two problems (that you would immediately hit when you implement DVC itself):

1. For a file we have a simple one-to-one relationship - file to its hash (`md5` in our case). We need this hash to find and save it in DVC cache/remote. We need it to being able to check if we need to run a stage again (if there are some changes in a dependency), etc. DVC saves hash value in its YAML files a one line: `md5: 12324dedefdsfd`. But what about dirs? How do we get a simple _hash_ for the whole directory? We can't store 1M md5s into a `.dvc` file.

2. Directories are hierarchical - they have subdirectories, they have a lot of files with different names inside. How do we capture all these file names? All this complex structure? Again, we can't save it into a `.dvc` file - too much info for a simple file, it can be megabytes in size, etc.

So, to address those two problems, in DVC we store a special JSON file `122sdsd323e42.dir` with a content of the directory. You can read a bit about this here  - https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory .

It means that if I want to remove `data/path/file-a.txt`, I would need to find it's cache first by parsing the `.dir` files that corresponds to the `data`.

Please, let me know if it's more or less makes sense :)

andronovhopf: oh, i see- if the file was in a directory that had been DVC tracked at a directory level. Got it, thanks! I think i can answer it now

andronovhopf: @shcheklein just gave writing the answer a try. let me know what you think!

andronovhopf:
the circle-ci commit check seems to be failing; possibly due to something about the new image. Did anything change about the way we address/format filepaths to images?

shcheklein:
> the circle-ci commit check seems to be failing; possibly due to something about the new image. Did anything change about the way we address/format filepaths to images?

no, disregard this please. We'll try to fix this in #1000

content/blog/2020-05-26-may-20-community-gems.md
shcheklein: It's close! :) One correction - when `.dir` is located we can't edit this file (it's name is md5 of the content and we need to preserve it). The only things we can do is to find an entry inside, that entry will have md5 on its own and then delete file in the way described above. The caveat here is that DVC will rely on the information in `.dir` in certain cases. E.g. it will never push a file to remote storage again.

Btw, we don't mention remote storage here at all - it's another aspect of getting rid of data in DVC.

andronovhopf: aha! ok, fixed that and pushed :D

shcheklein:
Looks good,  ping me when you want this to be merged

andronovhopf:
@shcheklein can we merge now and i'll publish on social media tomorrow am?

shcheklein:
@andronovhopf done!

PR #1363, opened 2020-05-27
Overhaul how GitHub stars are handled

### LINKED ISSUE #1019: landing page: proxying and handling Github stars fetch requests
We have stars counter for the `dvc` repo on the title page. And it is not always works:

https://sentry.io/organizations/iterative/issues/1510361492/

Also the it not works we didn't handle rejections in GithubLine:

https://sentry.io/organizations/iterative/issues/1519509024/

Maybe we should start proxying and caching GR request on our and as we do now for community page. And also we should add `catch` to request.

rogermparent:
- Uses `gatsby-source-github-api` to get GitHub data at build time

- Splits the GitHub Issues proxy API into one for Issues and Stars, sharing the
  query data. The new endpoint is `/api/github/stars`.

- Updates the GithubLine component to display the build-time value at first, then update
  via the proxy as soon as possible to reflect any changes. If the value hasn't 
  changed or there was an error, the component keeps the static value.

Fixes #1019

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,5 +1,6 @@
 /* eslint-env node */
 
+require('dotenv').config()
 const path = require('path')
 
 require('./config/prismjs/dvc')
@@ -39,6 +40,25 @@ const plugins = [
     options: {
       name: 'content',
       path: path.join(__dirname, 'content')
+    }
+  },
+  {
+    resolve: `gatsby-source-github-api`,
+    options: {
+      // token: required by the GitHub API
+      token: process.env.GITHUB_TOKEN,
+
+      // GraphQLquery: defaults to a search query
+      graphQLQuery: `
+          {
+            repository(owner: "iterative", name: "dvc") {
+              stargazers {
+                totalCount
+              }
+            }
+          }
+      `,
+      variables: {}
     }
   },
   {
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -43,25 +43,6 @@ const plugins = [
     }
   },
   {
-    resolve: `gatsby-source-github-api`,
-    options: {
-      // token: required by the GitHub API
-      token: process.env.GITHUB_TOKEN,
-
-      // GraphQLquery: defaults to a search query
-      graphQLQuery: `
-          {
-            repository(owner: "iterative", name: "dvc") {
-              stargazers {
-                totalCount
-              }
-            }
-          }
-      `,
-      variables: {}
-    }
-  },
-  {
     resolve: 'gatsby-source-filesystem',
     options: {
       name: 'images',
@@ -218,6 +199,28 @@ const plugins = [
     }
   }
 ]
+
+if (process.env.GITHUB_TOKEN) {
+  plugins.push({
+    resolve: `gatsby-source-github-api`,
+    options: {
+      // token: required by the GitHub API
+      token: process.env.GITHUB_TOKEN,
+
+      // GraphQLquery: defaults to a search query
+      graphQLQuery: `
+          {
+            repository(owner: "iterative", name: "dvc") {
+              stargazers {
+                totalCount
+              }
+            }
+          }
+        `,
+      variables: {}
+    }
+  })
+}
 
 if (process.env.CONTEXT === 'production') {
   plugins.push({
--- a/src/gatsby/hooks/stars.ts
+++ b/src/gatsby/hooks/stars.ts
@@ -18,7 +18,7 @@ export default function useStars(): number {
   // Run an IIFE to update from the server on the client side.
   useEffect(() => {
     ;(async (): Promise<void> => {
-      const res = await fetch(`/api/github/stars?current=${stars}`)
+      const res = await fetch(`/api/github/stars`)
       if (res.status === 200) {
         const json = await res.json()
         setStars(json.stars)
--- a/src/gatsby/hooks/stars.ts
+++ b/src/gatsby/hooks/stars.ts
@@ -23,7 +23,7 @@ export default function useStars(): number {
 
   // Run an IIFE to update from the server on the client side.
   useEffect(() => {
-    ;(async (): undefined => {
+    ;(async (): Promise<void> => {
       const res = await fetch(`/api/github/stars?current=${stars}`)
       if (res.status === 200) {
         const json = await res.json()
--- a/src/gatsby/hooks/stars.ts
+++ b/src/gatsby/hooks/stars.ts
@@ -1,0 +1,36 @@
+import { useEffect, useState } from 'react'
+import { useStaticQuery, graphql } from 'gatsby'
+import fetch from 'isomorphic-fetch'
+
+export default function useStars(): number {
+  // Get the amount of stars from build time
+  const staticStars = useStaticQuery(graphql`
+    query GithubStarsQuery {
+      githubData {
+        data {
+          repository {
+            stargazers {
+              totalCount
+            }
+          }
+        }
+      }
+    }
+  `).githubData.data.repository.stargazers.totalCount
+
+  // Maintain an updatable state so we can update stars on delivery
+  const [stars, setStars] = useState(staticStars)
+
+  // Run an IIFE to update from the server on the client side.
+  useEffect(() => {
+    ;(async (): undefined => {
+      const res = await fetch(`/api/github/stars?current=${stars}`)
+      if (res.status === 200) {
+        const json = await res.json()
+        setStars(json.stars)
+      }
+    })()
+  }, [])
+
+  return stars
+}
--- a/src/gatsby/hooks/stars.ts
+++ b/src/gatsby/hooks/stars.ts
@@ -22,6 +22,10 @@ export default function useStars(): number {
       if (res.status === 200) {
         const json = await res.json()
         setStars(json.stars)
+      } else {
+        console.log(
+          `Stars update response status was ${res.status}! Skipping update.`
+        )
       }
     })()
   }, [])
--- a/src/gatsby/hooks/stars.ts
+++ b/src/gatsby/hooks/stars.ts
@@ -6,17 +6,11 @@ export default function useStars(): number {
   // Get the amount of stars from build time
   const staticStars = useStaticQuery(graphql`
     query GithubStarsQuery {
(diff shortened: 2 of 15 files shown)


gatsby-config.js
```diff
@@ -41,6 +42,25 @@ const plugins = [
       path: path.join(__dirname, 'content')
     }
   },
+  {
+    resolve: `gatsby-source-github-api`,
+    options: {
+      // token: required by the GitHub API
+      token: process.env.GITHUB_TOKEN,
```
shcheklein, line 49: is it build time? what happens if it's not defined? (like local dev machine)

rogermparent: It is build time, and currently the site throws when the token is removed. I can fix it such that this doesn't happen and build without the token present just use dummy data, it'll just take a little extra logic to handle that.

src/gatsby/hooks/stars.ts
```diff
@@ -0,0 +1,36 @@
...
+  `).githubData.data.repository.stargazers.totalCount
+
+  // Maintain an updatable state so we can update stars on delivery
+  const [stars, setStars] = useState(staticStars)
+
+  // Run an IIFE to update from the server on the client side.
+  useEffect(() => {
+    ;(async (): Promise<void> => {
+      const res = await fetch(`/api/github/stars?current=${stars}`)
+      if (res.status === 200) {
```
shcheklein, line 28: woud be great to have a sentry notification if APIs fail

src/gatsby/hooks/stars.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  `).githubData.data.repository.stargazers.totalCount
+
+  // Maintain an updatable state so we can update stars on delivery
+  const [stars, setStars] = useState(staticStars)
+
+  // Run an IIFE to update from the server on the client side.
+  useEffect(() => {
+    ;(async (): Promise<void> => {
+      const res = await fetch(`/api/github/stars?current=${stars}`)
```
shcheklein, line 27: since we include query - need to make sure that cache is handled properly

src/server/index.js
```diff
@@ -19,6 +19,7 @@ const express = require('express')
 const compression = require('compression')
 const { s3Url } = require('./config')
 const { isProduction } = require('./utils')
+require('dotenv').config()
```
shcheklein, line 22: ? do we need this?

src/server/middleware/api/index.js
```diff
@@ -2,10 +2,11 @@ const routes = require('express').Router()
 
 const comments = require('./comments')
 const discourse = require('./discourse')
-const github = require('./github')
+const { issues, stars } = require('./github')
 
 routes.get('/comments', comments)
 routes.get('/discourse', discourse)
-routes.get('/github', github)
+routes.get('/github', issues)
```
shcheklein, line 9: do we want to change the route to `/github/issues`?

src/server/middleware/api/github.js
```diff
@@ -5,61 +5,113 @@ const { isProduction } = require('../../utils')
...
+    return { error: e.message || e }
+  }
+}
+
+async function issues(_, res) {
   if (!process.env.GITHUB_TOKEN) {
-    res.status(200).json({ issues: [] })
+    res
+      .status(200)
+      .json({ issues: [], error: 'No GITHUB_TOKEN specified on the server!' })
```
shcheklein, line 65: we don't want to expose implementation details to public

src/server/middleware/api/github.js
```diff
@@ -5,61 +5,113 @@ const { isProduction } = require('../../utils')
...
-      res.status(200).json({ issues })
-    } catch (e) {
+async function stars(req, res) {
+  if (!process.env.GITHUB_TOKEN) {
+    res.status(403).json({ error: 'No GITHUB_TOKEN specified on the server!' })
+  } else {
+    const stars = await getStars()
+    if (stars) {
+      // If the data in unchanged, return a minimal response indicating so.
+      if (Number(req.query.current) === stars) {
```
shcheklein, line 102: feels like a premature optimization ... we save like a few bytes I suppose?

src/gatsby/hooks/stars.ts
shcheklein: figured this out ... but I think we don't need this at the moment? we save a few bytes but complicate code

src/server/middleware/api/github.js
```diff
@@ -5,61 +5,113 @@ const { isProduction } = require('../../utils')
...
+    const { stars } = await getFreshGithubData()
+    return stars
+  }
+}
 
-      res.status(200).json({ issues })
-    } catch (e) {
+async function stars(req, res) {
+  if (!process.env.GITHUB_TOKEN) {
+    res.status(403).json({ error: 'No GITHUB_TOKEN specified on the server!' })
```
shcheklein, line 97: 403 here, was 200 for issues? also the same small concern - it's better to keep impl details private (I mean error message)

src/server/middleware/api/index.js
rogermparent: I kept it the same for backward compatibility reasons, but it would make sense to change it and I can do it.

src/server/middleware/api/github.js
rogermparent: Fair, I'll remove it

src/server/middleware/api/github.js
rogermparent: I suppose you're right for stars. Something like this would serve better in an API passes around more data.

src/server/middleware/api/github.js
rogermparent: I can change it back to the previous behavior if that's what we want. I made the fail states more explicit because I was thrown off at the start of the PR when diagnosing why I got a successful result with no issues nor indication anything went wrong.

shcheklein: I don't have any specific suggestions :) just some concerns - exposing impl details probably not good and errors codes probably should be consistent. Probably, we can use logs? And 403 itself sounds good if it helps to debug.

src/gatsby/hooks/stars.ts
rogermparent: I can remove the cache checking, if that's what you're talking about, but this IIFE in general is what fetches the data that updates stars if we got more since last build.

src/server/middleware/api/github.js
rogermparent: Sure! I chose 403 forbidden because we're giving an error for the lack of an API key. That code alone with no message should be expressive enough to tell devs what's going on without actually exposing implementation details.

src/server/index.js
rogermparent: We need it to use `dotenv` with the server, which helps a lot in setting up the required environment variables like the GitHub token in a dev environment without having to manually re-add the API key to the command invocation each time.

As far as if it's necessary to the functionality, it isn't, it's just a harmless but big improvement to dev QOL.

src/gatsby/hooks/stars.ts
shcheklein: yep, I mean the `?current=${stars}` part!

src/server/index.js
shcheklein: yeah, just curious why specifically this place

src/server/middleware/api/github.js
shcheklein: 👍

src/server/index.js
rogermparent: Totally! It's weird to see it in three places including here, but there's a reason.

shcheklein: 👍

gatsby-config.js
rogermparent: Fixed! Now stars displays as 8888 when no GitHub API token is given.

shcheklein: is it easy enough to skip and return empty value?

src/server/middleware/api/github.js
rogermparent: Done! Now it's an empty 403 for both instances.

src/server/middleware/api/index.js
rogermparent: Changed!

rogermparent:
I got that cache issue again, and since I want to investigate it I stopped the current build so I can use the cache it's failing on locally.

shcheklein:
@rogermparent looks like you need to update the exclusion list - check the CI check, please.

Also, I don't like how the default value (build-time) "blinks" into the new one. May be we should fall back only if API is not accessible and show the empty value first.

rogermparent:
> @rogermparent looks like you need to update the exclusion list - check the CI check, please.

Oops! I mistook this for another check that doesn't require manual intervention. I have it fixed for the next push.

> Also, I don't like how the default value (build-time) "blinks" into the new one. May be we should fall back only if API is not accessible and show the empty value first.

I highly doubt that any user would notice that, but if we prefer having a '---' initial placeholder like before I can do that. Doing this will break the star count for anyone who can't/won't run JS (noscript, text browsers, some screen readers and crawlers) as '---' will be delivered with the static HTML. I'm starting on the changes now.

shcheklein:
> Subjectively, I'd also argue it looks more clunky and noticeable to go from a completely differently-shaped '---' to a number instead of one number to a slightly higher one.

agreed! I think we should not show the whole block "⭐️ ----", show empty space and switch to "⭐️ NNNN" only when we get the number.

rogermparent:
> I think we should not show the whole block "star ----", show empty space and switch to "star NNNN" only when we get the number.
Sure! That's not a bad solution: it certainly would look better than "star ----". The other non-design caveats still apply, but I'll move forward with this suggestion for now.

gatsby-config.js
```diff
@@ -199,6 +200,28 @@ const plugins = [
   }
 ]
 
+if (process.env.GITHUB_TOKEN) {
```
shcheklein, line 203: if we don't show it by default - do we still need this build-time block?

rogermparent: Yes, we do. Basically, the original plugin throws if invoked without a token and this gets around that.

shcheklein: 👍

PR #1367, opened 2020-05-27
Get Started: structure and refactoring (without DVC 1.0)

jorgeorpinel:
> Branch rename from #1074

Should change PR target to master branch when #1051 is merged.

- [x] Addresses part of #1073 (put docs Home in /doc)
- [x] too many sections. maybe split/combine into 3? sections (basics, pipelines, experiment mgmt)
- [x] "the biggest potion of this is content - examples, preparing repos, writing text, etc - more than moving sections around" (from https://github.com/iterative/dvc.org/pull/1051#issuecomment-599883040)
- [x] Per https://github.com/iterative/dvc.org/pull/1097#discussion_r401737826:
  > ideally we should update the getting started and remove all the raw metrics and xpaths (expect metrics show).
- [x] Per https://github.com/iterative/dvc.org/pull/1128#issuecomment-611829288:
  > (dvc param) should become part of the Experiments section of the new Get started tutorial

---

- ~~Update for DVC 1.0!~~ See #1370

Issue #1366, opened 2020-05-27
term: stop using "DVC-file" and "stage file"


jorgeorpinel:
> Extracted from #1255 

After 1.0, the term DVC-file only refers to raw/initial data versioned with `dvc add/import/import-url`. however users seem to mostly refer to such files as "dvc file" or `.dvc` file or just `[data].dvc`. Since we have all these alternatives, let's remove the formal term DVC-file? Also since there's not `dvc.yaml` (pipelines file) which adds to the confusion.

We could also address the "import stage" check box from #608:
> -  Eliminate "import stage" term. Just use imported .dvc file or something like that

~~I suggest we always use the specific `[data].dvc` form in our docs.~~

See these changes as an example of what to do: https://github.com/iterative/dvc.org/pull/1370/commits/1008321bb58a6fc23514e12cfa6b97d0d41f27e6

UPDATE: Decided for `.dvc` file instead. See https://github.com/iterative/dvc.org/issues/1366#issuecomment-634975331 just below.

UPDATE: Similarly, we should stop using "stage file" in favor for `.dvc` file as well.

UPDATE: Do not include instances in DVC command output examples.

UPDATE: See https://github.com/iterative/dvc.org/issues/1366#issuecomment-643123274 for known progress

shcheklein:
Agreed. I haven't seen the `[data].dvc` form though - looks too complicated. We can just use `.dvc` files - very explicit and clear.

jorgeorpinel:
@utkarshsingh99 I think this chore is a little too boring TBH, I don't want you to suffer with this any more! Can you just please remind us the list of files that have been done so far up to your 2nd PR (#1408)? Thanks again for the help, BTW! I'll take it from there.

> If you want to upload the remaining changes you have to a new branch in your fork and open a draft PR so I can take it over that would also be OK, up to you.

utkarshsingh99:
Sure.
1. **Command Reference**
  - _PR 1 #1372_ 
     - add.md
     - get-url.md
     - get.md
     - init.md
     - move.md
     - repro.md
  - _PR 2 #1408_ 
     - fetch.md
     - import.md
     - import-url.md
     - list.md
     - pull.md
     - push.md
     - remove.md
     - status.md
     - update.md
2. **API Reference**
  - _PR 1 #1372_ 
     - get_url.md
3. **Install**
  - _PR 1 #1372_ 
     - plugins.md
4. **User-Guide**
  - _PR 2 #1408_
[...]

Issue #1255, opened 2020-05-06
docs: Major DVC 1.x discrepancies


jorgeorpinel:
#### Mission critical
- [x] Get Started! See https://github.com/iterative/dvc.org/issues/599#issuecomment-559233708 - Ivan is on it iterative/example-repos-dev/pull/14, 
- [x] Installation
- [x] `dvc add` — mainly .dvc file examples #1411 + merge #1414 
- [x] DVC-file guide(s) and maybe cache guide - included in #1370
- [x] `dvc run` PR #1420 + merge #1387
- [x] DVC-file guide(s) and maybe cache guide - included in #1370
- [x] document dvc.lock in user guide (see #1422) - #1448
- [x] dvc metrics - #1360
- [x] dvc plots - #1382 

#### Finally

- ~~document run cache? (#1289)~~ ~~#1464~~ ~~#1655~~ (Will happen when experiments are more stable)
- [x] BLOCKER: Figure out auto linking for dvc.yaml and .dvc files to avoid manual management of hundreds of instances (see #1431). - #1576
- NEW: Address https://github.com/iterative/dvc.org/issues/1663 to simplify dvc.yaml/lock and .dvc file mentions (extracted)
- [x] Remove term `Dvcfile` (see #1445)
- + Full sync of cmd ref usage strings and option names (see iterative/dvc/issues/4024)

#### Next priorities
- [ ] Review all .dvc file examples throughout docs (in YAML code blocks) — note we require YAML 1.2 (se iterative/dvc/pull/4415 and #1764) - #1608, 
- [ ] ^ Also push/pull examples
- [ ] Review DVC cache info everywhere, especially in the internal files/dirs guide.
- ~~Replace all instances of term "DVC-file(s)" 😱  as well as "stage file(s)". Use `dvc.yaml` and/or `dvc.yaml` as appropriate. (Review stage concept itself, see #1366)~~ - #1372 + extracted to https://github.com/iterative/dvc.org/pull/2047#issue-543854582
- ~~Update/remove Katacoda courses accordingly (e.g. fix #1586, see also #1400, #670)~~ Not 1.x related
- ~~Update all other command references 😱 (make a list of which ones are pending first...) e.g. #1456, #1701~~ Extracted to #1824
- [x] Review all `dvc add` examples throughout docs. Avoid having to include the command output unless it's important given the context, definitely don't include progress bars - #1605, #1608
[...]

PR #1370, opened 2020-05-28
user guide: DVC Files and Directories - 1.0 updates

jorgeorpinel:
~~Completes #1367~~

See also iterative/example-repos-dev/pull/14

LOTS OF FILES CHANGES but these are the main ones:

- user-guide/dvc-files-and-directories.md now also explains both .dvc files and dvc.yaml.
- Most other changes are just the link update to the above guide.
- Changes in content\docs\user-guide\basic-concepts update the DVC-file term usage (rel: #1366).
- Avoid "meta" and "metafile" terms

---
- [ ] PENDING: After this PR and all the linked occurrences of "DVC-file" are updated to ".dvc file" (see #1366), update the link to the specific subsection of the docs, either `#dvc-files` or `#dvcyaml-files`.


content/docs/user-guide/dvc-metafile-formats.md
```diff
@@ -0,0 +1,128 @@
...
+- `path` - path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` - (optional) whether or not DVC should cache the output. `true` by
+  default
+
+Metrics entries can contain these fields:
+
+- `type`: Type of the metric file (`json`)
+- `xpath`: Path within the metric file to the metrics data(e.g. `AUC.value` for
+  `{"AUC": {"value": 0.624321}}`)
```
jorgeorpinel, line 123: I'm not sure yet about all these details. Probably some things are missing and others have been deprecated.

shcheklein:
I think we should keep a simple name - DVC files ... we describe a lot of stuff in that section, not all of those file are "meta" ... arguably, even "dvc.yaml" is not _meta_. And I don't see a reason to introduce some tech specific "metafile" to be honest.

jorgeorpinel:
> I think we should keep a simple name - DVC files

@shcheklein but we separately have this guide https://dvc.org/doc/user-guide/dvc-files-and-directories which uses the term "DVC files" for the internal files. That's why I thought of lumping .dvc files and dvc.yaml as "DVC metafiles". Def. open to discuss terminology here.

shcheklein:
It feels like we can do a section DVC Files and Directories with a few subsections inside:

1. .dvc directory
2. data `.dvc` files
3. pipelines .dvc.yaml`

In the index page we can show a tree of a DVC project for people to visualize what is going on.

WDYT?

jorgeorpinel:
> we can do a section DVC Files and Directories with a few subsections inside
From https://github.com/iterative/dvc.org/pull/1370#issuecomment-635639551

Yep, that's is a good plan @shcheklein. Will do that.

Hi @MerlynMShelley! I think you're commenting in the wrong place probably? There's an issue about that, yes. But this PR is not specifically aiming to solve that. Can you find and comment on the right issue so we can follow up? Thanks

jorgeorpinel:
Hi @MerlynMShelley! I think you're commenting in the wrong place probably? There's an issue about that, yes. But this PR is not specifically aiming to solve that. Can you find and comment on the right issue so we can follow up? Thanks

jorgeorpinel:
p.s. but yes what you mentioned sounds correct. For that other issue that we have to find.

content/docs/install/plugins.md
```diff
@@ -16,15 +16,3 @@ autocmd! BufNewFile,BufRead Dvcfile,*.dvc setfiletype yaml
...
-
-A community member, [@prihoda](https://github.com/prihoda), maintains a plugin
-for IntelliJ IDEs, it offers a more robust integration than just syntax
-highlighting.
-
-You can download the plugin from
-[JetBrains Plugins repository](https://plugins.jetbrains.com/plugin/11368-dvc-support-poc)
-
-For more information, visit the plugin's repository:
-[iterative/intellij-dvc/](https://github.com/iterative/intellij-dvc/)
```
jorgeorpinel, line 30: Removed since it's outdated but opened iterative/intellij-dvc/issues/7 about it.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,136 @@
...
+by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
+the following fields:
+
+- `cmd` - executable command defined in this stage
+- `deps` - list of <abbr>dependencies</abbr> for this stage
+- `params` - (optional) list of the [parameter](/doc/command-reference/params)
+  names and their current values
+- `outs` - list of <abbr>outputs</abbr> for this stage
+- `metric` - (optional) list of [metric](/doc/command-reference/metrics) files
+- `locked` - (optional) whether or not this stage is locked from reproduction
```
efiop, line 89:
```suggestion
- `frozen` - (optional) whether or not this stage is locked from reproduction
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,136 @@
...
+When you add a file or directory to a <abbr>DVC project</abbr> with `dvc add` or
+`dvc import`, a `.dvc` file is created based on the data file name (e.g.
+`data.xml.dvc`). These files contain the basic information needed to track the
+data with DVC.
+
+They use a simple YAML format that can be easily written or altered manually.
+Here is a sample:
+
+```yaml
+outs:
```
efiop, line 33: Btw, this is correct for `dvc add`ed files, but for `dvc import`ed (or import-url) it is a bit more complex, same as it was before.

jorgeorpinel:
Let's split this to only focus on the DVC file format guide — huge enough by itself 😢

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK updated here and in the import cmd ref in 8fbf1294. Thanks

content/docs/user-guide/dvc-metafile-formats.md
jorgeorpinel: Has been reviewed by Ruslan in https://github.com/iterative/dvc.org/pull/1370#pullrequestreview-426091399.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Good point. I've added the deps section for .dvc files as well now. See bdab8aee

Please lmk if the latest version has any other errors or missing info. Thanks @efiop

content/docs/command-reference/add.md
```diff
@@ -141,22 +145,16 @@ $ tree
 Let's check the `data.xml.dvc` file inside:
 
 ```yaml
-md5: aae37d74224b05178153acd94e15956b
 outs:
   - cache: true
```
efiop, line 149:
```suggestion
```

this one is no longer present by default.

content/docs/command-reference/import.md
```diff
@@ -223,8 +222,6 @@ outs:
   - md5: b6923e1e4ad16ea1a7e2b328842d56a2.dir
     path: cats-dogs
     cache: true
```
efiop, line 224:
```suggestion
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+```
+
+`dvc.yaml` files consists of a group of `stages` with names provided explicitly
+by the user with the `--name` (`-n`) option of `dvc run`.
+
+Each stage can contain the following fields:
+
+- `cmd`: Executable command defined in this stage
+- `deps`: List of <abbr>dependencies</abbr> for this stage
+- `outs`: List of <abbr>outputs</abbr> for this stage
```
efiop, line 107: Strictly speaking, these are also optional. A stage might not have deps and/or outs.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `params` (optional): List of the [parameter](/doc/command-reference/params)
+  names and their current values
+- `metric` (optional): List of [metric](/doc/command-reference/metrics) files
+- `frozen` (optional): Whether or not this stage is frozen from reproduction
+- `always_changed` (optional) : Whether or not this stage is considered as
+  changed by commands such as `dvc status` and `dvc repro`. `false` by default
+
+An output entry consists of these fields:
+
+- `md5`: Hash value for the output file
```
efiop, line 117: Strictly speaking, `md5` is optional. User could totally remove it and dvc will be fine.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `path`: Path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` (optional): Whether or not DVC should cache the output. `true` by
+  default
+
+A `dvc.yaml` dependency entry consists of a these possible fields:
+
+- `path`: Path to the dependency, relative to the `wdir` path (always present)
+- `md5`: MD5 hash for the dependency (most [stages](/doc/command-reference/run))
+- `etag`: Strong ETag response header (only HTTP <abbr>external
```
efiop, line 127: Both optional.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+  > See the examples in
+  > [External Dependencies](/doc/user-guide/external-dependencies) for more
+  > info.
+
+Metric entries can contain these fields:
+
+- `type`: Type of the metric file (`json`)
+- `xpath`: Path within the metric file to the metrics data(e.g. `AUC.value` for
+  `{"AUC": {"value": 0.624321}}`)
+
```
efiop, line 139: Not true, these were removed. `metric/plots` have the same structure as `outs`.

efiop: Also, missing `plots`

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+by the user with the `--name` (`-n`) option of `dvc run`.
+
+Each stage can contain the following fields:
+
+- `cmd`: Executable command defined in this stage
+- `deps`: List of <abbr>dependencies</abbr> for this stage
+- `outs`: List of <abbr>outputs</abbr> for this stage
+- `params` (optional): List of the [parameter](/doc/command-reference/params)
+  names and their current values
+- `metric` (optional): List of [metric](/doc/command-reference/metrics) files
```
efiop, line 110: Missing plots.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: Not optional, `dvc.yaml` cannot contain `md5` or `etag` at all.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+Metric entries can contain these fields:
+
+- `type`: Type of the metric file (`json`)
+- `xpath`: Path within the metric file to the metrics data(e.g. `AUC.value` for
+  `{"AUC": {"value": 0.624321}}`)
+
+`dvc.yaml` files also support `# comments`.
+
+> `dvc.yaml` comments are preserved among executions of `dvc run`, `dvc repro`,
+> and `dvc commit`.
```
skshetry, line 143: I think, this is unnecessary as it preserves comment for all of the supported operations.

skshetry: Have you tried it, btw @jorgeorpinel? I did preserve comments before, not sure I have messed it up.

content/docs/user-guide/dvc-files-and-directories.md
skshetry, line 110:
```suggestion
- `metrics` (optional): List of [metric](/doc/command-reference/metrics) files
```

content/docs/command-reference/add.md
jorgeorpinel: Thanks! Out of scope though, being removed in https://github.com/iterative/dvc.org/pull/1411/files#diff-1a0ca14da9037fa6403dcd8a33b4f16dL144 🙂

jorgeorpinel: Lots of overlapping docs PRs rn 😭

content/docs/command-reference/import.md
jorgeorpinel: Out of scope but noted in #1255 thanks!

>  Review all .dvc file examples throughout docs 😱

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Updated in e957993e

jorgeorpinel: I wonder if "optional" is the best term for some of these... Oh well resolving for now.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: But they get always get written by DVC commands every time right? Maybe except with --no-commit

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Ugh good point. I need to make a new dvc.lock section. I'm going to extract this task...

jorgeorpinel: Extracted to #1422 for now, let's pretend dvc.yaml combines dvc.lock for now... Thanks

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Good catch. Updated in f32ffd57

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,162 @@
...
+Metrics entries can contain these fields:
+
+- `type`: Type of the metrics file (`json`)
+- `xpath`: Path within the metric file to the metrics data(e.g. `AUC.value` for
+  `{"AUC": {"value": 0.624321}}`)
+
+`dvc.yaml` files also support `# comments`.
+
+> `meta` fields and `#` comments are preserved among executions of `dvc run`,
+> `dvc repro`, and `dvc commit`.
```
skshetry, line 151: They should be always preserved. Unless user does `dvc remove stage-name` of-course. So, I think, it's redundant to even have a note.

jorgeorpinel: For .dvc files these are not preserved when running add again (I think), so we have a similar note for that section. So if you read that you may wonder about the same field in dvc.yaml, which is why we also have the note here. But maybe I can just say "they're always preserved".

jorgeorpinel:
Only missing documenting dvc.yaml (#1422) now (and this Q: https://github.com/iterative/dvc.org/pull/1370#discussion_r438550276), I think. Can we merge? CC @shcheklein

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Updated in 910e6dbe

content/docs/user-guide/dvc-files-and-directories.md
skshetry: `dvc.yaml` does not have `md5`, it's in the `dvc.lock`. For `.dvc` files, @jorgeorpinel, you are right. In case of `--no-commit` for `dvc.yaml` stages, there should not be a lockfile changes at all.

content/docs/user-guide/basic-concepts/external-dependency.md
```diff
@@ -3,8 +3,9 @@ name: 'External Dependency'
 match: ['external dependency', 'external dependencies']
 ---
 
-A stage dependency (`dep` field in [`dvc.yaml`](/doc/user-guide/dvc-file-format)
-or in an [import stage](/doc/command-reference/import) `.dvc` file) with origin
-in an external source, for example HTTP, SSH, Amazon S3, Google Cloud Storage
-remote locations, or even other DVC repositories. See
+A stage dependency (`dep` field in
```
skshetry, line 6:
```suggestion
A stage dependency (`deps` field in
```

jorgeorpinel: Thanks, committing.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `md5`: Hash value for the output file
+- `path`: Path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` (optional): Whether or not DVC should cache the output. `true` by
+  default
+
+A `dvc.yaml` dependency entry consists of a these possible fields:
+
+- `path`: Path to the dependency, relative to the `wdir` path (always present)
+- `md5`: MD5 hash for the dependency (most [stages](/doc/command-reference/run))
```
shcheklein, line 141: same here - we don't store md5 in the `dvc.yaml` anymore

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `path`: Path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` (optional): Whether or not DVC should cache the output. `true` by
+  default
+
+A `.dvc` file dependency entry consists of a these possible fields:
+
+- `path`: Path to the dependency, relative to the `wdir` path (always present)
+- `md5`: MD5 hash for the dependency (most [stages](/doc/command-reference/run))
+- `etag` (optional): Strong ETag response header (only HTTP <abbr>external
```
shcheklein, line 65: if we keep optional for etag, md5 should be optional as well - they play the same role

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+  - md5: a304afb96060aad90176268345e10355
+    path: data.xml
+
+# Comments and user metadata are supported.
+meta:
+  name: 'John Doe'
+  email: john@doe.com
+```
+
+`.dvc` files can contain the following two fields:
```
shcheklein, line 44: two -> three? (or just omit number - easier to maintain)

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I've extracted dvc.lock to #1422

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `meta` (manual): Arbitrary information can be added here manually. Any YAML
+  contents can be added. `meta` contents are ignored by DVC, but they can be
+  useful for user processes that read `.dvc` files.
+
+An output entry can consist of these fields:
+
+- `md5`: Hash value for the output file
+- `path`: Path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` (optional): Whether or not DVC should cache the output. `true` by
```
shcheklein, line 58: @efiop @skshetry does cache=false still make sense in DVC 1.0 for .dvc files?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,148 @@
...
+- `frozen` (optional): Whether or not this stage is frozen from reproduction
+- `always_changed` (optional): Whether or not this stage is considered as
+  changed by commands such as `dvc status` and `dvc repro`. `false` by default
+- `meta` (optional): Arbitrary information can be added here manually. Any YAML
+  contents can be added. `meta` contents are ignored by DVC, but they can be
+  useful for user processes that read `.dvc` files.
+
+> `meta` fields and `#` comments are always preserved in `dvc.yaml` stages.
+
+An output entry (`outs`, `metrics`, or `plots`) consists of these fields:
```
shcheklein, line 130: plots entry has more fields

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: For the record, this has been noted in #1422

content/docs/user-guide/dvc-files-and-directories.md
skshetry: It does not (except old-style stagefiles), but I guess it's fine to write it as an `optional` field.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Fixed in de1f12c6 but per private conversation will probably remove all these "optional" marks. Maybe just add "always present" to `outs` and `path` (everything else is optional).

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Fixed along with other things in de1f12c6

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Hm yeah also `metrics` and `path` doesn't even exist anymore in dvc.yaml... No one caught that one 😋 fixed

jorgeorpinel: Well, extracted params.yaml into https://github.com/iterative/dvc.org/issues/1422#issuecomment-643753423 for now

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Sorry, I'm confused. Does the `cache` field exist in dvc.yaml? Thanks

skshetry: It does, as an attribute of an output. But, isn't this about `.dvc` files?

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: @jorgeorpinel let's remove `optional`s for now everywhere.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,152 @@
...
+- `deps` (optional): List of <abbr>dependency</abbr> entries for this stage,
+  only present when `dvc import` and `dvc import-url` are used. Typically there
+  is only one (but several can be added manually).
+- `meta` (optional): Arbitrary metadata can be added manually with this field.
+  Any YAML contents is supported. `meta` contents are ignored by DVC, but they
+  can be meaningful for user processes that read `.dvc` files.
+
+An _output entry_ can consist of these fields:
+
+- `md5` (optional): Hash value for the output file
```
shcheklein, line 57: it's not optional I guess ... it depends on the definition of optional of course ... that's why let's just remove them for now everywhere

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,152 @@
...
+- `path`: Path to the output in the <abbr>workspace</abbr>, relative to the
+  location of the `.dvc` file
+- `cache` (optional): Whether or not DVC should cache the output. `true` by
```
shcheklein, line 60: let's use "output" term as less a possible .. I'm a bit fine with it when we use it with <abbr> or at least when we use it with `dvc.yaml` .... but it's better to be explicit and do not use some strange DVC terminology

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: let's deprecate this? what is the meaning for this?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,152 @@
...
+```
+
+`.dvc` files can contain the following fields:
+
+- `outs`: List of <abbr>output</abbr> entries for this `.dvc` file. Typically
+  there is only one (but several can be added manually).
+- `deps` (optional): List of <abbr>dependency</abbr> entries for this stage,
+  only present when `dvc import` and `dvc import-url` are used. Typically there
+  is only one (but several can be added manually).
+- `meta` (optional): Arbitrary metadata can be added manually with this field.
```
shcheklein, line 51: this probably the only meaningful place for `optional`

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,152 @@
...
+- `frozen` (optional): Whether or not this stage is frozen from reproduction
+- `always_changed` (optional): Whether or not this stage is considered as
+  changed by commands such as `dvc status` and `dvc repro`. `false` by default
+- `meta` (optional): Arbitrary metadata can be added manually with this field.
+  Any YAML contents is supported. `meta` contents are ignored by DVC, but they
+  can be meaningful for user processes that read `.dvc` files.
+
+`dvc.yaml` files also support `# comments`.
+
+> `meta` fields and `#` comments are always preserved in `dvc.yaml` stages.
```
shcheklein, line 147: are they preserved if I overwrite the stage? cc @skshetry @efiop ?

shcheklein: @jorgeorpinel no need to put it as a note - it can be part of the previous paragraph(s)

shcheklein: also, if they are preserved I would omit it completely - this what anyone would expect anyway - anythings else is a bug that is not fixed yet in DVC

shcheklein: that we should mention about

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,152 @@
...
+          x: epoch
+          x_label: Epoch
+    meta: 'For deployment'
+    # User metadata and comments are supported.
+```
+
+`dvc.yaml` files consists of a group of `stages` with names provided explicitly
+by the user with the `--name` (`-n`) option of `dvc run`.
+
+Each stage's contents are similar to individual [`dvc` files](#dvcfiles) but
```
shcheklein, line 123: `.dvc`? 

link is broken

(how many other links are broken like this???)

No need to put any parallels - stage file vs .dvc. They are play very different role. Or you can put something like "It has similar to .dvc files fields"

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -13,16 +157,17 @@ operation:
...
 
   > Note that DVC includes the cache directory in `.gitignore` during
   > initialization. No data tracked by DVC will ever be pushed to the Git
-  > repository, only [DVC-files](/doc/user-guide/dvc-file-format) that are
-  > needed to download or reproduce them.
+  > repository, only [DVC-files](/doc/user-guide/dvc-files-and-directories) that
+  > are needed to download or reproduce them.
 
 - `.dvc/plots`: Directory for
   [Plot templates](/doc/command-reference/plots#plot-templates).
```
shcheklein, line 173: either put `.` after all sentences or remove here

content/docs/user-guide/dvc-files-and-directories.md
skshetry: Yes, they are preserved when overwriting too (I don't know if that's what we want, but when implementing the "maintain the order when overwrite" I cut the corner and did it the fastest/easiest way).

skshetry: I agree with @shcheklein  and I have mentioned this already somewhere. There's no need for the note at all.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > isn't this about .dvc files?

Both .dvc files and dvc.files are specified in this doc. Soon also dvc.yaml and maybe params.yaml

All the special DVC files

> let's deprecate this?

Should we open an issue in the core repo?

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Yes, all removed now. I failed to push my last commit this morning sorry 🤦

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Yep, this one was left in.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > if they are preserved I would omit it completely - this what anyone would expect anyway

Saugat also mentioned this earlier, same answer:  for .dvc files the behavior is different and has a note about it (making both notes part of the paragraph now though) so if you read that first in the .dvc file section you may expect that dvc.yaml files behave similarly, thus we need to note here as well, I think.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: This paragraph was removed per a previous comment you had (but I forgot to push this last commit this morning). I cannot find any more instances of `dvc` file without . or this broken link anchor in docs so it was the only one.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: These weren't changed in this PR but sure, fixed. Only a couple of these

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,142 @@
...
+  `dvc plots modify`).
+- `frozen`: Whether or not this stage is frozen from reproduction
+- `always_changed`: Whether or not this stage is considered as changed by
+  commands such as `dvc status` and `dvc repro`. `false` by default
+- `meta` (optional): Arbitrary metadata can be added manually with this field.
+  Any YAML contents is supported. `meta` contents are ignored by DVC, but they
+  can be meaningful for user processes that read or write `.dvc` files directly.
+
+`dvc.yaml` files also support `# comments`. `meta` fields and `#` comments are
+always preserved in `dvc.yaml` stages.
```
shcheklein, line 137: @jorgeorpinel we don't need to mention that they are preserved I think. That's what would 100% users expect as I mentioned before. It's confusing to see even this mentioned since you start thinking why is that important to mention, may something else is not preserved, etc.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -1,8 +1,142 @@
...
+```
+
+`dvc.yaml` files consists of a group of `stages` with names provided explicitly
+by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
+the possible following fields:
+
+- `cmd` (always present): Executable command defined in this stage
+- `deps`: List of <abbr>dependency</abbr> file or directory paths of this stage
+- `params`: List of the [parameters](/doc/command-reference/params). These are
+  key paths referring to another YAML file (`params.yaml` by default).
```
shcheklein, line 123: only YAML is supported? or JSON as well? just to double check

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: also, double check is they are preserved when we overwrite the stage

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: I think we introduced quite a few items in this PR also, and looks like all of them are still not consistent (def not a blocker)

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Good point. Updated in d4705d63

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I think we do because .dvc files has this note "not (preserverd) when a `.dvc` file is overwritten". But OK, removing from here. It's not a big deal either way.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Good catch, also JSON indeed according to https://dvc.org/doc/command-reference/params#examples. It's just not something I've ever tried or was very aware of. Updating...

jorgeorpinel:
Done with the last few items. Merging!

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Thanks. I double checked all the bullets just now and I did miss just 2 more. Fixed in cd7bca05

Keep in mind that some of the incomplete sentences still get a period because a sentence is coming after and we need to separate them.

jorgeorpinel:
OK! Double checked CI test fail and it's just the new link. Also looked at preview app and it looks fine. Merging.

jorgeorpinel:
> looked at preview app and it looks fine

p.s. it's getting pretty long though @shcheklein, should we make each file a sub-page? esp. when adding dvc.lock

PR #1364, opened 2020-05-27
blog: author name clickable, description and posts by the author visible

utkarshsingh99:
Fix #1120 

- [X] Please read the guidelines in the [Contributing to the Documentation](https://dvc.org/doc/user-guide/contributing/docs) list if you make any substantial changes to the documentation or JS engine.

- [X] 🐛 Please make sure to mention `Fix #issue` (if applicable) in the description of the PR. This causes GitHub to close it automatically when the PR is merged.

- [X] Please choose to [allow us](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to edit your branch when creating the PR.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

**Built a new Author Page that looks like this**:
snapshot 1
snapshot 2
The link for every author is their `sourcePath` except ~~.md~~
If the sourcePath is: _authors/dmitry_petrov.md_, the link would be: **authors/dmitry_petrov**.

> make author name clickable

Resolved: The link on every blog post for the author's social media handle has been replaced with the link for their author page. 

> show description and posts by the author

Done. Unfortunately, the authors node in graphql does not have the description field unlike the markdown files of authors. Thus, I was not able to add a great description and had to just add their social media links.

For the future, as #1349 mentions, we can add more links in `Blog/Author/index.tsx` by passing the required props.

PR #1375, opened 2020-05-29
social icons and better author links

### LINKED ISSUE #1349: multiple author links
#1341 allows us to pick one link for each author to link their name to in blog posts.

At some point, I expect us to want to expand on this and allow for multiple links in the very common case of multiple social media websites.

I intend for this issue to expand the `link` field to a `links` array, allowing for us to specify external links in any way we like while also having the order they're in as a useful piece of data.

We can use a combination of strings and objects to handle simple and complex link cases respectively (think Twitter vs personal portfolio site- we'd like to handle both I'd think), and also use the ordering as an indication of priority (One author can list their Twitter over their LinkedIn because their Twitter is more active, while another can do the opposite) and also use some build-time logic to recognize common social media sources and use the data to fill in icons.

Solving this issue provides a data source that feeds into #1120's individual Author pages, but this specific issue can be implemented in a way that doesn't require that one to be fully solved yet (i.e. I add this schema, adapt the authors, and then plug it into our current templates without requiring any design changes)

Improves current solution of #1161

I'll also accept requests for adding/changing Author links in this PR, so leave a comment if you want them changed.

If we want to roll a design change on posts that integrates social icons into the existing blog page, I can do that in this branch as well. Alternatively, that can be a separate issue where we discuss the design of the icons.

rogermparent:
This branch builds on #1341 and fixes #1349 by converting the current author `link` field to `links`, a fully-fledged processed field in the Author model that takes an array of links represented in the markdown frontmatter in either string shortform (`"https://twitter.com/myprofile"`) or object longform (`site: "twitter", url: "https://twitter.com/myprofile", username: "myprofile"`).

Strings are converted into the object format at build-time, and they can be queried throughout the rest of the site for anywhere they're needed, like these social icons or a bigger descriptive list in an Author page.

Very likely will feed into #1364
Addresses #1161 even more


src/components/SocialIcon/index.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import Link from '../Link'
+import { ReactComponent as TwitterIcon } from '../../../static/img/community/icon-twitter.svg'
+import { ReactComponent as GithubIcon } from '../../../static/img/community/icon-github.svg'
```
shcheklein, line 4: I think we should move them out of community if we use them in multiple places ... to be honest I don't even see a problem in certain cases to duplicate them so that we can manage them independently

src/components/SocialIcon/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+   Given the situation where either the given link has no site or we don't have
+   an icon for it, we return null such that nothing is rendered in map
+   functions.
+*/
+const SocialIcon: React.FC<ISocialIconProps> = ({
+  site,
+  url,
+  className
+}): JSX.Element | null => {
+  if (!site) return null
```
shcheklein, line 34: do we need this check? next line will return null or undefined anyway right?

src/gatsby/models/authors/onCreateMarkdownContentNode.js
```diff
@@ -1,16 +1,50 @@
+function parseLink(input) {
```
shcheklein, line 1: would be good to write a test for this

src/components/SocialIcon/index.tsx
shcheklein: in this case probably just move them to `./`

src/components/SocialIcon/index.tsx
rogermparent: Technically no, but this solved a TS issue about using undefined as an index. I'll solve this better with build-time TS.

src/gatsby/models/authors/onCreateMarkdownContentNode.js
```diff
@@ -1,16 +1,50 @@
...
+  if (typeof input === 'string') {
+    const url = new URL(input)
+    // Slice the 'www.' off of given hostnames to normalize them.
+    const hostname = url.hostname.startsWith('www.')
```
shcheklein, line 5: nitpick: I would probably do a one liner with regex

shcheklein: you can even remove trailing slash from the path if it exists

src/gatsby/models/authors/onCreateMarkdownContentNode.js
```diff
@@ -1,16 +1,50 @@
+function parseLink(input) {
+  if (typeof input === 'string') {
```
shcheklein, line 2: why do we need this check?

src/components/SocialIcon/index.tsx
rogermparent: I have no problem with that, the existing icons aren't used in many other places anyway.

Also, with the SVGR resizing fix, we can use imported SVGs instead of static ones with no downsides.
I'll just do the ones involved here for now, but we may want to consider giving most of our static SVG icons this treatment later so they can be run through SVGR and maybe inlined.

src/gatsby/models/authors/onCreateMarkdownContentNode.js
rogermparent: I mentioned in the original spec that we should allow for both strings or objects when defining Author links. Both methods handle different situations well, and allowing for both not only makes up for the weaknesses of each alone but is also very easy and natural to YAML Markdown frontmatter.

Strings are simple, easy to use and pack a lot of info for known sites like Twitter. Supporting custom objects, however, allows for total control in special cases like having a personal blog, a portfolio site, or simply wanting the Twitter link to have a description associated with it if we source this data in a more detailed context.

This support would be particularly useful for the new Author pages at #1364.

I agree that the function isn't very expressively written though, I'll try to improve that.

src/components/SocialIcon/index.tsx
rogermparent: I had to use a combination of `eslint-disable-next-line` and, weirdly enough, some more TS (the `!` at the end of the index) to solve this issue.

From my research, it seems that TS is hard-coded to only accept numbers or strings as object indexes despite what we're doing working fine in vanilla JS. The solution I use now pacifies the TS lint check and doesn't introduce another build-time check.

src/gatsby/models/authors/onCreateMarkdownContentNode.js
rogermparent: I more or less completely remade the implementation that is mostly based off regex- not just become of this nitpick, but to handle edge cases I found when writing the tests.

src/gatsby/models/authors/onCreateMarkdownContentNode.js
rogermparent: The new parser makes it more obvious that it's handling both objects and strings.

rogermparent:
The new changes I made address all the prior points, but also introduce enough new code (at build time) to warrant a re-review.

rogermparent:
> I think it's hard to justify that we have 3 files for a function that parses links. I can appreciate the generalization and modularity of the approach you has taken, but I would say in this case it's a bit too much. Ideally we all should keep in mind - best code == no code :) In this specific case, I would recommend to try to keep it within a file (at least do not extract utils that are actually being used only in one place and specific to that place - they should be private members). Embed some helpers. Try to a bit less verbose comments - a lot of comments, large comments sometimes a sign of some too complicated code (consider making function names more explicit? split one line into two lines with some vars that describe the flow, etc).

I hear you on the split files and big comments. They come from my personal style, but I'll go back over them and see what's worth cutting. At the very least the string parser can be combined with the multi-type parser.

I think two files would be best because the strength of this implementation is that we abstract away the details of common patterns found in social links while still being able to mix-and-match them alongside JS-based custom logic, especially as we add more sites to the recognized list.

Thanks for the feedback!

rogermparent:
I ended up going with one file, two lacks the advantages of both three and one. It's all the same code, and I ran another build just to make sure, so we should be good to merge!

src/gatsby/models/authors/parse-link.js
```diff
@@ -1,31 +1,77 @@
...
+    ({ pathname }) => ({
+      site: 'linkedin',
+      username: /^\/in\/(.*)/.exec(pathname)[1]
+    }),
+    urlHTTPS
+  )
+}
+const defaultProcessor = processor(asSite(null), urlDefault)
+
+const processStringLink = groups =>
```
shcheklein, line 68: I would embed this - it's one line and never used anywhere else, right?

src/gatsby/models/authors/parse-link.js
```diff
@@ -1,31 +1,77 @@
...
-            equivalent to no provided subdomain. Other subdomains will be
-            included in "host"
-
-         3. Capture everything between here and the next / as "host"
-            This technically includes a port if we ever include one.
-
-         4. Capture the rest of the string, excluding leading and trailing
-            slash, as "pathname"
-      */
       const result = /^(?<scheme>.*?\/\/)?(?:www\.)?(?<host>[^\/]*)(?<pathname>.*?)\/?$/.exec(
```
shcheklein, line 75: can we use URL or URI class (we have some code around) to get all these parts?  + run simple function right here to trim `/` and `www.` (better to make is case insensitive)

src/gatsby/models/authors/parse-link.js
```diff
@@ -1,31 +1,77 @@
...
+})
+
+const pathnameAsUsername = ({ pathname }, input) => ({
+  ...input,
+  username: trimSlashes(pathname)
+})
+
+const urlHTTPS = ({ host, pathname }, input) => ({
+  ...input,
+  url: 'https://' + host + pathname
```
shcheklein, line 40: where do we get `/` between host and path?

src/gatsby/models/authors/parse-link.js
```diff
@@ -1,31 +1,77 @@
...
 
 function WrongTypeError(type) {
   return new Error(`A ${type} cannot be used as input for parseLink!`)
 }
 
+// Building processors for recognized sites
+
+const processors = {
+  'twitter.com': processor(asSite('twitter'), urlHTTPS, pathnameAsUsername),
+  'github.com': processor(asSite('github'), urlHTTPS, pathnameAsUsername),
```
shcheklein, line 52: I reminds me lodash flow (or may be something similar exists?)? should we just use that instead of writing our own?

src/gatsby/models/authors/parse-link.js
```diff
@@ -1,31 +1,77 @@
...
+
+// Processor helpers
+
+function trimSlashes(input) {
+  return /^\/?(.*?)\/?$/.exec(input)[1]
+}
+
+const asSite = site => (groups, input) => ({
+  ...input,
+  site
```
shcheklein, line 30: should it be "site: site"?

src/gatsby/models/authors/parse-link.js
rogermparent: Sort of, but not quite- at least for the built-in URL Node class. Builtin URL can't handle "twitter.com/profile" or even "www.twitter.com/profile" without "http://" before it. We'd have to manually add the scheme to links that don't have it before we can parse them with builtin URL, and at that point we're defeating the simplicity angle.

I can make the `www` case-insensitive easily enough, but I didn't feel the need to handle the situation where an editor puts a capital `WWW.` in their social media links and that gets accepted and merged. Still `[wW]{3}` is just a better way of handling that so I'll make that change.

As far as the regex method itself, I believe it's the best way going forward. This isn't just some generic link parser, it's specifically parsing editor-written frontmatter data into GraphQL nodes, handling the input as a generic string with regex handles more user-facing cases than the URL class.

src/gatsby/models/authors/parse-link.js
rogermparent: That's just how the regex breaks up the string now. I used to strip the leading slash in the regex, but it involved wrapping the pathname in a non-capturing group just to strip a slash and then have it be re-added here later.

I decided to simplify the regex and just let the processors strip the leading slash when they need to. Not only is the regex cleaner, but more operations use the pathname with the slash than without currently- most obviously these url functions that are run on every parsed string.

src/gatsby/models/authors/parse-link.js
rogermparent: I actually started with lodash flow, but writing this with that ended up being more complicated specifically when sharing the parsed data. This implementation is a bit similar to Models, but sequential instead of parallel and return values matter.

src/gatsby/models/authors/parse-link.js
```diff
@@ -0,0 +1,94 @@
...
+  return new Error(`A ${type} cannot be used as input for parseLink!`)
+}
+
+function parseLink(input) {
+  switch (typeof input) {
+    // Handle shorthand string links
+    case 'string':
+      const {
+        groups
+      } = /^(?<scheme>.*?\/\/)?(?:www\.)?(?<host>[^\/]*)(?<pathname>.*?)\/?$/.exec(
```
shcheklein, line 75: ok, makes sense regarding using regex.

just a few thing to check

do we actually need `\/?` at the end - I think `.*` captures everything enyway.

can we normalize pathname right here (like we essentially do with `www.`) - trim `/`. instead of using additional helper?

src/gatsby/models/authors/parse-link.js
```diff
@@ -0,0 +1,94 @@
...
+  ...input,
+  url: (scheme || 'https://') + host + pathname
+})
+
+// Building processors for recognized sites
+
+const processors = {
+  'twitter.com': processor(asSite('twitter'), urlHTTPS, pathnameAsUsername),
+  'github.com': processor(asSite('github'), urlHTTPS, pathnameAsUsername),
+  'linkedin.com': processor(
```
shcheklein, line 53: for linkedin it's better to include www? www.linkedin.com is canonical

src/gatsby/models/authors/parse-link.js
rogermparent: This is ES5 syntactic sugar for the same thing.

shcheklein: 👍

src/gatsby/models/authors/parse-link.js
rogermparent: `\/?` is specifically not captured, and removes trailing slashes from the URL.

```
can we normalize pathname right here (like we essentially do with www.) - trim /. instead of using additional helper?
```

That's exactly what I'm doing here. `www.` is trimmed to make `website.com` and `www.website.com` equivalent, and this regex also handles trailing slashes. I did my best to frontload all the work I could into this regex, down to the named capture groups because positional ones break with how much it's doing.

src/gatsby/models/authors/parse-link.js
rogermparent: Can do!

src/gatsby/models/authors/parse-link.js
shcheklein: 👍

src/gatsby/models/authors/parse-link.js
shcheklein: you should be able to simplify trimSlashes into trimLeadingSlash if the trailing is trimmed here?

src/gatsby/models/authors/parse-link.js
shcheklein: I see. I think this is good. I think it's consistent even with URI (or how is it named in JS?) class.

src/gatsby/models/authors/parse-link.js
```diff
@@ -0,0 +1,94 @@
...
+      (currentResult, _, i) => mods[i](groups, currentResult),
+      {}
+    )
+  }
+}
+
+// Processor helpers
+
+function trimSlashes(input) {
+  return /^\/?(.*?)\/?$/.exec(input)[1]
```
shcheklein, line 25: same here - are you sure it actually trims the trailing slash? do we care about it at all btw?

src/gatsby/models/authors/parse-link.js
rogermparent: Ah, yes, I can do that. Coming up in the next push!

src/gatsby/models/authors/parse-link.js
rogermparent: I know it works on both because I've tested it alone, but only the leading slash matters because it's how we extract the username from the path on many sites (`website.com/username`)

shcheklein: yep, missed the `*?` - lazy part of it

src/gatsby/models/authors/parse-link.js
rogermparent: Done! LinkedIn now specifically adds `www` to match the canonical link.

src/gatsby/models/authors/parse-link.js
rogermparent: Done! It's embedded now and simplified to only work on leading slashes.

shcheklein:
Nicely done @rogermparent !

Issue #1033, opened 2020-03-04
guide: add more links to the .dvcignore guide from other docs


jorgeorpinel:
To help increase awareness about this feature!

> Per https://github.com/iterative/dvc/pull/3257#discussion_r385531995

UPDATE: Here's a non-inclusive list of some ideas of the most obvious places where mentions of .dvcignore should exist. Each one could be addressed in different PRs:

- [x] Definitely in the description of https://dvc.org/doc/command-reference/add, and maybe add an example (not sure) which could be copied from [these](https://dvc.org/doc/user-guide/dvcignore#examples). Check how dvcignore interacts with `dvc move` - #1346
- [ ] Note that the `.hg/` dir is ignored by default along with .git/ (in git-enabled DVC repos), per same discussion. This can go in https://dvc.org/doc/command-reference/init
- [ ] See how .dvcignore interacts with `dvc checkout` and `dvc pull`
- [ ] ^ Same for `dvc list`, `dvc get`, and `dvc import`
- [ ] ^ Same for `metrics` and `plots`
- [ ] ^ `push` and `pull`...
- [ ] ^ etc. just go down the command reference 🙂 
- [ ] Maybe in a get-started page or tutorial? Just as a note with link (no need for actual example).

BTW if while attempting any of the above bugs or strange behavior arise, please ask in Discord or open an issue in the core repo (iterative/dvc)

sarthakforwet:
Hey @jorgeorpinel  . As far as I got this all you mean to say is that we have to refer the https://dvc.org/doc/user-guide/dvcignore from other sections of DVC docs?

jorgeorpinel:
Yes @sarthakforwet! You would first need to play with and fully understand the .dvcignore feature, then explore all the existing docs and try to find all the places where it would be appropriate to link to the .dvcignore guide.

But I see you're also interested in https://github.com/iterative/dvc.org/issues/1172#issuecomment-629842491, let's focus on a single one for now 🙂 please pick one and feel free to submit a PR to close it, thanks.

utkarshsingh99:
If no one is on this issue, I would like to take it up. @jorgeorpinel

jorgeorpinel:
@sarthakforwet are you still interested in this issue or #1172? Thanks

jorgeorpinel:
> p.s. Both involve .dvcignore which I'm guessing you've studied already so at this point focusing on these is probably the best bet.

jorgeorpinel:
[...]

PR #1346, opened 2020-05-24
user-guide: add more links to the .dvcignore guide from other docs. Subtask one completed

sarthakforwet:
Subtask : 
> Definitely in the description of https://dvc.org/doc/command-reference/add, and maybe add an > example (not sure) which could be copied from these.

of issue #1033 completed.

PR #1378, opened 2020-05-30
cmd: example about Dvcignore in move

sarthakforwet:
refers #1033 

- [x] Check how dvcignore interacts with dvc move

Added what happens when moving a file (referenced in `.dvcignore`) with `dvc move`.

### DIFF
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -174,11 +174,6 @@ $ tree
 $ mkdir dir1 dir2
 $ echo data1 > dir1/data1
 $ echo data2 > dir2/data2
-```
-
-Create a `.dvcignore` file and insert pattern matching one of the files:
-
-```dvc
 $ echo dir1/* >> .dvcignore
 ```
 
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -164,3 +164,48 @@ $ tree
     │   └── foo
     └── data3.dvc
 ```
+
+## Example: Moving ignored files
+
+`dvc move` raises an error if it finds a file with pattern available in
+`.dvcignore` file.
+
+Let's imagine it with following scenerio:
+
+```dvc
+$ mkdir dir1 dir2
+$ echo data1 > dir1/data1
+$ echo data2 > dir2/data2
+```
+
+Create a `.dvcignore` file and insert pattern matching one of the files:
+
+```dvc
+$ echo dir1/* >> .dvcignore
+```
+
+Now lets try to move the files with `dvc move`.
+
+```dvc
+$ dvc move dir2/data2 dir3/data3
+$ dvc move dir1/data1 dir4/data4
+```
+
+The second command that we executed above will raise an error and the transfer
+of that file won't happen. The final structure would look like this:
+
+```dvc
+$ tree .
+.
+├── dir1
+│   ├── data1
+│   └── data1.dvc
+├── dir2
+└── dir3
+    ├── data3
+    └── data3.dvc
+```
+
+Here we see that `data2` file from `dir2` directory got transferred to
+`dir3/data3` and `dir1` remained unchanged. See
+[Dvcignore](/doc/user-guide/dvcignore) for more details.
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -167,10 +167,8 @@ $ tree
 
 ## Example: Moving ignored files
 
-`dvc move` raises an error if it finds a file with pattern available in
-`.dvcignore` file.
-
-Let's imagine it with following scenerio:
+`dvc move` raises an error if it finds a file or directory listed in
+`.dvcignore`:
 
 ```dvc
 $ mkdir dir1 dir2
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -174,7 +174,7 @@ $ tree
 $ mkdir dir1 dir2
 $ echo data1 > dir1/data1
 $ echo data2 > dir2/data2
-$ echo dir1/* >> .dvcignore
+$ echo "dir1/*" >> .dvcignore
 ```
 
 Now lets try to move the files with `dvc move`.
@@ -182,12 +182,9 @@ Now lets try to move the files with `dvc move`.
 ```dvc
 $ dvc move dir2/data2 dir3/data3
 $ dvc move dir1/data1 dir4/data4
-```
+ERROR: failed to move 'dir1/data1' -> 'dir4/data4'
+- Unable to find DVC-file with output 'dir1/data1'
 
-The second command that we executed above will raise an error and the transfer
-of that file won't happen. The final structure would look like this:
-
-```dvc
 $ tree .
 .
 ├── dir1
@@ -200,5 +197,8 @@ $ tree .
 ```
 
 Here we see that `data2` file from `dir2` directory got transferred to
-`dir3/data3` and `dir1` remained unchanged. See
-[Dvcignore](/doc/user-guide/dvcignore) for more details.
+`dir3/data3` and `dir1` remained unchanged.
+
+Note that if ignored files are to be moved then `.dvcignore` file must be
+updated accordingly. See [Dvcignore](/doc/user-guide/dvcignore) for more
+details.


content/docs/command-reference/move.md
```diff
@@ -164,3 +164,48 @@ $ tree
     │   └── foo
     └── data3.dvc
 ```
+
+## Example: Moving ignored files
+
+`dvc move` raises an error if it finds a file with pattern available in
+`.dvcignore` file.
+
+Let's imagine it with following scenerio:
```
jorgeorpinel, line 173: The idea for this example is pretty good. Let's also review https://dvc.org/doc/user-guide/dvcignore#example-moving-ignored-data please. After reading it now it seems like it may be outdated or it could be improved in its explanation and extended with `dvc move`.

jorgeorpinel, line 173:
```suggestion
`dvc move` raises an error if it finds a file listed in `.dvcignore`:
```

content/docs/command-reference/move.md
```diff
@@ -164,3 +164,48 @@ $ tree
...
+```dvc
+$ mkdir dir1 dir2
+$ echo data1 > dir1/data1
+$ echo data2 > dir2/data2
+```
+
+Create a `.dvcignore` file and insert pattern matching one of the files:
+
+```dvc
+$ echo dir1/* >> .dvcignore
```
jorgeorpinel, line 184:
```suggestion
$ echo dir1/* > .dvcignore
```

content/docs/command-reference/move.md
```diff
@@ -164,3 +164,48 @@ $ tree
...
+
+Now lets try to move the files with `dvc move`.
+
+```dvc
+$ dvc move dir2/data2 dir3/data3
+$ dvc move dir1/data1 dir4/data4
+```
+
+The second command that we executed above will raise an error and the transfer
+of that file won't happen. The final structure would look like this:
```
jorgeorpinel, line 195: Instead of saying that an error will be thrown, just copy the error output please. Probably no need for that whole paragraph, just continue the code block to include the `tree` command.

content/docs/command-reference/move.md
```diff
@@ -164,3 +164,48 @@ $ tree
...
+```dvc
+$ echo dir1/* >> .dvcignore
+```
+
+Now lets try to move the files with `dvc move`.
+
+```dvc
+$ dvc move dir2/data2 dir3/data3
+$ dvc move dir1/data1 dir4/data4
+```
```
jorgeorpinel, line 192: Wait... You're missing a step here. `dvc add dir1` and `dvc add dir2` — but not sure you can do the latter given the .dvcignore setup. The point is `dvc move` shouldn't be used on just any file, it has to be used on DVC-tracked files or diresctories. If that's not the case please lmk and/or open a bug report in the core repo! https://github.com/iterative/dvc/issues

image

jorgeorpinel:
@sarthakforwet however please focus on finishing #1346 first. Thanks

jorgeorpinel:
> major problem

p.s. they're not problems, just major questions

content/docs/command-reference/move.md
sarthakforwet: > Wait... You're missing a step here. dvc add dir1 and dvc add dir2

Ops! missed accidently.. corrected

> The other major question I see is that this new example doesn't actually show how to move ignored files!

Why would one needs to move ignored files? 
Even if someone wants to move that file (with dvc files also re-named) so they only need to again bring  that file to tracking stage and then apply `dvc move` otherwise normal `mv` will do the trick?  
And even if we wanna describe that we could add a short description of above point, no need to add an example. If you prefer.

jorgeorpinel: > missed accidently.. corrected

I don't see any changes to the PR. You haven't pushed any commits @sarthakforwet. But please read my comment closely as it has some important caveats. Thanks

jorgeorpinel: Please also note several more review comments above this review.

jorgeorpinel:
> > The other major question I see is that this new example doesn't actually show how to move ignored files!
>
> Why would one needs to move ignored files? ... normal mv will do the trick?

Good point. I guess it's an unnecessary example. It should be obvious that .dvcignore has to be kept up to date.

> if we wanna describe that we could add a short description of above point

- [ ] Yes, it would be nice to add a note, something like "note that if ignored files are moved .dvcignore should be updated". Thanks

content/docs/command-reference/move.md
jorgeorpinel: I guess this had not much to do with this PR or the issue in question. Out of scope, never mind.

content/docs/command-reference/move.md
jorgeorpinel: Actually only https://github.com/iterative/dvc.org/pull/1378#pullrequestreview-421518565 is pending before this one. I resolved the other ones myself.

sarthakforwet: > I don't see any changes to the PR. You haven't pushed any commits @sarthakforwet.

I did all the changes that you suggested on my local machine and decided to push when all your suggested changes are adressed.

sarthakforwet: > Please also note several more review comments above this review.

Okay will do that..

sarthakforwet:
I have seen that [Dvcignore](https://github.com/iterative/dvc.org/blob/master/content/docs/user-guide/dvcignore.md) needs some of these changes : 
- [ ] Delete unnecessary newlines
- [ ] Change "checksums" to "hashvalues"
- [ ] fix the bug about adding files to `.dvcignore` one file in one line.

If you prefer that these changes should be made so I would suggest I should open a new PR with these changes.

jorgeorpinel:
Sorry for the delay @sarthakforwet we're really busy this week still. Will get back to you ASAP

shcheklein:

content/docs/command-reference/move.md
jorgeorpinel: @sarthakforwet please don't mark this as resolved as no changes have been pushed yet. I can mark it as resolved when it is. Unresolving. Thanks

jorgeorpinel: > Why would one needs to move ignored files?

This is a good question. I don't remember what the motivation was for this new example originally but indeed, there's no point since you can't even `add` an ignored file, so you can't `move` it... 🧠 thinking

jorgeorpinel:
@sarthakforwet so the point of this PR was to reflect your findings for

>  Check how dvcignore interacts with dvc move

What are those findings? How do these 2 DVC features intersect? I think this new example is unnecessary as it is now, since you can't even `dvc add` ignored files (thus you can't move them). Have you tried other things like adding a file or dir first, then putting it (or parts of it) into .dvcignore and see what `dvc move` does? This is the kind of exploration we need to check first to come up with what needs to be updated in the docs.

If nothing major arises, we can just close this PR and check that checkbox in #1033

sarthakforwet:
> What are those findings?  

Basically my findings were that if we create two directories example `dir1` and `dir2` and we put file `data1` and `data2` respectively in them and we execute - 
`dvc add dir1/data1`  
`dvc add dir2/data2` 
Now if we do - 
`echo "dir1/*" >> .dvcignore`  
and finally -
`dvc move dir2/data2 dir3/data3`
`dvc move dir1/data1 dir4/data4`
then we would get an error in the second command as now DVC won't be able to find the file which ultimately shows we can't move ignored files.

jorgeorpinel:
OK but that's not what the example shows, because it's missing the dvc add step, and the error seems incorrect to me as well. What is the error when you do add them first? in any case, seems not worth an example and it should definitely not be called "Moving ignored files" since you can't do that.

But maybe we shold note somewhere in the Dvcignore guide what happens if you add files/dirs that are already being tracked by DVC. Or maybe even request that this is not allowed in the iterative/dvc repo... What do you think?

jorgeorpinel:
and can you please try other combinations like I explained above? Thanks

jorgeorpinel:
As per https://github.com/iterative/dvc.org/pull/1378#issuecomment-640016721:

> [Dvcignore ](https://dvc.org/doc/user-guide/dvcignore) needs some of these changes
> - Delete unnecessary newlines

Agree. Please submit a separate PR.

> - Change "checksums" to "hashvalues"

I don't see any instances of "checksum".

> - fix the bug about adding files to .dvcignore one pattern per line.

Can you elaborate? Not sure which bug you're talking about.

All of this is completely unrelated to this PR though, can you please create another PR or at least an issue about this? Thanks

sarthakforwet:
> OK but that's not what the example shows, because it's missing the dvc add step

Yeah agreed. Apologies! 
Actually I made the correction long before and decided to push when all possible mistakes been solved from `move.md` but due to change in hardware, I had to setup the project on my local machine again due to which the thought of pushing the updated file slipped through mind.

> What is the error when you do add them first?

We get something similar to "failed to move dir1/data1 -> dir4/data4. Unable to find DVC-file with output dir1/data1"

sarthakforwet:
> Or maybe even request that this is not allowed in the iterative/dvc repo...

Sorry I did not get this one. Where should we request and what should be not allowed?

sarthakforwet:
> Agree. Please submit a separate PR.

Sure.

>I don't see any instances of "checksum".

They were changed in updations of DVC 1.0 docs maybe. They were present before the updations and I also also wrote this at that time.

> Can you elaborate? Not sure which bug you're talking about.

If we do - 
`echo dir1/* >> .dvcignore`  and let's suppose `dir1` contains `data1` and `data1.dvc` then if we check contents of `.dvcignore` we would see - 
`dir1/data1 dir1/data1.dvc`.
Hence I proposed to change the command to  `echo "dir1/*" >> .dvcignore` . 
Maybe I shouldn't call it as a bug rather a correction.

jorgeorpinel:
I don't think that's the case or maybe it depends on the OS? I'm on Windows/ Git Bash:

image

Please double check and open a separate issue or change it in the other PR you are going to open if so.

jorgeorpinel:
> We get something similar to "failed to move dir1/data1 -> dir4/data4. Unable to find DVC-file with output dir1/data1"

OK, well I think this is a problem with the message. It could be more helpful. But that would have to be reported separately to the core repo in https://github.com/iterative/dvc/issues

I'm not sure it's worth if but you could just do so and they'll let us know what they think. But so it seems indeed the example in this PR is indeed unnecessary so let's just close this PR and open a new one if needed, if you find other things that need being documented about Dvcignore + dvc move.

jorgeorpinel:
Thanks @sarthakforwet! Good research and findings for other PRs but closing this one for now.

sarthakforwet:
> I don't think that's the case or maybe it depends on the OS? I'm on Windows/ Git Bash:

I am also using Windows/Git Bash. The example that you tried does not involve any pattern in it. 
chutiya
This is what I'm sayin..

jorgeorpinel:
Please notice I actually updated the image in https://github.com/iterative/dvc.org/pull/1378#issuecomment-652669633 after the comment was notified by email. I did use that pattern and it still shows 2 lines. I'm not sure why your terminal has this issue but I'm pretty sure the regular behavior of echo on Linux is to make new lines. If you are able to try on Linux or Max that would be nice though. I'll ask in the team here for confirmation as well...

jorgeorpinel:
UPDATE: OK yes @sarthakforwet these patterns should definitely be put in quotes since different systems can interpret  `dir/*` in a bunch of different ways. Let's wrap it in `"` please.

jorgeorpinel:
UPDATE'S UPDATE: No, let it be single quotes actually `'` please! Thanks

sarthakforwet:
> Please notice I actually updated the image in #1378 (comment) after the comment was notified by email.

I think I know the answer why this is reacting in different manner. It is because maybe you does not have any file inside `dir1` and `dir2` directories and in my case I had two files in the directory. 

Can you reproduce the same output you got by following this structure - 
 - dir1
   - file1
   - file2 

and then `echo dir1/* >> .dvcignore`.
Expected behaviour - 
`cat  .dvcignore`
`dir1/*` 

>  Let's wrap it in single quotes ' please.

Should I open a PR with this small change?

jorgeorpinel:
> because maybe you does not have any file inside dir1 and dir2 directories and in my case I had two files in the directory

Yep, `echo` expands the `*` wildcard in some situations. That's why we need single quotes around any `echo` that gets redirected, just as a best practice. I would even go as far as reviewing all the `echo` statements in our docs and add single quotes as much as possible.

> Should I open a PR with this small change?

This is up to you but would definitely be appreciated! If not, please just open a new issue about it and someone will address it eventually. Thanks!

PR #1402, opened 2020-06-08
term: DVC-file -> .dvc file (dvc run)

jorgeorpinel:
> Continuation of https://github.com/iterative/dvc.org/pull/1372#issuecomment-640461482

Work done by @utkarshsingh99


jorgeorpinel:
Boo @utkarshsingh99 FYI everything here is wrong haha but it's not your fault. `dvc run` for 1.0 has to be completely rewritten anyway. Please don't worry about this one.

PR #1410, opened 2020-06-08
Small correction in checkout example

imhardikj:

### DIFF
--- a/content/docs/command-reference/checkout.md
+++ b/content/docs/command-reference/checkout.md
@@ -135,7 +135,6 @@ The workspace looks something like this:
 ├── evaluate.dvc
 ├── featurize.dvc
 ├── prepare.dvc
-├── train.dvc
 ├── src
 │   └── ...
 └── train.dvc


shcheklein:
thanks @imhardikj !

content/docs/command-reference/checkout.md
```diff
@@ -135,7 +135,6 @@ The workspace looks something like this:
 ├── evaluate.dvc
 ├── featurize.dvc
 ├── prepare.dvc
-├── train.dvc
 ├── src
 │   └── ...
 └── train.dvc
```
jorgeorpinel, line 140: It was probably easier to read before?

imhardikj: Do you mean `train.dvc` should be above `src`?  This will have better readability sure.
I just left this same as output (displayed alphabetically).

jorgeorpinel: Yes above src. Oh well, Ivan already approved and merged this one so let's go with it. Thanks

PR #1420, opened 2020-06-10
cmd ref: dvc run 1.0 update

jorgeorpinel:
> Stolen some info from #1387

Bring `dvc run` up to date.

- [x] Include new -f behavior. See iterative/dvc/pull/4011


content/docs/command-reference/run.md
```diff
@@ -125,22 +122,44 @@ data pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
...
+- `--plots-no-cache <path>` - the same as `--plots` except that DVC does not
+  track the plots metrics file. This means that the file is not cached, so it's
+  up to the user to save and version control it. See also the difference between
+  `-o` and `-O`.
+
+- `--external` - allow outputs that are outside of the DVC repository. See
+  [Managing External Data](/doc/user-guide/managing-external-data).
+
+- `--file <path>` - specify name of the YAML file this command will generate
+  instead of the default, `dvc.yaml`.
```
skshetry, line 162: This is ignored for `dvc.yaml`. We don't have support for custom `dvc.yaml` files.

jorgeorpinel: OK thanks. How do you create a dvc.yaml file inside a subdirectory?

skshetry: Not supported at this point. `dvc.yaml` is created at the location where `dvc run` was run from.

jorgeorpinel: OK. So multiple `dvc.yaml` files are supported, you just move to another directory and use `dvc run` from there, right?

content/docs/command-reference/run.md
```diff
@@ -20,28 +22,56 @@ positional arguments:
...
+specifies it, there should be no cycles, etc.
+
+A special type of key/value dependencies,
+[parameters](/doc/command-reference/params) (`-p` option), is supported.
+Multiple parameter dependencies can be specified from one or more YAML or JSON
+parameters files (`params.yaml` by default).
+
+Special types of output files, [metrics](/doc/command-reference/metrics) (`-m`
+and `-M` options) and [plots](/doc/command-reference/plots), are also supported.
+Metrics files have specific formats (JSON, YAML, CSV, or TSV) and allow for
```
jorgeorpinel, line 65: Note I'm using "metrics files" here @shcheklein. I have been thinking about it and I think we should consider always using plural "metrics" for everything. We don't use "Metric file" because it sounds strange. Well, I think "metric files" has the same problem: "metric" out of context sounds like we're talking about a measure of distance (as in kilometers).

And BTW I also think we shouldn't call plots files metrics files which we do now because technically they are metrics, but always call them plots files.

WDYT? I would extract this to a separate issue of course. It may even impact the core repo.

shcheklein: I would say, it's definitely not the right time to discuss this. Let's do it consistent in some way. Otherwise up to you regarding the ticker or changing it everywhere - I'm fine to have it either way (though `metrics files` sound like wrong from the language perspective). You might ask @andronovhopf as well, btw.

shcheklein: plots vs metrics - I'm fine with either for now

jorgeorpinel: > metrics files sound like wrong from the language perspective

A little bit but we also talk about "pipelines file" which would become "pipelines files" if pluralized, I think it's fine.

Anyway yes, totally not the right place to discuss this, you're right. I'll change this instance to singular for consistency and open a separate ticket, ping Dmitry and Elle.

jorgeorpinel: Extracted to #1442

jorgeorpinel: But I don't think I can roll this back to "metric files" because it's really out of context. Even when the concept of "metrics" is introduced in the same paragraph, it's not obvious what "metric files" are, whereas "metrics files" makes more sense since it's the exact same word and there's no ambiguity, so you can relate the concepts.

jorgeorpinel:
Ivan if you can review up to examples for now that would be great. I'm working on those still...

content/docs/command-reference/run.md
jorgeorpinel: Actually I'm going to go with "Metrics and plots files" here anyway, ha. See 5962ce06

content/docs/command-reference/run.md
jorgeorpinel: I checked this and it is the case. Confirmed

content/docs/install/plugins.md
```diff
@@ -14,7 +14,7 @@ In order to recognize `.dvc` files as YAML in Vim, you should add:
 
 ```vim
 " DVC
-autocmd! BufNewFile,BufRead Dvcfile,*.dvc setfiletype yaml
+autocmd! BufNewFile,*.dvc setfiletype yaml
```
jorgeorpinel, line 17: Can someone who uses Vim confirm this though? Maybe @efiop (since I'm about to ping you about something else 😬)

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -63,7 +63,7 @@ An _output entry_ can consist of these fields:
 
 A _dependency entry_ consists of a these possible fields:
 
-- `path`: Path to the dependency, relative to the `wdir` path (always present)
+- `path`: Path to the dependency, relative to the `.dvc` file location
```
jorgeorpinel, line 66: @efiop or perhaps @skshetry can you please confirm this?

content/docs/install/plugins.md
efiop: @jorgeorpinel Just from looking at it, seems like it is broken and should be:
```
autocmd! BufNewFile,BufRead *.dvc setfiletype yaml
```
but I would actually keep Dvcfile for now, since it might still be present in older repos.

content/docs/user-guide/dvc-files-and-directories.md
efiop: It is still relative to the `wdir`.

content/docs/install/plugins.md
jorgeorpinel: Good point, will roll this back for now.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK... But there's no longer `wdir` in .dvc files, is there? Only in dvc.yaml

So for .dvc files it's relative to the file location and for dvc.yaml its relative to `wdir` (which defaults to the dvc.yaml file location)?

efiop: wdir is not explicitly present, but it could still be there (same with dvc.yaml). Path resolution logic stays the same.

jorgeorpinel: OK, I see. But the .dvc or dvc.yaml location is the default wdir, right?

efiop: @jorgeorpinel `wdir` is resolved relative to the dvcfile location, and `wdir` is `.`(curdir) by default, so yeah.

jorgeorpinel: Got it.

content/docs/command-reference/run.md
```diff
@@ -378,53 +378,53 @@ alternatives.
...
+├── dvc.lock
+├── dvc.yaml
 ├── result.out
-├── stages
-│   └── test.dvc
 └── test.txt
 ```
 
-## Example: Using granular hyperparameter dependencies
+## Example: Using granular parameter dependencies
```
shcheklein, line 427: let's further simplify it - just parameter dependencies - I don't see any reason to emphasize it in the title

content/docs/tutorials/deep/reproducibility.md
```diff
@@ -27,8 +27,7 @@ defined in the <abbr>project</abbr>: There's nothing to reproduce.
 $ dvc repro model.p.dvc
 ```
 
-> By default, `dvc repro` tries to read the DVC-file with name `Dvcfile`, like
-> the one we define in the previous chapter.
+> By default, `dvc repro` tries to read all stages available.
```
skshetry, line 30: It only tries to read from the `dvc.yaml` from the current directory.

content/docs/command-reference/run.md
jorgeorpinel: This title intends to emphasize that parameter dependencies are granular, in contrast to regular dependencies. I remember writing it when `params` was intruduced. How about something like "Using file contents as dependencies" ?

shcheklein: I think that `parameter dependencies` (vs just "dependencies") is already enough to emphasize everything considering that we already introduced them in the description

content/docs/command-reference/run.md
```diff
@@ -125,29 +282,50 @@ data pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
...
-  files, because they are small enough to be tracked by Git directly. See also
-  the difference between `-o` and `-O`.
-
-- `-f <filename>`, `--file <filename>` - specify a path and/or file name for the
-  stage file generated by this command (e.g. `-f stages/stage.dvc`). This
-  overrides the default file name: `<file>.dvc`, where `<file>` is the file name
-  of the first output or metric.
+- `-p [<path>:]<params_list>`, `--params [<path>:]<params_list>` - specify a set
+  of [parameter dependencies](/doc/command-reference/params) the stage depends
+  on, from a parameters file. This is done by sending a coma separated list as
```
skshetry, line 287:
```suggestion
  on, from a parameters file. This is done by sending a comma separated list as
```

content/docs/command-reference/run.md
```diff
@@ -125,29 +282,50 @@ data pipeline (e.g. random numbers, time functions, hardware dependency, etc.)
...
+
+- `--plots-no-cache <path>` - the same as `--plots` except that DVC does not
+  track the plots metrics file. This means that the file is not cached, so it's
+  up to the user to save and version control it. See also the difference between
+  `-o` and `-O`.
+
+- `--external` - allow outputs that are outside of the DVC repository. See
+  [Managing External Data](/doc/user-guide/managing-external-data).
+
+- `--file <path>` - specify name of the YAML file this command will generate or
```
skshetry, line 321: Not true. `--file` is ignored for any stages created with `-n`.

content/docs/command-reference/run.md
jorgeorpinel: Good point. OK, changed in 1345ca4d

content/docs/tutorials/deep/reproducibility.md
jorgeorpinel: OMG yes this tutorial is very outdated. You're totally right but out of scope for this PR. I was just removing the Dvcfile term here, I think. Thanks though!

content/docs/command-reference/run.md
jorgeorpinel: Yeah that's right but so hmmm what does --file do? This is what the master branch of the core repo outputs:

`  --file <filename>     Specify name of the DVC-file this command will generate.`

We don't even have DVC-files anymore!

jorgeorpinel: Also

> ges created with -n

-n is always required no? Lil confused rn

skshetry: Oops, looks like `--single-stage` is hidden. `--single-stage` flag is available in `dvc run` that keeps the same behavior as the 0.X dvc had, regarding single stage files. This `-f` is a vestige of that. We might reuse `-f` if we introduce custom named dvc.yaml file someday. But, for `dvc.yaml`, it's ignored, so, we can just drop it from the documentation altogether.

skshetry: Wait, this is a command reference. So, maybe, we should document `-f` and `--single-stage` here too.

content/docs/tutorials/deep/reproducibility.md
skshetry: Still, the change you made is incorrect. Please fix or rollback.

content/docs/command-reference/run.md
jorgeorpinel: Interesting! I didn't know about `--single-stage`.

> for dvc.yaml, it's ignored, so, we can just drop it from the documentation altogether

Yep, I'll just remove it for now. But it should probably be hidden from the help output in core too? Idk 😕

shcheklein: @skshetry @efiop let's hide in DVC core if --single-stage is hidden.

shcheklein: ... and it's ignored for dvc.yaml

content/docs/tutorials/deep/reproducibility.md
jorgeorpinel: This is fine don't worry about it. The whole tutorial is going away soon. Resolving. Thanks

shcheklein:
Merged since we are releasing DVC 1.0 tmr. Please let's keep iterating in the new PR

jorgeorpinel:
Continuation PR: #1479

content/docs/command-reference/run.md
jorgeorpinel: Created #4084

Issue #1414, opened 2020-06-09
cmd: pending plots changes by 1.0 release


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/1382#issuecomment-640435301 et al.

- [x] Apply changes from iterative/dvc/pull/3963 once that is merged. - https://github.com/iterative/dvc.org/issues/1414#issuecomment-730566207
- [x] Add YAML and TSV examples to plot show? Or maybe just translate JSON example to YAML and CSV to TSV.
- [x] Per https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427733327 double check about "All plots are used" in plot show -h usage (targets arg). - https://github.com/iterative/dvc.org/issues/1414#issuecomment-730558360
- [x] modify: how to make a regular output into a plot? See https://github.com/iterative/dvc/pull/3994#pullrequestreview-429085947 - #1946
- ~~Review term "props" in plots commands once the code base is final/stable (per https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427522723)~~
- [x] term: "column index" vs "column number"? See https://github.com/iterative/dvc.org/pull/1382#discussion_r438996574 and https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427729775 - #1946
- [x] For JSON plot files we say "every object should have the same structure" but actually I'm not sure what the exact restriction is. Ask/test/check. See https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427730625
- ~~Double check how to apply all Vega spec props in plots: modify subcmd only has a limited no. of options. Why isn't it like `dvc config` BTW, where you can give any key/value pair (core repo Q)?~~ Moved to iterative/dvc/issues/4909
---

- EXTRA: Possibly change the modify example explained in https://github.com/iterative/dvc.org/pull/1382#discussion_r436435646
- EXTRA: Maybe address some of the discrepancies mentioned in this https://github.com/iterative/dvc.org/pull/1380#issuecomment-640181973
- EXTRA: Is plots diff really a diff? See discussion in https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427524008
- EXTRA: See #1800

jorgeorpinel:
@imhardikj some answers to https://github.com/iterative/dvc.org/pull/1946#issuecomment-727563679:

> > Apply changes from iterative/dvc/pull/3963
>
> These changes are already applied

[...]

PR #1444, opened 2020-06-17
cmd ref: metrics 1.0 update

nik123:
This PR updates cmd ref according to https://github.com/iterative/dvc/pull/3963

* cmd ref: metrics diff: targets desc updated per https://github.com/iterative/dvc/pull/3963

* cmd ref: metrics show: targets desc updated per https://github.com/iterative/dvc/pull/3963

Part of #1414


shcheklein:
Thanks @nik123 !

Issue #1289, opened 2020-05-13
guide: document run-cache


efiop:
So far affected commands are run/repro which create and/or use run cache. And `push/pull/fetch/potentially status`) that transfer ~~(options for that are hidden for now)~~ run cache to/from remote. Run cache is located in `.dvc/cache/runs`.

efiop:
Will also need to update run-cache flags in run/repro to explain them through the run-cache concept.

sarthakforwet:
Hey there I wanna work on this issue . Could you assign me this ?

efiop:
Hi @sarthakforwet ! I don't think it is suitable for contribution, as it is a WIP feature. Please check out `need-help` and `good-first-issue` labels, they should have something more suitable :slightly_smiling_face:

jorgeorpinel:
Hi @sarthakforwet indeed please further explore the issues for something that seems a little smaller and more defined. I believe I sent some specific links in my GSoD email reply to you to some recommended issue labels. Feel free to reach out to me directly via email or on Discord for help on issue selection it you're still unsure. Thanks!

jorgeorpinel:
A suggested place to start is to add an "execution and reproduction" section to the `dvc run` cmd ref. similar to this: https://github.com/iterative/dvc.org/pull/1464/files

jorgeorpinel:
More info that would be good to include for this ticket, from https://github.com/iterative/dvc.org/pull/1448#discussion_r443124577:

> 1. Run-Cache can generate outputs in a suitable condition and that too indirectly (eg: only on run/repro), whereas a lockfile will always provide you with outputs without rerunning.
> 2. RunCache can generate lockfile. Even though you have a run-cache, you should still use lockfile because for the outputs.
> 3. get/import commands work well only if you have a lockfiile.

(Thanks @skshetry)

jorgeorpinel:
Hello! So this continues to be a major discrepancy between released DVC and published docs. I know it's being worked on but at least basic mentions of this should exist in the docs for now. Is it still the plan for someone from engineering to submit a draft proposal adding such mentions? Thanks! Cc @efiop @shcheklein

efiop:
@jorgeorpinel Likely no drafts until experiments are here or https://github.com/iterative/dvc/issues/4223 .

jorgeorpinel:
Hi @efiop! I'm finally finishing up some basic docs for this (sort of needed to doc `dvc exp`). Quick questions:

> potentially status

[...]

PR #1448, opened 2020-06-18
user guide: dvc.lock section of Files & Dirs

### LINKED ISSUE #1422: user guide: document dvc.lock (DVC 1.0)
Extracted from https://github.com/iterative/dvc.org/pull/1370#pullrequestreview-427482603

> - [x] ...dvc.yaml cannot contain md5 or etag at all. They are in dvc.lock.

Also from https://github.com/iterative/dvc.org/pull/1370#pullrequestreview-427482065:

> - [x] Strictly speaking, md5 is optional. User could totally remove it and dvc will be fine.
> - [x] dvc.yaml does not have md5, it's in the dvc.lock. For .dvc files, in case of --no-commit for dvc.yaml stages, there should not be a lockfile changes at all.

jorgeorpinel:
Closes #1422


content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -149,6 +155,56 @@ the possible following fields:
 
 `dvc.yaml` files also support `# comments`.
 
+> Note that unlike `.dvc` files, `dvc.yaml` doesn't store any `md5` hash values,
```
shcheklein, line 158: let's remove this note ... people don't expect those md5 to be stored in the dvc.yaml in the first place (only user who had experience with DVC < 1.0)

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -149,6 +155,56 @@ the possible following fields:
...
+> making it extra easy to edit manually. These are kept in `dvc.lock` instead.
+
+### dvc.lock file
+
+The purpose of `dvc.lock` is to track the latest version of all files involved
```
shcheklein, line 163: track -> pin, capture (check docs for Gemfile.lock, or yarn.lock, etc)

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -149,6 +155,56 @@ the possible following fields:
...
+    metrics:
+      - performance.json
+  ...
+    plots:
+      - logs.csv:
+          x: epoch
+          x_label: Epoch
+```
+
+It replicates the structure of `dvc.yaml`, but with some key differences and
```
shcheklein, line 191: can we summarize the difference in some simpler way? w/o a long list that repeats everything else. What is the actual difference - only md5 and etag fields for outputs? anything else?

cc @skshetry

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: it also plays a role of `.dvc` file for pipeline ouputs.

I guess we need to be more explicit here.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Removed, but I think since there's `md5` in .dvc files, some users may expect them in dvc.yaml

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > pin, capture (check docs for Gemfile.lock, or yarn.lock

How about "lock"? 🙂  I checked https://docs.npmjs.com/files/package-lock.json and came up with a new intro. PTAL

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I only listed the differences.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > plays a role of .dvc file for pipeline ouputs

Should I add a note to connect these 2 concepts? Per https://github.com/iterative/dvc.org/pull/1448#pullrequestreview-433572182 I don't think that's necessary.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,56 @@ the possible following fields:
...
+      - path: data/clean
+        md5: d8b874c5fa18c32b2d67f73606a1be60
+    params:
+      params.yaml:
+        levels.no: 5
+    outs:
+      - path: features
+        md5: 2119f7661d49546288b73b5730d76485
+    metrics:
+      - performance.json
```
skshetry, line 187: No metrics or plots in dvc.lock.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: Yes, for deps and outs, `md5/checksum/etag` and `path` are present in dvc.lock. For params, it's a key-value pair of {file, values}.

skshetry: I don't think, we need to explain the structure of `dvc.yaml` at all. No one needs to edit it, and the differences are clear from above example.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,56 @@ the possible following fields:
 
 `dvc.yaml` files also support `# comments`.
 
+### dvc.lock file
```
skshetry, line 165: Should user commit lockfile to the VCS?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,56 @@ the possible following fields:
...
+
+The purpose of `dvc.lock` is to lock the exact contents of all files involved in
```
skshetry, line 167: Needs a description for what this lock provides for the `dvc.`.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Agree with both of you. That's what I have done.

> md5/checksum

Can you please clarify the field name @skshetry ? Using DVC 1.0 I've only seen `md5`. Is `checksum` the name of the field going forward or in certain cases? Thanks

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Oops, thanks. Fixed in dcd09105! (Moved to `outs`.)

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: It's a good note but I'm not sure we want to talk about that in this doc. We would need similar notes on all the other types of docs too. WDYT @shcheklein ?

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Sorry, I don't understand your comment. The "dvc." ?

jorgeorpinel:
LGTM?

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: @skshetry could you summarize it? I mean the value for DVC

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: @skshetry please, give some suggestion on this?

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: still too verbose to my mind @jorgeorpinel . I still don't think we need a list of params here.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,53 @@ the possible following fields:
...
+    dependencies</abbr>)
+- `params`: List of [parameters files](/doc/command-reference/params#examples)
+  (typically only `params.yaml`), as well as sub-lists of the specific
+  parameters tracked by this stage (per params file), and their latest value.
+- `outs`: Each item of the list becomes an object with 2 properties:
+
+  - `path`: Same as the `deps` value from `dvc.yaml`
+  - `md5`: The hash value of the file
+
+  This includes any [metrics files](/doc/command-reference/metrics) or
```
shcheklein, line 208: not sure I understand this tbh

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,53 @@ the possible following fields:
 
 `dvc.yaml` files also support `# comments`.
 
+### dvc.lock file
+
+The purpose of `dvc.lock` is to lock the exact contents of all files involved in
+the pipeline(s) at the time that `dvc.yaml` was last updated.
```
shcheklein, line 168: This does not sound like correct to me ... e.g.  I updated the `dvc.yaml` manually?

Also, `dvc.lock` also locks some other fields - params, not only files

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK. I summarized it into a few short paragraphs instead of making another list (in b6d68dab). PTAL @shcheklein

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: This no longer applies (it's been rewritten anyway) per https://github.com/iterative/dvc.org/pull/1448#discussion_r443045904. Resolving

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > This does not sound like correct to me ... e.g. I updated the dvc.yaml manually?

Good point. Rephrased to "at the time that the pipeline was last committed
(after using `dvc add`, `dvc run`, `dvc repro`, or `dvc commit`, for example)" in 2560ffed. Better?

> dvc.lock also locks some other fields - params, not only files

Yeah actually params files are not locked by md5. Only the param key/value pairs. This was already explained in my previous commit.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,45 @@ the possible following fields:
...
+separately under `params`, grouped by parameters file. (No hash value is needed
+to lock params files, as these are not tracked as a whole.)
+
+Here's an example `dvc.lock` based on the one in [`dvc.yaml`](#dvcyaml-file)
+above:
+
+```yaml
+stages:
+  features:
+    cmd: jupyter nbconvert --execute featurize.ipynb
```
jorgeorpinel, line 188: Question @skshetry: Why do we need the `cmd` in dvc.lock? Thanks

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,45 @@ the possible following fields:
 
 `dvc.yaml` files also support `# comments`.
 
+### dvc.lock file
+
+For every `dvc.yaml` file, a corresponding `dvc.lock` (YAML) file is created or
+updated by DVC as well. Its purpose is to lock the exact contents of all files
```
shcheklein, line 168: it's not only about files I think

shcheklein: it's not the only purpose - it also replaces .dvc files for pipeline outputs for example

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,45 @@ the possible following fields:
...
+involved in the pipeline(s) at the time that the pipeline was last committed
+(after using `dvc add`, `dvc run`, `dvc repro`, or `dvc commit`, for example).
+
+This means that regular <abbr>dependencies</abbr> and all <abbr>outputs</abbr>
+(including [metrics](/doc/command-reference/metrics) and
+[plots](/doc/command-reference/plots) files) are also listed in `dvc.lock`, but
+with additional fields that store the `md5` hash value (or `etag` for HTTP
```
shcheklein, line 175: let's just use hash, w/o going to deep (like etag for HTTP, etc), especially since you include an example

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,45 @@ the possible following fields:
...
+<abbr>external dependencies</abbr>) of each file or directory tracked by DVC.
+
+[parameter](/doc/command-reference/params#examples) key/value pairs are listed
```
shcheklein, line 178: not sure we need that verbose descritpion

content/docs/user-guide/dvc-files-and-directories.md
skshetry: It is used to find out if `cmd` for stage changed, by comparing the `cmd` between `.lock` and `.yaml`. Kinda like hash that we had before.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: Also, it might lock the values of params, not just files.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: > You should commit the `dvc.lock` file to your project repo so that all team-members can reuse the outputs of your experiment without re-running.

skshetry: Note that we might need to explain difference with `run-cache` later and might need to compare these two. 

1. Run-Cache can generate outputs in a suitable condition and that too indirectly (eg: only on `run/repro`), whereas a lockfile will always provide you with outputs without rerunning.
2. RunCache can generate lockfile. Even though you have a run-cache, you should still use lockfile because for the outputs. 
3. `get`/`import` commands work well only if you have a lockfiile.

skshetry: But, this can be added later when documenting run-cache.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: @jorgeorpinel, `checksum` was always there. It's used for `hdfs`. For http(s)/s3/azure/gs, we have `etag`, for `ssh/local` we have `md5` and for `hdfs`, we use `checksum` key.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: Something like this, but maybe a bit redundant:

> Using `dvc.lock`, `dvc` can provide support for checkouts for the outputs of the experiments and even allows importing that particular version of the outputs to a different repo via `get`/`import` ...

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: > Yeah actually params files are not locked by md5. 

they are locked by values, values play the same role as md5s

> "at the time that the pipeline was last committed

no, sounds still confusing ... I think the whole approach should be different? For every stage in the pipeline it captures specific versions of deps, outs, etc involved. It serves a few purposes - detect changes (e.g. on repro, or status), the same as `.dvc` files for outputs and intermediate pipeline results - for commands like checkout to work, for get and import to work, etc

`dvc add` probably does not affect it, btw.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I've never seen `checksum` nor does it exist in docs 😱 and currently we only mention eTag for HTTP I think — will have to document all this... Created #1469

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Interesting. Is there a run-cache issue we can add all these comments to? Even if it's in the core repo.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: And actually I addressed it for this PR's scope in f57f830e. Leaving this unresolved for future reference.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Gotcha. I feel like those benefits can be split between dvc.yaml and dvc.lock really. Also, this part has already changed considerably and we have 3 paragraphs and a bullet list already to explain dvc.lock, so we may have to pick between adding this info or keeping it relatively brief. Up to you @shcheklein please lmk. I'm down to add this info.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > they are locked by values, values play the same role as md5s

Not exactly, you can add any contents to a params file, change indentation or spacing, maybe add comments, etc. and nothing will change from for `dvc status`. That's why I'm saying the params file itself is not locked to (tracked).

Anyway, will review this again ⏳

jorgeorpinel: > it captures specific versions of deps, outs, etc

Idk if mentioning "versions" here is best. DVC doesn't actually provide versioning: you create versions by 1. tracking stuff with DVC and 2. committing the meta files to Git. That's the way I've been seeing it recently (since I began writing the new Get Started).

But I'll review this text along these lines:

> It serves a few purposes - detect changes (e.g. on repro, or status), the same as .dvc files for outputs and intermediate pipeline results - for commands like checkout to work, for get and import to work, etc

as well as trying to incorporate Saugat's suggestion from https://github.com/iterative/dvc.org/pull/1448#discussion_r443125758 (on the value of dvc.lock).

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Ok so per https://github.com/iterative/dvc.org/pull/1448#discussion_r443167630 I went ahead and incorporated this info in d7500f4a. PTAL guys. Resolving here.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > detect changes (e.g. on repro, or status), the same as .dvc files for outputs

`.dvc` files help detect changes? Only for import stages which have dependencies, I think.

Anyway, I just updated this whole intro to dvc.lock in d7500f4a. PTAL

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Makes sense. Info added.

jorgeorpinel: Thanks!

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: This has already been rewriten (PTAL) but it OK, what does dvc.lock lock to exactly? What's the broader concept? Saying "locks to the exact content of files, and  parameter key/value pairs" seems too specific for an introductory paragraph.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Hmm per https://github.com/iterative/dvc.org/pull/1448#discussion_r443124777 this is already a new bullet list actually. Isn't this the document where we want to have the complete tech spec of what dvc.lock contains? The `checksum` field, for example, is not even mentioned anywhere else in docs yet.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK. Can you PTAL now? First there's a description of what dvc.lock is for, then the example, and at the bottom, these verbose descriptions of what's in it: stage commands, deps and outs and a bullet list of kinds of file hashes, and params key/value pairs.

jorgeorpinel: Is all the bottom text after the example unnecessary in your opinion?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -16,11 +16,14 @@ Additionally, there are two special kind of files created by certain
   that form the pipeline(s) of a project, and their connections (_dependency
   graph_ or DAG).
 
+  These come with a matching [`dvc.lock` file](#dvclock-file) to track the
+  latest file versions.
```
shcheklein, line 20: it tracks not only file versions

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -49,21 +52,25 @@ meta:
 - `deps`: List of <abbr>dependency</abbr> entries for this stage, only present
   when `dvc import` and `dvc import-url` are used. Typically there is only one
   (but several can be added manually).
+- `wdir` (optional): Working directory for the stage command to run in (relative
```
shcheklein, line 55: not sure we need to specify `optional` here

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -83,10 +90,15 @@ Note that comments can be added to `.dvc` files and `dvc.yaml` using the
...
 
-## dvc.yaml files
+## dvc.yaml file
 
 When you add commands to a pipeline with `dvc run`, the `dvc.yaml` file is
-created or updated. Here's a simple example:
+created or updated. `dvc.yaml` files describe data pipelines, similar to how
+[Makefiles](https://www.gnu.org/software/make/manual/make.html#Introduction)
+work for building software. Additionally, a [`dvc.lock`](#dvclock-file) file is
+also created to record the latest file versions.
```
shcheklein, line 99: not only files

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,67 @@ the possible following fields:
 
 `dvc.yaml` files also support `# comments`.
 
+### dvc.lock file
+
+For every `dvc.yaml` file, a corresponding `dvc.lock` (YAML) file is created or
+updated by certain DVC commands (`dvc run`, `dvc repro`, `dvc commit`). It
+describes the latest contents of the files involved in the pipeline(s). It has
```
shcheklein, line 169: it's not about files only

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,67 @@ the possible following fields:
...
+several purposes:
+
+- Basic tracking of intermediate and final results of a pipeline (similar to
```
shcheklein, line 172: I would remove `Basic`?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,67 @@ the possible following fields:
...
+  [`.dvc` files](#dvc-files))
+- Allow DVC to detect when stage definitions, or their dependencies have
+  changed. Such conditions invalidate states, requiring their reproduction (see
+  `dvc status`, `dvc repro`).
+- Provides a mapping of file and directory paths tracked in the
```
shcheklein, line 177: this sounds like a part of the data tracking (fist point in the list), or I'm missing something

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +162,67 @@ the possible following fields:
...
+- `md5` is the most common type of file hash, used for local file system
+  dependencies, as well as SSH <abbr>external dependencies</abbr>.
+- `etag` is used for HTTP, S3, Azure, and Google Cloud external dependencies.
+- `checksum` is only used for HDFS external dependencies.
+
+> Refer to [External Dependencies](/doc/user-guide/external-dependencies) for
+> more information.
+
+[Parameter](/doc/command-reference/params#examples) key/value pairs are listed
+separately under `params`, grouped by parameters file. (No hash value is needed
```
shcheklein, line 223: `(No has  ` part can be confusing and probably is not needed

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I did change it to "hash value" in the paragraph though. Resolving this one.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Moved to https://github.com/iterative/dvc.org/issues/1289#issuecomment-648595194

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I'm going to go with "pipeline state" for now.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Good point. Not even about versions either. Using "record the pipeline state" now.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I think we do because wdir is not added by any command to this file anymore. But you can add it manually (like meta).

jorgeorpinel: Although... Unlike meta, wdir has an effect on the way DVC interprets the file. So not sure. Maybe "optional" just isn't the best word for either field.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Updated in d29e8d81

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Yes. Updated in 664351a1

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Removed.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: No, you're right. It's repetitive. The intention of this bullet was to say that it's needed by a bunch of commands to work, let me try to rewrite these 2 bullets... ⏳

jorgeorpinel: Updated in 9638a131. PTAL

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Removed.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +168,64 @@ the possible following fields:
...
+### dvc.lock file
+
+For every `dvc.yaml` file, a matching `dvc.lock` (YAML) file usually exists.
+It's created or updated by DVC commands such as `dvc run` and `dvc repro`.
+`dvc.lock` describes the latest pipeline state. It has several purposes:
+
+- Tracking of intermediate and final results of a pipeline — similar to
+  [`.dvc` files](#dvc-files). This is done with a mapping of file and directory
+  paths in the <abbr>project</abbr> to their locations (hash values) in the
+  <abbr>cache</abbr> or [remote storage](/doc/command-reference/remote).
```
jorgeorpinel, line 180: I wonder if this explanation is too long. Could remove the whole part about mapping between locations (hash values) to cache or remote storage. Although it's important and not sure it's stated this clearly in any other doc...

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +168,64 @@ the possible following fields:
...
+```
+
+Stage commands are listed again in `dvc.lock`, in order to know when their
+definitions change in the `dvc.yaml` file.
+
+Regular <abbr>dependencies</abbr> and all kinds of <abbr>outputs</abbr>
+(including [metrics](/doc/command-reference/metrics) and
+[plots](/doc/command-reference/plots) files) are also listed (per stage) in
+`dvc.lock`, but with an additional field to store the hash value of each file or
+directory tracked by DVC:
```
jorgeorpinel, line 216: ...on the other hand, if we keep the explanation about mapping to the cache/storage, should we mention it here again or even provide a small example of the corresponding .dvc/cache/ `tree` ?

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: yes, I would say that it's a similar mechanism and we can do the explanation somewhere else - e.g. we have a cache structure section somwhere

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +168,64 @@ the possible following fields:
...
+
+- `md5` is the most common type of file hash, used for local file system
+  dependencies, as well as SSH <abbr>external dependencies</abbr>.
+- `etag` is used for HTTP, S3, Azure, and Google Cloud external dependencies.
+- `checksum` is only used for HDFS external dependencies.
```
shcheklein, line 221: we don't need this list - it's excessive I think - we can just embed in

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -138,6 +168,64 @@ the possible following fields:
...
+
+> Refer to [External Dependencies](/doc/user-guide/external-dependencies) for
```
shcheklein, line 223: let's remove this - it's also excessive here

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -5,22 +5,28 @@ directory (`.dvc/`) with the
...
   tracked.
 - The [`dvc.yaml` file](#dvcyaml-file) or _pipeline(s) file_ specifies stages
   that form the pipeline(s) of a project, and their connections (_dependency
   graph_ or DAG).
 
-Both use human-friendly YAML schemas, described below. We encourage you to get
-familiar with them so you may edit them freely, as needed. Both type of files
-should be versioned with Git (for Git-enabled <abbr>repositories</abbr>).
+  These typically come with a matching [`dvc.lock` file](#dvclock-file) to
+  record the pipeline state.
```
shcheklein, line 20: and track pipeline data artifacts (vs creating a new .dvc file for each of those)

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -5,22 +5,28 @@ directory (`.dvc/`) with the
...
 
-## .dvc file
+Both `.dvc` files and `dvc.yaml` use human-friendly YAML schemas, described
```
shcheklein, line 22: mention `dvc.lock` again? that it is auto generated?

content/docs/user-guide/dvc-files-and-directories.md
shcheklein: yes, let's just omit it for simplicity. We can mention explicity that this field can be added manually.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK, removing.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: This no longer applies per https://github.com/iterative/dvc.org/pull/1448#discussion_r445116497

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Done.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Removed.

shcheklein:
A conflict to resolve ^^ and we are good to merge I think. Good stuff!

shcheklein:
A couple of unresolved comments still? ^^ Up to you.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK, added in 7022368f. But "record the pipeline state" is supposed to include that, conceptually.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: dvc.lock doesn't have a human-friendly YAML schema 😋  It's included in the following paragraph though: "All these should be versioned with Git..."

PR #1471, opened 2020-06-21
Added a note that gc might try to pull missing .dir cache files

### LINKED ISSUE #1075: cmd-ref: add a note that gc might try to pull missing .dir cache files
https://discordapp.com/channels/485586884165107732/563406153334128681/690239392094486641

imhardikj:
Fixes #1075 
Changes as per observations in https://github.com/iterative/dvc.org/issues/1075#issuecomment-643669420

### DIFF
--- a/content/docs/command-reference/gc.md
+++ b/content/docs/command-reference/gc.md
@@ -28,11 +28,12 @@ corresponds to keeping the data for the current workspace, and for a certain set
 of commits (determined by reading the DVC-files in them). See the
 [Options](#options) section for more details.
 
-> Please note that `dvc gc` pulls
-> [`.dir` cache](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
-> files for corresponding directories from remote. It pulls them when they are
-> missing and referenced in the commits that `gc` analyzed in `.dvc` or
-> `dvc.yaml` files.
+> Note that `dvc gc` tries to fetch any missing
+> [`.dir` files](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
+> from [remote storage](/doc/command-reference/remote) to the local
+> <abbr>cache</abbr>, in order to know which files should exist inside cached
+> directories. These files may be missing if the cache directory was previously
+> garbage collected, in a newly cloned copy of the repo, etc.
 
 Unless the `--cloud` option is used, `dvc gc` does not remove data files from
 any remote. This means that any files collected from the local cache can be
--- a/content/docs/command-reference/gc.md
+++ b/content/docs/command-reference/gc.md
@@ -42,6 +42,9 @@ most cases removing data locally and in remote storage is irreversible.
 The default remote is cleaned (see `dvc config core.remote`) unless the
 `--remote` option is used.
 
+> Please note that `dvc gc` might try to pull `.dir` <abbr>cache</abbr> files
+> for corresponding directories from remote if they are missing.
+
 ## Options
 
 - `-w`, `--workspace` - keep files and directories _only_ referenced in the
--- a/content/docs/command-reference/gc.md
+++ b/content/docs/command-reference/gc.md
@@ -20,13 +20,18 @@ DVC cache but are no longer needed. With `--cloud` it also removes data in
 To avoid accidentally deleting data, it raises an error and doesn't touch any
 files if no scope options are provided. It means it's user's responsibility to
 explicitly provide the right set of options to specify what data is still needed
-(so that DVC can figure out what fils can be safely deleted).
+(so that DVC can figure out what files can be safely deleted).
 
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/gc.md
```diff
@@ -42,6 +42,9 @@ most cases removing data locally and in remote storage is irreversible.
 The default remote is cleaned (see `dvc config core.remote`) unless the
 `--remote` option is used.
 
+> Please note that `dvc gc` might try to pull `.dir` <abbr>cache</abbr> files
```
shcheklein, line 45: - This note applies to the general case - consider moving it up
- May be we should change the language from "might" to "pulls"?
- Consider including a link to the https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory - it contains information about `.dir` files.
- Consider being more explicit - it pulls them when they are missing and they are referenced in the commits GC analyzed in `.dvc` or `dvc.yaml` files.

content/docs/command-reference/gc.md
```diff
@@ -20,14 +20,20 @@ DVC cache but are no longer needed. With `--cloud` it also removes data in
...
-(so that DVC can figure out what fils can be safely deleted).
+(so that DVC can figure out what files can be safely deleted).
 
 One of the scope options (`--workspace`, `--all-branches`, `--all-tags`,
 `--all-commits`) or a combination of them must be provided. Each of them
 corresponds to keeping the data for the current workspace, and for a certain set
 of commits (determined by reading the DVC-files in them). See the
 [Options](#options) section for more details.
 
+> Please note that `dvc gc` pulls `.dir` >
```
imhardikj, line 31: @shcheklein pre commit hook is adding ">" at end of this line.

jorgeorpinel: Hi. This is resolved right? Doesn't seem correct.

content/docs/command-reference/gc.md
```diff
@@ -20,14 +20,20 @@ DVC cache but are no longer needed. With `--cloud` it also removes data in
...
 `--all-commits`) or a combination of them must be provided. Each of them
 corresponds to keeping the data for the current workspace, and for a certain set
 of commits (determined by reading the DVC-files in them). See the
 [Options](#options) section for more details.
 
+> Please note that `dvc gc` pulls
+> [`.dir` cache](/doc/user-guide/dvc-files-and-directories#structure-of-cache-directory)
+> files for corresponding directories from remote. It pulls them when they are
+> missing and referenced in the commits that `gc` analyzed in `.dvc` or
+> `dvc.yaml` files.
```
jorgeorpinel, line 35: This note is directly about `dvc gc` so no need for the block quote (no `>`s please).

jorgeorpinel: Actually no... It's kind of a very technical implementation detail so the block quotes can stay. Never mind!

jorgeorpinel, line 35: > Note that `dvc gc` tries to fetch any missing [`.dir` files]{linked} from [remote storage]{linked} to the local \<abbr>cache\</abbr>, in order to know which files should exist inside cached directories. These files may be missing if the cache directory was previously garbage collected, in a newly cloned copy of the repo, etc.

jorgeorpinel: I'm not sure about this part:

> It pulls them when they are missing and referenced in the commits that `gc` analyzed in `.dvc` or `dvc.yaml` files.

jorgeorpinel: Is that trye @efiop ? ☝️  And also, is the operation performed recursively as deeply as needed? Thanks

content/docs/command-reference/gc.md
imhardikj: I have added `.dir` to next line. The problem seems with using backtick " ` " at end of line(in a note). Pre commit hook doesn't except this.

jorgeorpinel: Strange. Never saw that before. Good to know...

content/docs/command-reference/gc.md
efiop: @jorgeorpinel Yes, looks correct. 

> And also, is the operation performed recursively as deeply as needed? 

Not sure what you mean here.

shcheklein:
Thanks : 🙏

content/docs/command-reference/gc.md
jorgeorpinel: I meant: Does it pull .dir files recursively from cache? .dir files mapped inside .dir files and so on — as many as needed.

efiop: > .dir files mapped inside .dir files and so on — as many as needed

That doesn't ever happen, so it doesn't.

PR #1482, opened 2020-06-23
Get started 1.0 - experiments (feedback) et al.

jorgeorpinel:
Continuation of https://github.com/iterative/dvc.org/pull/1472#pullrequestreview-435326282


jorgeorpinel:
I like your version of the Get Started @shcheklein 👍

content/docs/start/experiments.md
```diff
@@ -43,71 +42,87 @@ evaluate:
...
+[`precision_recall_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_recall_curve.html))
+into plots file `prc.json`:
 
 ```json
 {
   "prc": [
     { "precision": 0.021473008227975116, "recall": 1.0, "threshold": 0.0 },
-    ...{ "precision": 1.0, "recall": 0.009345794392523364, "threshold": 0.64 }
+    ...
+    { "precision": 1.0, "recall": 0.009345794392523364, "threshold": 0.64 }
```
jorgeorpinel, line 74: The formatter wants this to be broken though. Since `...` isn't valid JSON. Comments are also not supported.

shcheklein: try to put `'...',`?

content/docs/start/experiments.md
```diff
@@ -43,71 +42,87 @@ evaluate:
...
+parameters.
 
 ## Defining parameters
 
-It's pretty common for data processing pipelines to use a separate YAML or JSON
-config file that defines adjustable parameters you use to train a model, do
-pre-processing, etc. DVC provides a mechanism for a stage to depend on values
-from such a config file. That's how stage is defined in the `dvc.yaml` file:
+It's pretty common for data science pipelines to include configuration files
+that defines adjustable parameters to train a model, do pre-processing, etc. DVC
```
shcheklein, line 98: you making everything a bit too formal to my taste (here, in some other places) :)

like config file vs configuration files.

anyways, `defines` -> `define` in your case

content/docs/start/experiments.md
```diff
@@ -43,71 +42,87 @@ evaluate:
...
+```
 
 <details>
 
-### 💡 Expand to recall how it was generated
+### 💡 Expand to recall how it was generated.
 
-As a reminder, the `featurize` stage was created with this command. No need to
-run it again, but pay more attention to the `-p` option this time:
+The `featurize` stage
```
shcheklein, line 122: here as well ... it's not CR, informal language is fine ... I wish we had even more of this

content/docs/start/experiments.md
```diff
@@ -155,9 +158,9 @@ train:
 
 ## Tuning and running experiments
 
-We are definitely not happy with the `AUC` value, we got so far! Let's now tune
-and run the new experiment. Edit the `params.yaml` file to use bigrams and
-increase number of features:
+We are definitely not happy with the `AUC` value we got so far (`0.573`)! Let's
```
shcheklein, line 161: let's not put precise values - it's hard to maintain (it's clear that you are never happy with it :) )

content/docs/start/experiments.md
```diff
@@ -167,25 +170,25 @@ increase number of features:
...
+code, update datasets — you do the changes, use `dvc repro`, and DVC runs what
+needs to be run.
 
 ## Comparing experiments
 
-Finally, we are now ready to compare experiments. DVC has a few commands to see
-metrics and parameter changes, visualize plots for one or more experiments.
-Let's compare the last "bigrams" run with the last committed "baseline"
-iteration:
+Finally, we are now ready to compare results. DVC has a few commands to see
```
shcheklein, line 188: we compare everything, not results only

content/docs/start/experiments.md
jorgeorpinel: That works, strangely. Still invalid JSON I would think. Oh well, thanks.

content/docs/start/experiments.md
jorgeorpinel: I do have a weakness for using complete words once at least, I'm OK with abbreviations later. Probably I shouldn't force this sort of thing on other people's text, sorry. It's just a habit at this point.

And yes, changing to "define". Thanks

content/docs/start/experiments.md
jorgeorpinel: I think I only removed "As a reminder" here, because you mentioned the different Get Started pages are not supposed to be connected to each other necessarily. Instead I linked to that page. "pay more attention" sounds strange to me, like we're telling the user they're not paying enough attention 😋 (angry teacher).

content/docs/start/experiments.md
jorgeorpinel: Removed. I added it as a reminder as well, since that section is not immediately before this text. But I guess the user should still remember about AUC, at least if they read/follow this whole page at once.

jorgeorpinel: Should we link `[we got so far](#collecting-metrics)` though?

content/docs/start/experiments.md
jorgeorpinel: Agree. We had "compare experiments" before and I thought "results" is a more general concept. Should we roll back? Or just say "ready to compare everything"?

jorgeorpinel: Going for "compare everything!" for now.

jorgeorpinel:
Merging. But I can open another one if you have any more feedback per our discussions above @shcheklein. Thanks

PR #1493, opened 2020-06-24
cmd ref: freeze/unfreeze 1.0 and improvements

jorgeorpinel:
Continuation of #1488


content/docs/command-reference/freeze.md
```diff
@@ -1,32 +1,32 @@
...
+(non-frozen) stages only.
 
-Note that <abbr>import stages</abbr> are considered always frozen. Use
-`dvc update` to update the corresponding <abbr>data artifacts</abbr> from the
-external data source. [Unfreeze](/doc/command-reference/unfreeze) them before
-using `dvc repro` on a pipeline that needs their outputs.
+Note that <abbr>import stages</abbr> are frozen by default. Use `dvc update` to
+update the corresponding <abbr>data artifacts</abbr> from the external data
+source. [Unfreeze](/doc/command-reference/unfreeze) them before using
+`dvc repro` on a pipeline that needs their outputs.
```
jorgeorpinel, line 29: Is this paragraph correct? Are import stages frozen by default, but they can be unfrozen so that repro re-downloads their outputs?

content/docs/command-reference/freeze.md
```diff
@@ -1,32 +1,32 @@
...
+changed_ until `dvc unfreeze` is used on it.
 
 ## Synopsis
 
 ```usage
 usage: dvc freeze [-h] [-q | -v] targets [targets ...]
 
 positional arguments:
-  targets        stages to freeze.
+  targets        Stages or .dvc files to freeze
```
jorgeorpinel, line 12: @pared please notice these updates which should probably be applied to the help output (dvc freeze -h) in the core repo.

jorgeorpinel, line 12: @pared please notice these updates which should probably be applied to the help output (dvc freeze -h) in the core repo.

jorgeorpinel: Similar changes to unfreeze.

jorgeorpinel: > p.s. there's an issue both in core and here to sync all these at some point.

content/docs/command-reference/freeze.md
shcheklein: It's tricky and not very consistent (`import` vs `import-url`) - @efiop could you clarify how does it work these days? Also, we should start getting rid of the "import stage" terminology. At least while they stay out of `dvc.yaml`.

content/docs/command-reference/freeze.md
```diff
@@ -1,32 +1,32 @@
 # freeze
 
-Freeze a [stage](/doc/command-reference/run). Use `dvc unfreeze` to unfreeze the
-stage.
+Fix the status of one or more [stages](/doc/command-reference/run) to _not
```
shcheklein, line 3: Can we do something simpler? Frozen stage is never executed on `dvc repro` no matter it has changes dependencies or not.

content/docs/command-reference/freeze.md
```diff
@@ -39,21 +39,22 @@ using `dvc repro` on a pipeline that needs their outputs.
...
 ```
 
+> See `dvc run` for more details.
+
 Then, let's change the file `foo` that the stage `make_copy` depends on:
 
 ```dvc
 $ rm foo
-$ echo foo1 > foo
+$ echo foo > foo
```
pared, line 56: foo is unchanged

content/docs/command-reference/unfreeze.md
```diff
@@ -35,56 +38,38 @@ be unfrozen. Use `dvc update` on them to update the file, directory, or
...
+> See `dvc run` for more details.
 
-Now, let's freeze the `bar` stage:
+Then, let's change the file `foo` that the stage `make_copy` depends on, and
+freeze stage as well, to see what's the project status after that:
 
 ```dvc
-$ dvc freeze bar.dvc
+$ dvc freeze make_copy
 $ dvc status
```
pared, line 56: We mention change of `foo` but we don't do that

pared:
I will take on changing the messages in core

content/docs/command-reference/freeze.md
jorgeorpinel: Maybe @pared knows. I can try this later if not.

content/docs/command-reference/freeze.md
jorgeorpinel: Actually. Are .dvc files still considered stages? Not sure we want to mention them separately but probably yes: "Stages" refers to their names (in dvc.yaml), .dvc file are paths to such files.

content/docs/command-reference/freeze.md
jorgeorpinel: I didn't want to use the word "freeze" to explain `freeze` but OK. I like the sentence you suggested but it doesn't say what the command does. Maybe start with "Freeze a stage. Frozen stages..."

content/docs/command-reference/freeze.md
jorgeorpinel: Ah woops yes haha, changing...

content/docs/command-reference/unfreeze.md
jorgeorpinel: Fixed!

content/docs/command-reference/freeze.md
jorgeorpinel: Updated in cca7692d. PTAL

content/docs/command-reference/freeze.md
jorgeorpinel: I see this is already applied in iterative/dvc/pull/4102. Resolving

content/docs/command-reference/freeze.md
pared: > Is this paragraph correct? Are import stages frozen by default, but they can be unfrozen so that repro re-downloads their outputs?

Seems to be true
Example:
```
#!/bin/bash

rm -rf repo other
mkdir repo other
main=$(pwd)

echo data >> data

pushd other

git init --quiet && dvc init --quiet
echo data >> data
dvc add data

git add -A
git commit -am "ble"
popd
pushd repo

git init --quiet && dvc init --quiet

dvc import-url $main/data imp-url
dvc import $main/other data -o imp 

popd
pushd other
echo modify >> data
dvc commit -f data.dvc
git commit -am "something"

popd
echo modify >> data

pushd repo
dvc status

dvc unfreeze imp-url.dvc
dvc unfreeze imp.dvc
dvc repro imp-url.dvc imp.dvc
```

> It's tricky and not very consistent (import vs import-url) - @efiop could you clarify how does it work these days? Also, we should start getting rid of the "import stage" terminology. At least while they stay out of dvc.yaml.

the main difference between `import-url` and `import` results is how `dvc status` will react to them - in case of `import-url` we will get warning that it is frozen, in case of `import`-ed file the status will be checked and returned appropriately.

jorgeorpinel: Thanks @pared! So this paragraph is correct so the main point here is resolved. However we have these 2 side issues that may be worth exporting somewhere else:

> we should start getting rid of the "import stage" terminology. At least while they stay out of dvc.yaml.

@shcheklein Idk. import .dvc files have both deps and outs so they're full stages.

> import-url we will get warning that it is frozen, in case of import-ed file the status will be checked

@pared do you think both should behave the same though? Maybe both like regular `import`, just check the status, no warning. Or both warn + check status properly. If so maybe we should open a separate issue about this in the core repo.

jorgeorpinel:
@pared do you approve the changes now?

content/docs/command-reference/freeze.md
shcheklein: @pared 

> the main difference between import-url and import results is how dvc status will react to them - in case of import-url we will get warning that it is frozen, in case of import-ed file the status will be checked and returned appropriately.

should it be a bug?

> have both deps and outs so they're full stages. 

`deps` and `outs` do not necessarily mean that something is a stage. It's an implementation detail. Probably a leftover for the backward compatibility. From a user perspective it's just an input data (similar to `dvc add`) to my mind. It's imported, yes, but it does not make any difference I think.

I agree that it is a separate discussion though and a separate ticket.

jorgeorpinel: > It's an implementation detail... From a user perspective it's just an input data (similar to dvc add)... It's imported, yes, but it does not make any difference I think.

Yeah, you're right. We should get rid of the term then 🙂 (noted in #608)

pared:
> @pared do you approve the changes now?
@jorgeorpinel, sorry for delay, I do

PR #1496, opened 2020-06-24
cmd ref: more updates on new dag command

jorgeorpinel:
Continuation of #1383

### DIFF
--- a/content/docs/command-reference/dag.md
+++ b/content/docs/command-reference/dag.md
@@ -1,8 +1,7 @@
 # dag
 
-Show [stages](/doc/command-reference/run) in a pipeline that lead to the
-specified stage. By default it lists
-[DVC-files](/doc/user-guide/dvc-files-and-directories).
+Display connected [stages](/doc/command-reference/run) in the pipelines that
+contains them.
 
 ## Synopsis
 
@@ -10,8 +9,8 @@ specified stage. By default it lists
 usage: dvc dag [-h] [-q | -v] [--dot] [--full] [target]
 
 positional arguments:
-  targets         Stage or output to show pipeline for (optional)
-                  Finds all stages in the workspace by default.
+  target          Stage or output to show pipeline for (optional)
+                  Uses all stages in the workspace by default.
 ```
 
 ## Description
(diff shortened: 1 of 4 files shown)


content/docs/command-reference/dag.md
```diff
@@ -1,49 +1,49 @@
 # dag
 
-Show [stages](/doc/command-reference/run) in a pipeline that lead to the
-specified stage. By default it lists
-[DVC-files](/doc/user-guide/dvc-files-and-directories).
+Display connected [stages](/doc/command-reference/run) in the pipelines that
+contains them.
```
efiop, line 4:
```suggestion
contain them.
```
?

efiop: Also, is `that contain them` really necessary? Stages belong to pipelines by definition.

jorgeorpinel: Yes but there could be more than one pipeline right? This command shows the one that contains the target stage (if a target is provided).

jorgeorpinel: But yeah this could be simplifies as much as just "Displays pipeline(s)."

The help output reads `Visualize DVC project DAG.` though. We should try to have these descriptions match. Isn't "pipeline" still the preferred term?

content/docs/command-reference/dag.md
```diff
@@ -1,49 +1,49 @@
 # dag
 
-Show [stages](/doc/command-reference/run) in a pipeline that lead to the
-specified stage. By default it lists
-[DVC-files](/doc/user-guide/dvc-files-and-directories).
+Display connected [stages](/doc/command-reference/run) in the pipelines that
```
shcheklein, line 3: Visualize pipelines from the `dvc.yaml` file as a graph of connected stages.

content/docs/command-reference/dag.md
```diff
@@ -1,49 +1,49 @@
...
 A data pipeline, in general, is a series of data processing
 [stages](/doc/command-reference/run) (for example console commands that take an
 input and produce an <abbr>output</abbr>). A pipeline may produce intermediate
-data, and has a final result. Machine learning (ML) pipelines typically start a
-with large raw datasets, include intermediate featurization and training stages,
-and produce a final model, as well as accuracy
-[metrics](/doc/command-reference/metrics).
+data, and has a final result.
+
+Machine learning (ML) pipelines typically start a with large raw datasets,
```
shcheklein, line 23: let's use Data processing or ML pipelines ...

content/docs/command-reference/dag.md
efiop: Yes, there could be more than one pipeline.

DAG and pipeline(s) are not really conflicting, not sure there should be a prefered term necessarily.

jorgeorpinel: From multiple experiences with key terminology, it's best to avoid synonyms. They can cause confusion and make docs harder to maintain. Unfortunately I don't see us moving away from "pipeline" and the command is now called `dag` so these will have to coexist but still I would prefer "pipelines" as much as possible (and some links to "DAG" concept [Wikipedia] and `dvc dag` of course).

content/docs/command-reference/dag.md
jorgeorpinel: Done

content/docs/command-reference/dag.md
jorgeorpinel: and done

content/docs/command-reference/dag.md
```diff
@@ -1,49 +1,50 @@
 # dag
 
-Show [stages](/doc/command-reference/run) in a pipeline that lead to the
-specified stage. By default it lists
-[DVC-files](/doc/user-guide/dvc-files-and-directories).
+Visualize the pipeline(s) in
+[`dvc.yaml`](/doc/user-guide/dvc-files-and-directories#dvclock-file) as a graph
```
shcheklein, line 4: should we use `graph(s)`?

content/docs/command-reference/dag.md
```diff
@@ -1,49 +1,50 @@
...
 A data pipeline, in general, is a series of data processing
 [stages](/doc/command-reference/run) (for example console commands that take an
 input and produce an <abbr>output</abbr>). A pipeline may produce intermediate
-data, and has a final result. Machine learning (ML) pipelines typically start a
-with large raw datasets, include intermediate featurization and training stages,
-and produce a final model, as well as accuracy
-[metrics](/doc/command-reference/metrics).
+data, and has a final result.
+
+Data processing or ML pipelines typically start a with large raw datasets,
```
shcheklein, line 24: all this text is pretty bad :( let's create a ticket to rewrite it since it's def not a focus of this PR.

content/docs/command-reference/dag.md
jorgeorpinel: The output is a single graph right? Depending what we mean by graph. One output "graph" (as in "figure") but it can have several DAGs (thus "pipeline(s)"). I think it's fine...

jorgeorpinel: Maybe change to figure actually, to avoid confusion. Or use the term DAG. See... This is why having synonyms is problematic 😢

content/docs/command-reference/dag.md
jorgeorpinel: Extracted to #1500

content/docs/command-reference/dag.md
jorgeorpinel: Updated to "DAGs" (linked to Wikipedia) in 78519465

shcheklein: `visualize as graph` definitely means create a picture with graph (math term) - thus it should graph(s) since each pipeline is an independent graph.

shcheklein: if we use DAGs here - let's use graph(s) (DAGs) ... DAG is too specific to be used as a main term.

jorgeorpinel: OK done!

PR #1519, opened 2020-07-01
docs: skipping hidden sidebar items

### LINKED ISSUE #1240: [accessibility] nav: skip hidden sidebar items for keyboard navigation
> Extracted from https://github.com/iterative/dvc.org/pull/1170#issuecomment-620911937.

Once you get to the left sidebar in https://dvc.org/doc/home, it seems to traverse all the collapsed (hidden) items..

mvshmakov:
Fix #1240 

Currently, there is no way to implement it efficiently without a wrapper (https://github.com/nkbt/react-collapse/issues/255). There is an ancient pending PR (https://github.com/nkbt/react-collapse/pull/262) that tries to fix it, but it is about the `aria-hidden` attribute, so the tab navigation would be still broken even (if?) after the merge. I suggest living with this wrapper approach.

### DIFF
--- a/src/components/Community/Section/index.tsx
+++ b/src/components/Community/Section/index.tsx
@@ -89,7 +89,9 @@ const Section: React.FC<ICommunitySection> = ({
       {background && <img className={styles.picture} src={background} alt="" />}
 
       {isTablet ? (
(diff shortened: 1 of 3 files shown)


shcheklein:
Nice! thanks @mvshmakov !

PR #1537, opened 2020-07-06
cmd: rephrase intro of run

jorgeorpinel:


content/docs/command-reference/run.md
```diff
@@ -1,13 +1,13 @@
...
-and execute the command.
+Helper command to create or update _stages_ in
+[`dvc.yaml`](/doc/user-guide/dvc-files-and-directories). Requires a name and a
+command.
 
 ## Synopsis
 
 ```usage
-usage: dvc run [-h] [-q | -v] [-d <path>] [-n <name>] [-o <path>]
+usage: dvc run [-h] [-q | -v] -n <name> [-d <path>] [-o <path>]
```
jorgeorpinel, line 10: If -n is required, why is is printed in `[]` here?

content/docs/command-reference/run.md
```diff
@@ -17,7 +17,7 @@ usage: dvc run [-h] [-q | -v] [-d <path>] [-n <name>] [-o <path>]
                command
 
 positional arguments:
-  command               Command to execute.
+  command               Command for the stage. Executed by default
```
jorgeorpinel, line 20: Help desc. update!

jorgeorpinel: I can submit a PR on the core repo if you agree, Ruslan. Cc @pared

shcheklein: Looks really strange w/o a dot at the end

jorgeorpinel: My bad, I think this is actually a full sentence (with implicit subject & verb "It's").

jorgeorpinel: Fixing in #1540

pared: @jorgeorpinel you want to do the PR on core, or would you rather like us to do it?

skshetry: @jorgeorpinel, I think this is quite opposite. Rather, `--no-exec` is a special case.

skshetry: And, technically, the command may or may not execute (think of run-cache and already-run stages).

content/docs/command-reference/run.md
efiop: @jorgeorpinel Because it is not strictly required, we have a hiddlen `--single-stage` hack for which it is not required.

content/docs/command-reference/run.md
jorgeorpinel: > technically, the command may or may not execute (think of run-cache

If it's in the tun cache it means the results are re-applied at `dvc run` again. From the user's perspective this is the same as executing it again (instantly)

> you want to do the PR on core

Yes, will do... ⏳

jorgeorpinel: > Yes, will do... ⏳

Actually, since there's some controversy here, I'll let you guys decide @pared and @skshetry and if you do not change it just lmk in order to roll this one back.

content/docs/command-reference/run.md
jorgeorpinel: Ok good to know.

P/s should the `-h` output

```
  -n NAME, --name NAME  Stage name.
```

Say it's (usually) required?

content/docs/command-reference/run.md
skshetry: @jorgeorpinel, I don't see any difference between previous and this version. It only adds confusion (what does `by default` mean?).  It'd be better to explain the cases where it won't be executed.

pared: I think @skshetry is right, I don't think we need to explain when it's not executed, thats is what  `--no-exec` reference is for. Maybe we should replace `Only create stage file without actually running it.` -> `Create stage file without actually executing its command.` Since we are operating on `command` and `execute` in main ref

jorgeorpinel: @skshetry you don't see a difference between "Command to execute" and "Command for the stage" ? The former reflected run being a command to run commands (what it used to be), the latter reflects it being a helper to define stages (what it is now, see the cmd ref Description).

I thought we were just discussing the "Executed by default" which I don't have a strong opinion on, tbh. Removed in #1550 (let's continue this part of the discussion there?)

> Maybe we should replace Only create stage file without actually running it. -> Create stage file without actually executing its command.

@pared I agree. Or use the text from the cmd ref: `create a stage file, but do not execute the `command` defined in
  it, nor cache dependencies or outputs`

PR #1543, opened 2020-07-07
Add an alert box above the header to advertise CML

rogermparent:
I'm sure we'll want changes to the text and color, but here's the implementation!

The current implementation is a simple div placed above the header, which collapses to zero height whenever the header collapses. This behavior can also be changed, especially if it has problems on corner cases.

### DIFF
--- a/src/components/LayoutHeader/index.tsx
+++ b/src/components/LayoutHeader/index.tsx
@@ -29,8 +29,10 @@ const LayoutHeader: React.FC<Required<ILayoutModifiable>> = ({ modifiers }) => {
       />
       <div className={styles.header}>
         <div className={cn(styles.alert, collapsed && styles.collapsed)}>
-          <span role="img" aria-label="rocket">🚀</span>
-          Check out the newest tool, {' '}
+          <span role="img" aria-label="rocket">
+            🚀
+          </span>
+          Check out the newest tool,{' '}
           <Link href="https://cml.dev">CML</Link>!
         </div>
         <LayoutWidthContainer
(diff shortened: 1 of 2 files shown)


src/components/LayoutHeader/index.tsx
```diff
@@ -28,6 +28,10 @@ const LayoutHeader: React.FC<Required<ILayoutModifiable>> = ({ modifiers }) => {
         )}
       />
       <div className={styles.header}>
+        <div className={cn(styles.alert, collapsed && styles.collapsed)}>
+          Check out the newest tool from Iterative,{' '}
```
shcheklein, line 32: let's add a bit more fun here:

🚀 Check out the newest tool 

- make font bigger
- do not mention Iterative

src/components/LayoutHeader/index.tsx
```diff
@@ -28,6 +28,13 @@ const LayoutHeader: React.FC<Required<ILayoutModifiable>> = ({ modifiers }) => {
         )}
       />
       <div className={styles.header}>
+        <div className={cn(styles.alert, collapsed && styles.collapsed)}>
+          <span role="img" aria-label="rocket">
+            🚀
+          </span>
+          Check out the newest tool,
```
jorgeorpinel, line 35: Needs a space e.g.
```suggestion
          Check out the newest tool,&nbsp;
```

src/components/LayoutHeader/index.tsx
```diff
@@ -28,6 +28,13 @@ const LayoutHeader: React.FC<Required<ILayoutModifiable>> = ({ modifiers }) => {
...
+          <Link href="https://cml.dev">CML</Link>!
```
jorgeorpinel, line 36: Needs a space e.g.
```suggestion
          Check out the newest tool,&nbsp;
          <Link href="https://cml.dev">CML</Link>!
```

PR #1544, opened 2020-07-07
cmd ref: doc plots diff --targets with revisions usage

### LINKED ISSUE #1425: cmd ref: doc plots diff --targets with revisions usage
Mixing `--targets` option and providing multiple `revisions` for `dvc plots diff` might end messy.
Eg:
`dvc plots diff --targets t1.json t2.csv HEAD v1 v2` - parser does not know how to handle that.

We should include some note for `dvc plots diff` to hint possible use cases:
`dvc plots diff HEAD v1 v2 --targets t1.json t2.csv`
`dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2`

Related https://github.com/iterative/dvc/issues/3962

imhardikj:
Fixes #1425

### DIFF
--- a/content/docs/command-reference/plots/diff.md
+++ b/content/docs/command-reference/plots/diff.md
@@ -105,11 +105,20 @@ file:///Users/dmitry/src/plots/logs.html
 Compare two specific versions (commit hashes, tags, or branches):
 
 ```dvc
-$ dvc plots diff --targets logs.csv HEAD 0135527
+$ dvc plots diff HEAD 0135527 --targets logs.csv
 file:///Users/usr/src/plots/logs.csv.html
 ```
 
 ![](/img/plots_diff.svg)
+
+> Alternatively, you can also run above statement as:
+>
+> ```dvc
+> $ dvc plots diff --targets logs.csv -- HEAD 0135527
+> ```
+>
+> When you're specifying multiple revisions after `--targets`, use `--` so that
+> argument parser doesn't confuse them as options for `--targets` argument.
 
 ## Example: Confusion matrix
 


content/docs/command-reference/plots/diff.md
```diff
@@ -105,12 +105,21 @@ file:///Users/dmitry/src/plots/logs.html
...
 
+> Alternatively, you can also run above statement as:
+>
+> ```dvc
+> $ dvc plots diff --targets logs.csv -- HEAD 0135527
+> ```
+>
+> When you're specifying multiple revisions after `--targets`, use `--` so that
+> argument parser doesn't confuse them as options for `--targets` argument.
+
```
jorgeorpinel, line 122: Nice, I didn't even realize we can do that. But this doesn't close #1425

jorgeorpinel:
Would you like to try again to fully close #1425, @imhardikj ?

> And please, be careful about using the keyword "Fixes" in PR descriptions when they only address an issue partially.

content/docs/command-reference/plots/diff.md
shcheklein: hmm ... what else would expect here to be mentioned, @jorgeorpinel ?

pared: @shcheklein I think we should also include info that:
`dvc plots diff --targets t1.json t2.csv HEAD v1 v2` -> this is bad and won't work. This should be explicitly stated, as in most cases it feels natural to first provide options and only then the positional arguments.

imhardikj:
@jorgeorpinel I'll make PR again to include @pared's suggestion.

content/docs/command-reference/plots/diff.md
jorgeorpinel: > We should include some note for dvc plots diff to hint possible use cases:
dvc plots diff HEAD v1 v2 --targets t1.json t2.csv
dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2

And let's put it in the description please @imhardikj

shcheklein: I think it's better to include this in the `--target` option description. I would try not to explode description too much with examples. At least that's not what we usually do.

jorgeorpinel: Good point. It's better to put it in --targets indeed. It's the first option listed, right after the Description anyway.

I have nothing against super short examples in the cmd ref descriptions though, we use them a little in run I think. It will also be "strange" to have it in the option bullets. But again, I agree this one should go in --target.

shcheklein: I would inline and simplify them as much as possible. And would try to have a single example.

PR #1549, opened 2020-07-08
Adding a note on dvc plots diff --targets with revisions usage

### LINKED ISSUE #1425: cmd ref: doc plots diff --targets with revisions usage

imhardikj:
Fixes #1425 
Continuation of #1544 
@jorgeorpinel @pared I have included possible uses cases in the note

### DIFF
--- a/content/docs/command-reference/plots/diff.md
+++ b/content/docs/command-reference/plots/diff.md
@@ -111,7 +111,8 @@ file:///Users/usr/src/plots/logs.csv.html
 
 ![](/img/plots_diff.svg)
 
-> When you're specifying multiple `--targets` options with multiple `revisions`:
+> Please note that when you're specifying `--targets` options with multiple
+> `revisions`:
 >
 > ```dvc
 > $ dvc plots diff --targets t1.json t2.csv HEAD v1 v2
--- a/content/docs/command-reference/plots/diff.md
+++ b/content/docs/command-reference/plots/diff.md
@@ -47,7 +47,21 @@ please see `dvc plots`.
 
 - `--targets <path>` - specific metric files to visualize. These must be listed
   in a [`dvc.yaml`](/doc/user-guide/dvc-files-and-directories#dvcyaml-file) file
-  (see the `--plots` option of `dvc run`).
+  (see the `--plots` option of `dvc run`). When you are specifying multiple
+  arguments for `--targets` and `revisions`, use `--` after all the arguments
+  for `--targets`:
+
+  ```dvc
+  $ dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2
+  ```
+
+  Without `--`. this statement will show an error as argument parser confuses
+  `revisions` as arguments for `--targets` option. Alternatively, you can also
+  run above statement as:
+
+  ```dvc
+  $ dvc plots diff HEAD v1 v2 --targets t1.json t2.csv
+  ```
 
 - `-o <path>, --out <path>` - name of the generated file. By default, the output
   file name is equal to the input filename with a `.html` file extension (or
@@ -105,32 +119,11 @@ file:///Users/dmitry/src/plots/logs.html
 Compare two specific versions (commit hashes, tags, or branches):
 
 ```dvc
-$ dvc plots diff --targets logs.csv -- HEAD 0135527
+$ dvc plots diff HEAD 0135527 --targets logs.csv
 file:///Users/usr/src/plots/logs.csv.html
 ```
 
 ![](/img/plots_diff.svg)
-
-> Please note that when you're specifying `--targets` options with multiple
-> `revisions`:
->
-> ```dvc
-> $ dvc plots diff --targets t1.json t2.csv HEAD v1 v2
-> ```
->
-> This statement will show an error because argument parser confuses `revisions`
-> as arguments for `--targets` option. To avoid this error use `--` after all
-> the arguments for `--targets`:
->
-> ```dvc
-> $ dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2
-> ```
->
-> Alternatively, you can also run above statement as:
->
-> ```dvc
-> $ dvc plots diff HEAD v1 v2 --targets t1.json t2.csv
-> ```
 
 ## Example: Confusion matrix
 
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/plots/diff.md
```diff
@@ -105,20 +105,32 @@ file:///Users/dmitry/src/plots/logs.html
...
+> Please note that when you're specifying `--targets` options with multiple
+> `revisions`:
+>
+> ```dvc
+> $ dvc plots diff --targets t1.json t2.csv HEAD v1 v2
+> ```
+>
+> This statement will show an error because argument parser confuses `revisions`
+> as arguments for `--targets` option. To avoid this error use `--` after all
+> the arguments for `--targets`:
```
jorgeorpinel, line 123: Let's move this note up to the description as a regular paragraph, and just make a quick reminder in this note (block quote) here in the example.

content/docs/command-reference/plots/diff.md
```diff
@@ -105,20 +105,32 @@ file:///Users/dmitry/src/plots/logs.html
...
+> $ dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2
 > ```
 >
-> When you're specifying multiple revisions after `--targets`, use `--` so that
-> argument parser doesn't confuse them as options for `--targets` argument.
+> Alternatively, you can also run above statement as:
+>
+> ```dvc
+> $ dvc plots diff HEAD v1 v2 --targets t1.json t2.csv
+> ```
```
jorgeorpinel, line 133: This block quote is way too long now :P

content/docs/command-reference/plots/diff.md
jorgeorpinel: Actually, per https://github.com/iterative/dvc.org/pull/1544#discussion_r452406892 please move up into the `--targets` option bullet text instead. Feel free to use several paragraphs and even code blocks or block quote notes in there, but do try to summarize all this a little in both places. Thanks

shcheklein: please, let's try to keep it really simple. The problem itself is very simple and clear and one sentence should be enough in the description or option description to my mind.

jorgeorpinel: Yeah, by "summarize all this a little" I meant summarize it a lot too 🙂

content/docs/command-reference/plots/diff.md
imhardikj, line 133: @jorgeorpinel is this part needed in `--target` option. I think rest of the part can be moved up as it and here instead we can use example in line 108 itself -> `$ dvc plots diff --targets logs.csv -- HEAD 0135527`. Along with one line note reminding to use `--`. 
Thoughts?

jorgeorpinel: I'm not sure, was this part of the example before? What about just use this one for the example (no --) and remove the note? Just try to summarize everything as much as possible, make your change, and we'll review that again. Thanks

content/docs/command-reference/plots/diff.md
```diff
@@ -47,7 +47,21 @@ please see `dvc plots`.
 
 - `--targets <path>` - specific metric files to visualize. These must be listed
   in a [`dvc.yaml`](/doc/user-guide/dvc-files-and-directories#dvcyaml-file) file
-  (see the `--plots` option of `dvc run`).
+  (see the `--plots` option of `dvc run`). When you are specifying multiple
+  arguments for `--targets` and `revisions`, use `--` after all the arguments
+  for `--targets`:
```
jorgeorpinel, line 52:
```suggestion
  (see the `--plots` option of `dvc run`). When specifying multiple targets and
  `revisions`, you may use `--` after this option's arguments, e.g.:
```

content/docs/command-reference/plots/diff.md
```diff
@@ -47,7 +47,21 @@ please see `dvc plots`.
...
+
+  ```dvc
+  $ dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2
+  ```
+
+  Without `--`. this statement will show an error as argument parser confuses
+  `revisions` as arguments for `--targets` option. Alternatively, you can also
+  run above statement as:
```
jorgeorpinel, line 60: Just
```suggestion
Alternatively, you can also run above statement as:
```

jorgeorpinel: I wonder is this should be the first recommendation though, not `--`. I also wonder which other commands can benefit from the `--` syntax.

content/docs/command-reference/plots/diff.md
jorgeorpinel: Continued in https://github.com/iterative/dvc.org/pull/1549#pullrequestreview-446819554

content/docs/command-reference/plots/diff.md
jorgeorpinel: WDYT @imhardikj ?

imhardikj: I think order is correct. As mentioned in https://github.com/iterative/dvc.org/pull/1544#discussion_r451395365 it's almost natural instinct to give `targets` first and then `revisions`.

imhardikj: `--` will be used only when a option accepts multiple arguments and at least one positional argument. This can be used in `metrics diff`. But right now `metrics diff` doesn't have any example which shows usage of `--targets` option.

content/docs/command-reference/plots/diff.md
imhardikj: I'll incorporate these changes in a new PR

content/docs/command-reference/plots/diff.md
jorgeorpinel: But the same note applies to metrics diff? If so please copy it to that reference as well in your new PR, or a subsequent one 🙂 Adding an targets example (similar to the one here) would be a nice addition too but not necessary.

content/docs/command-reference/plots/diff.md
jorgeorpinel: I see it's done in #1567, thanks.

PR #1553, opened 2020-07-10
docs: add intruction to create virtualenv

dmesquita:
Hello!
[edited]
This pull request changes the virtual virtual environment link from [python tutorials](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments) to [python docs](https://python.readthedocs.io/en/stable/library/venv.html). 
The tutorials page has some text before getting to the command itself, the docs page is more straightforward so the user can paste the code get back to DVC guides :)

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -38,6 +38,8 @@ Now let's install the requirements:
 > first.
 
 ```dvc
+$ python3 -m venv .env
+$ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -34,12 +34,10 @@ requirements.txt  train.py
 Now let's install the requirements:
 
 > We **strongly** recommend creating a
-> [virtual environment](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments)
+> [virtual environment](https://python.readthedocs.io/en/stable/library/venv.html)
 > first.
 
 ```dvc
-$ python3 -m venv .env
-$ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
 


content/docs/start/data-pipelines.md
```diff
@@ -38,6 +38,8 @@ Now let's install the requirements:
 > first.
 
 ```dvc
+$ python3 -m venv .env
```
shcheklein, line 41: this is great, didn't know that it's part of the standard library ... have been using `virtualenv -p python3 .env` so far

dmesquita: Haha yeah it's awesome! `venv` is recommended since Python 3.6: [28.3. venv — Creation of virtual environments](https://python.readthedocs.io/en/stable/library/venv.html)

content/docs/start/data-pipelines.md
```diff
@@ -38,6 +38,8 @@ Now let's install the requirements:
...
+$ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
```
jorgeorpinel, line 44: If this stays, it should be part of the md block quote above, please.

But should it stay? I'm not sure. We purposefully don't include this in our docs any more (we try to avoid having to maintain docs/examples of non-DVC tools). We assume people will know some basics. THAT SAID, maybe in this case it's a good idea to keep it, as this is the most introductory of our docs.

jorgeorpinel: Also,

> source .env/bin/activate

On Windows it's

`source .env/Scripts/activate`

I know 😠 but that's another possible reason to avoid this.

dmesquita: Hmm, got it. One cool thing spaCy has is [this quick section to generate the command](https://spacy.io/usage#quickstart). But it makes more sense there because it's the first page of the Get Started guide there. Here the page in question is the "third one" of the guide, I don't know if the pages should follow a sequence or not.

About assuming people will know some basics, maybe it would be good to add [here](https://dvc.org/doc/user-guide/contributing/docs) a little more about what you guys assume about the audience of the docs?

shcheklein:
thanks @dmesquita !

content/docs/start/data-pipelines.md
shcheklein: @dmesquita I've seen a few projects do a widget likes this for the installation page ... we can do something similar for the dvc.org/doc/install - that would be great. Please, feel free to create a ticket.

content/docs/start/data-pipelines.md
```diff
@@ -34,7 +34,7 @@ requirements.txt  train.py
 Now let's install the requirements:
 
 > We **strongly** recommend creating a
-> [virtual environment](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments)
+> [virtual environment](https://python.readthedocs.io/en/stable/library/venv.html)
```
jorgeorpinel, line 37: Thanks! We should probably change this link in other places as well. I think it's in a few places.

jorgeorpinel: Moved to #1556

shcheklein:
One downside of using built in venv is that `pip` is outdated, `wheel` is not includes which makes it very inefficient to use to install DVC to my mind. Additional command is needed:

`env/bin/python -m pip install --upgrade pip setuptools wheel`

more on this here - https://chriswarrick.com/blog/2018/09/04/python-virtual-environments/ .

jorgeorpinel:
Are you saying the change to use the `python -m venv .env` command was not thought through? Should we roll back to `virtualenv -p python3 .env` (and perhaps the link as well)? See https://github.com/iterative/dvc.org/pull/1556/files

shcheklein:
@jorgeorpinel @dmesquita let you decide on this :) (ping the team as well - what do they think).

PR #1567, opened 2020-07-12
Modifying note on dvc plots diff --targets with revisions usage

imhardikj:
Changes as described in https://github.com/iterative/dvc.org/pull/1549#discussion_r453220288 and https://github.com/iterative/dvc.org/pull/1549#discussion_r453220364


imhardikj:
@jorgeorpinel this one can be merged.
I'll open a new PR for `metrics diff` after #1568 is finished.

PR #1504, opened 2020-06-27
term: remove Dvcfile from repro cmd ref.

sarthakforwet:
The following PR completes one subtask of  #1445 by removing/updating `Dvcfile` from `repro.md`.
This issue has also been referenced under #1255 .
> Also remove `Dvcfile` 

Updations:
- [x] Changed the output of `dvc repro` to the new one.
- [x] Updated the usage of `dvc run` command in Examples.

Issue #1445, opened 2020-06-17
term: stop using Dvcfile


jorgeorpinel:
since there's no longer a default stage name in `dvc run` 1.0

This update was started in https://github.com/iterative/dvc.org/pull/1420/commits/435f4103f031e03d7ca194890c4ec59c59430fd3 — needs to be continued in its own PR.

Docs where the term is mentioned:
- [x] push cmd ref. (#1591)
- [x] repro cmd ref. (#1504)
- ~~install/plugins (#1570)~~
- [x] versioning use case tutorial (#1526)
- Double check any other mentions (the one in install/plugins can stay for legacy compatibility)

sarthakforwet:
@jorgeorpinel  in `push.md` file under the example `With dependencies`, following two problems come up - 
 
![dvc](https://user-images.githubusercontent.com/46163196/87216344-7ecd1b00-c35c-11ea-8fc0-db880027978f.PNG)
Here the `stages` has link to `run.md` which no clearly describes the pipeline stages. Was it linked to some pipeline stages before the new release? 

Secondly, the command `dvc pipeline show` is to be replaced with `dvc dag` and its output needs to be updated with the new syntax to output stages specified in `dvc.yaml` file. However I am not getting what new stages I should replace it with? Do these stages exist somewhere in the docs or should I go with a new example?

jorgeorpinel:
Yes, the word "stage" is linked to the `dvc run` cmd ref in many places. Currently it's the best place where the concept of stages is clearly described, did you read it?

![image](https://user-images.githubusercontent.com/1477535/87231230-edb96c80-c37a-11ea-9efd-3516c11e35bd.png)

How does it relate to this issue, @sarthakforwet ? See #550. and replacing `pipeline` with `dag` is in #1255. Please comment in those issues accordingly, I'm hiding these comments from here. Thanks

jorgeorpinel:
Or feel free to open a PR with any progress on the pipeline->dag issue and comment there so we continue the conversation and address your doubts. Thanks

shcheklein:
Reopening since there are still Dvcfile in the docs?

jorgeorpinel:
Where? Only one in content\docs\install\plugins.md which Ruslan recommended we keep for backward compatibility.

shcheklein:
Yep, probably got removed in another PR or commits!

jorgeorpinel:
Yeah sorry that PR was confusing. There was an overlapping one and I merged them in the right order 🙂

PR #1571, opened 2020-07-13
Updated repro.md for the DVC 1.0 release

sarthakforwet:
A PR as an outcome of the dicsussion at [#1445(review)](https://github.com/iterative/dvc.org/pull/1504#discussion_r448617904). 

This PR intends to completely update `dvc repro` documentation to sync with the 1.0 version of DVC.

### DIFF
--- a/content/docs/command-reference/repro.md
+++ b/content/docs/command-reference/repro.md
@@ -18,12 +18,15 @@ positional arguments:
 
 ## Description
 
-`dvc repro` provides an way to regenerate data pipeline results, by restoring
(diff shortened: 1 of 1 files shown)


sarthakforwet:
Should the first sentence in -c option be updated as  - "directory within the project containing `dvc.yaml` file to reproduce from.

> Instead of using --cwd, one can alternately specify a target in a subdirectory as path/to/target.dvc

Should this line be removed as not any use of this now ( also we cannot specify a YAML file directly in this command ).

PR #1384, opened 2020-05-31
cmd: document target granularity for push/pull/etc, et al.

efiop:
Fixes #886

UPDATE: See also https://github.com/iterative/dvc.org/pull/1384#issuecomment-657744994 (could be extracted into another PR)

UPDATE: Merge #1602 after this.

PR #1572, opened 2020-07-13
repro: 1.x update

sarthakforwet:
A PR as an outcome of the dicsussion at #1445(review).

This PR intends to completely update dvc repro documentation to sync with the 1.0 version of DVC.


sarthakforwet:
Should the first sentence in -c option be updated as - 
"directory within the project containing `dvc.yaml` file to reproduce from".

And is there any need to update -R option?

content/docs/command-reference/repro.md
```diff
@@ -37,7 +40,7 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 needed after a `git commit`. See `dvc install` for more details.
 
 There's a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage file `targets`, or by using the `--single-item`,
+command: by specifying stage `targets`, or by using the `--single-item`,
 `--cwd`, or other options.
```
jorgeorpinel, line 44:
```suggestion
`--cwd`, among other options.
```

jorgeorpinel:
> Should the first sentence in -c option be updated as

No, `dvc repro` also works with .dvc files. Maybe this should be clarified, although it's clearly mentioned in the `targets` argument desc.

> And is there any need to update -R option?

Why? What is the discrepancy? What do you suggest? Thanks

content/docs/command-reference/repro.md
```diff
@@ -18,12 +18,15 @@ positional arguments:
...
 
-`dvc repro` provides an way to regenerate data pipeline results, by restoring
-the dependency graph (a
-[DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)) implicitly defined
-by the [stage files](/doc/command-reference/run) (DVC-files with dependencies)
-that are found in the <abbr>project</abbr>. The commands defined in these stages
-can then be executed in the correct order, reproducing pipeline results.
+`dvc repro` provides a way to regenerate data pipeline results, by restoring the
+dependency graph ( a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)
+) implicitly defined by the [stages](/doc/command-reference/run) (DVC-files with
```
jorgeorpinel, line 23: Thanks for the typo fix but now you introduced spaces inside the parentheses. Please revert that (strange style).

content/docs/command-reference/repro.md
```diff
@@ -18,12 +18,15 @@ positional arguments:
...
+dependencies) that are found in the <abbr>project</abbr>. The commands defined
```
jorgeorpinel, line 24: Stages are no longer "DVC-files with dependencies" found in the project. Need to rewrite to mention and link to dvc.yaml here.

content/docs/command-reference/repro.md
```diff
@@ -18,12 +18,15 @@ positional arguments:
...
+in these stages can then be executed in the correct order, reproducing pipeline
+results.
+
+`dvc repro` relies on the DAG definition that it reads from `dvc.yaml` file and
+uses `dvc.lock` file to determine what exactly needs to be run.
```
jorgeorpinel, line 29: Just put it in the previous paragraph like I mentioned, no need for an extra p. And no need to mention dvc.lock yet, if at all — maybe in the examples.

content/docs/command-reference/repro.md
```diff
@@ -18,12 +18,15 @@ positional arguments:
...
 
 > Pipeline stages are typically defined using the `dvc run` command, while
 > initial data dependencies can be registered by the `dvc add` command.
```
jorgeorpinel, line 32: This note should  be updated, please refer to the `dvc run` cmd ref (it's now considered a helper cmd). The recommended way to define stages is manually write/edit dvc.yaml.

content/docs/command-reference/repro.md
```diff
@@ -37,8 +40,8 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 needed after a `git commit`. See `dvc install` for more details.
 
 There's a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage file `targets`, or by using the `--single-item`,
-`--cwd`, or other options.
+command: by specifying stage `targets`, or by using the `--single-item`,
+`--cwd`, among other options.
```
jorgeorpinel, line 44: Maybe clarify .dvc files area also accepted as targets around here.

sarthakforwet:
> -R, --recursive - determines the stages to reproduce by searching each target directory and its subdirectories for DVC-files to inspect

This option also searches for `dvc.yaml` files and maybe we should mention it?

content/docs/command-reference/repro.md
```diff
@@ -53,7 +56,7 @@ corresponding commands. <abbr>Outputs</abbr> are deleted from the
 
 It saves all the data files, intermediate or final results into the <abbr>DVC
 cache</abbr> (unless the `--no-commit` option is used), and updates the hash
-values of changed dependencies and outputs in the corresponding stage files.
+values of changed dependencies and outputs in the corresponding stages.
```
jorgeorpinel, line 59:
```suggestion
values of changed dependencies and outputs in the corresponding stages (or
`.dvc` files).
```

content/docs/command-reference/repro.md
```diff
@@ -124,12 +127,12 @@ only execute the final stage.
 - `-i`, `--interactive` - ask for confirmation before reproducing each stage.
   The stage is only executed if the user types "y".
 
-- `-p`, `--pipeline` - reproduce the entire pipelines that the stage file
-  `targets` belong to. Use `dvc dag <target>` to show the parent pipeline of a
-  target stage.
+- `-p`, `--pipeline` - reproduce the entire pipelines that the stage `targets`
+  belong to. Use `dvc dag <target>` to show the parent pipeline of a target
+  stage.
```
jorgeorpinel, line 132: This may supports .dvc files as well, which can be the beginning of a pipeline. Please test it to confirm, and if so update to just:
```suggestion
- `-p`, `--pipeline` - reproduce the entire pipelines that the `targets`
  belong to. Use `dvc dag <target>` to show the parent pipeline of a target.
```
> p formatting needed too.

jorgeorpinel:
Actually yes, something is definitely wrong with the `-R` option desc @sarthakforwet because it talks about "target directories" but `targets` doesn't support dirs. Probably refers to the `--cwd` option. Can you please test to check and update both options accordingly to make reference of one another?

> See also https://github.com/iterative/dvc/issues/4191#issuecomment-657702111

jorgeorpinel:
> > reproduce by searching each target directory and its subdirectories for DVC-files to inspect

> This option also searches for dvc.yaml

Yep @sarthakforwet , like I mentioned in https://github.com/iterative/dvc.org/pull/1572#issuecomment-657704976 that option desc. is outdated. "DVC-files" don't even exist any more, we call them `.dvc` files now. Let's just end the sentence at "...searching each target directory and its subdirectories."

I also said "targets doesn't support dirs." but actually I think that's wrong (per https://github.com/iterative/dvc/issues/4191#issuecomment-657744483). But also notice:

> Probably refers to the --cwd option. Can you please test to check and update both options accordingly to make reference of one another?"

Thanks

content/docs/command-reference/repro.md
```diff
@@ -18,12 +18,13 @@ positional arguments:
...
-`dvc repro` provides an way to regenerate data pipeline results, by restoring
-the dependency graph (a
-[DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)) implicitly defined
-by the [stage files](/doc/command-reference/run) (DVC-files with dependencies)
-that are found in the <abbr>project</abbr>. The commands defined in these stages
-can then be executed in the correct order, reproducing pipeline results.
+`dvc repro` provides a way to regenerate data pipeline results, by restoring the
+dependency graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))
+implicitly defined by the [stages](/doc/command-reference/run) listed in
+[`dvc.yaml` file](/doc/user-guide/dvc-files-and-directories#dvcyaml-file). The
```
jorgeorpinel, line 24: **Please**, basic grammar. Very important.
```suggestion
implicitly defined by the [stages](/doc/command-reference/run) listed in the
[`dvc.yaml` file](/doc/user-guide/dvc-files-and-directories#dvcyaml-file). The
```

content/docs/command-reference/repro.md
```diff
@@ -36,9 +37,10 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 💡 For convenience, a Git hook is available to remind you to `dvc repro` when
 needed after a `git commit`. See `dvc install` for more details.
 
-There's a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage file `targets`, or by using the `--single-item`,
-`--cwd`, or other options.
+There are a few ways to restrict the stages that will be regenerated by this
+command: by specifying stage `targets`( can be a `.dvc` file or one of the
+stages in `dvc.yaml` file ), or by using the `--single-item`, `--cwd`, among
```
jorgeorpinel, line 42: .dvc files are not stages, please rephrase. Feel free to break up the sentences, it's all about the writing. BTW, strange formatting again (spaces around paragraphs, and missing after comma) — please keep in mind this is a writing program. We'd prefer not having to worry much about copy editing GSoD-related contributions. Thanks

content/docs/command-reference/repro.md
jorgeorpinel: @sarthakforwet I see you did update this, did you test to confirm first? May I ask what test you ran? Thanks

sarthakforwet: Well, I ran the following test - First created a file named `text.txt` then tracked it with `dvc add` and finally built my rest of the pipeline by adding stages to dvc.yaml file.

dvc

`dvc.yaml` file -

yaml_file

Directory structure - 

dvc_tree
  

  
After then I reproduced the complete pipeline with the last stage as target whose output is - 

pipeline

content/docs/command-reference/repro.md
sarthakforwet: > BTW, strange formatting again (spaces around paragraphs, and missing after comma) — please keep in mind this is a writing program. We'd prefer not having to worry much about copy editing GSoD-related contributions.

I would keep that in mind from the next commits.

> .dvc files are not stages, please rephrase.

Maybe we should change the sentence to - 

"There are a few ways to restrict the stages that will be regenerated by this command: by specifying `targets`, or by using the --single-item, --cwd, among other options."

As targets are already defined in the "Synopsis", specifying `targets` here would by default mean that .dvc files are also accepted. WDYT @jorgeorpinel ?

sarthakforwet:
> I also said "targets doesn't support dirs." but actually I think that's wrong

Yeah actually, targets do support dirs.

So basically we have the following difference between usage of -R and -c option  - 

R_ cdiff

Directory structure - 

directory

Hence I think the -c option should be updated by removing the following line - 
" Instead of using --cwd, one can alternately specify a target in a subdirectory as path/to/target.dvc." 

and adding an example of using `dvc repro` with specific targets.

content/docs/command-reference/repro.md
jorgeorpinel: OK. How does that check whether this option supports .dvc as targets? (not just stage names). That is the change applied here to the doc.

jorgeorpinel: * .dvc files

jorgeorpinel: \* .dvc files

content/docs/command-reference/repro.md
```diff
@@ -18,15 +18,19 @@ positional arguments:
...
+[`dvc.yaml` file](/doc/user-guide/dvc-files-and-directories#dvcyaml-file). The
+commands defined in these stages can then be executed in the correct order,
+reproducing pipeline results. `dvc repro` relies on the DAG definition that it
+reads from `dvc.yaml` file.
+
+> Pipeline stages are typically defined by manually adding or manipulating them
+> in the
+> [`dvc.yaml` file](/doc/user-guide/dvc-files-and-directories#dvcyaml-file) or
+> by using `dvc run` helper command, while initial data dependencies can be
+> registered by the `dvc add` command.
```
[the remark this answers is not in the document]
jorgeorpinel: P/s was missing "the" again the way you had written it. No longer relevant in the way I suggested though...

content/docs/command-reference/repro.md
sarthakforwet: > How does that check whether this option supports .dvc files as targets?

Even if you write the `.dvc` filename there, this option would work fine. 

image

sarthakforwet: And actually I'm its kind of confusing with the output for `.dvc` file. As you said `.dvc` files are not stages but here we can see from the output that it is considering them as one. So don't you think the output for .dvc files needs to be something different?

content/docs/command-reference/repro.md
jorgeorpinel: > Maybe we should change the sentence to "... by specifying targets, or by using the --single-item, --cwd, among other options." As targets are already defined in the "Synopsis"

It's a good idea but the synopsis isn't enough. We want to explain everything completely in the description again. It's OK if it repeats the text from the Synopsis, for example.

content/docs/command-reference/repro.md
```diff
@@ -105,8 +111,8 @@ only execute the final stage.
   independent unit.
 
 - `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory and its subdirectories for DVC-files to inspect. If there are
-  no directories among the `targets`, this option is ignored.
+  target directory and its subdirectories. If there are no directories among the
+  `targets`, this option is ignored.
```
jorgeorpinel, line 115: Let's further simplify this, actually (and fix grammar error I just noticed):
```suggestion
  target directory (if any) and their subdirectories.
```

content/docs/command-reference/repro.md
jorgeorpinel: Sound good, thanks.

> Its kind of confusing with the output for .dvc file...

I agree. Can you please open an issue on the core DVC repo about this output? Please link to this discussion as context (I'll leave it unresolved for that purpose.)

content/docs/command-reference/repro.md
jorgeorpinel: > Maybe we should change the sentence to "... by specifying targets, or by using the --single-item, --cwd, among other options." As targets are already defined in the "Synopsis"

It's a good idea but the synopsis isn't enough. We want to explain everything completely in the description again. It's OK if it repeats the text from the Synopsis, for example.

content/docs/command-reference/repro.md
jorgeorpinel: P/s also do you think we should also report the weird -R output you posted in https://github.com/iterative/dvc.org/pull/1572#issuecomment-658272251 when you provide `./` as the target? That output is also confusing! It repeats some "stages" but using different relative paths, it seems.

jorgeorpinel: Ah never mind about that last comment. I see you created an `x/` directory. I was confused by that.

sarthakforwet: >  I see you created an x/ directory

Yeah! actually I created the stage names same in both `dvc.yaml` files to check if we can have do something like that.

jorgeorpinel:
> I think the -c option should be updated by...

Hard to give feedback on comments like this, please just apply the changes and I'll review that. You can even open a comment on the file change to point something specific out for me to notice.

But first, did you check if it's different to run `dvc repro -c ./` vs. just `dvc repro ./` ?

> and adding an example of using dvc repro with specific targets.

Not sure it's needed but see the file changes in #1384 for several examples of that, and decide if you think it's appropriate to add here too.

content/docs/command-reference/repro.md
```diff
@@ -95,7 +101,7 @@ only execute the final stage.
 
 - `-s`, `--single-item` - reproduce only a single stage by turning off the
   recursive search for changed dependencies. Multiple stages are executed
-  (non-recursively) if multiple stage files are given as `targets`.
+  (non-recursively) if multiple stages are given as `targets`.
 
 - `-c <path>`, `--cwd <path>` - directory within the project to reproduce from.
   Instead of using `--cwd`, one can alternately specify a target in a
```
jorgeorpinel, line 107: Like I just suggested, I'd like to test whether this option even makes provides any value now that dvc repro accepts directories as targets. But even if it does, the paragraph is outdated, for example "specify a target in a subdirectory as `path/to/target.dvc`" — `target.dvc` is not a subdir.

sarthakforwet:
> But first, did you check if it's different to run dvc repro -c ./ vs. just dvc repro ./

Obviously they are different. And I think `dvc repro ./` would give an error as we cannot specify a directory with `dvc repro` command without using any option.

content/docs/command-reference/repro.md
jorgeorpinel: > Obviously they are different... I think dvc repro ./ would give an error...
From https://github.com/iterative/dvc.org/pull/1572#issuecomment-659879072

I don't see how it's obvious. Did you test each option @sarthakforwet ? You are making lazy assumptions. Please share your process and results. Thanks

sarthakforwet: > I don't see how it's obvious

Oh sorry! Maybe I interpreted the review incorrectly but I think you are asking for this - 

image

Thanks

sarthakforwet:
@jorgeorpinel  Although we cannot perform parallel stage execution but we can perform sequential stage execution (even when having different branches in the pipeline) like this-

image

would it be beneficial to mention this too?

content/docs/command-reference/repro.md
jorgeorpinel: I see. You are correct then. The thing is you can actually use `dvc repro path/to/dir` if `path/to/dir` is tracked by DVC. That's what I'd like you to play with please, compare using or not using `-c` and `-R` in that scenario, and find a good way to express all your findings in the description; as well, probably, as the new **Specific files or directories** example you're suggesting.

jorgeorpinel:
> Although we cannot perform parallel stage execution but we can perform sequential stage execution

- [ ] Good idea @sarthakforwet it can be a small change in https://dvc.org/doc/command-reference/repro#parallel-stage-execution just to clarify that if you just run repro, the final result will be regenerated, but the stages will be executed sequentially. If you want it to be parallel... etc.

content/docs/command-reference/repro.md
```diff
@@ -36,9 +39,9 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 💡 For convenience, a Git hook is available to remind you to `dvc repro` when
 needed after a `git commit`. See `dvc install` for more details.
 
-There's a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage file `targets`, or by using the `--single-item`,
-`--cwd`, or other options.
+There are a few ways to restrict the stages that will be regenerated by this
+command: by specifying stage or `.dvc` file to reproduce, or by using the
```
jorgeorpinel, line 43: Is "specifying stage or `.dvc` file to reproduce" correct grammatically? It was "specifying stage file `targets`" before (which is correctly written).

sarthakforwet:
One more thing @jorgeorpinel . I found (dunno if already been tested) that we can specify more than one stage/`.dvc` file in the `dvc repro` command. 
Just like this - 

image

 So should we also include information regarding this in the doc?

content/docs/command-reference/repro.md
sarthakforwet: > which is correctly written

But stages are not files now so we can't write "stage file `targets`". We have to specify stage and `.dvc` file separately. Maybe we can write "specifying `targets` (stage or .dvc file).."

jorgeorpinel:
> I found (dunno if already been tested) that we can specify more than one stage

This is already explained everywhere. `targets` is plural, it's used in plural everywhere in the ref...

sarthakforwet:
> This is already explained everywhere. targets is plural, it's used in plural everywhere in the ref...

Yeah that's why I deleted that comment. Thanks

content/docs/command-reference/repro.md
jorgeorpinel: I know we need to change it. What I'm saying is the way you wrote it is incorrect grammatically. Please fix it.

sarthakforwet: > Please fix it.

So is "specifying `targets` (stage or .dvc file), ..." correct?

jorgeorpinel: Nope, the fix is much simpler than that. It has to do with singular vs plural... If you use singular then an article is needed (a, the, etc.) yet missing rn. But should it be singular in the first place?

sarthakforwet: Even the idea to introduce  `.dvc` files here doesn't make any sense in the first place. Since a `.dvc` file can only be used as the first node in a branch and hence no stages will be regenerated if we specify `.dvc` file here.

For an example - 

image

content/docs/command-reference/repro.md
jorgeorpinel: Were you able to open that issue @sarthakforwet ?

sarthakforwet: Yeah @jorgeorpinel. I opened that issue [#4233](https://github.com/iterative/dvc/issues/4233) and also mentioned this discussion there.

jorgeorpinel: Great. Resolving this then

content/docs/command-reference/repro.md
jorgeorpinel: If you haven't done this, let's leave it for a separate PR @sarthakforwet

content/docs/command-reference/repro.md
jorgeorpinel: > a .dvc file can only be used as the first node in a branch and hence no stages will be regenerated if we specify .dvc file here.

If the tracked data changes, the .dvc file will become outdated and `repro` will commit the new data to the cache. So it's perfectly valid to use .dvc files here

I think we're over-complicating this discussion. Please just fix the grammar of the current text.

content/docs/command-reference/repro.md
jorgeorpinel: Same for the note on parallelism (https://github.com/iterative/dvc.org/pull/1572#issuecomment-660263924).

content/docs/command-reference/repro.md
```diff
@@ -40,7 +40,7 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 needed after a `git commit`. See `dvc install` for more details.
 
 There are a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage or `.dvc` file to reproduce, or by using the
+command: by specifying a stage or a `.dvc` file to reproduce, or by using the
```
jorgeorpinel: Only one stage or one .dvc file is accepted as `targets`?

sarthakforwet: No, it can be many. The sentence should be "specifying stages or .dvc files..."  but only the grammar fix in the current text was asked so only did that. Should I update it?

sarthakforwet:
> What about the Examples though? Need to test them all and update if needed.

The "Example: Downstream" probably needs updation @jorgeorpinel. Following updates could be needed-

a.) The "Example: Downstream"  describes `text.txt` as the dependency in the last stage as - "The reason being that the text.txt file is a dependency in the last stage of the pipeline " but it is a dependency in the stage "filter". So we should update it accordingly.

image

b. Performing `dvc repro --downstream` in the example would reproduce the complete pipeline as no stages are specified and `text.txt` is also changed. So I think we should specify a stage to depict its true usage.

c. Running `dvc repro --downstream` does not update `.dvc` file as -

image

So should we also mention it in the doc?

I would open a new PR and mention this discussion if the changes are valid.

content/docs/command-reference/repro.md
jorgeorpinel: "specifying stages or .dvc files" is the grammar fix I was looking for.

sarthakforwet:
@jorgeorpinel  Are we ready to merge this PR?

jorgeorpinel:
Thanks again @sarthakforwet 

On the examples, I've started updating them in #1615 since it's mostly typos and corrections, not really 1.x-related updates. Feel free to contribute to that branch (although you would need to open a PR from your fork to my branch... or you can leave suggestions in the changes).

> b. Performing dvc repro --downstream in the example would reproduce the complete pipeline as no stages are specified and text.txt is also changed. So I think we should specify a stage to depict its true usage.

Are you sure? Meaning have you actually tried this? If so, that would probably be 1.x-related, actually.

> c. Running dvc repro --downstream does not update .dvc file as -

Interesting finding... I don't know if that's expected, or some bug! Please open a report in the core repo (link here for context).

sarthakforwet:
> Are you sure? 

Yeah I'm sure about this. It is evident from the screenshot in the next point. If we take another example - 

image
  
We can say that by-default `dvc repro --downstream` starts with the first stage of the pipeline.

jorgeorpinel:
In that case like I said the --downstream option description definitely needs an update, as well as that example itself (see https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-453636252). Would you like to submit a PR to my `jorge` branch with that?

Feel free to also include the 2 pending items listed under https://github.com/iterative/dvc.org/pull/1572#pullrequestreview-450392422

sarthakforwet:
> Would you like to submit a PR to my jorge branch with that?

Yeah absolutely.

> Feel free to also include the 2 pending items listed under #1572 (review) above.

Sure!

content/docs/command-reference/repro.md
```diff
@@ -18,15 +18,18 @@ positional arguments:
...
-can then be executed in the correct order, reproducing pipeline results.
-
-> Pipeline stages are typically defined using the `dvc run` command, while
-> initial data dependencies can be registered by the `dvc add` command.
+`dvc repro` provides a way to regenerate data pipeline results, by restoring the
+dependency graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))
+implicitly defined by the [stages](/doc/command-reference/run) listed in the
+[`dvc.yaml` file](/doc/user-guide/dvc-files-and-directories#dvcyaml-file). The
+commands defined in these stages can then be executed in the correct order,
+reproducing pipeline results. `dvc repro` relies on the DAG definition that it
```
shcheklein, line 26: last sentence is excessive .. repeats the first one

content/docs/command-reference/repro.md
```diff
@@ -36,9 +39,9 @@ and caches relevant <abbr>data artifacts</abbr> along the way.
 💡 For convenience, a Git hook is available to remind you to `dvc repro` when
 needed after a `git commit`. See `dvc install` for more details.
 
-There's a few ways to restrict the stages that will be regenerated by this
-command: by specifying stage file `targets`, or by using the `--single-item`,
-`--cwd`, or other options.
+There are a few ways to restrict the stages that will be regenerated by this
+command: by specifying stages or `.dvc` files to reproduce, or by using the
```
shcheklein, line 43: can it accept an output? 

stages -> stage names

.dvc - probably it's better not to metion this? what is the use case for this? can it be done w/o repro?

content/docs/command-reference/repro.md
```diff
@@ -53,7 +56,8 @@ corresponding commands. <abbr>Outputs</abbr> are deleted from the
 
 It saves all the data files, intermediate or final results into the <abbr>DVC
 cache</abbr> (unless the `--no-commit` option is used), and updates the hash
-values of changed dependencies and outputs in the corresponding stage files.
+values of changed dependencies and outputs in the corresponding stages (or
+`.dvc` files).
```
shcheklein, line 60: do we need to complicate everything with `.dvc` files?

content/docs/command-reference/repro.md
```diff
@@ -95,7 +99,7 @@ only execute the final stage.
 
 - `-s`, `--single-item` - reproduce only a single stage by turning off the
   recursive search for changed dependencies. Multiple stages are executed
-  (non-recursively) if multiple stage files are given as `targets`.
+  (non-recursively) if multiple stages are given as `targets`.
```
shcheklein, line 102: stage names?

content/docs/command-reference/repro.md
```diff
@@ -124,12 +127,11 @@ only execute the final stage.
...
 
-- `-p`, `--pipeline` - reproduce the entire pipelines that the stage file
-  `targets` belong to. Use `dvc dag <target>` to show the parent pipeline of a
-  target stage.
+- `-p`, `--pipeline` - reproduce the entire pipelines that the `targets` belong
+  to. Use `dvc dag <target>` to show the parent pipeline of a target.
 
-- `-P`, `--all-pipelines` - reproduce all pipelines, for all the stage files
-  present in `DVC` repository.
+- `-P`, `--all-pipelines` - reproduce all pipelines, for all the stages present
```
shcheklein, line 133: does it reproduce all pipelines within one dvc.yaml or within all of them? (search recusrsively)?

content/docs/command-reference/repro.md
sarthakforwet: > can it be done w/o repro?

Yes, `.dvc` files can easily be updated via `dvc add`.

content/docs/command-reference/repro.md
sarthakforwet: It reproduces all of the pipelines (recursive search).

content/docs/command-reference/repro.md
jorgeorpinel: Will be addressing these in #1615

content/docs/command-reference/repro.md
jorgeorpinel: > can it accept an output?

No, repro/run don't support this nor granularity (files inside tracked dirs)

> .dvc - what is the use case for this?

Same as using `dvc add` again on a changed data file.

Keeping it in this paragraph because it's not an edge case to have the raw data as .dvc files at the beginning of a pipeline and `dvc repro` will by default update these if the data changed.

Updated in #1615.

content/docs/command-reference/repro.md
jorgeorpinel: Unfortunately this is a pretty common thing as I just explained. And we no longer call .dvc files "stages" so we need to keep mentioning both some times...

jorgeorpinel: In fact I removed the parentheses.

content/docs/command-reference/repro.md
jorgeorpinel: Updated.

content/docs/command-reference/repro.md
jorgeorpinel: I'm assuming all pipelines in all dvc.yaml files but not sure. Let's test this @sarthakforwet ? Keep in mind you can have one dvc.yaml in each subdirectory of the project. Let's update in #1615 if needed. Thanks

sarthakforwet: > Let's test this @sarthakforwet ?

Already tested it @jorgeorpinel. Here's the results - 

image

Directory Structure - 

image

jorgeorpinel: Just tested. -P makes it recursive indeed (checks all dvc.yaml files). I rewrote the description anyway since it didn't make much sense. See #1615.

jorgeorpinel: The funny think is that `status` is recursive by default, but `repro` needs `-P`... Is that a bug? Cc @efiop

content/docs/command-reference/repro.md
shcheklein: Even if it the case (repro updates .dvc) - it's a side effect, not a main goal of the `dvc repro` ... `dvc repro something.dvc` for `dvc add`-ed files is confusing and not worth complicating docs for I think.

content/docs/command-reference/repro.md
shcheklein: let's just use  `... in the DVC files (.dvc, dvc.lock, etc)` ... essence first, details second

content/docs/command-reference/repro.md
shcheklein: thanks @sarthakforwet, for testing this. I would make the note more explicit then

content/docs/command-reference/repro.md
jorgeorpinel: > dvc repro something.dvc for dvc add-ed files is confusing

Good point. Generalized in ebc15605.

content/docs/command-reference/repro.md
jorgeorpinel: OK. Updated to "in the DVC files (`dvc.lock` and `.dvc`)" (those are the only 2 that apply here) also in ebc15605.

jorgeorpinel: > DVC files (dvc.lock and .dvc)

Wouldn't it be better to just have "dvc.lock and .dvc files" thouogh? Shorter

content/docs/command-reference/repro.md
jorgeorpinel: I already updated the note.

content/docs/command-reference/repro.md
shcheklein: yep, can be that as well

jorgeorpinel: Changed. I like it also because so far I've managed to avoid the term "DVC file" It's confusing vs. .dvc file, which is just one type of DVC file...

jorgeorpinel: I would even rename the DVC Files and Dirs guide to Internal Files and Dirs... But anyway, moving on!

content/docs/command-reference/repro.md
jorgeorpinel: > status is recursive by default, but repro needs -P

Extacted this matter iterative/dvc/issues/4273 with much more detail. PTAL

content/docs/command-reference/repro.md
sarthakforwet: >  The thing is you should be able to use dvc repro path/to/dir (without flags) if path/to/dir is tracked by DVC,

We can specify the corresponding `.dvc` file of the DVC tracked directory in `dvc repro`  (which would update any change in the directory) but not the directory name itself. Moreover we cannot build a pipeline in DVC tracked directories ( as per my experimentation dunno if documented already). Also I think directories have no logical use as a dependency in a pipeline. (although we can add them if we wish). And hence I think it is not incorrect if `dvc repro` does not take directories as targets (without flags).

Since we cannot specify a directory in `dvc repro`, to reproduce pipelines inside a directory/sub-directory (which is untracked), we make use of -c and -R options.

However no example in the doc for using `dvc repro` with a specific `target` stage or .dvc file is provided and hence I think this should be done. WDYT @jorgeorpinel?

jorgeorpinel: > We can specify the corresponding .dvc file ... but not the directory name itself

(or a stage name). OK got it. I think you're right. Except when you use `-R`. 
- [ ] This should be clarified in the -R option desc: It expands what the `targets` argument accepts.

> we cannot build a pipeline in DVC tracked directories

I never thought about that... I guess it's to be expected. May need a core repo issue, will think about it 🧠 thanks

> I think directories have no logical use as a dependency in a pipeline

Why not? Of course you can use a directory as a dependency. They can also be outputs. I don't see how this affects our conversation though.

> to reproduce pipelines inside a directory/sub-directory (which is untracked), we make use of -c and -R options

And -P.

> - [ ] no example in the doc for using dvc repro with a specific target stage or .dvc file is provided and hence I think this should be done

Agree. I'll create an issue about these 2 check boxes, ans possibly something else... I'll tag you there. Thanks!

jorgeorpinel: See #1632

sarthakforwet: > (or a stage name)

I am referencing DVC tracked directories specifically in [#1572(review)](https://github.com/iterative/dvc.org/pull/1572#discussion_r460402286) and I think "stage name" should not be mentioned with that..

> And -P.

Yeah sure, but I was talking specifically in comparison of `dvc repro` with/without `-R` and `-c` options.

> Agree. I'll create an issue about these 2 check boxes, ans possibly something else...

Thanks!

jorgeorpinel: > I am referencing DVC tracked directories specifically

Stages in dvc.yaml can output directories too. It't not exclusive to .dvc files.

PR #1573, opened 2020-07-13
Adding a note on dvc metrics diff --targets with revisions usage

imhardikj:
As per suggestion by @jorgeorpinel  in https://github.com/iterative/dvc.org/pull/1549#discussion_r453765990


content/docs/command-reference/metrics/diff.md
```diff
@@ -43,7 +43,19 @@ lists all the current metrics without comparisons.
 ## Options
 
 - `--targets <paths>` - limit command scope to these metric files. Using -R,
-  directories to search metric files in can also be given.
+  directories to search metric files in can also be given. When specifying
+  multiple `--targets` before `revisions`, you may use `--` after this option's
+  arguments, e.g.:
+
+  ```dvc
+  $ dvc metrics diff --targets t1.json t2.csv -- HEAD v1
```
jorgeorpinel, line 51: metrics diff doesn't support CSV. Please think through anything copy pasted among commands. See https://dvc.org/doc/command-reference/metrics#supported-file-formats

content/docs/command-reference/plots/diff.md
```diff
@@ -47,14 +47,17 @@ please see `dvc plots`.
 
 - `--targets <path>` - specific metric files to visualize. These must be listed
   in a [`dvc.yaml`](/doc/user-guide/dvc-files-and-directories#dvcyaml-file) file
-  (see the `--plots` option of `dvc run`). When specifying multiple `--targets`
-  and `revisions`, you may use `--` after this option's arguments, e.g.:
+  (see the `--plots` option of `dvc run`). When specifying arguments for
+  `--targets` before `revisions`, you may use `--` after this option's arguments
+  , e.g.:
```
imhardikj, line 52: @jorgeorpinel `--` will be in only when `--targets` precede `revisions`. Also even if only one `--targets` file is specified, `--` will have to be used. I think current reason is ambiguous so have made some changes here.

content/docs/command-reference/metrics/diff.md
imhardikj: Thanks! I will now verify everything before pushing changes.

content/docs/command-reference/plots/diff.md
```diff
@@ -47,14 +47,17 @@ please see `dvc plots`.
...
 
   ```dvc
   $ dvc plots diff --targets t1.json t2.csv -- HEAD v1 v2
   ```
 
-  Alternatively, you can also run above statement as:
+  Without `--`, this statement will show an error as argument parser confuses
+  `revisions` as arguments for `--targets` option. Alternatively, you can also
```
imhardikj, line 59: In https://github.com/iterative/dvc.org/pull/1567#pullrequestreview-447399307, I guess you asked me whether to keeps this part on not.  This will be helpful in case a user is still not sure why `--` has to be used.

content/docs/command-reference/metrics/diff.md
```diff
@@ -43,7 +43,21 @@ lists all the current metrics without comparisons.
 ## Options
 
 - `--targets <paths>` - limit command scope to these metric files. Using -R,
-  directories to search metric files in can also be given.
+  directories to search metric files in can also be given. When specifying
+  arguments for `--targets` before `revisions`, you may use `--` after this
+  option's arguments, e.g.:
+
+  ```dvc
+  $ dvc metrics diff --targets t1.json t2.json -- HEAD v1
```
jorgeorpinel, line 51: > Metrics can be organized as tree hierarchies in JSON or YAML files

Form https://dvc.org/doc/command-reference/metrics#supported-file-formats

```suggestion
  $ dvc metrics diff --targets t1.json t2.yaml -- HEAD v1
```

content/docs/command-reference/metrics/diff.md
```diff
@@ -43,7 +43,21 @@ lists all the current metrics without comparisons.
...
+  arguments for `--targets` before `revisions`, you may use `--` after this
+  option's arguments, e.g.:
+
+  ```dvc
+  $ dvc metrics diff --targets t1.json t2.yaml -- HEAD v1
+  ```
+
+  Without `--`, this statement will show an error as argument parser confuses
+  `revisions` as arguments for `--targets` option. Alternatively, you can also
+  run above statement as:
```
jorgeorpinel, line 56: Let's just avoid this sentence and change "may" for "must" or "should" in the previous one, please. Also, "the" is missing again 😞 

```suggestion
  Alternatively, you can also run the above statement as:
```

Same for plots diff, tnx

content/docs/command-reference/plots/diff.md
jorgeorpinel: Yeah but we also want this to be a short note. Probably no need to exact explanation as to why. People don't know what a parser is anyway.

Issue #1112, opened 2020-02-17
lint/pre-commit: forbid and fix long comments and string constants


fabiosantoscode:
All of our lints accept very long comments and strings, which go way past 80 characters per line.

The purpose of this issue is to make one of the lints (eslint or prettier) complain and fail the build.

Edit: Fail the build if any requires cannot be resolved.

UPDATE: See https://github.com/iterative/dvc.org/issues/1112#issuecomment-611822294 below.

shcheklein:
@fabiosantoscode @pavelgrinchenko am I right that we use `eslint` now? Is this one fixed?

shcheklein:
So, it's not fixed for long string constants, repurposed it to revisit and fix linters.

jorgeorpinel:
HI. Is this still relevant and if so can we remove the `migration` flag? Cc @rogermparent WDYT 🙂

jorgeorpinel:
In fact I took the liberty to remove the flag for now, please reinstate it or LMK if I'm wrong.

rogermparent:
> In fact I took the liberty to remove the flag for now, please reinstate it or LMK if I'm wrong.

I think it's safe to say this doesn't need to be marked as a migration issue.

jorgeorpinel:
> it's not fixed for long string constants

Do you know whether this is still an issue @rogermparent ?

To be honest I haven't bumped into this problem much if at all: in code blocks we manually keep them short to prevent horizontal scrolling anyway (and the col limit varies depending on browser, from 68 to 72). I guess if that can be linted it would be nice, but from what I remember `eslint` doesn't have that ability so I would understand if you prefer to close this.

rogermparent:
Do we have an example of one of these failing long string constants? My assumption is we're talking about something like `const LONG_CONSTANT = 'aaaaaaaaaaaaaaa'` where the amount of `a`s breaks 80 lines. I'd consider that fine for cases where it's needed, but if we want to break lines for something like:

```ts
const LONG_CONSTANT = 'aaaaaaaaaaaaaaaaaaaaaaaa'
  + 'aaaaaaaaaaaaaaaaaaaaa'
```

I have very little passion toward linting for this or not, but I'll take a look into using the `max-len` rule, see if it works, then close this whether or not we add it since that's mostly all it warrants IMO. Open to other opinions.

rogermparent:
[...]

PR #1576, opened 2020-07-14
Add functionality to link arbitrary strings in code blocks

### LINKED ISSUE #1431: basic concepts: pipeline stages
Other than explaining what a pipeline stage is, this can help us group 2 terms that are used extensively and sometimes confusingly throughout docs: `.dvc` files (formerly DVC-files) and `dvc.yaml` — which already have their own guide so maybe this can just use that page instead of a new basic concepts one. Not sure

See more context: https://github.com/iterative/dvc.org/pull/1408#issuecomment-643121300:

> Mentioning dvc.yaml and .dvc (everywhere) is too noisy

### Automatic linking

> Maybe should be a separate issue

- [x] `dvc.yaml` could be like a command name (done in #1370), which automatically links to its user guide. #1137
- [x] Not sure about `.dvc` file — I think here we don't want to link automatically. #1137
- [ ] dvc.lock ?
- ~~"stage" could have a tooltip and glossary entry.~~ Extracted to #1579
- [x] remove all existing manual links from dvc.yaml, dvc.lock, and .dvc file.
- "orphan stage" ~~we could link manually to dvc add~~ (term will be removed)
- "import stage" ~~to dvc import~~ (term will be removed)

Jump to https://github.com/iterative/dvc.org/issues/1431#issuecomment-658384344 for an update.

rogermparent:
currently the only usage is `dvc.yaml`, but this addition to the remark plugin is able to handle any matchers, simple to complex. If we find the need for generating urls based on content I can add it, but since it isn't necessary for the currently requested additional links it can be a low priority concern.

Fixes #1431


plugins/gatsby-remark-dvc-linker/simpleLinker.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env node */
+
+const { createLinkNode } = require('./helpers')
+
+const entries = [
```
shcheklein, line 5: better to move it into some configuration file

rogermparent: Sure, where you do think it should go? I'd assume at the root like `redirects-list.json`. Maybe `linked-terms.js`?

shcheklein: we can make part of `content/` I guess ... along the community.json?

rogermparent: Sure!

plugins/gatsby-remark-dvc-linker/simpleLinker.js
```diff
@@ -0,0 +1,32 @@
...
+const useMatcher = (matcher, item) => {
+  switch (typeof matcher) {
+    case 'string':
+      return item === matcher
+    case 'object':
+      if (Array.isArray(matcher))
+        return matcher.find(submatcher => useMatcher(submatcher, item))
+      if (matcher instanceof RegExp) return matcher.match(item)
+    default:
+      throw `gatsby-remark-dvc-linker simpleLinker given bad matcher of type "${typeof matcher}"`
```
shcheklein, line 16: looks like linters are not applied to this file?

rogermparent: Maybe? I'm not sure what format rules are being broken, as running prettier manually doesn't force make any changes.

shcheklein: ah, I see it the same as https://github.com/iterative/dvc.org/issues/1112 probably

rogermparent:
An example of this working can be found [here](https://dvc-landing-new-auto-li-fwr1dg.herokuapp.com/doc/command-reference/run#options), (or rather, the first instance of "dvc.yaml" after that heading)

shcheklein:
let's run some script to remove already existing links?

@jorgeorpinel could you please provide a list of things to link?

(@rogermparent @jorgeorpinel it can be sone as a separate ticket, feel free to merge this one PR then)

rogermparent:
> let's run some script to remove already existing links?

I can ripgrep for `[dvc.yaml]` and remove all existing instances, replacing them with this behavior.

shcheklein:
@rogermparent sounds good to me, let's start with that

rogermparent:
Just pushed a commit that "unwraps" all links that contain `dvc.yaml`, including those that contain other text like "`dvc.yaml` file", making usage of the auto-link feature for this term consistent across the repo.

rogermparent:
(Originally from #1431)

From what I understand, there's three currently proposed entries for the automatic linker

* [x] dvc.yaml could be like a command name, which automatically links to its user guide
* [ ] "orphan stage" we could link manually to dvc add
* [ ] "import stage" to dvc import

shcheklein:
@jorgeorpinel could you please review and share the list related to the DVC files?

rogermparent:
It seems like "import stage" and "orphan stage" aren't used in code blocks like the current linker checks for, but I could add additional functionality to operate on italic and bold text in the same way if we want these terms auto-linked but not in code quotes.

content/linked-terms.js
```diff
@@ -0,0 +1,6 @@
+module.exports = [
+  {
+    matches: 'dvc.yaml',
+    url: '/doc/user-guide/dvc-files-and-directories#dvcyaml-file'
+  }
```
jorgeorpinel, line 5:
```suggestion
  }
  {
    matches: 'dvc.yaml',
    url: '/doc/user-guide/dvc-files-and-directories#dvcyaml-file'
  }
  {
    matches: 'dvc.lock',
    url: '/doc/user-guide/dvc-files-and-directories#dvclock-file'
  }
  {
    matches: ['`.dvc` file', '`.dvc` file'],
    url: '/doc/user-guide/dvc-files-and-directories#dvclock-file'
  }
```
The last one probably won't work like this though. It's tricky because the back quotes don't wrap the full text...

jorgeorpinel: > The last one probably won't work like this

If that's too difficult to implement or problematic for some reason, we can just link `.dvc` I guess. That should be fine and it's easier, so that last one would be:

```js
  {
    matches: '`.dvc`,
    url: '/doc/user-guide/dvc-files-and-directories#dvclock-file'
  }
```

jorgeorpinel: > We just have to be careful to always use `.dvc/` when referring to the directory.

rogermparent:
> Sorry, just noticed this.

No problem!

>     * .dvc file(s) to /doc/user-guide/dvc-files-and-directories#dvc-files
> 
>     * dvc.yaml to /doc/user-guide/dvc-files-and-directories#dvcyaml-file
> 
>     * dvc.lock to /doc/user-guide/dvc-files-and-directories#dvclock-file

I can definitely do these!

> Or maybe all to just /doc/user-guide/dvc-files-and-directories ?

That's up to preference, no difference either way.

> Don't worry about "orphan/import stgae" — we're not using those terms anymore.

Sure!

> Yes, this is needed, but for all `` [`dvc.yaml`] ``, `` [`dvc.lock`] ``, and `` `.dvc` file ``/`` `.dvc` files ``. (We could do this separately if to keep this PR smallish.)

Considering I've already done it for dvc.yaml, I think it's fine to do this with the rest to introduce the feature.

> The last one probably won't work like this though. It's tricky because the back quotes don't wrap the full text...

Correct, currently the auto-linker specifically checks individual MDAST nodes. This works for us because code blocks are always their own individual nodes, even when inline. To capture words around the block, like file in "`.dvc` file", would require a large amount of re-rigging of the autolinker to consider surrounding words. Autolinking `.dvc` and unwrapping links containing the text will convert all existing instances to auto-linked ones.

jorgeorpinel:
> To capture words around the block, like file in ".dvc file", would require a large amount of re-rigging of the autolinker

I say we just do `.dvc` then. I already changed all the ones referring to the directory to `.dvc/` anyway, I think.

BTW maybe we could also add `.dvc/` -> https://dvc.org/doc/user-guide/dvc-files-and-directories#internal-directories-and-files — or even //`.dvc/[^/]`// if regexps are allowed (to catch things like .dvc/cache and .dvc/config). However the scope would further expand so def. should extract that part to a separate issue/PR

rogermparent:
Here's what we've got in this PR

new auto-linked terms:

- [x] `.dvc` to /doc/user-guide/dvc-files-and-directories#dvc-files
- [x] `dvc.yaml` to /doc/user-guide/dvc-files-and-directories#dvcyaml-file
- [x] `dvc.lock` to /doc/user-guide/dvc-files-and-directories#dvclock-file

terms must be surrounded in code quotes (backticks) to be recognized by the auto-linker. Terms with parts that aren't code-quoted (like "`.dvc` file") will only have the code-quoted part linked by the auto-linker. This issue can be fixed in a few ways, but is more of a "nice-to-have" when compared to the rest of this whole feature which is intended to make the editing process easier.

rogermparent:
Unless there's an unexpected regression something else to add, I think we're ready to merge this PR!

content/docs/command-reference/move.md
```diff
@@ -1,7 +1,6 @@
 # move
 
-Rename a file or a directory and modify the corresponding
-[`.dvc` file](/doc/user-guide/dvc-files-and-directories#dvc-files) (see
+Rename a file or a directory and modify the corresponding `.dvc` file (see
 `dvc add`) to reflect the change. If the file or directory has the same name as
 the corresponding `.dvc` file, it also renames it.
```
jorgeorpinel, line 5:
```suggestion
the `.dvc` file, it also renames it.
```

content/docs/command-reference/import.md
```diff
@@ -114,11 +111,9 @@ Importing 'data/data.xml (git@github.com:iterative/example-get-started)'
 ```
 
 In contrast with `dvc get`, this command doesn't just download the data file,
-but it also creates an import stage
-([`.dvc` file](/doc/user-guide/dvc-files-and-directories#dvc-files)) with a link
-to the data source (as explained in the description above). (This import stage
-can later be used to [update](/doc/command-reference/update) the import.) Check
-`data.xml.dvc`:
+but it also creates an import stage (`.dvc` file) with a link to the data source
```
jorgeorpinel, line 114: Note for self: replace all these "import stage" term instances in a separate PR.

content/linked-terms.js
```diff
@@ -0,0 +1,14 @@
+module.exports = [
```
jorgeorpinel, line 1: Could this be extracted to a simple JSON or YAML config file so we can easily edit the list later? I mean... It's already easy but would be even easier 🙂

Also, should a test be written for this fn?

shcheklein:
thanks @rogermparent and @jorgeorpinel !

content/linked-terms.js
rogermparent: It could, but using `.js` allows us to use RegExp literal matchers and dynamic replacement values in a future. It's possible to switch to JSON by simply converting what's there now if we decide we don't want those features.

PR #1582, opened 2020-07-16
dvc run tool-tip

### LINKED ISSUE #1579: glossary: create "stage" entry
and replace all the links from "stage" to the `dvc run` cmd ref for `<abbr>` tags, so the term gets a tooltip.

utkarshsingh99:
Fix #1579 

I have only made changes to one file for adding <abbr> tag -> `cmd ref/dvc add` for preview of the tooltip formatting
This is because markdown's code syntax conflicts with the <abbr> tag in the Tooltip component.
Thus, our only option currently is to lose the code block and make `dvc run` look like this:
Screenshot from 2020-07-16 19-33-09
Also, I've kept a link to the cmd ref page of dvc run too in the tooltip so it's easier to navigate in case of more complex doubts by the user.
P.S: Also, a small grammatical correction, that I've added along with this.

@jorgeorpinel @shcheklein


content/docs/user-guide/basic-concepts/dvc-run.md
```diff
@@ -0,0 +1,9 @@
+---
+name: 'dvc run'
```
jorgeorpinel, line 2: `dvc run` is not a basic concept nor does it need a glossary entry. All cmd names are already linked automatically to their refs.

Original ticket: **glossary: create "stage" entry** #1579

jorgeorpinel:
@utkarshsingh99 that ticket is about the term "stage", not "dvc run". Please try to work on the **Next priorities** in #1255 for now like I mentioned in a GSoD email. Thanks

PR #1587, opened 2020-07-17
Improve CloudFlare cache clear logging and GitHub Stars bugfix

rogermparent:
This PR was made to help diagnose the CloudFlare cache purge issue we've been having, though Review Apps don't actually have CloudFlare credentials.

The changes of this PR move the CloudFlare cache cleaning logic into a regular Node module instead of a CLI-invoked script. This module is then used at the end of the deploy script. Effectively, this moves the CloudFlare cache clean to "the end of the deploy phase" as opposed to "the start of the runtime process".

As I understand it, this effectively doesn't change anything except for logs. The log changes, however, mean that the CloudFlare script outputs in the build logs instead of the runtime logs- this is good because runtime logs expire without add-ons, but build-time logs last forever by default.

You can see in the Deploy Preview for this PR that the cache clear module is being run, as it outputs the line `Skipping CloudFlare cache purge because CONTEXT is not set to production.`. To actually test it out, I just need to manually add keys to the deploy app's env vars. I'll do another update when I do.

Update: When testing with the keys in prod on the Heroku review app, this error appears: 
```
Error: CloudFlare cache clear failed! {
  "status": 403,
  "errors": [
    {
      "code": 10000,
      "message": "Authentication error"
    }
  ]
}
```
When testing with a new token on a Review App, the CloudFlare cache is successfully cleared. This could either be because the token is invalid or deploy apps fail an auth check on the old key that prod passes. Either way, I think it's worth trying this new log behavior and a new key on the production Heroku instance.

Update: While working on this issue, I ran into a bug related to GitHub stars that would break our cache re-use occasionally and force a full rebuild- the fix was to turn the static stars GraphQL endpoint into a simple resolver instead of a true Node, solving the issue that necessitated the separate node as well as the cache issue the first iteration of the fix introduced.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -13,6 +13,7 @@ exports.createSchemaCustomization = api =>
 exports.sourceNodes = api => callOnModels(models, 'sourceNodes', api)
 exports.onCreateNode = api => callOnModels(models, 'onCreateNode', api)
 exports.createPages = api => callOnModels(models, 'createPages', api)
+exports.createResolvers = api => callOnModels(models, 'createResolvers', api)
 exports.onPostBuild = api => callOnModels(models, 'onPostBuild', api)
 
 exports.onCreatePage = ({ page, actions }) => {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "develop": "gatsby develop",
     "dev": "gatsby develop",
     "build": "gatsby build",
-    "start": "./scripts/clear-cloudflare-cache.js; node ./src/server/index.js",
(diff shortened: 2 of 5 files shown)


shcheklein:
do we want to merge this?

rogermparent:
I believe we do want to merge this, the primary change will mean any future CloudFlare cache cleaner problems will be much more visible (i.e. if the new token I make ever expires for some reason)
On top of that, I managed to pin down and fix a hard-to-replicate bug while working with this PR on Heroku that would occasionally break Gatsby cache and trigger a full rebuild.

PR #1584, opened 2020-07-17
pipeline: links and examples update for 1.x

utkarshsingh99:
- [ ] UPDATE: Merge https://github.com/iterative/dvc.org/pull/1591#pullrequestreview-453804426 after this one.

---
- Continuation of https://github.com/iterative/dvc.org/pull/1383#issuecomment-648363086

- One of the *Next Priorities* in DVC 1.x discrepancies #1255:
  - [x] dvc pipeline -> dvc dag #1383 Need to update pipeline links, for now they just redirect to dag. Also push/pull examples.

~~Line 169 in `repro.md` links to `/doc/tutorial/dags` and previously linked to `/doc/tutorial/pipelines`. Both these links just redirect to `doc/start`. Maybe they should have a different link (`/doc/start/data-pipelines`)?~~

PR #1591, opened 2020-07-18
cmd: update push deps example

### LINKED ISSUE #1445: term: stop using Dvcfile

imhardikj:
Close #1445 by addressing the last pending checkbox:
* [x] push cmd ref.

### DIFF
--- a/content/docs/command-reference/push.md
+++ b/content/docs/command-reference/push.md
@@ -155,40 +155,14 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
 [stages](/doc/command-reference/run):
 
 ```dvc
-$ cat dvc.yaml
-stages:
-  Posts-xml:
-    cmd: unzip data/Posts.xml.zip -d data/
-    deps:
-    - data/Posts.xml.zip
-    outs:
-    - data/Posts.xml
-  Posts-tsv:
-    cmd: python3 src/Posts.py
-    deps:
-    - data/Posts.xml
-    - src/Posts.py
-    outs:
-    - Posts.tsv
-  Posts-test-tsv:
-    cmd: python3 src/Posts-test.py
-    deps:
-    - Posts.tsv
-    - src/Posts-test.py
-    outs:
-    - Posts-test.tsv
-  matrix-train:
-    cmd: python3 src/matrix-train.py
-    deps:
-    - Posts-test.tsv
-    - Posts.tsv
-    - src/matrix-train.py
-    outs:
-    - data/matrix-test.p
-    - data/matrix-train.p
-    - data/model.p
-    params:
-    - matrix-train.train
+$ dvc pipeline show
+data/Posts.xml.zip.dvc
+Posts.xml.dvc
+Posts.tsv.dvc
+Posts-test.tsv.dvc
+matrix-train.p.dvc
+model.p.dvc
+Dvcfile
 ```
 
 Imagine the <abbr>projects</abbr> has been modified such that the
@@ -207,7 +181,7 @@ One could do a simple `dvc push` to share all the data, but what if you only
 want to upload part of the data?
 
 ```dvc
-$ dvc push --with-deps Posts-tsv
+$ dvc push --with-deps test-posts
 
 ... Do some work based on the partial update
 
@@ -220,9 +194,9 @@ $ dvc status --cloud
 Data and pipelines are up to date.
 ```
 
-We specified a stage in the middle of this pipeline (`Posts-tsv`) with the first
-push. `--with-deps` caused DVC to start with that stage, and search backwards
-through the pipeline for data files to upload.
+We specified a stage in the middle of this pipeline (`test-posts`) with the
+first push. `--with-deps` caused DVC to start with that `.dvc` file, and search
+backwards through the pipeline for data files to upload.
 
 Because the `matrix-train` stage occurs later (it's the last one), its data was
 not pushed. However, we then specified it in the second push, so all remaining
--- a/content/docs/command-reference/push.md
+++ b/content/docs/command-reference/push.md
@@ -155,14 +155,28 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
 [stages](/doc/command-reference/run):
 
 ```dvc
-$ dvc pipeline show
-data/Posts.xml.zip.dvc
-Posts.xml.dvc
-Posts.tsv.dvc
-Posts-test.tsv.dvc
-matrix-train.p.dvc
-model.p.dvc
-Dvcfile
+$ dvc dag
+             +-------------+
+             | prepare.dvc |
+             +-------------+
+                    *
+                    *
+                    *
+            +---------------+
+            | featurize.dvc |
+            +---------------+
+             **            **
+           **                **
+         **                    **
++-----------+                    **
+| train.dvc |                  **
++-----------+                **
+             **            **
+               **        **
+                 **    **
+            +--------------+
+            | evaluate.dvc |
+            +--------------+
 ```
 
 Imagine the <abbr>projects</abbr> has been modified such that the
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/push.md
```diff
@@ -155,37 +155,51 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
...
-$ dvc status --cloud
-
-  new:            data/model.p
-  new:            data/matrix-test.p
-  new:            data/matrix-train.p
+$ dvc status -c
+    new:            data/featurize/train.pkl
+    new:            data/featurize/train.pkl
+    new:            data/prepared/train.tsv
+    new:            data/prepared/test.tsv
```
imhardikj, line 191: @jorgeorpinel `status`, `pull`, `push` doesn't seem to work for example-get-started repo. I have made up this output myself.

content/docs/command-reference/push.md
```diff
@@ -155,37 +155,51 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
...
-data/Posts.xml.zip.dvc
-Posts.xml.dvc
-Posts.tsv.dvc
-Posts-test.tsv.dvc
-matrix-train.p.dvc
-model.p.dvc
-Dvcfile
+$ dvc dag
+         +---------+
+         | prepare |
```
jorgeorpinel, line 160: Idk if we want to do this change or just something else completely e.g. `cat dvc.yaml` but seems out of scope for this PR given the title?

content/docs/command-reference/push.md
jorgeorpinel: Should work. What errors are you getting?

But let's not change the example stages. You can make dummy data files and stages based on the existing file and stage names in this example e.g. `Posts.xml`, `data/model.p, etc.

content/docs/command-reference/push.md
```diff
@@ -155,37 +155,51 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
...
+    new:            data/prepared/train.tsv
+    new:            data/prepared/test.tsv
 ```
 
 One could do a simple `dvc push` to share all the data, but what if you only
 want to upload part of the data?
 
 ```dvc
-$ dvc push --with-deps matrix-train.p.dvc
+$ dvc push --with-deps featurize
```
jorgeorpinel, line 198: Stage name should be something like `test-posts`

content/docs/command-reference/push.md
```diff
@@ -155,37 +155,51 @@ a [pipeline](/doc/command-reference/pipeline) has been setup with these
...
 
 ... Do some work based on the partial update
 
-$ dvc push --with-deps model.p.dvc
+$ dvc push --with-deps evaluate
```
jorgeorpinel, line 202: `matrix-train`

content/docs/command-reference/push.md
imhardikj: I can think the problem is with the remote of `example-get-started` repo . This was the error:
`ERROR: unexpected error - Unable to locate credentials`

content/docs/command-reference/push.md
imhardikj: This is actually based on example get started repo. I excluded the `data/data.xml.dvc`, as it was taking to much space. This is similar to [dag example](https://dvc.org/doc/command-reference/dag#examples)

imhardikj:
@jorgeorpinel I have removed `example-get started` repo and instead set up a project for original example using dummy data and files.
I have used `$cat dvc.yaml` here currently but the `dvc dag` visualization of this dummy project can be viewed in this db08e42997734fcc38bf9655eab34f48ce41d655 commit.

content/docs/command-reference/push.md
jorgeorpinel: Strange. At `dvc status -c`?

content/docs/command-reference/push.md
jorgeorpinel: What do you mean this is based on Get Started? No, look at the stage names, Posts.xml.dvc, Posts.tsv.dvc, etc.

Anyway... How is this related to removing "Dvcfile" again? The cat dvc.yaml idea wasn't good, now it's even longer. Way too long. Let's just not change this for now and focus on the original PR scope?

content/docs/command-reference/push.md
jorgeorpinel: I see you went for `Posts-tsv`. Not a great stage name. It should be a descriptive action. I left specific suggestions...

content/docs/command-reference/push.md
jorgeorpinel: Anyway, is this even relevant anymore? 🙂

content/docs/command-reference/push.md
jorgeorpinel: Yeah just rever this pleas.e It's being taken care of in #1584

jorgeorpinel: P/a I see it's related because this is where `Dvcfile` gets removed. But yeah, lets just revert thia part since it's being fixed elsewhere already. This way we can merge the rest of this work and move on 🙂

shcheklein:
hey, thanks @imhardikj , but I still see `Dvcfile` in the push.md - I think we should get rid of it completely to close this ticket?

content/docs/command-reference/push.md
```diff
@@ -181,11 +181,11 @@ One could do a simple `dvc push` to share all the data, but what if you only
 want to upload part of the data?
 
 ```dvc
-$ dvc push --with-deps matrix-train.p.dvc
+$ dvc push --with-deps test-posts
```
shcheklein, line 184: we need to rewrite the example above, the project structure for this command to make sense?

shcheklein: also, why do change the stage name here, intentional?

content/docs/command-reference/push.md
```diff
@@ -181,11 +181,11 @@ One could do a simple `dvc push` to share all the data, but what if you only
...
 
 ... Do some work based on the partial update
 
-$ dvc push --with-deps model.p.dvc
+$ dvc push --with-deps matrix-train
```
shcheklein, line 188: same here - why do we need to change the stage name?

jorgeorpinel:
> thanks @imhardikj , but I still see Dvcfile in the push.md

Sorry that wsa confusing but I set it up that way: someone else was working on that part of the code so I asked Hardik to leave it alone. Now both PRs are merged and `Dvcfile` is gone from the code base 🙂

content/docs/command-reference/push.md
jorgeorpinel: > rewrite the example above, the project structure for this command to make sense

Yes! It was also done in the complementary PR. Both are merged, please check in https://dvc.org/doc/command-reference/push#example-with-dependencies

> why do change the stage name here, intentional?

Intentional. I came up with these new names. But now that you mention it `matrix-train` would've been the obvious choice here 🤦

jorgeorpinel: Wait no, that't the name of the next stage. That's shy I came up with `test-posts`.

content/docs/command-reference/push.md
jorgeorpinel: I don't think `model` is a good stage name.

PR #1600, opened 2020-07-20
Improve CloudFlare cache cleaner logging + Stars bugfix (second attempt)

rogermparent:
This is the same PR as before, but with a slight change to fix the server invocation.

This time I'll confirm the review app before I suggest a merge.


rogermparent:
Yep, adding `node` back to the invocation makes the proxy server actually work!

https://dvc-landing-fix-cloudfl-jkfbz8.herokuapp.com/

shcheklein:
Deploying? :)

rogermparent:
Yep, it's deploying successfully now! With the slight change to the server, deploy previews are successful. I must have been using only `gatsby serve` before and missed where I broke the server invocation because it can't be directly run like `clear-cloudflare-cache` was before.

rogermparent:
Yay! The site's not broken!

PR #1602, opened 2020-07-20
cmd: rewrite push/pull et al.

jorgeorpinel:
Continuation of #1384 with out-of-scope details. Waiting for that PR to close.


jorgeorpinel:
I think this is ready to merge. Needs an approval though.

content/docs/command-reference/push.md
```diff
@@ -18,55 +18,48 @@ positional arguments:
...
-remote is used (see `dvc config core.remote`) unless the `--remote` option is
-used. See `dvc remote` for more information on how to configure a remote.
+Under the hood, a few actions are taken:
 
-With no arguments, just `dvc push` or `dvc push --remote REMOTE`, it uploads
-only the files (or directories) that are new in the local repository to remote
-storage. It will not upload files associated with earlier commits in the
-<abbr>repository</abbr> (if using Git), nor will it upload files that have not
-changed.
+- The push command checks the appropriate `dvc.lock` and `.dvc` files in the
```
shcheklein, line 52: dvc.yaml?

skshetry: It's a command-reference, but I don't think this information should be here (perhaps on Dvc files?).

Also, @jorgeorpinel, `dvc.yaml` is a _discovery_ file which dvc reads to find outputs/dependencies, etc. and `dvc.lock` is a file to correlate those outputs/deps with the exact version through the checksums. `dvc.lock` is never used if there's no corresponding deps/outs/stages in dvc.yaml. So, either both dvc.yaml and dvc.lock are read together or we could just say dvc.yaml is read (and consider dvc.lock as an implementation detail).

shcheklein: > It's a command-reference, but I don't think this information should be here 

could you clarify please what information exactly do you mean?

> So, either both dvc.yaml and dvc.lock are read together or we could just say dvc.yaml is read (and consider dvc.lock as an implementation detail) ..

we come back to a single term "DVC files" w/o specifying details with some tooltip that mentions that DVC files == dvc.yaml + dvc.lock + .dvc . All of those files play their role here. We can expand later and bit explicit if needed where it is needed.

jorgeorpinel: dvc.yaml is mentioned just before: "`targets` ... limit what to pull. It accepts ... and stage names (found in `dvc.yaml`)". Then, when we say "push checks the appropriate `dvc.lock` files", by "appropriate" we mean corresponding to what was found in dvc.yaml files. But let me see if I can clarify this further... ⌛ 

> just say dvc.yaml is read (and consider dvc.lock as an implementation detail

This is not a bad idea... We do go over implementation details in cmd refs though, in fact they'r probably the most low-level docs we have. But we should prob. hide such details in expandable sections. Just not sure dvc.lock should be considered a low-level detail, it's pretty visible to the user.

> we come back to a single term "DVC files" w/o specifying details with some tooltip that mentions that DVC files

Yes, let's try to merge this one as best as possible though, and follow up on this in #1663 ?

jorgeorpinel: > See also iterative/dvc/issues/4393

jorgeorpinel: > let me see if I can clarify this further... ⌛

Turn this whole bullet list into a more simple paragraph in https://github.com/iterative/dvc.org/pull/1602#pullrequestreview-465516419. PTAL

content/docs/command-reference/push.md
```diff
@@ -18,55 +18,44 @@ positional arguments:
...
-these commands.
+[metrics](/doc/command-reference/metrics), etc.) remotely are the most common
+use cases for these commands.
 
-The `dvc push` command allows us to upload data to remote storage. It doesn't
-save any changes to the code, `dvc.yaml`, or `.dvc` files (those should be saved
-with `git commit` and `git push`).
+The `dvc push` command uploads data to
+[remote storage](/doc/command-reference/remote). It doesn't save any changes to
+the code, `dvc.yaml`, or `.dvc` files (that should be saved with `git commit`
```
shcheklein, line 31: should include `dvc.lock` here ...

content/docs/command-reference/push.md
```diff
@@ -18,55 +18,44 @@ positional arguments:
...
-Under the hood a few actions are taken:
+The default remote is used (see `dvc config core.remote`) unless the `--remote`
+option is used. See `dvc remote` for more information on how to configure a
+remote.
 
-- The push command by default uses all stages (in `dvc.yaml` and `dvc.lock`) and
-  `.dvc` files in the <abbr>workspace</abbr>. The command options will either
-  limit or expand the set of stages or `.dvc` files to consult.
+Without arguments, it uploads all files and directories missing from remote
+storage, found as <abbr>outputs</abbr> in the stages (in `dvc.lock`) or `.dvc`
```
shcheklein, line 39: dvc.yaml?

content/docs/command-reference/push.md
```diff
@@ -18,55 +18,44 @@ positional arguments:
...
+`git push`. See `dvc install` for more details.
 
-With no arguments, just `dvc push` or `dvc push --remote REMOTE`, it uploads
-only the files (or directories) that are new in the local repository to remote
-storage. It will not upload files associated with earlier commits in the
-<abbr>repository</abbr> (if using Git), nor will it upload files that have not
-changed.
+For all <abbr>outputs</abbr> referenced in each target, DVC finds the
+corresponding files and directories in the <abbr>cache</abbr> (identified by
+hash values saved in `dvc.yaml` and `.dvc` files). DVC then gathers a list of
```
shcheklein, line 52: hash values are not saved in dvc.yaml

content/docs/command-reference/push.md
jorgeorpinel: True. This whole sentence is kind of a note though, now that I think about it. I changed it a little in 4717b78b. It's still not perfect but we'll have to review this again soon, when addressing #1663.

content/docs/command-reference/push.md
jorgeorpinel: Hmmm yeah... Kinda. dvc.lock is where you find the output file names — that's what this meant. But probably this paragraph should be simplified and not even mention either one. See 246f446c.

content/docs/command-reference/push.md
jorgeorpinel: Typo... Thanks for catching that one. Fixed.

jorgeorpinel:
> stop and find a simple solution

I agree. But I don't think there's a simple solution if we want to be thorough in the explanations until iterative/dvc/issues/4278 or related discussions are figured out. What we can do now is avoid going into details (where stage names are listed, where hash values are, etc.) and just talk about stages and maybe .dvc files. Users will have to know that stages are in dvc.yaml and what dvc.lock does.

> something similar to DVC files + expand where we explain exact mechanism

Unfortunately, even if I agreed on that term, I think that we can't go back to pre 1.x simplicity. This is because DVC 1.x is more complex, since it introduced dvc.yaml and lockfiles **in addition** to DVC-files.

But I agree expandable sections can be our friend here.

My only question is can we merge this so far and follow the above going fwd? Or do you see still specific problems in these changes @shcheklein? Thanks

shcheklein:
> My only question is can we merge this work so far and follow the above going fwd? (even have an immediate PR on these same docs) Or do you see still specific problems in these changes @shcheklein? Thanks

the problem we don't merge this is not the "DVC metafiles" issue, but some actual factual problems (e.g. dvc.lock -> dvc.yaml).

content/docs/command-reference/pull.md
```diff
@@ -18,40 +18,42 @@ positional arguments:
...
-The `dvc pull` command allows one to retrieve data from remote storage.
-`dvc pull` has the same effect as running `dvc fetch` and `dvc checkout`
-immediately after.
+`dvc pull` downloads data from [remote storage](/doc/command-reference/remote)
+and places it in the <abbr>workspace</abbr>. It has the same effect as running
+`dvc fetch` and `dvc checkout`.
 
-The default remote is used (see `dvc config core.remote`) unless the `--remote`
+> Note that pulling data does not change any `dvc.yaml` or `.dvc` files, nor
+> does it save any changes to the code, `dvc.lock`, or `.dvc` files (that should
```
shcheklein, line 34: not sure why would I expect it to save changes in the first place?

content/docs/command-reference/pull.md
```diff
@@ -18,40 +18,42 @@ positional arguments:
...
-[metrics](/doc/command-reference/metrics), etc) remotely (S3, SSH, GCS, etc.)
-are the most common use cases for these commands.
+[metrics](/doc/command-reference/metrics), etc.) remotely are the most common
+use cases for these commands.
 
-The `dvc pull` command allows one to retrieve data from remote storage.
-`dvc pull` has the same effect as running `dvc fetch` and `dvc checkout`
-immediately after.
+`dvc pull` downloads data from [remote storage](/doc/command-reference/remote)
+and places it in the <abbr>workspace</abbr>. It has the same effect as running
```
shcheklein, line 30: `places into workspace` - not a good term, sounds like it literally downloads them into workspace

content/docs/command-reference/pull.md
```diff
@@ -18,40 +18,42 @@ positional arguments:
...
-`.dvc` files and stages (in `dvc.yaml` and `dvc.lock`) currently in the
-<abbr>project</abbr>. It will not download files associated with earlier commits
-in the <abbr>repository</abbr> (if using Git), nor will it download files that
-have not changed.
-
-The command `dvc status -c` can list files referenced in current stages (in
-`dvc.yaml`) or `.dvc` files, but missing from the <abbr>cache</abbr>. It can be
-used to see what files `dvc pull` would download.
+Without arguments, it downloads all files and directories missing from the
+project, found as <abbr>outputs</abbr> of the stages or `.dvc` files present in
```
shcheklein, line 42: not only .dvc right?

shcheklein: Ah I see that stages as well here .. I completely missed this

shcheklein: stages should be highlighted I guess somehow?

content/docs/command-reference/pull.md
jorgeorpinel: True. I copied this from pull without editing accordingly. Updated.

> I've also applied similar notes to fetch and commit, but didn't push it in this PR for now. I'll make another small PR soon.

content/docs/command-reference/pull.md
jorgeorpinel: True. Updated this language to be more precise in both refs.

content/docs/command-reference/pull.md
jorgeorpinel: I linked it to the `run` ref where stages are explained for now. Those links will be changed to either a glossary tooltip or basic concepts link soon. Also phrases like "stages or/and .dvc files" should also be replaced soon when we figure out the best term to encompass all these things...

jorgeorpinel:
OK. Well, I'll merge this to move forward and come back to cleaning up all these refs.

One thing that improved was the correctness of DVC metafile mentions, although it will have to be updated again soon. Another thing that improved is the consistency among these as well as fetch, and to a lesser degree also commit and status.

PR #1606, opened 2020-07-20
Refactor header and fix banner+placeholder positioning issues

### LINKED ISSUE #1592: banner breaks positions for other elements
After we introduced the banner we broke a few things: 

Screen Shot 2020-07-18 at 11 30 23 AM

Screen Shot 2020-07-18 at 11 31 47 AM

rogermparent:
This PR primarily aims to make the Header placeholder grow and shrink while
accounting for the extra height of the alert banner, fixing many positioning
issues that were introduced with the banner addition.

This PR also refactors the Hamburger menu into three relocatable 
parts: a hook for the state and logic, a component for the button, and another 
component for the menu overlay. This helps with readability of the different parts, 
and will make possible changes in a future a bit easier since these disparate parts
are now less glued together in source code.

This fixes the positioning issues I've seen with the banner, but if others still exist feel 
free to post them in this thread so I can get to them.

Fixes #1592


src/utils/front/hamburgerMenu.ts
```diff
@@ -0,0 +1,52 @@
+import {
```
shcheklein, line 1: should we put it close to the HamburgerMenu code? it's not generic enough unless I'm missing something to keep it outside

rogermparent: I can do that! I just placed it next to the other hooks by default. It probably makes more sense to keep all three in the same module since they must be used together to work properly.

shcheklein:
thanks @rogermparent !

PR #1615, opened 2020-07-22
Misc. updates

### LINKED ISSUE #1638: cmd: explain init dvc nested dvc projects/repos and relationship to --subdir
## Report

UPDATE: Jump to https://github.com/iterative/dvc.org/issues/1638#issuecomment-665745290

---
> Context: https://discord.com/channels/485586884165107732/563406153334128681/737729577992061059

```console
$ mkdir test && cd test
$ git init && dvc init
  # So far so good
$ mkdir dir && cd dir
$ git init && dvc init
  # No problem (expecting ERROR?)

$ cd ..
$ dvc destroy -f
$ dvc init
  # No problem (should WARN?)
```

- (2nd example above:) should it be enforced or warned in dirs that contain a subdir with a dvc repo?
- ~~Also, should it also be enforced even with `dvc init --no-scm`?~~ OK no point in this, but maybe should be mentioned in the docs.

### Please provide information about your setup

**Output of `dvc version`:**

```console
$ dvc version
DVC version: 1.2.1+b14ad4.mod
Python version: 3.8.2
Platform: Windows-10-10.0.18362-SP0
Binary: False
Package: None
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Repo: dvc, git
Filesystem type (workspace): ('NTFS', 'C:\\')
```

**Additional Information (if any):**

If applicable, please also provide a `--verbose` output of the command, eg: `dvc add --verbose`.

jorgeorpinel:
Triggered as out-of-scope from other tickets/PRs
- more repro updates per https://github.com/iterative/dvc.org/pull/1572#pullrequestreview-453722844 +
- repro examples per https://github.com/iterative/dvc.org/pull/1572#pullrequestreview-447454431 
- updates to init to close #1638
- std. `--jobs` option description in all refs.
- other


content/docs/command-reference/repro.md
```diff
@@ -286,3 +286,7 @@ $ dvc dag
       |  count  |
       `---------'
 ```
+
+> Note that using `dvc repro --downstream` without a target will always have a
+> similar effect, where all previous stages are ignored — only if the last stage
+> is changed will it have any effect.
```
jorgeorpinel, line 292: I guess let's just update the example to insert a 3rd stage and target the 2nd one with --downstream, instead of needing this note. Although the note could be useful in the option desc.

sarthakforwet: We can also extend the `--downstream` opt. with the sentence "If no `targets` are provided, this option is ignored" along with an update in the example.

jorgeorpinel: Yes. But it's not exactly ignored. It just doesn't have any effect.

sarthakforwet: Yeah actually, could we replace "this option is ignored" with "this option has no effect"?

content/docs/command-reference/repro.md
```diff
@@ -268,14 +269,13 @@ $ dvc repro --downstream
...
 ```
 
-The reason being that the `text.txt` file is a dependency in the last stage of
-the pipeline (used by default by `dvc repro`), This last `count` stage is
-dependent on `filter` stage, which happens first in this pipeline (shown in the
-following figure):
+The reason being that the `text.txt` file is not a dependency in the last stage
+of the pipeline, used as the default target by `dvc repro`. `text.txt` is a
+dependency of the `filter` stage, which happens earlier (shown in the figure
+below), so it's skipped given the `--downstream` option.
```
sarthakforwet, line 275: I think there is a mistake here as `filter` stage would not be skipped but it will be executed and hence the complete pipeline with an output -

image

content/docs/command-reference/repro.md
jorgeorpinel: @sarthakforwet please apply the changes without waiting for me to confirm every action. It's not a sustainable workflow. I can always review them later.

content/docs/command-reference/repro.md
jorgeorpinel: The whole example has to be rewritten as mentioned in https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-453636252

sarthakforwet:
@jorgeorpinel  would you also be updating the examples? I have updated them locally.

jorgeorpinel:
@sarthakforwet I thought we agreed you'd do that in a PR to this branch? See https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-453636252

jorgeorpinel:
I mean https://github.com/iterative/dvc.org/pull/1572#issuecomment-662673042. But the conversation in this PR is also relevant. Thanks

sarthakforwet:
> @sarthakforwet I thought we agreed you'd do that in a PR to this branch in #1572 (comment)?

Yes, added it. Thanks!

content/docs/command-reference/init.md
```diff
@@ -32,67 +32,60 @@ that are hidden from the user. This directory is automatically staged with
...
-- Not enough isolation/granularity - commands like `dvc pull`, `dvc checkout`,
-  and others analyze the whole repository to look for `dvc.yaml` or `.dvc` files
-  to download files and directories, to reproduce <abbr>pipelines</abbr>, etc.
-  It can be expensive in the large repositories with a lot of projects.
-
-- Not enough isolation/granularity - commands like `dvc metrics diff`, `dvc dag`
-  and others by default dump all the metrics, all the pipelines, etc.
+- DVC commands like `dvc pull`, `dvc checkout`, `dvc metrics diff`, and plenty
+  others analyze the whole repository to find `dvc.yaml` and `.dvc` files to
+  work with. This may be undesirable and it's inefficient for large repositories
```
jorgeorpinel, line 58: I'm not sure this is true. For example `dvc repro` only uses ./dvc.yaml by default and needs `-R .` or `-P` to use everything.

jorgeorpinel: Hm... It's also repetitive with the next paragraph. I'll update it... ⏳

jorgeorpinel: Done.

content/docs/command-reference/init.md
```diff
@@ -102,18 +94,37 @@ project-A
...
+...
+```
+
+> This is a questionable Git practice though, unless employing submodules.
 
-won't download or checkout data for the `data-B.dvc` file.
+In these cases all projects are also isolated from each other and commands run
+in one or the other only affects itself. Inner ones wouldn't search for
+data/pipelines above them anyway, and outer ones know to ignore sub-repos by
+default.
```
jorgeorpinel, line 127: Here's the main NEW text in my init change so far.

jorgeorpinel: But thinking about it maybe I can merge this with the previous section smoehow... ⏳

jorgeorpinel: > ones ignore sub-repos by default

I tested this with `dvc status` but I'm not sure it's 100% true with all commands... Need to test a few more things on this too ⏳

content/docs/command-reference/repro.md
```diff
@@ -130,8 +127,8 @@ only execute the final stage.
 - `-p`, `--pipeline` - reproduce the entire pipelines that the `targets` belong
   to. Use `dvc dag <target>` to show the parent pipeline of a target.
 
-- `-P`, `--all-pipelines` - reproduce all pipelines, for all the stages present
-  in `DVC` repository.
+- `-P`, `--all-pipelines` - reproduce all pipelines for all the `dvc.yaml` files
+  in present the DVC repository (there can be one in every directory).
```
shcheklein, line 131: `(there can be one in every directory).` not sure this is needed - it's clear that there can be only one dvc.yaml per directory

shcheklein, line 131: present the DVC ....  ??

content/docs/command-reference/init.md
```diff
@@ -32,67 +32,59 @@ that are hidden from the user. This directory is automatically staged with
 
 `--subdir` must be provided to initialize DVC in a subdirectory of a Git
 repository. DVC still expects to find the Git repository (will check all
-directories up to the root to find `.git`). This options does not affect any
-config files, `.dvc` directory is created the same way as in the default mode.
-This way multiple DVC projects (including nested ones) could be initialized in a
```
shcheklein, line 37: why are we removing points about nesting, about granularity?

content/docs/command-reference/init.md
```diff
@@ -32,67 +32,59 @@ that are hidden from the user. This directory is automatically staged with
...
 
 #### When is this useful?
 
-`--subdir` is mostly used in the scenario of a
-[monorepo](https://en.wikipedia.org/wiki/Monorepo), but also can be used in
-other workflows when such isolation and/or advanced granularity is needed.
+This option is mostly useful in the scenario of a
+[monorepo](https://en.wikipedia.org/wiki/Monorepo) (Git repository split into
+several project directories), but can also be used with other patterns when such
+isolation is needed. `dvc init --subdir` mitigates the issues of initializing
```
shcheklein, line 45: same here - what is the reason in removing granularity?

content/docs/command-reference/init.md
```diff
@@ -32,67 +32,59 @@ that are hidden from the user. This directory is automatically staged with
...
-  different remote storages, for example, for different sub-projects, etc.
-
-- Not enough isolation/granularity - commands like `dvc pull`, `dvc checkout`,
-  and others analyze the whole repository to look for `dvc.yaml` or `.dvc` files
-  to download files and directories, to reproduce <abbr>pipelines</abbr>, etc.
-  It can be expensive in the large repositories with a lot of projects.
-
-- Not enough isolation/granularity - commands like `dvc metrics diff`, `dvc dag`
-  and others by default dump all the metrics, all the pipelines, etc.
+- Many DVC commands can explore the whole <abbr>DVC repository</abbr> to find
```
shcheklein, line 56: I would prefer to keep it more explicit as it was before. Even if we merge two points.

content/docs/command-reference/init.md
```diff
@@ -32,67 +32,59 @@ that are hidden from the user. This directory is automatically staged with
...
-No matter what mode is used, DVC looks for the `.dvc` directory when it starts
-(from the current working directory and up). Location of the found `.dvc`
-directory determines the root of the DVC project. (In case of `--subdir` it
-might happen that Git repository root is located at different path than the DVC
-project root.)
+The <abbr>project</abbr> root defines the possible scope of action for most DVC
+commands (e.g. `dvc repro`, `dvc pull`, `dvc metrics diff`), meaning that only
+`dvc.yaml`, `dvc.lock`, and `.dvc` files under that location are usable by the
+commands. To determine the root of the project, DVC looks for `.dvc/` from the
+current working directory, up. With `--subdir`, the project root will be found
```
shcheklein, line 66: not sure what Git root has to do with root search

content/docs/command-reference/init.md
```diff
@@ -102,18 +94,37 @@ project-A
...
+    ...
 ```
 
-Nothing changes for the `project-B`. But for any DVC command being run in the
-`project-A` ignores the whole directory `project-B/`, meaning for example:
+Nothing changes for inner `project-B`. But any DVC command being run in outer
+`project-A` ignores the nested directory `project-B/`. For example, `dvc pull`
+(run in `project-A/`) wouldn't download data for the `data-B.dvc` file.
+
+### Nested DVC repositories
```
shcheklein, line 104: I guess confusion comes from the our usage of the DVC project term. DVC project to me is a generic term - it can be Git repo or not. So the whole section above should apply to "DVC repos" (which I would avoid using).

content/docs/command-reference/repro.md
jorgeorpinel: I mean sure, you can't repeat file names. So true, I'm removing this note here (see a5db93fe).

But actually, we don't mention that DVC supports multiple `dvc.yaml` files here, in `run`, or in the DVC files & dirs guide. Should we?

jorgeorpinel: Thanks. Also fixed in a5db93fe.

content/docs/command-reference/init.md
jorgeorpinel: > about nesting

I think "multiple DVC projects can be initialized in a single Git repository" implies "including nested ones". In general I didn't remove the nesting explanation, in fact I clarified it and expanded on it quite a bit.

> about granularity

The doc mentioned "isolated and granular" everywhere. We're really just talking about one thing I think, not 2, and "isolation" seems to me like the more correct concept.

content/docs/command-reference/init.md
jorgeorpinel: Simplifying the language. "isolation" is enough?

content/docs/command-reference/init.md
jorgeorpinel: OK. You mean to add "Not enough isolation" to the beginning of every bullet? It seemed repetitive to me especially since "isolation" is already mentioned several times in preceding paragraphs but as you prefer.

content/docs/command-reference/init.md
jorgeorpinel: This paragraph is about what the project root is, how it's found, how it affects DVC commands, and finally, how --subdir plays into all this:

> With `--subdir`, the project root will be found before the Git repo root, reducing this scope.

Maybe the p is too complicated, I'll split it in 2...

jorgeorpinel: Improved explanation in 40512a11.

content/docs/command-reference/init.md
jorgeorpinel: Yes, a DVC repo is a kind of DVC project — we use these terms consistently (only call it "repo" when the Git context matters). In the monorepo pattern (--subdir), there's sub-projects in a single Git repo, so I don't call those DVC repos (that would be a little confusing). But anyway...

I'm working on merging this with the previous section as mentioned in https://github.com/iterative/dvc.org/pull/1615#discussion_r463272826, I'll get back to you... ⏳

content/docs/command-reference/init.md
jorgeorpinel: > maybe I can merge this with the previous section

Done in 1f77e8fe.

content/docs/command-reference/init.md
jorgeorpinel: Merged in 1f77e8fe. But maybe it should be an Example instead of this long explanation with 2 code blocks?

content/docs/command-reference/init.md
jorgeorpinel: > not sure it's 100% true with all commands

Well, I'll just assume this is the case, at least it's the intention — otherwise it would be a bug anyway.

content/docs/command-reference/repro.md
jorgeorpinel: Extracted to #1641

content/docs/command-reference/init.md
```diff
@@ -32,88 +33,93 @@ that are hidden from the user. This directory is automatically staged with
...
+`dvc checkout`) can access the commit history from the Git root (`.`), when run
+in `project-A`.
 
-On the other hand, if there _are_ nested DVC projects, e.g.:
+> `.` is not a DVC project in this case, so most DVC commands can't be run
+> there.
 
-```
-project-A
+**Advanced structures**: If there are nested projects, either `--subdir`,
```
shcheklein, line 95: I still don't quite understand the motivation behind this tbh? This case is not different from what we already had - we just didn't mention Git in some cases only because it does not matter - projects are isolated git or not ... am I missing something?

jorgeorpinel: Yes, a quick summary is that projects are always isolated. Maybe (I should simplify some of the explanations this way) but users specifically wonder about nested Git repos and submodules (seen it in Discord a few times). That's the main motivation here.

content/docs/command-reference/init.md
shcheklein: hard to tell ... I guess my thought process was something around trying to include both terms since they resonate with different people (easier to search). `isolation` (implies granularity, but not the other way around) is def enough but a bit too technical, while `granularity` can be a bit more problem related. 

anyways, it's a minor thing - I'm fine if we omit this - your call.

content/docs/command-reference/init.md
shcheklein: isolation +/- granularity is minor thing. I guess two paragraphs with specific commands to give some sense, etc.

content/docs/command-reference/init.md
```diff
@@ -32,88 +33,92 @@ that are hidden from the user. This directory is automatically staged with
...
 
-DVC project root defines the scope for most DVC commands. Mostly meaning that
-all `dvc.yaml` and `.dvc` files under the root path are being analyzed.
+With `--subdir`, the project root will be found before the Git root, making sure
+the scope of DVC commands run here is constrained to this project alone, even if
+there are more DVC-related files elsewhere in the repo. Similarly, DVC commands
+run outside this project root will ignore its contents.
 
-If there are multiple DVC sub-projects but they _are not_ nested, e.g.:
+**Simple structures**: multiple `--subdir` projects, not nested, e.g.:
```
shcheklein, line 73: I would not call these simple or advanced. Nesting alone does not make them simple or advanced.

content/docs/command-reference/init.md
```diff
@@ -32,88 +33,92 @@ that are hidden from the user. This directory is automatically staged with
...
-├── .git
-|
+├── .git      # plain Git repo
 ├── project-A
-│   └── .dvc
+│   ├── .dvc    # dvc init --subdir
 │   ...
 ├── project-B
-│   └── .dvc
+│   ├── .dvc    # dvc init --no-scm
```
shcheklein, line 82: two different comments are confusing, `no-scm` here is confusing

content/docs/command-reference/init.md
shcheklein: all these became quite complicated and hard to read with `projects either --subdir or --no-scm` - even for me it's very challenging to wrap my mind around it ... I know that all this subdir stuff hurts the brain- thus it's important to keep it simple, try to not mix, explain essence.

So, for submodules - there no difference at all with regular `dvc init` , right? `--subdir` has nothing to do in this case. DVC commands are isolated (since .dvc defines the scope). If we want to mention this then we can mention we should probably mention it in the description itself? Somewhere where we explain how `dvc init` connects dvc with git.

Same goes for nested Git repos - for DVC (like for git) in this case nothing changes and `--subdir` does not play a role.

content/docs/command-reference/init.md
jorgeorpinel: I see. But I just think "isolation" is a much more common term: https://trends.google.com/trends/explore?date=today%205-y&q=isolation,granularity

shcheklein: They are not the same, so doesn't make sense to my mind to compare them directly. Isolation probably implies granularity, granularity does not necessarily mean isolation.

content/docs/command-reference/fetch.md
```diff
@@ -90,9 +90,9 @@ or `-T` options are used).
 
 - `-j <number>`, `--jobs <number>` - number of threads to run simultaneously to
   handle the downloading of files from the remote. The default value is
-  `4 * cpu_count()`. For SSH remotes, the default is just `4`. Using more jobs
-  may improve the total download speed if a combination of small and large files
-  are being fetched.
+  `4 * cpu_count()`. For SSH remotes, the default is `4`. Using more jobs may
+  improve the total download speed if a combination of small and large files are
```
shcheklein, line 94: no sure about this note - why only combination?

content/docs/command-reference/push.md
```diff
@@ -105,9 +105,9 @@ the target [stage files](/doc/command-reference/run), through the corresponding
 
 - `-j <number>`, `--jobs <number>` - number of threads to run simultaneously to
   handle the uploading of files from the remote. The default value is
-  `4 * cpu_count()`. For SSH remotes, the default is just `4`. Using more jobs
-  may improve the total download speed if a combination of small and large files
-  are being fetched.
+  `4 * cpu_count()`. For SSH remotes, the default is `4`. Using more jobs may
+  improve the total download speed if a combination of small and large files are
```
shcheklein, line 109: same here, note looks strange to me

content/docs/command-reference/pull.md
```diff
@@ -98,9 +98,9 @@ reflinks or hardlinks to put it in the workspace without copying. See
 
 - `-j <number>`, `--jobs <number>` - number of threads to run simultaneously to
   handle the downloading of files from the remote. The default value is
-  `4 * cpu_count()`. For SSH remotes, the default is just `4`. Using more jobs
-  may improve the total download speed if a combination of small and large files
-  are being fetched.
+  `4 * cpu_count()`. For SSH remotes, the default is `4`. Using more jobs may
+  improve the total download speed if a combination of small and large files are
```
shcheklein, line 102: same here

content/docs/command-reference/init.md
jorgeorpinel: I just don't see what granularity is provided by --subdir. TBH I don't even know what we meant by "granular" here.

image

Definitely isolation, though.

content/docs/command-reference/init.md
jorgeorpinel: OK, added back a couple specific command examples in 221ed757.

content/docs/command-reference/init.md
jorgeorpinel: These 2 block examples (one for not nested and one for nested) were already there (see https://dvc.org/doc/command-reference/init#how-does-it-affect-dvc-commands). I was just expanding the nested one to include other possibilities, but I see how that can be confusing...

> for submodules - there no difference at all with regular dvc init , right? --subdir has nothing to do in this case... we can mention we should probably mention it in the description itself

Correct. OK, I changed it this way in d14e960e.

content/docs/command-reference/init.md
jorgeorpinel: Renamed to "Nested".

content/docs/command-reference/init.md
jorgeorpinel: Removed.

content/docs/command-reference/fetch.md
jorgeorpinel: I didn't cange this note, only removed the word "just" for consistency with a corresponding change you made in `status`.

content/docs/command-reference/push.md
jorgeorpinel: I only removed the word "just" here.

content/docs/command-reference/pull.md
jorgeorpinel: Same

content/docs/command-reference/init.md
shcheklein: simple example:

granularity - you split project in a way that `dvc pull` becomes _granular_ (on a project level) because of _isolation_

content/docs/command-reference/init.md
jorgeorpinel: OK, removed both "Simple" and "Nested" now (in 5b47bd4f).

content/docs/command-reference/fetch.md
shcheklein: I see, yes, let's then remove it as well. (`just` is already removed btw)

jorgeorpinel: > Using more jobs may improve the total download speed if a combination of small and large files are being fetched.

I think this was written by the engineering team a while back. (maybe by @pared ?) So I just accepted it as truth.

jorgeorpinel: But anyway. Removed the note and standardized this option desc. for all refs. in 80a0f092.

content/docs/command-reference/push.md
jorgeorpinel: Standardized in 80a0f092.

content/docs/command-reference/pull.md
jorgeorpinel: Same (80a0f092)

content/docs/command-reference/fetch.md
shcheklein: thanks! we can probably keep generic explanation like "Using more jobs may improve the total download speed."

there are a lot of situations when that might happen (high latency connection, qos per connection on both ends). Curios about large and small files combination though- may be I'm missing something.

content/docs/command-reference/init.md
```diff
@@ -14,106 +14,103 @@ DVC works best in a Git repository. This enables all features, providing the
...
+The command [options](#options) can be used to start an alternative workflow for
+advanced scenarios:
+
+- [Initializing DVC in subdirectories](#initializing-dvc-in-subdirectories)
+  (`--subdir`) - for monorepos and nested <abbr>DVC projects</abbr>
+- [Initializing DVC without Git](#initializing-dvc-without-git) (`--no-scm`) -
+  for very simple projects, SCM other than Git, deployment automation, among
+  other uses
+
+> Note that <abbr>DVC repositories</abbr> nested inside other DVC repos (for
```
shcheklein, line 32: very hard to understand ... mostly because we use DVC repository term here probably.

shcheklein: and probably because it comes almost at the very beginning - a bit unexpected

content/docs/command-reference/init.md
```diff
@@ -14,106 +14,103 @@ DVC works best in a Git repository. This enables all features, providing the
...
-  different remote storages, for example, for different sub-projects, etc.
-
-- Not enough isolation/granularity - commands like `dvc pull`, `dvc checkout`,
-  and others analyze the whole repository to look for `dvc.yaml` or `.dvc` files
-  to download files and directories, to reproduce <abbr>pipelines</abbr>, etc.
-  It can be expensive in the large repositories with a lot of projects.
-
-- Not enough isolation/granularity - commands like `dvc metrics diff`, `dvc dag`
-  and others by default dump all the metrics, all the pipelines, etc.
+- By default, DVC commands like `dvc checkout` and `dvc repro` explore the whole
```
shcheklein, line 61: still, case with not isolated/granular `dvc metrics diff` seems a bit different than `dvc pull` - one is mostly about performance, second is about actual output - might resonate with different people.

It's fine though to keep it this way. Not a strong one.

content/docs/command-reference/init.md
```diff
@@ -14,106 +14,103 @@ DVC works best in a Git repository. This enables all features, providing the
...
 
-DVC considers them a two separate DVC projects. Any DVC command that is being
-run in the `project-A` is not aware about DVC `project-B`. DVC does not consider
-Git repository root an initialized DVC project in this case and commands that
-require DVC project will raise an error.
-
-On the other hand, if there _are_ nested DVC projects, e.g.:
+DVC considers A and B separate projects. Any DVC command run in `project-A` is
+not aware of `project-B`. However, commands that involve versioning (like
+`dvc checkout`) access the commit history from the Git root (`.`).
```
shcheklein, line 92: `dvc checkout` does not require GIt

content/docs/command-reference/init.md
```diff
@@ -14,106 +14,103 @@ DVC works best in a Git repository. This enables all features, providing the
...
+> there.
 
-Nothing changes for the `project-B`. But for any DVC command being run in the
-`project-A` ignores the whole directory `project-B/`, meaning for example:
+If there are nested `--subdir`projects e.g.:
 
 ```dvc
-$ cd project-A
-$ dvc pull
+project-A       # full DVC + Git repo
```
shcheklein, line 100: why do we need this comment?

content/docs/command-reference/init.md
```diff
@@ -130,15 +127,19 @@ include:
...
 
 DVC sets the `core.no_scm` config option value to `true` in the DVC
-[config](/doc/command-reference/config) when it is initialized this way. It
-means that even if the project was Git-tracked already or Git is initialized in
-it later, DVC keeps operating in the detached from Git mode.
+[config](/doc/command-reference/config) when initialized this way. This means
+that even if the project is tracked by Git, or if Git is initialized in it
+later, DVC will keep operating detached from Git in this project.
+
+> Note that like with nested <abbr>repositories</abbr> and `--subdir` projects,
```
shcheklein, line 140: so, if behavior is the same, do we actually need this note?

content/docs/command-reference/fetch.md
jorgeorpinel: Added "Using more jobs may improve the overall transfer speed." to all for now 🙂 

> Curios about large and small files combination

I think it means one big file is downloaded in a thread while many small ones are downloaded in the other ones so somehow there's no bottle neck... But I'm also a bit skeptical of this logic now that you mention it.

jorgeorpinel:
OMG sorry I linked the other PR to this one 😋 Reopening...

content/docs/command-reference/init.md
jorgeorpinel: True, it's not the best example. Changed to `diff` now.

content/docs/command-reference/init.md
jorgeorpinel: Hmmm they are informative. Changed to `git init && dvc init` in cac92e58. WDYT?

shcheklein: so, that's probably what I don't understand - if we put them they highlight importance of something to my mind, I spend a cycle thinking why did author put a comment here - is .git important in this case? does it change anything?

content/docs/command-reference/init.md
jorgeorpinel: Changed a little but I'm not sure how to make it super clear. It's just a minor note very specific for people looking for that specific use case (DVC + Git submodules). Don't you think they will get it?

jorgeorpinel: > because it comes almost at the very beginning

It's at the end of the main **Description** section. Should I move it to after all the other sub-headings? It would look as it pertains to the --no-scm explanation...

shcheklein: it's fine let's keep it


> Don't you think they will get it?

I would expect from people who use submodules have the same intuition by default - isolated in Git, isolated in DVC - nothing special or specif. unless I'm missing something. That's why I'm a bit hesitant about all this changes and notes. It's hard to cover every case with a note.

content/docs/command-reference/init.md
jorgeorpinel: Agree. I changed the last sentence to "This can produce undesirable results and/or be inefficient for large monorepos." to better capture both cases.

content/docs/command-reference/init.md
jorgeorpinel: They do intent to highlight exactly what we mean by nested or not-nested --subdir projects:

Not nested: single-level DVC projects in a Git repo.
Nested: multi-level DVC projects inside (DVC projects inside) DVC/Git repo

This could be also just undetstood by looking at the directories shown (.git, .dvc) but it's not as obvious).

content/docs/command-reference/init.md
shcheklein: yep, but this is a good example of what I meant with implicit vs explicit. `undesirable results` sounds and reads cool, and one paragraph looks better for sure, but it almost does not mean anything to a reader .. it's hard (or at least takes an effort) to get an example of what does it mean.

again, not a big deal in this case, just want to further explain my thought-process on being explicit if possible.

content/docs/command-reference/init.md
shcheklein: .git still not important here to my mind unless I'm missing something. Nesting would work with `--no-scm` the same way effectively.

content/docs/command-reference/init.md
jorgeorpinel: - [ ] Maybe not. This isolation and nested situation is too complicated... I should probably make a separate sub-section to cover it after explaining all the 3 kinds of projects (typical or git-enabled, --subdir, --no-scm).

Can we extract to another issue/PR and get this merged maybe? I have some other changes I also want to throw in on another regular updates PR 😬

content/docs/command-reference/init.md
jorgeorpinel: True, I can just remove it and handle them via support channel if needed. So far there was only one case that I noticed.

content/docs/command-reference/init.md
shcheklein: Let's remove it then, create a ticket and discuss the changes in advance?

content/docs/command-reference/init.md
jorgeorpinel: I see. OK, split into 2 bullets for clarity in be5bc6dc.

content/docs/command-reference/fetch.md
pared: @jorgeorpinel That might have been me, though I do not remember this one. Anyway, I could agree that it might get faster for any combination of files. Using multiple jobs parallelizes tasks such as opening file descriptors, so even in case of limited connection speed, it might help (to some extent). Anyway, I don't think its something user might want to spend time on, and such inclusion hinders understanding this option.

jorgeorpinel: Thanks for the comment Pawel!

content/docs/command-reference/init.md
jorgeorpinel: I'm removing these block examples completely for now, and extracting to a following PR...

content/docs/command-reference/init.md
jorgeorpinel: Removed. I have all the text already in another branch though so I'll just open a draft PR to discuss, if you don't mind... ⌛

shcheklein:
I've approved it already a bit ago. Merging ...

jorgeorpinel:
Thanks

PR #1619, opened 2020-07-23
GitHub Actions link checker

### LINKED ISSUE #1000: scripts: link check improvements

rogermparent:
# GitHub Actions Link Checker

This PR introduces a big feature overhaul: A completely new JS-based link checker that runs in both GitHub Actions and the CLI!

You can find the new app's repo at https://github.com/iterative/link-check. It has a submodule for distributing the lambda-esque compiled action, but the source is all in the same repo.

This link checker primarily aims to solve two issues:

- Root-relative links are checked against the deploy preview, not master (new links to files added in the same commit don't fail)
- Initial request uses HEAD, but falls back to GET on 405 (affects towardsdatascience.com links, which are removed from the exclude list now)

These two features, if I understand correctly, make this PR fix #1000 barring any regressions

## Output examples

- [Failing deploy check with various common reasons](https://github.com/iterative/dvc.org/runs/1074713822?check_suite_focus=true)

- [Successful deploy check](https://github.com/iterative/dvc.org/pull/1619/checks?check_run_id=1076497631)

- [Sitewide check reporting unused link patterns](https://github.com/iterative/dvc.org/pull/1619/checks?check_run_id=1076506044)

- [Full check with all patterns used](https://github.com/iterative/dvc.org/pull/1619/checks?check_run_id=1074810465)


rogermparent:
Seems like PR workflows do show up. Nice! No need for prod testing.

shcheklein:
I love the transition from bash!

qq:

Does it supersede the the bash scripts completely? it's not good to support both - we should completely get rid of bash and use JS

rogermparent:
> I love the transition from bash!

Thanks!

> Does it supersede the the bash scripts completely? it's not good to support both - we should completely get rid of bash and use JS

Yup! I left the bash ones in there in case we wanted to ensure all the matches are the same, but they can be safely removed from the repo. Afterward, the link check bash script will have to be removed from CircleCI.

shcheklein:
@rogermparent 

> Yup! I left the bash ones in there in case we wanted to ensure all the matches are the same, but they can be safely removed from the repo. Afterward, the link check bash script will have to be removed from CircleCI.

how about some easy way to run it locally/on a pre-commit hook?

I'd say let's start this transition. And let's transfer the action repo to the iterative account?

rogermparent:
> how about some easy way to run it locally/on a pre-commit hook?

Sure! I made a CLI runner for testing during development, so I can just expand that into something end-users and/or a pre-commit hook can use.

> I'd say let's start this transition. And let's transfer the action repo to the iterative account?

Sounds good to me!

rogermparent:
Alright, at this point I've made the plugin work on dvc.org and now generalized it enough to run on any website (with some opinions that can be loosened later, like exclusion filters)

The report output is better now as well, check it out in the updated report examples in the OP!

shcheklein:
@rogermparent amazing stuff! A few initial questions:

- does it cover the full run that we do on cron everyday (vs diff)?
- does it check for links in the exclusion list that do not exist in the files anymore?

package.json
```diff
@@ -16,9 +16,7 @@
     "format": "prettier --write",
     "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "link-check": "./scripts/link-check-git-all.sh",
-    "link-check-diff": "./scripts/link-check-git-diff.sh",
-    "link-check-exclude": "./scripts/exclude-links-check.sh"
+    "link-check": "LINK_CHECK_EXCLUSION_FILE='./scripts/exclude-links.txt' LINK_CHECK_FILE_FILTER='content/*' LINK_CHECK_ROOT_URL='https://dvc.org' yarn run link-check-git-diff"
```
shcheklein, line 19: should we use localhost for the ROOT URL here?

package.json
```diff
@@ -16,9 +16,7 @@
     "format": "prettier --write",
     "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "link-check": "./scripts/link-check-git-all.sh",
```
shcheklein, line 19: do we account for all three commands with this change? do we need all of them?

package.json
```diff
@@ -157,7 +157,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn format-staged && yarn lint-staged"
+      "pre-commit": "yarn format-staged && yarn lint-staged && yarn run link-check"
```
shcheklein, line 160: if we don't use localhost for the base url it will fail, right?

shcheklein: also means that it fails if local server is not running ... probably too much complication to run it on pre-commit

shcheklein:
Some questions on the action code:

- do we use linters, prettier, pre-commit to run them all, CI to check, etc? Let's use the same style as we do in other projects?
- I would say too much split - sometimes one function per file. I like modularity, but it seems excessive. Don't forget about the overhead it brings (switching among too many files).

- in the https://github.com/iterative/link-check-git-diff/blob/master/exclusions.js#L12
```
? (subject) => !micromatch.isMatch(subject, filters, { bash: true })
    : (subject) => micromatch.isMatch(subject, filters, { bash: true });
```

can it be deduplicated?

- we should be using `.github/ content/ src/`  and `*.css *.js *.jsx *.md *.tsx *.ts *.json` extenstions.
- why don't we use `git diff` the same way as we do in bash to get file names and then read them - look simpler and more robust then parsing diff.

rogermparent:
>     * does it cover the full run that we do on cron everyday (vs diff)?

I must have glossed over that requirement, but I can pretty easily add a full repo run feature and rig it up for the same things.

>     * does it check for links in the exclusion list that do not exist in the files anymore?

This would be a sub-feature of a full run, but regardless completely possible to do.

>     * do we use linters, prettier, pre-commit to run them all, CI to check, etc? Let's use the same style as we do in other projects?

Not yet, but probably a good idea. I'll probably try to put something together to make it easier for me to add this without forgetting it every time- even if it just ends up being an MD checklist and not some sort of metapackage.

>     * I would say too much split - sometimes one function per file. I like modularity, but it seems excessive. Don't forget about the overhead it brings (switching among too many files).

I personally have the opposite inclination and find large files much harder to browse, which is why I keep tending toward that naturally, but I can take another look through and condense the files to as few as is needed.

>     * we should be using `.github/ content/ src/`  and `*.css *.js *.jsx *.md *.tsx *.ts *.json` extenstions.

Sure! That sounds easy enough.

>     * why don't we use `git diff` the same way as we do in bash to get file names and then read them - look simpler and more robust then parsing diff.

The bash diff script didn't work like that, it used `git diff` to get a list of files and then `git diff`ed those files individually with a hash retrieved by `git merge-base`; even if it did work like that, I heavily disagree that that kind of change will bring any improvements.  
Using one `git diff` call for everything is the simplest and most robust option for the diff feature, as it means that only one call to bash `git` needs to be made *and* only links that are added are tested as opposed to situations like re-testing every link in one file. As far as I researched, no other method can boast these two features, nor another benefit that outweighs them.

That said, I will need to be able to read whole files like that to add the "whole project check". Thankfully, I made the codebase modular enough that I just need to make a new entry point that feeds different data into the same link scraping and link checking tools and this addition shouldn't take very long.

package.json
rogermparent: That would only be if we could guarantee the server was up during every run.

Though a separate invocation that does that would be a great feature for offline testing!

package.json
rogermparent: The CLI script works completely fine on pre-commit with no local server, with the exception that new root-relative links that don't exist will be reported as nonexistent. 
To mitigate root-relative issues, I left the pre-commit as just an informative check that doesn't block commits. This way, editors can see ahead of time if a link fails, but can also continue if it's just root-relative links that will be fulfilled on the preview with the real GH Actions check.

The ability to optionally run this against the dev server would be good though, so I'll keep that in mind.

shcheklein:
> I personally have the opposite inclination and find large files much harder to browse, which is why I keep tending toward that naturally, but I can take another look through and condense the files to as few as is needed.

Def not goal to minimize number of files :) Your call on this.

> Using one git diff call for everything is the simplest and most robust option for the diff feature

there can be potentially an overhead of parsing a huge diff if it includes. Again, no strong opinion. Up to you on this. The logic with regex looked hard to read and fragile that's why I mentioned this.

package.json
rogermparent: I didn't see where this was used and thought it was legacy, but I'll replicate all the old functionality now that I know!

rogermparent:
> there can be potentially an overhead of parsing a huge diff if it includes. Again, no strong opinion. Up to you on this. The logic with regex looked hard to read and fragile that's why I mentioned this.

Working with `git` programmatically tends to lead to fragile-looking code no matter how it's implemented because there isn't a good way to get a more "computer-native" output like JSON without intermediate parsing like regex.

.github/workflows/link-check.yml
```diff
@@ -13,11 +13,11 @@ jobs:
 
     - name: Run the link checker
       id: check
-      uses: "iterative/link-check-git-diff.action@master"
+      uses: "iterative/link-check.action@master"
```
shcheklein, line 16: love the new name!

content/blog/2020-08-10-august-20-dvc-heartbeat.md
```diff
@@ -20,6 +20,12 @@ tags:
   - Meetup
 ---
 
+[Google](https://www.google.com)
```
shcheklein, line 23: this is for testing?

package.json
```diff
@@ -16,9 +16,7 @@
     "format": "prettier --write",
     "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "link-check": "./scripts/link-check-git-all.sh",
-    "link-check-diff": "./scripts/link-check-git-diff.sh",
-    "link-check-exclude": "./scripts/exclude-links-check.sh"
+    "link-check": "repo-link-check --rootURL='https://www.dvc.org' --fileIncludePattern='**' --linkExcludeFile='scripts/exclude-links.txt' --always-exit-zero"
```
shcheklein, line 19: a few questions:

1. why do we call it `repo-link-check` not just "link-check"?
2 does it include "link-check-exclude"?
3. --fileIncludePattern='**' - does it literally mean all files? no matter what extension is?

shcheklein:
@rogermparent hey, Roger! What is the status of this? Anything else left?

Also, actions look great - we should add some readme to them, tags, how to use and publish them properly for good - so that other people could use them, wdyt?

rogermparent:
> @rogermparent hey, Roger! What is the status of this? Anything else left?
> Also, actions look great - we should add some readme to them, tags, how to use and publish them properly for good - so that other people could use them, wdyt?

Unless an issue pops up in usage, the code should be good to go! All that's left is the README and integrating it everywhere it needs to go on the DVC site.
Making this usable by others is definitely on the roadmap, as the bulk of the work I've done is to that end. I was changing the API a lot up until now, but I think I have it where I want it.

package.json
rogermparent: I named the package's binary `repo-link-check` because it's the name of the npm package, as `link-check` is already taken by another library.

I meant to come back here to get the exact file patterns to use, `**` was a testing stopgap that does mean all files.

content/blog/2020-08-10-august-20-dvc-heartbeat.md
rogermparent: yep! I'll reset to master's content before merge.

.github/workflows/link-check.yml
rogermparent: Thanks! With the new filesystem feature, I'm more comfortable naming it something more generic.

package.json
rogermparent: Now that I think about it, I can still name the binary 'link-check', as the binary names aren't coupled with package names. I just wanted to avoid confusion for if a user installs `repo-link-check` from NPM and has to run `link-check` as a binary.

shcheklein: up to you Roger, I was curious

shcheklein:
@rogermparent okay 👌  let me know when it's ready to merge

content/blog/2020-08-10-august-20-dvc-heartbeat.md
shcheklein: 👍

package.json
shcheklein: does it include the `link-check-exclude` as well?

rogermparent: Do you mean the `exclude-links.txt` file? If so, yes! Excluded links are displayed in the report as always passing with a description of "Excluded". I made sure to re-test that just recently to ensure that behavior still works.
If you mean something else, I must not have been aware of it.

shcheklein: I mean a test that check the `exclude-links.txt` file itself, that for each entry we have a match in the repo.

rogermparent: Ah, no, I don't currently have that. Since all exclusion patterns are kept in an array, we could track to make sure all are accessed by marking them in the script as the program uses them and then checking if any weren't touched by the end to report them.

rogermparent: This feature is in now! Using the CLI with the "--report-unused-patterns" will have the script output any link exclusion patterns that it didn't use. I added this flag to the `link-check-all` invocation, and the results can be seen at the top of the script's output, though when doing so I would recommend piping the output into less or a text file because a full repo report is enough to push the unused patterns out of scroll range on most terminal emulators.

rogermparent:
I adapted all the scripts and was ready to merge, but CircleCI just... can't find the node scripts I guess. I certainly didn't expect that.

```
#!/bin/bash -eo pipefail
yarn link-check-exclude

yarn run v1.22.4
$ repo-link-check -r='https://www.dvc.org' --fi='{.github,content,src}/**/*.{css,js,jsx,md,tsx,ts,json}' --fef='./scripts/exclude-files.txt' --lef='./scripts/exclude-links.txt' -s=filesystem -u='only'
/bin/sh: 1: repo-link-check: not found
```

This looks like the final barrier to merging, but I'm at a loss about how to fix it.

rogermparent:
Sorry for the wait, it's been a *while*, but I'm finally ready to say this is ready for review!

I'm not sure how we should change the two remaining dead links that the site-wide check picked up. One is for a Spotify podcast link to "The Data Stream" that's since been removed (though they're still up on ITunes)  
The other is to https://conf.mec.dev/2019/sprint/, where the site is completely down- the event's site seems to have moved to https://hasgeek.com/mec-dev/conf/, but I'm not sure if that's actually equivalent to the given link.

.github/workflows/link-check-all.yml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Run the link check script
+      id: check
+      uses: "iterative/link-check.action@v0.3.0"
+      with:
+        source: "filesystem"
+        configFile: "./.linkcheckrc.json"
+
+    - uses: LouisBrunner/checks-action@v0.1.0
```
shcheklein, line 23: curious - do we need this for the cron-style check? what would it give us?

.github/workflows/link-check-exclude.yml
```diff
@@ -0,0 +1,27 @@
+name: Link Check Exclude
+on: deployment
```
shcheklein, line 2: hmm ... why on deployment?

.github/workflows/link-check-all.yml
rogermparent: Good point that this invocation doesn't necessarily need to be a GitHub Check, but currently we need it as the GHA script always makes output for `checks-action`.
That said, it'd be pretty simple for me to add an option to the Action inputs for using the console instead of GHA outputs for runs that don't need to be a Check if that's desirable.

.github/workflows/link-check-exclude.yml
rogermparent: That's where it was in the CircleCI config, unless I misread it

.github/workflows/link-check-exclude.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - name: Run the link check script
+      id: check
+      uses: "iterative/link-check.action@v0.3.0"
+      with:
+        source: "filesystem"
```
shcheklein, line 17: tbh, hard to tell how is it different from the other two checks ... `reportUnusedPatterns` - does it mean that it checks only exclude links? or everything else + exclude links?

Can we run exclude links checks as part of the regular cron check? It should be enough I guess if it's not very simple and obvious to make it part of the PR check

.github/workflows/link-check-exclude.yml
rogermparent: Here's the bit I'm referring to:

```
jobs:
  test:
    <<: *defaults
    steps:
      - install
      - run: yarn test
      - run: yarn lint-css
      - run: yarn lint-ts
      - run: yarn format-check
      - run: yarn link-check-diff
      - run: yarn link-check-exclude

  test_full:
    <<: *defaults
    steps:
      - install
      - run: yarn test
      - run: yarn lint-css
      - run: yarn lint-ts
      - run: yarn format-check
      - run: yarn link-check
      - run: yarn link-check-exclude

workflows:
  version: 2

  commit:
    jobs:
      - test
```

`test` includes `link-check-exclude`, and `test` is run on `commit`. If you're implying that it would make more sense to run this as part of the daily check, I agree and can make that change.

scripts/exclude-links.txt
```diff
@@ -1,41 +1,16 @@
-http://127.0.0.1:10000/devstoreaccount1;
-http://localhost:8000/
-http://millionsongdataset.com/pages/getting-dataset/
-https://$
-http://s3-external-1.amazonaws.com/bucket/path
-https://accounts.google.com/o/oauth2/auth
-https://api.cloudflare.com/client/v4/zones/$
-https://circleci.com/gh/iterative/dvc.org
-https://discuss.$
-https://drive.google.com/drive/folders/0AIac4JZqHhKmUk9PDA
```
shcheklein, line 10: did we change rules a bit?

scripts/exclude-files.txt
```diff
@@ -0,0 +1,4 @@
+src/consts.js
```
shcheklein, line 1: should we move these two files out of scripts now? Not sure where though .... don't want to put too many files into root

scripts/exclude-links.txt
rogermparent: Yeah, the scraper on the new checker specifically inspects `[]()`-style links on markdown files, and doesn't go through code blocks like the bash one did. I checked, and most of these exclusions are from code blocks- see the ones ending in `$`: those are string interpolations.

package.json
```diff
@@ -16,9 +16,10 @@
     "format": "prettier --write",
     "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "link-check": "./scripts/link-check-git-all.sh",
-    "link-check-diff": "./scripts/link-check-git-diff.sh",
-    "link-check-exclude": "./scripts/exclude-links-check.sh"
+    "link-check": "repo-link-check -c='./.linkcheckrc.json' -s='filesystem' -u",
```
shcheklein, line 19: hmm ... why `=` though? it's not a very common convention for CLI, just `-c ./.linkcheckrc.json` should be enough

scripts/exclude-files.txt
rogermparent: Good idea, maybe something like `config/link-check`?

content/blog/2020-07-16-devops-for-data-scientists.md
```diff
@@ -127,7 +127,7 @@ Git refresher?
 When a feature branch passes the automated tests, it becomes a candidate to be
 merged into the master branch.
 
-![](/static/uploads/2020-07-16/basic_ci_system.png) _Here's what continuous
+![](/uploads/images/2020-07-16/basic_ci_system.png) _Here's what continuous
```
shcheklein, line 130: hmm ... why is it working w/o this change?

package.json
rogermparent: That's just how I happened to write it, I can make that change and I'm sure `minimist` can handle it.

shcheklein: yeah, I would prefer to have it simpler if possible. thanks!

scripts/exclude-files.txt
shcheklein: yep! sounds good!

content/blog/2020-07-16-devops-for-data-scientists.md
rogermparent: This one confused me as well, I think something in the Gatsby pipeline that I'm not familiar with automatically corrected this.
The `/static/uploads/` version of the path doesn't map to a real directory on the filesystem nor on the site, but the images do get picked up by `gatsby-image` without the change.

scripts/exclude-links.txt
shcheklein: sounds good!

.github/workflows/link-check-exclude.yml
rogermparent: Basically, this was an adaptation to match the circleCI config and check exclusions on commit. 

`reportUnusedPatterns` is only useful when done on a full check, so I made the `reportUnusedPatterns=only` flag to enable `dryRun`, letting the script run on the whole site to check exclusions without actually pinging every link in the site.

So there's two checks on commit: A full-site sweep that checks exclusions without actually sending requests and a diff-only check that actually sends requests and checks if the links are alive.

As we discussed around the other comments, it seems we agree it'd be better to just do this on the daily check.

rogermparent:
We've hit another sweet spot! I've addressed the prior issues and added some extras that were asked for in the original issue.

## New Workflows

The DVC configuration for Link Check now has two different run modes:

### Link Check Deploy

After every successful deploy, Link Check will be run on the git diff between the deployment's branch and `master`, effectively checking all added or modified links.

This workflow currently needs two files because of a quirk in the way the link check action works- the `deployment_status` trigger we use to wait until build completion doesn't "attach" its Checks to the deployment like the `deployment` trigger, so we have to initialize the check on `deployment` and then update it with the result on the `deployment_status` the indicates success.

This approach has some advantages in that the check run is shown as the deployment starts instead of just appearing, but there is an annoying cosmetic drawback in that the previous behavior is mandatory *and* the check initializer shows up as its own entry in the Check list.
The only ways around the consequences of this workaround are either to fix the Check Action such that the Check creation call is able to point to the deployment from a `deployment_status` trigger, or to drop the Check Action and directly use the GitHub Checks REST API, which would entail even more dev work.

### Link Check All

Once per day, as well as on manual request via `workflow_dispatch`, Link Check will be run on the entire repo from GitHub Actions in Console Log output mode.

Also, the `.txt` config syntax was removed in favor of YAML or JSON. While JSON is built-in and sufficient, YAML allows for some extra readability and features that were asked for in #1000.

## YAML config files

DVC uses YAML, GitHub Actions uses YAML, a ton of things use YAML, so it's natural that Link Check should as well. We maintain support for JSON config as well, but drop the raw "file lines" approach in favor of top-level JSON/YAML.

For example:
```
excluded.link
**.excluded.**
```
becomes
```
- 'excluded.link'
- '**.excluded.**'
```

I believe this is more readable, but more importantly YAML allows for comments- a feature asked for in #1000.

Unquoted strings can be used, except that patterns that start with `*` will be interpreted as YAML pointers and this must be quoted so quoting everything is the only way to be consistent.

## Outputs refactor

All output strategies, like `console.log` and the GitHub Checks action, are now able to be modularly switched out and combined. There are currently three:

- `consoleLog` outputs the link check report via `console.log`
- `checksAction` uses the GitHub Actions `core.setOutput` API to generate output in the shape taken by `LouisBrunner/checks-action`
- `exitCode` ends the process with an exit code of 2 if any link checks failed. It's separate from `consoleLog` in case you want to always exit with 0 in favor of other output strategies.

The full check uses `[consoleLog, exitCode]` and the deploy check uses `checksAction`.

shcheklein:
thanks @rogermparent , I'll be reviewing it today-tomorrow

shcheklein:
@rogermparent could you please clarify, show screenshots re the `Link Check Deploy` explanation- it's hard to get this w/o going deep into details?

config/link-check/excluded-links.yml
```diff
@@ -0,0 +1,17 @@
+- 'http://localhost:8000**'
+- 'http://millionsongdataset.com/pages/getting-dataset/#subset'
+- 'https://marketplace.visualstudio.com/items?itemName=stkb.rewrap'
+- 'https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml'
+- 'https://remote.dvc.org/get-started'
+- 'https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&amp;id=24c0ecc49a'
+- 'https://www.meetup.com/San-Francisco-Machine-Learning-Meetup/events/264846847/'
+- 'https://www.reddit.com/r/MachineLearning/comments/bx0apm/d_how_do_you_manage_your_machine_learning/'
+- 'https://www.amazon.com/DevOps-Handbook-World-Class-Reliability-Organizations-ebook/dp/B01M9ASFQ3'
+- '**linkedin.com/in/**'
```
shcheklein, line 10: we were testing them before? (had a custom handler as far as I remember). I would keep it if it was able detecting actual 404 from some "I'm busy" error code LI sends back

config/link-check/excluded-files.yml
```diff
@@ -0,0 +1,4 @@
+- 'src/consts.js'
```
shcheklein, line 1: should we make part of the config for simplicity?

shcheklein: though, I think it makes sense to keep it separate, since this list might become large

config/link-check/config.yml
```diff
@@ -0,0 +1,12 @@
+rootURL: https://dvc.org
+fileIncludePatterns: '{.github,content,src}/**/*.{css,js,jsx,md,tsx,ts,json}'
+fileExcludePatternFiles: config/link-check/excluded-files.yml
+linkExcludePatternFiles: config/link-check/excluded-links.yml
+bottlenecks:
```
shcheklein, line 5: let's call it custom link handler or something, bottleneck sounds a bit strange. Also, it might happen we'll want to define some additional params here per link type/domain

config/link-check/excluded-links.yml
rogermparent: The reason I ignored them is that LinkedIn always returns status code 999 specifically for profiles specifically, both with GET and HEAD and whether the profile exists or not.

shcheklein: got it! that's unfortunate

config/link-check/excluded-files.yml
rogermparent: I'd say it's pretty safe to put the file patterns in the main config if needed, as with the pattern match feature on both whitelists and blacklists we have quite a bit of control over the files checked with only a few entries.

That said, keeping them separate is at least a bit more future-proof and keeps consistent with the link filters.

.github/workflows/link-check-all.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - name: Run Link Check
+      uses: "iterative/link-check.action@master"
+      with:
+        source: "filesystem"
```
shcheklein, line 19: I would recommend to make `filesystem` the default mode and instead `git diff` can become a special option. This is way more intuitive. 

Anyways, source + filesystem - names are very confusing. It's hard to understand that this simply means that it checks all the files.

package.json
```diff
@@ -16,9 +16,10 @@
     "format": "prettier --write",
     "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "link-check": "./scripts/link-check-git-all.sh",
-    "link-check-diff": "./scripts/link-check-git-diff.sh",
-    "link-check-exclude": "./scripts/exclude-links-check.sh"
+    "link-check": "repo-link-check -c config/link-check/config.yml -s filesystem -u",
```
shcheklein, line 19: reminder: probably need to check and fix the docs contrib guide if it mentions links check

shcheklein, line 19: what does `-u` mean?

.github/workflows/link-check-all.yml
```diff
@@ -0,0 +1,23 @@
...
+        configFile: "config/link-check/config.yml"
+        reportUnusedPatterns: true
+        output: >-
+          ["consoleLog", "exitCode"]
```
shcheklein, line 23: we mix two different abstractions here to my mind- actual output + exit code. It might be quite confusing for users.

shcheklein: I would recommend to keep exit code setting a separate explicit option

shcheklein: something similar to linters - failOnError: true/false? we might even include it into both to make it extra explicit

rogermparent:
> @rogermparent could you please clarify, show screenshots re the `Link Check Deploy` explanation- it's hard to get this w/o going deep into details?

Fair point, a screenshot will certainly help

deploy_explanation

You can see in the deployment's Checks that Link Check make two Check entries: Initialize and Check.

The "Initialize" workflow simply creates the Check seen below it in a "queued" state, and more importantly attaches it to the deployment in question which currently cannot be done on the `deployment_status` trigger that we must use to run the action when Heroku is finished as opposed to when the build is first requested. Actions run on the `deployment` trigger are automatically added to the Check list, otherwise I would hide the `Initialize` check, and because of this, we get the extra entry that shows it has started.

The reason this works is that the Checks Action has a sort of between-run cache, and this is their documented solution to this kind of problem which happens to mesh uncleanly with how we're using it.

Without the Initialize workflow, the Check is never attached to the deployment and doesn't show up in the Check Run. This is more a limitation of Checks Action than Link Check itself.

.github/workflows/link-check-deploy.yml
```diff
@@ -0,0 +1,48 @@
+name: Link Check Deploy
```
shcheklein, line 1: I wonder if the whole workflow can be simplified- looks a bit too verbose

.github/workflows/link-check-all.yml
rogermparent: Sure, we can just treat the exit code as a special output controlled by a flag. It's already hard-coded to always run last, so it needs special treatment one way or another.

.github/workflows/link-check-all.yml
rogermparent: Fair point, and it matches with the previous usage. This is a remnant of the project originally only having git diff functionality.

.github/workflows/link-check-deploy.yml
rogermparent: I could use YAML pointers to DRY things out, and the first item isn't strictly necessary.

shcheklein:
A few comments:

1. check names are confusing (see how they are different from everything else, codeclimate, etc). I think it should be simple like `link-check (initialize)`
2. it looks like I need to read about checks action that we use to understand this complexity ...

package.json
rogermparent: `-u` is shorthand for `--report-unused-patterns`, [Here's a link to the README section that lists the CLI flags with shorthands](https://github.com/iterative/link-check#link-check-option-flags)

shcheklein: let's use explicit long name?

also, `-u only` is very confusing- I don't quite understand what does it do from the docs

.github/workflows/link-check-deploy-init.yml
```diff
@@ -0,0 +1,16 @@
+name: Link Check Deploy
+on: deployment
+
+jobs:
+  run:
+    name: Initialize
+    runs-on: ubuntu-latest
+    steps:
+    - uses: LouisBrunner/checks-action@v0.1.0
```
shcheklein, line 9: btw, looks like 1.0 has been release a few hours ago

shcheklein:
> Note: The Checks API only looks for pushes in the repository where the check suite or check run were created. Pushes to a branch in a forked repository are not detected and return an empty pull_requests array.

Have we tested running this from a fork?

shcheklein:
@rogermparent  what is the main reason to use checks API, btw? In our case I though that a simple binary status is enough, or am I missing something?

.github/workflows/link-check-deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+        output: >-
+          {"summary": "Running master diff Link Check on ${{ github.event.deployment.payload.web_url }}"}
+
+    - name: Run the link check script
+      id: check
+      uses: "iterative/link-check.action@master"
+      with:
+        configFile: "config/link-check/config.yml"
+        rootURL: "${{ github.event.deployment.payload.web_url }}"
+        output: checkAction
```
shcheklein, line 29: could you clarify what actually do we output here? some json structure? what does it contain?

shcheklein: also, re the exitCode - why can't we make it always return something non zero on a failure?

shcheklein:
okay, I better understand what check runs are about, still curious- if we need them to make everything look nicer or there some strong reason to use them?

.github/workflows/link-check-deploy.yml
rogermparent: There's no reason we can't, I just figured it'd be better to have the option than not.

rogermparent: > could you clarify what actually do we output here? some json structure? what does it contain?

It is a JSON structure, as that's what the Check Action takes for input on its output.
name: "build-test"
on: [push]

See this slice of its README, particularly `output`

It takes a required `summary`, which is used as the headline, and a markdown `text_description` to include details, like the report of failed links for us.

```
    - uses: LouisBrunner/checks-action@v0.1.0
      if: always()
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
        name: Test XYZ
        conclusion: ${{ job.status }}
        output: |
          {"summary":${{ steps.test.outputs.summary }}}
```

rogermparent:
New requested changes list for convenience:

- [ ] Make "filesystem" the default source
- [ ] Make Git Diff a special option
- [ ] Make exit code output a boolean option
- [ ] Update Check Action to 1.0
- [ ] Rename "bottlenecks" to something more evocative and general
- [ ] Check docs about existing link check
- [ ] Use longhand `--report-unused-patterns` instead of `-u`
- [ ] Improve documentation around `--report-unused-patterns only`

rogermparent:
> okay, I better understand what check runs are about, still curious- if we need them to make everything look nicer or there some strong reason to use them?

We need to make a Check for the deploy check because it must be triggered from `deployment_status` to execute when the deploy is done rather than `deployment` when the deploy has started. Only actions run from the `deployment` trigger are automatically added to the Check Run.

That actually gives me an idea on how to hide the initializer, though, as I think there's also a `deployment_status` trigger on deployment start which isn't automatically added to the Check Run.

shcheklein:
@rogermparent 

> New requested changes list for convenience:

excellent summary! thanks! 🙏 

> We need to make a Check for the deploy check because it must be triggered from deployment_status

I'm a bit worried about the forks workflow with Checks API ...

rogermparent:
- use Commander over Minimist for extra features like auto-help

- change all arguments to be consistently camelCase, even in CLI

- change `reportUnusedPatterns` to boolean `unusedPatternsOnly`, default
  behavior is now to always report unused but otherwise be silent unless using
  this flag.

- rename bottlenecks to linkOptions

- change `source` into boolean `diff`, using diff source when true and FS otherwise

- change exit code to always apply on CLI and GHA with CLI output mode

- add configurable default `minTime` and `maxConcurrent`

- all filePatternFiles options renamed to singular (PatternFile) because one
  file per is the only reasonable usage even though multiple can be used.

- exitCode output is now an internal part of the program.

.github/workflows/link-check-all.yml
rogermparent: `filesystem` is now the default mode, and `diff` is a boolean flag

.github/workflows/link-check-all.yml
rogermparent: The exit code is now separate from the output strategies, always applying on CLI and the CLI mode of the GHA

.github/workflows/link-check-deploy.yml
rogermparent: Turns out GitHub Actions doesn't support YAML pointers. Where it's at now is about as simple as the workflow can be while working around GHA's weird, inconsistent treatment of Checks.

config/link-check/config.yml
rogermparent: This is now renamed to `linkOptions`

package.json
rogermparent: `report-unused-patterns` is now reworked into `unused-patterns-only`.
Now unused link patterns are always reported, and `-u / --unused-patterns-only` makes it so the program exclusively checks for unused patterns using a dry run.

I also used longhand `--unused-patterns-only` in the configuration here.

rogermparent: A case-insensitive ripgrep for `link.?check` shows that no docs mention the link check, so I think we're good here.

shcheklein (approved):

PR #1624, opened 2020-07-24
repro 1.x : Updated downstream example and added info for Sequential execution.

sarthakforwet:
PR as an outcome of discussion at [#1572(review)](https://github.com/iterative/dvc.org/pull/1572#issuecomment-662673042).


content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,41 @@ The answer to universe is 42
 - The Hitchhiker's Guide to the  Galaxy
 ```
 
-Now, using the `--downstream` option results in the following output:
+And add a new stage to the pipeline:
 
 ```dvc
-$ dvc repro --downstream
+$ dvc run -n final -d count.txt -o alphabet.txt \
```
shcheklein, line 269: may be I'm missing something, but why do we need an additional stage?

content/docs/command-reference/repro.md
```diff
@@ -83,11 +82,12 @@ $ dvc dag
 ```
 
 This pipeline consists of two parallel branches (`A` and `B`), and the final
-"result" stage, where the branches merge. To reproduce both branches at the same
-time, you could run `dvc repro A2` and `dvc repro B2` at the same time (e.g. in
-separate terminals). After both finish successfully, you can then run
-`dvc repro train`: DVC will know that both branches are already up-to-date and
-only execute the final stage.
+"result" stage, where the branches merge. If you run `dvc repro` at this point,
```
jorgeorpinel, line 85:
```suggestion
`train` stage, where the branches merge. If you run `dvc repro` at this point,
```

content/docs/command-reference/repro.md
```diff
@@ -83,11 +82,12 @@ $ dvc dag
...
 
 This pipeline consists of two parallel branches (`A` and `B`), and the final
-"result" stage, where the branches merge. To reproduce both branches at the same
-time, you could run `dvc repro A2` and `dvc repro B2` at the same time (e.g. in
-separate terminals). After both finish successfully, you can then run
-`dvc repro train`: DVC will know that both branches are already up-to-date and
-only execute the final stage.
+`train` stage, where the branches merge. If you run `dvc repro` at this point,
+it would reproduce the complete pipeline with all stages executing sequentially.
+To reproduce both branches at the same time, you could run `dvc repro A2` and
```
jorgeorpinel, line 87:
```suggestion
To reproduce both branches simultaneously , you could run `dvc repro A2` and
```

content/docs/command-reference/repro.md
```diff
@@ -151,7 +151,8 @@ only execute the final stage.
   each execution, meaning the cache cannot be trusted for such stages.
 
 - `--downstream` - only execute the stages after the given `targets` in their
-  corresponding pipelines, including the target stages themselves.
+  corresponding pipelines, including the target stages themselves. This option
+  doesn't have any effect if no `targets` are provided.
```
jorgeorpinel, line 155:
```suggestion
  has no effect if no `targets` are provided.
```

content/docs/command-reference/repro.md
jorgeorpinel: Because it only had 2 stages which doesn't allow us to demo the --downstream option.

content/docs/command-reference/repro.md
```diff
@@ -83,11 +82,12 @@ $ dvc dag
 ```
 
 This pipeline consists of two parallel branches (`A` and `B`), and the final
-"result" stage, where the branches merge. To reproduce both branches at the same
-time, you could run `dvc repro A2` and `dvc repro B2` at the same time (e.g. in
-separate terminals). After both finish successfully, you can then run
-`dvc repro train`: DVC will know that both branches are already up-to-date and
-only execute the final stage.
+`train` stage, where the branches merge. If you run `dvc repro` at this point,
+it would reproduce the complete pipeline with all stages executing sequentially.
```
jorgeorpinel, line 86: Let's use

"If you run `dvc repro` at this point, it would reproduce each branch sequentially before `train`."

please.

content/docs/command-reference/repro.md
jorgeorpinel: However this command `"cat count.txt | egrep -o '[a-zA-Z]+' > alphabet.txt"` is probably not very meaningful... 🧠

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,41 @@ The answer to universe is 42
...
-Now, using the `--downstream` option results in the following output:
+And add a new stage to the pipeline:
 
 ```dvc
-$ dvc repro --downstream
+$ dvc run -n final -d count.txt -o alphabet.txt \
+            "cat count.txt | egrep -o '[a-zA-Z]+' > alphabet.txt"
+```
+
+Now, using the `--downstream` option with `count` as a target stage results in
```
jorgeorpinel, line 273:
```suggestion
Now, using the `--downstream` option with `count` as a target stage, results in
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,41 @@ The answer to universe is 42
 - The Hitchhiker's Guide to the  Galaxy
 ```
 
-Now, using the `--downstream` option results in the following output:
+And add a new stage to the pipeline:
 
 ```dvc
-$ dvc repro --downstream
+$ dvc run -n final -d count.txt -o alphabet.txt \
+            "cat count.txt | egrep -o '[a-zA-Z]+' > alphabet.txt"
```
jorgeorpinel, line 270: Let's try to change this last stage a little to something closer to real life. It needs a meaningful name and some new command. It probably shouldn't use count.txt as a dependency too because that's an untracked metric file (-M option) so how about using the numbers.txt file as dependency here? It doesn't matter that it was already used for `count`. Just process it in some other hypothetical way similar to how `count` has a hypothetical `process.py` Python script.

content/docs/command-reference/repro.md
shcheklein: it does, right? since downstream include the target stage itself. But I see it can be not so obvious. I just have a slight concern that example got bigger and more complicated .

jorgeorpinel:
Also, the changes per https://github.com/iterative/dvc.org/pull/1572#pullrequestreview-450392422 may also be missing still, but if it's too confusing or doesn't seem to apply now please lmk @sarthakforwet and I'll take care of it.

content/docs/command-reference/repro.md
jorgeorpinel: Hmmm it's true... the 3rd stage here doesn't really show much at the moment since nothing gets repro! We would need to also change numbers.txt manually or something like that to show that only the 2nd and 3rd stages get reproduced.

jorgeorpinel: ...which makes it even longer. You call Ivan!

content/docs/command-reference/repro.md
jorgeorpinel: ACTUALLY, this whole new stage may not be needed at all. Please follow the discussion above with Ivan and change accordingly. See https://github.com/iterative/dvc.org/pull/1624#pullrequestreview-455125609

content/docs/command-reference/repro.md
jorgeorpinel: p.s. the change to both files could be a single line though, instead of the "Hitchhikers..." reference:

```console
$ echo changed | tee text.txt numbers.txt
```

content/docs/command-reference/repro.md
sarthakforwet: > ACTUALLY, this whole new stage may not be needed at all.

Yeah you are right. Updated the example.

content/docs/command-reference/repro.md
sarthakforwet: >  since downstream include the target stage itself.

Yes you are correct @shcheklein. I also feel now that we can omit adding a new stage and change the `process.py` and `text.txt` files and show that `dvc repro --downstream count` only updates the count stage and not the stages above it. Thanks!

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,42 @@ The answer to universe is 42
...
+
+```python
+import sys
+num_digits = 0
+with open(sys.argv[1], 'r') as f:
+    for number in f:
+        num_digits += len(number) - 1
+print("Number of digits:",end=" ")
+print(num_digits)
+```
```
jorgeorpinel, line 276: This change is way to complex, why not something simple again like in lines 227-34? Also the Python code isn't properly formatted...

jorgeorpinel: I'm not even sure this update is needed...

jorgeorpinel: I guess it does clarify the example. Let's keep it! But please simplify the Python change. Should be 1 line.

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,42 @@ The answer to universe is 42
...
+Updating lock file 'dvc.lock'
 ```
 
-The reason being that the `text.txt` file is not a dependency in the last stage
-of the pipeline, used as the default target by `dvc repro`. `text.txt` is a
-dependency of the `filter` stage, which happens earlier (shown in the figure
-below), so it's skipped given the `--downstream` option.
+The reason being that the `text.txt` file is a dependency in the `filter` stage
+of the pipeline which happens before the `count` stage (shown in the following
+figure) and hence did not get updated.
```
jorgeorpinel, line 290: The reason for what? This explanation needs more detail.
"and hence did not get updated" what does this refer to? one of the stages? which one? the pipeline?
should be broken up into several sentences too.

content/docs/command-reference/repro.md
```diff
@@ -83,11 +82,12 @@ $ dvc dag
 ```
 
 This pipeline consists of two parallel branches (`A` and `B`), and the final
-"result" stage, where the branches merge. To reproduce both branches at the same
-time, you could run `dvc repro A2` and `dvc repro B2` at the same time (e.g. in
-separate terminals). After both finish successfully, you can then run
-`dvc repro train`: DVC will know that both branches are already up-to-date and
-only execute the final stage.
+`train` stage, where the branches merge. If you run `dvc repro` at this point,
+it would reproduce each branch sequentially before train. To reproduce both
```
jorgeorpinel, line 86: You missed the block quotes (hidden but existent in https://github.com/iterative/dvc.org/pull/1624#discussion_r460272991)
```suggestion
it would reproduce each branch sequentially before `train`. To reproduce both
```
I'll just commit this.

content/docs/command-reference/repro.md
jorgeorpinel: Thanks

content/docs/command-reference/repro.md
sarthakforwet: Thanks!

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,39 @@ The answer to universe is 42
 - The Hitchhiker's Guide to the  Galaxy
 ```
 
-Now, using the `--downstream` option results in the following output:
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print('Number of lines in %s:'%(sys.argv[1]))
```
jorgeorpinel, line 270: Still too complicated: Need to `import sys`, uses outdated `%` string formatting (f-string are the py3 standard), and it's still not well formatted (lacking proper spacing).

content/docs/command-reference/repro.md
jorgeorpinel: Continues in https://github.com/iterative/dvc.org/pull/1624#pullrequestreview-457247058 ...

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,39 @@ The answer to universe is 42
...
 ```
 
-The reason being that the `text.txt` file is not a dependency in the last stage
-of the pipeline, used as the default target by `dvc repro`. `text.txt` is a
-dependency of the `filter` stage, which happens earlier (shown in the figure
-below), so it's skipped given the `--downstream` option.
+The change in the `text.txt` file is ignored as it is a dependency in the
+`filter` stage which did not get updated in the above command. This is because
+the `filter` stage happens before the `count` stage in the pipeline (shown in
+the following figure).
```
jorgeorpinel, line 287: Thanks, better. But the text needs a little more improvement and can be simplified:
```suggestion
The change in `text.txt` is ignored because that file is a dependency in the
`filter` stage, which did not get updated in the above command. This is because
`filter` happens before `count` in the pipeline (shown below).
```
May need formatting.

jorgeorpinel: > which did not get updated in the above command

Still doesn't fully explain this. Should mention --downstream somewhere, or update the paragraph before the code block where --downstream is mentioned, to explain what it does (very briefly) please.

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,39 @@ The answer to universe is 42
...
+  +--------+
+  | filter |
+  +--------+
+      *
+      *
+      *
+  +-------+
+  | count |
+  +-------+
+```
```
jorgeorpinel, line 301:
````suggestion
```

> Refer to `dvc dag` for more details on that command.
````

File: content/docs/command-reference/repro.md (hunk 11 above)
Comment (sarthakforwet) [reply content/docs/command-reference/repro.md:270]: > Need to import sys 

`sys` is already imported in the file.

>  uses outdated % string formatting 

I think most of the people still uses it as I have seen this format on various Kaggle competitions and open source projects. (although I would change it to f-string).

>  it's still not well formatted (lacking proper spacing).

I have checked this code with py-lint and its showing no errors with it. Please can you suggest a linter to check for these errors. Thanks

File: content/docs/command-reference/repro.md [hunk 14]
```diff
@@ -281,10 +281,11 @@ Running stage 'count' with command:
...
 ```
 
-The change in the `text.txt` file is ignored as it is a dependency in the
-`filter` stage which did not get updated in the above command. This is because
-the `filter` stage happens before the `count` stage in the pipeline (shown in
-the following figure).
+The change in the `text.txt` file is ignored because that file is a dependency
+in the `filter` stage, which did not get updated in the above command. This is
+because `filter` happens before `count` in the pipeline (shown below) and the
+`--downstream` option only execute the stages after a given target stage
```
sarthakforwet, line 287: Should the line be "... and using the --downstream option..." ?

content/docs/command-reference/repro.md
jorgeorpinel: I see `sys.argv[1]` is already used indeed. And you changed to f-string, so resolving.

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,42 @@ The answer to universe is 42
...
+Running stage 'count' with command:
+        python process.py numbers.txt > count.txt
+Updating lock file 'dvc.lock'
 ```
 
-The reason being that the `text.txt` file is not a dependency in the last stage
-of the pipeline, used as the default target by `dvc repro`. `text.txt` is a
-dependency of the `filter` stage, which happens earlier (shown in the figure
-below), so it's skipped given the `--downstream` option.
+The change in the `text.txt` file is ignored because that file is a dependency
```
jorgeorpinel, line 284: Missed a change from my suggestion in https://github.com/iterative/dvc.org/pull/1624#discussion_r462085536
```suggestion
The change the `text.txt` is ignored because that file is a dependency
```

content/docs/command-reference/repro.md
jorgeorpinel:
```suggestion
because `filter` happens before `count` in the pipeline (shown below) and with the
`--downstream` option, `dvc repro` only executes the stages after a the target stage.
```

jorgeorpinel: But I think the other alternative I suggested in https://github.com/iterative/dvc.org/pull/1624#discussion_r462086427 is better, actually:

> or update the paragraph before the code block where --downstream is mentioned, to explain what it does (very briefly) please

sarthakforwet: > But I think the other alternative I suggested in #1624 (comment) is better, actually:

Okay. Updated accordingly.

I see that you added a full stop after "the target stage", so don't we need to mention that the target stage is also included when using `--downstream` option or is it evident from the output above this para?

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print(f'Number of lines in {sys.argv[1]}:')
+print(num_lines)
+```
+
+Now, using the `--downstream` option with `dvc repro`, results in the execution
+of stages after the target stage (`count` in this case) in the pipeline.
```
jorgeorpinel, line 275:
```suggestion
Now, using the `--downstream` option with `dvc repro` results in the execution
only of stages after the target (`count`).
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print(f'Number of lines in {sys.argv[1]}:')
+print(num_lines)
+```
+
+Now, using the `--downstream` option with `dvc repro` results in the execution
+only of stages after the target (`count`).
```
jorgeorpinel, line 275:
```suggestion
Now, using the `--downstream` option with `dvc repro` results in the execution
only of stages after the target (`count`):
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print(f'Number of lines in {sys.argv[1]}:')
+print(num_lines)
+```
+
+Now, using the `--downstream` option with `dvc repro` results in the execution
+only of stages after the target (`count`):
```
jorgeorpinel, line 275:
```suggestion
only of the target stage (`count`) and any following ones:
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Updating lock file 'dvc.lock'
 ```
 
-The reason being that the `text.txt` file is not a dependency in the last stage
-of the pipeline, used as the default target by `dvc repro`. `text.txt` is a
-dependency of the `filter` stage, which happens earlier (shown in the figure
-below), so it's skipped given the `--downstream` option.
+The change in `text.txt` is ignored because that file is a dependency in the
+`filter` stage, which did not get updated in the above command. This is because
+`filter` happens before `count` in the pipeline (shown below).
```
jorgeorpinel, line 286:
```suggestion
`filter` stage, which wasn't executed by the `dvc repro` above. This is because
`filter` happens before the target (`count`) in the pipeline, as shown below:
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print(f'Number of lines in {sys.argv[1]}:')
+print(num_lines)
+```
+
+Now, using the `--downstream` option with `dvc repro` results in the execution
+only of the target stage (and any following ones):
```
jorgeorpinel, line 275:
```suggestion
only of the target stage, and following ones (none in these case):
```

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
...
+Let's say we want to print the filename also in the description and so we update
+the `process.py` as:
+
+```python
+print(f'Number of lines in {sys.argv[1]}:')
+print(num_lines)
+```
+
+Now, using the `--downstream` option with `dvc repro` results in the execution
+only of the target stage, and following ones (none in these case):
```
shcheklein, line 275: this ?

content/docs/command-reference/repro.md
```diff
@@ -262,31 +263,40 @@ The answer to universe is 42
 - The Hitchhiker's Guide to the  Galaxy
 ```
 
-Now, using the `--downstream` option results in the following output:
+Let's say we want to print the filename also in the description and so we update
```
shcheklein, line 266: ... we also want ..

content/docs/command-reference/repro.md
jorgeorpinel: Yes, thanks. Committing:
```suggestion
only of the target stage, and following ones (none in this case):
```

content/docs/command-reference/repro.md
jorgeorpinel: Agree:
```suggestion
Let's say we also want to print the filename in the description, and so we update
```

jorgeorpinel:
Merged #1648!

PR #1634, opened 2020-07-28
Make image caption styling apply to plain `img`s

rogermparent:
This PR makes it so that plain `img` elements with captions are properly styled, as the previous rules only applied to images wrapped with `gatsby-image`, which are divs with a specific class.

I just took the relevant bits from the current selector and moved them into a new one that selects both `img` and the `gatsby-image` divs.

Example of the issue in the last few images of [this post](https://dvc.org/blog/shtab-completion-release).

[Here's the same blog post, but on this PR's preview](https://dvc-landing-img-caption-zfazra.herokuapp.com/blog/shtab-completion-release)

### DIFF
--- a/src/components/Blog/Post/Markdown/styles.module.css
+++ b/src/components/Blog/Post/Markdown/styles.module.css
@@ -58,6 +58,10 @@
     color: var(--color-gray);
   }
 
+  img + em {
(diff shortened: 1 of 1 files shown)


shcheklein:
thanks @rogermparent !! 🙏

PR #1642, opened 2020-07-31
Add a "Thank You" page for MailChimp subscribers

rogermparent:
This PR adds a simple "Thank You" page for MailChimp to redirect to upon a successful subscription.

Hosting this page ourselves gives us more control over the page in general, but a specific issue brought this up:

Having one Audience for CML and DVC is easier to manage, but there can only be one embedded form per Audience. This is fine for the most part, except that the "return to website" button on the "Thank you" page "returns" to dvc.org even if you got there via cml.dev! Even worse, we can't just remove the thank you button in MailChimp's provided editor!

Thankfully, we *can* redirect to a custom page, which is where this PR comes in.

I fully expect the text to be changed, and maybe we want return links to both DVC and CML to make up for the fact that the page can't change between sources when using one Audience.

See how it looks on the [deploy preview](https://dvc-landing-mailchimp-t-nywlsi.herokuapp.com/subscriber-thank-you)

### DIFF
--- a/src/components/ThankYou/index.tsx
+++ b/src/components/ThankYou/index.tsx
@@ -1,0 +1,11 @@
+import React from 'react'
+import styles from './styles.module.css'
+
+const ThankYouPage: React.FC = () => (
+  <div className={styles.wrapper}>
+    <h1 className={styles.heading}>Thanks for subscribing!</h1>
+    <p className={styles.copy}>Keep an eye on your inbox</p>
+  </div>
+)
+
+export default ThankYouPage
--- a/src/components/ThankYou/styles.module.css
+++ b/src/components/ThankYou/styles.module.css
@@ -1,0 +1,29 @@
+.wrapper {
+  margin: 12rem auto;
+  text-align: center;
+
+  @media (--md-scr) {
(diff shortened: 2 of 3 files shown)


shcheklein:
@rogermparent look great! out of curiosity - how do we setup the redirect?  it should be unique per website, right? can it be setup in the Mailchimp itself on the code level?

rogermparent:
> it should be unique per website, right? can it be setup in the Mailchimp itself on the code level?

Unfortunately, MailChimp's Embedded Forms doesn't allow us to setup separate redirects for separate sites, nor does it give any parameter to tell between the two. If we want to continue with one MailChimp Audience for both sites, we'll have to accept anyone that subscribes will be brought to the DVC thank you page. This is just a limitation of MailChimp, as far as I can tell.

This is a marginal improvement because the "Return to Site" button isn't present on the custom page, so even though a CML subscriber will see a DVC newsletter thank you page, there isn't an unintuitive return button and the user is led to close the success page instead.

shcheklein:
I see. Thanks @rogermparent , let's ship it!

PR #1643, opened 2020-07-31
Update external deps examples dvc run command

Robbie-Palmer:
Since the change to v1 of DVC, the description of generating a separate DVC file is no longer true

### DIFF
--- a/content/docs/user-guide/external-dependencies.md
+++ b/content/docs/user-guide/external-dependencies.md
@@ -35,8 +35,8 @@ directory.
 ## Examples
 
 As examples, let's take a look at a [stage](/doc/command-reference/run) that
-simply moves a local file from an external location, producing a `data.txt.dvc`
-stage file (DVC-file).
+simply downloads a file from an external location, adding a `download_file`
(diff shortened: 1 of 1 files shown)


shcheklein:
thanks @Robbie-Palmer !!

content/docs/user-guide/external-dependencies.md
```diff
@@ -35,40 +35,44 @@ directory.
...
 > Linux distributions.
 
 ### Local file system path
 
 ```dvc
-$ dvc run -d /home/shared/data.txt \
+$ dvc run -n download_file
+          -d /home/shared/data.txt \
           -o data.txt \
           cp /home/shared/data.txt data.txt
```
jorgeorpinel, line 50: It's not downloaded in the case of the local file system, where `cp` is used. But OK, the top case covers the base case/ intention. Maybe we should just add a note about this under the **Local file system path** header.

jorgeorpinel: Done in 42b670f1.

PR #1645, opened 2020-07-31
Update data-pipelines.md

noahshpak:
Saw a typo and got a bit carried away. I'm a big DVC fan! 💫 👨‍💻

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -301,10 +301,7 @@ important problems:
   files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
   ML pipelines allow CI/CD systems to retrain models on fresh
-  datasets with identical preprocessing and training stages, version upstream
-  models and datasets, and easily
-  [compare](https://dvc.org/doc/start/experiments#comparing-experiments) metrics
-  with currently deployed models.
+  datasets with identical training, and save the results.
 
 ## Visualize
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -296,8 +296,9 @@ important problems:
   your project faster. DVC caches "runs" and results in stages and automatically
   determines which parts of a project need to be run to avoid unnecessary
   re-runs.
-- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
-  data to use and what commands to run to produce an ML model. Storing these
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
+  and which commands will generate the pipeline results (such as an ML
+  model). Storing these
   files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
   ML pipelines allow CI/CD systems to retrain models on fresh
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -293,8 +293,8 @@ DVC pipelines (`dvc.yaml` file, `dvc run`, and `dvc repro` commands) solve a few
 important problems:
 
 - _Automation_ - run a sequence of steps in a "smart" way that makes iterating on
-  your project faster. DVC caches "runs" and results in stages and automatically
-  determines which parts of a project need to be run to avoid unnecessary
+  your project faster. DVC automatically determines which parts of a project
+  need to be run, and it caches "runs" and their results, to avoid unnecessary
   re-runs.
 - _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
   and which commands will generate the pipeline results (such as an ML
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -302,7 +302,8 @@ important problems:
   files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
   ML pipelines allow CI/CD systems to retrain models on fresh
-  datasets with identical training, and save the results.
+  datasets with identical training, save the results, and even produce reports
+  about the whole process. See [CML.dev](https://cml.dev/) for some examples.
 
 ## Visualize
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -300,8 +300,7 @@ important problems:
   data to use and what commands to run to produce an ML model. Storing these
   files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
-  ML pipelines allow CI/CD systems to retrain
-  ([repro](https://dvc.org/doc/command-reference/repro)) models on fresh
+  ML pipelines allow CI/CD systems to retrain models on fresh
   datasets with identical preprocessing and training stages, version upstream
   models and datasets, and easily
   [compare](https://dvc.org/doc/start/experiments#comparing-experiments) metrics
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -292,7 +292,7 @@ prepare:
 DVC pipelines (`dvc.yaml` file, `dvc run`, and `dvc repro` commands) solve a few
 important problems:
 
-- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
+- _Automation_ - run a sequence of steps in a "smart" way that makes iterating on
   your project faster. DVC caches "runs" and results in stages and automatically
   determines which parts of a project need to be run to avoid unnecessary
   re-runs.
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -292,14 +292,20 @@ prepare:
 DVC pipelines (`dvc.yaml` file, `dvc run`, and `dvc repro` commands) solve a few
 important problems:
 
-- _Automation_ - run a sequence of steps in a "smart" way to iterate on your 
-  project faster. DVC caches "runs" and results in stages to avoid unnecessary
+- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
   re-runs.
-- _Reproducibility_ - YAML files describe and capture what data to use and
-  what commands to run to produce an ML model. Storing these files in Git
-  makes it easy to version and share.
-- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  reproducible ML pipelines (builds) facilitates CI/CD systems.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
+  data to use and what commands to run to produce an ML model. Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
+  ML pipelines allow CI/CD systems to retrain
+  ([repro](https://dvc.org/doc/command-reference/repro)) models on fresh
+  datasets with identical preprocessing and training stages, version upstream
+  models and datasets, and easily
+  [compare](https://dvc.org/doc/start/experiments#comparing-experiments) metrics
+  with currently deployed models.
 
 ## Visualize
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -300,10 +300,9 @@ important problems:
   and which commands will generate the pipeline results (such as an ML
   model). Storing these
   files in Git makes it easy to version and share.
-- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
-  ML pipelines allow CI/CD systems to retrain models on fresh
-  datasets with identical training, save the results, and even produce reports
-  about the whole process. See [CML.dev](https://cml.dev/) for some examples.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
+  projects in way that it can be reproduced (built) is the fist necessary step
+  before introducing CI/CD systems.
 
 ## Visualize
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -292,17 +292,14 @@ prepare:
 DVC pipelines (`dvc.yaml` file, `dvc run`, and `dvc repro` commands) solve a few
 important problems:
 
-- _Automation_ - run sequence of steps in a "smart" way that makes iterating on
-  the project faster. It automatically determines which parts of a project need
-  to be run, it caches "runs" and results — all to avoid running the same stage
-  again.
-- _Reproducibility_ - it can describe and capture what data should be used and
-  what commands to run to produce an ML model, for example. It's described and
-  captured in way that is easy to put into Git. It means that it's easy to
-  version and share.
+- _Automation_ - run a sequence of steps in a "smart" way to iterate on your 
+  project faster. DVC caches "runs" and results in stages to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - YAML files describe and capture what data to use and
+  what commands to run to produce an ML model. Storing these files in Git
+  makes it easy to version and share.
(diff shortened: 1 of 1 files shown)


content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,14 @@ prepare:
...
-  the project faster. It automatically determines which parts of a project need
-  to be run, it caches "runs" and results — all to avoid running the same stage
-  again.
-- _Reproducibility_ - it can describe and capture what data should be used and
-  what commands to run to produce an ML model, for example. It's described and
-  captured in way that is easy to put into Git. It means that it's easy to
-  version and share.
+- _Automation_ - run a sequence of steps in a "smart" way to iterate on your 
+  project faster. DVC caches "runs" and results in stages to avoid unnecessary
+  re-runs.
```
jorgeorpinel, line 297: - "makes iterating on the project faster" was more accurate I think.
- "It automatically determines which parts of a project need to be run" was very important"

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,14 @@ prepare:
...
+- _Reproducibility_ - YAML files describe and capture what data to use and
+  what commands to run to produce an ML model. Storing these files in Git
+  makes it easy to version and share.
```
jorgeorpinel, line 300: Let's use `` `dvc.yaml` and `dvc.lock` files describe and capture what data...``

I agree with the other changes to this bullet 👍

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,14 @@ prepare:
...
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  project in way that it can be reproduced (built) is the fist necessary step
-  before introducing CI/CD systems.
+  reproducible ML pipelines (builds) facilitates CI/CD systems.
```
jorgeorpinel, line 302: Definitely better. But not sure about "(builds)" What builds? Builds what? "builds" the plural noun? No need to summarize so much that it's ambiguous, let's try to be as explicit as necessary. Same with "facilitates": how does it facilitate them?

noahshpak: Thank you for the feedback! I agree. Hopefully my additions are more explicit.

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,20 @@ prepare:
...
-  what commands to run to produce an ML model, for example. It's described and
-  captured in way that is easy to put into Git. It means that it's easy to
-  version and share.
-- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  project in way that it can be reproduced (built) is the fist necessary step
-  before introducing CI/CD systems.
+- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
+  re-runs.
```
jorgeorpinel, line 298: - Again, "way that makes iterating" is more accurate.
- `DVC caches "runs" and results in stages` - this isn't accurate. The text here was referring to a feature called run-cache which we haven't documented yet. Stages are not where results are cached.

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,20 @@ prepare:
...
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
+  data to use and what commands to run to produce an ML model. Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
+  ML pipelines allow CI/CD systems to retrain
+  ([repro](https://dvc.org/doc/command-reference/repro)) models on fresh
```
jorgeorpinel, line 304:
```suggestion
  ML pipelines allow CI/CD systems to retrain models on fresh
```

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,20 @@ prepare:
...
-  project in way that it can be reproduced (built) is the fist necessary step
-  before introducing CI/CD systems.
+- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
+  data to use and what commands to run to produce an ML model. Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
```
shcheklein, line 302: Noah, first of all thanks for the contribution! 🙏 

Can we though keep this CI/CD piece a bit simpler? Potentially, replace "forward" links like repro and metrics to the CML.dev project that has some visual examples?

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,20 @@ prepare:
...
+  ML pipelines allow CI/CD systems to retrain
+  ([repro](https://dvc.org/doc/command-reference/repro)) models on fresh
+  datasets with identical preprocessing and training stages, version upstream
+  models and datasets, and easily
+  [compare](https://dvc.org/doc/start/experiments#comparing-experiments) metrics
+  with currently deployed models.
```
jorgeorpinel, line 308: Thanks but now we're introducing too many concepts here, I think. It's just an introductory doc 🙂 
```suggestion
  datasets with identical training, and save the results.
```

content/docs/start/data-pipelines.md
jorgeorpinel: I already simplified it but good idea about CML.dev !

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,16 @@ prepare:
...
-  version and share.
-- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  project in way that it can be reproduced (built) is the fist necessary step
-  before introducing CI/CD systems.
+- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
+  data to use and what commands to run to produce an ML model. Storing these
```
jorgeorpinel, line 300: > to produce an ML model

It used to be "model, for example". That's because data pipelines can have any results, it doesn't even have to be an ML application. Would be ideal to have it more general (but still mention the "ML model" key word).

content/docs/start/data-pipelines.md
jorgeorpinel: How about something like

DVC automatically determines which parts of a project need to be run, and caches "runs" and their results, to avoid unnecessary re-runs.

content/docs/start/data-pipelines.md
jorgeorpinel: Maybe something like

...files describe what data to use and what commands to execute to generate results (such as ML models).

content/docs/start/data-pipelines.md
noahshpak: That sounds great to me! Thank you for the assist here

content/docs/start/data-pipelines.md
noahshpak: ah yes I am definitely tunnel visioned toward my use case. Pipeline results is nice and general 👍

content/docs/start/data-pipelines.md
jorgeorpinel: OK, will apply 🙂

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,16 @@ prepare:
...
-  to be run, it caches "runs" and results — all to avoid running the same stage
-  again.
-- _Reproducibility_ - it can describe and capture what data should be used and
-  what commands to run to produce an ML model, for example. It's described and
-  captured in way that is easy to put into Git. It means that it's easy to
-  version and share.
-- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  project in way that it can be reproduced (built) is the fist necessary step
-  before introducing CI/CD systems.
+- _Automation_ - run a sequence of steps in a "smart" way to make iterating on
```
jorgeorpinel, line 295:
```suggestion
- _Automation_ - run a sequence of steps in a "smart" way that makes iterating on
```
Per https://github.com/iterative/dvc.org/pull/1645#discussion_r463785986

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,16 @@ prepare:
...
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
+  re-runs.
```
jorgeorpinel, line 298:
```suggestion
  your project faster. DVC automatically determines which parts of a project
  need to be run, and caches "runs" and their results, to avoid unnecessary
  re-runs.
```
Also per https://github.com/iterative/dvc.org/pull/1645#discussion_r463785986

content/docs/start/data-pipelines.md
jorgeorpinel, line 300:
```suggestion
- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
  and which commands will generate the pipeline results (such as an ML
  model). Storing these
```

jorgeorpinel: OK, I'll update it.

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,16 @@ prepare:
...
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe and capture what
+  data to use and what commands to run to produce an ML model. Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
+  ML pipelines allow CI/CD systems to retrain models on fresh
+  datasets with identical training, and save the results.
```
jorgeorpinel, line 304:
```suggestion
  datasets with identical training, save the results, and even produce reports
  about the whole process. See [CML.dev](https://cml.dev/) for some examples.
```
Per https://github.com/iterative/dvc.org/pull/1645#pullrequestreview-459404692

content/docs/start/data-pipelines.md
jorgeorpinel: Applied in https://github.com/iterative/dvc.org/pull/1645#discussion_r463802976.

jorgeorpinel:
(Planing to merge the Restyled PR.)

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,18 @@ prepare:
...
+- _Automation_ - run a sequence of steps in a "smart" way that makes iterating on
+  your project faster. DVC caches "runs" and results in stages and automatically
+  determines which parts of a project need to be run to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
+  and which commands will generate the pipeline results (such as an ML
+  model). Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
+  ML pipelines allow CI/CD systems to retrain models on fresh
```
shcheklein, line 304: this describes a very specific use case. CI/CD can include running the last stage for example, full retraining, just training remotely instead of local ... etc

content/docs/start/data-pipelines.md
shcheklein: are we merging this change? ^^

jorgeorpinel: Oh yeah forgot to commit this... Thanks

content/docs/start/data-pipelines.md
```diff
@@ -292,17 +292,18 @@ prepare:
...
+  need to be run, and it caches "runs" and their results, to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
+  and which commands will generate the pipeline results (such as an ML
+  model). Storing these
+  files in Git makes it easy to version and share.
+- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
+  ML pipelines allow CI/CD systems to retrain models on fresh
+  datasets with identical training, save the results, and even produce reports
+  about the whole process. See [CML.dev](https://cml.dev/) for some examples.
```
jorgeorpinel, line 306: How about this @shcheklein ?
```suggestion
- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - reproducible
  ML pipelines allow CI/CD systems to reproduce all or part of the process (e.g.
  model training) on production datasets. Using DVC, they can then save the
  results, and even produce reports about the whole process. See
  [CML.dev](https://cml.dev/) for some examples.
```

jorgeorpinel: Feel free to edit/commit it and merge the Restyled PR if you agree 🙂 or lmk and I'll do so.

shcheklein: it still describes one very specific case "reproduce all or part of the process ... on production datasets"

let's go back a bit, gusy. What was the problem with initial version? So, that I understand how to help you solve it.

jorgeorpinel: It just had a small typo, Noah decided to rewrite it all, for readability I think. Here's the original version (typo fixed) in case we want to just go back to that:
```suggestion
- _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
  projects in way that it can be reproduced (built) is the fist necessary step
  before introducing CI/CD systems.
```

shcheklein: So, let's just fix the typo then? if the problem is not clear. Changing language a bit is fine either if it sounds/reads better, but let's not change the essence if there is no clear understanding what exactly are we fixing.

jorgeorpinel: Yes, that's what I meant. The typo is fixed in the suggestion I left. We just need to commit it. Allow me...

noahshpak: To me the original version echos reproducibility rather than pipeline flexibility. The piece that matters for CI/CD is DVC can figure out when something important has changed (data, code, etc) and re-run a pipeline correctly. Like what @jorgeorpinel said "pipelines allow CI/CD systems reproduce all or part of the process."

content/docs/start/data-pipelines.md
```diff
@@ -292,16 +292,16 @@ prepare:
...
+  your project faster. DVC automatically determines which parts of a project
+  need to be run, and it caches "runs" and their results, to avoid unnecessary
+  re-runs.
+- _Reproducibility_ - `dvc.yaml` and `dvc.lock` files describe what data to use
+  and which commands will generate the pipeline results (such as an ML
+  model). Storing these
+  files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  project in way that it can be reproduced (built) is the fist necessary step
+  projects in way that it can be reproduced (built) is the fist necessary step
```
shcheklein, line 304: it -> they?

jorgeorpinel: Oops I merged the REstyled PR #1647. I'll continue this elsewhere... ⏳

content/docs/start/data-pipelines.md
jorgeorpinel: Thanks Noah, I'll chat with Ivan and continue this elsewhere if needed, since this has been merged. But feel free to contribute another PR if you prefer 🙂

jorgeorpinel:
Thanks again @noahshpak !

content/docs/start/data-pipelines.md
noahshpak: No problem! Again, thanks for all of the feedback

shcheklein: yep. I agree with the `pipelines allow CI/CD systems reproduce all or part of the process` part. `on production datasets` and mentioning `train` make it specific.

> The piece that matters for CI/CD is DVC can figure out when something important has changed (data, code, etc) and re-run a pipeline correctly

not sure this it true. Since it's already committed it can force `dvc repro` for example to guarantee that external system gets the same result. There are other examples when we can "abuse" CI/CD to do the training first time- we don't change anything per-se - everything comes with a commit, system runs it first time.

shcheklein: Also, I think CML.dev link still make a lot of sense.

PR #1649, opened 2020-08-01
docs: Std. SSH URLs and rel. notes about SFTP

jorgeorpinel:
per https://github.com/iterative/dvc/issues/4167#issuecomment-663318837


content/docs/command-reference/remote/modify.md
```diff
@@ -440,7 +440,7 @@ more information.
 - `url` - remote location URL.
 
   ```dvc
-  $ dvc remote modify myremote url ssh://user@example.com:1234/path/to/remote
+  $ dvc remote modify myremote url ssh://user@example.com:1234/absolute/path
```
jorgeorpinel, line 443: BTW @efiop HDFS doesn't have this config param? `url` I mean (it's not in the doc currently). Thanks

content/docs/user-guide/external-dependencies.md
```diff
@@ -54,11 +54,13 @@ $ dvc run -n download_file
...
 ```dvc
 $ dvc run -n download_file
-          -d ssh://user@example.com:/home/shared/data.txt \
+          -d ssh://user@example.com/home/shared/data.txt \
           -o data.txt \
-          scp user@example.com:/home/shared/data.txt data.txt
+          scp user@example.com/home/shared/data.txt data.txt
 ```
 
+> Please notice `/home/...` is an absolute path from the remote system's root.
```
efiop, line 62: Strictly speaking this is not about system's root, but rather sftp root directory, which doesn't always match system's root.

content/docs/command-reference/remote/modify.md
efiop: All remotes do. That's what `dvc remote add` sets.

content/docs/user-guide/managing-external-data.md
```diff
@@ -66,21 +66,23 @@ $ dvc run -d data.txt \
 
 ```dvc
 # Add SSH remote to be used as cache location for SSH files
-$ dvc remote add sshcache ssh://user@example.com:/cache
+$ dvc remote add sshcache ssh://user@example.com/home/.../cache
```
efiop, line 69: `/cache` is absolute anyway. `...` is clearly not needed.

content/docs/user-guide/external-dependencies.md
shcheklein, line 62: instead of notes - let's do /absolute/root/path/from... ? (I don't like notes - there should be a very strong reason to use them).

content/docs/user-guide/managing-external-data.md
```diff
@@ -66,21 +66,23 @@ $ dvc run -d data.txt \
...
-$ dvc add --external ssh://user@example.com:/mydata
+$ dvc add --external ssh://user@example.com/home/.../mydata
 
 # Create the stage with an external SSH output
 $ dvc run -d data.txt \
           --external \
-          -o ssh://user@example.com:/home/shared/data.txt \
-          scp data.txt user@example.com:/home/shared/data.txt
+          -o ssh://user@example.com/home/shared/data.txt \
+          scp data.txt user@example.com/home/shared/data.txt
```
efiop, line 81: I'm not sure that this is a valid scp url, I would rather not touch it.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Will address this in #1680.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Note updated.

jorgeorpinel: Good idea, done. But `/absolute/path/from/sftp/root/which/istnt/always/system/root` would be excessive. Even without the "which..." part, I think. So I kept the note for now to warn about the common SFTP root confusion (see https://github.com/iterative/dvc.org/pull/1649#pullrequestreview-459591916 above).

I agree it's best to avoid notes (like avoiding comments in code) but at the same time this is exactly what ⚠️ side-notes (about external tools) are for, and we have plenty of them.

content/docs/user-guide/external-dependencies.md
```diff
@@ -54,11 +54,14 @@ $ dvc run -n download_file
...
 $ dvc run -n download_file
-          -d ssh://user@example.com:/home/shared/data.txt \
+          -d ssh://user@example.com/abs/path/to/data \
           -o data.txt \
-          scp user@example.com:/home/shared/data.txt data.txt
+          scp user@example.com/abs/path/to/data data
 ```
 
+> ⚠️ Please notice `/abs/path/to/data` is an absolute path from the SFTP root
+> (not always configured to be the system root).
```
jorgeorpinel, line 63: Althouogh... TBH Idk if we can call this "absolute". Isn't it relative to the SFTP root? (which may or may not be the system root, in which case this would be an abs path indeed)

content/docs/user-guide/managing-external-data.md
jorgeorpinel: `/cache` is an absolute path from a SSH system root? Not getting this.

`...` means "more stuff here", what do you mean it's clearly not needed?

jorgeorpinel: p.s. Updating the `...` part per https://github.com/iterative/dvc.org/pull/1649#pullrequestreview-459592083 below anyway.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: p.s. I did remove the notes completely from HDFS sections.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Good catch, thanks. But we do want to remove `:` from the SSH URL above, right?

jorgeorpinel:
Sorry, this one fell through the cracks for over a week... Addressed all the feedback now.

content/docs/user-guide/external-dependencies.md
efiop: Yes, as noted in https://github.com/iterative/dvc.org/pull/1649#discussion_r463982286

jorgeorpinel: OK I updated the note... But it's confusing now 🤔🧠

content/docs/user-guide/external-dependencies.md
jorgeorpinel: UPDATE: With the note (only on SSH) now being

> ⚠️ Please notice `/abs/path/to/...` is a path relative to the SFTP root
> (typically the system root, in which case it's an absolute path).

it may be confusing again since the URL starts with `abs`/`absolute` but the note calls it "relative"... So I guess just `/path/from/sftp/root` + further simplify or remove the note may be better after all...

jorgeorpinel:
Just one discussion pending here @shcheklein but we should probably try to merge #1649 first? Please check it when you have a change — I think it's ready.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: alright, I assume those `:` are OK to have been removed 🙂

content/docs/command-reference/remote/add.md
```diff
@@ -322,7 +322,7 @@ like `ssh` and `sftp` (GNU/Linux).
 ### Click for HDFS
 
 ```dvc
-$ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
+$ dvc remote add -d myremote hdfs://user@example.com/absolute/path
```
shcheklein, line 325: we have abs in some places in some absolute - not a big deal. jfyi

content/docs/user-guide/external-dependencies.md
```diff
@@ -54,11 +54,14 @@ $ dvc run -n download_file
...
 ```dvc
 $ dvc run -n download_file
-          -d ssh://user@example.com:/home/shared/data.txt \
+          -d ssh://user@example.com/abs/path/to/data.txt \
           -o data.txt \
-          scp user@example.com:/home/shared/data.txt data.txt
+          scp user@example.com:/abs/path/to/data.txt data.txt
 ```
 
+> ⚠️ Please notice `/abs/path/to/data` is a path relative to the SFTP root
```
shcheklein, line 62: it reads strange absolute path that is relative :)

content/docs/user-guide/managing-external-data.md
```diff
@@ -66,21 +66,24 @@ $ dvc run -d data.txt \
...
-$ dvc add --external ssh://user@example.com:/mydata
+$ dvc add --external ssh://user@example.com/abs/path/to/mydata
 
 # Create the stage with an external SSH output
 $ dvc run -d data.txt \
           --external \
-          -o ssh://user@example.com:/home/shared/data.txt \
-          scp data.txt user@example.com:/home/shared/data.txt
+          -o ssh://user@example.com/abs/path/to/data.txt \
+          scp data.txt user@example.com:/abs/path/to/data.txt
```
shcheklein, line 81: are you sure that scp will convert it the same way as ssh dependency? (if they have different roots)

content/docs/command-reference/get-url.md
```diff
@@ -33,14 +33,14 @@ directory will be placed inside.
...
-| `ssh`   | SSH server     | `ssh://user@example.com:/path/to/data`     |
-| `hdfs`  | HDFS to file\* | `hdfs://user@example.com/path/to/data.csv` |
-| `http`  | HTTP to file\* | `https://example.com/path/to/data.csv`     |
+| Type    | Description    | `url` format                                   |
+| ------- | -------------- | ---------------------------------------------- |
+| `local` | Local path     | `/path/to/local/data`                          |
+| `s3`    | Amazon S3      | `s3://mybucket/data`                           |
+| `gs`    | Google Storage | `gs://mybucket/data`                           |
+| `ssh`   | SSH server     | `ssh://user@example.com/abs/path/to/data`      |
+| `hdfs`  | HDFS to file\* | `hdfs://user@example.com/abs/path/to/data.csv` |
```
shcheklein, line 42: why do we change HDFS? was there a confusion about it?

content/docs/command-reference/remote/add.md
jorgeorpinel: Yeah it was to keep it short but maybe "abs" is confusing. Anyway, will probably have to come up with another sample path per https://github.com/iterative/dvc.org/pull/1649#discussion_r468679544... ⏳

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Agree. Will address.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: No, I will have to check ⏳ 

@efiop can you please confirm we want to remove all those `:` from SSH URLs? I think that was a problem on a recent support case and you mentioned we shouldn't have them in docs to avoid confusion.

content/docs/command-reference/get-url.md
jorgeorpinel: I think so. See https://github.com/iterative/dvc/issues/4167#issuecomment-656468558

content/docs/user-guide/external-dependencies.md
jorgeorpinel: OK... Changed again, using longer `/path/from/sftp/root`... now.

content/docs/command-reference/remote/add.md
jorgeorpinel: Left only full "absolute" ones now.

content/docs/command-reference/get-url.md
shcheklein: that comment applies to SSH only, I have no idea about HDFS. With SSH the problem that commands like SCP resolve path in a certain way when you specify it, and you can pass a relative one, etc. So, there was error and trial process to get the path right.

content/docs/user-guide/external-dependencies.md
```diff
@@ -41,64 +41,57 @@ stage to your list of stages in dvc.yaml.
...
+          -d gs://mybucket/data \
+          -o data \
+          gsutil cp gs://mybucket/data data
 ```
 
+### SSH
+
+```dvc
+$ dvc run -n download_file
+          -d ssh://user@example.com/path/from/sftp/root/to/data \
```
shcheklein, line 79: I'm not sure .. but I think it got too complicated

content/docs/user-guide/managing-external-data.md
efiop: from our urls we can remove `:` , from scp urls - we can't remove `:`

jorgeorpinel: > Sorry for the delay, I fell behind again in recent days. Catching up now-Sun.

So I could not confirm this from Windows because while I was able to setup OpenSSH it won't let me private key auth and so I have to type in the password to connect. And apparently this is a limitation of DVC? See iterative/dvc/issues/4392

As a consequence I can't confirm this. @efiop can you please confirm what the text above should be for now? (with or without `:` in the SSH URL)

content/docs/command-reference/get-url.md
jorgeorpinel: OK. I rolled back the "absolute" mentions in HDFS URLs.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: I thought so too at first as discussed in https://github.com/iterative/dvc.org/pull/1649#pullrequestreview-459592083 but this way a longer note about paths relative to SFTP root is avoided.

jorgeorpinel: I'm happy to go back to a shorter sample path like just `/data` and expand on the note below. But what would we do with the other `path/from/sftp/root` example URLs in get/import-url ? May need to add a similar note in those docs, which we can do.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Ah I see Ruslan already confirmed the current text version should be fine!

efiop: @jorgeorpinel For the record: that issue is not a dvc limitation, but just misconfiguration on your side. Left a comment in that issue.

jorgeorpinel: K. Will answer over there.

content/docs/command-reference/import-url.md
```diff
@@ -55,16 +55,16 @@ source.
...
-| `hdfs`   | HDFS to file (explanation below)                    | `hdfs://user@example.com/path/to/data.csv` |
-| `http`   | HTTP to file with _strong ETag_ (explanation below) | `https://example.com/path/to/data.csv`     |
-| `remote` | Remote path (see explanation below)                 | `remote://myremote/path/to/data`           |
+| Type     | Description                                         | `url` format                                 |
+| -------- | --------------------------------------------------- | -------------------------------------------- |
+| `local`  | Local path                                          | `/path/to/local/data`                        |
+| `s3`     | Amazon S3                                           | `s3://mybucket/data`                         |
+| `azure`  | Microsoft Azure Blob Storage                        | `azure://my-container-name/path/to/data`     |
+| `gs`     | Google Cloud Storage                                | `gs://mybucket/data`                         |
+| `ssh`    | SSH server                                          | `ssh://user@example.com/path/from/sftp/root` |
```
shcheklein, line 64: I'm worried a bit about this exposing too much information that can just scary people when it's not needed.

e.g. regular SSH server on ubuntu works perfectly - I don't care if it's sftp or not. sftp in the path might confuse.

content/docs/command-reference/remote/add.md
```diff
@@ -299,7 +299,7 @@ $ export OSS_ACCESS_KEY_SECRET='AccessKeySecret'
 ### Click for SSH
 
 ```dvc
-$ dvc remote add -d myremote ssh://user@example.com/path/to/dir
+$ dvc remote add -d myremote ssh://user@example.com/absolute/path
```
shcheklein, line 302: why is it just absolute path here if we use sftp in other places?

content/docs/user-guide/external-dependencies.md
```diff
@@ -41,64 +41,57 @@ stage to your list of stages in dvc.yaml.
...
+          -o data \
+          scp user@example.com:/path/from/sftp/root/to/data data
```
shcheklein, line 81: not sure that scp is using sftp and it's the same convention ... how often sftp root is not /?

content/docs/command-reference/import-url.md
jorgeorpinel: True. I guess it's not something we need to bring up in these refs. Rolled back in 6ec6717f.

content/docs/command-reference/remote/add.md
jorgeorpinel: Thanks. I removed these now and standardized path examples and SFTP notes in remote add/modify as well as the external data guides. See f68eb406de63c813baa2f550e302117a2da3bf0b.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Good point about `scp` not being SFTP! Missed that 🤦 I'm going to remove these stfp mentions from the paths... ⌛ 

>how often sftp root is not /?

Not sure but it's a note we already have in remote add so it has been deemed important before. Plus per this https://github.com/iterative/dvc/issues/4167#issuecomment-663318837 I thought we definitely wanted to mention this clearly in all the docs where it may cause issues.

jorgeorpinel: OK so... I'm removing all the `/from/sftp/...` paths again. Having the note underneath should be enough I think.

In fact, should we just remove these SFTP notes completely? And or consider renaming `ssh://` to `sftp://` if that's the actual protocol we're using... Then users could be expected to know what kinds of URLs and paths are supported. Cc @efiop

content/docs/command-reference/remote/add.md
```diff
@@ -307,16 +307,11 @@ $ dvc remote add -d myremote ssh://user@example.com/path/to/dir
...
-⚠️ DVC requires both SSH and SFTP access to work with SSH remote storage. Please
-check that you are able to connect both ways to the remote location, with tools
-like `ssh` and `sftp` (GNU/Linux).
-
-> Note that your server's SFTP root might differ from its physical root (`/`).
-> (On Linux, see the `ChrootDirectory` setting in `/etc/ssh/sshd_config`.) In
-> these cases, the path component in the SSH URL (e.g. `/path/to/dir` above)
-> should be specified relative to the SFTP root instead. For example, on some
-> Sinology NAS drives, the SFTP root might be in directory `/volume1`, in which
-> case you should use path `/path/to/dir` instead of `/volume1/path/to/dir`.
```
jorgeorpinel, line 319: I decided to remove these details. They seem more like tips unrelated to DVC. Users can research SFTP details on other sites easily, I think. Left a quick note: "Note that the server's SFTP root might differ from its physical root (`/`)." that should be enough of a tip for the few users running into that issue.

content/docs/user-guide/external-dependencies.md
```diff
@@ -41,24 +41,6 @@ stage to your list of stages in dvc.yaml.
...
-
-### SSH
-
-```dvc
-$ dvc run -n download_file
-          -d ssh://user@example.com:/home/shared/data.txt \
-          -o data.txt \
-          scp user@example.com:/home/shared/data.txt data.txt
-```
-
```
jorgeorpinel, line 61: Sorry, slightly unrelated: reordered these 2 sections to match the order in remote add/modify.

jorgeorpinel: Same in content/docs/user-guide/external-dependencies.md

content/docs/user-guide/external-dependencies.md
```diff
@@ -90,15 +72,29 @@ $ dvc run -n download_file
...
 
 ```dvc
 $ dvc run -n download_file
-          -d hdfs://user@example.com/home/shared/data.txt \
+          -d hdfs://user@example.com/data.txt \
           -o data.txt \
           hdfs fs -copyToLocal \
-                  hdfs://user@example.com/home/shared/data.txt \
-                  data.txt
+                  hdfs://user@example.com/data.txt data.txt
```
jorgeorpinel, line 97: Also did some changes to HDFS URls since from the original issue/comment that produced this PR it sounded like they also needed an update.

content/docs/command-reference/get-url.md
```diff
@@ -39,7 +39,7 @@ DVC supports several types of (local or) remote locations (protocols):
 | `local` | Local path     | `/path/to/local/data`                      |
 | `s3`    | Amazon S3      | `s3://mybucket/data`                       |
 | `gs`    | Google Storage | `gs://mybucket/data`                       |
-| `ssh`   | SSH server     | `ssh://user@example.com:/path/to/data`     |
+| `ssh`   | SSH server     | `ssh://user@example.com/path/to/data`      |
```
shcheklein, line 42: do we need the trailing space?

content/docs/command-reference/import-url.md
```diff
@@ -62,7 +62,7 @@ DVC supports several types of (local or) remote locations (protocols):
 | `s3`     | Amazon S3                                           | `s3://mybucket/data`                       |
 | `azure`  | Microsoft Azure Blob Storage                        | `azure://my-container-name/path/to/data`   |
 | `gs`     | Google Cloud Storage                                | `gs://mybucket/data`                       |
-| `ssh`    | SSH server                                          | `ssh://user@example.com:/path/to/data`     |
+| `ssh`    | SSH server                                          | `ssh://user@example.com/path/to/data`      |
```
shcheklein, line 65: same

content/docs/command-reference/get-url.md
shcheklein: sounds good! and we can elaborate when we have a separate page per remote I guess

content/docs/command-reference/remote/add.md
shcheklein: ounds good! and we can elaborate when we have a separate page per remote I guess

shcheklein:
approved! just a few minor comments ...

content/docs/user-guide/external-dependencies.md
efiop: > In fact, should we just remove these SFTP notes completely? 

Could just create 1 note in `remote modify/add` help. The rest could be removed, yes.

> nd or consider renaming ssh:// to sftp:// if that's the actual protocol we're using...

Not really worth renaming. Plus, we do use ssh too, but only for external data management, so it is not a direct swap. Would just keep as is for now.

content/docs/command-reference/get-url.md
jorgeorpinel: Yes. It's all formatted by Prettier this way.

content/docs/command-reference/import-url.md
jorgeorpinel: Auto-formatted. To align columns I guess. Looks "prettier" tehehe

content/docs/user-guide/external-dependencies.md
jorgeorpinel: I left the notes everywhere just in case. It's a shorter note version now, anyway.

Agreed about not renaming. Thanks @efiop

content/docs/command-reference/remote/add.md
jorgeorpinel: We're planning to do separate pages per remote? I don't recall that decision. It sounds like a good plan except that the remote types are already burried in docs->cmd ref->remote->add/modify->expandable sections so not sure adding yet another level to that tree is a good idea.

Maybe the remote types should all be in the remote ref. index? And the remote concept explanation in Basic Concepts.

Issue #1507, opened 2020-06-28
dvc run step in tutorial does not give expected results


shachibista:
## Bug Report

```console
$ dvc version
DVC version: 1.0.1
Python version: 3.7.1
Platform: Linux-4.4.0-176-generic-x86_64-with-debian-stretch-sid
Binary: True
Package: deb
Supported remotes: azure, gdrive, gs, hdfs, http, https, s3, ssh, oss
Cache: reflink - not supported, hardlink - supported, symlink - supported
Filesystem type (cache directory): ('ext4', '/dev/mapper/qwerty--vg-root')
Repo: dvc, git
Filesystem type (workspace): ('ext4', '/dev/mapper/qwerty--vg-root')
```

Following https://dvc.org/doc/use-cases/versioning-data-and-model-files/tutorial, `dvc run` step returns an error instead of creating the `Dvcfile`/`dvc.yaml` file:

```
dvc run -f Dvcfile \
          -d train.py -d data \
          -M metrics.csv \
          -o model.h5 -o bottleneck_features_train.npy -o bottleneck_features_validation.npy \
          python train.py

ERROR: `-n|--name` is required
```

iterative/dvc#4077 states that `-f` option is deprecated. Running with the `-n` option results in a subsequent error:

```
dvc run \
    -d train.py \
    -d data \
    -n training \
    -M metrics.csv \
    -o model.h5 \
    -o bottleneck_features_train.npy \
    -o bottleneck_features_validation.npy \
    python train.py --verbose

ERROR: output 'model.h5' is already specified in stage: 'model.h5.dvc'.
```

Please update the tutorial to reflect changes in v1.0.

Thanks!

shcheklein:
@shachibista sorry about that. It's a little bit outdated (DVC version <0.94). We changed the `dvc run` interface in the DVC 1.0 and haven't had time to update some tutorials.

shachibista:
@shcheklein Would it require a lot of (conceptual) changes to update the tutorial? If it is only a single command, I could send a pull request if you could tell the correct parameters.

shcheklein:
@shachibista no huge conceptual changes! Just changing commands, altering text here and there (e.g. no Dvcfile anymore, but `dvc.yaml`).

To start I think we can try to remove `-f something` and put `-n train` - that alone should probably fix the issue.

jorgeorpinel:
[...]

PR #1658, opened 2020-08-05
[hotfix] cases: fix Automating capturing section of versioning tutorial

jorgeorpinel:
Re-close #1615 

Specifically https://github.com/iterative/dvc.org/issues/1507#issuecomment-662913116:

> Removing the model.h5.dvc file and re-running the updated tutorial runs successfully! ...
Perhaps it would be clearer if the documentation re-stated this before the Automating capturing section.

### DIFF
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -309,11 +309,13 @@ When you have a script that takes some data as an input and produces other data
 
 > If you tried the commands in the
 > [Switching between workspace versions](#switching-between-workspace-versions)
-> section, go back to the master branch code and data with:
+> section, go back to the master branch code and data, and remove the
+> `model.h5.dvc` file with:
 >
 > ```dvc
 > $ git checkout master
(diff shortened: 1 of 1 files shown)


content/docs/use-cases/versioning-data-and-model-files/tutorial.md
```diff
@@ -309,11 +309,13 @@ When you have a script that takes some data as an input and produces other data
...
 > If you tried the commands in the
 > [Switching between workspace versions](#switching-between-workspace-versions)
-> section, go back to the master branch code and data with:
+> section, go back to the master branch code and data, and remove the
+> `model.h5.dvc` file with:
 >
 > ```dvc
 > $ git checkout master
 > $ dvc checkout
+> $ dvc remove model.h5.dvc
```
shcheklein, line 318: let's use a regular `rm -f` - no need for a fancy command here

shcheklein: rm -f might be wrong to use though ... okay, let's keep dvc remote, disregard please

PR #1661, opened 2020-08-05
cmd: explain project/repo nesting in init ref.

jorgeorpinel:
> Extracted from #1661 

- [ ] Ideally not but yes in the way things are structured now in this ref. This isolation and nested situation got too complicated... I should probably make a separate sub-section to cover it after explaining all the 3 kinds of projects (typical or git-enabled, --subdir, --no-scm).
- [ ] On Git submodules: "I would expect from people who use submodules have the same intuition by default - isolated in Git, isolated in DVC - nothing special or specif", "True, I can just remove it and handle them via support channel if needed" (from https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-461335110)
- [ ] On block code examples for nested repos: ".git still not important here" (from https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-461337818)


content/docs/command-reference/init.md
```diff
@@ -100,6 +100,54 @@ DVC sets the `core.no_scm` config option value to `true` in the DVC
 that even if the project is tracked by Git, or if Git is initialized in it
 later, DVC will keep operating detached from Git in this project.
 
+### Nested DVC projects and repositories
```
shcheklein, line 103: Doing PR this way makes it hard to review (we lost the context since we removed the whole section somewhere else). :(

shcheklein: Let's revert the way it was before please and then do changes on top?

jorgeorpinel: We already reverted all of the changes about this in the previous PR.

jorgeorpinel: OK I think you mean having the code blocks back in the --subdir section... Hold on ⌛

jorgeorpinel: Reverted for now but those changes were already merged to master so reverting here looks like a change. This is still a draft BTW, not quite ready for review.

jorgeorpinel: I can remove the new section here, and then we can merge this, so a following PR shows you the changes they way I think you'd like to see them ⌛

jorgeorpinel: Ready.

PR #1664, opened 2020-08-06
cmd: simplify project nesting info in init

jorgeorpinel:
Notes extracted from #1661: 

> Ideally not but yes in the way things are structured now in this ref. This isolation and nested situation got too complicated... I should probably make a separate sub-section to cover it after explaining all the 3 kinds of projects (typical or git-enabled, --subdir, --no-scm).

> On Git submodules: "I would expect from people who use submodules have the same intuition by default - isolated in Git, isolated in DVC - nothing special or specif", "True, I can just remove it and handle them via support channel if needed" (from https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-461335110)

> On block code examples for nested repos: ".git still not important here" (from https://github.com/iterative/dvc.org/pull/1615#pullrequestreview-461337818)

### DIFF
--- a/content/docs/command-reference/init.md
+++ b/content/docs/command-reference/init.md
@@ -100,6 +100,54 @@ DVC sets the `core.no_scm` config option value to `true` in the DVC
 that even if the project is tracked by Git, or if Git is initialized in it
 later, DVC will keep operating detached from Git in this project.
 
+### Nested DVC projects and repositories
+
+> Note that Git-enabled <abbr>DVC repositories</abbr> nested inside parent DVC
+> repositories (for example when using Git submodules) are isolated from the
+> parent, and vice versa. This is because each one has their own Git root.
+
+If there are multiple `--subdir` projects, but not nested, e.g.:
+
+```dvc
+.           # git init
+├── .git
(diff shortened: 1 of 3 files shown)


shcheklein:
@jorgeorpinel I think it was in a decent shape. Let's leave as is for now?

jorgeorpinel:
Yeah sorry, I couldn't help myself from pushing these changes. Only took me like 10 min. Basically the code block examples in --subdir are removed, this simplifies the cmd ref a little. Everything else are copy edits. See https://dvc-landing-cmd-init-ne-juucow.herokuapp.com/doc/command-reference/init

But if you prefer to close this that's also OK, I'll re-apply the copy edits in a small regular updates PR that I probably need to submit tomorrow. Thanks

shcheklein:
@jorgeorpinel we have some unrelated changes here now? was it intentional?

I still don't see a strong reason to touch it (except small copy edits). Let's create a ticket if needed, prioritize and discuss it there first.

jorgeorpinel:
> some unrelated changes here now?

Oops those were meant for another branch...

PR #1668, opened 2020-08-06
import: clarify -o behavior for subdirectories

pmrowla:

* Clarify `dvc import -o` behavior when output path contains subdirectories.
* See https://github.com/iterative/dvc/pull/4340

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -78,7 +78,8 @@ from the source repo.
 - `-o <path>`, `--out <path>` - specify a path (directory and/or file name) to
   the desired location to place the imported file in (instead of using the
   current working directory). If an existing directory is specified, the target
-  data will be placed inside.
+  data will be placed inside. If `path` contains a parent directory which does
+  not exist, this command will fail.
 
 - `--rev <commit>` - commit hash, branch or tag name, etc. (any
   [Git revision](https://git-scm.com/docs/revisions)) of the repository to


shcheklein:
Thanks @pmrowla , really appreciate this! 

Just curious - is it the same for get?

pmrowla:
@shcheklein For get we do a fetch + checkout and create the output subdirectory if it doesn't already exist, since that's the normal behavior for checkout.

content/docs/command-reference/import.md
```diff
@@ -78,7 +78,8 @@ from the source repo.
 - `-o <path>`, `--out <path>` - specify a path (directory and/or file name) to
   the desired location to place the imported file in (instead of using the
   current working directory). If an existing directory is specified, the target
-  data will be placed inside.
+  data will be placed inside. If `path` contains a parent directory which does
+  not exist, this command will fail.
```
jorgeorpinel, line 82: Sorry for late review @pmrowla, I have a Q:

> If `path` contains a parent directory

What does that mean? "Containing a parent" sounds contradictory. Parents contain you 😋

jorgeorpinel: Plus how can you contain something that does not exist? Then you don't contain it... Sorry, just confused by this.

pmrowla: If you run `dvc import -o foo/bar` and `foo/` doesn't already exist, `import` will fail

jorgeorpinel: I see, got it. I may reword a little. Thanks Peter!

jorgeorpinel:
> For get we ... create the output subdirectory if it doesn't already exist

I'll add this note too.

BTW I get the technicality involving internal processes (checkout) but it's kind of weird that get creates them but import doesn't... And why doesn't import use checkout? Hmmmm 🤔

pmrowla:
It's because in import we are defining an actual stage/dependency, but for get it's treated more like a standalone file

from https://github.com/iterative/dvc/pull/4340#discussion_r466158245

> We also don't makedirs in dvc run, but rather raise an error that wdir doesn't exist, so it seems it would be more consistent if we raise and error here too. Same with dvc add that shares the same logic with dvc run.

jorgeorpinel:
Makes sense, that's relevant context for sure. My idea would be that for consistency maybe no DVC commands should makedirs... In fact probably only `get` does that. But anyway, I won't open a ticket about this, up to you to bring it up with the core team if you wish Peter 🙂

PR #1677, opened 2020-08-07
Revert "Blog CML self-hosted runners on demand with GPUs"

DavidGOrtega:
The merge should have been squashed


DavidGOrtega:
@elleobrien this is going to add a new commit in the history but there is nothing else I can do. I dont have permissions in the repo

shcheklein:
I guess not point if we just updating dates on top.

PR #1680, opened 2020-08-08
docs: misc. updates

### LINKED ISSUE #1683: docs: insufficient description of params
https://github.com/iterative/dvc.org/blob/7546c1f45a435ebd7840426a9349c402e72bace1/content/docs/user-guide/dvc-files-and-directories.md#L153-L155

The link to _parameter dependencies_ takes to the `params` where it's not clear where I find what it means (unless I read 341 words, who has time for it).

Also, the meaning of  _key paths_ is not clear, and makes it even more confusing since it talks about "referring to a ... file".

It also does not mention any way to specify a custom params file.

Discord Context: https://discord.com/channels/485586884165107732/563406153334128681/741913494991142932

jorgeorpinel:
- Matches https://github.com/iterative/dvc/pull/4364 (rewrite sections and short descs. about URLs (remote, get/import-url))
- Review of term "feature(s)"
- Closes #1683
- Misc. copy edits


content/docs/command-reference/remote/add.md
```diff
@@ -12,30 +12,15 @@ usage: dvc remote add [-h] [--global | --system | --local] [-q | -v]
                       [-d] [-f] name url
 
 positional arguments:
-  name           Name of the remote.
-  url            URL. (See supported URLs in the examples below.)
+  name           Name of the remote
+  url            URL (see supported URLs in the examples below.)
```
jorgeorpinel, line 16:
```suggestion
  url            Remote location.
                 See full list of supported URLs below.
```

content/docs/command-reference/remote/modify.md
```diff
@@ -122,12 +135,6 @@ these settings, you could use the following options:
   $ dvc remote modify myremote secret_access_key my-secret_access_key
   ```
 
-- `url` - remote location URL:
```
shcheklein, line 125: Let's not optimize out this? People don't read texts as you expect. A lot of them would just jump to the section they need.

shcheklein, line 125: I think it worth keeping them, just for another piece of explicit information in the right context that might help to figure out what's going on

content/docs/command-reference/remote/add.md
```diff
@@ -44,6 +30,25 @@ url = /tmp/dvc-storage
 remote = myremote
 ```
 
+`name` and `url` are required. The `name` is used to identify the remote and
+should be unique for the project.
+
+`url` specifies a location to store your data. It can represent a cloud storage
+service, an SSH server, network-attached storage, or even a directory in the
+local file system (see all the supported remote storage types in the examples
+below). Whenever possible, DVC will create a remote directory if it doesn't
```
shcheklein, line 39: this is not true for all remotes (e.g. not for GDrive)

content/docs/command-reference/remote/add.md
```diff
@@ -44,6 +30,25 @@ url = /tmp/dvc-storage
 remote = myremote
 ```
 
+`name` and `url` are required. The `name` is used to identify the remote and
+should be unique for the project.
```
shcheklein, line 34: should -> must?

content/docs/command-reference/remote/modify.md
jorgeorpinel: But we have a section called **Available parameters for all remotes** and `url` is one of them. Should I add all the examples in that section? Currently I only added a few types. Or if you're sure I'll roll back the change... But that whole section only exists for `verify` right now. Seems a little pointless (i.e. I could also copy `verify` to all the type sections.

content/docs/command-reference/remote/add.md
jorgeorpinel: The part about creating dirs? Yeah it is pre-existing text. I think that's what who ever wrote it meant by "Whenever possible" and "it won't create an S3 bucket though".

Should we remove this or add a full list of which types support this? Cc @efiop hi! Do you happen to know for which remote types this is the case? Thanks 🙂

content/docs/command-reference/remote/add.md
jorgeorpinel: Updated.

jorgeorpinel:
Sorry @shcheklein , I added more stuff to this PR before realizing you had already approved it. Please check once more. Won't add anything else.

content/docs/command-reference/remote/add.md
efiop: Bucket/Dir creation differs from remote to remote, so let's not mention it for now at all. We need to unify it in the code first.

content/docs/command-reference/remote/modify.md
shcheklein: I would expect a lot of people to start reading right from the section and skip the description. Let's not optimize it for the sake of optimization only.

Also, example (multiple dvc remote add commands) looks scary :) - it's not cleat that this is a variation of a single command, or is it something I would have do to even start with remotes.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -150,9 +150,9 @@ the possible following fields:
   (the file's location).
 - `deps`: List of <abbr>dependency</abbr> file or directory paths of this stage
   (relative to `wdir` which defaults to the file's location)
-- `params`: List of [parameter dependencies](/doc/command-reference/params).
-  These are key paths referring to a YAML, JSON or TOML file (`params.yaml` by
-  default).
+- `params`: List of <abbr>parameter<abbr> dependencies. Contains subfields that
+  specify key/value pairs that are read from an arbitrary YAML, JSON, or TOML
```
shcheklein, line 154: They are field names from an arbitrary ....

content/docs/user-guide/troubleshooting.md
```diff
@@ -1,6 +1,7 @@
 # Troubleshooting
 
-Here we provide help for some of the problems that DVC user might stumble upon.
+Here we provide help for some of the problems that DVC user might stumble upon,
+based on frequent support questions (FAQ).
```
shcheklein, line 4: FAQ != Frequent Support questions though. Also, not clear if this change is useful.

content/docs/command-reference/remote/modify.md
jorgeorpinel: Optimizing wasn't even my main goal here. We currently explain `url` for each type in remote add, but in [remote modify](https://dvc.org/doc/command-reference/remote/modify#available-parameters-per-storage-type) we only mention it in 5 of them. I noticed this inconsistency and wanted to complete the info., but decided to put them all in the "for all remotes" section to avoid bloating the doc (that's the optimization part) and since it makes sense — what is that section for otherwise, just for `verify`?

> I would expect a lot of people to start reading right from the section

I personally would also check the "for all remotes" section along with the type I'm interested on.

But anyway: for now I've put URL both in the "all remotes" (general desc. and local remote example) and in every "per storage type" (specific examples except local, which doesn't have a section here). WDYT?

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: True. Updated but kept "key/value pairs" because both the name and last value are saved in dvc.yaml.

content/docs/user-guide/troubleshooting.md
jorgeorpinel: It was introduced for SEO. We don't really have the term "FAQ" anywhere. Rolled this back for now but should we put it somewhere? Maybe in the User Guide index even.

content/docs/command-reference/remote/add.md
jorgeorpinel: Removed.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -150,8 +150,8 @@ the possible following fields:
   (the file's location).
 - `deps`: List of <abbr>dependency</abbr> file or directory paths of this stage
   (relative to `wdir` which defaults to the file's location)
-- `params`: List of [parameter dependencies](/doc/command-reference/params).
-  These are key paths referring to a YAML, JSON or TOML file (`params.yaml` by
+- `params`: List of <abbr>parameter<abbr> dependencies. They are key/value pairs
```
shcheklein, line 153: they are not values though, just keys, right?

content/docs/command-reference/remote/modify.md
```diff
@@ -58,6 +58,15 @@ manual editing could be used to change the configuration.
 
 The following config options are available for all remote types:
 
+- `url` - the remote location can always be modified. This is how DVC determines
+  what type of remote it is, and thus which other config options can be modified
+  (see each type in the next section for more details). Here's a _local remote_
+  example:
+
+  ```dvc
+  $ dvc remote modify myremote url /home/user/dvcstore
```
shcheklein, line 67: let's do s3 example and can even explain about s3 prefix, or do another example if needed in a separate block?

content/docs/command-reference/remote/modify.md
shcheklein: sounds good!

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Ah yes I was a bit confused with dvc.loc. BTW I just updated the dvc.lock part as well accordingly. See d5e7a468.

content/docs/command-reference/remote/modify.md
jorgeorpinel: OK added S3 example before local example although I'm not sure it's needed as it's also in the S3 section. The S3 URL format is also explained in its section. See 3b7898e3

PR #1687, opened 2020-08-11
term: must vs. should

jorgeorpinel:
> Non-extensive in the case of "should" — there's too many instances.

\+ some copy edits around the instances I did review


content/docs/user-guide/basic-concepts/parameter.md
```diff
@@ -0,0 +1,8 @@
+---
```
shcheklein, line 1: unrelated change?

shcheklein:
it looks like we have a mix of changes in this PR, please clean it up first a bit

content/docs/user-guide/basic-concepts/parameter.md
jorgeorpinel: Shoot. This is from another PR (#1680)...

jorgeorpinel:
Something very wrong happened with the branching here... Sorry

jorgeorpinel:
Trying again in #1692.

PR #1705, opened 2020-08-20
guide: begin How To section

jorgeorpinel:
> Extracted from #1581

JUST A DRAFT TO START THIS GOING — Please don't review yet.

- Moves Updating Tracked Files into the new How To section
- ~~Creates the Best Practices how to guide (for #72)~~ Extracted to #1748
- How to undo `dvc add` (Fixes #625)

Ref: #899 and https://documentation.divio.com/how-to-guides/#how-to

AFTER MERGE

- [ ] This doc can be expanded for stage outputs.
  > From https://github.com/iterative/dvc.org/pull/1705#issuecomment-695832489

Issue #820, opened 2019-11-26
use-cases: revise so they're more high level "landing pages"


jorgeorpinel:
Use cases should have a different focus, so they are more high level and serve as landing pages – single web page that appears in response to clicking on a search engine optimized search result – leaving most details to the user guide. Some relevant quotes from https://github.com/iterative/dvc.org/pull/784#issuecomment-558403171:

> This section should answer the question of an infrastructure, DevOps, MLOps folks on how DVC can help them. How can it be used, in what high level scenarios. (And it's even okay if they overlap in this case.)

> To reiterate on the Use Cases vs User Guide:
**Users Guide** is like a car owner's manual - a lot of details, how tos, etc.
**Use Cases** would be like explaining how that car can be used (potentially referring a lot to the owner's manual to cover the technical details) - driving from home to work, renting, taxi, etc

~~Note: perhaps we could split this issue into 3. Not 4 because Data Registry is already written in this direction in #818.~~

Ideas:

- [ ] IN PROGRESS #654 extract details from use cases and possibly other redundant docs (e.g. some tutorials) into ~~blog posts~~ user guide? - #2482

---

Extracted to #2490:

- [x] #1722 write use cases correctly and extract how-to/tutorials into separate pages
- [ ] IN PROGRESS? #862 new case: **DVC in Production** (probably touch MLOps topics) - #2404
- [x] #857 Special content previews for SEO
- [x] #986 small **shared dev server** case update
- [ ] #194 About ML on a remote machine
- [ ] #2431 About specific ways to deploy ML models
- [ ] #2270 About Experiment Management
- ~~QUESTIONED #682 new case about data **partitioning and compression** ?~~
   > Use cases should have a different focus, so they are more high level and serve as landing pages – single web page that appears in response to clicking on a search engine optimized search result – leaving most details to the user guide.

jorgeorpinel:
Priority @shcheklein? 🙂

shcheklein:
I think p2 for now. Too many things in play already.

jorgeorpinel:
Summary on MLOps-related use case ideas:

- **Training on a remote machine** #194 was the first idea and would narrowly focus on remote SSH executors. We can prob consider it part 3.0 experiments now, I think? (I can't find the related issue or discussion but I know there's one).
- **DVC in Production** #862 sounds broad but it intends to define and doc best practices around "synchronizing" DVC pipelines between experimental local dev and changing prod envs (I think). Cc @dmpetrov 
[...]

Issue #899, opened 2020-01-06
how: new (sub)section (How To's)


jorgeorpinel:
It seems from several open issues that we may need a section or subsection somewhere in the docs to provide certain "how to" guides. For example, these are the issues with "how to" in their titles atm:

> UPDATE: See https://documentation.divio.com/how-to-guides/#how-to

- [x] Organize existing how-to material into the new section (some may be under guides now).
- [x] explain **how to** "undo" 'dvc add' + other "**how to** undo mistakes" recipes #625
- [x] commit: add example on **how to** add missing dependency to existing pipeline #612
- [ ] how to update tracked directories (part of #3297)
- [ ] https://github.com/iterative/dvc.org/issues/1079 ?

Other ideas:

- search old "how to" questions in `dvc` repo e.g. https://github.com/iterative/dvc/issues/599, forum, etc.
- cmd ref: improve `run` and `commit` / **how to** add output without re-running stage? #460
- "**How to** remove data" — what other ways to remove data (other than un-doing `add`) are there?
- "**How to** host a public DVC project" (private-write, public-read)
- ~~"**How to** uninstall dvc" for both binary and pip packages~~ #65
- This https://github.com/iterative/dvc.org/pull/732#pullrequestreview-306762834 discussion:
  > we'll have to introduce a how to section soon... May be we can move things like "Running on Windows" into this section
- Also, this https://github.com/iterative/dvc.org/pull/868#pullrequestreview-338379858 comment:
  > **Parallel stage execution**  from the `dvc repro` cmd ref can be probably moved to a **How to** section, with a link to it from the cmd ref
- How tos on `dvc init --no-scm` usage? See https://github.com/iterative/dvc.org/issues/100#issuecomment-657783092

Let's analyze this list, look for any others, and see whether it makes sense to have a dedicated how-to section, or whether an existing section of the docs e.g. use-cases or specific command references or user guides can cover all of these.

jorgeorpinel:
💡 If we do need this new section, perhaps it could live near the new **Troubleshooting** section (in **User Guide**).

PseudoNerd:
Can we expand the troubleshooting section by researching and discovering known errors and beginner mistakes while using DVC to make a list, and then analyze the formed list to make a dedicated how-to page with information pulled from the docs for better access? 


I'd like to participate in the first project for this year's edition of GSoD and DVC is something that I have used and had some experience with.

jorgeorpinel:
Hey @PseudoNerd!

[...]

PR #1716, opened 2020-08-23
cases: Versioning Data and Model Files (review)

jorgeorpinel:
Addresses #820

- [x] + Simplify the Use Cases index (dvc.org/doc/use-cases)
- [x] Language review + copy edits
- [x] Refocus writing (remove distracting parts).
- [x] Give it a better structure.
- [x] Reference other docs more.
---
- [x] Be very clear and explicit about the role of DVC vs Git
- [x] Expand on run/dag (pipelines) and/or dvc.yaml/lock versioning.
---
- [x] Rewrite as a higher level _Explanation_. See https://github.com/iterative/dvc.org/pull/1716#issuecomment-679316702
- [x] Fix and improve links from other docs to this case and its's tutorial.
- ~~Extract _How-to_ steps into actual how-to guide? (hidden for now, see #1705)~~ Extracted to #1722


jorgeorpinel:
Requesting a general review of the doc's structure and direction. Probably easiest to read it here directly: https://dvc-landing-use-cases-iezica0y.herokuapp.com/doc/use-cases/versioning-data-and-model-files.

shcheklein:
To be honest I don't see any major changes. What I would expect is that we go at least one level up in terms of how we explain this use case. It is still very technical, with exact commands, etc. We should migrate this into tutorials, how-tos

jorgeorpinel:
Agree. My first step was to check that at the current level the doc covers everything it should. Now I'm working on rewriting it to fit in the use case category.

jorgeorpinel:
BTW, based on https://documentation.divio.com/introduction/ Use Cases are not fully documentation. They are part _Explanation_ I guess, and they can contain some _How-to_ steps (in fact the Shared Cache one is one of our most popular "tutorials", I think). But they're also 50% marketing, closer to our current https://dvc.org/features page sections.

May be interesting to pull them out of /docs, even merge with /features perhaps. Moving the How-to parts to the actual /how-to docs of course (and link from the use case). This would make them proper landing pages (possibly with their own design, even).

jorgeorpinel:
> I would expect is that we go at least one level up in terms of how we explain this use case. It is still very technical

I simplified the intro and **DVC is not Git!** sections of the use case. See https://dvc-landing-use-cases-iezica0y.herokuapp.com/doc/use-cases/versioning-data-and-model-files.

The rest of the content was also updated but it will need to be migrated like you mentioned, and just briefly mentioned//linked to the new location. However that's also the case for all other use cases including [Data Registries](https://dvc.org/doc/use-cases/data-registries), and depends a little on having started the How-to section (in progress, see #899). So can we try to merge this and let me do the same initial simplification to the remaining use cases @shcheklein ?

I'll open a separate issue for migrating the how-to/tutorial parts.

I also improved mentions of the versioning case and it's tutorials throughout the docs.

jorgeorpinel:
> So can we try to merge this and let me

Per call with Ivan, we'll use this same PR to try and simplify the 2nd half of the Versioning use case, moving low-level code blocks to its tutorial... ⏳

jorgeorpinel:
OK @shcheklein I shortened the more instructional parts of the use case (from **Track data and models for versioning** on). It's still long but much more like a generic demo than a how-to / tutorial. Similar to the level of the **Data Registries** use case. PTAL

Track data and models for versioning

We may need to further shorten these use cases and make/reuse their tutorials instead, but I feel like that should be a separate issue? Thanks

jorgeorpinel:
See https://dvc-landing-use-cases-iezica0y.herokuapp.com/doc/use-cases/versioning-data-and-model-files

content/docs/command-reference/import.md
```diff
@@ -65,7 +65,7 @@ path in the <abbr>workspace</abbr>. It records enough metadata about the
 imported data to enable DVC efficiently determining whether the local copy is
 out of date.
 
-To actually [track the data](/doc/tutorials/get-started/data-versioning),
+To actually [version the data](/doc/tutorials/get-started/data-versioning),
```
shcheklein, line 68: not sure this is the best term ... but okay, let' keep it for now

content/docs/command-reference/add.md
```diff
@@ -19,17 +19,18 @@ The `dvc add` command is analogous to `git add`, in that it makes DVC aware of
...
+with Git.
 
 The `targets` are the files or directories to add, which are turned into
 <abbr>data artifacts</abbr> of the <abbr>project</abbr>. These are stored in the
 <abbr>cache</abbr> by default (use the `--no-commit` option to avoid this, and
 `dvc commit` to finish the process when needed).
 
-> See also `dvc run` for more advanced ways to version intermediate and final
-> results (like ML models).
+> See also `dvc.yaml` and `dvc run` for more advanced ways to track & version
```
shcheklein, line 32: & -> and

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,77 +1,121 @@
...
+$ git commit -m "Train model via DVC."
+$ git tag -a "v1.0" -m "Fist model via DVC" # We'll use this soon ;)
 ```
 
-There are two ways to get to the previous version of the dataset or model: a
-full <abbr>workspace</abbr> checkout, or checkout of a specific data or model
-file. Let's consider the full checkout first. It's quite straightforward:
+> See also `dvc.lock`.
+
+## Switching versions
```
shcheklein, line 114: we don't need 3-4 technical sections, we need to give some sense how does it look like, but it's not "how to".

shcheklein:
@jorgeorpinel looks better, but it is still stays on the same level (it feels this at least), we still need to go level or two up.

content/docs/command-reference/import.md
jorgeorpinel: "To actually version the data, git add (and git commit) the import stage." seems accurate to me 🙂

jorgeorpinel: Plus the link goes to the Versioning use case.

content/docs/command-reference/add.md
jorgeorpinel: Updated.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK, in order to remove more stuff I need to make sure whatever I remove here is covered by the tutorial so I'll review both now ⏳

jorgeorpinel: OK, it wasn't that difficult. See latest version please. I think this is pretty much the shortest it can be without completely removing the 2 technical sections left now, I think.

jorgeorpinel:
Alright. I went just 1 level up now, I think. Not sure anything in the technical sections would be left if we go up 2 levels (depending what you meant by that).

shcheklein:
@jorgeorpinel good progress, but overall I feel that we moved 0.5 level, not 1 👍 

What I would expect to see to make my feedback more actionable:

- Why versioning of large files is important at all. In ML or not.
- Why DVC is beneficial for this
- Give some sense how does DVC feel and look, etc. See the philosophy
- Connect with other cases

I guess, my point is that most likely we would need to write a new doc, not edit the existing one, to actually move +1 level.

jorgeorpinel:
Understood. I'll work on rewriting it completely then ⏳ 

Should we merge this for now?

jorgeorpinel:
Or should we close it in favor of #1747 ?

shcheklein:
> Or should we close it in favor of #1747 ?

@jorgeorpinel not sure I can recommend on this to be honest, it depends on your flow. From my end, it would be great to have a single PR.

jorgeorpinel:
I think this one is done. Maybe it's not what we need but if it's an improvement and you found no errors, etc. then I would like to merge it. But we can also close it since it will be overwritten by the other PR at some point, hopefully soon.

shcheklein:
Fine, we can merge it!

> Maybe it's not what we need but if it's an improvement and you found no errors, etc

I didn't look for errors to be honest, was not reviewing it to deep since to my mind we need to redo this case on a different level.

jorgeorpinel:
OK thanks! I think it's good to merge this because people are reading it and it's quite outdated.

PR #1721, opened 2020-08-26
docs: misc updates

jorgeorpinel:
A few unrelated notes I had stashed from other work.


content/docs/user-guide/external-dependencies.md
```diff
@@ -110,6 +110,11 @@ $ dvc run -n download_file
 
 ### Local file system path
 
+For local paths outside of your project:
+
+> This includes different storage devices or partitions mounted on the same file
```
shcheklein, line 115: I think it's more or less clear w/o that note

jorgeorpinel: Note removed.

PR #1728, opened 2020-08-27
docs: misc. updates

jorgeorpinel:
- 3 unrelated corrections in cmd ref explanations.
- Improvements to basic remote refs.

### DIFF
--- a/content/docs/command-reference/remote/add.md
+++ b/content/docs/command-reference/remote/add.md
@@ -19,8 +19,9 @@ positional arguments:
 ## Description
 
 This command creates a `remote` section in the <abbr>DVC project</abbr>'s
-[config file](/doc/command-reference/config) and optionally assigns a default
-`remote` in the `core` section, if the `--default` option is used:
+[config file](/doc/command-reference/config) and optionally assigns a _default
+remote_ in the `core` section, if the `--default` option is used (recommended
+for the first remote):
 
 ```ini
 ['remote "myremote"']
@@ -28,6 +29,10 @@ url = /tmp/dvc-storage
 [core]
 remote = myremote
 ```
+
+> 💡 Default remotes are expected by commands that accept a `-r`/`--remote`
+> option (`dvc pull`, `dvc push`, `dvc status`, `dvc gc`, `dvc fetch`) when that
+> option is omitted.
 
 `name` and `url` are required. The `name` is used to identify the remote and
 must be unique for the project.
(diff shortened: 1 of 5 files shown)


content/docs/command-reference/remote/add.md
```diff
@@ -47,13 +47,6 @@ DVC will determine the [type of remote](#supported-storage-types) based on the
 > include them all. The command should look like this: `pip install "dvc[s3]"`.
 > (This example installs `boto3` library along with DVC to support S3 storage.)
 
-DVC supports the concept of a _default remote_. For the commands that accept a
```
shcheklein, line 50: It looks quite important, no?

jorgeorpinel: Hmmm yeah I didn't mean to just remove this. I thought it was rewritten rather. Let me check, maybe a bad stash/unstash operation on my side... ⌛

content/docs/command-reference/remote/add.md
```diff
@@ -47,13 +47,6 @@ DVC will determine the [type of remote](#supported-storage-types) based on the
...
-`--remote` option (`dvc pull`, `dvc push`, `dvc status`, `dvc gc`, `dvc fetch`),
-the default remote is used if that option is not used.
```
jorgeorpinel, line 52: OK this part I'm not sure I should've removed indeed... ⌛

content/docs/command-reference/remote/add.md
```diff
@@ -47,13 +47,6 @@ DVC will determine the [type of remote](#supported-storage-types) based on the
...
-
-Use `dvc config` to unset/change the default remote as so:
-`dvc config -u core.remote`.
```
jorgeorpinel, line 55: This was just moved down into the -d option.

content/docs/command-reference/remote/add.md
jorgeorpinel: I think it should go in the remote cmd ref index though. I rewrote that text and included this note. See e8713864 please @shcheklein.

shcheklein: I think it's import to keep it here (not sure about index). Moreover I would even move up and highlight. Note: if you add the first remote you most likely want to use `-d` option as well. 

It's important since this is most likely the command people would to use.

jorgeorpinel: I agree. I was just assuming people would spot to read the index first but now that you mention it, yeah probably a lot of people will only skim over it the first time or skip it altogether on their way to the `remote add` ref.

OK, I rolled back some of the changes to the index and emphasized on the default remote in add. See https://github.com/iterative/dvc.org/compare/ab13a8c..e657efd for a summary of the changes related to this matter.

jorgeorpinel: I guess since this was the only concern and I've addressed it, I should merge this now 🎉 Thanks

PR #1729, opened 2020-08-27
Squash a few warnings and errors

### LINKED ISSUE #1727: warnings and errors in prod deployment
E.g.

```
warning code block language not specified in markdown. applying generic code block

lscpu: failed to determine number of CPUs: /sys/devices/system/cpu/possible: No such file or directory

warning Query takes too long: ...  # Several of these

error Command failed with exit code 1. ...
Error: Command failed: yarn build ...
Retrying with a cleared cache:

error Building static HTML failed for path "/community"

warning Deprecation warning - adding inferred resolver for field CommunityEvent.date. In Gatsby v3, only fields with an explicit directive/extension will get a resolver.

warning There are expired Nodes in community.json!
```

> Taken from the latest deploy log, https://dashboard.heroku.com/apps/dvc-org/activity/builds/598f8b63-09f9-4ce9-aa37-f7545a8d8a45

rogermparent:
- Remove expired events (which is all of them right now)
- Log expired event titles
- Add a `text` prismjs extension to suppress warnings on generic blocks
- Add explicit date format to event date fields

This cleans up many of the errors pointed out in #1727, but the two left are the most important as well as the most difficult.

- A cache issue related to the Community page schema that causes builds to occasionally fail in production. It's hard to pin down, but I think it can be re-created and addressed with improvements to the event schema.

- Repeated `warning: Query takes too long`s on index pages because they generate a **lot** of images, making their pages vastly stand out in build time compared to everything else.

We could either make issues for each to track them individually, or leave #1727 open until they're closed.

UPDATE: Closes #1727


jorgeorpinel:
Build log looks better for sure: https://dashboard.heroku.com/apps/dvc-landing-squash-erro-btioy4/activity/builds/4f3ec23f-b66d-4533-9ed1-36be99cc0043

content/community.json
```diff
@@ -69,24 +69,7 @@
...
-      "date": "2020-06-09",
-      "pictureUrl": "/img/community/events/WiDS_logo.jpg"
-    },
-    {
-      "url": "https://mlopsworld.com/",
-      "title": "MLOps: Production and Engineering World",
-      "description": "Elle O'Brien is speaking about adapting CI/CD to ML.",
-      "city": "Virtual event",
-      "date": "2020-06-17",
-      "pictureUrl": "/img/community/events/toronto_ml.jpg"
```
shcheklein, line 87: remove images if they are not used anymore? Should we just keep one of the events (or all of them) as an example? I'm not sure we should even log warnings on expired events to be honest. It's an intended behavior

gatsby-config.js
```diff
@@ -58,7 +58,13 @@ const plugins = [
         {
           resolve: 'gatsby-remark-prismjs',
           options: {
-            noInlineHighlight: true
+            noInlineHighlight: true,
+            languageExtensions: [
+              {
+                language: 'text',
```
shcheklein, line 64: on what specific blocks do we get this error? should they be `dvc`?

shcheklein:
Yeah, the most concerning is that build takes long times and we still don't cache images properly (?) . I would create a separate one for this.

gatsby-config.js
rogermparent: Here's a few I found:

docs/command-reference/install.md

    ## Disable Git hooks
    
    When you use `dvc install`, it creates three files under the `.git/hooks`
    directory:
    
    ```
    .git/hooks
    ├── post-checkout
    ├── pre-commit
    └── pre-push
    ```

content/docs/command-reference/repro.md

    ## Examples
    
    For simplicity, let's build a pipeline defined below. (If you want get your
    hands-on something more real, see this short
    [pipeline tutorial](/doc/start/data-pipelines)). It takes this `text.txt` file:
    
    ```
    dvc
    1231
    is
    3
    the
    best
    ```

content/docs/command-reference/pull.md

    ## Description
    
    [...]
    
    It has the same effect as running `dvc fetch` and `dvc checkout`:
    
    ```
    Controlled files             Commands
    ---------------- ---------------------------------
    
    remote storage
         +
         |         +------------+
         | - - - - | dvc fetch  | ++
         v         +------------+   +   +----------+
    project's cache                  ++ | dvc pull |
         +         +------------+   +   +----------+
         | - - - - |dvc checkout| ++
         |         +------------+
         v
     workspace
    ```

From what I can tell, each usage is for visual examples made in monospace ASCII and text files that really don't map to a language.

content/community.json
rogermparent: Good point on the images. If we don't want the logging behavior though, I can remove it.

rogermparent:
> Yeah, the most concerning is that build takes long times and we still don't cache images properly (?) . I would create a separate one for this.

Sure! This'll let me make another branch to diagnose this and catch a failed build's cache without worrying about outside variables.

I'm sure image caching is still working, as if they weren't cached the build would be *way* longer than even when this bug happens. I think the queries that are warned about for taking too long are taking a while because they're fetching a huge amount of data even when the image transforms are cached. 
For example, take an index page has 8 thumbnails with our current setup that gives each thumbnail two sourcesets with 4 images: that means `gatsby-transformer-sharp`, even with working cache, reads the FS cache to provide 64 image paths along with any extra blur-up data for each image. While this certainly isn't long as it takes when the images are actually resized, it's still much more intensive than the average query and gets flagged by Gatsby with the extra warnings that were introduced a bit ago.

content/community.json
shcheklein: @rogermparent up to you to decide on this :)

gatsby-config.js
shcheklein: yeah, I see - they are indeed generic. Thanks @rogermparent !

shcheklein:
>  reads the cache and the FS to provide 64 image paths along with any extra blur-up data for each image

hmm, it does not sound scary at all ... it should be within milliseconds to read them I guess

shcheklein:
> I'd say the cache issue should come first

Up to you on this - I don't have enough context to make a call.

I would mention though that if it's only about the community page then we don't change it often (once a month at the very best I guess).

rogermparent:
> > reads the cache and the FS to provide 64 image paths along with any extra blur-up data for each image
> 
> hmm, it does not sound scary at all ... it should be within milliseconds to read them I guess

You're right, the problem isn't very scary with cache, it just takes comparatively much longer than the average query that grabs some basic page content from a markdown file that's already been read to memory. That's why the warning only occurs in the Heroku containers and not local builds- the difference in computing power is enough to drive the time below Gatsby's warning threshold. The warning is much scarier than the thing it's warning about.

rogermparent:
I made #1733 and #1732 to track the remaining two issues, so I think we're good to merge and close here.

shcheklein:
thanks @rogermparent !

PR #1735, opened 2020-08-28
guide: external data updates

jorgeorpinel:
and other related updates

Issue #1614, opened 2020-07-22
links: automatic command links are too lax now?


jorgeorpinel:
Should `dvc {cmd} {arg}` always be linked? E.g.:

![image](https://user-images.githubusercontent.com/1477535/88220850-2b46b100-cc29-11ea-9eef-53c882f10025.png)
> From https://dvc.org/doc/command-reference/repro#examples

I have the impression this was not the case before. It used to only do that for commands with subcommands, and perhaps some valid options of `dvc config`.

jorgeorpinel:
Cc @rogermparent (let's see what Ivan says though).

rogermparent:
I don't know when this was introduced, but I'm pretty sure it's not the `simpleLinker` I added. Its current settings only perform exact string matches. This is looks like behavior from `commandLinker` that theoretically has existed for a while.

I can change the behavior now that I'm familiar with the mechanism, but I'll need some specific rules.

You can see in the source that `commandLinker` still links commands even if the subcommand isn't valid.

```
if (isSubcommandPageExists) {
  url = `${COMMAND_ROOT}${parts[1]}/${parts[2]}`
} else if (isCommandPageExists && hasThirdSegment) {
  url = `${COMMAND_ROOT}${parts[1]}#${parts[2]}`
} else if (isCommandPageExists) {
  url = `${COMMAND_ROOT}${parts[1]}`
}
```

shcheklein:
Yeah, I think that was always the case.

Do we want this? I would keep it. May be it should be smart enough to allow only proper third level sub commands and `dvc dag <tag>` should link to the `dag`, not to `dag#tag`

rogermparent:
> Do we want this? I would keep it. May be it should be smart enough to allow only proper third level sub commands and `dvc dag <tag>` should link to the `dag`, not to `dag#tag`

This is the current behavior as `hasThirdSegment` is only true when the third segment matches a regex determining that it's a proper command. Here's more context around that same snippet:
```
const DVC_REGEXP = /dvc\s+[a-z][a-z-.]*/
const COMMAND_REGEXP = /^[a-z][a-z-]*$/
const COMMAND_ROOT = '/doc/command-reference/'

module.exports = astNode => {
  const node = astNode[0]
  const parent = astNode[2]

  if (parent.type !== 'link' && DVC_REGEXP.test(node.value)) {
    const parts = node.value.split(/\s+/)
    let url

[...]

PR #1736, opened 2020-08-28
Fix details triggers that contain markup

rogermparent:
Before, as pointed out [here](https://github.com/iterative/dvc.org/pull/1735#discussion_r479549491) by @jorgeorpinel, any markup like code quotes would break a header as only the first child
had its text extracted. Now, the `Collapsible` uses the whole `children` value of the
header it's made from as the `trigger` text, so anything that can go in a heading can go in a `details` `trigger`.

`simpleLinker` is also changed to not apply to code quotes within headings, partially because of this and also because auto-links in headers feels like it clutters the UX a bit since the same contents will very likely be linked in the following paragraph. I can re-enable this behavior if we find it desirable.

### DIFF
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -1,4 +1,10 @@
-import React, { useCallback, useEffect, useRef } from 'react'
+import React, {
+  useCallback,
+  useEffect,
+  useRef,
+  ReactNode,
+  ReactElement
+} from 'react'
 import cn from 'classnames'
 import { navigate } from '@reach/router'
 import rehypeReact from 'rehype-react'
@@ -12,6 +18,8 @@ import { getPathWithSource } from '../../../utils/shared/sidebar'
 import 'github-markdown-css/github-markdown.css'
 import sharedStyles from '../styles.module.css'
 import styles from './styles.module.css'
+
+import util from 'util'
 
 const isInsideCodeBlock = (node: Element): boolean => {
   while (node?.parentNode) {
@@ -30,17 +38,31 @@ const isInsideCodeBlock = (node: Element): boolean => {
 }
 
 const Details: React.FC<{
-  children: Array<{ props: { children: Array<string> } } | string>
+  children: Array<{ props: { children: ReactNode } } | string>
 }> = ({ children }) => {
   const filteredChildren = children.filter(child => child !== '\n')
 
   if (!filteredChildren.length) return null
   if (typeof filteredChildren[0] === 'string') return null
 
-  const text = filteredChildren[0].props.children[0]
+  const headingChildren: ReactNode[] = filteredChildren[0].props
+    .children as ReactNode[]
+
+  // Remove header auto-link if present
+  const finalHeadingChild = headingChildren[headingChildren.length - 1]! as {
+    props: {
+      className: string
+    }
+  }
+
+  const triggerChildren: unknown =
+    finalHeadingChild.props !== undefined &&
+    finalHeadingChild.props.className === 'anchor after'
+      ? headingChildren.slice(0, headingChildren.length - 1)
+      : headingChildren
 
   return (
-    <Collapsible trigger={text} transitionTime={200}>
+    <Collapsible trigger={triggerChildren as ReactElement} transitionTime={200}>
       {filteredChildren.slice(1)}
     </Collapsible>
   )
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -51,8 +51,8 @@ const Details: React.FC<{
      which we currently have as an external package.
    */
   const triggerChildren: ReactNode[] = firstChild.props.children.slice(
-    1,
-    firstChild.props.children.length
+    0,
+    firstChild.props.children.length - 1
   ) as ReactNode[]
 
   /*
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -19,8 +19,6 @@ import 'github-markdown-css/github-markdown.css'
 import sharedStyles from '../styles.module.css'
 import styles from './styles.module.css'
 
-import util from 'util'
-
 const isInsideCodeBlock = (node: Element): boolean => {
   while (node?.parentNode) {
     if (node.tagName === 'PRE') {
@@ -49,7 +47,7 @@ const Details: React.FC<{
     .children as ReactNode[]
 
   // Remove header auto-link if present
-  const finalHeadingChild = headingChildren[headingChildren.length - 1]! as {
+  const finalHeadingChild = headingChildren[headingChildren.length - 1] as {
     props: {
       className: string
     }
--- a/src/components/Documentation/Markdown/index.tsx
+++ b/src/components/Documentation/Markdown/index.tsx
@@ -41,16 +41,19 @@ const Details: React.FC<{
   const filteredChildren: ReactNode[] = children.filter(child => child !== '\n')
   const firstChild = filteredChildren[0] as JSX.Element
 
-  const triggerChildren: ReactNode[] = firstChild.props.children as ReactNode[]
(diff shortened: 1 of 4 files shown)


src/components/Documentation/Markdown/index.tsx
```diff
@@ -30,17 +36,31 @@ const isInsideCodeBlock = (node: Element): boolean => {
...
+    .children as ReactNode[]
+
+  // Remove header auto-link if present
+  const finalHeadingChild = headingChildren[headingChildren.length - 1] as {
+    props: {
+      className: string
+    }
+  }
+
+  const triggerChildren: unknown =
```
shcheklein, line 56: is it actually `unknown`?

src/components/Documentation/Markdown/index.tsx
```diff
@@ -30,17 +36,31 @@ const isInsideCodeBlock = (node: Element): boolean => {
...
 
   if (!filteredChildren.length) return null
   if (typeof filteredChildren[0] === 'string') return null
 
-  const text = filteredChildren[0].props.children[0]
+  const headingChildren: ReactNode[] = filteredChildren[0].props
+    .children as ReactNode[]
+
+  // Remove header auto-link if present
+  const finalHeadingChild = headingChildren[headingChildren.length - 1] as {
```
shcheklein, line 50: not sure why do we filter only the last node - could you clarify how does it work please?

rogermparent: Sure! It's kind of weird and a workaround to adapt this fix to our ecosystem.

Basically, this bit removes the link icon that `gatsby-remark-autolink-headers` adds on the headings that eventually become the trigger for a `details` box. Since `details` is a component and the autolinker is a Remark plugin, there's no way to prevent the headers from getting an auto-link from within our codebase.

To prevent our `details` component from having to do this, we'd need to change how the header auto-links work by forking the auto-link plugin or rebuilding the functionality ourselves, which is totally doable but obviously more intensive than with this quick fix.

The reason this wasn't an issue before was that the previous behavior sliced off all but the first element, which is very similar to what this is doing but with the caveat of markup causing the break that this fix is for.

Now that I think about it though, if all `details` blocks are started with headers that are always auto-linked, this could just be a simple `slice` that removes the last item each time and skip the ternary.

src/components/Documentation/Markdown/index.tsx
rogermparent: Kind of. It should be a `ReactNode`, but since `Collapsible`'s interface for the `trigger` prop only accepts `text` or `ReactElement` despite `ReactNode`s being a viable input.

Because of this, I TS casted `triggerChildren` to `unknown` to "force" a cast from `ReactNode` to `ReactElement` at the suggestion of the compiler.
Maybe making this a `ReactNode` and doing `as unknown as ReactComponent` would be a little more expressive, so readers know exactly what we're doing.

jorgeorpinel:
What parts of #1614 does this address? 🙂

rogermparent:
> What parts of #1614 does this address? slightly_smiling_face

Ah, I see there was a discussion about linked headers further down about this in the comments. I hadn't looked at #1614 for a while until we discovered the bugged interaction with markup in `details` headers.

src/components/Documentation/Markdown/index.tsx
shcheklein: can we do `ReactNode` here, and  `as ReactElement ` below? (I'm not very familiar with TS rules tbh, especially when it comes to TS being passed to JS + PropTypes, just curious how this is handled usually). Btw, does TS "understand" propTypes?

src/components/Documentation/Markdown/index.tsx
```diff
@@ -30,17 +36,31 @@ const isInsideCodeBlock = (node: Element): boolean => {
...
+    props: {
+      className: string
+    }
+  }
+
+  const triggerChildren: unknown =
+    finalHeadingChild.props !== undefined &&
```
shcheklein, line 57: how can it be undefined btw?

src/components/Documentation/Markdown/index.tsx
shcheklein: > Now that I think about it though, if all details blocks are started with headers that are always auto-linked, this could just be a simple slice that removes the last item each time and skip the ternary.

 May be there should be some check in place that check `filteredChildren[0]` type properly (not only `=== string`)? Otherwise looks like the main method expects it to be the header.

shcheklein:
> simpleLinker is also changed to not apply to code quotes within headings, partially because of this and also because auto-links in headers feels like it clutters the UI

it's fine to keep it disabled, looks like no one likes it

src/components/Documentation/Markdown/index.tsx
```diff
@@ -30,17 +36,31 @@ const isInsideCodeBlock = (node: Element): boolean => {
...
   const filteredChildren = children.filter(child => child !== '\n')
 
   if (!filteredChildren.length) return null
   if (typeof filteredChildren[0] === 'string') return null
 
-  const text = filteredChildren[0].props.children[0]
+  const headingChildren: ReactNode[] = filteredChildren[0].props
+    .children as ReactNode[]
+
+  // Remove header auto-link if present
```
shcheklein, line 49: we def need to improve the comment. Very hard to read and understand this low level code that depends on some external presence of the plugin that does auto-link - it's very non-obvious

src/components/Documentation/Markdown/index.tsx
rogermparent: Because of a quirk of how this conversion works with raw children, the closest we can get is `ReactNode` here, and `as unknown as ReactElement` below. Since `ReactNode`s don't cleanly translate into `ReactComponent`, the variable has to pass through `unknown` at some point to let the compiler know it isn't a mistake.

src/components/Documentation/Markdown/index.tsx
rogermparent: Yeah, `filteredChildren[0]` is expected to be the MD heading in the `details` that becomes the `Collapsable` trigger. I can make a change to more specifically make the details box only when the first element is a heading. I'm not sure if `details` is used in any other way, but if not we could either throw or provide a default `trigger` text for `details` that don't lead with a heading.

src/components/Documentation/Markdown/index.tsx
rogermparent: Plaintext children in React children are represented by strings which don't have a `props` like component children.

src/components/Documentation/Markdown/index.tsx
shcheklein: 👍

rogermparent:
All the feedback has been addressed, check out a live example of this working [on the preview](https://dvc-landing-fix-details-t4ihvr.herokuapp.com/doc/user-guide/external-dependencies#example-import-url-command)!

The behavior now throws if the `details` tag isn't led with a heading like we do everywhere now. This is one of a few possible behaviors we could implement for this situation, but I'm assuming that we want to keep `details` tag invocations relatively uniform and always topped with a heading.

src/components/Documentation/Markdown/index.tsx
rogermparent: I took your advice and now we check that the first child is a heading, regardless of its level. Otherwise, the component throws and is caught at build-time with an error message informing the user a heading must be used.

While the throw behavior is only for source readability, I think we want to keep `details` blocks uniform in the sense that they're always started with a heading instead of accepting any first as `trigger` and letting them become inconsistent.

src/gatsby/models/github/index.js
```diff
@@ -7,34 +7,6 @@
    alone does.
  */
 
-async function createStaticGithubDataNode(api, fieldData = {}) {
```
shcheklein, line 10: is it unused now?

rogermparent: Yeah, it gets picked up by `lint-ts`.

shcheklein: hmm ... why didn't it pick it up earlier?

PR #1737, opened 2020-08-29
cases: update and shorten Data Registries

jorgeorpinel:
> - [x] ~~Change base branch to `master` when #1716 is merged.~~

Mostly 1.x updates and copy edits. Similar changes to those in #1716 (for consistency among use cases. Also, addresses #820.

UPDATE: Jump to https://github.com/iterative/dvc.org/pull/1737#issuecomment-684120533


jorgeorpinel:
Can't merge before #1716 due to weird branching, sorry... But you could review this if you have the time, @shcheklein 🙂 — see https://dvc-landing-use-cases-d-mojnxd.herokuapp.com/doc/use-cases/data-registries

jorgeorpinel:
p.s. we could even merge this as is (to the `use-cases` branch).

content/docs/use-cases/data-registries.md
```diff
@@ -2,22 +2,21 @@
 
 One of the main uses of <abbr>DVC repositories</abbr> is the
 [versioning of data and model files](/doc/use-cases/data-and-model-files-versioning),
-with commands such as `dvc add`. With the aim to enable reusability of these
-<abbr>data artifacts</abbr> between different projects, DVC also provides
-commands like `dvc import` and `dvc get`. This means that you projects can
-depend on data from other DVC repositories, **similar to a package management
-systems for data science**.
+with tools like the `dvc add` command and `dvc.yaml` files. DVC also enables
```
shcheklein, line 5: let's just avoid mentioning specific commands and files in the first paragraph ?

shcheklein: also, we didn't mention `dvc run` before

content/docs/use-cases/data-registries.md
```diff
@@ -41,56 +40,52 @@ _middleware_ between ML projects and cloud storage. Here are it's advantages:
...
-`git init` and `dvc init`. A good way to organize their contents is by using
-different directories to group similar data, e.g. `images/`,
-`natural-language/`, etc. For example, our
-[dataset registry](https://github.com/iterative/dataset-registry) uses a
-directory for each part in our docs like `get-started/` and `use-cases/`.
-
 Adding datasets to a registry can be as simple as placing the data file or
-directory in question inside the <abbr>workspace</abbr>, and telling DVC to
-track it, with `dvc add`. For example:
+directory in question inside the <abbr>workspace</abbr>, and track it with
```
shcheklein, line 44: I think the previous version was pretty good ... at least I don't see the current one superior? what is the reason changing this?

content/docs/use-cases/data-registries.md
jorgeorpinel: OK, removed specific commands and dvc.yaml mentions.

content/docs/use-cases/data-registries.md
jorgeorpinel: I compressed 2 paragraphs and 2 code blocks into a single pair.

jorgeorpinel:
We want to change this later?

> what was the primary reason to do this one update?

I'm reviewing all the use cases as agreed. I rewrote Versioning in #1716 using this one (Data Registry) partly as example, but ended up summarizing there much more. Also we agreed that this one was also still too long, so I simplifies this one as well.

> can we scope and simplify it?

It's already extracted from #1716 and only 1 file changed. Maybe when you looked it was more files due to unmerged changes in the base branch.

shcheklein:
> Also we agreed that this one was also still too long

I don't think this is the problem. The problem might be that it is still too low level.

> It's already extracted from #1716 and only 1 file changed.

yeah, when I talk about the scope I mostly mean that to me this PR feels that we just polishing and existing doc, not changing it. And that's why the question comes to my mind- do we need this kind of polishing, what is the reason behind it? May be I'm missing something - that's why I asked - is there something specific in this doc I should review, or is it just refactoring w/o changing the meaning and how doc feels?

jorgeorpinel:
Gotcha. Nah, it's just polishing. I think we agree this is the better-written use case already. As I was re-reading it for "inspiration" when working on the other ones, I found things that could be summarized/omitted, better explained, reformatted, etc. That's what this is. Maybe took 20 min. total

It looks like a huge change from the Github diff because of our paragraph formatting but comparing https://dvc-landing-use-cases-d-mojnxd.herokuapp.com/doc/use-cases/data-registries vs https://dvc.org/doc/use-cases/data-registries may be an easier and faster way to quickly review visually.

content/docs/use-cases/data-registries.md
```diff
@@ -1,92 +1,87 @@
...
   chances someone can delete or rewrite an ML model, for example.
-- Storage optimization: centralize data
+- **Storage optimization**: centralize data
   [shared](/doc/use-cases/sharing-data-and-model-files) by multiple projects in
   a single location (distributed copies are possible too). This simplifies data
   management and optimizes space requirements.
-- Data as code: leverage Git workflows such as commits, branching, pull
+- **Data as code**: leverage Git workflows such as commits, branching, pull
   requests, reviews, and even CI/CD for your data and models lifecycle. Think
   "Git for cloud storage", but without ad-hoc conventions.
```
jorgeorpinel, line 33: Should it be "Codification" like in the upcoming Versioning use case?

content/docs/sidebar.json
```diff
@@ -74,8 +74,8 @@
         "label": "Sharing Data and Model Files",
         "slug": "sharing-data-and-model-files"
       },
-      "shared-development-server",
-      "data-registries"
+      "data-registries",
+      "shared-development-server"
```
jorgeorpinel, line 78: I feel like Data Registries has become quite popular so the shared dev env seems like more of an edge case now?

shcheklein: 👍

content/docs/use-cases/data-registries.md
shcheklein: I'd like to keep `data as code` as well (to coin the term)

also Codification is a general term (applies to anything, not only data) - you will need to clarify the meaning in this case, means will have to rewrite it quite a bit (which is fine, but we don't have time for this)

unrelated: `but without ad-hoc conventions.` sounds a bit strange- I guess it's not easy to understand

shcheklein (approved):

content/docs/use-cases/data-registries.md
jorgeorpinel: > keep data as code as well (to coin the term

So should we put it in the Versioning use case too?

> without ad-hoc conventions. sounds a bit strange

It does come out of nowhere. Should I remove it for now? I think we were referring to file names and that is explained in the Versioning one now.

PR #1738, opened 2020-08-30
cmd: move 1.x updates

dmpetrov:
- Update `move` ref.
- Review it's examples carefully
- Add note and example about renaming stages

### DIFF
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -19,7 +19,7 @@ positional arguments:
 `dvc move` is useful when a `src` file or directory has previously been added to
 the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
 (with `src` as a dependency). `dvc move` doesn't support renaming stage
-<abbr>outputs</abbr> (see `dvc run`), they have to be renamed manually.
+<abbr>outputs</abbr> (see `dvc.yaml`), they have to be renamed manually.
 
 > Note that `src` may be either a
 > [link](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -169,7 +169,7 @@ $ tree
 ## Example: manually renaming stage outputs
 
 `dvc move` doesn't support renaming of <abbr>outputs</abbr> in a stage, you will
-have to do that manually. Let's look at our sample workspace and `dvc.yaml:
+have to do that manually. Let's look at our sample workspace and `dvc.yaml`:
 
 ```dvc
 $ tree
@@ -195,15 +195,16 @@ stages:
 ```
 
 In this example we have to rename `keras.h5` as `model.h5`. First change the
-name of model file in code (`train.py`). Next change output name in train stage
-of `dvc.yaml`. After this change `/keras.h5` to `/model.h5` in `.gitignore`
-file. Lastly, rename the existing model file:
+name of model file in code (`train.py`). Next change output name in `train`
+stage of `dvc.yaml`. After this change `/keras.h5` to `/model.h5` in
+`.gitignore` file. Lastly, rename the existing model file:
 
 ```dvc
 $ mv keras.h5 model.h5
 ```
 
-In final step run `dvc commit` to save the changes and update the lock file:
+In final step run `dvc commit` to save the changes and update the `dvc.lock`
+file:
 
 ```dvc
 $ dvc commit -f
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -202,5 +202,10 @@ $ git status -s
  M .gitignore
  M dvc.yaml
  M train.py
-$ dvc repro
+$ dvc commit --force # save the changes and update the lock file
+$ git status -s
+ M .gitignore
+ M dvc.lock
+ M dvc.yaml
+ M train.py
 ```
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -4,6 +4,10 @@ Rename a file or a directory and modify the corresponding `.dvc` file (see
 `dvc add`) to reflect the change. If the file or directory has the same name as
 the `.dvc` file, it also renames it.
 
+Renaming is supported only for data source files and directories (creaded by
+`dvc add` and `dvc import`), not for outputs of pipelines (`dvc run`). The
+pipeline outputs can be renamed only manualy.
+
 ## Synopsis
 
 ```usage
@@ -160,4 +164,43 @@ $ tree
     │   ├── bar
     │   └── foo
     └── data3.dvc
+```
+
+## Example: manual renaming of a pipeline output
+
+The command does not support renaming of pipeline outputs. This example shows
+how the renaming can be done manualy. An alternative approach - remove an entire
+stage and recreate it again.
+
+```dvc
+$ tree
+.
+├── dvc.lock
+├── dvc.yaml
+├── keras.h5
+├── params.yaml
+├── train.py
+├── users.csv
+└── users.csv.dvc
+$ vi train.py # change model file name in code to model.h5
+$ cat dvc.yaml
+stages:
+  train:
+    cmd: python train.py
+    deps:
+    - users.csv
+    - train.py
+    params:
+    - epochs
+    - opt
+    outs:
+    - keras.h5
+$ vi dvc.yaml # change output name keras.h5 to model.h5
+$ vi .gitignore # change output name /keras.h5 to /model.h5
+$ mv keras.h5 model.h6
+$ git status -s
+ M .gitignore
+ M dvc.yaml
+ M train.py
+$ dvc repro
 ```
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -28,7 +28,7 @@ the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
 
 `dvc move` behaves like `mv src dst`, moving `src` to the given `dst` path, but
 it also renames and updates the corresponding `.dvc` file appropriately. If the
-destination path (`dst`) already exists and is a directory, the source code file
+destination path (`dst`) is a directory and already exists, the source code file
 or directory (`src`) is moved unchanged into this folder along with the
 corresponding `.dvc` file.
 
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -19,7 +19,8 @@ positional arguments:
 `dvc move` is useful when a `src` file or directory has previously been added to
 the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
 (with `src` as a dependency). `dvc move` doesn't support renaming stage
-<abbr>outputs</abbr> (see `dvc.yaml`), they have to be renamed manually.
+<abbr>outputs</abbr> (see `dvc.yaml`), they have to be
+[renamed manually](doc/user-guide/dvc-files-and-directories#renaming-stage-outputs).
 
 > Note that `src` may be either a
 > [link](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
@@ -28,9 +29,8 @@ the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
 
 `dvc move` behaves like `mv src dst`, moving `src` to the given `dst` path, but
 it also renames and updates the corresponding `.dvc` file appropriately. If the
-destination path (`dst`) is a directory and already exists, the source code file
-or directory (`src`) is moved unchanged into this folder along with the
-corresponding `.dvc` file.
+destination path is a directory and already exists, the source file or directory
+is moved into this folder along with the corresponding `.dvc` file.
 
 Let's imagine the following scenario:
 
@@ -38,8 +38,8 @@ Let's imagine the following scenario:
 $ dvc add data.csv
 ```
 
-The `dvc add` command would create a `.dvc` file named as `data.csv.dvc` with
-the following content:
+The `dvc add` command would create a `.dvc` file named `data.csv.dvc` with the
+following content:
 
 ```yaml
 outs:
@@ -73,42 +73,31 @@ outs:
 
 ### Manually renaming stage outputs
 
-In this example, let's rename stage output `keras.h5` to `model.h5`:
+Let's take the following scenario and rename stage output `keras.h5` to
+`model.h5`:
 
-```dvc
+```yaml
 $ cat dvc.yaml
 stages:
   train:
-    cmd: python train.py
     ...
     outs:
     - keras.h5
 ```
 
-First we need to change the name of model file in code (`train.py`). Next we
-change the output name in the `train` stage of `dvc.yaml`. After this we change
-`/keras.h5` to `/model.h5` in the `.gitignore` file. Lastly, we rename the
-existing model file:
+First, we change the name of the model file in code (`train.py`). Next, change
+the output name in the `train` stage of `dvc.yaml` and update `/keras.h5` to
+`/model.h5` in `.gitignore` file. In the next step, we rename the existing model
+file:
 
-```dvc
+```yaml
 $ mv keras.h5 model.h5
 ```
 
-In the final step we run `dvc commit` with `-f` option to force save the
-changes:
+Finally, we run `dvc commit` with the `-f` option to force save the changes:
 
 ```dvc
 $ dvc commit -f
-```
-
-Alternatively, we can also use `dvc run` with `-f` to overwrite the stage
-changes in `dvc.yaml`:
-
-```dvc
-$ dvc run -f -n train \
-          -d data.csv -d train.py -o model.h5 \
-          -p train.epochs,train.lr \
-          python train.py
 ```
 
 ## Options
@@ -144,9 +133,9 @@ $ tree
 ## Example: change the location
 
 We use `dvc add` to track a file with DVC, then we use `dvc move` to change its
-location. If the target path already exists and is a directory, data file is
-moved with unchanged name into this folder. Note that the corresponding
-`.dvc` file, `data.csv.dvc` is also moved.
+location. If the target path is a directory and already exists, the data file is
+moved with unchanged name into this folder. Note that the corresponding `.dvc`
+file, `data.csv.dvc` is also moved.
 
 ```dvc
 $ tree
--- a/content/docs/command-reference/move.md
+++ b/content/docs/command-reference/move.md
@@ -17,22 +17,20 @@ positional arguments:
 ## Description
 
 `dvc move` is useful when a `src` file or directory has previously been added to
-the <abbr>project</abbr> with `dvc add`, creating a `.dvc` file (with `src` as a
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/move.md
```diff
@@ -4,6 +4,10 @@ Rename a file or a directory and modify the corresponding `.dvc` file (see
 `dvc add`) to reflect the change. If the file or directory has the same name as
 the `.dvc` file, it also renames it.
 
+Renaming is supported only for data source files and directories (creaded by
+`dvc add` and `dvc import`), not for outputs of pipelines (`dvc run`). The
+pipeline outputs can be renamed only manualy.
+
```
jorgeorpinel, line 10: Great info. We should move it down to the description though, the intro is supposed to be as short and on-point as possible.

jorgeorpinel: Actually this whole ref is still outdated (pre 1.x) unfortunately. dvc.yaml or dvc run are not even mentioned yet so this text is a bit out of context... I.e. you can just leave the PR there and we'll reuse it to update the whole ref. it may just take a few days to get to this.

jorgeorpinel: Cc @imhardikj probably a good one to start for you 1.x updates. It should be simple enough and let us practice this side of your project.

jorgeorpinel:
Hi @imhardikj. We need to carefully read and rewrite the whole text as needed. For example:

- `Note that `src` may be a copy or a link...`
  should be
  `Note that `src` may be a either a link or a copy to the corresponding data in the cache. The cached...`

- `already exists and is a directory`
  shouold be
  `is a directory and already exists`

content/docs/command-reference/move.md
```diff
@@ -30,6 +30,10 @@ If the destination path (`dst`) already exists and is a directory, the source
 code file or directory (`src`) is moved unchanged into this folder along with
 the corresponding `.dvc` file.
 
+`dvc move` doesn't support renaming outputs of pipeline stages (`dvc run`), it
+only renames data source files and directories (added using `dvc add` and
+`dvc import`). The pipeline outputs have to be renamed manually.
```
jorgeorpinel, line 35: Some of the changes you made to Dmitry's text here didn't improve it:

- He started with "data source files and directories" (generic concept). You start with "outputs" (DVC jargon). What is that? New users won't know.
  > We have a glossary entry for that term BTW, so please use `<abbr>outputs</abbr>` for the first instance.

- He uses "outputs of pipelines", you appended "stages" to the phrase, which is redundant. All pipelines have stages. Most of the times we only need to mention one or the other. I would actually go for "stage outputs (**see** `dvc run`)" (I reserve the term "pipeline" for actual documents about pipelines).

- You correctly identified the phrase "creaded by `dvc add` and `dvc import`" and improved it to "added using `dvc add` and `dvc import`). But it's still not perfect: "Added using add"? Sounds strange. Also, is that parenthesis needed?

- Does the last sentence need "The"? We're not talking about any specific pipeline outputs (I'd also change that to "Stage outputs BTW".

jorgeorpinel: Anyway, I'd rather not rewrite every sentence but for now during bonding I'll be giving you these specific tips as writing training. I will leave more general feedback progressively 🙂

jorgeorpinel: All that said... Really this information is super important and should be part of the very first paragraph. Please rewrite the first paragraphs or 2 to make extra clear that `move` is meant for .dvc files only. You don't have to keep the existing text, it can be totally different, as long as it's correct, complete, and as brief as possible. Thanks

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
     │   └── foo
     └── data3.dvc
 ```
+
+## Example: manually renaming stage outputs
+
+`dvc move` doesn't support renaming of <abbr>outputs</abbr> in a stage, you will
+have to do that manually. Let's look at our sample workspace and `dvc.yaml`:
```
jorgeorpinel, line 172: I don't think this qualifies as an example, as `dvc move` is not being used. It can be a subsection of the description perhaps (H3) but it should also probably be much shorter. Or it could be a section in `dvc run` (about using its `-f` option for renaming) and link to there. And/or a section in the dvc.yaml spec (https://dvc.org/doc/user-guide/dvc-files-and-directories#dvcyaml-file) for the fully manual method + also link. WDYT?

jorgeorpinel: Also writing could improve a little e.g. "renaming of outputs in a stage" is a complicated phrase, not sure... Let's use simple, direct language like "renaming stage outputs". Thanks

> The whole phrase may not be needed though, as if this is moved to the description, it will be repetitive.

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
...
+    - data.csv
+    - train.py
+    params:
+    - epochs
+    - lr
+    outs:
+    - keras.h5
+```
+
+In this example we have to rename `keras.h5` as `model.h5`. First change the
```
jorgeorpinel, line 197: as->to

jorgeorpinel: We start with "we" and change to "you" (implicit) in the next sentence.

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
...
+name of model file in code (`train.py`). Next change output name in `train`
```
jorgeorpinel, line 198: change output name - grammar

jorgeorpinel: in train stage - grammar

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
...
+stage of `dvc.yaml`. After this change `/keras.h5` to `/model.h5` in
```
jorgeorpinel, line 199: in .gitignore file - grammar

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
...
+`.gitignore` file. Lastly, rename the existing model file:
+
+```dvc
+$ mv keras.h5 model.h5
+```
+
+In final step run `dvc commit` to save the changes and update the `dvc.lock`
```
jorgeorpinel, line 206: in final step - grammar

jorgeorpinel: the dvc.lock file -> just "dvc.lock" should be enough.

jorgeorpinel: Although not sure we even want to mention it, as it's not mentioned much less explained anywhere in the doc. Is there a better phrase that explains what `commit` is for here without that technical detail?

jorgeorpinel: May also want to mention why -f

content/docs/command-reference/move.md
```diff
@@ -161,3 +165,57 @@ $ tree
...
+file:
+
+```dvc
+$ dvc commit -f
+```
+
+Alternatively, you can also use `dvc run -f` to overwrite the changes in the
+stage in `dvc.yaml`:
```
jorgeorpinel, line 214: the stage changes (simple/direct language, and avoids using "in" twice in a row).

jorgeorpinel (commented):
As usual good start, with logical updates. But still missing careful reading and understanding of the meaning (and grammar!) in each sentence, and then updating or completely rewriting, accordingly. Don't be afraid to make bigger changes 🙂

content/docs/command-reference/move.md
imhardikj: Thanks for grammar inputs. Updated the doc.

content/docs/command-reference/move.md
imhardikj: I think its better suited in this document as it is related to renaming. Moved into Description as it is not a `move` example. The 'dvc run' part can be example under `dvc run` with a link to example here.

content/docs/command-reference/move.md
imhardikj: changed this to `dvc commit with -f option to force save the changes`. I think this explains minimally without technical details.

content/docs/command-reference/move.md
```diff
@@ -17,27 +17,29 @@ positional arguments:
 ## Description
 
 `dvc move` is useful when a `src` file or directory has previously been added to
-the <abbr>project</abbr> with `dvc add`, creating a `.dvc` file (with `src` as a
-dependency). `dvc move` behaves like `mv src dst`, moving `src` to the given
-`dst` path, but it also renames and updates the corresponding `.dvc` file
-appropriately.
+the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
+(with `src` as a dependency). `dvc move` doesn't support renaming stage
+<abbr>outputs</abbr> (see `dvc run`), they have to be renamed manually.
```
jorgeorpinel, line 22:
```suggestion
<abbr>outputs</abbr> (see `dvc.yaml`), they have to be renamed manually.
```

content/docs/command-reference/move.md
```diff
@@ -17,27 +17,29 @@ positional arguments:
...
-> to a file in cache. The cached file is not changed by this command.
+> or a copy to the corresponding data in the cache. The cached file is not
+> changed by this command.
 
-If the destination path (`dst`) already exists and is a directory, the source
-code file or directory (`src`) is moved unchanged into this folder along with
-the corresponding `.dvc` file.
+`dvc move` behaves like `mv src dst`, moving `src` to the given `dst` path, but
+it also renames and updates the corresponding `.dvc` file appropriately. If the
+destination path (`dst`) already exists and is a directory, the source code file
```
jorgeorpinel, line 31:
```suggestion
destination path (`dst`) is a directory and already exists, the source code file
```

content/docs/command-reference/move.md
```diff
@@ -17,27 +17,29 @@ positional arguments:
...
+> or a copy to the corresponding data in the cache. The cached file is not
+> changed by this command.
 
-If the destination path (`dst`) already exists and is a directory, the source
-code file or directory (`src`) is moved unchanged into this folder along with
-the corresponding `.dvc` file.
+`dvc move` behaves like `mv src dst`, moving `src` to the given `dst` path, but
+it also renames and updates the corresponding `.dvc` file appropriately. If the
+destination path (`dst`) is a directory and already exists, the source code file
+or directory (`src`) is moved unchanged into this folder along with the
```
jorgeorpinel, line 32: Good! But
- the `src` and `dst` mentions have become repetitive.
- We probably don't need "unchanged".

jorgeorpinel: Also "source code"? I think that's a (pre-existing) typo.

content/docs/command-reference/move.md
```diff
@@ -17,27 +17,29 @@ positional arguments:
...
 
 Let's imagine the following scenario:
 
 ```dvc
 $ dvc add data.csv
 ```
 
-The `dvc add` command would create a `data.csv.dvc` `.dvc` file with the
-following content:
+The `dvc add` command would create a `.dvc` file named as `data.csv.dvc` with
```
jorgeorpinel, line 41: "name as" ?

content/docs/command-reference/move.md
```diff
@@ -17,27 +17,29 @@ positional arguments:
 ## Description
 
 `dvc move` is useful when a `src` file or directory has previously been added to
-the <abbr>project</abbr> with `dvc add`, creating a `.dvc` file (with `src` as a
-dependency). `dvc move` behaves like `mv src dst`, moving `src` to the given
-`dst` path, but it also renames and updates the corresponding `.dvc` file
-appropriately.
+the <abbr>project</abbr> with `dvc add` or `dvc import`, creating a `.dvc` file
+(with `src` as a dependency). `dvc move` doesn't support renaming stage
+<abbr>outputs</abbr> (see `dvc.yaml`), they have to be renamed manually.
```
jorgeorpinel, line 22: Link "renamed manually" to that H3 section?

content/docs/command-reference/move.md
```diff
@@ -103,8 +145,8 @@ $ tree
 
 We use `dvc add` to track a file with DVC, then we use `dvc move` to change its
 location. If target path already exists and is a directory, data file is moved
```
jorgeorpinel, line 147: Pre-existing grammar issue:
```suggestion
location. If the target path already exists and is a directory, data file is moved
```
> This may break the formatting, sorry.

content/docs/command-reference/move.md
```diff
@@ -102,9 +144,9 @@ $ tree
 ## Example: change the location
 
 We use `dvc add` to track a file with DVC, then we use `dvc move` to change its
-location. If target path already exists and is a directory, data file is moved
-with unchanged name into this folder. Note that the `data.csv.dvc` `.dvc` file
-is also moved.
+location. If the target path already exists and is a directory, data file is moved
+with unchanged name into this folder. Note that the corresponding `.dvc` file
+`data.csv.dvc` is also moved.
```
jorgeorpinel, line 149:
```suggestion
location. If the target path already exists and is a directory, data file is
moved with unchanged name into this folder. Note that the corresponding
`.dvc` file, `data.csv.dvc` is also moved.
```

content/docs/command-reference/move.md
jorgeorpinel, line 22: Link "renamed manually" to that H3 section?

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
     path: other.csv
 ```
 
+### Manually renaming stage outputs
+
+In this example, let's rename stage output `keras.h5` to `model.h5`:
```
jorgeorpinel, line 76: I thought we agreed it's not an example 🙂

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
...
+
+```dvc
+$ cat dvc.yaml
+stages:
```
jorgeorpinel, line 80: Better to use a YAML code block, no need for `$ cat` here. May need to change the context a little too.

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
...
+  train:
+    cmd: python train.py
+    ...
+    outs:
+    - keras.h5
+```
+
+First we need to change the name of model file in code (`train.py`). Next we
```
jorgeorpinel, line 88: grammar, grammar

jorgeorpinel: But maybe just skip the whole `cmd:` like to simplify this explanation and avoid distractions.

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
...
+change the output name in the `train` stage of `dvc.yaml`. After this we change
+`/keras.h5` to `/model.h5` in the `.gitignore` file. Lastly, we rename the
+existing model file:
```
jorgeorpinel, line 91: Let's not use "lastly" if it's not the last step.

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
...
+
+```dvc
+$ mv keras.h5 model.h5
+```
+
+In the final step we run `dvc commit` with `-f` option to force save the
+changes:
```
jorgeorpinel, line 98: Grammar

jorgeorpinel: Maybe explain a but more what commit -f does. Very briefly

content/docs/command-reference/move.md
```diff
@@ -69,6 +71,46 @@ outs:
...
+
+Alternatively, we can also use `dvc run` with `-f` to overwrite the stage
+changes in `dvc.yaml`:
+
+```dvc
+$ dvc run -f -n train \
+          -d data.csv -d train.py -o model.h5 \
+          -p train.epochs,train.lr \
+          python train.py
+```
```
jorgeorpinel, line 112: Getting too long and this doesn't seem "manual" anymore (meaning not using DVC commands I guess, because everything here is manual in a way... — consider renaming the section if you can think of a better word BTW)

content/docs/command-reference/move.md
jorgeorpinel: I'm not so sure... It's not avout `dvc move` at all really. I think it belongs somewhere else. Probably both in dvc.yaml (the manual move) and in `dvc run` (for ita -f option)

jorgeorpinel (commented):
Review on the manual renaming section ☝️

content/docs/command-reference/move.md
jorgeorpinel: "in the `.gitignore` file" -> "in `.gitignore`"

content/docs/command-reference/move.md
jorgeorpinel: but what does commit do? Why is force needed?

content/docs/command-reference/move.md
jorgeorpinel: You know what, I guess there's no need to explain commit here. Users can click on the link if they're interested.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -226,6 +226,35 @@ directory tracked by DVC. Specifically: `md5`, `etag`, or `checksum` (same as in
 Full <abbr>parameters</abbr> (key and value) are listed separately under
 `params`, grouped by parameters file.
 
+### Renaming stage outputs
+
+Let's take the following scenario and rename stage <abbr>output</abbr>
+`keras.h5` to `model.h5` for the `train` stage:
+
+```yaml
```
imhardikj, line 234: @jorgeorpinel I have modified this based on your inputs and added to this doc as well for comparison. Which doc do you think is better for this section? (I'll remove from the other one). I thought it will look odd in `dvc-files` doc since it only contains explanation and as `dvc move` is related to renaming and this section also is more like a code walkthrough it will be better suited in 'move' doc. Shall I remove it from `move` doc?
___
Also I have removed the `dvc run -f` part since it more complicated. The user will still need to change the name in code and already existing model file (its `.gitignore`  entry) is not deleted.

content/docs/command-reference/move.md
imhardikj: Without the "Alternatively...." part now it's comparatively shorter. Also if it's in `dvc-files-and-dir` doc then the above `$cat`  block might be needed for context.

content/docs/command-reference/move.md
imhardikj: `.dvc` is repeating that's why I changed it.

jorgeorpinel:
OK... This is still very incomplete. I'll take it over this time because it's been open way too long.

jorgeorpinel (changes requested):
@imhardikj please take a look at my last commit, https://github.com/iterative/dvc.org/pull/1738/commits/1fe960bc1600cc1657d6630e0c32011f29c0908f to get an idea of what we're looking for in terms of updating cmd refs.

This one is a bit more complex due to the new **Renaming stage outputs** section but still...

jorgeorpinel (approved):
Oops, I meant to approve ^

OK I'm merging this but please @dmpetrov, @shcheklein if you see any problems here lmk and I'll send a quick follow-up PR.

jorgeorpinel:
p.s. sorry for the wait and thanks again for your contribution Dmitry! It's definitely not typical that we'll take this long but we were in the GSoD bonding phase until last week.

PR #1744, opened 2020-08-31
Rule change for ESLint to error on unused variables.

### LINKED ISSUE #1291: CI does not fail on unused imports
As discovered and discussed [here](https://github.com/iterative/dvc.org/pull/1281#pullrequestreview-411397364), CI doesn't block commits on eslint warnings like unused imports. I ended up discovering some of my own when running `yarn lint-ts` after all TS issues are solved.

Getting the CI to fail on unused imports would be very nice, but considering that we can work around this by manually running `yarn lint-ts` I think it deserves a p1 over a p0.

I can work on this, but someone more familiar with our CI may be more effective here.

rogermparent:
This effectively makes commits with an unused variable blocked, fixes #1291

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,8 @@
       { "prefixWithI": "always" }
     ],
     "no-tabs": "error",
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "@typescript-eslint/no-unused-vars": "error"
   },
   "overrides": [
     {


.eslintrc.json
```diff
@@ -28,7 +28,8 @@
       { "prefixWithI": "always" }
     ],
     "no-tabs": "error",
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "@typescript-eslint/no-unused-vars": "error"
```
shcheklein, line 32: I wonder if we want to make some other checks an error?

rogermparent: It's likely, this is just the only one that comes to mind right now.

PR #1747, opened 2020-09-01
cases: Versioning Data and Models (rewrite)

jorgeorpinel:
> Continuation of https://github.com/iterative/dvc.org/pull/1716#issuecomment-684122937

- Moves the use case location and replaces all the links, adds a redirect.
- Includes same related changes to other docs as #1716
- Removes term "SCM" in favor of "version control" (and other phrases) — see https://trends.google.com/trends/explore?geo=US&q=Source%20code%20management,%2Fm%2F015_g6,%2Fm%2F0fj8g
- Redefines "data artifact" glossary term (could be removed from the glossary actually, not sure).

UPDATE: Pending stuff

- [x] PR scope https://github.com/iterative/dvc.org/pull/1747#pullrequestreview-521165697
- 2-column section? https://github.com/iterative/dvc.org/pull/1747#pullrequestreview-526848296


jorgeorpinel:
@shcheklein WDYT about the direction of this brand new WIP version of the use case? https://dvc-landing-use-cases-v-nb9tpb.herokuapp.com/doc/use-cases/versioning-data-and-model-files

shcheklein:
It's good in a sense that we are starting it from scratch 👍 

Regarding the intro- it still does not give any sense why would versioning data and models is important (reference to a problem), how DVC solves it and why DVC is the way to go (selling the philosophy).

jorgeorpinel:
> still does not give any sense why would versioning data and models is important

OK I updated the intro along these lines now.

> reference to a problem), how DVC solves it and why DVC is the way to go (selling the philosophy

This part I'm leaving for the the **## How it looks** section.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,28 +1,38 @@
...
 
-DVC enables versioning large files and directories such as datasets, data
-science features, and machine learning models using Git, but without storing the
-contents in Git.
+[Version control](https://en.wikipedia.org/wiki/Version_control) has become a
+staple in software engineering because it allows effective collaboration on
+source code. This means having a change history to go back to (commits),
+developing features in parallel (branching), assisted merging, peer-reviews
+(pull requests), tagging key revisions, etc. Imagine if we could use these
+features for data modeling!
```
jorgeorpinel, line 8: Is "data modeling" the right term? I'm trying to keep this use case generic, beyond ML or even data science.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,28 +1,62 @@
...
+
+DVC brings the best of both worlds together by replacing the data in the repo
+with small, human-readable
+[metafiles](/doc/user-guide/dvc-files-and-directories). Tracked data is
+<abbr>cache</abbr> locally outside the Git repo, and can easily be synchronized
+with on-premises or cloud storage. But unlike other alternatives (like Git-LFS),
+[remote storage](/doc/command-reference/remote) is optional — no server setup or
+special services are required.
+
+## Why bother?
```
jorgeorpinel, line 31: Too informal?

jorgeorpinel:
Full first draft is ready. I'd say it has the essential but
- It may be too verbose (at least the ratio of text vs sample is high)
- It's short i.e. could add more scenarios, better demo look&feel, etc.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Yea plus the WHY is explained earlier, going back to **How it looks**.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,30 +1,67 @@
...
+the data, we can repeat the procedure to generate more commits. Use
+`dvc checkout` to go back:
+
+```dvc
+$ git checkout v1
+$ dvc checkout
+M       data\raw
+```
+
+![](/img/versioning.png) _Full project restoration_
```
jorgeorpinel, line 62: The image may need an update to match the file names mentioned above, or to make it more generic. It's currently:

image

shcheklein:
@jorgeorpinel we got to the right level of the abstraction in the introduction, but it still doesn't cover the problems (use cases), nor it explains well _why_ DVC. If needed we can discuss this and brainstorm together.

jorgeorpinel:
> still doesn't cover the problems (use cases)

What about "having a change history to go back (commits), developing features in parallel (branching), assisted merging, peer-reviews (pull requests), tagging key revisions, etc."

Aren't the benefits of versioning basically the use case?

> nor it explains well why DVC

What about "What if we could **combine big data storage with versioning features**? ... DVC brings the best of both worlds together..." ?

shcheklein:
> Aren't the benefits of versioning basically the use case?

It's not explicit. It won't be easy to connect the dots. You have to imagine yourself how would it looks like for data. It feels that something similar to the Data Registry is required (even if we overlap a lot) - reproducibility, ability to debug, other stuff applied to data and models.

> What if we could combine big data storage with versioning features? ... DVC brings the best of both worlds together...

`big data storage` - people not necessarily want/do use "big data storage". It's quite indirect that we go from files, models, artifacts to the storage itself.

> Unlike other solutions (like Git-LFS) though, remote storage is optional in DVC

I'm not sure I understand this also. Remote storage is as optional in LFS as in DVC to my mind?

> How it looks

should be high level- focus on .dvc /dvc.yaml/dvc.lock files explain that there are commands that read them and synchronize data, etc

> DVC is not Git!

it's a strange place for this after the How it looks section. It should be part of the intro?

jorgeorpinel:
> won't be easy to connect the dots. You have to imagine yourself how would it looks like for data

Oh gotcha. Ah yes that's what the [**How it looks**](https://dvc-landing-use-cases-v-mskh4d.herokuapp.com/doc/use-cases/versioning-data-and-model-files#how-it-looks) section is for. Need to work on that section more...

> something similar to the Data Registry is required (even if we overlap

OK, added a benefits bullet list now. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I guess that term is more related to databases (e.g. https://www.datasciencegraduateprograms.com/data-modeling/). I'll use "data science" for now. Cc @jeremydesroches have you noticed this term or related in searches? May have potential

jorgeorpinel:
> It's quite indirect that we go from files, models, artifacts to the storage itself.

True... I skipped a step here originally @shcheklein, but I've tried to cover the gap now. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
jeremydesroches: dvc.org had some impressions for ["data modeling" and related terms](https://search.google.com/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&num_of_months=12&query=*data%20modeling&metrics=CLICKS%2CIMPRESSIONS). That's a start. Looking at overall search volume, "data modelling" and related terms have decent volume but the subject matter tends to the data architect role — not so much for data science or machine learning, etc.

jorgeorpinel: Thanks for checking. I went for something else in the end actually, after asking Elle about it. And now I found out how to use the Search Console haha so won't be bothering you with these basic questions much 😬

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
-`dvc repro` can now execute the `train` stage for you. DVC will track all of its
-outputs (`outs`) automatically. Let's do that, and commit this project version:
+[Version control](https://en.wikipedia.org/wiki/Version_control) has become a
+staple in software engineering because it allows effective collaboration on
+source code. This means having a change history to traverse (commits),
+developing parallel features (branching and merging), peer-reviews (pull
+requests), release management, etc. Imagine using these features for data
+engineering and machine learning!
+
+Unfortunately, versioning tools like [Git](https://git-scm.com/) are designed to
```
shcheklein, line 10: Before we jump into motivation behind DVC (Git does not work) ... we should go into some of the actual advantages of versioning/tracking data/models in ML workflow. Some of the advantages from below should in some way here.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
+handle small text files. DVC begins by enabling the [tracking](#how-it-looks) of
+large datasets and other <abbr>data artifacts</abbr> by replacing them with
+small, human-readable [metafiles](/doc/user-guide/dvc-files-and-directories)
+that `git` can handle. The data itself is <abbr>cached</abbr> outside of Git.
+Other features of DVC build upon this foundation to tackle hurdles specific to
+data science.
+
+> 💡 Please see [Get Started](/doc/start) for a primer on all of DVC's features.
+
+Another limitation of source code management is storage. While some data and
```
shcheklein, line 20: it looks more or less repetitive or overlaps a lot with the previous paragraph

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
+binary assets can exist in the Git repository, it's not usually recommended. And
+storage is severely limited by hosting services
+[like GitHub](https://docs.github.com/en/github/managing-large-files/what-is-my-disk-quota).
+DVC addresses this problem by integrating with
+[dedicated storage](#versioned-storage) platforms.
+
+## Summary of advantages
```
shcheklein, line 27: if we remove general versioning advantages to the top, we can keep it as advantages of using DVC here

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
+- Debugging: trace problems using the exact data that was used during
+  development.
+- Releasing: Tag stable data and models like you do with code. Employ semantic
+  versioning, connect branches to automatic deployment, etc.
+- Security: Git versioning of
+  [DVC metafiles](/doc/user-guide/dvc-files-and-directories) allows us to audit
+  data changes. And using cloud storage, data access control can be setup per
+  user or project.
+
+> ## ⚠️ DVC is not Git!
```
shcheklein, line 47: we should do something about it- looks artificial. It's better for this to be clear from the introduction. One sentence should be enough I guess?

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
+data science.
+
+> 💡 Please see [Get Started](/doc/start) for a primer on all of DVC's features.
+
+Another limitation of source code management is storage. While some data and
+binary assets can exist in the Git repository, it's not usually recommended. And
+storage is severely limited by hosting services
+[like GitHub](https://docs.github.com/en/github/managing-large-files/what-is-my-disk-quota).
+DVC addresses this problem by integrating with
+[dedicated storage](#versioned-storage) platforms.
```
shcheklein, line 25: we need some image at the very top

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
+> ## ⚠️ DVC is not Git!
+>
+> DVC metafiles such as `dvc.yaml` and `.dvc` files serve as placeholders to
+> track data files and directories (among other purposes). They point to
+> specific data contents in the <abbr>cache</abbr>, providing the ability to
+> store multiple data versions out-of-the-box. Full-fledged
+> [version control](https://git-scm.com/book/en/v2/Getting-Started-About-Version-Control)
+> itself is left for Git to handle, however.
+
+## How it looks
```
shcheklein, line 56: let's not focus on CLI commands, but on some visual representation of the workspace tree- large files - data/models are replaced with .dvc

also we can mention that DVC is more than that - dvc.yaml, metrics, etc

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,129 +1,103 @@
...
 
-```dvc
-$ git checkout v1.0 images.dvc
-$ dvc checkout images.dvc
-M       images
-```
+> For more hands-on experience, we recommend following the
+> [versioning tutorial](/doc/use-cases/versioning-data-and-model-files).
+
+## Versioned storage
```
shcheklein, line 91: this whole section looks disconnected at this moment

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK I removed the repetitive part and left only the information specifically about storage, which is what I wanted to point out here (not directly mentioned before).

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Agreed.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Yep, still working on that. Will send over a sketch ASAP.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Yeah I know. It's linked from the top (paragraph about storage) but it should be better in the next iteration... ⏳

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK, I made the first paragraph a little shorter and added a 2nd one about this (and an image placeholder after that). The motivation comes after that now. Not 100% sure about the summary of advantages though, it may be too abstract or redundant.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Hmm. The paragraph about advantages is very summarized, there's lots more details here (about data versioning in general).

jorgeorpinel: I can try to shorten it (it is a little long).

jorgeorpinel: > keep it as advantages of using DVC here

That would imply some sort of comparison vs other tools that can provide similar features. Airflow? MLFlow? Database-centric tools? I'm not sure, it could blur the use case (no clear focus) — if I'm understanding you correctly.

jorgeorpinel: I shortened this section a little for now.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: In fact it doesn't really belong in this use case any more. I moved it (shorter version) to the What is DVC? guide.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I think the section flow seems more natural now that DVC is not Git! is gone. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I guess you were right about this, I applied it now (more or less).

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: This is a short bullet list now.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: So here's an idea that tries to show a disorganized folder above ("before") and what it looks as different workspace versions below ("after")

IMG_20201012_205412

Cons:
- Too much info for a single figure, probably (could be summarized to less files above and only 3 versions below.
- Focuses mainly on one of the benefits mainly (better organization, efficient storage) although it also implies the enabling of git workflow.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I expanded on this now, with the focus you suggested. It's not 100% ready but PTAL.

content/docs/start/data-versioning.md
```diff
@@ -77,7 +77,7 @@ outs:
 
 > \* See
 > [Large Dataset Optimization](/doc/user-guide/large-dataset-optimization) and
-> `dvc config cache` for more information on file linking.
+> `dvc config cache` for more info on file linking.
```
shcheklein, line 80: intentional?

content/docs/use-cases/data-registries.md
```diff
@@ -16,28 +16,26 @@ repository would have all the metadata and history of changes in the different
 datasets. We could see who updated what and when, and use pull requests to
 update data, the same way we do with code.
 
-This is what we call a **data registry** — a kind of data management
-_middleware_ between ML projects and cloud storage. Here are its advantages:
+That's a **data registry**: a data management _middleware_ between ML projects
```
shcheklein, line 19: not sure - why do we touch the second use case here?

content/docs/user-guide/what-is-dvc.md
```diff
@@ -47,3 +47,12 @@ can version experiments, manage large datasets, and make projects reproducible.
 
   > Git servers, as well as SSH and cloud storage providers are supported,
   > however.
+
+## DVC is not Git!
+
+DVC metafiles such as `dvc.yaml` and `.dvc` files serve as placeholders to track
+large data files and directories for versioning (among other purposes). They
+provide the ability to store multiple data versions out-of-the-box, but this
+does not replace any SCM features. Full-fledged
```
shcheklein, line 56: SCM -> Git

`They provide the ability to store multiple data versions out-of-the-box` - sound like a single file is responsible to multiple versions. (vs multiple versions of the same file)

I would try to rephrase (simplify this)- it's a bit confusing, is Git still does versioning? etc...

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: yep, it feels like too much text

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,167 @@
 # Versioning Data and Model Files
```
shcheklein, line 1: btw, should we just name it `Data and Models` (remove Files, I'm not sure when we added this and why? or was it always like this?)

content/docs/start/data-versioning.md
jorgeorpinel: Yes, keeps the note in a single line when rendered. Could be "info." to make it a little more formal.

content/docs/use-cases/data-registries.md
jorgeorpinel: I initially reused some of the advantages from the bullet list in this doc. In the end we only kept one fo them as overlap, I think, but still I worked on the 2 docs in parallel a little bit. This particular line is an unrelated copy edit I noticed and decided to throw in.

content/docs/start/data-versioning.md
jorgeorpinel: BTW I copied this existing note to the use case (and changed to "info"), that's why I also edited it here — consistency.

content/docs/user-guide/what-is-dvc.md
jorgeorpinel: True. I rewrote this (added a 2nd p too) in f447b143. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I was just thinking that! (reviewing #1856)

It was "Data and Model Files Versioning" before, and I changed it to this at some point.

I guess "files" only intends to refer to models (i.e. versioning data and versioning model files), but it's not clear. Will update

jorgeorpinel: Initial update in 384218da

But I need to change a bunch of links from other docs which will add like 8 files to this PRs change... You may want to review the changes so far first?

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: How about just the bottom half?

image

I guess without the (before) part it may not make much sense... I'll think of something else ⏳

jorgeorpinel:
From call
- Versioning use case: much less text (3x shorter), a couple images. Just give the essence, why versioning, why DVC, how does it feel — that's it. Remove all the structure? Remove things from previous versions. More emphasis on codification (read about GitOps, Hashicorp, etc. Key terms: immutable history, level of indirection, integrate with CI systems, data as code, ML artifacts as code)
- "Versioned storage": Confusing term. Just mention it briefly in the how it feels. May be a use case in itself (overlapping with Versioning) (2 sides of the "data management" problem)

content/docs/use-cases/versioning-data-and-models/index.md
```diff
@@ -0,0 +1,156 @@
...
+to track the evolution of data and ML models? How to organize and store the
+revisions of these data artifacts for easy access and sharing?
+
+It turns out that we can find some answers in software engineering, like
+[version control](https://en.wikipedia.org/wiki/Version_control). This means
+keeping a change history (commits), automatic file deduplication, enabling
+parallel teamwork (branching), peer-reviews (pull requests), etc. Imagine if we
+could do the same with data!
+
+![](/img/data-as-code.png) _Matching data descriptions to a working environment_
```
jorgeorpinel, line 14: See proposed figure in https://dvc-landing-use-cases-v-6d9wm9.herokuapp.com/doc/use-cases/versioning-data-and-models:

image

content/docs/use-cases/versioning-data-and-models/index.md
```diff
@@ -0,0 +1,156 @@
...
+
+Versioning large data files and directories with Git is made possible by
+replacing them with small, human-readable
+[metafiles](/doc/user-guide/dvc-files-and-directories). The data itself is
+<abbr>cached</abbr> outside of Git (locally or externally), and can be
+synchronized automatically with
+[dedicated storage](/doc/use-cases/versioned-storage) — a good way to share data
+for collaboration.
+
+![](/img/404) _Parts of a DVC project_
```
jorgeorpinel, line 58: Pending another (more simple) figure here.

content/docs/use-cases/versioned-storage.md
```diff
@@ -0,0 +1,13 @@
+# Versioned storage
+
+What if we could **combine data and ML model versioning features with large file
+storage** solutions like traditional hard drives, NAS, or cloud services such as
+Amazon S3 and Google Drive? DVC brings together the best of both worlds by
+implementing easy synchronization between the data <abbr>cache</abbr> and
+on-premises or cloud storage for sharing.
```
jorgeorpinel, line 7: This is a draft/ proposal but not very firm. Could be completely removed.

redirects-list.json
```diff
@@ -24,10 +24,11 @@
   "^/doc/get-started(/.*)?$                                                               /doc/start",
   "^/doc/tutorial(/.*)?$                                                                  /doc/start",
   "^/doc/tutorials/get-started(/.*)?$                                                     /doc/start",
-  "^/doc/tutorials/versioning(/.*)?$                                                      /doc/use-cases/versioning-data-and-model-files/tutorial",
+  "^/doc/tutorials/versioning(/.*)?$                                                      /doc/use-cases/versioning-data-and-models/tutorial",
   "^/doc/tutorials(/.*)?                                                                  /doc/start",
 
-  "^/doc/use-cases/data-and-model-files-versioning/?$                                     /doc/use-cases/versioning-data-and-model-files",
+  "^/doc/use-cases/data-and-model-files-versioning/?$                                     /doc/use-cases/versioning-data-and-models",
+  "^/doc/use-cases/versioning-data-and-model-files(/.*)?$                                 /doc/use-cases/versioning-data-and-models$1",
```
jorgeorpinel, line 31: Does this change seem correct? I need to double check its behavior on the preview app anyway... ⏳

shcheklein: we don't necessarily need to change the URL ... i would not bother

jorgeorpinel: OK, reverted this in 579cc5b6 for now.

@jeremydesroches does it matters that the URL path `/doc/use-cases/versioning-data-and-model-files` doesn't exactly match the title `Versioning Data and Models`, SEO-wise? Thanks

jeremydesroches: > does it matter that the URL path `/doc/use-cases/versioning-data-and-model-files` doesn't exactly match the title `Versioning Data and Models`, SEO-wise?

@jorgeorpinel nope, doesn't need to match. The main consideration for URL paths is to keep them short (< 5 words) and coherent (no weird parameters or hashes etc.). Any key terms worth having in the title should be carried into the URL if possible, as you have done. This one checks out. 👍

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,135 @@
...
-## Track data and models for versioning
+# Versioning Data and Models
+
+Data science teams today face data management questions around versioning
+datasets, data artifacts, and machine learning models. How do we keep track of
+changes in data, code, and ML models? What’s the best way to organize and store
+multiple versions of data files for safe, persistent access? How can the
+lifecycle of data and models be followed and enforced?
+
+![](/img/data_ver_complex.png) _Exponential complexity of DS projects_
```
jorgeorpinel, line 9: New draft figure idea

image

jorgeorpinel: Although it would probably be better to represent the different lifecycles of code/data/models. A bit like figures from https://martinfowler.com/articles/cd4ml.html e.g.:

image

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,135 @@
...
+- Enforce data lifecycle policies — all project changes have to go through the
+  repository history.
+- Low coupling: separate code from data by caching large files/directories
+  automatically. This makes your project easier to maintain (high cohesion) and
+  improves data persistence.
+- Simple CLI: work with simple terminal [commands](/doc/command-reference)
+  similar to `git`.
+- Treating data as code also allows for other advanced features, see
+  [Get Started](/doc/start) for a primer.
+- Security: Audit an immutable history of changes to your data and models.
```
jorgeorpinel, line 34: Most of these benefits are about data versioning in general but I don't want to add another motivation paragraph or have 2 separate bullet lists.

jorgeorpinel: This list is already kind of long...

jorgeorpinel:
@shcheklein please check out the intro in while I rework the **How it looks** section. Thanks

https://dvc-landing-use-cases-v-6d9wm9.herokuapp.com/doc/use-cases/versioning-data-and-model-files

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: > Simple CLI...

The purpose of this bullet is to also mention the Viewer and VSCode plugin when they are released. Otherwise I would merge it with the first one about the "unified toolset".

jorgeorpinel: p.s. I did some updates to this list.

jorgeorpinel:
Alright, I worked on the **How it looks (& feels)** section. A few details I'll points out below but if that doesn't work, I'll probably have to re-write it from scratch again... PTAL @shcheklein and hopefully we can chat about it soon. Thanks

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -15,7 +15,8 @@ continuous integration (CI), and other best practices that improve productivity.
 This means writing a description of which data, ML artifacts, etc. should be in
 the environment at a given time. DVC can restore the <abbr>workspace</abbr>
 files and directories (from a separate data storage) to match that description.
-Here are some benefits of this approach:
+
+## Why do it with DVC
```
jorgeorpinel: Is this too salesy? It's nice to have the right-hand CONTENT nav entry.

image

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -113,34 +94,32 @@ stages:
         md5: ede2872bedbfe10342bb1c416e2f049f
 ```
 
-**Remember: DVC doesn’t replace Git.** Having codified data
-<abbr>artifacts</abbr> and [pipelines](/doc/start/data-pipelines), regular `git`
-commands are used to create versions (commits, tags, branches, etc.):
+> See [Data Pipelines](/doc/start/data-pipelines) to learn about stages.
```
jorgeorpinel, line 97: Not sure we want to include the dvc.lock example and link to data pipelines, may be too much/ confusing. Should we only have the data.dvc example? It would definitely help make the doc shorter, but it looses some completeness.

jorgeorpinel: Plus maybe we want the pipelines teaser.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -113,34 +94,32 @@ stages:
...
+$ git commit -m 'First modeling experiment'
 
-$ git commit -m 'Data cleanup v1.0'
+# Iterate, repeat!
 ```
 
-> By matching versions of code with snapshots of the data at the time, we can
-> achieve full [reproducibility](/doc/start/data-pipelines).
+> Note that DVC prevents Git from tracking data in the workspace (via
+> `.gitignore`).
```
jorgeorpinel: Probably no need for this note but then the `git add data/.gitignore` is totally out of context. Should we remove both `data/.gitignore` in the `git add` line AND the note, or just the note? (Or keep both)

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -113,34 +94,32 @@ stages:
...
+be rewound ⏪:
 
 ```dvc
 $ git checkout v1.0
 $ dvc checkout
-M       data\raw
+M       data
 ```
 
 ![](/img/versioning.png) _Full project restoration_
```
jorgeorpinel: Keeping this old image is a bit forced. May not be necessary at all if we make a new figure for the **How it looks & feels** section.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -26,82 +27,62 @@ Here are some benefits of this approach:
...
-to switch between versions
-
-Versioning large data files and directories with Git is made possible by
-replacing them with small, human-readable
-[metafiles](/doc/user-guide/dvc-files-and-directories). The data itself is
-<abbr>cached</abbr> outside of Git (locally or externally), and can be
-synchronized automatically with
-[dedicated storage](/doc/use-cases/versioned-storage) — a good way to share data
-for collaboration.
+## How it looks & feels
```
jorgeorpinel, line 37: BTW is this a good title? "look & feel" is kind of a design term, I think.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I decided to remove this after all. Resolving for now

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I removed both (replaces with `...` above). Resolving

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Actually, I think it's good: "looks" refers to the worspace files, dvc cache, etc, while "feels" refers to the DVC/Git workflow. Resolving

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -282,22 +282,31 @@ Full <abbr>parameters</abbr> (key and value) are listed separately under
 
 ## Structure of the cache directory
 
+The DVC cache is a
+[content-addressable storage](https://en.wikipedia.org/wiki/Content-addressable_storage)
+(CAS), which adds a layer of indirection between code and data. It's hidden from
+the user (and from Git), and typically local. However, it can be synchronized
+with [remote storage](/doc/command-reference/remote) for sharing.
```
jorgeorpinel, line 289: This I extracted from a previous iteration of the use case, that's why this file was modified. But then I also simplified everything else in this section which is pretty out of scope. Feel free to ignore and I'll move it to it's own PR... Sry

jorgeorpinel: OK, reverted these changes for now.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,100 @@
...
+the environment at a given time. DVC can restore the <abbr>workspace</abbr>
+files and directories (from a separate data storage) to match that description.
+
+Data versioning lets us identify exact research inputs later, so anyone can
+understand and reproduce results. DVC brings additional benefits on top of that:
+make projects easier to work on by separating code from data (which also
+increases data persistence); enforce lifecycle and security policies by having
+all changes go though a Git repo (immutable history); collaborate with a toolset
+everyone is familiar with (data scientists, engineers, managers), and others
+(see [Get Started](/doc/start)).
```
jorgeorpinel, line 25: Should I turn this back into a bullet list? It's kind of boring just a paragraph after the next, and so on.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,100 @@
...
+
+Data science teams today face data management questions around versioning
+datasets, data artifacts, and machine learning models. How do we keep track of
+changes in data, code, and ML models? What’s the best way to organize and store
+multiple versions of data files for safe, persistent access? How can the
+intertwined lifecycles of data and models be followed and enforced?
+
+![](/img/data_ver_complex.png) _Exponential complexity of DS projects_
+
+DVC proposes to _codify_ data projects in order to adopt existing engineering
```
jorgeorpinel, line 11: "codify data", "data-as-code", etc.: should we switch to Hashicorp's existing concept of "Versioning through Codification
"? (See https://www.hashicorp.com/tao-of-hashicorp) 

@jeremydesroches do you think one or the other has better SEO potential?

jorgeorpinel: "Automation through Codification" also seems relevant:

> Codification allows for knowledge to be executed by machines, but still readable by operators.
(Same ref.)

content/docs/command-reference/cache/index.md
```diff
@@ -25,7 +25,7 @@ Learn more about it's
 > [File link types](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
 > for more information on file links on different platforms.
 
-> For more cache-related configuration options refer to `dvc config cache`.
+For cache configuration options, refer to `dvc config cache`.
```
shcheklein, line 28: let's not make it part of this PR please? it looks unrelated

content/docs/command-reference/index.md
```diff
@@ -14,10 +14,10 @@ does not change directories in your terminal).
   `dvc init`.
 - Copy data files or dataset directories for modeling into the repository, and
   track them with DVC using the `dvc add` command.
-- Process raw data with your own source code, using `dvc.yaml` and/or the
```
shcheklein, line 17: same, looks like changes are unrelated

content/docs/use-cases/data-registries.md
```diff
@@ -8,36 +8,34 @@ commands like `dvc import` and `dvc get`. This means that your projects can
 depend on data from other DVC repositories, **similar to a package management
 systems for data science**.
 
-![](/img/data-registry.png) _Data and models as code_
```
shcheklein, line 11: please, let's not touch data registry in this PR

content/docs/user-guide/what-is-dvc.md
```diff
@@ -47,3 +47,18 @@ can version experiments, manage large datasets, and make projects reproducible.
 
   > Git servers, as well as SSH and cloud storage providers are supported,
   > however.
+
+## DVC is not Git!
```
shcheklein, line 51: DVC is not a new Git? DVC does not replace Git?

content/docs/user-guide/what-is-dvc.md
```diff
@@ -47,3 +47,18 @@ can version experiments, manage large datasets, and make projects reproducible.
...
+
+DVC metafiles such as `dvc.yaml` and `.dvc` files serve as placeholders to track
+large data files and directories for versioning (among other
+[purposes](/doc/user-guide/dvc-files-and-directories)). These metafiles change
+along with your data, and you can use Git to place them under
+[version control](https://git-scm.com/book/en/v2/Getting-Started-About-Version-Control)
+as a proxy to the actual data versions, which are stored in the <abbr>DVC
+cache</abbr> (outside of Git). This does not replace features of Git.
+
+DVC does, however, provide several commands that are somewhat analogous to their
```
shcheklein, line 61: `somewhat analogous` can we replace this with something explicit?

shcheklein: e.g. from use perspective?

jorgeorpinel: I did list examples below but OK, I'm moving them higher up in the text here.

content/docs/command-reference/cache/index.md
jorgeorpinel: OK rolling this back...

content/docs/command-reference/index.md
jorgeorpinel: Indeed. Rolling back as well.

content/docs/use-cases/data-registries.md
jorgeorpinel: OK please ignore these changes on Data Registries here for now. It's a little harder to roll back right now for me (git stash mess) but it will be reverted ASAP.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,108 @@
...
+outside of Git in an efficient way that prevents file duplication.
+
+<div style="display: flex; flex-flow: row wrap; margin: 0 -0.5rem;">
+<div style="flex: 1 0 50%; padding: 0.5rem;">
+
+**Before**: multiple filename versions — inefficient and error-prone
+
+```git
+ .
+ ├── data
```
jorgeorpinel, line 26: Hi @rogermparent 👋 Question!

As you may or may not see in https://dvc-landing-use-cases-v-6d9wm9.herokuapp.com/doc/use-cases/versioning-data-and-model-files (Ivan reported this is not rendered properly on his side), I'm trying to have a 2-column section in this use case:

image

I just copied some HTML from https://dvc.org/doc and it works on my side but not for every window width. What's the best way to do this? Can we perhaps generalize this layout and attach it to some tag?

p.s. I'm not 100% sure we'll keep it for this use case but it can surely be a useful tool in the engine, I think. Let's confirm with @shcheklein of course. Thanks

jorgeorpinel: > it works on my side but not for every window width

On this, of course we don't want 2 cols for slim windows, but I mean the current code breaks for some longer widths as well.

rogermparent: >the current code breaks for some longer widths as well.

I'm actually not sure where the code is breaking here, unless you simply mean that the columns should collapse sooner than they do now so the columns have a smaller minimum width.

I that is the case, the simplest quick fix would be to change the `flex: 1 0 50%` styles to something like `flex: 1 0 325px`, changing the `flex-basis` to specify a preferred minimum width (325px is roughly around that the columns end up at with max screen width).

>  Can we perhaps generalize this layout and attach it to some tag?

Certainly! While nested elements like this can be tricky and have a bit of a clunky syntax, we can at least get a better syntax than raw copy+pasted HTML.

jorgeorpinel: > I'm actually not sure where the code is breaking here, unless you simply mean that the columns should collapse sooner

Hmmm, no. I swear it wasn't working on certain widths, but now it's OK!

@shcheklein could you confirm about this working for you and whether you think making a special component for this in our engine would be worth it?

Thanks

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+👩‍💻.
+
+DVC's approach brings the following benefits:
+
+- **Tidy project**: Work with a natural file structure. No need for ad hoc
+  naming conventions like `data/20190922/labels_v7_final`.
+- **Reproducibility**: Restore any project version and find the corresponding
+  data instantly. Identify past research inputs to understand its results, or
+  for debugging.
+- **Easy storage**: All versions of data are
```
jorgeorpinel, line 37: @shcheklein here's the latest list of benefits with the advanced scenarios included. Known problems:
- It's too long again.
- Not sure about "Integration" but can't think of a better term rn.

PTAL.

content/docs/use-cases/data-registries.md
jorgeorpinel: OK I finally rolled back the changes to Data Regs. Resolving this one

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+changes in data, code, and ML models together? What's the best way to organize
+and store multiple versions of data and model files? How can data lifecycles be
+defined and enforced?
+
+![](/img/data-ver-complex.png) _Exponential complexity of data science projects_
+
+Data Version Control (DVC) can address these issues by helping you organize your
+data and models effectively and capture their versions in
+[Git commits](<(https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository)>),
+while storing the actual data contents outside ot Git. This way there's a single
```
dmpetrov, line 14: "outside ot Git" - where is it? also, `ot` - typo.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
-versioning large data files.
+# Versioning Data and Models
+
+Data science teams today face data management questions around versioning
+datasets, data artifacts, and machine learning models. How do we keep track of
+changes in data, code, and ML models together? What's the best way to organize
+and store multiple versions of data and model files? How can data lifecycles be
+defined and enforced?
+
+![](/img/data-ver-complex.png) _Exponential complexity of data science projects_
```
dmpetrov, line 9: I'm not sure I understand `train feats` from the image. I have some guesses but I'm not sure my understanding is correct.
Is there a way to use a more common term?

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+project history for all of its assets and artifacts, and at the same time the
+data stays with you.
+
+![](/img/project-versions.png) _DVC matches the right versions of code and
+data._
+
+DVC enables data _versioning through codification_. This means describing which
+data, ML artifacts, etc. should be in the environment at any given time. This is
+achieved with special [metafiles](/doc/user-guide/dvc-files-and-directories)
+that can be put in Git — but we won't much deeper into how it works here, since
```
dmpetrov, line 24: `but we won't much deeper....` - it sounds like an excuse for removing the previous version of this doc. You can remove this sentence without losing anything.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+
+DVC's approach brings the following benefits:
+
+- **Tidy project**: Work with a natural file structure. No need for ad hoc
+  naming conventions like `data/20190922/labels_v7_final`.
+- **Reproducibility**: Restore any project version and find the corresponding
+  data instantly. Identify past research inputs to understand its results, or
+  for debugging.
+- **Easy storage**: All versions of data are
+  [cached](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory)
```
dmpetrov, line 38: Easy storage - not clear what is it about. What is cache (first reference from this page)? You can remove Easy storage from the list.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+we have many [guides](/doc/user-guide) and [references](/doc/command-reference)
+for that. To learn how it looks and feels, try our
+[versioning tutorial](/doc/use-cases/versioning-data-and-model-files/tutorial)
+👩‍💻.
+
+DVC's approach brings the following benefits:
+
+- **Tidy project**: Work with a natural file structure. No need for ad hoc
+  naming conventions like `data/20190922/labels_v7_final`.
+- **Reproducibility**: Restore any project version and find the corresponding
```
dmpetrov, line 34: A bit too abstract. I'd suggest moving it down in the list.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+  [shared](/doc/use-cases/sharing-data-and-model-files) and
+  [reused](/doc/start/data-access) via on-premises or cloud
+  [storage remotely](/doc/command-reference/remote).
+- **Integrations**: Adopting a
+  [Git workflow](https://about.gitlab.com/topics/version-control/what-is-git-workflow/)
+  opens the door to advanced scenarios such as continuous integration (CI/CD
+  e.g. [CML](https://cml.dev/)), specialized patterns like
+  [data registries](/doc/use-cases/data-registries), and other best practices
+  that improve productivity and scalability.
+- **Data security**: The project history (in Git) is immutable, which allows for
```
dmpetrov, line 54: I'd separate this into two:
- **Data security** - set up security in underlying storage like S3 or NFS....
- **Data compliance** - auditing processes to change data and models using regular GitFlow and Pull Requests. Visibility on data and ML model modification using a regular Git - who modified a dataset, when it was happening, who approved the change.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,60 @@
...
+achieved with special [metafiles](/doc/user-guide/dvc-files-and-directories)
+that can be put in Git — but we won't much deeper into how it works here, since
+we have many [guides](/doc/user-guide) and [references](/doc/command-reference)
+for that. To learn how it looks and feels, try our
+[versioning tutorial](/doc/use-cases/versioning-data-and-model-files/tutorial)
+👩‍💻.
+
+DVC's approach brings the following benefits:
+
+- **Tidy project**: Work with a natural file structure. No need for ad hoc
```
dmpetrov, line 32: It is not clear how it is achieved. How about ... "Data metafiles allow addressing file names like `labels.tsv` directly while the actual links are codified in metafiles like `data/20190922/labels_v7_final`. Modify data or links to data independently from the modeling code."

Tidy project --> Tidy file naming ?

dmpetrov (commented):
It is a good writen use case. A few suggestions are inline.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: The point was to emphasize data doesn't go into Git. I 've now changed "storing" for "keeping" to avoid causing the question of where they are exactly. Caching/storage is mentioned a bit later.

And `t` looks a lot like `f`! You pass the dyslexia test! 👍 Fixed

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I was thinking that too. Maybe just "features"?

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: The intention here is linking to the user guide and cmd reference areas of the docs, so the user is aware that all the details are there.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Using **Simple file names** now.

> It is not clear how it is achieved.

The figures are supposed to convey this but otherwise we could add text explaining this better somewhere above in the story, but I don't think the benefits list is the best place for this explanation. Thanks for the suggested text though, let me see if I can incorporate it naturally ⌛

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: True. We used to have more info on storage and cache before. Removed!

jorgeorpinel: But... I'm starting to think we should cover cache/storage a bit more as part of the story at least, as it's in the "problems" presented initially:

> best way to organize and **store** multiple versions of data and model files?

OR we could just no talk about storage anywhere at all in this doc, leaving that for **Sharing Data and Model Files** (needs rewrite too).

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK, I moved it down for now.

But I thought this was a pretty key purpose of versioning in general. Maybe we can rewrite it in a less conceptual terms, but I'm not sure how.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I agree that _compliance_ is a better term, thanks! Updated along the lines you suggested.

Data security: maybe we don't need this benefit? Especially if we decide not to cover storage in this doc (per https://github.com/iterative/dvc.org/pull/1747#discussion_r528449200 above). I removed it for now.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: Updated to "features" in both figures.

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK! I added the following sentence after "special metafiles that can be put in Git.":

> These use constant file names but link to variable contents (via the metadata in Git), so that the data can be modified independently from the source code that addresses it.

Does this help give context to the benefit?

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: This part has been moved to the next short paragraph now and I think it makes more sense this way. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: there are a few things to storage story that should be ideally covered in the DVC benefits:

1. data stays with you - runs on top of your cloud, on-prem, etc. 
2. efficient data management - local cache that utilizes links, shared cache scenario to share data efficiently

`easy` in this case doesn't provide any meaningful information (what is Hard storage? Easy from what perspective?)

also, no need to emphasize on all versions here - that's what all the previous text is about to my mind

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: How can we make it less abstract?

it's an important point to my mind (even more important than a Tidy project (Simple names now))

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: Data compliance should be probably part of the Git flow? or Git flow can be replaced by it?

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: 1. "data stays with you" is in the story as:

> you can have a single, immutable history for all of your assets and artifacts, and at the same time, the **data stays with you**.

Separately, the **Collaboration** benefit mentions that data can be shared/reused via **on-prem or cloud storage** (which currently is the only mention of "storage" in the doc).

Working on 2. but we may end up with too many bullets again ⌛

jorgeorpinel: > 2. efficient data management - local cache... shared cache...

OK I added this benefit but we're back to square one here. The cache mention is kind of out of context, maybe unnecessary. I think it's probably better to revisit the next use case to focus more on storage (both local cache and remotes).

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I put in on the top now 😆 for now. But we should come up with a criteria do decide which ones go first... After we decide which benefits to even keep (too many rn probably).

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: cache should not be considered as DVC cache ... there is a general notion of caching that most people would understand. E.g. if you say that `DVC gives you a way to granularly cache data on your local or training machine` most people should understand this to my mind.

content/docs/use-cases/versioning-data-and-model-files/index.md
shcheklein: @dmpetrov 's point stays - if it feels abstract, should we reconsider the description? make Reproducibility and something in bold?

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: OK I reworked the doc per the comments above and some private feedback. PTAL

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I generalized all benefits into 5 now, PTAL. But the problem with **Git workflows** is that it has too many sub-benefits and it would be a huge paragraph if we include everything. I think Data compliance is a high-level benefit of it's own anyway?

We're even missing data security but per the comments above it sounds like it would be better to list it in https://dvc.org/doc/use-cases/sharing-data-and-model-files (as it has more to do with remote storage).

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: > Tidy project... not clear how it is achieved...

All the text is significantly different now and I think this has been addressed. Resolving!

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: > > but we won't much deeper.... - it sounds like an excuse
>
> has been moved to the next short paragraph

Resolving per ^

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: This is back around the middle and the other benefits have been generalized more so maybe now they all have a consistent level of abstraction?

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,76 @@
...
+machine learning models. How do we keep track of changes in data, source code,
+and ML models together? What's the best way to organize and store variations of
+these files and directories?
+
+![](/img/data-ver-complex.png) _Exponential complexity of data science projects_
+
+Another problem in the field has to do with bookkeeping: being able to identify
+past research inputs and processes to understand results, for knowledge sharing,
+or for debugging. There's also the matter of defining and enforcing data
+lifecycles.
```
jorgeorpinel, line 13: Not sure about the data lifecycles mention since it's not addressed directly later as part of the solution.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,76 @@
...
+
+- **Data compliance**: Review data modification attempts as Git
+  [pull requests](https://www.dummies.com/web-design-development/what-are-github-pull-requests/).
+  Audit this process later to learn when data or model versions were approved,
+  and why.
+
+- **GitOps**: Connect your data science projects with the Git-powered universe.
+  Git workflows open the door to advanced tools such as continuous integration
+  (like [CML](https://cml.dev/) CI/CD), specialized patterns such as
+  [data registries](/doc/use-cases/data-registries), and other best practices.
```
jorgeorpinel, line 72: I thought about adding another sentence here about production-ready systems, productivity and scalability, but it's already a long text. We can add it later when we have a "DVC in production" use case, perhaps.

jorgeorpinel (commented):
Another iteration of this is ready for review. Here are 2 known (minor) issues I see:

content/docs/use-cases/versioning-data-and-model-files/index.md
jorgeorpinel: I'm going to move this all the way to the bottom instead (summary) where lifecycles are actually mentioned again.

content/docs/use-cases/versioning-data-and-model-files/index.md
```diff
@@ -1,130 +1,88 @@
...
+  [free](https://github.com/iterative/dvc/blob/master/LICENSE), open-source
+  [command line](/doc/command-reference) tool that doesn't require databases,
+  servers, or any other special services.
+
+- **Consistency**: Keep your projects readable with stable file names — they
+  don't need to change because they represent variable data. No need for
+  complicated paths like `data/20190922/labels_v7_final` or for constantly
+  editing these in source code.
+
+- **Efficient data management**: Use a familiar and cost-effective storage
```
shcheklein, line 58: A final ask in private to update this a bit

shcheklein (approved):
Cool 😎 stuff!

jeremydesroches:
@jorgeorpinel would you still like me to review this for SEO-related updates, as you mentioned before?

jorgeorpinel:
That would be useful @jeremydesroches but completely up to you. Also at this point, since the rewrite was merged recently, do you think we should wait more weeks to be able to compare search patterns vs the previous version?

I guess giving it a read to make sure the key terms you added in https://github.com/iterative/dvc.org/pull/1806/files#diff-34b5f307478d996f5f763b6e71d8e9689543d9ef99f85d36a505c8ef47e57fee are still there (if possible) would be ideal.

PR #1759, opened 2020-09-04
s/open/read in dvc.api.read reference

skshetry:
(rant starts)
Feels like the `dvc.api` could be better overall (it's written in a man-page style instead of easy to use, gets-out-of-way reference.
The block that tries to show the function signature is split into multiple lines and has a `def` in front of them.
Formatting is also broken in many places and seems that there's too much information than there should be for what's it meant to be very simple APIs.

Another issue, I noticed is that we have too many exceptions for `dvc.api`. Not a documentation issue, but docs do reflect the issue on the core side.
(rant ends)

Just fixing a typo/mistake on `dvc.api.read()`. :slightly_smiling_face:

### DIFF
--- a/content/docs/api-reference/read.md
+++ b/content/docs/api-reference/read.md
@@ -3,7 +3,7 @@
 Returns the contents of a tracked file.
 
 ```py
-def open(path: str,
+def read(path: str,
          repo: str = None,
          rev: str = None,
          remote: str = None,


shcheklein (approved):

shcheklein:
> it's written in a man-page style instead of easy to use, gets-out-of-way reference

API reference/command references are meant to be man-page style. We need a place we describe the behavior, edge cases, etc. (e.g. where would you describe that `read` streams data vs making a copy?).

I guess `reference` itself means something that is not tutorial, not get started, but rather something you could refer to set expectations, prove something, etc. 

It's a place you go to find _details_, not an easy way to get started.

 > gets-out-of-way reference

there are ways to solve this:

1. consider improving reference itself- include simple examples at the top? (usually it's hard to make one simple example that covers everything and this section grows, that's why we keep it at the bottom)
2. put a link to examples section explicitly at the top and restructure examples to be a "gets-out-of-way reference"
3. use Katakoda
4. add Python API to the get started section + mention it in the references.

I would personally prefer to have less docs and refactor/update the examples sections + put some explicit links if needed.

@skshetry WDYT?

> The block that tries to show the function signature is split into multiple lines and has a def in front of them.

it's a minor, so don't want to focus too much, but I'm not sure I understand the issue with this one ...

shcheklein:
@skshetry thanks for the fix, btw!

jorgeorpinel:
Yeah, thanks for the fix.

As for the discussion, what would an example of a "easy to use, gets-out-of-way reference" be? I mean what would it get out of the way of?

> Formatting is also broken in many places

Not seeing any broken format at least in api.read

> there's too much information than there should be for what's it meant to be very simple APIs
> we have too many exceptions

Maybe they're not as simple as intended 🙂 

Don't get me wrong @skshetry, appreciate the feedback. But it's not very actionable. If you can better explain, give examples, suggest alternatives, find users having these confusions, etc. that would be great 🙂 

If you do have insights please open an issue and we'll figure it out. Maybe we need an API guide separate to the references, for example. Or to write some tutorials or blogs. Or some of Ivan's great ideas. Thanks

PR #1760, opened 2020-09-05
docs: fix non-intuitive diff offset formatting in start/data-pipelines.md

bobertlo:
The `diff` notation breaks the tree visualization in this quote block.

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -102,9 +102,9 @@ prepare:
 +      │       └── train.tsv
 +      ├── dvc.yaml
 +      ├── dvc.lock
-        ├── params.yaml
-        └── src
-            ├── ...
+       ├── params.yaml
+       └── src
+           ├── ...
   ```
 
 - The last line, `python src/prepare.py ...`, is the command to run in this
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -97,11 +97,11 @@ prepare:
        ├── data
        │   ├── data.xml
        │   ├── data.xml.dvc
-  +   │   └── prepared
-  +   │       ├── test.tsv
-  +   │       └── train.tsv
-  +   ├── dvc.yaml
-  +   ├── dvc.lock
++      │   └── prepared
++      │       ├── test.tsv
++      │       └── train.tsv
++      ├── dvc.yaml
++      ├── dvc.lock
         ├── params.yaml
         └── src
             ├── ...
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -93,10 +93,10 @@ prepare:
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
   ```
-        .
-        ├── data
-        │   ├── data.xml
-        │   ├── data.xml.dvc
+       .
+       ├── data
+       │   ├── data.xml
+       │   ├── data.xml.dvc
   +   │   └── prepared
   +   │       ├── test.tsv
   +   │       └── train.tsv
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -92,7 +92,7 @@ prepare:
 - `-o data/prepared` specifies an output directory for this script, which writes
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
-  ```diff
+  ```
         .
         ├── data
         │   ├── data.xml
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -93,18 +93,18 @@ prepare:
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
   ```
-       .
-       ├── data
-       │   ├── data.xml
-       │   ├── data.xml.dvc
-  +    │   └── prepared
-  +    │       ├── test.tsv
-  +    │       └── train.tsv
-  +    ├── dvc.yaml
-  +    ├── dvc.lock
-       ├── params.yaml
-       └── src
-           ├── ...
+   .
+   ├── data
+   │   ├── data.xml
+   │   ├── data.xml.dvc
+  +│   └── prepared
+  +│       ├── test.tsv
+  +│       └── train.tsv
+  +├── dvc.yaml
+  +├── dvc.lock
+   ├── params.yaml
+   └── src
+       ├── ...
   ```
 
 - The last line, `python src/prepare.py ...`, is the command to run in this
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -97,11 +97,11 @@ prepare:
        ├── data
        │   ├── data.xml
        │   ├── data.xml.dvc
-+      │   └── prepared
-+      │       ├── test.tsv
-+      │       └── train.tsv
-+      ├── dvc.yaml
-+      ├── dvc.lock
+  +    │   └── prepared
+  +    │       ├── test.tsv
+  +    │       └── train.tsv
+  +    ├── dvc.yaml
+  +    ├── dvc.lock
        ├── params.yaml
        └── src
            ├── ...
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -93,18 +93,18 @@ prepare:
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
   ```diff
-      .
-      ├── data
-      │   ├── data.xml
-      │   ├── data.xml.dvc
+        .
+        ├── data
+        │   ├── data.xml
+        │   ├── data.xml.dvc
   +   │   └── prepared
   +   │       ├── test.tsv
   +   │       └── train.tsv
   +   ├── dvc.yaml
   +   ├── dvc.lock
-      ├── params.yaml
-      └── src
-          ├── ...
+        ├── params.yaml
+        └── src
+            ├── ...
   ```
 
 - The last line, `python src/prepare.py ...`, is the command to run in this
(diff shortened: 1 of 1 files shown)


content/docs/start/data-pipelines.md
```diff
@@ -93,18 +93,18 @@ prepare:
...
-      .
-      ├── data
-      │   ├── data.xml
-      │   ├── data.xml.dvc
+        .
+        ├── data
+        │   ├── data.xml
+        │   ├── data.xml.dvc
   +   │   └── prepared
   +   │       ├── test.tsv
```
jorgeorpinel, line 101: Hmmm I don't know. Yeah, this aligns the `|` characters but there's several other diff examples here and there and they all have indentation problems when rendered.

jorgeorpinel: It also makes it extra obvious which lines are being added the way it is now.

efiop: @jorgeorpinel `+` make that obvious, but the visual layout looks outright broken. Maybe the issue here is that we are using `diff` format, while this is not actually a patch. We could higlight new entries some other way, e.g. by making it bald  or by adding a bash-like comment suffix that looks smth like  `# added` or something.

efiop: This patch keeps `+` to make added entries obvious and also fixes the layout. :+1:

jorgeorpinel: > the visual layout looks outright broken
Agree, but we have an issue to fix this 🙂 #1761

> using diff format, while this is not actually a patch
So `git diff` doesn't output proper diff format? Or is something missing in this snippet to make it valid?

> We could higlight new entries some other way
Sure, that's a good idea. IMO this can be left alone though (since the bug should be addressed eventually).

> making it bald
👨‍🦲

shcheklein: @jorgeorpinel how would that output look in git diff? How would it look on the GH? What about if we copy/paste it from git diff or GH - will we have extra spaces?

jorgeorpinel: Let me check... ⌛

jorgeorpinel: Oh sorry I was confused, this has nothing to do with `git diff`, I think it's just supposed to be a regular `diff` of outputs of `tree`..? Not sure, you created this one Ivan 🙂

Unfortunately, `tree` on Windows uses different characters and it doesn't look good the way I tried it but please copy/paste the commands to double check:

```console
λ mkdir test && cd test
λ mkdir data/raw
λ touch data/raw/words.txt
λ touch data/raw/names.csv
λ tree > b
λ mkdir data/clean
λ touch data/clean/words.txt
λ tree > 2
λ diff b a
2a3
> |-- 2
4c5,7
<                     `-- raw
---
>     |               `-- clean
>     |               `-- words.txt
>     `-- raw
8c11
< 2 directories, 3 files
---
> 3 directories, 5 files
```

jorgeorpinel: Let me check what `git diff` does though... ⌛

jorgeorpinel: OK `git diff` would show something like this:

```console
λ mkdir -p data/raw
λ touch data/raw/words.txt
λ touch data/raw/names.csv
λ tree > tree
λ git add tree
λ mkdir data/clean
λ touch data/clean/words.txt
λ git diff
diff --git a/tree b/tree
index e60fc51..c90b158 100644
--- a/tree
+++ b/tree
@@ -1,8 +1,10 @@
 .
 |-- data
 |
+|               `-- clean
+|   |               `-- words.txt
+|
    `-- raw
 |       |           `-- names.csv
 |                   `-- words.txt
 `-- tree

-2 directories,
 3
+directories, 4
  files
```

OK so never mind, I was confused again... I guess the intention here was to output whatever format `git diff` does, and the conclusions from above are:

1. `git diff` does align stuff adding **one space** to lines without `+`
2. Indeed this is not `diff` format. We should probably not use that language name in any of the existing blocks throughout docs...

content/docs/start/data-pipelines.md
```diff
@@ -93,18 +93,18 @@ prepare:
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
   ```diff
```
jorgeorpinel, line 95:
````suggestion
  ```
````

File: content/docs/start/data-pipelines.md [hunk 3]
```diff
@@ -93,18 +93,18 @@ prepare:
...
-      .
-      ├── data
-      │   ├── data.xml
-      │   ├── data.xml.dvc
+        .
+        ├── data
+        │   ├── data.xml
+        │   ├── data.xml.dvc
```
jorgeorpinel, line 99:
```suggestion
       .
       ├── data
       │   ├── data.xml
       │   ├── data.xml.dvc
```

content/docs/start/data-pipelines.md
```diff
@@ -93,18 +93,18 @@ prepare:
...
   +   │   └── prepared
   +   │       ├── test.tsv
   +   │       └── train.tsv
   +   ├── dvc.yaml
   +   ├── dvc.lock
```
jorgeorpinel, line 104:
```suggestion
+      │   └── prepared
+      │       ├── test.tsv
+      │       └── train.tsv
+      ├── dvc.yaml
+      ├── dvc.lock
```

content/docs/start/data-pipelines.md
```diff
@@ -93,18 +93,18 @@ prepare:
...
-      ├── params.yaml
-      └── src
-          ├── ...
+        ├── params.yaml
+        └── src
+            ├── ...
```
jorgeorpinel, line 107:
```suggestion
       ├── params.yaml
       └── src
           ├── ...
```

jorgeorpinel (commented):
OK here's the proposed fix per my last comment. Let me commit these changes and check the result... ⏳

content/docs/start/data-pipelines.md
shcheklein: >  We should probably not use that language name in any of the existing blocks throughout docs...

is there some other prisma.js lang that we should use that would highlight it similar to the GH interface?

jorgeorpinel: Good Q. Idk. Do we know what format does `git diff` use BTW? I'll try some options from https://prismjs.com/#supported-languages ⏳

content/docs/start/data-pipelines.md
```diff
@@ -92,19 +92,19 @@ prepare:
 - `-o data/prepared` specifies an output directory for this script, which writes
   two files in it. This is how the <abbr>workspace</abbr> should look like now:
 
-  ```diff
-      .
-      ├── data
```
jorgeorpinel, line 97: Why does this block start with 6 trailing spaces though?

content/docs/start/data-pipelines.md
```diff
@@ -92,19 +92,19 @@ prepare:
...
+  ```
+       .
+       ├── data
+       │   ├── data.xml
+       │   ├── data.xml.dvc
++      │   └── prepared
++      │       ├── test.tsv
++      │       └── train.tsv
++      ├── dvc.yaml
++      ├── dvc.lock
```
jorgeorpinel, line 104:
```suggestion
  +    │   └── prepared
  +    │       ├── test.tsv
  +    │       └── train.tsv
  +    ├── dvc.yaml
  +    ├── dvc.lock
```

content/docs/start/data-pipelines.md
jorgeorpinel: Going to remove all the extra indentation...

content/docs/start/data-pipelines.md
```diff
@@ -92,19 +92,19 @@ prepare:
...
+       │   ├── data.xml
+       │   ├── data.xml.dvc
+  +    │   └── prepared
+  +    │       ├── test.tsv
+  +    │       └── train.tsv
+  +    ├── dvc.yaml
+  +    ├── dvc.lock
+       ├── params.yaml
+       └── src
+           ├── ...
```
jorgeorpinel, line 107:
```suggestion
   .
   ├── data
   │   ├── data.xml
   │   ├── data.xml.dvc
  +│   └── prepared
  +│       ├── test.tsv
  +│       └── train.tsv
  +├── dvc.yaml
  +├── dvc.lock
   ├── params.yaml
   └── src
       ├── ...
```

content/docs/start/data-pipelines.md
```diff
@@ -92,19 +92,19 @@ prepare:
...
+   │   ├── data.xml.dvc
+  +│   └── prepared
+  +│       ├── test.tsv
+  +│       └── train.tsv
+  +├── dvc.yaml
+  +├── dvc.lock
+   ├── params.yaml
+   └── src
+       ├── ...
   ```
```
jorgeorpinel, line 108: OK so this is what `git diff` would output. Unfortunately it's still not being rendered nicely by our engine, which eats up the single trailing spaces used for indentation 🙁 

image

jorgeorpinel: This is only a problem in this block because the whole block is indented inside an MD bullet.

So we're back to square one 😒 should we manually align this? Or look for a solution in https://github.com/iterative/dvc.org/issues/1761#issuecomment-692439599?

content/docs/start/data-pipelines.md
```diff
@@ -92,19 +92,19 @@ prepare:
...
-      │   ├── data.xml.dvc
-  +   │   └── prepared
-  +   │       ├── test.tsv
-  +   │       └── train.tsv
-  +   ├── dvc.yaml
-  +   ├── dvc.lock
-      ├── params.yaml
-      └── src
-          ├── ...
+  ```
```
jorgeorpinel, line 95:
````suggestion
  ```git
````

content/docs/start/data-pipelines.md
jorgeorpinel: p.s. OK `diff` uses diff format, while `git diff` seems to use "patch" format. Prism doesn't have a `patch` language, BUT it has a `git` language, which looks pretty good for this 🙂 — will update here.

> Leaving this thread unresolved for future ref.

jorgeorpinel (approved):
The indentation problem is still happening but I think we should address that in #1761

jorgeorpinel:
Alright. Merging this for now. The formatting is still off though, see #1761.

Issue #1743, opened 2020-08-31
docs: List instances of stage file concepts, dvc run/checkout/repro for example review and update to 1.0


jeremydesroches:
From #1255 : 

> Create a list of places where the **concept of stage file(s)** and `dvc run/checkout/repro` are used...

List of places here, including initial example:

**Get Started**
- [x] [https://dvc.org/doc/start/data-pipelines](https://dvc.org/doc/start/data-pipelines)
- [x] [https://dvc.org/doc/start/experiments](https://dvc.org/doc/start/experiments)
- [x] Several cmd examples/links need updating (see **Command Reference** section below)

**User Guide**
- [x] [https://dvc.org/doc/user-guide/dvc-files-and-directories](https://dvc.org/doc/user-guide/dvc-files-and-directories)
- [x] [https://dvc.org/doc/user-guide/managing-external-data](https://dvc.org/doc/user-guide/managing-external-data)
- [x] [https://dvc.org/doc/user-guide/external-dependencies](https://dvc.org/doc/user-guide/external-dependencies)
- [x] [https://dvc.org/doc/user-guide/related-technologies](https://dvc.org/doc/user-guide/related-technologies)
- [x] [https://dvc.org/doc/user-guide/large-dataset-optimization](https://dvc.org/doc/user-guide/large-dataset-optimization)
- [x] [https://dvc.org/doc/user-guide/merge-conflicts](https://dvc.org/doc/user-guide/merge-conflicts)
- [x] [https://dvc.org/doc/user-guide/dvcignore](https://dvc.org/doc/user-guide/dvcignore)

**Use Cases**
- [x] [https://dvc.org/doc/use-cases/versioning-data-and-model-files/tutorial](https://dvc.org/doc/use-cases/versioning-data-and-model-files/tutorial)

**Install**
- [x] [https://dvc.org/doc/install/completion](https://dvc.org/doc/install/completion) - #1766

**Command Reference**
- [x] [https://dvc.org/doc/command-reference/run](https://dvc.org/doc/command-reference/run)
- [x] [https://dvc.org/doc/command-reference/repro](https://dvc.org/doc/command-reference/repro)
- [x] [https://dvc.org/doc/command-reference/remove](https://dvc.org/doc/command-reference/remove)
- [x] [https://dvc.org/doc/command-reference/freeze](https://dvc.org/doc/command-reference/freeze) - #1766
[...]

PR #1766, opened 2020-09-08
docs: replace stage file example in /install/completion, fix typos found in review

jeremydesroches:
Found in review for #1743. Replaced out-of-date `dvc repro` stage file example with `dvc plots` completion example.

### DIFF
--- a/content/docs/command-reference/install.md
+++ b/content/docs/command-reference/install.md
@@ -178,7 +178,7 @@ $ git tag
 
 These tags are used to mark points in the development of the project, and to
 document specific experiments conducted in it. To take a look at one, we
-checkout the `6-featurization` tag:
+checkout the `7-ml-pipeline` tag:
 
 ```dvc
 $ git checkout 7-ml-pipeline
--- a/content/docs/command-reference/unfreeze.md
+++ b/content/docs/command-reference/unfreeze.md
@@ -49,7 +49,7 @@ $ dvc run -n make_copy -d foo -o bar cp foo bar
 > See `dvc run` for more details.
 
 Then, let's change the file `foo` that the stage `make_copy` depends on, and
-freeze stage as well, to see what's the project status after that:
+freeze the stage as well, to see what's the project status after that:
 
 ```dvc
 $ echo zoo > foo
@@ -61,7 +61,7 @@ foo.dvc:
 ```
 
 DVC notices that `foo` changed due to the `foo.dvc` file that tracks this file
-(as `outs`), but the `make_copy` stage doesn't records the change among it's
+(as `outs`), but the `make_copy` stage no longer records the change among its
 dependencies. Run `dvc unfreeze` to get the regular/full project status:
 
 ```dvc
--- a/content/docs/command-reference/unfreeze.md
+++ b/content/docs/command-reference/unfreeze.md
@@ -61,7 +61,7 @@ foo.dvc:
 ```
 
 DVC notices that `foo` changed due to the `foo.dvc` file that tracks this file
-(as `outs`), but the `make_copy` stage no longer records the change among its
+(as `outs`), but the `make_copy` stage doesn't record the change among its
 dependencies. Run `dvc unfreeze` to get the regular/full project status:
 
 ```dvc
--- a/content/docs/install/completion.md
+++ b/content/docs/install/completion.md
@@ -27,12 +27,13 @@ run     -- Generate a stage file from a command and execute the command.
 
 Depending on what you typed on the command line so far, it completes:
 
-- Available DVC commands.
+- Available DVC commands and subcommands. For example, `dvc plots` completes
+  with `diff`, `modify`, and `show`.
 - Options (flags) that are available for a particular command.
 - File names that make sense in a given context, such as using them as a target
   for some commands.
-- Values for certain command arguments and subcommands. For example, `dvc plots`
-  completes with `diff`, `modify`, and `show`.
+- Values for certain command arguments. For example, `dvc repro` completes with
+  stages or `.dvc` files to reproduce.
 
 ## What shell do you have?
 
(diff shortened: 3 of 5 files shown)


jeremydesroches:
@jorgeorpinel for the other docs listed in #1743, do you prefer that I compile all updates in this PR or separately?

content/docs/install/completion.md
```diff
@@ -31,8 +31,8 @@ Depending on what you typed on the command line so far, it completes:
 - Options (flags) that are available for a particular command.
 - File names that make sense in a given context, such as using them as a target
   for some commands.
-- Values for certain command arguments. For example, `dvc repro` completes with
-  stage files to reproduce.
+- Values for certain command arguments and subcommands. For example, `dvc plots`
+  completes with `diff`, `modify`, and `show`.
```
jorgeorpinel, line 35: Hmmm I think this should stay as just "command arguments" and find another example. In fact you can keep `repro` just replace "stage file" for the updated term (see https://dvc.org/doc/command-reference/repro to figure that one out 😉 )

jorgeorpinel: It's a good idea to mention "subcommands" and maybe even the `plots` example too though, but in the first bullet "Available DVC commands."

jorgeorpinel:
Small PRs are preferred @jeremydesroches thanks. BTW you don't need to address everything in #1743. It's good practice so it would be good for you to get a few in each category, but it doesn't seem that related to your GSoD project [SEO, keywords, etc]. Thanks!

content/docs/install/completion.md
jeremydesroches: This makes sense, I moved the `plots` example to the first bullet as you suggest. I restored the `repro` example with updated terminology.

When started working on this the first thing I tested out was the completion for `repro`. It didn't show me available stages. I assumed this meant it was changed, but it might just be a problem with my example project.

jeremydesroches:
> Small PRs are preferred @jeremydesroches thanks. BTW you don't need to address everything in #1743. It's good practice so it would be good for you to get a few in each category, but it doesn't seem that related to your GSoD project [SEO, keywords, etc]. Thanks!

I understand. I'll do a few this week and then move on to SEO-focussed items only.

jeremydesroches:
Hey @jorgeorpinel, I added a few fixes for simple typos found along the way in completing the review for #1743. Will leave the other open items in #1743 for other PRs.

content/docs/install/completion.md
```diff
@@ -27,12 +27,13 @@ run     -- Generate a stage file from a command and execute the command.
...
 
-- Available DVC commands.
+- Available DVC commands and subcommands. For example, `dvc plots` completes
+  with `diff`, `modify`, and `show`.
 - Options (flags) that are available for a particular command.
 - File names that make sense in a given context, such as using them as a target
   for some commands.
 - Values for certain command arguments. For example, `dvc repro` completes with
-  stage files to reproduce.
+  stages or `.dvc` files to reproduce.
```
jorgeorpinel, line 36: I highly doubt the autocompletion tool reads stages from dvc.yaml files. It probably only works with .dvc files. In fact it probably completes any file in the cwd. Please try it. Editing most of these docs usually involves actually trying the tool in question. Thanks

content/docs/install/completion.md
jorgeorpinel: OK. I see you mentioned trying the tool here. Yeah it's not that smart probably. You can just say something like "existing files" here.

content/docs/command-reference/install.md
```diff
@@ -178,7 +178,7 @@ $ git tag
 
 These tags are used to mark points in the development of the project, and to
 document specific experiments conducted in it. To take a look at one, we
-checkout the `6-featurization` tag:
+checkout the `7-ml-pipeline` tag:
```
jorgeorpinel, line 181: Unrelated but ok, thanks.

content/docs/command-reference/unfreeze.md
```diff
@@ -61,7 +61,7 @@ foo.dvc:
 ```
 
 DVC notices that `foo` changed due to the `foo.dvc` file that tracks this file
-(as `outs`), but the `make_copy` stage doesn't records the change among it's
+(as `outs`), but the `make_copy` stage no longer records the change among its
```
jorgeorpinel, line 64: "No longer" would mean it did before, which isn't shown.
```suggestion
(as `outs`), but the `make_copy` stage doesn't record the change among its
```

jorgeorpinel (approved):
2 small things pending above but this is good overall. Thanks

p.s. Let's avoid adding more unrelated changes here but feel free to open other PRs as needed.

content/docs/command-reference/unfreeze.md
jeremydesroches: My mistake, I thought this made it consistent with wording `freeze` but I see now the tense should be different.

content/docs/install/completion.md
jeremydesroches: > OK. I see you mentioned trying the tool here. Yeah it's not that smart probably. You can just say something like "existing files" here.

Yes, this was the motivation for replacing the `dvc repro` example in the first place. It didn't complete with stages from dvc.yaml. It does complete with .dvc files so "existing files" makes sense.

content/docs/install/completion.md
```diff
@@ -27,12 +27,13 @@ run     -- Generate a stage file from a command and execute the command.
...
 
-- Available DVC commands.
+- Available DVC commands and subcommands. For example, `dvc plots` completes
+  with `diff`, `modify`, and `show`.
 - Options (flags) that are available for a particular command.
 - File names that make sense in a given context, such as using them as a target
   for some commands.
 - Values for certain command arguments. For example, `dvc repro` completes with
-  stage files to reproduce.
+  existing files to reproduce.
```
jorgeorpinel, line 36:
```suggestion
  existing files to use as targets.
```

jorgeorpinel (approved):

jorgeorpinel:
Thanks @jeremydesroches

Issue #1635, opened 2020-07-29
cmd: update version ref.


jorgeorpinel:
Per iterative/dvc/pull/4114

jorgeorpinel:
Another good one for when you're free @imhardikj

PR #1787, opened 2020-09-18
misc. regular updates

jorgeorpinel:
- Complete config file type table
-

### DIFF
--- a/content/docs/command-reference/config.md
+++ b/content/docs/command-reference/config.md
@@ -30,12 +30,12 @@ information (passwords, SSH keys, etc). Use the `--local` command option (flag)
 instead, to set (or override) secrets. `--global` and `--system` are also
 available to set config options for multiple projects and users, respectively.
 
-| Command option | Priority | Linux location         | Windows location                                          |
-| -------------- | -------- | ---------------------- | --------------------------------------------------------- |
-| `--local`      | 1        | `.dvc/config.local`    | `.dvc/config.local`                                       |
-| None (default) | 2        | `.dvc/config`          | `.dvc/config`                                             |
-| `--global`     | 3        | `~/.config/dvc/config` | `%LocalAppData%/iterative/dvc/config`                     |
-| `--system`     | 4        | `/etc/dvc.config`      | `%AllUsersProfile%/Application Data/iterative/dvc/config` |
+| Flag                                                | Priority | Mac location                            | Linux location             | Windows location                                          |
+| --------------------------------------------------- | -------- | --------------------------------------- | -------------------------- | --------------------------------------------------------- |
+| <span style="white-space: nowrap">`--local`</span>  | 1        | `.dvc/config.local`                     | Same                       | Same                                                      |
+| None (default)                                      | 2        | `.dvc/config`                           | Same                       | Same                                                      |
+| <span style="white-space: nowrap">`--global`</span> | 3        | `$HOME/Library/Application Support/dvc` | `$HOME/.config/dvc/config` | `%LocalAppData%/iterative/dvc/config`                     |
+| <span style="white-space: nowrap">`--system`</span> | 4        | `/Library/Application Support/dvc`      | `/etc/dvc.config`          | `%AllUsersProfile%/Application Data/iterative/dvc/config` |
 
 ## Command options (flags)
 


content/docs/command-reference/config.md
```diff
@@ -27,15 +27,20 @@ This command reads and updates the DVC configuration files. By default, the
...
+| `--local`      | 1        | `.dvc/config.local`  |
+| None (default) | 2        | `.dvc/config`        |
+
+The `--global` and `--system` flags are also available to set config options for
+multiple projects and users, respectively:
+
+| Flag                                                | Priority | Mac location                            | Linux location             | Windows location                                          |
+| --------------------------------------------------- | -------- | --------------------------------------- | -------------------------- | --------------------------------------------------------- |
+| <span style="white-space: nowrap">`--global`</span> | 3        | `$HOME/Library/Application Support/dvc` | `$HOME/.config/dvc/config` | `%LocalAppData%/iterative/dvc/config`                     |
+| <span style="white-space: nowrap">`--system`</span> | 4        | `/Library/Application Support/dvc`      | `/etc/dvc.config`          | `%AllUsersProfile%/Application Data/iterative/dvc/config` |
```
jorgeorpinel, line 43: Notice the ugly HTMl needed so that `--` isn't separated from `global/system`. I tried a bunch of MD tricks and couldn't fix this otherwise... Any ideas anyone? Cc @rogermparent

jorgeorpinel: p.s. this is how it looks right now:

image
> From https://dvc.org/doc/command-reference/config

rogermparent: You can use this snippet in the Markdown to apply this style page-wide without the ugly in-table repeats.

```
<style>
  #markdown-root td:first-child code {
    white-space: nowrap;
  }
</style>
```

This also removes the risk of collision if any other table depends on having a code block in the first column wrap.

jorgeorpinel: OK thanks Roger, I'll use that. It's also not ideal and needs an HTML comment to explain what's happening but it's better.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -50,6 +50,10 @@ folder i.e. `gdrive://<base>/path/to/folder`. The base can be one of:
    and [shared drives](https://support.google.com/a/users/answer/9310351)\*
    (these two can only be referenced by ID).
 
+   > ⚠️ Files and folders should be shared to specific users (or groups) in
```
shcheklein, line 53: Just `Folder should be ...` should be enough?

content/docs/command-reference/version.md
```diff
@@ -14,7 +14,7 @@ usage: dvc version [-h] [-q | -v]
 
 | Line                                        | Detail                                                                                                                                                                |
 | ------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| [`DVC version`](#components-of-dvc-version) | Version of DVC (along with a Git commit hash in case of a development version)                                                                                        |
+| [`DVC version`](#components-of-dvc-version) | Version of DVC (along with a Git commit hash if running a development version)                                                                                        |
```
shcheklein, line 17: It looks like the output and fields have changed significantly- do we have a ticket for this?

content/docs/command-reference/version.md
```diff
@@ -44,6 +44,8 @@ The detail of DVC version depends upon the way of installing DVC.
   [in our core repo](https://github.com/iterative/dvc/blob/master/dvc/version.py).
   For example `0.40.2`.
 
+  > 💡 To only get this number, use `dvc -V`.
```
shcheklein, line 47: No need to make it a note?

content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -28,12 +28,18 @@ perfect solution, but here are some ideas:
...
-performance when the user has over 300K files in a single directory.
+A major feature of DVC requires file links for
+[optimization](/doc/user-guide/large-dataset-optimization) purposes. In Windows,
+symlinks are available but the
+[Create symbolic links](<https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/dn221947(v=ws.11)>)
+privilege has to be enabled (using the _Local Security Policy_ console, for
+example). It's enabled for the _Administrators_ group by default, so running
+`dvc` in a terminal as an admin should do the trick for occasional use.
+
+> Note that DVC's Windows [installer](/doc/install/windows) enables symlink
```
shcheklein, line 41: better to move to the top, something like- It's enabled if you installed ... You can skip it.

content/docs/command-reference/version.md
jorgeorpinel: I think so, because it's not about `dvc version`.

> Still it feels it belongs here (as a note) more than in the cmd ref index (where `dvc -C` will be, for example).

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Agree, updated.

content/docs/command-reference/version.md
jorgeorpinel: There should be one yeah, since this was updated in the core repo... ⌛

jorgeorpinel: #1635

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: Updated.

content/docs/command-reference/config.md
```diff
@@ -27,15 +27,27 @@ This command reads and updates the DVC configuration files. By default, the
...
 meant to be tracked by Git and should not contain sensitive and/or user-specific
 information (passwords, SSH keys, etc). Use the `--local` command option (flag)
-instead, to set (or override) secrets. `--global` and `--system` are also
-available to set config options for multiple projects and users, respectively.
-
-| Command option | Priority | Linux location         | Windows location                                          |
-| -------------- | -------- | ---------------------- | --------------------------------------------------------- |
-| `--local`      | 1        | `.dvc/config.local`    | `.dvc/config.local`                                       |
-| None (default) | 2        | `.dvc/config`          | `.dvc/config`                                             |
-| `--global`     | 3        | `~/.config/dvc/config` | `%LocalAppData%/iterative/dvc/config`                     |
```
shcheklein, line 37: should we use Windows style path for Windows?
should we put escape before " " in all of these string so that it's easier to copy/paste them?

content/docs/command-reference/version.md
shcheklein: 👍

content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -28,12 +28,18 @@ perfect solution, but here are some ideas:
...
   partition (dual boot).
 
-## Disable short-file name generation
+## Enable symbolic links
 
-With NTFS, users may want to disable `8dot3` as per
-[this article](https://support.microsoft.com/en-us/help/121007/how-to-disable-8-3-file-name-creation-on-ntfs-partitions)
-to disable the short-file name generation. It is important to do so for better
-performance when the user has over 300K files in a single directory.
+> This is done automatically by the DVC's Windows
```
shcheklein, line 33: This is done automatically if you installed DVC from the installer package ... You would probably need this if you install it via other methods- pip, conda, etc

(be more explicit)

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -50,6 +50,9 @@ folder i.e. `gdrive://<base>/path/to/folder`. The base can be one of:
    and [shared drives](https://support.google.com/a/users/answer/9310351)\*
    (these two can only be referenced by ID).
 
+   > ⚠️ Folders should be shared to specific users (or groups) in order for them
```
shcheklein, line 53: Folders -> Folder

content/docs/command-reference/config.md
jorgeorpinel: Good points. Applied.

content/docs/user-guide/running-dvc-on-windows.md
jorgeorpinel: OK. Updated.

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Hmmm I think it's better in plural:

> ...including shared folders...
> Folders should be shared...

shcheklein: _Folder ID_ - means you specify a single folder. And sharing should be setup for that folder.

jorgeorpinel: I would just change this if it didn't make a difference but actually the form in which the context is written I think implies plural here:

> There's a few alternatives to construct a GDrive remote URL... such as... shared **folders** [plural]... The base can be one of:
> 1. _Folder ID_... identifier for every Google Drive folder, including shared **folders** [plural]
> > **Folders** should [plural, referring to the above mentions]

But I guess if it was "The folder" it would also make sense (referring to the folder which ID we are using — a bit indirect but OK). No strong opinion really, up to you 🙂

shcheklein: To be precise we should communicate the following:

Folder that corresponds to the _Folder ID_ mentioned must be shared to specific users (account emails) or groups in order for them to ...

jorgeorpinel: O_o I see it has a typo! "in order for them to access them with DVC" should've been just "in order to access them..." maybe that's what was confusing? Fixed now...

jorgeorpinel: > To be precise we should communicate:
Folder that corresponds to the Folder ID...

OK, I think that was implied. But I've re-written it now so it's extra explicit. PTAL

shcheklein (approved):

PR #1788, opened 2020-09-20
Document --backup import behaviour

charlesbaynham:
Documenting the changes in https://github.com/iterative/dvc/pull/4581

REVIEW UPDATE: Jump to https://github.com/iterative/dvc.org/pull/1788#issuecomment-697993141

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -105,8 +105,8 @@ from the source repo.
 
 To avoid duplicating large files, `dvc import` does not push imported files to
 any [remotes](/doc/command-reference/remote) by default. This means that when
-`dvc pull/fetch` is run, DVC retrieves imports from their source instead of from
-your remote. It also means that if the source of the import is deleted then the
+`dvc pull/fetch` is run, DVC retrieves imports from their source.
+So if the original source of an import is deleted, then the
 files are not backed up in your remote, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -84,9 +84,10 @@ the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
 
 To prevent this, you can use the `--backup` option to mark the imports to be
-pushed to a DVC remote when `dvc push` is run.  This also means that
-`dvc pull/fetch/repro` no longer access import sources, so these imports
-can be restored from the remote like regular data tracked with `dvc add`.
+pushed to a DVC remote when `dvc push` is run. This also means that
+`dvc pull/fetch/repro` no longer access import sources, going to the remote
+instead, like with regular project data tracked with `dvc add`. Use
+`dvc update` to get the (latest) data from its import source.
 
 Backup mode is enabled with a `backup: true` key/value pair in the `outs`
 section of the import `.dvc` file. To switch to regular import behavior, set
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -110,7 +110,7 @@ So if the original source of an import is deleted, then the
 files are not backed up in your remote, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
-data imports to your project's remote. This also means that `dvc pull/fetch` no
+data imports to remote storage. This also means that `dvc pull/fetch` no
 longer needs to access import sources, so interacts only with the DVC remote,
 and that backed-up imports can be restored just like files that were
 `dvc add`ed.
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -92,8 +92,8 @@ project data tracked with `dvc add`. You can still use `dvc update` to get the
 latest import data from its source.
 
 Remote storage is enabled with a `store: true` key/value pair in the `outs`
-section of the import `.dvc` file. To switch to regular import behavior, set its
-value to `false`, or re-import the data without using the `--store` option.
+section of the import `.dvc` file. To switch to regular import behavior, change
+its value to `false` manually, or re-import the data without the `--store` option.
 
 ## Options
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -76,7 +76,7 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
-### Store mode
+### Enabling remote storage
 
 To avoid duplicating large files, imported data is not pushed to
 [remote storage](/doc/command-reference/remote) by default (see `dvc push`).
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -91,8 +91,8 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
-- `-b`, `--backup` - imports files in "backup mode": see Backup Mode section for
-  details.
+- `-b`, `--backup` - turns _backup mode_ on for this import, so it can be pushed
+  to remote storage (see [Backup mode](#backup-mode) for details).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -91,7 +91,8 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
-- `-b`, `--backup` - imports files in "backup mode": see Backup Mode section for details. 
+- `-b`, `--backup` - imports files in "backup mode": see Backup Mode section for
+  details.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
@@ -103,18 +104,18 @@ from the source repo.
 ## Backup Mode
 
 To avoid duplicating large files, `dvc import` does not push imported files to
-any configured remotes for your project by default. This means that when 
-`dvc pull/fetch` is run, DVC retrieves imports from their source instead of from your
-remote. It also means that if the source of the import is deleted then the files
-are not backed up in your remote, and data may be lost. 
+any configured remotes for your project by default. This means that when
+`dvc pull/fetch` is run, DVC retrieves imports from their source instead of from
+your remote. It also means that if the source of the import is deleted then the
+files are not backed up in your remote, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
 data imports to your project's remote. This also means that `dvc pull/fetch` no
 longer needs to access import sources, so interacts only with the DVC remote,
-and that backed-up imports can be restored just like files that were 
-`dvc add`ed. 
+and that backed-up imports can be restored just like files that were
+`dvc add`ed.
 
-To update backed-up imports, use `dvc update` like usual. 
+To update backed-up imports, use `dvc update` like usual.
 
 ## Examples
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -91,9 +91,9 @@ no longer access import sources, going to the remote instead, like with regular
 project data tracked with `dvc add`. You can still use `dvc update` to get the
 latest data from its import source.
 
-Store mode is enabled with a `store: true` key/value pair in the `outs` section
-of the import `.dvc` file. To switch to regular import behavior, set its value
-to `false`, or re-import the data without using the `--store` option.
+Remote storage is enabled with a `store: true` key/value pair in the `outs`
+section of the import `.dvc` file. To switch to regular import behavior, set its
+value to `false`, or re-import the data without using the `--store` option.
 
 ## Options
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -10,7 +10,7 @@ the import.
 ## Synopsis
 
 ```usage
-usage: dvc import [-h] [-q | -v] [-o <path>] [--rev <commit>] url path
+usage: dvc import [-h] [-q | -v] [-o <path>] [--rev <commit>] [--backup] url path
 
 positional arguments:
   url              Location of DVC or Git repository to download from
@@ -91,12 +91,30 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-b`, `--backup` - imports files in "backup mode": see Backup Mode section for details. 
+
 - `-h`, `--help` - prints the usage/help message, and exit.
 
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
+
+## Backup Mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any configured remotes for your project by default. This means that when 
+`dvc pull/fetch` is run, DVC retrieves imports from their source instead of from your
+remote. It also means that if the source of the import is deleted then the files
+are not backed up in your remote, and data may be lost. 
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to your project's remote. This also means that `dvc pull/fetch` no
+longer needs to access import sources, so interacts only with the DVC remote,
+and that backed-up imports can be restored just like files that were 
+`dvc add`ed. 
+
+To update backed-up imports, use `dvc update` like usual. 
 
 ## Examples
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -87,9 +87,9 @@ data imports to a remote. This also means that `dvc pull/fetch/repro`
 no longer need to access import sources, so these imports can be restored from
 the remote like regular data tracked with `dvc add`.
 
-Imports using _backup mode_ have a flag in their `.dvc` file `outs` section:
-`backup: true`. To revert to normal import behaviour, remove this flag or set it
-to `false`.
+Backup mode is enabled with a `backup: true` key/value pair in the `outs`
+section of the import `.dvc` file. To switch to regular import behavior, set
+its value to `false` (or remove the field altogether).
 
 ## Options
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -77,8 +77,8 @@ from the source repo.
 ### Backup mode
 
 To avoid duplicating large files, `dvc import` does not push imported files to
-any [remotes](/doc/command-reference/remote) by default. This means that when
-`dvc pull` or `dvc fetch` is run, DVC retrieves imports from their source. So if
+any [remotes](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
 the original source of an import is deleted, then the files are not backed up in
 your remote, and data may be lost.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -85,9 +85,8 @@ anywhere, and data may be lost.
 
 To prevent this, you can use the `--backup` option to mark the imports to be
 pushed to a DVC remote when `dvc push` is run.  This also means that
-`dvc pull/fetch/repro`
-no longer needs to access import sources, so these imports can be restored from
-the remote like regular data tracked with `dvc add`.
+`dvc pull/fetch/repro` no longer access import sources, so these imports
+can be restored from the remote like regular data tracked with `dvc add`.
 
 Backup mode is enabled with a `backup: true` key/value pair in the `outs`
 section of the import `.dvc` file. To switch to regular import behavior, set
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -108,8 +108,8 @@ to `false`.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
-- `-b`, `--backup` - turns _backup mode_ on for this import, so it can be pushed
-  to remote storage (see [Backup mode](#backup-mode) for details).
+- `-b`, `--backup` - turns [backup mode](#backup-mode) on for the import,
+  so it can be pushed to remote storage.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -79,8 +79,8 @@ from the source repo.
 To avoid duplicating large files, `dvc import` does not push imported files to
 any [remotes](/doc/command-reference/remote) by default. This means that
 `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
-the original source of an import is deleted, then the files are not backed up in
-your remote, and data may be lost.
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
 data imports to remote storage. This also means that `dvc pull` or `dvc fetch`
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -82,9 +82,10 @@ To avoid duplicating large files, `dvc import` does not push imported files to
 the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
 
-To prevent this, you can use the `--backup` option to request DVC to backup the
-data imports to a remote. This also means that `dvc pull/fetch/repro`
-no longer need to access import sources, so these imports can be restored from
+To prevent this, you can use the `--backup` option to mark the imports to be
+pushed to the remote when `dvc push` is run. 
+This also means that `dvc pull/fetch/repro`
+no longer needs to access import sources, so these imports can be restored from
 the remote like regular data tracked with `dvc add`.
 
 Backup mode is enabled with a `backup: true` key/value pair in the `outs`
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -115,8 +115,6 @@ longer needs to access import sources, so
 these imports can be restored from remote storage like typical
 data tracked with `dvc add`.
 
-To update backed-up imports, use `dvc update` like usual.
-
 ## Examples
 
 A simple case for this command is to import a dataset from an external <abbr>DVC
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -104,7 +104,7 @@ from the source repo.
 ## Backup Mode
 
 To avoid duplicating large files, `dvc import` does not push imported files to
-any configured remotes for your project by default. This means that when
+any [remotes](/doc/command-reference/remote) by default. This means that when
 `dvc pull/fetch` is run, DVC retrieves imports from their source instead of from
 your remote. It also means that if the source of the import is deleted then the
 files are not backed up in your remote, and data may be lost.
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -83,7 +83,7 @@ the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
-data imports to a remote. This also means that `dvc pull` or `dvc fetch`
+data imports to a remote. This also means that `dvc pull/fetch/repro`
 no longer need to access import sources, so these imports can be restored from
 remote storage like typical data tracked with `dvc add`.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -78,14 +78,14 @@ from the source repo.
 
 To avoid duplicating large files, `dvc import` does not push imported files to
 any [remotes](/doc/command-reference/remote) by default. This means that when
-`dvc pull/fetch` is run, DVC retrieves imports from their source. So if the
-original source of an import is deleted, then the files are not backed up in
+`dvc pull` or `dvc fetch` is run, DVC retrieves imports from their source. So if
+the original source of an import is deleted, then the files are not backed up in
 your remote, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
-data imports to remote storage. This also means that `dvc pull/fetch` no longer
-needs to access import sources, so these imports can be restored from remote
-storage like typical data tracked with `dvc add`.
+data imports to remote storage. This also means that `dvc pull` or `dvc fetch`
+no longer need to access import sources, so these imports can be restored from
+remote storage like typical data tracked with `dvc add`.
 
 Imports using _backup mode_ have a flag in their `.dvc` file `outs` section:
 `backup: true`. To revert to normal import behaviour, remove this flag or set it
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -76,8 +76,8 @@ from the source repo.
 
 ### Backup mode
 
-To avoid duplicating large files, `dvc import` does not push imported files to
-[remote storage](/doc/command-reference/remote) by default. This means that
+To avoid duplicating large files, imported data is not pushed to
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`). This means that
 `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
 the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -87,6 +87,10 @@ data imports to remote storage. This also means that `dvc pull/fetch` no longer
 needs to access import sources, so these imports can be restored from remote
 storage like typical data tracked with `dvc add`.
 
+Imports using _backup mode_ have a flag in their `.dvc` file `outs` section:
+`backup: true`. To revert to normal import behaviour, remove this flag or set it
+to `false`.
+
 ## Options
 
 - `-o <path>`, `--out <path>` - specify a path to the desired location in the
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -113,7 +113,8 @@ its value to `false` manually, or re-import the data without the `--store` optio
   > example below).
 
 - `-s`, `--store` - [enables remote storage](#enabling-remote-storage) for the
-  import, so it can be pushed (e.g. in case the original source is lost).
+  import, so it can be pushed and pulled like regular project data (e.g. in
+  case the original source is lost).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -77,8 +77,9 @@ from the source repo.
 ### Backup mode
 
 To avoid duplicating large files, imported data is not pushed to
-[remote storage](/doc/command-reference/remote) by default (see `dvc push`). This means that
-`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`).
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
 the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -10,7 +10,7 @@ the import.
 ## Synopsis
 
 ```usage
-usage: dvc import [-h] [-q | -v] [-o <path>] [--rev <commit>] [--backup] url path
+usage: dvc import [-h] [-q | -v] [-o <path>] [--rev <commit>] [--store] url path
 
 positional arguments:
   url              Location of DVC or Git repository to download from
@@ -74,24 +74,24 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
-### Backup mode
+### Store mode
 
 To avoid duplicating large files, imported data is not pushed to
 [remote storage](/doc/command-reference/remote) by default (see `dvc push`).
 This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
-their source (found in the `deps` field of the `.dvc` file). So if
-the original source of an import is deleted, then the files are not backed up
-anywhere, and data may be lost.
+their source (found in the `deps` field of the `.dvc` file). So if the original
+source of an import is deleted, then the files are not backed up anywhere, and
+data may be lost.
 
-As an alternative, you can use the `--backup` option so the import is
-pushed to a DVC remote when `dvc push` is run. This also means that
-`dvc pull/fetch/repro` no longer access import sources, going to the remote
-instead, like with regular project data tracked with `dvc add`. You can still
-use `dvc update` to get the latest data from its import source.
+As an alternative, you can use the `--store` option so the import is pushed to a
+DVC remote when `dvc push` is run. This also means that `dvc pull/fetch/repro`
+no longer access import sources, going to the remote instead, like with regular
+project data tracked with `dvc add`. You can still use `dvc update` to get the
+latest data from its import source.
 
-Backup mode is enabled with a `backup: true` key/value pair in the `outs`
-section of the import `.dvc` file. To switch to regular import behavior, set
-its value to `false`, or re-import the data without using the `--backup` option.
+Store mode is enabled with a `store: true` key/value pair in the `outs` section
+of the import `.dvc` file. To switch to regular import behavior, set its value
+to `false`, or re-import the data without using the `--store` option.
 
 ## Options
 
@@ -110,8 +110,8 @@ its value to `false`, or re-import the data without using the `--backup` option.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
-- `-b`, `--backup` - turns [backup mode](#backup-mode) on for the import,
-  so it can be pushed to remote storage (in case the original source is lost).
+- `-s`, `--store` - turns [store mode](#store-mode) on for the import, so it can
+  be pushed to remote storage (e.g. in case the original source is lost).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -83,7 +83,7 @@ the original source of an import is deleted, then the files are not backed up
 anywhere, and data may be lost.
 
 To prevent this, you can use the `--backup` option to request DVC to backup the
-data imports to remote storage. This also means that `dvc pull` or `dvc fetch`
+data imports to a remote. This also means that `dvc pull` or `dvc fetch`
 no longer need to access import sources, so these imports can be restored from
 remote storage like typical data tracked with `dvc add`.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -77,7 +77,7 @@ from the source repo.
 ### Backup mode
 
 To avoid duplicating large files, `dvc import` does not push imported files to
-any [remotes](/doc/command-reference/remote) by default. This means that
+[remote storage](/doc/command-reference/remote) by default. This means that
 `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
 
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+## Backup Mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any configured remotes for your project by default. This means that when
```
jorgeorpinel, line 107:
```suggestion
any remotes for your project by default. This means that when
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+## Backup Mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that when
+`dvc pull/fetch` is run, DVC retrieves imports from their source instead of from
+your remote. It also means that if the source of the import is deleted then the
```
jorgeorpinel, line 109:
```suggestion
`dvc pull/fetch` is run, DVC retrieves imports from their source.
So if the original source of an import is deleted, then the
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
+So if the original source of an import is deleted, then the
+files are not backed up in your remote, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to your project's remote. This also means that `dvc pull/fetch` no
+longer needs to access import sources, so interacts only with the DVC remote,
+and that backed-up imports can be restored just like files that were
+`dvc add`ed.
+
+To update backed-up imports, use `dvc update` like usual.
```
jorgeorpinel, line 118: No new info, it seems 🙂 
```suggestion
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
+## Backup Mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that when
+`dvc pull/fetch` is run, DVC retrieves imports from their source.
+So if the original source of an import is deleted, then the
+files are not backed up in your remote, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to your project's remote. This also means that `dvc pull/fetch` no
```
jorgeorpinel, line 113:
```suggestion
data imports to remote storage. This also means that `dvc pull/fetch` no
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
+any [remotes](/doc/command-reference/remote) by default. This means that when
+`dvc pull/fetch` is run, DVC retrieves imports from their source.
+So if the original source of an import is deleted, then the
+files are not backed up in your remote, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to remote storage. This also means that `dvc pull/fetch` no
+longer needs to access import sources, so interacts only with the DVC remote,
+and that backed-up imports can be restored just like files that were
+`dvc add`ed.
```
jorgeorpinel, line 116:
```suggestion
longer needs to access import sources, so
these imports can be restored from remote storage like typical
data tracked with `dvc add`.
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
+  details.
+
 - `-h`, `--help` - prints the usage/help message, and exit.
 
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+## Backup Mode
```
jorgeorpinel, line 104:
```suggestion
## Backup mode
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
+files are not backed up in your remote, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to remote storage. This also means that `dvc pull/fetch` no
+longer needs to access import sources, so
+these imports can be restored from remote storage like typical
+data tracked with `dvc add`.
+
+To update backed-up imports, use `dvc update` like usual.
+
```
jorgeorpinel, line 119: Let's put this whole section as an H3 inside the **Description**, just before **Options**, please.

jorgeorpinel: - [ ] Also, it should probably briefly describe what happens in the .dvc file (is there a new field?) and how to remove backup mode.
- [ ] Info about this field should also be added to https://dvc.org/doc/user-guide/dvc-files-and-directories#dvc-files please.

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-b`, `--backup` - imports files in "backup mode": see Backup Mode section for
+  details.
```
jorgeorpinel, line 95:
```suggestion
- `-b`, `--backup` - turns _backup mode_ on for this import, so it can be pushed
  to remote storage (see [Backup mode](#backup-mode) for details).
```

content/docs/command-reference/import.md
```diff
@@ -91,13 +91,32 @@ from the source repo.
...
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+## Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that when
+`dvc pull/fetch` is run, DVC retrieves imports from their source.
```
jorgeorpinel, line 108: > when `dvc pull/fetch` is run, DVC retrieves imports from their source.

I didn't know that. Are you sure? I think you have to `dvc repro`. Please try it out 🙂

jorgeorpinel (changes requested):
Thanks a lot for this! Great start. Needs some clarification and expansion still. Oh, and formatting! Please see https://dvc.org/doc/user-guide/contributing/docs#development-environment

content/docs/command-reference/import.md
charlesbaynham: It does! That was the motivation for this change actually, since I have a lot of git imports and all the repos were getting cloned each time I ran `dvc pull`. As a test, I ran `dvc pull -v` in a repo with a single import in it, like this:

```
md5: ca4361da725ee43b1e01ffe75f96b42a
frozen: true
deps:
- path: file.txt
  repo:
    url: ../imp
    rev_lock: f7d942abf891b4a0c29f8dec9e6728870862b012
outs:
- md5: f701279bc9ac499afb1f556e22c23cb2
  path: file.txt
```

Even after the import and the cache were present and up to date, the output contains

```
...
2020-09-20 22:20:08,091 DEBUG: Collecting information from local cache...
2020-09-20 22:20:08,094 DEBUG: Creating external repo ../imp@f7d942abf891b4a0c29f8dec9e6728870862b012                                                                                
2020-09-20 22:20:08,095 DEBUG: erepo: git clone '../imp' to a temporary dir
2020-09-20 22:20:08,192 DEBUG: Saving '../../../../private/var/folders/s8/gg_nx8mn23sbqxbj1021tn6c0000gn/T/tmpbg9ac34ldvc-clone/file.txt' to '.dvc/cache/f7/01279bc9ac499afb1f556e22c23cb2'.
...
```

content/docs/command-reference/import.md
charlesbaynham: Done and done. I already had some text in content/docs/user-guide/dvc-files-and-directories.md, though not loads. Is that fine?

content/docs/command-reference/import.md
jorgeorpinel: Interesting. I should know this 😅 but also `repro` will do it so we just need to mention that one too in the sentence, please.

I'll make a separate issue for us to mention that fetch and pull re-import data...

jorgeorpinel: > Even after the import and the cache were present and up to date (the repo is cloned)

On this, did DVC actually re-download the data? OR just clone the repo to check what the latest version was? Because if the former case, it sound like a bug to report in the core repo (shouldn't re-download imports if you already have the latest version).

content/docs/command-reference/import.md
jorgeorpinel: Great, thanks. I didn't see those changes in the other file, yes that's enough! I'll take it from there 🙂

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `backup`: Whether or not this file or directory is backed-up to project
+  remotes. (`true` by default for all `outs` except imports, where it is `false`
+  by default). Mostly relevant for imports in
```
jorgeorpinel, line 76: Can it be set to `false` for regular .dvc files? (Non-imports)

jorgeorpinel: I feel like it should probably only have an effect for imports (and if I;m right, we can say so here 🙂)

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that when
+`dvc pull` or `dvc fetch` is run, DVC retrieves imports from their source. So if
```
jorgeorpinel, line 81:
```suggestion
any [remotes](/doc/command-reference/remote) by default. This means that
`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+the original source of an import is deleted, then the files are not backed up in
+your remote, and data may be lost.
```
jorgeorpinel, line 83:
```suggestion
the original source of an import is deleted, then the files are not backed up
anywhere, and data may be lost.
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
...
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to remote storage. This also means that `dvc pull` or `dvc fetch`
```
jorgeorpinel, line 86: > `dvc import` does not push imported files

Will `import` do the pushing when using --backup? Or that just tells `push` to do it when/if run? I'm guessing the latter but it's not clear from the text above (also in the 2nd paragraph). Feel free to update it or just lmk what the answer is and I will 👍

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+any [remotes](/doc/command-reference/remote) by default. This means that
```
jorgeorpinel, line 80:
```suggestion
[remote storage](/doc/command-reference/remote) by default. This means that
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
...
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+[remote storage](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to remote storage. This also means that `dvc pull` or `dvc fetch`
```
jorgeorpinel, line 86:
```suggestion
data imports to a remote. This also means that `dvc pull` or `dvc fetch`
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
...
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+[remote storage](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to a remote. This also means that `dvc pull` or `dvc fetch`
```
jorgeorpinel, line 86:
```suggestion
data imports to a remote. This also means that `dvc pull/fetch/repro`
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
...
+To avoid duplicating large files, `dvc import` does not push imported files to
+[remote storage](/doc/command-reference/remote) by default. This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to a remote. This also means that `dvc pull/fetch/repro`
+no longer need to access import sources, so these imports can be restored from
+remote storage like typical data tracked with `dvc add`.
```
jorgeorpinel, line 88:
```suggestion
the remote like regular data tracked with `dvc add`.
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +108,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-b`, `--backup` - turns _backup mode_ on for this import, so it can be pushed
+  to remote storage (see [Backup mode](#backup-mode) for details).
```
jorgeorpinel, line 112:
```suggestion
- `-b`, `--backup` - turns [backup mode](#backup-mode) on for the import,
  so it can be pushed to remote storage.
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,23 @@ Note that import stages are considered always
...
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to request DVC to backup the
+data imports to a remote. This also means that `dvc pull/fetch/repro`
+no longer need to access import sources, so these imports can be restored from
+the remote like regular data tracked with `dvc add`.
+
+Imports using _backup mode_ have a flag in their `.dvc` file `outs` section:
+`backup: true`. To revert to normal import behaviour, remove this flag or set it
+to `false`.
```
jorgeorpinel, line 92:
```suggestion
Backup mode is enabled by the `backup: true` key/value pair set in the `outs`
section of the import `.dvc` file. To switch to regular import behavior, set its
value to `false` (or remove the field altogether).
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +108,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-b`, `--backup` - turns [backup mode](#backup-mode) on for the import,
+  so it can be pushed to remote storage.
```
jorgeorpinel, line 112:
```suggestion
  so it can be pushed to remote storage (in case the original source is lost).
```

jorgeorpinel (commented):
Almost! Just a couple more doubts here and I'll wrap it up.

content/docs/command-reference/import.md
charlesbaynham: I think just clone to check but, in the case of git imports, that's the same since you need to redownload the entire git repo to clone it.

content/docs/user-guide/dvc-files-and-directories.md
charlesbaynham: @efiop and I had a bit of a chat about this already - see https://github.com/iterative/dvc/issues/4527 for a summary. 

My thinking is that it's good to try and avoid special treatment for particular types of file where possible: it's better to have the behaviour explicit in the .dvc file. The code PR https://github.com/iterative/dvc/pull/4581 really removes special treatment for `outs` which result from imports vs `outs` which result from adds or pipelines: the difference is now that the former have `backup: false` whereas the latter have `backup: true`. To me it seems odd that something in the `deps` part of the file would affect how the `outs` are treated, but that's what you have to do if you want to treat imports as special cases. 

It would also be surprising for the user if you had `push: false` in a normal dvc file but which still resulted in that file getting pushed to the remote. And hey, if someone wants a normal dvc added file which doesn't get pushed to remotes, why shouldn't we let them have one?

content/docs/command-reference/import.md
charlesbaynham: Yes, the latter: I'll tweak

content/docs/command-reference/import.md
jorgeorpinel: Hmmm. I think it could/should just clone the bare repo and check the latest revision where the file was modified (blame) is <= the `rev` in the .dvc file (`deps.repo.rev` field, to be precise.) Anyway... Do you see any more -v output that indicates whether the file was actually pulled/checked-out from the Git remote? It may be hard to determine, actually. I can take a look later if not so don;t worry too much about this! 🙂 Thanks!

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > it seems odd that something in the deps part of the file would affect how the outs are treated, but that's what you have to do if you want to treat imports as special cases

Yeah OK, makes sense. From an implementation perspective at least

> surprising for the user if you had push: false in a normal dvc file but which still resulted in that file getting pushed

But DVC would never add backup:false to regular .dvc files so users doing it manually should also be an edge case (in which case they should know about the field and the docs would explain it only affects imports)? But anyway, let's just document the way it works for now, and maybe I'll jump into the discussions in the core repo.

> if someone wants a normal dvc added file which doesn't get pushed to remotes, why shouldn't we let them have one?

Good point.

---
BTW:

> The idea is for imports with --backup to be severed from their import source completely unless the user calls dvc update explicitly.

- [ ] Great explanation in that ticket! We should probably include this sentence more or less as is in the docs 🙂

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,24 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Backup mode
+
+To avoid duplicating large files, `dvc import` does not push imported files to
+[remote storage](/doc/command-reference/remote) by default. This means that
```
jorgeorpinel, line 80:
```suggestion
To avoid duplicating large files, imported data is not pushed to
[remote storage](/doc/command-reference/remote) by default (see `dvc push`). This means that
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,24 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Backup mode
+
+To avoid duplicating large files, imported data is not pushed to
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`). This means that
+`dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from their source. So if
```
jorgeorpinel, line 81:
```suggestion
[remote storage](/doc/command-reference/remote) by default (see `dvc push`).
This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
their source (found in the `deps` field of the `.dvc` file). So if
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +110,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-b`, `--backup` - turns [backup mode](#backup-mode) on for the import,
+  so it can be pushed to remote storage (in case the original source is lost).
```
jorgeorpinel, line 114: What about `import-url`, does it have `--backup`? Probably should?

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,25 @@ Note that import stages are considered always
...
+To avoid duplicating large files, imported data is not pushed to
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`).
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to mark the imports to be
+pushed to the remote when `dvc push` is run. 
+This also means that `dvc pull/fetch/repro`
```
jorgeorpinel, line 88:
```suggestion
pushed to a DVC remote when `dvc push` is run.  This also means that
`dvc pull/fetch/repro`
```

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: p.s. I see now you were trying to add that info about `update` but I removed it haha sorry — it didn't seem necessary the way it was written. I'll re-add it.

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,25 @@ Note that import stages are considered always
...
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to mark the imports to be
+pushed to a DVC remote when `dvc push` is run.  This also means that
+`dvc pull/fetch/repro`
+no longer needs to access import sources, so these imports can be restored from
+the remote like regular data tracked with `dvc add`.
```
jorgeorpinel, line 90:
```suggestion
`dvc pull/fetch/repro` no longer access import sources, so these imports can be restored from
the remote like regular data tracked with `dvc add`.
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,24 @@ Note that import stages are considered always
...
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`).
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to mark the imports to be
+pushed to a DVC remote when `dvc push` is run.  This also means that
+`dvc pull/fetch/repro` no longer access import sources, so these imports
+can be restored from the remote like regular data tracked with `dvc add`.
```
jorgeorpinel, line 89:
```suggestion
pushed to a DVC remote when `dvc push` is run. This also means that
`dvc pull/fetch/repro` no longer access import sources, going to the remote
instead, like with regular project data tracked with `dvc add`. Use
`dvc update` to get the (latest) data from its import source.
```

jorgeorpinel:
From iterative/dvc/issues/4527:

> The idea is for imports with --backup to be severed from their import source completely unless the user calls dvc update explicitly.

Does `dvc update` (or running `dvc import` again without `--backup`) remove the backup flag?

jorgeorpinel (approved):
Still some questions we can get to as the core PR is still under review. But for the most part this looks ready 🙂 

Thanks again @charlesbaynham! Great stuff

jorgeorpinel:
Thanks again @charlesbaynham! Great stuff

content/docs/command-reference/import.md
charlesbaynham: `import-url` always made files in "backup mode", and still does. I haven't provided any option to disable backup mode for it though. I could do, or alternatively its behaviour could be documented.

charlesbaynham:
> 
> 
> From [iterative/dvc/issues/4527](https://github.com/iterative/dvc/issues/4527):
> 
> > The idea is for imports with --backup to be severed from their import source completely unless the user calls dvc update explicitly.
> 
> Does `dvc update` (or running `dvc import` again without `--backup`) remove the backup flag?

`dvc update` won't remove the flag, it'll behave as expected and retrieve the latest version from the update source. This will then be pushed to / retrieved from the remote. So the file are "severed" from the update source in that you don't need access to the source to pull them from your remote, but the link still exists so that you can go get the latest version when you want to (with `dvc update`).

I don't actually know what happens if you call `dvc import` again without the flag. I should find out

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
...
+  [backup mode](/doc/command-reference/import#backup-mode).
```
jorgeorpinel, line 77:
```suggestion
- `backup`: Whether or not this file or directory is pushed to
  [remote storage](/doc/command-reference/remote) (`true` by default for
  all outputs except for imports, where it is `false` by default). See
  [import backup mode](/doc/command-reference/import#backup-mode).
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `backup`: Whether or not this file or directory is pushed to
```
jorgeorpinel, line 74:
```suggestion
- `backup`: Whether or not this file or directory is pushed and pulled to/from
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `backup`: Whether or not this file or directory is pushed and pulled to/from
+  [remote storage](/doc/command-reference/remote) (`true` by default for
```
jorgeorpinel, line 75: Wait... So what happens to a regular output after it has been pushed if you then set it to backup:false manually, and run `pull`? Do `pull/fetch` throw an error? What's the message?

charlesbaynham: Good catch: there wasn't one, it was an unhandled error. Now, you get an error for `pull/fetch` if the file isn't already in your local cache:

```
ERROR: failed to pull data from the cloud - Checkout failed for following targets:
file.txt
Is your cache up to date?
```

content/docs/command-reference/import.md
jorgeorpinel: Ah yeah I was confused. No need to mention backup mode in import-url, as it's explained in the DVC metafiles doc.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Nice, thanks. Will follow up in the core PR.

jorgeorpinel:
From https://github.com/iterative/dvc/pull/4581#issuecomment-696276154

> you can use import to alter the settings of the import, or update to retrigger it with whatever settings it was set up with

- [ ] Seems as expected. We should prob add that info to the docs somewhere, though.

content/docs/command-reference/import.md
jorgeorpinel: So yes, `dvc pull` does download and overwrite the existing cached file even if it's the latest version (sounds like a bug to me, opened iterative/dvc/issues/4597). But only for Git imports, I think.

Anyway, resolving here!

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,25 @@ Note that import stages are considered always
...
+
+To avoid duplicating large files, imported data is not pushed to
+[remote storage](/doc/command-reference/remote) by default (see `dvc push`).
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+To prevent this, you can use the `--backup` option to mark the imports to be
+pushed to a DVC remote when `dvc push` is run. This also means that
```
jorgeorpinel, line 87:
```suggestion
As an alternative, you can use the `--backup` option so the import is
pushed to a DVC remote when `dvc push` is run. This also means that
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,25 @@ Note that import stages are considered always
...
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if
+the original source of an import is deleted, then the files are not backed up
+anywhere, and data may be lost.
+
+As an alternative, you can use the `--backup` option so the import is
+pushed to a DVC remote when `dvc push` is run. This also means that
+`dvc pull/fetch/repro` no longer access import sources, going to the remote
+instead, like with regular project data tracked with `dvc add`. Use
+`dvc update` to get the (latest) data from its import source.
```
jorgeorpinel, line 90:
```suggestion
instead, like with regular project data tracked with `dvc add`. You can still
use `dvc update` to get the latest data from its import source.
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +74,25 @@ Note that import stages are considered always
...
+
+As an alternative, you can use the `--backup` option so the import is
+pushed to a DVC remote when `dvc push` is run. This also means that
+`dvc pull/fetch/repro` no longer access import sources, going to the remote
+instead, like with regular project data tracked with `dvc add`. You can still
+use `dvc update` to get the latest data from its import source.
+
+Backup mode is enabled with a `backup: true` key/value pair in the `outs`
+section of the import `.dvc` file. To switch to regular import behavior, set
+its value to `false` (or remove the field altogether).
```
jorgeorpinel, line 94:
```suggestion
its value to `false`, or re-import the data without using the `--backup` option.
```

jorgeorpinel:
- [ ] We'll probably want to update the notes about pull/push for imports being added in #1801 (import(-url) refs, external deps guide)

content/docs/command-reference/import.md
```diff
@@ -74,6 +76,25 @@ Note that import stages are considered always
 they won't be updated. Use `dvc update` to update the downloaded data artifact
 from the source repo.
 
+### Store mode
```
jorgeorpinel, line 79:
```suggestion
### Enabling remote storage
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +112,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-s`, `--store` - turns [store mode](#store-mode) on for the import, so it can
+  be pushed to remote storage (e.g. in case the original source is lost).
```
jorgeorpinel, line 116:
```suggestion
- `-s`, `--store` - [enables remote storage](#enabling-remote-storage) on for
  the import, so it can be pushed (e.g. in case the original source is lost).
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +76,25 @@ Note that import stages are considered always
...
+
+As an alternative, you can use the `--store` option so the import is pushed to a
+DVC remote when `dvc push` is run. This also means that `dvc pull/fetch/repro`
+no longer access import sources, going to the remote instead, like with regular
+project data tracked with `dvc add`. You can still use `dvc update` to get the
+latest data from its import source.
+
+Store mode is enabled with a `store: true` key/value pair in the `outs` section
+of the import `.dvc` file. To switch to regular import behavior, set its value
+to `false`, or re-import the data without using the `--store` option.
```
jorgeorpinel, line 96:
```suggestion
Remote storage is enabled with a `store: true` key/value pair in the `outs`
section of the import `.dvc` file. To switch to regular import behavior, set its
value to `false`, or re-import the data without using the `--store` option.
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `store`: Whether or not this file or directory is pushed and pulled to/from
+  [remote storage](/doc/command-reference/remote) (`true` by default for all
+  outputs except for imports, where it is `false` by default). See
+  [import store mode](/doc/command-reference/import#store-mode).
```
jorgeorpinel, line 77:
```suggestion
  [import storage](/doc/command-reference/import#enabling-remote-storage).
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,16 @@ Unable to detect supported link types, as the
...
+Unable to checkout this output since it is not present in the cache and has been
+marked to not be stored on remotes.
+
+If `store: false` is added to the `outs` section of a `.dvc` file, that output
+will not be stored on remotes when `dvc push` is used, and dvc will not attempt
+to retreive it when `dvc pull` or `dvc fetch` is called. This error means that
+the file is not already in the local cache, so dvc cannot find it. You must
+provide the file locally since it is not tracked on dvc-managed remote storage.
+
+See [import store mode](/doc/command-reference/import#store-mode).
```
jorgeorpinel, line 68:
```suggestion
See [import storage](/doc/command-reference/import#enabling-remote-storage).
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +76,25 @@ Note that import stages are considered always
...
+This means that `dvc pull`, `dvc fetch`, and `dvc repro` retrieve imports from
+their source (found in the `deps` field of the `.dvc` file). So if the original
+source of an import is deleted, then the files are not backed up anywhere, and
+data may be lost.
+
+As an alternative, you can use the `--store` option so the import is pushed to a
+DVC remote when `dvc push` is run. This also means that `dvc pull/fetch/repro`
+no longer access import sources, going to the remote instead, like with regular
+project data tracked with `dvc add`. You can still use `dvc update` to get the
+latest data from its import source.
```
jorgeorpinel, line 92:
```suggestion
latest import data from its source.
```

content/docs/command-reference/import.md
```diff
@@ -74,6 +76,25 @@ Note that import stages are considered always
...
+
+As an alternative, you can use the `--store` option so the import is pushed to a
+DVC remote when `dvc push` is run. This also means that `dvc pull/fetch/repro`
+no longer access import sources, going to the remote instead, like with regular
+project data tracked with `dvc add`. You can still use `dvc update` to get the
+latest import data from its source.
+
+Remote storage is enabled with a `store: true` key/value pair in the `outs`
+section of the import `.dvc` file. To switch to regular import behavior, set its
+value to `false`, or re-import the data without using the `--store` option.
```
jorgeorpinel, line 96:
```suggestion
section of the import `.dvc` file. To switch to regular import behavior, change
its value to `false` manually, or re-import the data without using the `--store` =
option.
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +112,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-s`, `--store` - [enables remote storage](#enabling-remote-storage) on for
+  the import, so it can be pushed (e.g. in case the original source is lost).
```
jorgeorpinel, line 116:
```suggestion
- `-s`, `--store` - [enables remote storage](#enabling-remote-storage) for the
  import, so it can be pushed (e.g. in case the original source is lost).
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +112,9 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-s`, `--store` - [enables remote storage](#enabling-remote-storage) for the
+  import, so it can be pushed (e.g. in case the original source is lost).
```
jorgeorpinel, line 116:
```suggestion
  import, so it can be pushed and pulled (e.g. in case the original source is
  lost).
```

content/docs/command-reference/import.md
```diff
@@ -91,6 +112,10 @@ from the source repo.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `-s`, `--store` - [enables remote storage](#enabling-remote-storage) for the
+  import, so it can be pushed and pulled like regular project data (e.g. in
```
jorgeorpinel, line 116: Maybe link pushed and pulled to the cmd refs?

jorgeorpinel: Yeah
```suggestion
  import, so it can be [pushed](/doc/command-reference/push) and 
  [pulled](/doc/command-reference/pull) like regular project data (e.g. in
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,10 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `store`: Whether or not this file or directory is pushed and pulled to/from
+  [remote storage](/doc/command-reference/remote) (`true` by default for all
+  outputs except for imports, where it is `false` by default). See
+  [import storage](/doc/command-reference/import#enabling-remote-storage).
```
jorgeorpinel, line 77:
```suggestion
- `store`: Whether this file or directory can be
  [pushed](/doc/command-reference/push) and
  [pulled](/doc/command-reference/pull) to/from
  [remote storage](/doc/command-reference/remote) (`true` by default for
  all outputs except for imports, where it is `false` by default). See
  [import storage](/doc/command-reference/import#enabling-remote-storage).
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,16 @@ Unable to detect supported link types, as the
 [cache directory](/doc/command-reference/config#cache) doesn't exist. It is
 usually created automatically by DVC commands that need it, but you can create
 it manually (e.g. `mkdir .dvc/cache`) to enable this check.
+
+## Checkout failed with `store: false` files {#checkout-failed-no-store}
```
jorgeorpinel, line 57:
```suggestion
## Checkout failed with `store: false` outputs {#checkout-no-store}
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,16 @@ Unable to detect supported link types, as the
...
 it manually (e.g. `mkdir .dvc/cache`) to enable this check.
+
+## Checkout failed with `store: false` outputs {#checkout-no-store}
+
+Unable to checkout this output since it is not present in the cache and has been
+marked to not be stored on remotes.
+
+If `store: false` is added to the `outs` section of a `.dvc` file, that output
+will not be stored on remotes when `dvc push` is used, and dvc will not attempt
+to retreive it when `dvc pull` or `dvc fetch` is called. This error means that
```
jorgeorpinel, line 64:
```suggestion
DVC is unable to checkout any output that is not present in the
<abbr>cache</abbr>.This can usually be solved by fetching from
[import storage](/doc/command-reference/remote). However, If `store: false`
is used in a `.dvc` file, `dvc pull` and `dvc fetch` won't attempt to retrieve the
data from a DVC remote.

This error means that
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,17 @@ Unable to detect supported link types, as the
...
+<abbr>cache</abbr>.This can usually be solved by fetching from
+[import storage](/doc/command-reference/remote). However, If `store: false`
+is used in a `.dvc` file, `dvc pull` and `dvc fetch` won't attempt to retrieve the
+data from a DVC remote.
+
+This error means that
+the file is not already in the local cache, so dvc cannot find it. You must
+provide the file locally since it is not tracked on dvc-managed remote storage.
+
+See [import storage](/doc/command-reference/import#enabling-remote-storage).
```
jorgeorpinel, line 69:
```suggestion
To solve this, you must provide the file locally.

See also
[Enabling remote storage](/doc/command-reference/import#enabling-remote-storage)
for imports.
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,17 @@ Unable to detect supported link types, as the
...
 usually created automatically by DVC commands that need it, but you can create
 it manually (e.g. `mkdir .dvc/cache`) to enable this check.
+
+## Checkout failed with `store: false` outputs {#checkout-no-store}
+
+DVC is unable to checkout any output that is not present in the
+<abbr>cache</abbr>.This can usually be solved by fetching from
+[import storage](/doc/command-reference/remote). However, If `store: false`
+is used in a `.dvc` file, `dvc pull` and `dvc fetch` won't attempt to retrieve the
+data from a DVC remote.
```
jorgeorpinel, line 63: We may need to update this title, anchor, and description a little per https://github.com/iterative/dvc/pull/4581#pullrequestreview-501492896...

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,17 @@ Unable to detect supported link types, as the
...
+
+To solve this, you must provide the file locally.
```
jorgeorpinel, line 65: Is it possible to suggest more elaborate possible solution(s) that will cover most of the cases? Not sure what the options are but off the top of my head: push the data from the original repo; remove store: false and try fetch/pull again; re-import; what else? If you can just write them here I'll copy edit it as needed. Thanks!

jorgeorpinel (changes requested):
Will need another round of updates but almost there. Maybe let's wait until the core PR is finalized too? Oh and it needs merge of master branch (solve conflicts) 😅

jorgeorpinel:
- [ ] Oh, and after merging `master`, we should probably also update this new note that is now on https://dvc.org/doc/command-reference/import:

image

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -71,6 +71,12 @@ An _output entry_ (`outs`) consists of these fields:
   _checksum_ for HDFS.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `store`: Whether this file or directory can be
+  [pushed](/doc/command-reference/push) and
+  [pulled](/doc/command-reference/pull) to/from
+  [remote storage](/doc/command-reference/remote) (`true` by default for
+  all outputs except for imports, where it is `false` by default). See
```
jorgeorpinel, line 78: > except for imports

Per https://github.com/iterative/dvc.org/issues/1865#issuecomment-714433118 this isn't the only exception. External outputs are also never stored remotely (since they are already in an external cache).

content/docs/user-guide/troubleshooting.md
```diff
@@ -53,3 +53,17 @@ Unable to detect supported link types, as the
 [cache directory](/doc/command-reference/config#cache) doesn't exist. It is
 usually created automatically by DVC commands that need it, but you can create
 it manually (e.g. `mkdir .dvc/cache`) to enable this check.
+
+## Checkout failed with `store: false` outputs {#checkout-no-store}
+
+DVC is unable to checkout any output that is not present in the
+<abbr>cache</abbr>.This can usually be solved by fetching from
```
jorgeorpinel, line 60:
```suggestion
<abbr>cache</abbr>. This can usually be solved by fetching from
```

jorgeorpinel:
Closing this as stale. We can reopen when/if the core PR gets merged. Thanks

PR #1794, opened 2020-09-22
Skip running deploy check on master.

rogermparent:
It will never check any links because we're comparing to master.

Also, this change provides a convenient excuse to test the deploy check on a new PR!

### DIFF
--- a/.github/workflows/link-check-deploy.yml
+++ b/.github/workflows/link-check-deploy.yml
@@ -2,14 +2,18 @@ name: Check new links against deployment
 # This workflow "triggers" and skips on deployment because GitHub Actions /
 # Checks refuses to show the check on deployment_status
 on:
-  - deployment
-  - deployment_status
+  deployment:
+    branches-ignore:
+      - master
+  deployment_status:
+    branches-ignore:
+      - master
 
 jobs:
   run:
     name: Initialize
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/link-check-deploy.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   run:
     name: Initialize
     runs-on: ubuntu-latest
-    if: github.event.deployment_status.state == 'success'
+    if: github.ref != 'master' && github.event.deployment_status.state == 'success'
```
shcheklein, line 12: extra space at the end is not needed

shcheklein (approved):

shcheklein:
Actually you can do something like this?

```
on:
  push:
    branches-ignore:
      - master
```

rogermparent:
> Actually you can do something like this?
> 
> ```
> on:
>   push:
>     branches-ignore:
>       - master
> ```

I'm not sure if that's an option on the `deployment` and `deployment_status` triggers, but I can try it.

rogermparent:
While we can't really tell if it's right for sure until this PR hits master, the fact that GitHub Actions doesn't reject `branches-ignore` as malformed config is a good sign!

shcheklein:
Okay, merging! Let's see I guess ...

rogermparent:
Master check was skipped, looks like it worked!

shcheklein:
👍

PR #1795, opened 2020-09-22
No deploy check on master

rogermparent:
Turns out `branches-ignore` syntax didn't work on `deployment_status`, as seen [here](https://github.com/iterative/dvc.org/actions/runs/266344304)

### DIFF
--- a/.github/workflows/link-check-deploy.yml
+++ b/.github/workflows/link-check-deploy.yml
@@ -2,14 +2,18 @@ name: Check new links against deployment
 # This workflow "triggers" and skips on deployment because GitHub Actions /
 # Checks refuses to show the check on deployment_status
 on:
-  - deployment
-  - deployment_status
+  deployment:
+    branches-ignore:
+      - master
+  deployment_status:
+    branches-ignore:
+      - master
 
 jobs:
   run:
     name: Initialize
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


shcheklein (approved):

.github/workflows/link-check-deploy.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   run:
     name: Initialize
     runs-on: ubuntu-latest
-    if: github.event.deployment_status.state == 'success'
+    if: github.ref != 'master' && github.event.deployment_status.state == 'success'
```
shcheklein, line 12: should we remove ignore branch?

rogermparent: It is removed, I assume it looks different on the default view because I reused the old branch

See the commit that removes it [here](https://github.com/iterative/dvc.org/pull/1795/commits/5e7889a3b58877a8bbe4f386f3866011e89a8843)

shcheklein: I merged it and it hasn't been not removed.

rogermparent: Sorry about that, I guess I have to merge master back when re-using a branch and the file changes tab was more trustworthy than I thought

PR #1801, opened 2020-09-23
misc. regular updates

### LINKED ISSUE #1792: sync: mention that fetch and pull re-import data
> Extracted from https://github.com/iterative/dvc.org/pull/1788#pullrequestreview-492211092

`dvc fetch` and `dvc pull` (re)download import data from their original source as indicated in the corresponding .dvc files created with `dvc import` or `dvc import-url`.

This should be clearly mentioned, perhaps even emphasized in the docs (in all the cmd refs involved, at least).

jorgeorpinel:
- Update `status` sample outputs (rel=iterative/dvc/pull/4490)
- Other copy edits around term "import"
- Notes about pull/push for imports (fixes #1792)
- Other copy edits
- Rewrote section about remote aliases in external dependencies guide


content/docs/user-guide/merge-conflicts.md
```diff
@@ -139,4 +139,4 @@ outs:
   - path: data.xml
 ```
 
-And then `dvc update` the `.dvc` file.
+And then `dvc pull` the import (`dvc.xml` in the example above).
```
jorgeorpinel, line 142: Hi @efiop, can you confirm this is actually more correct?
Rationale: update may bring an even newer version and overwrite the hash you selected during merge, while pull will just make sure you have the correct version in cache and workspace.
Thanks

content/docs/command-reference/get.md
```diff
@@ -36,16 +36,16 @@ the data source. Both HTTP and SSH protocols are supported for online repos
 to an "offline" repo (if it's a DVC repo without a default remote, instead of
 downloading, DVC will try to copy the target data from its <abbr>cache</abbr>).
 
+⚠️ Online repos should have a default
```
shcheklein, line 39: it's not only about online repos though

content/docs/user-guide/external-dependencies.md
```diff
@@ -220,6 +203,10 @@ Importing 'model.pkl (git@github.com:iterative/example-get-started)'
 The command above creates `model.pkl.dvc`, where the external dependency is
 specified (with the `repo` field).
 
+⚠️ DVC won't push or pull imported data to
```
shcheklein, line 206: feels excessive to explain the details in every example. That's what descriptions for to my mind

content/docs/user-guide/external-dependencies.md
```diff
@@ -243,3 +230,24 @@ The `url` and `rev_lock` subfields under `repo` are used to save the origin and
 [version](https://git-scm.com/docs/revisions) of the dependency, respectively.
 
 </details>
+
+## Example: DVC remote aliases
+
+If instead of a URL you'd like to use an alias that can be managed
```
shcheklein, line 236: should we explain why would someone want to manage them independently?

content/docs/user-guide/merge-conflicts.md
efiop: @jorgeorpinel The example shows deleting the hash, so `dvc update` is more correct, as it will record the new hash.

jorgeorpinel: Hmmm true. But still then you may get a newer version which may not be the goal of this merge. I'll add a pull option and keep update as well.

jorgeorpinel: Wait no, that's probably too difficult. OK I'll just add a note the implications of updating.

content/docs/user-guide/merge-conflicts.md
```diff
@@ -103,27 +103,22 @@ To resolve conflicted `.dvc` files generated by `dvc import` or
 `dvc import-url`, remove the conflicted hashes altogether:
 
 ```yaml
-< < < < < < < HEAD
-md5: 263395583f35403c8e0b1b94b30bea32
-=======
-md5: 520d2602f440d13372435d91d3bfa176
-> > > > > > > branch
 frozen: true
```
jorgeorpinel, line 106: @efiop another quick Q: this was added accidentally right? As .dvc files no longer have a top `md5` field. Just double checking

content/docs/command-reference/get.md
jorgeorpinel: Here's the whole idea (text fragments from the paragraph above):

> Both HTTP and SSH protocols are supported for online repos
> `url` can also be a local file system path
> Online repos should have a default DVC remote

I'm not sure what is missing. Or are you agains the term "online repo"? We could come up with another way of saying it.

jorgeorpinel: Yeah I guess the whole paragraph is too complicated, let me see how to re-write it and avoid terms online/offline.

jorgeorpinel: OK done in e6359056. PTAL

shcheklein: yep, LGTM!

content/docs/user-guide/external-dependencies.md
jorgeorpinel: This doc is a guide about external deps, there's no description per se, just a **How it works** section and I don't think this note would make any sense there. Importing is only mentioned in the examples.

jorgeorpinel: But I guess we don't really need these notes in this doc, as remote storage is also not the topic, and they're in the import/-url cmd refs. Removing.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Sure. But I'm not sure what the answer is. Let me think about that.

> p.s. I only moved this section, it's not new.

jorgeorpinel: OK, done in 1e75862e, check it out when you can. Thanks

jorgeorpinel:
Sorry, this PR got too large somehow (I had an irregular week). But 9/16 files only have small indentation or typo changes. Hopefully it's not too much to review, but lmk if I should split it. Thanks

content/docs/user-guide/merge-conflicts.md
```diff
@@ -103,27 +103,22 @@ To resolve conflicted `.dvc` files generated by `dvc import` or
...
 deps:
 - path: get-started/data.xml
   repo:
     url: https://github.com/iterative/dataset-registry
 < < < < < < < HEAD
     rev_lock: f31f5c4cdae787b4bdeb97a717687d44667d9e62
-=======
+= = = = = = =
```
shcheklein, line 113: they should be w/o spaces - https://wincent.com/wiki/Understanding_Git_conflict_markers ?

content/docs/user-guide/merge-conflicts.md
```diff
@@ -139,4 +134,8 @@ outs:
   - path: data.xml
 ```
 
-And then `dvc update` the `.dvc` file.
+And then `dvc update` the `.dvc` file to download the latest data from its
+original source.
+
+> Note that updating will bring in the latest version of the data found in its
```
shcheklein, line 140: @efiop could you confirm this?

content/docs/command-reference/import-url.md
```diff
@@ -109,8 +109,12 @@ $ dvc run -n download_data \
           wget https://data.dvc.org/get-started/data.xml -O data.xml
 ```
 
-`dvc import-url` generates an import stage `.dvc` file and `dvc run` a regular
-stage (in `dvc.yaml`).
+`dvc import-url` generates an <abbr>import stage</abbr> `.dvc` file and
+`dvc run` a regular stage (in `dvc.yaml`).
+
+⚠️ DVC won't push or pull imported data to
```
shcheklein, line 115: `pull to remote` sounds strange

content/docs/user-guide/external-dependencies.md
```diff
@@ -146,15 +146,18 @@ $ dvc run -n download_file \
...
-If instead of a URL you'd like to use an alias that can be managed
-independently, or if the external dependency location requires access
-credentials, you may use `dvc remote add` to define this location as a DVC
-Remote, and then use a special URL with format `remote://{remote_name}/{path}`
-to define an external dependency.
+You may want to encapsulate external locations as configurable entities that can
+be managed independently. This is useful if multiple dependencies (or stages)
+reuse the same location, or if its likely to change in the future. And if the
+location requires authentication, you need a way to configuring it in order to
+access the data.
```
shcheklein, line 155: also users might redefine using `--local` config, thus using their own external space

shcheklein (approved):
looks good, some comments to fix, otherwise ready to merge

content/docs/user-guide/merge-conflicts.md
jorgeorpinel: The problem is that most IDEs will think it's an actual conflict if we save it like that. Maybe even GitHub or some code analyzers will pick it up as a problem.

jorgeorpinel: And we can't use HTML encoding since it's in a md code block.

content/docs/command-reference/import-url.md
jorgeorpinel: Oops, thanks. Updated to "to/from" in f23fc5ce.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: True. That seems a too `dvc remote`-specific to add in this paragraph though, IMO. The concept of "DVC remote" isn't even mentioned yet, and it's not really explained in this section.

jorgeorpinel: BUT I could change the example from HTTP to something that requires setting up credentials to give a hint of that possibility... ⌛

jorgeorpinel: Done in d16cd33d.

jorgeorpinel:
Thanks! Merging this but I can do a quick followup if any of the recent comments above require more iterations to be addressed.

content/docs/user-guide/external-dependencies.md
```diff
@@ -146,27 +146,39 @@ $ dvc run -n download_file \
...
+`remote://{remote_name}/{path}` (remote alias) to define the external
+dependency.
+
+Let's see an example using SSH. First, register and configure the remote:
+
+```dvc
+$ dvc remote add myssh ssh://myserver.com
+$ dvc remote modify --local myssh user myuser
+$ dvc remote modify --local myssh password mypassword
+```
```
jorgeorpinel, line 168: This makes me realize our SSH examples for `remote add/modify` may be misleading (in other docs) because we sometimes use `ssh://user@example.com...` when adding, and then mention about modifying the credentials or even show an example of `remote modify example user ...` but I don't know if that will work correctly with DVC. Need to check...

content/docs/user-guide/external-dependencies.md
```diff
@@ -146,27 +146,39 @@ $ dvc run -n download_file \
...
+
+> Please refer to `dvc remote add` for more details like setting up access
+> credentials for the different remote types.
+
+Now, use an alias to this remote when defining the stage:
 
 ```dvc
```
jorgeorpinel, line 175: I also realized this doc only uses `dvc run` to define stages and doesn't show the resulting dvc.yaml files (except for import(-url) examples. Maybe we shouldn't even use `dvc run` in most of these examples? Since manually writing dvc.yaml is the recommended way.

But then again you can't easily write an external dependency in dvc.yaml ... Hmmm 🤔

jorgeorpinel (commented):
A couple things I noticed based on the changes above.

content/docs/user-guide/merge-conflicts.md
efiop: @jorgeorpinel This is an import file, so they do have the md5 and so this change is wrong :slightly_frowning_face:

content/docs/user-guide/merge-conflicts.md
efiop: Yes, correct.

content/docs/user-guide/merge-conflicts.md
jorgeorpinel: - [ ] OK np I'll revert it.

Why do import .dvc files need a top md5 though? We may need to document this in https://dvc.org/doc/user-guide/dvc-files-and-directories or in the import(-url) refs. at least. Thanks

efiop: We probably will be removing it in the future, it has been in discussions for a while now.

Indeed, looks like https://dvc.org/doc/user-guide/dvc-files-and-directories doesn't have imports at all :slightly_frowning_face:

content/docs/user-guide/external-dependencies.md
jorgeorpinel: I tried testing this but can't because of https://github.com/iterative/dvc/issues/4712.

PR #1803, opened 2020-09-23
Add ability to override SEO description in docs pages

### LINKED ISSUE #1785: docs: add metadata markup capability to docs template so docs can have unique SEO descriptions and metadata
**Requirements**
- [ ] default meta description for /doc items is blank (no description)
- [ ] when front matter tags are added to /doc items then meta tags are added, following /blog pattern
- [ ] front matter tags are optional, e.g., only description, only title, title + description, none. Unused tags don't need to be added to front matter.

**What?**
All docs pages currently use the same meta description and it's not possible to customize it for each doc. In the existing blog template, unique metadata is added by using markup at the top of each markdown file as follows:

`title: August '20 Community Gems`
`description: |
  A roundup of technical Q&A's from the DVC community...`

I propose adding the same capability to the docs template so that docs can have unique metadata (if added to the front matter). 

**Why?**
SEO analyzers/site audits flag this as an issue because a unique (and well written) description often improves clickthrough rates on search results, which leads to better rankings. 

There are currently ~90 /doc pages on the dvc.org domain with duplicate descriptions. Adding unique descriptions to all/most of the docs could have a material positive impact on overall rankings in the medium to long term. 

Adding this feature would allow for a partial fix for two open general SEO issues: it's a best practice to have unique descriptions (#1118), which could then be updated on all docs and
- [ ] the meta description tags considered "reviewed" (#1119).

If the docs template was updated to use custom metadata like blog posts, then
- [ ] #857 re: content previews should also be addressed.
  > The content preview (e.g., `og:description`) is not unique for the same reason: different tags but generated from the same text. 😉 

This should function as described in #982 ("easily replace default og-tags in header with the ones needed for current page"), but if no custom tags are added to a given doc, the default tags should remain.

**How?**
I investigated how this is working for blog posts and tried adding similar metadata to a doc on my local... just in case the renderer picked it up (it didn't 🙃). So it appears the docs template and page generation need to be updated.

To fix the duplicates issue, only the meta description would be required. There is no need or benefit to customizing the meta title for most of the docs as the title is simple and should only reference the command or feature. However, for tutorials, use cases, or other long-form docs it would be good to have the option to edit the meta title for SEO purposes.

rogermparent:
This PR adds the ability to use `description` and `title` frontmatter fields to override the SEO description and title of any docs page, in the same way as the Blog pages.

Since the Blog and Docs use the same SEO component, all the same SEO tags are used. If either is not provided, the page will fall back on the same SEO title and description it had before.

Fixes #1785 

Can be verified on the deploy preview with test data [here](https://dvc-landing-docs-descri-ytzq28.herokuapp.com/doc/start). I'll remove it before merge.

### DIFF
--- a/content/docs/start/index.md
+++ b/content/docs/start/index.md
@@ -1,4 +1,5 @@
 ---
+title: 'Get Started | DVC'
 description: 'Get started with DVC!'
 ---
 
--- a/content/docs/start/index.md
+++ b/content/docs/start/index.md
@@ -1,8 +1,3 @@
----
-title: 'Get Started | DVC'
-description: 'Get started with DVC!'
----
-
 # Get Started
 
 Assuming DVC is already [installed](/doc/install), let's initialize it by
--- a/content/docs/start/index.md
+++ b/content/docs/start/index.md
@@ -1,3 +1,7 @@
+---
+description: 'Get started with DVC!'
+---
+
 # Get Started
 
 Assuming DVC is already [installed](/doc/install), let's initialize it by
--- a/src/gatsby/models/docs/createSchemaCustomization.js
+++ b/src/gatsby/models/docs/createSchemaCustomization.js
@@ -11,7 +11,9 @@ async function createSchemaCustomization(api) {
       interfaces: ['Node'],
       fields: {
         ...markdownParentFields,
-        template: 'String'
+        template: 'String',
+        title: 'String',
+        description: 'String'
       }
     })
   ]
--- a/src/gatsby/models/docs/onCreateMarkdownContentNode.js
+++ b/src/gatsby/models/docs/onCreateMarkdownContentNode.js
@@ -18,7 +18,9 @@ async function createMarkdownDocsNode(api, { parentNode, createChildNode }) {
     slug,
     rawMarkdownBody: node.rawMarkdownBody,
     sourcePath: relativePath,
-    template: node.frontmatter.template
+    template: node.frontmatter.template,
+    title: node.frontmatter.title === '' ? null : node.frontmatter.title,
+    description: node.frontmatter.description
   }
 
   const docNode = {
--- a/src/templates/doc.tsx
+++ b/src/templates/doc.tsx
@@ -31,8 +31,6 @@ const DocPage: React.FC<IDocPageProps> = ({
 
   const { label } = getItemByPath(slug)
 
-  console.log({ title, description })
-
   return (
     <>
       <SEO title={title || label} description={description} />
--- a/src/templates/doc.tsx
+++ b/src/templates/doc.tsx
@@ -11,6 +11,8 @@ interface IDocPageProps {
   data: {
     page: {
       htmlAst: Node
+      title?: string
+      description?: string
     }
   }
   pageContext: {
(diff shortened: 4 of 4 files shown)


content/docs/start/index.md
```diff
@@ -1,3 +1,7 @@
+---
+description: 'Get started with DVC!'
```
jorgeorpinel, line 2: Let's try this:
```suggestion
title: 'Get Started | DVC'
description: 'Get started with DVC!'
```

rogermparent: I believe this will end up as "Get Started | DVC | DVC", as the bit is auto-added in the SEO component's helmet setup.

jorgeorpinel: We'll know soon... ⌛ (I'll report back.)

shcheklein:
So, can we and should we auto-generate these descriptions? We should also check how other tools, search engines do this.

rogermparent:
It's possible to auto-generate descriptions, and even make auto-generation manually overridable, but I don't know what template or rules would work to make a description out of here. There isn't much metadata to work with in these docs pages other than the page name.

shcheklein:
@rogermparent btw, why don't we see the link check after the deployment? No changes to content files?

shcheklein:
@rogermparent 

> but I don't know what template or rules would work to make a description out of here. There isn't much metadata to work with in these docs pages other than the page name.

that's a good task to figure this out indeed. I would first check existing libs, existing doc systems.

content/docs/start/index.md
```diff
@@ -1,3 +1,8 @@
+---
+title: 'Get Started | DVC'
+description: 'Get started with DVC!'
+---
```
jorgeorpinel, line 4: Seems to work 🙂 

See https://dvc-landing-docs-descri-xlzjup.herokuapp.com/doc/start
image

- [ ] Let's remove the changes to this file before merge.

jorgeorpinel (approved):

rogermparent:
> @rogermparent btw, why don't we see the link check after the deployment? No changes to content files?

Good catch, I'll have to figure that out.

jorgeorpinel:
> There isn't much metadata to work with in these docs

There's the title (H1 header), usually the first paragraph is a good intro (we could truncate it into a preview), and there are `<abbr>` terms in case we also want to add keywords (cc @jeremydesroches WDYT?)

All could be overwritten manually with this frontmatter.

jeremydesroches:
> > There isn't much metadata to work with in these docs
> 
> There's the title (H1 header), usually the first paragraph is a good intro (we could truncate it into a description), and there are `<abbr>` terms in case we also want to add keywords (cc @jeremydesroches WDYT?)
> 
> All could be overwritten manually with this frontmatter.

The behaviour as described in OP and as tested looks good. 

I don't think generating descriptions is needed. I think it's okay to leave the standard fallback if a custom description isn't added. 

For SEO purposes, leaving the description _blank_ is actually better than generating one (if there isn't a custom description available). When the description is left blank, Google and other search engines will do their own generation based on search query and document content — likely to be better than what we can generate.

rogermparent:
> For SEO purposes, leaving the description _blank_ is actually better than generating one (if there isn't a custom description available). When the description is left blank, Google and other search engines will do their own generation based on search query and document content — likely to be better than what we can generate.
>
> Also worth clarifying: the description doesn't directly contribute to how a page is indexed as it's not being checked for keywords, etc. It contributes indirectly through its effect on clickthrough and bounce, which are key engagement metrics that affect the index position (rank). A good description can increase likelihood of clickthrough and, if well matched to the content, lower the bounce rate.

Interesting, I never knew this. I could tweak our current setup to do this either everywhere or just on docs pages if that's desirable.

jorgeorpinel:
> leaving the description blank is actually better... search engines will do their own generation

Great! Let's leave the heavy lifting to the SEOs then!

jorgeorpinel:
>  I could tweak our current setup to use blank descriptions either on docs pages or everywhere

WDYT @shcheklein? If yes, let's create a separate issue though. Thanks!

shcheklein (approved):

PR #1804, opened 2020-09-23
Add Link Check status badge to README

rogermparent:
I didn't think the long workflow name would also reflect in the badge. We may want to rename the workflow to make the badge look better, as it doesn't seem to be configurable otherwise.

We can also use this branch for fixes for existing links, as adding the badge while failing may not be a good look.

Check out how it looks [on this branch's README](https://github.com/iterative/dvc.org/tree/link-check-status-badge#readme)


shcheklein:
yeah, up to you on naming (to my mind even PR current action names look different and longer then anything else). Not a big deal I guess.

PR #1815, opened 2020-09-26
remote modify [ssh]: Add doc about the parameter 'allow_agent'

gcoter:
This parameter was created per iterative/dvc/issues/4186 (for https://github.com/iterative/dvc/issues/4186)

### DIFF
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -539,10 +539,9 @@ more information.
   $ dvc remote modify myremote gss_auth true
   ```
 
-- `allow_agent` - whether to connect to the SSH agent (default is `true`).
-  Please refer to
-  [Paramiko's documentation](http://docs.paramiko.org/en/stable/api/client.html?highlight=allow_agent#paramiko.client.SSHClient.connect)
-  for more information.
+- `allow_agent` - whether to use [SSH agents](https://www.ssh.com/ssh/agent)
+  (`true` by default). Setting this to `false` is useful when `ssh-agent` is
+  causing problems, such as a "No existing session" error:
 
   ```dvc
   $ dvc remote modify myremote allow_agent false
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -545,7 +545,7 @@ more information.
   for more information.
 
   ```dvc
-  $ dvc remote modify myremote allow_agent true
+  $ dvc remote modify myremote allow_agent false
   ```
 
 </details>
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -539,8 +539,8 @@ more information.
   $ dvc remote modify myremote gss_auth true
   ```
 
-- `allow_agent` - whether to connect to the SSH agent (default is true). Please
-  refer to
+- `allow_agent` - whether to connect to the SSH agent (default is `true`).
+  Please refer to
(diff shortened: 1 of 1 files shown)


efiop (approved):

content/docs/command-reference/remote/modify.md
```diff
@@ -539,6 +539,15 @@ more information.
   $ dvc remote modify myremote gss_auth true
   ```
 
+- `allow_agent` - whether to connect to the SSH agent (default is true). Please
+  refer to
+  [Paramiko's documentation](http://docs.paramiko.org/en/stable/api/client.html?highlight=allow_agent#paramiko.client.SSHClient.connect)
+  for more information.
+
+  ```dvc
+  $ dvc remote modify myremote allow_agent true
```
efiop, line 548: Just to show off how to change the default
```suggestion
  $ dvc remote modify myremote allow_agent false
```

shcheklein (approved):

content/docs/command-reference/remote/modify.md
```diff
@@ -539,6 +539,15 @@ more information.
   $ dvc remote modify myremote gss_auth true
   ```
 
+- `allow_agent` - whether to connect to the SSH agent (default is `true`).
+  Please refer to
+  [Paramiko's documentation](http://docs.paramiko.org/en/stable/api/client.html?highlight=allow_agent#paramiko.client.SSHClient.connect)
+  for more information.
```
jorgeorpinel, line 545: OK thanks! This is all correct but it doesn't really explain what this is / why someone would use it. Let me see if I can research this a bit ⌛ 

> p.s. We probably don't need that Paramiko link? Again, doesn't explain much, and it's too low-level.

jorgeorpinel: From iterative/dvc/issues/4186:

> with an SSH remote, sometimes I get an error like `No existing session`... This is only on one specific machine...
> To fix it, I have to give the parameter allow_agent=False to paramiko.

OK so is this config option meant for that one specific problem? (If so we can add a section about it to https://dvc.org/doc/user-guide/troubleshooting) Or can it be generalized? Let me keep digging... ⌛

jorgeorpinel: https://linux.die.net/man/1/ssh-agent
> ssh-agent is a program to hold private keys used for public key authentication
(OpenSSH)

http://docs.paramiko.org/en/stable/api/client.html?highlight=allow_agent#paramiko.client.SSHClient.connect (so I guess this link is relevant after all)
> Authentication is attempted in the following order of priority:
> - The pkey or key_filename...
> - Any key we can find through an SSH agent
> - Any (default-named) key... in ~/.ssh/
> - Plain username/password auth, if... given

So basically, setting `allow_agent=false` is a workaround for misconfigured SSH agents running in a machine, right?

jorgeorpinel: If so, here's my suggestion:

```suggestion
- `allow_agent` - whether to use the SSH agent (`true` by default). Setting this to `false` is useful when a running `ssh-agent` is causing problems, such as a "No existing session" error.
```

> Notice that I decided to remove the link because I don't think DVC's authentication order of priority reflect that of `SSHClient.connect` — it depends on how the remote is configured. But no strong opinion.

- And I do think we should add a Troubleshooting section for this, if others agree (doesn't have to happen in this PR though, we can just open a new issue about it in this repo for now).

jorgeorpinel: OK I'm going to commit this and ask @shcheklein to merge it, if he agrees. But anyone has feedback on this please lmk or feel free to continue editing the text. Thanks

jorgeorpinel (approved):
Thanks again, @gcoter !

shcheklein (approved):

PR #1821, opened 2020-09-28
September Community Gems PR

elleobrien:


shcheklein:
@rogermparent what would the best way to fix the link in this PR? - https://anaconda.org/conda-forge/vega-lite-cli .. could we mimic browser better?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+  and outputs, and how to save models and data in continuous integration.
+picture: 2020-09-28/Gems_Sept_20.png
+author: elle_obrien
+commentsUrl: https://discuss.dvc.org/t/september-20-community-gems/512
+tags:
+  - Discord
+  - Gems
+  - CML
+  - Hyperparameters
+  - External
```
shcheklein, line 20: External Data? or do we have any similar tags already?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+Yes, you can have as many separate parameter files as you'd like. It's only
+important that they are correctly specified in your DVC pipeline stages.
+
+For example, if you have files `params_data_processing.yaml` and
+`params_model.yaml` in your project (perhaps to store hyperparameters of your
+data processing and model fitting stages, respectively), you'll want to call the
+right file at each stage. For example:
+
+```dvc
+dvc run -n preprocess \
```
shcheklein, line 54: should start with $

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+### [Q: I'm confused about external dependencies and outputs. What's the difference?](https://discordapp.com/channels/485586884165107732/485596304961962003/752478399326453840)
+
+First, let's talk about **external outputs**. You might use an external output
+when you have a dataset in storage that is so large, you can't reasonably
+transfer it to your local workspace to start DVC-tracking it. For example, if
+you have a TB of data in S3 storage, you want to treat it as an external output
+so DVC can version it without resorting to time-consuming data transfer
+operations.
+
+External outputs are configured using using `dvc add --external`
```
shcheklein, line 95: all good, except `dvc add` is not the only way ... `dvc run -o s3://something` also an external output.

In general, we can try to describe it here from a high level perspective- external outputs and dependencies are files, directories that you would `dvc add` or `dvc run -o` or `dvc run -d` but that physically reside outside of the local workspace (e.g. data is too large, or we are running a pipeline that writes directly to cloud, etc). To some extent I like to compare this with a "virtual workspace" or "expanded workspace". It just happened that some files that you add, produce, etc are located in some place outside.

We should mention that this is an advanced scenario. And we recommend to ask us on Discord for any help or confirm that this is needed, or if there are some performance bottlenecks.

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+your workspace- DVC will try to use the cache/link type you set up in the remote
+location where the external data and cache are.
+
+Now, let's talk about **external dependencies**. An external dependency might
+come into play when you run a DVC pipeline stage that depends on a file or
+directory in a remote location. For example, a pipeline stage could depend on a
+`model.pkl` file in your S3 bucket:
+
+```dvc
+$ dvc run -d s3://mybucket/model.pkl
```
shcheklein, line 112: it's probably better to use a dataset here, not end result (model)

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+`params_model.yaml` in your project (perhaps to store hyperparameters of your
+data processing and model fitting stages, respectively), you'll want to call the
+right file at each stage. For example:
+
+```dvc
+dvc run -n preprocess \
+  -p params_data_process.yaml:param1,param2,...
+```
+
+### [Q: Is there a way to automatically produce SVG plots from `dvc plot`? I don't like having to click through the Vega-Lite GUI to get an SVG, and my plots look so small when I access them in the browser.](https://discordapp.com/channels/485586884165107732/563406153334128681/750012082149392414)
```
shcheklein, line 58: Looks like two question? Should we split them?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+title: September '20 Community Gems
+date: 2020-09-28
+description: |
+  A roundup of technical Q&A's from the DVC community. This month, we discuss 
+  customizing your DVC plots, the difference between external dependencies
+  and outputs, and how to save models and data in continuous integration.
+descriptionLong: |
+  A roundup of technical Q&A's from the DVC community. This month, we discuss 
+  customizing your DVC plots, the difference between external dependencies
+  and outputs, and how to save models and data in continuous integration.
```
shcheklein, line 11: 💅 continuous integration (CI) ?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+[known issue](https://stackoverflow.com/questions/51955990/base64-decoding-error-incorrect-padding-when-loading-putty-ppk-private-key-to)
+that `paramiko` expects RSA keys in OpenSSH key format, and can throw an error
+if the keys are in an alternative format (such as default PuTTY formatted keys).
+If this is the case, you'll likely see:
+
+```
+ERROR: unexpected error - ('... ssh-rsa ...=', Error('Incorrect padding',))
+```
+
+To fix this, convert your RSA key to the OpenSSH format. Tools like PuTTYgen and
```
shcheklein, line 40: put links to these tools?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+in your Git repo.
+
+We created the DVC `run-cache` in part
+[to solve this issue](https://stackoverflow.com/questions/61245284/is-it-necessary-to-commit-dvc-files-from-our-ci-pipelines).
+Here's how it works: you'll setup a DVC remote with access credentials passed to
+your GitHub Action/GitLab CI via CML. Then you'll use the following protocol in
+your CI workflow (your workflow config file in GitHub/GitLab):
+
+```dvc
+dvc pull --run-cache
```
shcheklein, line 167: start with $

shcheklein: also, it might be a good idea on CI to run things w/o run cache (to make sure that run is independent?) ... or at least it should be using a separate remote storage that only CI has access to

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+of your CI workflow to commit any new artifacts to your Git repository. However,
+auto-commits have a lot of downsides- they don't make sense for a lot of users,
+and generally, it's better to re-create outputs as needed than save them forever
+in your Git repo.
+
+We created the DVC `run-cache` in part
+[to solve this issue](https://stackoverflow.com/questions/61245284/is-it-necessary-to-commit-dvc-files-from-our-ci-pipelines).
+Here's how it works: you'll setup a DVC remote with access credentials passed to
+your GitHub Action/GitLab CI via CML. Then you'll use the following protocol in
+your CI workflow (your workflow config file in GitHub/GitLab):
```
shcheklein, line 164: can we put a link to some workflow file?

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+
+```dvc
+dvc pull --run-cache
+dvc repro
+dvc push --run-cache
+```
+
+When you use this design, any artifacts of `dvc repro`, such as models or
+transformed datasets, will be saved in DVC storage and indexed by the pipeline
+version that generated them. You can access them via `dvc checkout` in your
```
shcheklein, line 174: but first you would need to pull run cache locally as well ... also, does `dvc checkout` actually restores things from the run cache ... I thought only repro does that

content/blog/2020-09-28-september-20-community-gems.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+If you're installing CML as part of your workflow, you may need to install Node
+first-
+[check out our docs](https://github.com/iterative/cml#install-cml-as-a-package)
+for how to do this in GitHub Actions and GitLab CI.
+
+### [Q: After running a GitHub Action workflow that runs a DVC pipeline, I want to save the output of the pipeline. Why doesn't CML automatically save the output?](https://discordapp.com/channels/485586884165107732/728693131557732403/757686601953312988)
+
+What you're talking about sounds like an auto-commit: a `git commit` at the end
```
shcheklein, line 154: should add some context here- why commit is needed at all?

rogermparent:
> @rogermparent what would the best way to fix the link in this PR? - https://anaconda.org/conda-forge/vega-lite-cli .. could we mimic browser better?

I believe all we need to do is broaden when the situations where we retry with GET, as we only do it for the specific code that towarddatascience throws while servers reject HEAD in multiple different ways. I've started on a modification that I'll push up.

rogermparent:
After the link check update, all new links pass here!

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: yep that's right. checkout only works if you've pulled locally.

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: We have neither, and nothing similar, so it's up to us. What do you prefer?

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: the question is about the difference between external outputs and external dependencies. i feel like it's not clear that they're very different, except in one potential case you need a cache (when you're versioning a remote dataset on cloud storage that you never move). 

do you think we should not write as if there's a difference between external outputs and dependencies? In other words, say they are the same thing?

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: I think this works as one, would rather keep it together since the two parts are so related. It's just about two different ways to handle customizing your plots (always go straight to SVG, or keep as html and fix size)

shcheklein: 👍

content/blog/2020-09-28-september-20-community-gems.md
shcheklein: Terminology with outputs is pretty bad to mind. Output term comes from the pipelines and is being used even for a regular `dvc add something` while in this case from a user perspective this can be input to the pipeline, raw data. So, at the end of the day- they both represent a file or a directory somewhere outside that you don't want to copy locally to start using it. It can be `dvc add something` and we call it external output (😢 ), or it can be used in place of a dependency in `dvc run -d` - we call it external dependency then, or it can be an output/intermediate artifact of the pipeline `dvc run -o` - we call it external output as well.

From the cache/non-cache they behave exactly like regular local DVC-tracked directories, files and not-necessarily-DVC-tracked files and directories. For outputs you define cache and DVC puts into cache, for dependencies - we don't care - we just save hash to check if they changed.

So, I would they are not the same from a lot of angles. They are conceptually are very similar to regular files and directories that just happen to be located and cache outside.

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: Added 

```
By design, artifacts generated in a CI workflow aren't saved anywhere- they disappear as soon as the runner shuts down. So a DVC pipeline executed in your CI system might produce outputs, like transformed datasets and model files, that will be lost at the end of the run. If you want to save them, there are a few methods. 
```

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: Hm, are you thinking of a scenario where you pull run-cache, then `dvc repro` pulls an output from the cache and skips executing that stage? Then it's possible that the stage could not execute on the runner, but you don't catch the error because it is never attempted. 

Is that what you had in mind?

content/blog/2020-09-28-september-20-community-gems.md
elleobrien: ahhh i see. yes it seems like the terminology should be reworked for clarity, eventually. but not the worst thing, we can work with it here.

elleobrien: think i got it in the revision!

shcheklein (approved):
Let's ship it !! 😎

rogermparent:
I just checked out the failing link check, and found that it's due to master being ahead of this PR, making it such that the diff between them is as if this is "adding" a failing link that's removed in master.

Merging master results in a successful check.

PR #1825, opened 2020-09-30
Link check upgrade

rogermparent:
Bump link check version to 0.7, with changes to the GET->HEAD retry system such that any failure on HEAD is retried with GET (previously this only happened on 405)

Also, this PR changes the GitHub link settings pattern to include the apex domain (github.com as opposed to www.github.com or api.github.com).


shcheklein (approved):

shcheklein:
@rogermparent 👍

Issue #1827, opened 2020-09-30
community: Contribute should have a clear link to docs/"Contributing to DVC"


brainless:
Hello everyone,

As an interested contributor, when I click on top navigation `Community > Contribute` I should see a clear CTA or link to [Contributing to DVC](https://dvc.org/doc/user-guide/contributing/core). This being the starting page for developers who are interested should be visible in the main sites Contribute page in my opinion. If the team agrees, then I will create a PR.

Thank you

jorgeorpinel:
Good idea @brainless ! Maybe where it says "We welcome contributions to our code and docs." it could be changed a little to link to both https://dvc.org/doc/user-guide/contributing/core and to https://dvc.org/doc/user-guide/contributing/docs.

iswarmondal:
Hey is this issue still active? 
I'm new to open source and want to give it a try.
What do you think @jorgeorpinel?

jorgeorpinel:
Pull requests are very much welcome @iswarmondal . Please see https://dvc.org/doc/contributing/docs first.

PR #1835, opened 2020-10-02
Added links to contribute pages in the Community header

sarthakforwet:
References #1827.

### DIFF
--- a/src/components/Community/Contribute/index.tsx
+++ b/src/components/Community/Contribute/index.tsx
@@ -31,6 +31,24 @@ const Contribute: React.FC<{ theme: ICommunitySectionTheme }> = ({ theme }) => {
         background="/img/community/contribute_bg.jpg"
         color={theme.color}
         description={description}
+        codeContrib={
+          <Link
+            className={sharedStyles.button}
+            style={theme}
+            href="https://dvc.org/doc/user-guide/contributing/core"
+          >
+            Contribute to Code
+          </Link>
+        }
+        docContrib={
+          <Link
+            className={sharedStyles.button}
+            style={theme}
+            href="https://dvc.org/doc/user-guide/contributing/docs"
+          >
+            Contribute to Docs
+          </Link>
+        }
         icon="/img/community/contribute.svg"
         mobileDescription={mobileDescription}
         title={title}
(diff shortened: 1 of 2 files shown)


shcheklein:
@sarthakforwet thanks!! it's great, but looks too excessive to my mind. We should integrate this properly into the page.

**Some ideas:**

- Make `code` and `docs` links in the very first sentence
- Instead of Go to Github - link to the Contribution section

Any other ideas are welcome, but it should be properly integrated onto Desktop/Mobile, etc.

sarthakforwet:
> Make code and docs links in the very first sentence

Nice idea @shcheklein! But I am kind of unaware as how we can club both the text and the link in the same sentence. (HTML tags do not work with json values). One midway could be to remove the sentence and add links to the guides for Doc contribution and Code Contribution.

  > Instead of Go to Github - link to the Contribution section

I think we must have this "Go to GitHub" button as a user would be more likely to click on it rather than going to GitHub on their own.

---------------------------------------------------------------------------------------------------------------------------------------------
The second idea that came to my mind is, what about adding cards for describing the Doc and Code contributions also.
Cards for doc and code contributions could be made similar to these-

image

shcheklein:
@sarthakforwet 

> But I am kind of unaware as how we can club both the text and the link in the same sentence. (HTML tags do not work with json values). One midway could be to remove the sentence and add links to the guides for Doc contribution and Code Contribution.

I'm not sure I understand this to be honest, could you elaborate please?

> I think we must have this "Go to GitHub" button as a user would be more likely to click on it rather than going to GitHub on their own.

sounds good.

> The second idea that came to my mind is, what about adding cards for describing the Doc and Code contributions also.
Cards for doc and code contributions could be made similar to these-

we'll have too many cards then?

sarthakforwet:
> I'm not sure I understand this to be honest, could you elaborate please?

Consider this-
image

Now I am unable to figure out how would I change those string "code" and "docs" to anchor text as this sentence is stored as a json value in the code.

> we'll have too many cards then?

Agreed! But we can add them in a manner that they would cause minimum distortion. The more we encourage contribution the more they'll come.  I think adding cards would impact more! Your call next @shcheklein, what should we do?

shcheklein:
> Your call next @shcheklein, what should we do?

So far the only idea that I like is to add links to the text- which we should find someone to implement. Otherwise we should ask for a designer input or come with more ideas on our own.

jorgeorpinel:
Thanks for the work so far. Closing as stale for now. Feel free to reopen if there are significant updates to this. Thanks!

PR #1840, opened 2020-10-05
how-to: add stage outputs existing in the WS but missing from dvc.yaml

### LINKED ISSUE #460: cmd ref: improve `run` and `commit` / how to add outs/deps without re-running stage?

imhardikj:
Add additional output to a stage without running it again.
Partially resolves #460
Refer - #899 (for How to doc)


content/docs/sidebar.json
```diff
@@ -102,7 +102,14 @@
         "label": "How To",
         "slug": "how-to",
         "source": false,
-        "children": ["undo-adding-data", "update-tracked-files"]
+        "children": [
+          {
+            "label": "Add Output to Stage",
+            "slug": "add-output-to-stage"
+          },
```
jorgeorpinel, line 109: Quick Q @rogermparent. Do you think it would be easy to NOT capitalize certain words from slugs like "to"? To avoid having to use labels for instances like this one. Other such words are 'a' and 'the'. We could keep a list somewhere. Just checking with you before even opening a feature request, may not be worth it.

jorgeorpinel (commented):
Hey @imhardikj a preliminary comment here is that to fix #460 you also need to edit the run and commit references to mention this scenario (and link to the How To for details).

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
+# Add Output to Stage
+
+There are situations where you might have forgotten to specify one of the
+<abbr>outputs</abbr> for a stage while creating it, and then later want to add
+an output to that stage.
```
jorgeorpinel, line 5: Correct but this is not the only case where this scenario happens. The more generalized explanation here I think has to do with having run a stage without specifying all the outpus in dvc.yaml (either via run or manually) and later wanting to make some of the files/dirs produces by the state as outputs formally.

jorgeorpinel: (Without rerunning the stage)

jorgeorpinel: So pleas give the more general explanation and/or list all the ways that you can think about that a user would end up with this problem.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+
+Running a stage to specify additional output is an expensive step, as it
+requires time and resources to execute the command. Follow the steps listed
+below to add an output to the stage without running it again.
+
+Let's first create a stage for the example <abbr>project</abbr>:
```
jorgeorpinel, line 11: What example project?

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+          -d src/prepare.py \
+          -o data/train \
+          python src/prepare.py
+```
+
+You can check stage entry in `dvc.yaml`:
+
+```dvc
+$ cat dvc.yaml
+stages:
```
jorgeorpinel, line 24: Yeah in fact the dvc.yaml file is what really matter here. Using `dvc run` is not required and in fact we encourage users to make dvc.yaml files manually and only consider `run` a helper. The `run` command block could stay but as a note after the dvc.yaml block.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+  prepare:
+    cmd: python src/prepare.py
+    deps:
+    - src/prepare.py
+    outs:
+    - data/train
+```
+
+Now to add an output to the `prepare` stage, you need to use `dvc run` with the
```
jorgeorpinel, line 33:
```suggestion
To add another output to the `prepare` stage, you need to use `dvc run` with the
```
> May need formatting.

jorgeorpinel: Nice writing BTW, thanks!

But again, this is very `run`-centric. Let's explain what should happen to the dvc.yaml file first and then add the `run -f --no-exec` command as an alternative (note) to manual editing.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+command defined in it.
+
+Let's check `dvc.yaml` again:
+
+```dvc
+$ cat dvc.yaml
+stages:
+  prepare:
+    ...
+    outs:
```
jorgeorpinel, line 56: I feel like we could compress the whole How To into a single code block that shows the "diff" between dvc.yaml before and after the new output is added. See https://dvc.org/doc/start/data-pipelines (expand the first **Expand to see what happens under the hood.** to see the block I mean):

image

jorgeorpinel: Same for the `run` command notes, it could be a single block quote after the dvc.yaml block with both run commands one after the other basically. Let's see what you can come up with 🙂

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+    - data/test
+    - data/train
+```
+
+As you can see, the new output has been added to the `train` stage in
```
jorgeorpinel, line 61: Avoid condescending language:
```suggestion
That' it! The new output has been added to the `train` stage in
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+`dvc.yaml`. Now you need to run `dvc commit` to update `dvc.lock` and save data
```
jorgeorpinel, line 62: Redundant/obvious:
```suggestion
As you can see, the new output has been added to the `train` stage.
Now you need to run `dvc commit` to update `dvc.lock` and save data
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,69 @@
...
+to the <abbr>cache</abbr>:
```
jorgeorpinel, line 63: I guess the whole paragraph about "the new output has been added" is repetitive to previous paragraphs or could be included in the previous p. Let's just leave the one about `commit` here? And also move the very last sentence up to this paragraph to keep things concise?

jorgeorpinel: But it feels like the `commit` part could use a little more explaining. Why is it necessary, what does it do in the cache, etc.

jorgeorpinel (changes requested):
Great start. Left a round of feedback ☝️ 

Please read all the comments before addressing them as some of them are connected. Thanks

content/docs/sidebar.json
rogermparent: I believe you are describing [Title Case](https://en.wikipedia.org/wiki/Title_case), and if you are it should be pretty easy to use [this small library](https://www.npmjs.com/package/title-case) in our sidebar label generation.

Let me refamiliarize myself with the code to make sure this works, but I'm pretty confident it's doable.

rogermparent: While it seems using an external library can cause issues with words specific to us, like "DVC", we can simply write a small function that does automatic casing the specific way we want. An external file for this is be a bit much in my opinion, but a small custom function would be a clean solution that can adapt to any of our needs.

jorgeorpinel: Yeah, title capitalization, basically. I think if any word is all caps in the slug it should let it be, but if the slug uses lower case we can just use the a sidebar.json label for titles with "DVC".

rogermparent: I made #1844 as a quick answer to this. It does title case and capitalizes "DVC" I figure we want it capitalized more often than not, and the behavior can be overridden with a `label`.

jorgeorpinel: @imhardikj please merge master and apply this, double check the label looks good. Thanks:
```suggestion
          "add-output-to-stage"
```

content/docs/user-guide/how-to/add-output-to-stage.md
imhardikj: The emphasis (in the update) is on editing `dvc.yaml` now I think. Also `dvc run` is moved to a block quote.

content/docs/user-guide/how-to/add-output-to-stage.md
imhardikj: I have added this "diff" part but it isn't showing green color in my local environment.
Also as you said I moved the `dvc run` code to a block quote. But I have only kept the second one (with `-f` and `--no-exec`) as this how-to is about what happens after creating a stage. That is, the user might have already created the stage so I think the `dvc run` code showing creating the stage won't be required here. The other reason being that the block quote became too long.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
+# Add Output to Stage
+
+There are situations where we run a stage without specifying all the
+<abbr>outputs</abbr>, and then later want to add more outputs to that stage. If
+we use `dvc run` to add outputs, it will re-run the stage.
+
+Running a stage is an expensive step, as it requires time and resources to
+execute the command. Follow the steps listed below to add an output to the stage
```
jorgeorpinel, line 8:
```suggestion
execute the command. Follow the steps below to add an output to the stage
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+
+There are situations where we run a stage without specifying all the
+<abbr>outputs</abbr>, and then later want to add more outputs to that stage. If
+we use `dvc run` to add outputs, it will re-run the stage.
+
+Running a stage is an expensive step, as it requires time and resources to
+execute the command. Follow the steps below to add an output to the stage
+without running it again.
+
+We have an example stage `prepare` defined in `dvc.yaml`. To add another output
```
jorgeorpinel, line 11:
```suggestion
We start with an example `prepare` stage in `dvc.yaml`. To add another output
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+`data/test` to this stage, we can edit `dvc.yaml`. This is how it looks after
+the change:
```
jorgeorpinel, line 13:
```suggestion
We start with an example `prepare` stage in `dvc.yaml`, which has a single
output. To add a missing output to this stage, we can edit the file like this:
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+```dvc
+$ cat dvc.yaml
+stages:
+  prepare:
+    cmd: python src/prepare.py
+    deps:
+    - src/prepare.py
+    outs:
+    - data/train
++   - data/test
```
jorgeorpinel, line 24: We have to decide whether we're using ` ```dvc ` and the `$ cat ...` console example OR the ` ```git ` format (no `$` line) which supports the `+` line highlighting. The latter seems more important it has already been mentioned that this is a dvc.yaml file. BTW I think all lines need an extra space at the beginning to align with the one starting with `+`, please check how it gets rendered in your local dev env.

jorgeorpinel: Also data/test -> data/validate please

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+> ```dvc
+> $ dvc run -f --no-exec \
+>           -n prepare \
+>           -d src/prepare.py \
+>           -o data/train \
+>           -o data/test \
+>           python src/prepare.py
+> ```
+>
+> The `-f` option is used to overwrite an existing stage in the `dvc.yaml`, and
```
jorgeorpinel, line 41: the dvc.yaml ? Either just dvc.yaml, or "the dvc.yaml file", remember? "the" needs a noun. "dvc.yaml" is a name so doesn't need an article (the).

Grammar https://owl.purdue.edu/owl/general_writing/grammar/using_articles.html

jorgeorpinel: "is used to overwrite an existing stage" is a little overcomplicated language and too generic, when we're talking about a specific use. It can just be "overwrites the stage". Simple, direct language 🙂

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+> the `--no-exec` option is used to create or update a stage without executing
+> the command.
+
+Finally, we need to run `dvc commit`. It saves the newly specified outputs to
+the <abbr>cache</abbr> and updates their hash values in `dvc.lock`:
```
jorgeorpinel, line 46:
```suggestion
the <abbr>cache</abbr> (and updates their hash values in `dvc.lock`):
```

content/docs/user-guide/how-to/add-output-to-stage.md
jorgeorpinel: > added this "diff" part but it isn't showing green color

It should. Did you check the example from where I took a screenshot to see what the code is? That MD code formatter is `git`.

> moved the dvc run code to a block quote

Goodm thanks

> the user might have already created the stage

Yep, not might: that's exactly the situation we're writing about 🙂

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+  prepare:
+    cmd: python src/prepare.py
+    deps:
+    - src/prepare.py
+    outs:
+    - data/train
++   - data/test
+```
+
+That's it, the new output has been added to the `prepare` stage.
```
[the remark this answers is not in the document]
jorgeorpinel: I see I suggested the was that sentence begins in https://github.com/iterative/dvc.org/pull/1840#pullrequestreview-503224505 but it no longer makes sense after moving stuff around.

content/docs/command-reference/run.md
```diff
@@ -144,6 +144,11 @@ $ dvc run -n my_stage "./my_script.sh > /dev/null 2>&1"
 $ dvc run -n my_stage './my_script.sh $MYENVVAR'
 ```
 
+Using `dvc run` with the `--no-exec` option doesn't execute the `command` and
+can be used to
+[add output to a stage](/docs/user-guide/how-to/add-output-to-stage) without
+re-running it.
+
```
jorgeorpinel, line 151: - This doesn't belong in the **The `command` argument** section, please find another place to insert this.
- It shouldn't focus on the `command`, probably not even mention it. It's about adding a missing output after the stage has already been run.

jorgeorpinel, line 151: - This doesn't belong in the **The `command` argument** section, please find another place to insert this.
- It shouldn't focus on the `command`, probably not even mention it. It's about adding a missing output after the stage has already been run.

content/docs/command-reference/commit.md
```diff
@@ -71,6 +71,10 @@ force-update the `dvc.lock` or `.dvc` files and save data to cache. They are
 still useful, but keep in mind that DVC can't guarantee reproducibility in those
 cases.
 
+We can [add outputs to a stage](/docs/user-guide/how-to/add-output-to-stage)
+without re-running it, and `dvc commit` is used to save the newly specified
+output to the cache and update `dvc.lock`.
+
```
jorgeorpinel, line 77: Looks like we're just appending the notes to the end of the Description. Let's read and understanding the context first to find the best place and manner to insert this info. Hint: there's actually a bullet list of scenarios to use this command in this ref.

jorgeorpinel: The explanation is missing a key piece of info: it's about EXISTING stages (that have already been executed). Let's also use the term "execute" instead of "run" here BTW, please. Also, no need to explain what commit does since that is explained in detail in this reference. Again, let's try to understand the context, read the doc, and make sure the text being inserted makes sense when reading it as a whole.

jorgeorpinel (changes requested):
Forgot to review the changes to run and commit:

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,47 @@
+# Add Output to Stage
+
+There are situations where we execute a stage without specifying all the
+<abbr>outputs</abbr> in `dvc.yaml` (either via `dvc run` or manually), and then
+later want to add more outputs to that stage.
+
+Re-executing a stage using `dvc run -f` to add an output is an expensive step,
+as it requires time and resources. Follow the steps below to add an output to
+the stage without executing it again.
```
jorgeorpinel, line 9: Still needs to be a little more general like we discussed. And no need to mention dvc run -f here.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,47 @@
...
+   prepare:
+     cmd: python src/prepare.py
+     deps:
+     - src/prepare.py
+     outs:
+     - data/train
++    - data/validate
+```
+
+We have added `data/validate` as an output to the `prepare` stage.
```
jorgeorpinel, line 25: Not sure about the usefulness of this sentence, maybe just mention it in the previous paragraph before the code block.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,47 @@
...
+>           -o data/train \
+>           -o data/validate \
+>           python src/prepare.py
+> ```
+>
+> The `-f` option overwrites the stage in `dvc.yaml`, and the `--no-exec` option
+> updates the stage without executing it.
+
+Finally, we need to run `dvc commit`. It saves the newly specified outputs to
+the <abbr>cache</abbr> (and updates their hash values in `dvc.lock`):
```
jorgeorpinel, line 43:
```suggestion
Finally, we need to run `dvc commit` to saves the newly specified output(s) to
the <abbr>cache</abbr> (and to update the corresponding hash values in
`dvc.lock`):
```
> May break the format

content/docs/command-reference/commit.md
```diff
@@ -34,9 +34,11 @@ scenarios are further detailed below.
...
 - It's always possible to manually execute the source code used in a stage
   without DVC (outputs must be unprotected or removed first in certain cases,
-  see `dvc unprotect`). Once a desirable result is reached, use `dvc add` or
-  `dvc commit` as appropriate to update the `dvc.lock` or `.dvc` files and store
-  changed data to the cache.
+  see `dvc unprotect`). We can also
+  [add outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without executing it again. Once a desirable result is reached, use `dvc add`
+  or `dvc commit` as appropriate to update the `dvc.lock` or `.dvc` files and
+  store changed data to the cache.
```
jorgeorpinel, line 41: Let's use 2 bullets. The existing one is having run a command without DVC and later wrapping the results with a dvc.yaml file. BTW this one should be the last bullet and could use some improvement in it's explanation e.g. it's not necessarily about source code, it could be a binary, etc. (maybe use the term "terminal command")...

The new one is about having run a stage with DVC (either writing dvc.yaml manually and using `repro`, or with `run`) but later noticing dvc.yaml is incomplete, missing one+ output(s) — which already exist in the workspace, so no need to re-execute the command here either.

So as you can see there's lots of subtleties here and we need to express everything in the docs as clearly and briefly as possible (difficult balance).

content/docs/command-reference/run.md
```diff
@@ -106,6 +106,10 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
+- Missing outputs can be
```
jorgeorpinel, line 109: @imhardikj please note this clarification, apply elsehwere:
```suggestion
- Outputs already in the workspace but missing from `dvc.yaml` can be
```

content/docs/command-reference/run.md
```diff
@@ -106,6 +106,10 @@ Relevant notes:
...
+  [added to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without running it again.
```
jorgeorpinel, line 111:
```suggestion
  without executing it again.
```
^ Note that we prefer that term since you can "run" stages either with `dvc run` or with `dvc repro` (sometimes even with `dvc pull`).

content/docs/command-reference/run.md
jorgeorpinel: p.s. I also updated the title of this PR to make extra clear what we're doing.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
...
+> $ dvc run -f --no-exec \
+>           -n prepare \
+>           -d src/prepare.py \
+>           -o data/train \
+>           -o data/validate \
+>           python src/prepare.py
+> ```
+>
+> The `-f` option overwrites the stage in `dvc.yaml`, and the `--no-exec` option
+> updates the stage without executing it.
```
jorgeorpinel, line 38:
```suggestion
> `-f` overwrites the stage in `dvc.yaml`, while `--no-exec` updates the stage
> without executing it.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
...
+>           -n prepare \
+>           -d src/prepare.py \
+>           -o data/train \
+>           -o data/test \
+>           python src/prepare.py
+> ```
+>
+> The `-f` option is used to overwrite an existing stage in the `dvc.yaml`, and
+> the `--no-exec` option is used to create or update a stage without executing
+> the command.
```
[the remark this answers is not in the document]
jorgeorpinel: This wasn't applied but I just did it so we can move on...

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that output(s) already existing in workspace are missing from `dvc.yaml`.
+Re-executing a stage to specify all the outputs is an expensive step, as it
+requires time and resources. Follow the steps below to add already existing
+output(s) to a stage without executing it again.
```
jorgeorpinel, line 8:
```suggestion
requires time and resources. Follow the steps below to add existing output(s)
to a stage without executing it again.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that output(s) already existing in workspace are missing from `dvc.yaml`.
+Re-executing a stage to specify all the outputs is an expensive step, as it
+requires time and resources. Follow the steps below to add existing output(s) to
+a stage without executing it again.
```
jorgeorpinel, line 8: This is pretty good writing, nice. But reading it now I just have one last concern. Will people understand what we mean by "output(s) already existing in workspace"? ("THE workspace" BTW 😕)
We could clarify what an output is in this case: file(s) written by the stage command (which we want to track with DVC).

jorgeorpinel: Also, please use `<abbr>` arround "output" in the 2nd or last instance of the word (once we're clearly referring to outputs tracked by DVC).

jorgeorpinel: > add existing output(s) to a stage

Hmmm. How about something like "add existing files or directories as outputs to a stage (`outs` field)" to make sure we separate the concepts of files in the workspace from tracked outputs, like I mentioned earlier? Please rewrite the p along these lines. Almost there...

content/docs/command-reference/commit.md
```diff
@@ -20,9 +20,10 @@ positional arguments:
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+force updating the `dvc.lock` or `.dvc` files without reproducing stages or
+pipelines; when adding already existing <abbr>outputs</abbr> to a stage that
+missing from `dvc.yaml`; or when manually editing or generating DVC outputs.
+These scenarios are further detailed below.
```
jorgeorpinel, line 26: OK to reorder these but adding another sentence here is TMI. It's already a very long/hard-to-read paragraph. Let's merge the 2 last cases as s single sentence similar to the original version. BTW please don't remove `<abbr>` tags unless there's a reason. Those make tooltips on the website:

image
> https://dvc.org/doc/command-reference/commit

content/docs/command-reference/commit.md
```diff
@@ -46,6 +41,19 @@ scenarios are further detailed below.
   reproduce the whole pipeline. If you're sure no pipeline results would change,
   use `dvc commit` to force update the `dvc.lock` or `.dvc` files and cache.
 
+- Sometimes we have executed a stage (either by writing `dvc.yaml` manually and
+  using `dvc repro`, or with `dvc run`), and later notice that output(s) which
+  already exist in workspace are missing in `dvc.yaml`. We can
```
jorgeorpinel, line 46: Good but I wouldn't start with "Sometimes" again same as the previous bullet (minor), and please apply the same changes you do to the paragraph in content/docs/user-guide/how-to/add-output-to-stage.md per the recent comments on files vs. outputs I left there. Thanks

content/docs/command-reference/commit.md
```diff
@@ -66,7 +74,7 @@ computed and added to the `dvc.lock` or `.dvc` file, but the actual data file is
 not saved in the cache. This is where the `dvc commit` command comes into play.
 It performs that last step (saving the data in cache).
 
-Note that it's best to avoid the last two scenarios. They essentially
+Note that it's best to avoid the last three scenarios. They essentially
```
jorgeorpinel, line 77: Good catch! Glad you're reading the whole context now 🙂

content/docs/command-reference/run.md
```diff
@@ -106,6 +106,10 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
+- Outputs already in the workspace but missing from `dvc.yaml` can be
+  [added to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without executing it again.
```
jorgeorpinel, line 111: Same. Maybe we can clarify files vs. outputs a little? And consider adding `<abbr>` if there's no existing output tooltip nearby already.

jorgeorpinel (changes requested):
90% there ^

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,48 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that output(s) already existing in workspace are missing from `dvc.yaml`.
```
jorgeorpinel, line 5:
```suggestion
that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` field).
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,48 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` field).
+Outputs are files or directories written by the stage commands which we want to
+track with DVC. Re-executing a stage to specify all the <abbr>outputs</abbr> is
```
jorgeorpinel, line 7:
```suggestion
Re-executing a stage to specify all the <abbr>outputs</abbr> is
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,47 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` field).
```
jorgeorpinel, line 5:
```suggestion
that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` fields).
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,47 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` fields).
+Re-executing a stage to specify all the <abbr>outputs</abbr> is
+an expensive step, as it requires time and resources. Follow the steps below to
+add existing files or directories as output(s) to a stage (`outs` field in
+`dvc.yaml`) without executing it again.
```
jorgeorpinel, line 9:
```suggestion
Follow the steps below to
add existing files or directories as outputs to a stage
without re-executing it again, which can be expensive/time-consuming, but is unnecessary.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace, are missing from `dvc.yaml` (`outs` fields).
+Follow the steps below to
+add existing files or directories as outputs to a stage
+without re-executing it again, which can be expensive/time-consuming, but is unnecessary.
```
jorgeorpinel, line 8: OK just need to format it now:
```suggestion
There are situations where we have executed a stage (either by writing
`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
that some of its output files or directories, which already exist in workspace,
are missing from `dvc.yaml` (`outs` fields). Follow the steps below to add
existing files or directories as outputs to a stage without re-executing it
again, which can be expensive/time-consuming, but is unnecessary.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace,
+are missing from `dvc.yaml` (`outs` fields). Follow the steps below to add
```
jorgeorpinel, line 6:
```suggestion
are missing from `dvc.yaml` (`outs` field). Follow the steps below to add
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which already exist in workspace,
+are missing from `dvc.yaml` (`outs` field). Follow the steps below to add
+existing files or directories as outputs to a stage without re-executing it
+again, which can be expensive/time-consuming, but is unnecessary.
```
jorgeorpinel, line 8: @imhardikj still missing that "the" 😢 
```suggestion
that some of its output files or directories, which are already in the
workspace, are missing from `dvc.yaml` (`outs` field). Follow the steps below to
add existing files or directories as outputs to a stage without re-executing it
again, which can be expensive/time-consuming, but is unnecessary.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
...
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which are already in the
+workspace, are missing from `dvc.yaml` (`outs` field). Follow the steps below to
+add existing files or directories as outputs to a stage without re-executing it
+again, which can be expensive/time-consuming, but is unnecessary.
+
+We start with an example `prepare` stage in `dvc.yaml`, which has a single
+output. To add a missing output `data/validate` to this stage, we can edit
+`dvc.yaml` like this:
```
jorgeorpinel, line 12:
```suggestion
We start with an example `prepare`, which has a single output. To add a missing
output `data/validate` to this stage, we can edit `dvc.yaml` like this:
```

content/docs/user-guide/how-to/add-output-to-stage.md
jorgeorpinel: I see you added a definition of what outputs are. That tries to address this but isn't subtle enough. I've finished this change for you as it's a little complicated I suppose, resolving here!

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,45 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories, which are already in the
+workspace, are missing from `dvc.yaml` (`outs` field). Follow the steps below to
+add existing files or directories as outputs to a stage without re-executing it
+again, which can be expensive/time-consuming, but is unnecessary.
```
jorgeorpinel, line 8: Here it is. Notice how I clarify the diff between files/dirs written by the stage vs. outputs (in `<abbr>`) tracked by DVC (`outs` field of dvc.yaml). Please apply similar changes to the other 2 docs if you haven't and request my review again to finalize this. Thanks

content/docs/command-reference/commit.md
jorgeorpinel: I still see the new separate sentence there, BTW. Let's merge those last 2 sentences before "These scenarios..." somehow please.

content/docs/user-guide/how-to/add-output-to-stage.md
jorgeorpinel: I fixed this BTW.

content/docs/user-guide/how-to/add-output-to-stage.md
jorgeorpinel: Actually, this is missing `<abbr>` tags around the 2nd instance of "outputs" 🤦. Please add them @imhardikj, thanks.

jorgeorpinel: ACTUALLY, you can just commit this (made another small copy edit for extra clarification between written files vs tracked outputs):
```suggestion
There are situations where we have executed a stage (either by writing
`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
that some of the files or directories it writes, which are already in the
workspace, are missing from `dvc.yaml` (`outs` field). Follow the steps below to
add existing files or directories as <abbr>outputs</abbr> to a stage without
re-executing it again, which can be expensive/time-consuming, but is
unnecessary.
```

jorgeorpinel: p.s. Not sure whether it should be "it writes" or "it creates" — up to you.

content/docs/user-guide/how-to/add-output-to-stage.md
imhardikj: Thank You. I tried to include that but the sentence became very complex so I thought I'd rather make a separate one. Thanks again btw.

content/docs/user-guide/how-to/add-output-to-stage.md
imhardikj: used "creates" because every time before execution existing outputs are deleted, and then new ones are created.

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its output files or directories it creates, which are already in
```
jorgeorpinel, line 5:
```suggestion
that some of the output files or directories it creates, which are already in
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the output files or directories it creates, which are already in
+the workspace, are missing from `dvc.yaml` (`outs` field). Follow the steps
+below to add existing files or directories as <abbr>outputs</abbr> to a stage
+without re-executing it again, which can be expensive/time-consuming, but is
+unnecessary.
```
jorgeorpinel, line 9:
```suggestion
the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs` field). Follow
the steps below to add existing files or directories as <abbr>outputs</abbr> to
a stage without re-executing it again, which can be expensive/time-consuming,
but is unnecessary.
```

content/docs/user-guide/how-to/add-output-to-stage.md
```diff
@@ -0,0 +1,46 @@
+# Add Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the output files or directories it creates, which are already in
+the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs` field). Follow
+the steps below to add existing files or directories as <abbr>outputs</abbr> to
+a stage without re-executing it again, which can be expensive/time-consuming,
+but is unnecessary.
```
jorgeorpinel, line 9:
```suggestion
the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs` field). Follow
the steps below to add existing files or directories as <abbr>outputs</abbr> to
a stage without re-executing it again, which can be expensive/time-consuming,
and is unnecessary.
```

content/docs/command-reference/commit.md
```diff
@@ -20,9 +20,9 @@ positional arguments:
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+force updating the `dvc.lock` or `.dvc` files without reproducing stages or
+pipelines; or when manually editing or generating DVC <abbr>outputs</abbr> or
+adding them to a stage. These scenarios are further detailed below.
```
jorgeorpinel, line 25: > or when manually editing or generating DVC <abbr>outputs</abbr> or
adding them to a stage

To many "or"s and the order is inverted compared to the bullet list below. But let's combine them conceptually. Generalize. Basically you just removed the comma so far 🙂  How can you describe all the cases of the last sentence in a single statement? Should be possible, I think.

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
   reproduce the whole pipeline. If you're sure no pipeline results would change,
   use `dvc commit` to force update the `dvc.lock` or `.dvc` files and cache.
 
+- In some situations we have executed a stage (either by writing `dvc.yaml`
```
jorgeorpinel, line 43:
```suggestion
- In cases where we have previously executed a stage (either by writing `dvc.yaml`
```

content/docs/command-reference/commit.md
jorgeorpinel: and please apply the same changes you do to the paragraph in content/docs/user-guide/how-to/add-output-to-stage.md

content/docs/command-reference/commit.md
jorgeorpinel: > stage commands

Commands? There's just one command per stage.

content/docs/command-reference/run.md
```diff
@@ -106,6 +106,11 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
+- <abbr>Output</abbr> files or directories created by the stage commands which
+  are already in the workspace but missing from `dvc.yaml` (`outs` field) can be
```
jorgeorpinel, line 110: Same here. Just use the same text as much as possible since we already have a solution in content/docs/user-guide/how-to/add-output-to-stage.md because this version still doesn't clarify between files/dir "outputs" of stages and formally tracked _outputs_ (only the 2nd case should have `abbr` tags. Thanks

imhardikj: Have updated it with almost same text as in `commit.md`

content/docs/command-reference/commit.md
imhardikj: Tried combine the last point in same sentence in the latest update. Please check.

content/docs/command-reference/commit.md
jorgeorpinel: Still pending... Will comment again.

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
   reproduce the whole pipeline. If you're sure no pipeline results would change,
   use `dvc commit` to force update the `dvc.lock` or `.dvc` files and cache.
 
+- In cases where we have previously executed a stage (either by writing `dvc.yaml`
+  manually and using `dvc repro`, or with `dvc run`), and later notice that
+  output files or directories created by the stage command, which are already in
+  the workspace, are missing from `dvc.yaml` (`outs` field). We can
```
jorgeorpinel, line 46: Why not just use the exact same text we have previously reviewed and agreed upon? As requested in https://github.com/iterative/dvc.org/pull/1840#pullrequestreview-512176615:

`that some of the output files or directories it creates, which are already in the <abbr>workspace</abbr>`

content/docs/command-reference/commit.md
```diff
@@ -20,9 +20,9 @@ positional arguments:
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+force updating the `dvc.lock` or `.dvc` files without reproducing stages or
```
jorgeorpinel, line 23:
```suggestion
force-updating the `dvc.lock` or `.dvc` files without reproducing stages or
```

content/docs/command-reference/commit.md
```diff
@@ -20,9 +20,9 @@ positional arguments:
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+force-updating the `dvc.lock` or `.dvc` files without reproducing stages or
+pipelines; or when DVC <abbr>outputs</abbr> are added to a stage, manually
+generated or edited. These scenarios are further detailed below.
```
jorgeorpinel, line 25:
```suggestion
generated, or edited. These scenarios are further detailed below.
```

content/docs/command-reference/commit.md
```diff
@@ -20,9 +20,9 @@ positional arguments:
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; when
```
jorgeorpinel, line 22:
```suggestion
[pipeline](/doc/command-reference/dag) is in development/experimentation; to
```

content/docs/command-reference/commit.md
```diff
@@ -19,10 +19,10 @@ positional arguments:
 
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
-[pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+[pipeline](/doc/command-reference/dag) is in development/experimentation; to
+force-updating the `dvc.lock` or `.dvc` files without reproducing stages or
```
jorgeorpinel, line 23:
```suggestion
force-update the `dvc.lock` or `.dvc` files without reproducing stages or
```

content/docs/command-reference/commit.md
```diff
@@ -19,10 +19,10 @@ positional arguments:
...
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
-[pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
-scenarios are further detailed below.
+[pipeline](/doc/command-reference/dag) is in development/experimentation; to
+force-update the `dvc.lock` or `.dvc` files without reproducing stages or
+pipelines; or when DVC <abbr>outputs</abbr> are added to a stage, manually
+generated, or edited. These scenarios are further detailed below.
```
jorgeorpinel, line 25: > when DVC <abbr>outputs</abbr> are added to a stage, manually generated, or edited

I don't even know what we were trying to say here, actually :/ Please go over this phrasing again, it should be simple language that makes sense, not a riddle that readers will just shrug off/ignore/skip. Just a general statement that doesn't have all the details is better here, since it will be explained in detail below in the bullets. No need to force a complete explanation in a single sentence, try to make it general yet correct.

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
...
 
+- In cases where we have previously executed a stage (either by writing `dvc.yaml`
+  manually and using `dvc repro`, or with `dvc run`), and later notice that
+  output files or directories created by the stage command, which are already in
+  the workspace, are missing from `dvc.yaml` (`outs` field). We can
+  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without having to execute it again. Use `dvc commit` to update the `dvc.lock`
+  file and save outputs to the cache.
+
+- It's always possible to manually execute the terminal command or source code
```
jorgeorpinel, line 51:
```suggestion
- It's always possible to manually execute the command or source code
```

jorgeorpinel: Please commit and fix formatting.

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
...
+  used in a stage without DVC (outputs must be unprotected or removed first in
+  certain cases, see `dvc unprotect`). Once a desirable result is reached, use
```
jorgeorpinel, line 53:
```suggestion
  certain cases, see `dvc unprotect`). Once the desired result is reached, use
```

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
...
+  the workspace, are missing from `dvc.yaml` (`outs` field). We can
+  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without having to execute it again. Use `dvc commit` to update the `dvc.lock`
+  file and save outputs to the cache.
+
+- It's always possible to manually execute the terminal command or source code
+  used in a stage without DVC (outputs must be unprotected or removed first in
+  certain cases, see `dvc unprotect`). Once the desired result is reached, use
+  `dvc add` or `dvc commit` as appropriate to update the `dvc.lock` or `.dvc`
+  files and store changed data to the cache.
```
jorgeorpinel, line 55: So actually this seems wrong. `dvc add` and `.dvc` files don't seem to have anything to do with this case. We're talking about stage commands so only commit and dvc.lock apply here. Please update

p.s. I know this text was there, before but since already we moved it around, let's also fix this up please.

jorgeorpinel: Maybe just ``... use `dvc commit` to update the `dvc.lock` file(s) and ...``

content/docs/command-reference/run.md
```diff
@@ -106,6 +106,12 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
+- In some situations we have executed a stage and later notice that output files
+  or directories created by the stage command, which are already in the
+  workspace, are missing from `dvc.yaml` (`outs` field). We can
```
jorgeorpinel, line 111: Also use the same text we have already agreed upon in content/docs/user-guide/how-to/add-output-to-stage.md please! Need to differentiate between files/dirs written by a command and formal DVC outputs. This is key... As discussed a few times.

`... later notice that some of the output files or directories it creates, which are already in the <abbr>workspace</abbr>, ...`

content/docs/command-reference/run.md
jorgeorpinel: why commit.md?

> we already have a solution in content/docs/user-guide/how-to/add-output-to-stage.md

jorgeorpinel (changes requested):
I'm afraid we're stuck at 90% of the way. We're close but we can't polish the consistency and other details. Details matter. Please not only address my feedback comments one by one in a vacuum, they connect with each other. Do address them individually but then please check the whole context, and that the changes to each doc are consistent with one another. Thanks

content/docs/command-reference/commit.md
imhardikj: Done.

content/docs/command-reference/run.md
imhardikj: Done.

content/docs/command-reference/commit.md
imhardikj: Have updated this. Tried to make it simpler and more general.

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
   reproduce the whole pipeline. If you're sure no pipeline results would change,
   use `dvc commit` to force update the `dvc.lock` or `.dvc` files and cache.
 
+- In cases where we have previously executed a stage (either by writing
+  `dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later
+  notice that some of the output files or directories it creates, which are
+  already in the workspace, are missing from `dvc.yaml` (`outs` field). We can
```
jorgeorpinel, line 46: `abbr` aroud "workspace"?

content/docs/command-reference/commit.md
```diff
@@ -19,9 +19,9 @@ positional arguments:
 
 The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
-[pipeline](/doc/command-reference/dag) is in development/experimentation; when
-manually editing or generating DVC <abbr>outputs</abbr>; or to force update the
-`dvc.lock` or `.dvc` files without reproducing stages or pipelines. These
+[pipeline](/doc/command-reference/dag) is in development/experimentation; to
+force-update the `dvc.lock` or `.dvc` files without reproducing stages or
+pipelines; or to add missing outputs or manually execute a stage. These
```
jorgeorpinel, line 24: > or to add missing outputs or manually execute a stage.

Yes, definitely simpler which is good. But this is still 2 separate things (so should use a comma, removing a coma is not = merging). This should be something like "or to mark existing files/dirs as stage outputs" — a single generalization that covers the last 2 bullets.

Also, `abbr` is needed around "outputs".

content/docs/command-reference/commit.md
```diff
@@ -46,6 +40,20 @@ scenarios are further detailed below.
...
+  notice that some of the output files or directories it creates, which are
+  already in the workspace, are missing from `dvc.yaml` (`outs` field). We can
+  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  without having to execute it again. Use `dvc commit` to update the `dvc.lock`
+  file and save outputs to the cache.
+
+- It's always possible to manually execute the command or source code used in a
+  stage without DVC (outputs must be unprotected or removed first in certain
+  cases, see `dvc unprotect`). Once the desired result is reached, use
+  `dvc commit` as appropriate to update the `dvc.lock` file(s) and store changed
```
jorgeorpinel, line 54:
```suggestion
  `dvc commit` to update the `dvc.lock` file(s) and store changed
```

jorgeorpinel: Needs formatting

jorgeorpinel (commented):
A few final items. We can merge after. Please address @imhardikj. Thanks

content/docs/command-reference/commit.md
imhardikj: Done.

content/docs/command-reference/commit.md
```diff
@@ -66,7 +75,7 @@ computed and added to the `dvc.lock` or `.dvc` file, but the actual data file is
 not saved in the cache. This is where the `dvc commit` command comes into play.
 It performs that last step (saving the data in cache).
 
-Note that it's best to avoid the last two scenarios. They essentially
+Note that it's best to avoid the last three scenarios. They essentially
```
jorgeorpinel, line 78:
```suggestion
Note that it's best to avoid the last two scenarios. They essentially
```

jorgeorpinel (approved):

jorgeorpinel:
Done here 🙂

PR #1844, opened 2020-10-07
Use modified Title Case for slug -> label conversion.

rogermparent:
Before, we used lodash startCase which simply made every word have its first
letter capitalized. This new function does title case (leaving 'and', 'or', etc.
uncapitalized) and also capitalizes "dvc"

Initially mentioned [here](https://github.com/iterative/dvc.org/pull/1840#discussion_r500466702)

### DIFF
--- a/content/docs/sidebar.json
+++ b/content/docs/sidebar.json
@@ -87,10 +87,7 @@
         "slug": "what-is-dvc",
         "source": "what-is-dvc.md"
       },
-      {
-        "label": "DVC Files and Directories",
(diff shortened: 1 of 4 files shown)


shcheklein (approved):

jorgeorpinel:
That was a quick feature, thanks @rogermparent !

> and also capitalizes "dvc"

Doesn't title case respect all acronyms by default? The slugs would need to be case sensitive though, e.g. `DVC-files-and-directories` — probably not a great idea.

PR #1855, opened 2020-10-12
start: add meta title, description, and key term (data versioning) [SEO]  

jeremydesroches:
**Meta title**
Added custom title to front matter to add "Get Started:" prefix. Originally suggested for the tutorial, this format should also help with ranking and clickthrough for the Get Started guides. As with the tutorial, I updated the H1 to match.

**Meta description**
Added a custom meta description to the front matter with overview and key terms. This doc appears on the [first page of search results](https://www.google.com/search?q=versioning+data) for 'data versioning' / 'versioning data', but the auto-generated description is too specific and doesn't reference the key terms. 
image

**Key term**
Added "versioning data" to the second paragraph to address this alternate version of the search term.


content/docs/start/data-versioning.md
```diff
@@ -1,3 +1,8 @@
+---
+description: |
+  Get started with data versioning in DVC. Learn how to use a regular Git workflow for large files, datasets, and ML models, without storing large files in Git.
```
jorgeorpinel, line 3: Hmmm. Restyled doesn't pick this up but it should be wrapped at 80 chars like in blogs, I think.

jorgeorpinel: Removing a term that repeats:
```suggestion
  Get started with data versioning in DVC. Learn how to use a regular Git workflow for datasets, and ML models, without storing large files in Git.
```

jorgeorpinel: - I see this comes from the GS index, good idea. Should we also include "Git for data" somewhere in the docs?
- "ML models" is also good but should the full term "machine learning" or "machine learning models" (and possibly other terms around "model" like "model versioning" be added somewhere to the doc too? It's not present at the moment.

jorgeorpinel (commented):
Left a couple comments but looks good overall.

~~I also noticed the intro paragraph has some strange language like "Imagine you can do a ..." and "Or do ... in a less than a second?" — would you like to rewrite that correctly @jeremydesroches? (out of scope so up to you)~~

jorgeorpinel:
Also, question:

> This doc appears on the first page of search results for 'data versioning' / 'versioning data', but the auto-generated description is too specific...
Added "versioning data" to the second paragraph

Sounds logical (and a necessary fix, thanks). My Q is what kind of impact can we expect from a change like this one, where we're already ranking on page #1 for the terms involved?

content/docs/start/data-versioning.md
jeremydesroches: > Hmmm. Restyled doesn't pick this up but it should be wrapped at 80 chars like in blogs, I think.

Yeah, fixed it now. FYI when I used the `|` character and tried to format it as in blog posts it broke the renderer. Seems work fine with single quotes like in Roger's initial example.

> Removing a term that repeats

Nice catch, thanks. Fixed and also removed extra comma.

> Should we also include "Git for data" somewhere in the docs?

Good idea. Added in rewrite of intro section that I just pushed. PTAL

> "ML models" is also good but should the full term "machine learning" or "machine learning models" (and possibly other terms around "model" like "model versioning" be added somewhere to the doc too? It's not present at the moment.

In this case I was only looking to add "data versioning". As per your request for one term/doc/PR, but also the subject matter and existing terms for this page are more related to versioning than ML. So it's OK to keep with the abbreviations here and apply the full set in other places.

jeremydesroches:
> I also noticed the intro paragraph has some strange language like "Imagine you can do a ..." and "Or do ... in a less than a second?" — would you like to rewrite that correctly @jeremydesroches? (out of scope so up to you)

I thought you would never ask... 😆 PTAL

jeremydesroches:
> what kind of impact can we expect from a change like this one, where we're already ranking on page # 1 for the terms involved?

On the page level, we're hoping to see an improved clickthrough rate (more appealing in the list of results) and lower bounce rate (better preview of actual doc content). Improvements on those metrics could result in a better position on the 1st page and less decay over time (by maintaining good performance metrics vs competing results). 

The overarching goal is in the aggregate though: small optimizations like this might only net small improvements at the page level, but done across many pages can add up to significant improvement at the site level. Improved performance across multiple pages works to improve "domain authority", which is a major component of the rank for any given page. It's sort of like we have 100 helium balloons under a net... they all provide a little bit of lift. 🎈

shcheklein:
> I also noticed the intro paragraph has some strange language like "Imagine you can do a ..." and "Or do ... in a less than a second?" — would you like to rewrite that correctly @jeremydesroches? (out of scope so up to you)

@jorgeorpinel @jeremydesroches let's keep the PR well scoped? 

also, the way we changed it is not correct. It was specifically made informal (imagine Git that can clone data != DVC can do this on `git clone`, etc). Making it formal requires deep rewrite and I'm not sure it's needed (e.g. we might decide to remove it), also not sure it is in the scope of this PR.

content/docs/start/data-versioning.md
jorgeorpinel: Sounds good.

content/docs/start/data-versioning.md
```diff
@@ -1,13 +1,22 @@
...
+---
 
-How cool would it be to make Git handle arbitrary large files and directories
-with the same performance as with small code files? Imagine you can do a
-`git clone` and see data files and ML model files in the workspace. Or do
-`git checkout` and switch to a different version of a 100Gb size file in a less
-than a second?
+# Data Versioning
 
 The core part of DVC is a few commands that you can run along with Git to track
```
jorgeorpinel, line 9:
```suggestion
How cool would it be to make Git handle arbitrary large files and directories
with the same performance as with small code files? Imagine you can do a
`git clone` and see data files and ML model files in the workspace. Or do
`git checkout` and switch to a different version of a 100Gb size file in a less
than a second?

The core part of DVC is a few commands that you can run along with Git to track
```

jorgeorpinel: Rolling this back for now.

jorgeorpinel:
> > the intro paragraph has some strange language
> 
> let's keep the PR well scoped?
also, the way we changed it is not correct

Indeed it's out of scope like I mentioned so never mind @jeremydesroches, sorry about the disctraction — I'll address it somewhere else.

content/docs/start/data-versioning.md
```diff
@@ -7,7 +13,16 @@ with the same performance as with small code files? Imagine you can do a
...
+a large file, ML model, or a directory. Think "Git for data". For example, you
+can use DVC and Git to:
+
+- Track large files and directories in Git with the same performance as small
+  code files.
+- See data files and ML model files next to code in your workspace with
+  `git clone`.
+- Switch between versions of a 100GB dataset instantly with `git checkout`.
+
+Read on or watch our video docs to learn about versioning data with DVC!
```
jorgeorpinel, line 25: We can definitely insert "Git for data" but everything else here is a huge change. Yeah sorry, Get Started pages need to be as simple as possible, this is too much in this case (and changes the precious tone/meaning as Ivan mentioned). I only wanted to do a very small copy edit to improve the language on the first paragraph (but again, never mind about that, I'll take care of it somewhere else).

jorgeorpinel: p.s. It's my bad for expanding the scope...

jorgeorpinel:
> Improved performance across multiple pages works to improve "domain authority", which is a major component of the rank

Ok cool thanks, definitely important. As discussed offline, this is a much needed but secondary (in terms of urgency) line of work.

content/docs/start/data-versioning.md
jorgeorpinel: p.p.s. here's basically what I had in mind: https://github.com/iterative/dvc.org/pull/1871/files#diff-ea052227e82952e1865a66baab65e3ef88513639e7a00d7dedab4e44c53e0891 cc @shcheklein

jeremydesroches: OK, PTAL at latest version that combines your latest and adds 'versioning data' to end of p2.

content/docs/start/data-versioning.md
```diff
@@ -7,8 +14,8 @@ to a different version of a 100Gb file in less than a second with a
 `git checkout`.
 
 The foundation of DVC consists of a few commands that you can run along with
-`git` to track large files and directories, or ML models. Read on, or watch our
-video guides!
+`git` to track large files, directories, or ML models. Think "Git for data".
+Read on or watch our video docs to learn about versioning data with DVC!
```
jorgeorpinel, line 18:
```suggestion
Read on or watch our video to learn about versioning data with DVC!
```

content/docs/start/data-versioning.md
```diff
@@ -7,8 +14,8 @@ to a different version of a 100Gb file in less than a second with a
 `git checkout`.
 
 The foundation of DVC consists of a few commands that you can run along with
-`git` to track large files and directories, or ML models. Read on, or watch our
-video guides!
+`git` to track large files, directories, or ML models. Think "Git for data".
+Read on or watch our video to learn about versioning data with DVC!
```
jorgeorpinel, line 18: Great. My only thing here is that the paragraph before the video looked nice in 2 lines but now it has a 3rd 😋 

image

Definitely very minor (probably nothing). We can merge this.

jorgeorpinel (approved):
Want to double check @shcheklein ?

jorgeorpinel:
Alright, merging this!

jeremydesroches:

**Get Started: Data Versioning** - [Link](https://search.google.com/u/0/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fstart%2Fdata-versioning&breakdown=query&metrics=CLICKS%2CIMPRESSIONS&start_date=20201023&end_date=20201204&compare_start_date=20200910&compare_end_date=20201022) - 41% higher clicks and 83% higher impressions. Clickthrough and position were down slightly (to be expected from large increase with impressions). On the key term "data versioning", clicks were up 46%, impressions up 85%, and position improved from 2.7 to 2.1.

jorgeorpinel:
Nice, where would we look for brute traffic figures to confirm the effects here? Somewhere in GA I suppose?

Also, I [changed the dates](https://search.google.com/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fstart%2Fdata-versioning&breakdown=query&metrics=IMPRESSIONS&start_date=20201021&end_date=20201202&compare_start_date=20200909&compare_end_date=20201021) to align the weeks, so that the graph pattern is less confusing:

image

> On the key term "data versioning", clicks were up 46%, impressions up 85%, and position improved from 2.7 to 2.1.

Great results 👍 

p.s. funny how "versioning data" is such a 2nd choice of words vs "data versioning".

jorgeorpinel:
Now that I see the graph closely though, I think that the URL didn't exist or was still pretty new a few weeks before this change so I [closed the range](https://search.google.com/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fstart%2Fdata-versioning&breakdown=query&metrics=CLICKS&start_date=20201021&end_date=20201111&compare_start_date=20200930&compare_end_date=20201021) to 3 weeks before and 3 after. The results are more moderate but still good:

image

jeremydesroches:
> I think that the URL didn't exist or was still pretty new a few weeks before this change ... The results are more moderate but still good

Yep, it was around, but started to grow impressions and clicks just a few weeks before this change. I think what we're seeing on this one is also the URL being "auditioned" (where the search engine assigns more impressions to test out how it performs), which is more apparent in the long-term view:

image

So in this case it's very important to look at the nature of the optimizations: we added meta title and description (goal is increase clickthrough) and a key term (goal is increase performance on that term). We covered the performance of the term above (good). Looking at the clickthrough, it increased across these terms:

image

...while decreasing across only a few, which produced more clicks anyway (CTR decreased while clicks increased because of large increase in impressions):

image

jorgeorpinel:
Perfect, that itemized report is much more informative. And yes, funny how the page grew steadily during those weeks without having any changes that I can think of...

PR #1856, opened 2020-10-12
cases: add custom meta title and meta description to (data versioning tutorial) [SEO]

jeremydesroches:
The data versioning tutorial has good placement in [search results](https://www.google.com/search?q=dataset+versioning) for key terms like "versioning data" and "dataset versioning". The results for this page typically show the default dvc.org meta description, and the title is limited to "Tutorial".
image

A unique description that matches the page content can increase clickthrough and reduce bounce (leading to better search ranking). Pages that include the search terms in their title typically rank better on those terms.

This PR adds a custom meta description and expands the meta title to "Tutorial: Data & Model Versioning".

### DIFF
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -1,15 +1,15 @@
 ---
 title: 'Tutorial: Data & Model Versioning'
-description: 'Get experience with data versioning in a basic machine learning
-version control scenario: managing multiple datasets and ML model versions using
-DVC commands.'
+description: 'Get hands-on experience with data versioning in a basic machine
+learning version control scenario: managing multiple datasets and ML models
+using DVC.'
 ---
 
 # Tutorial: Data & Model Versioning
 
 The goal of this example is to give you some hands-on experience with a basic
 machine learning version control scenario: managing multiple datasets and ML
-model versions using DVC commands. We'll work with a
+models using DVC. We'll work with a
 [tutorial](https://blog.keras.io/building-powerful-image-classification-models-using-very-little-data.html)
 that [François Chollet](https://twitter.com/fchollet) put together to show how
 to build a powerful image classifier using a pretty small dataset.
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -1,7 +1,8 @@
 ---
 title: 'Tutorial: Data & Model Versioning'
-description: |
-  Get experience with data versioning in a basic machine learning version control scenario: managing multiple datasets and ML model versions using DVC commands.
+description: 'Get experience with data versioning in a basic machine learning
+version control scenario: managing multiple datasets and ML model versions using
+DVC commands.'
 ---
 
 # Tutorial: Data & Model Versioning
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -1,5 +1,5 @@
 ---
-title: 'Tutorial: Data & Model Versioning'
+title: 'Data and Model Versioning Tutorial'
 description: 'Get hands-on experience with data versioning in a basic machine
 learning version control scenario: managing multiple datasets and ML models
 using DVC.'
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -1,3 +1,9 @@
+---
+title: 'Tutorial: Data & Model Versioning'
+description: |
+  Get experience with data versioning in a basic machine learning version control scenario: managing multiple datasets and ML model versions using DVC commands.
+---
+
 # Tutorial: Data & Model Versioning
 
 The goal of this example is to give you some hands-on experience with a basic
--- a/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
+++ b/content/docs/use-cases/versioning-data-and-model-files/tutorial.md
@@ -5,7 +5,7 @@ learning version control scenario: managing multiple datasets and ML models
 using DVC.'
 ---
 
-# Tutorial: Data & Model Versioning
+# Data and Model Versioning Tutorial
 
 The goal of this example is to give you some hands-on experience with a basic
 machine learning version control scenario: managing multiple datasets and ML


jorgeorpinel:
That's not what I see in the search result for those 2 terms, respectively 🤔 

image

image

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
```diff
@@ -1,3 +1,10 @@
+---
+title: 'Tutorial: Data & Model Versioning'
```
jorgeorpinel, line 2:
```suggestion
title: 'Data and Model Versioning Tutorial'
```
We should probably update the doc title and nav entry from & to "and".

jorgeorpinel: Same for "Sharing Data & Model Files" (nav entry)

jorgeorpinel: Unless "and" breaks the line (in the nav entries)

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
```diff
@@ -1,3 +1,10 @@
...
+description: 'Get experience with data versioning in a basic machine learning
+version control scenario: managing multiple datasets and ML model versions using
+DVC commands.'
```
jorgeorpinel, line 5: Keep in mind we're changing the use case a lot in #1747 — basically rewritten. It has the same ideas but maybe reuse some of the existing text + insert any needed terms?

jorgeorpinel: BTW can you please review the text there to make sure we won't lose any traffic from of the terms that current rank high? Thanks

jorgeorpinel: Oh never mind, this is the tutorial, I see. We're rewriting the use case only so far.

jorgeorpinel: What about the term "dataste versioning" you mentioned in the PR description, should it be included here or featured more in the text? (Currently it's buried under **Automating capturing**).

jorgeorpinel: - Why not "hands-on experience" like in the intro?
- Is "machine learning version control" too much? Should the terms be separated (same in the actual text).
- Related: term "version(ing)" is probably repeated too much in this description (3 times).
- End at "using DVC."

jorgeorpinel (commented):
Left specific comments on the proposed description. Also:

- What about the term "dataste versioning" you mentioned in the PR description, should it be included in the SEO desc. and/or featured more in the text? (Currently it's buried under **Automating capturing**).
- What other terms can we include in the doc/desc that currently don't rank as high? I feel like optimizing terms that already rank high is not the lowest hanging fruit (although improving CTR/ minimizing bounces seems helpful of course).

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
jeremydesroches: For the meta title and description fields, we have to be careful about character length. In this case, the expanded title with the & is very close to the max length (<60 characters for title, <160 for description). Beyond that, Google will cut it off and append with an ellipsis. Here's an example with the new title mocked up:

image

The meta title is the most important field SEO-wise, because it's actually checked for matching terms. Where it can be expanded with key terms (more likely for tutorials and use cases than command references), being economical with space helps to fit the needed terms. 

Yes, 'and' reads better, but it's okay to have both: the on page and navigation title can be 'and' while the meta title uses '&'.

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
jeremydesroches: > can you please review the new text in #1747 to make sure we won't lose any traffic from of the terms that current rank high?

Will do.

> * Why not "hands-on experience" like in the intro?

Trying to keep it under 160 characters, but by removing 'commands' from the end I was able to add this.

> * Is "machine learning version control" too much? Should the terms be separated (same in the actual text).
> * Related: term "version(ing)" is probably repeated too much in this description (3 times).

"machine learning version control" is an important search term that DVC.org already ranks in position 1 for, probably due to mentions like this. So we should keep it if possible. 😉

Agree, 'version(ing)' mentioned three times is too much, so I removed the last instance of 'version' in the description and mirrored in actual text. 'data versioning' and 'machine learning version control' are the two important mentions to keep.

> * End at "using DVC."

Done, and mirrored in actual text.

jeremydesroches:
> * What about the term "dataset versioning" you mentioned in the PR description, should it be included in the SEO desc. and/or featured more in the text? (Currently it's buried under **Automating capturing**).

For the meta description, "data versioning" matches to "dataset versioning" as far as being highlighted (shown in your latest screens). While the description doesn't affect rank, the matching variations and strong page position for "dataset versioning" (No. 3) make me think it doesn't warrant further optimization on that term.

> * What other terms can we include in the doc/desc that currently don't rank as high? I feel like optimizing terms that already rank high is not the lowest hanging fruit (although improving CTR/ minimizing bounces seems helpful of course).

This PR was meant to be only for optimization, indeed improving CTR and minimizing bounces, but the title addition could also have a significant effect on the position. A quick look doesn't show any other terms ranking lower for this page that would be easy to add. Like all the others, I'll consider this page for more in-depth changes to rank on new terms (out of scope for this PR).

jeremydesroches:
> That's not what I see in the search result for those 2 terms, respectively 🤔

I see the same. Good example of how quickly things change and of "experiments" the engine runs on its own: switching between descriptions, titles, etc. Sometimes it will even discard a custom description if it thinks it has a better idea. Keeps things interesting!

jeremydesroches:
> > That's not what I see in the search result for those 2 terms, respectively 🤔
> 
> I see the same. Good example of how quickly things change and of "experiments" the engine runs on its own: switching between descriptions, titles, etc. Sometimes it will even discard a custom description if it thinks it has a better idea. Keeps things interesting!

Further to this, just came across Google pre-emptively editing the title for us, in another set of results:

image

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
jorgeorpinel: > the expanded title with the & is very close to the max length (<60 characters

As long as it's the last bit of  `Data Version Control · DVC` that gets truncated, it doesn't sound like a big deal to me. In any case, my suggestion is 1 char longer so should we commit it? (and apply to the doc's H1 title)

> while the meta title uses '&'.

Actually removing & from the nav in the use case rewrite anyway...

jorgeorpinel: \*Actually we're removing...

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
jorgeorpinel: > > BTW can you please review the new text in #1747

> Will do.

Thanks but actually don't worry about this yet. I'll request your review from there when ready.

jorgeorpinel: OK about "machine learning version control"!

Thanks for the other updates. Resolving

jorgeorpinel:
> just came across Google pre-emptively editing the title

Woah 😮 I don't see that rn but that's crazy. I guess it makes sense that it experiments with lots of A/B tests to see what users engage with more.

jorgeorpinel (approved):
Looks good, I'd just remove that '&` altogether for consistency, as we won't be using any other ampersand in titles or navs soon.

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
jorgeorpinel: > my suggestion is 1 char longer so should we commit it? (and apply to the doc's H1 title)

Applying this myself to move on. Thanks again @jeremydesroches !

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
```diff
@@ -1,8 +1,15 @@
+---
+title: 'Tutorial: Data & Model Versioning'
+description: 'Get hands-on experience with data versioning in a basic machine
+learning version control scenario: managing multiple datasets and ML models
+using DVC.'
+---
+
 # Tutorial: Data & Model Versioning
```
jorgeorpinel, line 8:
```suggestion
# Tutorial: Data & Model Versioning
```

content/docs/use-cases/versioning-data-and-model-files/tutorial.md
```diff
@@ -1,8 +1,15 @@
-# Tutorial: Data & Model Versioning
+---
+title: 'Data and Model Versioning Tutorial'
+description: 'Get hands-on experience with data versioning in a basic machine
+learning version control scenario: managing multiple datasets and ML models
+using DVC.'
+---
+
+# Data and Model Versioning Tutorial
```
jorgeorpinel, line 8: If for reason having "Tutorial" first was intended as a SEO improvement, please open another quick PR for that. Thanks

jeremydesroches: Yes, as mentioned in #1857 that was the intent. Will wait on consensus there and open a new PR if needed.

jorgeorpinel: OK please reapply your change in a small PR or throw it in with the next SEO optimization PR you open. Thanks

jeremydesroches:

**Tutorial: Data Versioning** - [Link](https://search.google.com/u/0/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fuse-cases%2Fversioning-data-and-model-files%2Ftutorial&breakdown=query&metrics=CLICKS%2CIMPRESSIONS%2CCTR%2CPOSITION&start_date=20201023&end_date=20201204&compare_start_date=20200910&compare_end_date=20201022) - 12% more clicks, 8% more impressions, and 3.8% increase in clickthrough. Average position improved from 4.8 to 4.2. Notably, added clicks and impressions were on more "long tail" terms while maintaining performance for existing key terms. 👍 

Also see #1906 for data before/after prefix was added to the title and h1.

jorgeorpinel:
Great, so for this one every metric improved slightly. [Aligned 5w before and after](https://search.google.com/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fuse-cases%2Fversioning-data-and-model-files%2Ftutorial&breakdown=query&metrics=CLICKS%2CIMPRESSIONS%2CCTR%2CPOSITION&start_date=20201021&end_date=20201125&compare_start_date=20200916&compare_end_date=20201021) merge to confirm:

image

Some interesting (mixed) results per term below. But I suspect that some of the drops may have to do with new docs we have that have capture them instead.

image
image

image

image

> Notably, added clicks and impressions were on more "long tail" terms

Could you post some examples @jeremydesroches ?

jeremydesroches:
> Like "model versioning" above? Could you post some other examples @jeremydesroches ?

Yep, exactly. "model versioning" gained clicks and impressions and "version control data" went from 1 impression to 200. Using your timeframe:

image

Here's another group that is also typical for "long tail" in terms of format... usually longer terms and different combinations. (columns are Impressions post-merge, pre-merge, and difference).

image

Mixed results are always difficult to interpret, so sometimes it's easier to see effects in the lower-volume/long-tail terms. A 10x increase is a clear signal that _something_ happened.

jorgeorpinel:
Makes sense. I think we definitely improved the accuracy of the search results at least.

PR #1857, opened 2020-10-12
start: add meta title, description and key term (data pipelines) [SEO]

jeremydesroches:
**Meta title**
Added a custom title to prefix title with "Get Started:" in search result. Edited page H1 to match. Suggest using this pattern for all Get Started items and Tutorial. Adding prefixes like this should improve clickthrough and reduce bounce by clearly showing the result is a "Get Started" article (or Tutorial, in that case) as opposed to command reference, etc. 

**Meta description**
Added a custom meta description to the front matter with overview and key terms. 

**Key term**
Added second paragraph to expand on use cases and add "machine learning pipelines".

This page doesn't rank on the first page for either "data pipelines" or "machine learning pipelines". The intent of these changes is to improve placement for "machine learning pipelines". Current results for that term align well with the DVC use cases.

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -1,9 +1,11 @@
 ---
-description: 'Learn how DVC data pipelines can help you organize, version, and
-efficiently reproduce your data science or machine learning project.'
+title: 'Get Started: Data Pipelines'
+description: 'Get started with pipelines in DVC. Learn how to build data
+pipelines to organize, version, and reproduce your data science and machine
+learning processes.'
 ---
 
-# Data Pipelines
+# Get Started: Data Pipelines
 
 Versioning large data files and directories for data science is great, but not
 enough. How is data filtered, transformed, or used to train ML models? DVC
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -13,8 +13,8 @@ that produce a final result.
 
 DVC pipelines and their data can also be easily versioned (using Git). This
 allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally! This allows you to
-capture simple ETL workflows, better organize data science projects, or build detailed
+results later exactly as they were built originally! This allows you to capture
+simple ETL workflows, better organize data science projects, or build detailed
 machine learning pipelines (to name a few uses).
 
 ## Pipeline stages
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -312,7 +312,7 @@ important problems:
   and which commands will generate the pipeline results (such as an ML model).
   Storing these files in Git makes it easy to version and share.
 - _Continuous Delivery and Continuous Integration (CI/CD) for ML_ - describing
-  projects in way that it can be reproduced (built) is the fist necessary step
+  projects in way that it can be reproduced (built) is the first necessary step
   before introducing CI/CD systems. See our sister project,
   [CML](https://cml.dev/) for some examples.
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -1,8 +1,7 @@
 ---
 title: 'Get Started: Data Pipelines'
-description: 'Get started with pipelines in DVC. Learn how to build data
-pipelines to organize, version, and reproduce your data science and machine
-learning processes.'
+description: 'Get started with pipelines in DVC. Learn how to capture, organize,
+version, and reproduce your data science and machine learning workflows.'
 ---
 
 # Get Started: Data Pipelines
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -11,14 +11,11 @@ enough. How is data filtered, transformed, or used to train ML models? DVC
 introduces a mechanism to capture _data pipelines_ — series of data processes
 that produce a final result.
 
-A DVC pipeline uses stages to describe how models and other data artifacts are
-built, and provides an efficient way to reproduce them. This allows you to
-capture simple data workflows, organize data science projects, or build detailed
-machine learning pipelines (to name a few use cases).
-
-You can easily version DVC pipelines and their data (using Git). This allows you
-to better organize your project and reproduce your workflow and results later —
-exactly as they were built originally!
+DVC pipelines and their data can also be easily versioned (using Git). This
+allows you to better organize your project, and reproduce your workflow and
+results later exactly as they were built originally! This allows you to
+capture simple ETL workflows, better organize data science projects, or build detailed
+machine learning pipelines (to name a few uses).
 
 ## Pipeline stages
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -1,6 +1,6 @@
 ---
 title: 'Get Started: Data Pipelines'
-description: 'Get started with pipelines in DVC. Learn how to capture, organize,
+description: 'Learn how to build and use DVC pipelines to capture, organize,
 version, and reproduce your data science and machine learning workflows.'
 ---
 
@@ -12,10 +12,10 @@ introduces a mechanism to capture _data pipelines_ — series of data processes
 that produce a final result.
 
 DVC pipelines and their data can also be easily versioned (using Git). This
-allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally! This allows you to capture
-simple ETL workflows, better organize data science projects, or build detailed
-machine learning pipelines (to name a few uses).
+allows you to better organize projects, and reproduce your workflow and results
+later — exactly as they were built originally! For example, you could capture a
+simple ETL workflow, organize a data science project, or build a detailed
+machine learning pipeline.
 
 ## Pipeline stages
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -1,6 +1,6 @@
 ---
-description: |
-  Learn how DVC data pipelines can help you organize, version, and efficiently reproduce your data science or machine learning project.
+description: 'Learn how DVC data pipelines can help you organize, version, and
+efficiently reproduce your data science or machine learning project.'
 ---
 
 # Data Pipelines
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -1,3 +1,8 @@
+---
+description: |
+  Learn how DVC data pipelines can help you organize, version, and efficiently reproduce your data science or machine learning project.
+---
+
 # Data Pipelines
 
 Versioning large data files and directories for data science is great, but not
@@ -5,9 +10,14 @@ enough. How is data filtered, transformed, or used to train ML models? DVC
 introduces a mechanism to capture _data pipelines_ — series of data processes
 that produce a final result.
 
-DVC pipelines and their data can also be easily versioned (using Git). This
-allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally!
+A DVC pipeline uses stages to describe how models and other data artifacts are
+built, and provides an efficient way to reproduce them. This allows you to
+capture simple data workflows, organize data science projects, or build detailed
+machine learning pipelines (to name a few use cases).
+
+You can easily version DVC pipelines and their data (using Git). This allows you
+to better organize your project and reproduce your workflow and results later —
+exactly as they were built originally!
 
 ## Pipeline stages
 


content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,23 @@
+---
+description: |
+  Learn how DVC data pipelines can help you organize, version, and efficiently reproduce your data science or machine learning project.
```
shcheklein, line 3: We haven't found a good way to automate this, but let's still wrap this at 80 please

jeremydesroches: Thanks for catching this. Done, and I'll apply going forward.

content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,23 @@
+---
+description: 'Learn how DVC data pipelines can help you organize, version, and
+efficiently reproduce your data science or machine learning project.'
```
jorgeorpinel, line 3:
```suggestion
reproduce your data science or machine learning project efficiently.'
```

jorgeorpinel: But do we need that term? "efficient"

jorgeorpinel: data science or machine learning project -> data science and machine learning processes?

Or some other wording that makes it specifically about the aspect of codifying pipelines (as dvc.yaml files). "Project" seems too generic.

content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,23 @@
...
-allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally!
+A DVC pipeline uses stages to describe how models and other data artifacts are
+built, and provides an efficient way to reproduce them. This allows you to
+capture simple data workflows, organize data science projects, or build detailed
+machine learning pipelines (to name a few use cases).
+
+You can easily version DVC pipelines and their data (using Git). This allows you
+to better organize your project and reproduce your workflow and results later —
+exactly as they were built originally!
```
jorgeorpinel, line 20: The thing with Get Started pages is that we don't want to get in the way of the hands-on sections so 3 paragraphs for intro is probably too much. In this case I think inserting the target terms (machine learning pipelines, efficient?) in existing text would be better.

Not sure we need to expand on use cases here anyway, we have /doc/use-cases for that. Maybe link to the appropriate use case instead (if not linked yet).

content/docs/start/data-pipelines.md
jeremydesroches: Reworked this, PTAL.

content/docs/start/data-pipelines.md
jeremydesroches: As mentioned in the PR description (and discussed offline), "data pipelines" and "machine learning pipelines" are high potential terms which dvc.org does not yet rank well for. 

This doc is a good example of "lowest hanging fruit" (like you mentioned in #1856). It has the most relevant content for the above terms — matching the intent and use case for both. 

The expanded intro is a good way to add the key terms (necessary) and explain the context to people who might be visiting DVC for the first time (warranted, in lieu of other pipeline content). 

> Maybe link to the appropriate use case instead (if not linked yet).

What would that be for pipelines? Aside from this doc, the next best candidates I see are the command references for `dag` and `repro`. Am I missing something?

content/docs/start/data-pipelines.md
jeremydesroches: Also added a meta title, keeping with the pattern for other Getting Started and Tutorial docs.

content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,25 @@
-# Data Pipelines
+---
+title: 'Get Started: Data Pipelines'
+description: 'Get started with pipelines in DVC. Learn how to build data
+pipelines to organize, version, and reproduce your data science and machine
+learning processes.'
```
jorgeorpinel, line 5: Copy edit:
```suggestion
description: 'Get started with pipelines in DVC. Learn how to capture, organize,
version, and reproduce your data science and machine learning workflows.'
```
(will commit)

content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,24 @@
-# Data Pipelines
+---
+title: 'Get Started: Data Pipelines'
+description: 'Get started with pipelines in DVC. Learn how to capture, organize,
```
jorgeorpinel, line 3: Is it desirable to repeat the title more or less in the description? Curious

content/docs/start/data-pipelines.md
jorgeorpinel: OK thanks.

content/docs/start/data-pipelines.md
jorgeorpinel: > This doc is a good example of "lowest hanging fruit" (like you mentioned in #1856). It has the most relevant content for "data pipelines" and "machine learning pipelines" (high potential terms which dvc.org does not yet rank well for).

Agree, let's definitely target them here (and then in other related docs as well).

> The expanded intro is a good way to add the key terms (necessary) and explain the context to people who might be visiting DVC for the first time

We want to add "machine learning pipelines", yes. But we don't have to create a whole new paragraph for that. Can we try to find an alternative place to insert it instead?

The part about explaining more context seems out of scope for this PR to me. We could discuss that separately (feel free to open a separate issue) but the new text, esp. "A DVC pipeline uses stages to describe..." is already implicit when you read the doc. (GS pages try to be more demonstrative than descriptive).

> the next best candidates I see are the command references for dag and repro

As for where else to optimize these terms, indeed dat and repo are good, as well as `run` but probably the more impactful one may be in https://dvc.org/doc/user-guide/dvc-files-and-directories#dvcyaml-file

content/docs/start/data-pipelines.md
```diff
@@ -1,13 +1,24 @@
...
-allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally!
+A DVC pipeline uses stages to describe how models and other data artifacts are
+built, and provides an efficient way to reproduce them. This allows you to
+capture simple data workflows, organize data science projects, or build detailed
+machine learning pipelines (to name a few use cases).
+
+You can easily version DVC pipelines and their data (using Git). This allows you
+to better organize your project and reproduce your workflow and results later —
+exactly as they were built originally!
```
jorgeorpinel, line 21: Simple workaround to https://github.com/iterative/dvc.org/pull/1857#pullrequestreview-509809200:

```suggestion
DVC pipelines and their data can also be easily versioned (using Git). This
allows you to better organize your project, and reproduce your workflow and
results later exactly as they were built originally! This allows you to
capture simple data workflows, organize data science projects, or build detailed
machine learning pipelines (to name a few use cases).
```
> Needs formatting

content/docs/start/data-pipelines.md
jorgeorpinel: Posted a possible resolution in https://github.com/iterative/dvc.org/pull/1857#pullrequestreview-512241072 below.

jeremydesroches: > Posted a possible resolution in #1857 (review) below.

Nice! I like how you were able to condense it and preserve the meaning.

It's a good idea to address this particular term "machine learning pipelines" and add context across `dag` `repro` `run` and [files and directories](https://dvc.org/doc/user-guide/dvc-files-and-directories#dvcyaml-file). I'll take you up on that and create a separate issue for it.

content/docs/start/data-pipelines.md
jeremydesroches: Whatever is best to appeal to the audience. It isn't used for indexing purposes, but including terms that match popular searches (which might then get highlighted) can be helpful to that end. The key objective is increasing clickthrough and reducing bounce, so a copy/technical writing mindset works well.

In this case, a lot of the searches are quite general: "machine learning pipelines", "ml pipelines". So I wrote the description to appeal to people who know they want a pipeline, but might not have heard of DVC or built a pipeline before. (Sometimes challenging in < 160 characters, haha)

content/docs/start/data-pipelines.md
```diff
@@ -1,4 +1,10 @@
-# Data Pipelines
+---
+title: 'Get Started: Data Pipelines'
+description: 'Get started with pipelines in DVC. Learn how to capture, organize,
+version, and reproduce your data science and machine learning workflows.'
+---
+
+# Get Started: Data Pipelines
```
jorgeorpinel, line 7: > Added a custom title to prefix title with "Get Started:" in search result. Edited page H1 to match. Suggest using this pattern for all Get Started items and Tutorial. Adding prefixes like this should improve clickthrough and reduce bounce by clearly showing the result is a "Get Started"

Oh, I didn't realize we already did that for https://dvc.org/doc/start/data-versioning in fact!

Hmmmm... Is the reason we're changing the H1 that the HTML title is taken from there @jeremydouglass ? If so we could change the engine to use the SEO title instead, and keep a lean H1 in the page body. Please lmk before I request that feature unnecessarily.

content/docs/start/data-pipelines.md
jorgeorpinel: Right, but my question is more basic than that: if the search result title is "Get Started: Data Pipelines" , isn't starting the description with "Get started with pipelines..." a waste of real estate (since we only have 160 chars)? Seems kind of redundant to me (it's not like people don't read the title but only read the desc in a serach result).

content/docs/start/data-pipelines.md
```diff
@@ -7,7 +13,9 @@ that produce a final result.
 
 DVC pipelines and their data can also be easily versioned (using Git). This
 allows you to better organize your project, and reproduce your workflow and
-results later exactly as they were built originally!
+results later exactly as they were built originally! This allows you to capture
+simple ETL workflows, better organize data science projects, or build detailed
+machine learning pipelines (to name a few uses).
```
jorgeorpinel, line 18: Actually this text has a couple phrasing issues:
- DVC pipelines... allows you to... build detailed machine learning pipelines? Idk, seems like a circular explanation
- "This allows you to" is repeated.

content/docs/start/data-pipelines.md
jorgeorpinel: Thanks. Left a small follow up about that in https://github.com/iterative/dvc.org/pull/1857#pullrequestreview-516465031, actually.

> I'll take you up on that and create a separate issue for it.

Please do! Resolving here.

jorgeorpinel (commented):
Left a couple more (minor) comments/questions ☝️ 

Getting there. Let's wrap this up so we can see the next batch of updates! Thanks

content/docs/start/data-pipelines.md
jorgeorpinel: Actually I checked and the SEO title is what's used for the HTML head title tag. So do we need the H1 title to be the same? I personally prefer the old way but it's not a strong opinion so if there's a reason, even if it's just for a potential benefit, we can def. consider it.

content/docs/start/data-pipelines.md
jeremydesroches: OK, edited it to avoid repetition.

content/docs/start/data-pipelines.md
jeremydesroches: The Get Started docs make a better case for using the "Get Started: Data Pipelines" format for both H1 and SEO title. I think you'd agree that's better than "Data Pipelines Get Started". 

For the tutorial, (as in #1856) "Tutorial: Data and Model Versioning" fits the standard for how Google is recommending titles for search results. Another tactic/example, used elsewhere, is the title itself as the prefix, something like "Widgets: Do Cool Widgety Things | Widget Corp." 

Tying the two together, I think it's best for the SEO title (what's clicked on) to match the H1 so there isn't a disconnect on landing (page title matches the search result). So I'm advocating for "Get Started: Data Pipelines" format for both SEO title and H1.

What do you think @jorgeorpinel ?

content/docs/start/data-pipelines.md
jeremydesroches: OK, edited this to fix phrasing while avoiding too much copy editing. PTAL

content/docs/start/data-pipelines.md
jorgeorpinel: Ty

content/docs/start/data-pipelines.md
jorgeorpinel: OK sure, if there's a SEO-related reason then it's in scope. Let's do it!

content/docs/start/data-pipelines.md
jorgeorpinel: Looks good 👍

jorgeorpinel (approved):

jorgeorpinel:
Thanks @jeremydesroches, neeeext

jeremydesroches:

**Get Started: Data Pipelines** - [Link](https://search.google.com/u/0/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=*https%3A%2F%2Fdvc.org%2Fdoc%2Fstart%2Fdata-pipelines&breakdown=query&metrics=CLICKS%2CIMPRESSIONS&start_date=20201030&end_date=20201204&compare_start_date=20200924&compare_end_date=20201029) - 7% more clicks and 15% more impressions over the five-week comparison period. Clickthrough and average position down slightly. Promising improvement in position (but still on page 5-6) for high-potential general terms like "data pipeline examples" and "creating data pipelines". 👍

Issue #1824, opened 2020-09-29
cmd refs: Updates per 1.x


imhardikj:
All cmd refs that need to be updated per 1.x
Context: https://github.com/iterative/dvc.org/issues/1255#issuecomment-698134751

Mainly descriptions, or everything
- [x] `repro` - #1632
- ~~`dag` (rewrite description #1500)~~
- [x] `commit` (example for adding missing deps/outs #460/#612) #1840
- [x] `list` ~~+ #1521?~~
- [x]  `target` arg descriptions - #1574

Mainly examples
- [x] `run` examples in other docs ~~(also #1700 and #186)~~
- [x] `status` examples in all docs (including consistent indentation — may keep tabs) - #1952
- ~~pending `plots` updates #1175~~ #1414
- [x] `fetch` (examples) - #1903
- [x] `pull` (examples) - #1903
- [x] `get` (examples) - #1903
- [x] `params diff` (examples #1743) -
- [x] `version` - #1635
- [x] `root` (new example #404)

jorgeorpinel:
@imhardikj please lmk when you've reviewed this list to make each check box more specific as we discussed, and I'll review and prioritize the list myself then. Thanks

jorgeorpinel:
Closed by #2075.

PR #1861, opened 2020-10-13
repro cmdref updates

### LINKED ISSUE #1632: repro: improvements to targets arg and -R option
> Extracted from https://github.com/iterative/dvc.org/pull/1572#discussion_r461232283
And per private discussion with Ruslan rel. to iterative/dvc/issues/4292

- [x] Update `targets` description to `Stage or path to dvc.yaml or .dvc file to reproduce` for now,
  see resolution of iterative/dvc/issues/4292
- [x] This should be clarified in the -R option desc: It expands what the `targets` argument accepts (directory paths).
  Make sure options -c and -P are also clear and complete, as they also allow to target directories in different ways.
- [x] No example in the doc for using `repro` with a specific target stage or .dvc file is provided... I think this should be done

imhardikj:
Fixes #1632 

- [x] Update `targets` description
- [x] example for using repro with a specific target stage 
- [x] -R option desc

Partially addresses #1824


content/docs/command-reference/repro.md
```diff
@@ -110,7 +110,8 @@ up-to-date and only execute the final stage.
   reproduced independently.
 
 - `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory (if any) and their subdirectories.
+  directory and their subdirectories. These directories can be given as
+  `targets`.
```
imhardikj, line 114: @jorgeorpinel does the option description need to explicitly convey that -R will work when at least one of the target is directory.

content/docs/command-reference/repro.md
```diff
@@ -14,7 +14,7 @@ usage: dvc repro [-h] [-q | -v] [-f] [-s] [-c <path>] [-m] [--dry] [-i]
                  [targets [targets ...]]
 
 positional arguments:
-  targets        Stage or .dvc file to reproduce
+  targets        Stage, path to dvc.yaml or .dvc file to reproduce
```
jorgeorpinel, line 17: Why comma? Just need to copy/paste the text from https://github.com/iterative/dvc.org/pull/1861

jorgeorpinel: Or use commas consistently (one is missing after "dvc.yaml" above).

content/docs/command-reference/repro.md
jorgeorpinel: > searching each directory and their subdirectories

This is not clear. What directory? "target directory" was better, (even if the next sentence explains it — it should be clear first that -R expands what `targets` accepts).
Also "directory and their" mixes singular and plural (incorrect grammar).

> need to explicitly convey that -R will work when at least one of the target is directory

Yes, good idea. Let's be explicit about that. Use the phrase "only has an effect if" please.

jorgeorpinel: Also, you may have missed

> Make sure options -c and -P are also clear and complete, as they also allow to target directories in different ways.

From #1861

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,41 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
 information: updated dependency/output file hash values, and a new result,
 respectively.
 
+## Example: Specific stage
+
+> This example continues the previous one.
+
+We can use `dvc repro` to reproduce a pipeline partially till a specific stage.
```
jorgeorpinel, line 263: "till" is too informal

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,41 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
...
+Let's add this line to `text.txt` (the input of our first stage, created in the
+previous example):
```
jorgeorpinel, line 265:
```suggestion
Let's add this line to `text.txt`:
```
Removing some redundant info ^

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,40 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
 information: updated dependency/output file hash values, and a new result,
 respectively.
 
+## Example: Specific stage
+
+> This example continues the previous one.
+
+We can use `dvc repro` to reproduce a pipeline partially till a specific stage.
+Let's add this line to `text.txt`:
```
jorgeorpinel, line 264:
```suggestion
Let's add a line to `text.txt` , for example:
```

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,40 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
...
+
+We can use `dvc repro` to reproduce a pipeline partially till a specific stage.
+Let's add a line to `text.txt` , for example:
+
+```
+...
+4321
+```
+
+Now, specify the `count` stage as a `target` with `dvc repro`. This results in
```
jorgeorpinel, line 271: Let's use filter here, otherwise this is not really a partial reproduction as the whole pipeline will be reproduced. Also because it's the same as the next example, basically.

jorgeorpinel: Or introduce an intermediate step if you are trying to show that using a target can also reproduce the previous stages, so that the next example is not so repetitive (in it's effects).

jorgeorpinel: Actually it would be ideal to have that intermediate step so it can be used with --downstream in the next example, to show the contrast between this example (target the middle stage which runs the previous ones) vs downstream (BTW also targets a stage, but the behavior is different because of the --downstream option).

jorgeorpinel: BUT we shuold probably extract that part to a separate issue to keep this PR simple.

jorgeorpinel: > downstream (BTW also targets a stage

This makes me think that another solution is to combine the 2 examples by basically just clarifying in the old one that a stage name target is being used (combined with --downstream). For example the title could be **Example: Downstream from a target stage**.

jorgeorpinel: WDYT?

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,40 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
...
+Updating lock file 'dvc.lock'
+
+Running stage 'count' with command:
+	    python process.py numbers.txt > count.txt
+Updating lock file 'dvc.lock'
+```
+
 ## Example: Downstream
 
 > This example continues the previous one.
```
jorgeorpinel, line 288: This is no longer exactly true atm.

jorgeorpinel (changes requested):
Thanks @imhardikj. This needs some more work on the options, and to think through the examples. Please see discussions above.

rogermparent:
I see this failed deploy, and I believe it's been affected by the recent change to how the community page's schema works. I believe merging master will fix this, and I'm trying it out locally.

I can do this push if you'd like to see if it fixes the deploy.

jorgeorpinel:
Thank for the note Roger. Please merge `master` @imhardikj, thanks.

content/docs/command-reference/repro.md
imhardikj: `-c` and `-P` option are complete I think.
Like `-P` doesn't require `target` (It is working without them).
-c` option's description also looks completes.

content/docs/command-reference/repro.md
imhardikj: Combining them both is better I think. As both example requires a stage as target. This doesn't require much changes in current example. Maybe we can add a note(or separate paragraph) that says without `--downstream` pipeline would be reproduces until the target and not further.

content/docs/command-reference/repro.md
jorgeorpinel: Looks like -c is OK but -P is not complete. I'll reopen this review... Resolving here.

content/docs/command-reference/repro.md
```diff
@@ -110,7 +110,8 @@ up-to-date and only execute the final stage.
   reproduced independently.
 
 - `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory (if any) and their subdirectories.
+  target directory and its subdirectories. This option only has an effect if a
+  directory is given as one of the `targets`.
```
jorgeorpinel, line 114: This change doesn't clarify that this option "expands what the targets argument accepts (directory paths)." If you're unsure what the issue requirements mean please ping me @imhardikj, thanks.

jorgeorpinel: The same applies for -P, it should clarify the same idea.

jorgeorpinel: > expands what the targets argument accepts (directory paths)

This refers to the top description of `targets`: "Stage or path to dvc.yaml or .dvc file to reproduce" <-- notice it doesn't include directories, nor are dirs mentioned as targets anywhere in the desc. That's why clarification is needed.

content/docs/command-reference/repro.md
```diff
@@ -255,14 +256,14 @@ You can now check that `dvc.lock` and `count.txt` have been updated with the new
...
-## Example: Downstream
+## Example: Downstream from a target stage
 
 > This example continues the previous one.
 
-The `--downstream` option allows us to only reproduce results from commands
-after a specific stage in a pipeline. To demonstrate how it works, let's make a
-change in `text.txt` (the input of our first stage, created in the previous
-example):
+The `--downstream` option when used with a stage `target` allows us to only
```
jorgeorpinel, line 263:
```suggestion
The `--downstream` option, when used with a `target` stage, allows us to only
```

content/docs/command-reference/repro.md
jorgeorpinel: OK. Ideally examples should focus on a single thing but I guess in this case we can try the approach.

content/docs/command-reference/repro.md
```diff
@@ -279,7 +280,7 @@ print(num_lines)
 ```
 
 Now, using the `--downstream` option with `dvc repro` results in the execution
-only of the target stage, and following ones (none in this case):
+of only the `target` stage `count`, and following ones (none in this case):
```
jorgeorpinel, line 283:
```suggestion
of only the target (`count`) and following stages (none in this case):
```

jorgeorpinel (changes requested):
The options still need clarification. Thanks

content/docs/command-reference/repro.md
imhardikj: Yeah I realize my mistake. `-R` option will expand paths if they are given as targets. Then it will be search for `dvc.yaml` or `.dvc` in that path (and folders inside it), and then reproduce them.
In case of `-P`, I checked that it works without any targets given. So basically it expands the path of the DVC project and find all targets to reproduce from it.

jorgeorpinel: No, you're not getting me yet. Maybe I'm not being clear? I wasn't talking about "expanding paths". Once more, these options "expand what the `targets` argument accepts" i.e. usually targets only accept "Stage or path to dvc.yaml or .dvc file to reproduce", right? It doesn't say that paths to directories are accepted. With these options, paths to directories are also accepted.

Like I mentioned, please ping me if you're not 99% sure you're understanding me. Please go over this again and lmk when to review. Thanks

jorgeorpinel (changes requested):
Please see https://github.com/iterative/dvc.org/pull/1861#discussion_r511520908

content/docs/command-reference/repro.md
jorgeorpinel: > It turns out that with these options, paths to directories are also accepted even if it's not stated in the Synopsis/usage.

So another possible solution is to change the usage text to include "directories" but then we have to clarify somewhere that dirs are only used by the command if -PR are used) so I'm not sure which way is better/ easier to explain. Let's stick to the current approach unless you have a strong opinion.

content/docs/command-reference/repro.md
```diff
@@ -109,8 +109,10 @@ up-to-date and only execute the final stage.
   for example with subdirectories containing a separate pipeline that can be
   reproduced independently.
 
-- `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory (if any) and their subdirectories.
+- `-R`, `--recursive` - expands arguments that `targets` accepts to determine
+  the stages to be reproduced, by searching each target directory and its
+  subdirectories. This option only has an effect if directory paths are given as
+  `targets`.
```
jorgeorpinel, line 115: First, this is a confusing text. It should be explained in simpler terms, shorter sentences, etc. Also, it has grammar errors.

jorgeorpinel: Secondly we're missing the whole point of #1632... Maybe you're confused by the term "expands". One thing is to expand a path (explore recursively) — that's not what the issue is about at all. Just forget about that term...

This is about the targets argument of dvc repro (and how -RP affect it). Again (from https://github.com/iterative/dvc.org/pull/1861#discussion_r511520908), "i.e. usually targets only accept "Stage or path to dvc.yaml or .dvc file to reproduce", right? It does not say that paths to directories are accepted. But these options (-RP) only work for directories, so what's going on?"

If you can answer that question, you'll be closer to understanding what's needed here.

content/docs/command-reference/repro.md
jorgeorpinel: Continued in https://github.com/iterative/dvc.org/pull/1861#pullrequestreview-517073211. Resolving here

jorgeorpinel:
Please focus on this one @imhardikj (I noticed you submitted changes to other PRs first). 1.x updates are priority no. 1, the rest are for the meantime in between. Thanks

content/docs/command-reference/repro.md
```diff
@@ -14,7 +14,8 @@ usage: dvc repro [-h] [-q | -v] [-f] [-s] [-c <path>] [-m] [--dry] [-i]
                  [targets [targets ...]]
 
 positional arguments:
-  targets        Stage or .dvc file to reproduce
+  targets        Stage or path to dvc.yaml or .dvc file to reproduce. Using -R,
+                 directories to search for stages can also be given.
```
jorgeorpinel, line 18: Perfect. We will need to submit a matching change in the core `dvc` repo. Are you able to try that too @imhardikj ? Thanks

content/docs/command-reference/repro.md
```diff
@@ -110,7 +111,8 @@ up-to-date and only execute the final stage.
   reproduced independently.
 
 - `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory (if any) and their subdirectories.
+  target directory and its subdirectories. If there are no directories among the
+  `targets`, this option is ignored.
```
jorgeorpinel, line 115:
```suggestion
  target directory and its subdirectories. It has no effect if there are no
  directories among the `targets`.
```

content/docs/command-reference/repro.md
```diff
@@ -110,7 +111,8 @@ up-to-date and only execute the final stage.
   reproduced independently.
 
 - `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory (if any) and their subdirectories.
+  target directory and its subdirectories. It has no effect if there are no
+  directories among the `targets`.
```
jorgeorpinel, line 115:
```suggestion
  target directory and its subdirectories for stages (in `dvc.yaml`) to inspect.
  If there are no directories among the targets, this option is ignored.
```

content/docs/command-reference/repro.md
```diff
@@ -132,7 +134,7 @@ up-to-date and only execute the final stage.
   to. Use `dvc dag <target>` to show the parent pipeline of a target.
 
 - `-P`, `--all-pipelines` - reproduce all pipelines for all `dvc.yaml` files
-  present in the DVC project.
+  present in the <abbr>DVC project</abbr>.
```
jorgeorpinel, line 137: Why this change? Completely unrelated to the PR. Also, we don't use `<abbr>` in Options (it's too much to have tooltips here).
```suggestion
  present in the DVC project.
```

jorgeorpinel (commented):
OK this is good so far. I guess changing the usage block was way easier than what we were trying to do initially... 🤦 

Let's just check whether `-p` needs clarification too. Are directories accepted as `targets` when `-p` is used? Please test and if so, update that option to clarify @imhardikj 

Thanks

content/docs/command-reference/repro.md
imhardikj: I checked `-p`, it doesn't work with directories. So the current is correct I think.
> submit a matching change in the core dvc repo

Sure I'll do this. Currently this is displayed for `targets`: `Stages to reproduce. 'dvc.yaml' by default.`
This should be updated to exact same text as above right: `Stage or path to dvc.yaml or .dvc file to reproduce. Using -R, directories to search for stages can also be given.`
Also this (https://github.com/iterative/dvc/blob/master/dvc/command/repro.py) is the file that needs to updated?

content/docs/command-reference/repro.md
imhardikj: Done. 
> we don't use <abbr> in Options

I forgot this. Thanks.

content/docs/command-reference/repro.md
jorgeorpinel: Yes, https://github.com/iterative/dvc/blob/master/dvc/command/repro.py#L70-L73

jorgeorpinel (approved):

jorgeorpinel:
Woot woot 🎉

PR #1862, opened 2020-10-14
Update twitter link from @andronovhopf to @DrElleOBrien

soerface:
Just updated a broken twitter link


shcheklein (approved):
thanks @soerface !! 🙏 good catch!

PR #1863, opened 2020-10-14
Add events to community page

elleobrien:
Note- the thumbnails for the events look pixelated. Is there any way to shrink the fields for image displays? I couldn't find high-res images of the event logos.


shcheklein:
hmm ... I don't see the events for some reason ...

dmpetrov:
> hmm ... I don't see the events for some reason ...

the same

elleobrien:
@dmpetrov hm that's weird, I can see it when I run a local development server. The file has definitely been changed (see file edits) to add the events. Can we get a new deployment?

shcheklein:
@rogermparent could you please take a look when you have a minute why doesn't it pick the changes up?

rogermparent:
Sounds like a caching issue, let me see what I can do.

rogermparent:
The deployment has events now! I've changed the community page logic to an implementation that hopefully doesn't run into the same caching issues the current one does with our setup.

shcheklein:
> Note- the thumbnails for the events look pixelated. Is there any way to shrink the fields for image displays?

minor, I'm merging this. But we can add some empty black box around to make it effectively larger.

PR #1864, opened 2020-10-14
start: add meta description, 'machine learning' term (index) [SEO]

jeremydesroches:
**Meta description**
Added a custom meta description to the front matter. This doc commonly appears in results next to the Tutorial (shown below in results for "dvc tutorial"). In addition to normal benefits, a custom description will help users differentiate between the two (hopefully reducing bounces for both).

**Key term**
Added "machine learning" in two usages: one as "machine learning model" and another as "git for machine learning".

**Typos**
Fixed two typos that I came across incidentally.

image

### DIFF
--- a/content/docs/start/index.md
+++ b/content/docs/start/index.md
@@ -1,7 +1,7 @@
 ---
-description: 'A step-by-step introduction: Learn how to use DVC to organize,
-version, and efficiently reproduce your data science or machine learning
-project.'
+description: 'Learn how you can use DVC to manage data science and machine
+learning projects: version data, manage data access, build data pipelines, and
+capture experiments.'
 ---
 
 # Get Started
@@ -53,14 +53,14 @@ DVC features can be grouped into layers. We'll explore them one by one in the
 next few sections:
 
 - [**Data versioning**](/doc/start/data-versioning) is the core part of DVC for
-  large files, datasets, ML models versioning and efficient sharing. We'll show
-  how to use a regular Git workflow, without storing large files with Git. Think
-  "Git for data".
+  large files, datasets, machine learning models versioning and efficient
+  sharing. We'll show how to use a regular Git workflow, without storing large
+  files with Git. Think "Git for data".
 
 - [**Data access**](/doc/start/data-access) shows how to use data artifacts from
   outside of the project and how to import data artifacts from another DVC
-  project. This can help to download a specific version of a machine learning
-  model to a deployment server or import a model to another project.
+  project. This can help to download a specific version of an ML model to a
+  deployment server or import a model to another project.
 
 - [**Data pipelines**](/doc/start/data-pipelines) describe how models and other
   data artifacts are built, and provide an efficient way to reproduce them.
--- a/content/docs/start/index.md
+++ b/content/docs/start/index.md
@@ -1,3 +1,9 @@
+---
+description: 'A step-by-step introduction: Learn how to use DVC to organize,
+version, and efficiently reproduce your data science or machine learning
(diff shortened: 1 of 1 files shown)


content/docs/start/index.md
```diff
@@ -1,3 +1,9 @@
+---
+description: 'A step-by-step introduction: Learn how to use DVC to organize,
+version, and efficiently reproduce your data science or machine learning
+project.'
```
jorgeorpinel, line 4: Not bad but sounds a lot like the desc of the Versioning tutorial. Should we try to make them clearly different? The GS is basically a primer on all of DVC's feature layers (there's 3-4) and the Versioning tutorial is about that one layer specifically (so, closer to the Data Versioning page of the Get Started, but deeper and with a real repo example).

jorgeorpinel: In fact notice how the current autogenerated desc Google shows comes from the first bullet in the doc, which refers to Data Versioning only. Let's try to include all 4 bullets in the SEO desc. please, unless you have a better idea.

content/docs/start/index.md
```diff
@@ -53,12 +59,13 @@ next few sections:
 
 - [**Data access**](/doc/start/data-access) shows how to use data artifacts from
   outside of the project and how to import data artifacts from another DVC
-  project. This can help to download a specific version of an ML model to a
-  deployment server or import a model to another project.
+  project. This can help to download a specific version of a machine learning
+  model to a deployment server or import a model to another project.
```
jorgeorpinel, line 63: Why don't we change this in the previous bullet instead, which is the first instance of "ML"?

content/docs/start/index.md
jeremydesroches: OK, good idea. Reworked this to include all 4 bullets.

content/docs/start/index.md
jeremydesroches: OK, moved full term to previous bullet and reverted this.

content/docs/start/index.md
```diff
@@ -1,3 +1,9 @@
+---
+description: 'Learn how you can use DVC to manage data science and machine
+learning projects: version data, manage data access, build data pipelines, and
+capture experiments.'
```
jorgeorpinel, line 4: Slight edit:
```suggestion
learning projects: version data, access it anywhere, capture data pipelines, and
manage experiments.'
```

content/docs/start/index.md
jorgeorpinel: Thanks

jorgeorpinel (approved):
Thanks @jeremydesroches ! I'll commit my suggestion above and merge this.

jeremydesroches:

**Get Started (index)** - [Link](https://search.google.com/u/0/search-console/performance/search-analytics?resource_id=sc-domain%3Advc.org&page=!https%3A%2F%2Fdvc.org%2Fdoc%2Fstart&breakdown=query&metrics=CLICKS%2CPOSITION&start_date=20201020&end_date=20201201&compare_start_date=20200907&compare_end_date=20201019) - 20% less clicks, 4% less impressions, and clickthrough down by 13%. Average position improved from 5.3 to 4.1. However, the comparison period had two weeks of strong performance prior to Sept. 19 (which accounts for most of the difference and did not continue), and there is an uptrend in all metrics up to the current date. So not bad, actually. 👍

jorgeorpinel:
Interesting. I wonder what those 2-3 strong prior weeks were due to?

And maybe looking at results for the target terms inserted here can give us a better report of the results?

PR #1799, opened 2020-09-23
add example with python parameters file

aandrusenko:
Related to https://github.com/iterative/dvc/pull/4456

PR #1874, opened 2020-10-17
Regular updates

jorgeorpinel:
- Python params file copy edits/ code formatting
- Misc. copy edits

### DIFF
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -152,9 +152,13 @@ Consider this Python parameters file named `params.py`:
 # All standard variable types are supported.
 BOOL = True
 INT = 5
+FLOAT = 0.001
 STR = 'abc'
 DICT = {'a': 1, 'b': 2}
-# etc.
+LIST = [1, 2, 3]
+SET = {4, 5, 6}
+TUPLE = (10, 100)
+NONE = None
 
 # DVC can retrieve class constants and variables defined in __init__
 class TrainConfig:
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -24,8 +24,9 @@ dependencies: _parameters_. Parameters are defined using the the `-p`
 
 In contrast to a regular <abbr>dependency</abbr>, a parameter is not a file (or
 directory). Instead, it consists of a _parameter name_ (or key) to find inside a
-YAML, JSON, TOML, or Python _parameters file_. Multiple parameter dependencies
-can be specified from one or more parameters files.
+YAML, JSON, TOML, or [Python](#examples-python-parameters-file) _parameters
+file_. Multiple parameter dependencies can be specified from one or more
+parameters files.
 
 The default parameters file name is `params.yaml`. Parameters should be
 organized as a tree hierarchy inside, as DVC will locate param names by their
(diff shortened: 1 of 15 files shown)


content/docs/command-reference/params/index.md
```diff
@@ -145,51 +146,38 @@ $ dvc run -n train -d logs/ -o users.csv \
...
 ## Examples: Python parameters file
 
-Consider this parameters file in Python format, named `params.py`:
+Consider this Python parameters file named `params.py`:
 
 ```python
-IS_BOOL: bool = True
-CONST = 5
-
-# All standard variable types are supported
```
shcheklein, line 154: let's keep comments in this case serve as docs

content/docs/command-reference/params/index.md
```diff
@@ -145,51 +146,38 @@ $ dvc run -n train -d logs/ -o users.csv \
...
-FLOAT = 0.001
+BOOL = True
+INT = 5
 STR = 'abc'
-DICT = {
-    "a": 1,
-    "b": 2
-}
-LIST = [1, 2, 3]
```
shcheklein, line 161: what is the reason removing this?

content/docs/command-reference/params/index.md
```diff
@@ -145,51 +146,38 @@ $ dvc run -n train -d logs/ -o users.csv \
...
     EPOCHS = 70
 
     def __init__(self):
-        # TrainConfig.layers param will be 9
         self.layers = 5
-        self.layers = 9
-        # TrainConfig.foo will NOT be found because the complex expression
-        self.foo = 1 + 2
-        # TrainConfig.bar will NOT be found
-        bar = 1
```
shcheklein, line 179: same - why change this?

content/docs/command-reference/params/index.md
jorgeorpinel: I didn't mean to remove that comment, my bad. Restored now.

content/docs/command-reference/params/index.md
jorgeorpinel: Why do we have a full list of Python types? We already mentioned that "All standard variable types are supported". Probably users won't be learning Python on our site. Only `BOOL` and `INT` are used later in the example so I left those and a couple more to make the point.

> As mentioned in https://github.com/iterative/dvc.org/pull/1799#discussion_r504869926

content/docs/command-reference/params/index.md
jorgeorpinel: Went from 7 lines to 3, looks better IMO and has the same info.

If you refer to removing the `bar = 1` line and its comment above, again we're explaining how Python works there — the variable has local scope so it can't be accessed externally (unless I'm wrong).

content/docs/command-reference/params/index.md
jorgeorpinel: TBH it's not even ideal to have these notes inside the Python code example IMO but OK, we probably don't want to make a whole new section in the description above to explain all the rules around Python params files.

content/docs/command-reference/params/index.md
jorgeorpinel: TBH it's not even ideal to have these notes inside the Python code example IMO but OK, we probably don't want to make a whole new section in the description above to explain all the rules around Python params files.

content/docs/command-reference/params/index.md
shcheklein: I think it was very explicit and easy to see all the posibilities at once, when you move something into text you make it harder to grasp the idea ... it's not a strong preference, tbh, I think it was good enough we spent some time on it, it's not a very core functionality

content/docs/command-reference/params/index.md
shcheklein: I think in this case all of this was beneficial, you see at once pretty much different example and see what might work what not

content/docs/user-guide/basic-concepts/data-artifact.md
```diff
@@ -6,3 +6,5 @@ match: ['data artifact', 'data artifacts', 'artifact', 'artifacts']
 Any data file or directory, as well as intermediate or final result that is
 tracked by DVC, for example by using `dvc add` or `dvc run`. See
 [Versioning Data and Models](/doc/use-cases/versioning-data-and-model-files).
+
+A.k.a. _output_
```
shcheklein, line 10: tbh, not sure when data artifact became a synonym  ... it's not exactly true ... e.g. we can depend on some data artifact that is not tracked by DVC whatsoever

content/docs/command-reference/params/index.md
shcheklein: >  whole new section in the description above to explain all the rules around Python params files.

that's what I would try to avoid... it's not scalable in this case, it's way easier to get all possibilities from the code to my mind

jorgeorpinel: Yeah. But shouldn't we have a section in https://dvc.org/doc/user-guide/dvc-files-and-directories to describe params files and the supported formats? We could have these notes there (and link from here).

content/docs/command-reference/params/index.md
jorgeorpinel: No strong opinion either. Reverted this here for now.

content/docs/command-reference/params/index.md
jorgeorpinel: OK. I added `bar` back, PTAL.

content/docs/user-guide/basic-concepts/data-artifact.md
jorgeorpinel: Good point. I've been meaning to address these redundant terms for a while in fact (see https://github.com/iterative/dvc.org/issues/608). OK, I remove the cross-ref for now.

shcheklein (approved):

PR #1881, opened 2020-10-20
Misc. updates

### LINKED ISSUE #1800: plots: update docs
after iterative/dvc#4602 we accept any target for `dvc plots`. Reflect that in docs.

- [x] UPDATE: pending one thing (from https://github.com/iterative/dvc.org/pull/1809#issuecomment-708029242):

> if we explain in the plots concept itself (plots ref. index) that plots are not only dvc.yaml based, people won't expect dvc.yaml to be present in the first place.
It's better to specify at the top level that plots files by definition could be anywhere.

jorgeorpinel:
- Try to make existing run-cache mentions less confusing
- Better explain `repro --pull` per https://github.com/iterative/dvc.org/pull/1841#pullrequestreview-508584134
- Review "control" term (esp. "under control")
- Mention that (2) `plots` subcommands work with any valid plots files in the system per https://github.com/iterative/dvc.org/pull/1809#issuecomment-708029242 (closes #1800)
- Update `dvc status -c/r` sample outputs (see https://github.com/iterative/dvc.org/pull/1877#pullrequestreview-513983803)


content/docs/command-reference/repro.md
```diff
@@ -157,8 +157,9 @@ up-to-date and only execute the final stage.
   corresponding pipelines, including the target stages themselves. This option
   has no effect if `targets` are not provided.
 
-- `--pull` - try automatically [pulling](/doc/command-reference/pull) missing
-  cache for outputs restored from run-cache.
+- `--pull` - [pulls](/doc/command-reference/pull) dependencies and outputs
+  involved in the stages being reproduced, if they are found in remote storage.
+  Note that it checks the run-cache too (available history of stage runs).
```
jorgeorpinel, line 162: A few questions about this still remain (see https://github.com/iterative/dvc.org/pull/1841#discussion_r508696705).

jorgeorpinel: Answers in https://github.com/iterative/dvc.org/pull/1841#discussion_r508804055! Resolving

content/docs/command-reference/plots/index.md
```diff
@@ -34,10 +34,10 @@ experiments. Usual plot examples are AUC curves, loss functions, confusion
...
 
 This type of metrics files are created by users, or generated by user data
-processing code, and get defined with the `-p` (`--plots`) and
-`--plots-no-cache`) options of `dvc run`. `dvc plots` subcommands can work with
-plots files committed to a Git repo history, data files controlled by DVC, or
-any other file in system.
+processing code, and can be defined in `dvc.yaml` stages (using the `--plots`
+and `--plots-no-cache` options if using `dvc run`). `dvc plots show` and
+`dvc plots diff` can work with any valid plots files in the system, whether
+tracked by Git or DVC, or not.
```
jorgeorpinel, line 40: Fixes #1800 @pared @shcheklein but note that since `plots modify` does require that plots are defined in dvc.yaml, I couldn't change this to be as general as suggested initially (summarized in https://github.com/iterative/dvc.org/pull/1809#issuecomment-708029242):

> if explain in the plots concept itself that plots are not only dvc.yaml based people won't expect dvc.yaml to be present in the first place.

pared: I like the shift from `get defined` to `can be defined`. This is on point what we actually did.

> can work with any valid plots files in the system

While it is something we need to do (as part of iterative/dvc#4446, currently we cannot `dvc plots show {target}` in no-repo case, as we will get error that `'.' is not a git repo`.

pared: Created #4761 to address this

jorgeorpinel: OK great, thanks for addressing that. In that case we can leave this text as-is, I think.

content/docs/command-reference/fetch.md
```diff
@@ -277,4 +277,4 @@ into the workspace (with `dvc repro train.dvc`).
 > Note that in this example project, the last stage file `evaluate.dvc` doesn't
 > add any more data files than those form previous stages, so at this point all
 > of the data for this pipeline is cached and `dvc status -c` would output
-> `Data and pipelines are up to date.`
+> `Cache and remote '{remote}' are in sync.`
```
shcheklein, line 280: we can potentially use `{name}` instead? instead of `remote {remote}`

content/docs/command-reference/plots/index.md
shcheklein: I think we are still overcomplicating this. In this doc the most important points to say that users can output certain type of metrics (arrays, etc) into JSON, YAML or whatnot and DVC provides a bunch of commands to deal with them - visualize, compare, etc, etc. We should def mention that it is similar to metrics, but for "continuous", etc, etc

Details that certain commands require `dvc.yaml` to be present can be hidden into those commands. Or there should be a separate section that starts with some _motivation_ for creating a `dvc.yaml`- what benefits on top of regular files people can get from using it.

jorgeorpinel: Well, I was following the original idea to emphasize that plots don't have to be tracked by DVC... But you're right, this was too low level for an index. I just removed a bunch of redundant info and left the basics + some refs to where the details are. Also updated show and diff a bit again. See 27519b4a. Resolving here

content/docs/command-reference/fetch.md
jorgeorpinel: I changed it to `Cache and remote 'myremote' are in sync.` which is what the message would actually look like with our generic remote name (used a lot in the add/modify examples). In a couple of instances I used `r1` to match the context. See 91c55fef

shcheklein (approved):

pared (approved):

PR #1887, opened 2020-10-26
glossary: remove "data artifact", et al.

jorgeorpinel:
- Remove "(data) artifact" from glossary, as it's a pretty general DS/ML term (not DVC-specific) — per #608
- Other misc. updates around ^
- \+ Small review of glossary term "import stage"


jorgeorpinel:
Sorry, I need to keep these coming because I fix many small things that are out of scope for my main work and they mount up over time.

content/docs/api-reference/read.md
```diff
@@ -79,8 +79,9 @@ These are loaded to memory directly (without using any disc space).
 
 ## Example: Load data from a DVC repository
 
-Any <abbr>data artifact</abbr> hosted online can be loaded directly in your
-Python code with this API. For example, let's say that you want to load and
+Any file tracked tracked in a <abbr>DVC project</abbr> (and
```
shcheklein, line 82: tracked tracked

shcheklein (approved):

content/docs/api-reference/read.md
jorgeorpinel: Yes because you track in Git the data tracked in DVC...

jorgeorpinel: Jk, fixed in dac21d59.

jorgeorpinel:
🎉

PR #1888, opened 2020-10-26
docs: /etc/dvc.config -> /etc/dvc/config

efiop:
Looks like an old leftover/misunderstanding.

Discord context https://discordapp.com/channels/485586884165107732/485596304961962003/770294056902852650


efiop:
The changes in this PR are correct when XDG_CONFIG_DIRS are not set (usually not set for root). We are using https://pypi.org/project/appdirs/ internally, which complies with xdg standard, and for regular users XDG_CONFIG_DIRS might be set (e.g. on ubuntu), so there might be some difference in behavior. Need to clarify that. Also worth improving on dvc side to make it more informative.

shcheklein (approved):

PR #1900, opened 2020-10-31
docs: add --no-exec param in 'import' command

BurnzZ:
In line with https://github.com/iterative/dvc/pull/4817.

### DIFF
--- a/content/docs/command-reference/import-url.md
+++ b/content/docs/command-reference/import-url.md
@@ -124,9 +124,10 @@ source.
   the imported data (`out`).
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exist it can be used to skip download.
-  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
-  update the `.dvc` files, and save existing data to the cache.
+  the file or directory already exists, this can be used to skip the download.
+  The data hash is not calculated by this since only the metadata is saved into
+  the `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in
+  the new `.dvc` file (_this also updates the cache_).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import-url.md
+++ b/content/docs/command-reference/import-url.md
@@ -125,9 +125,9 @@ source.
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
   the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated by this since only the metadata is saved into
-  the `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in
-  the new `.dvc` file (_this also updates the cache_).
+  The data hash is not calculated by this, only the metadata is saved into the
+  `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in the
+  new `.dvc` file and save existing data to the cache.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -11,7 +11,7 @@ the import.
 
 ```usage
 usage: dvc import [-h] [-q | -v]
-                  [-o <path>] [--file <filename>] [--rev <commit>]
+                  [-o <path>] [--file <filename>] [--rev <commit>] [--no-exec]
                   url path
 
 positional arguments:
@@ -102,6 +102,11 @@ repo at `url`) are not supported.
   > revision. This can impact the behavior of `dvc update` (see the
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
+
+- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
+  the file or directory already exist it can be used to skip download.
+  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
+  update the `.dvc` files, and save existing data to the cache.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -11,7 +11,8 @@ the import.
 
 ```usage
 usage: dvc import [-h] [-q | -v]
-                  [-o <path>] [--file <filename>] [--rev <commit>] [--no-exec]
+                  [-o <path>] [--file <filename>]
+                  [--rev <commit>] [--no-exec]
                   url path
 
 positional arguments:
@@ -104,9 +105,10 @@ repo at `url`) are not supported.
   > example below).
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exist it can be used to skip download.
-  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
-  update the `.dvc` files, and save existing data to the cache.
+  the file or directory already exists, this can be used to skip the download.
+  The data hash is not calculated by this since only the metadata is saved into
+  the `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in
+  the new `.dvc` file (_this also updates the cache_).
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -104,11 +104,11 @@ repo at `url`) are not supported.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
-- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated by this since only the metadata is saved into
-  the `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in
-  the new `.dvc` file (_this also updates the cache_).
+- `--no-exec` - create `.dvc` file without actually downloading the file or
+  directory. E.g. if the file or directory already exists, this can be used to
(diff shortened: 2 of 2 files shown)


content/docs/command-reference/import.md
```diff
@@ -103,6 +103,11 @@ repo at `url`) are not supported.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
+  the file or directory already exist it can be used to skip download.
+  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
```
shcheklein, line 108: @BurnzZ not sure about the `calculate URL` part, could you clarify what's happening here?

shcheklein (changes requested):
Looks great! Thanks @BurnzZ . Let @jorgeorpinel and me check and see if we need some minor adjustments here.

content/docs/command-reference/import.md
```diff
@@ -11,7 +11,7 @@ the import.
 
 ```usage
 usage: dvc import [-h] [-q | -v]
-                  [-o <path>] [--file <filename>] [--rev <commit>]
+                  [-o <path>] [--file <filename>] [--rev <commit>] [--no-exec]
                   url path
```
jorgeorpinel, line 15: That new line is too long. Let's reformat this way please:
```
usage: dvc import [-h] [-q | -v] [-o <path>] [--file <filename>]
                  [--rev <commit>] [--no-exec]
                  url path
```

content/docs/command-reference/import.md
jorgeorpinel: It's a copy/paste of the homologue option in https://dvc.org/doc/command-reference/import-url#options

jorgeorpinel: I.e. yes, that part about calculating hashes doesn't apply to DVC repo imports, we can just remove it as no other similar process occurs I think.

content/docs/command-reference/import.md
```diff
@@ -103,6 +103,11 @@ repo at `url`) are not supported.
   > [Importing and updating fixed revisions](#example-importing-and-updating-fixed-revisions)
   > example below).
 
+- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
+  the file or directory already exist it can be used to skip download.
+  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
+  update the `.dvc` files, and save existing data to the cache.
```
jorgeorpinel, line 109:
```suggestion
- `--no-exec` - create a `.dvc` file without actually downloading `url`. E.g. if
  the file or directory already exist, this option can be used to skip the download.
  `dvc commit <out>.dvc` should be used to update the import `.dvc` files, and save existing data to the cache.
```

jorgeorpinel: BTW we should apply similar copy edits to `import-url --no-exec` — I'll do it separately 🙂

content/docs/command-reference/import.md
dmpetrov: It should be `existS` and probably `THE download` or `downloadING`.

`E.g. if the file or directory already exist it can be used to skip download.` I'm not sure I understand this scenario? 😄 Do we run `dvc import --no-exec` on top of an existing directory - what is the goal? I thought, the primary use case is to postpone the download because we need a partial run (that is supported in `pull` but not `import`) OR the download needs to be run in another machine (we just cook metadata).

shcheklein: @jorgeorpinel to my mind `--no-exec` in the import-url has to be updated as well. `Calculate URL and data hash` part is not clear to me.

@dmpetrov it is the same as `import-url --no-exec`. Scenario is - create .dvc but don't download (if data exists already). E.g. you did `wget something` and now want to have a .dvc files to track it, but don't want to download data again.

shcheklein: `the download needs to be run in another machine (we just cook metadata).` - this scenario fits as well, I guess

`we just cook metadata` - is a good description I think.

content/docs/command-reference/import.md
shcheklein: Btw, it should be not `url` in this case I think. `W/o downloading data or file` or something like that.

content/docs/command-reference/import.md
BurnzZ: Hi everyone! Thanks for the feedback!

> It's a copy/paste of the same option in https://dvc.org/doc/command-reference/import-url#options.
@jorgeorpinel That's right! I've also updated the docs for `import-url` as per @shcheklein's reminder.

I've refined the `--no-exec` definition as per your feedback.

content/docs/command-reference/import.md
BurnzZ: I've moved it to a new line.

content/docs/command-reference/import.md
```diff
@@ -104,9 +105,10 @@ repo at `url`) are not supported.
   > example below).
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
```
shcheklein, line 107: I think `url` is not the best term here (mind the difference in the command names - import vs import-_url_)

content/docs/command-reference/import.md
```diff
@@ -104,9 +105,10 @@ repo at `url`) are not supported.
...
-  the file or directory already exist it can be used to skip download.
-  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
-  update the `.dvc` files, and save existing data to the cache.
+  the file or directory already exists, this can be used to skip the download.
+  The data hash is not calculated by this since only the metadata is saved into
```
shcheklein, line 109: @efiop when we import objects do we trust the external source for the hash? if not, does it makes sense to do this?  It seems to me that `--no-exec` at least in the case of `import` can write hash as well if the goal is to avoid data download. WDYT?

content/docs/command-reference/import-url.md
```diff
@@ -124,9 +124,10 @@ source.
   the imported data (`out`).
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exist it can be used to skip download.
-  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
-  update the `.dvc` files, and save existing data to the cache.
+  the file or directory already exists, this can be used to skip the download.
+  The data hash is not calculated by this since only the metadata is saved into
+  the `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in
+  the new `.dvc` file (_this also updates the cache_).
```
shcheklein, line 130: minor - `_this also updates the cache_` - @BurnzZ I think it's better to use `and save existing data to the cache` (for consistency- we don't use the cache update terminology in other places)

content/docs/command-reference/import-url.md
```diff
@@ -124,9 +124,10 @@ source.
   the imported data (`out`).
 
 - `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exist it can be used to skip download.
-  `dvc commit <out>.dvc` should be used to calculate the URL and data hash,
-  update the `.dvc` files, and save existing data to the cache.
+  the file or directory already exists, this can be used to skip the download.
+  The data hash is not calculated by this since only the metadata is saved into
```
shcheklein, line 128: no sure we need `... since ...`

content/docs/command-reference/import.md
BurnzZ: Right, that makes sense since `dvc import` handles repo locations. Would `repo url` fit much better here?

Moreover, if we do replace the `url` with something else, we'll also need to update the CLI parser (_like the screenshot below_) as well as the rest of the doc for consistency.

image

content/docs/command-reference/import-url.md
BurnzZ: Updated.

content/docs/command-reference/import-url.md
BurnzZ: Updated.

content/docs/command-reference/import.md
BurnzZ: Updated.

shcheklein (approved):
Thanks @BurnzZ ! To be honest, I still have a lot of questions to the option itself - name, its mechanics, etc, but that would be out of scope for the ticket.

content/docs/command-reference/import.md
efiop: It is more complicated than that, but yeah, we trust it indirectly by trusting the cache. This is because we don't really `download`, but rather `pull + checkout`. We don't fill it on `--no-exec` for the `run` symmetry, but yeah, could consider it, though that has itsown implications. Let's leave it as is for now, since it is not a high priority.

shcheklein: >  but rather pull + checkout

hmm, how do we do `pull` if hash is not set? I thought that we don't do anything with data at all - kinda create .dvc with some meta in it and that it.

if we do `pull` + `checkout` what happens if data already exists in workspace and is in cache, or exists and not in cache?

> Let's leave it as is for now, since it is not a high priority. 

PR is merged, yep.

jorgeorpinel: Pretty late to this conversation but just wanted to double check that with --no-exec there's no connection to the data source, right? Because I'm reviewing this text for an upcoming docs update. Thanks

jorgeorpinel:
I came up with this text for an upcoming regular update of docs:

> - `--no-exec` - create the import `.dvc` file without actually downloading the
  target file or directory. No connection to the source location is made, so the
  data hash is not calculated. Only the import operation is recorded in the
  `.dvc` file. Use `dvc commit` to save any existing dep/out files to the cache
  and record their hashes to the `.dvc` file.
>  This can be useful to skip the download if the file or directory already
  exists locally, or to just save the operation's metadata (to be imported at
  another time or location).

(Doing a similar change in the same option of `run`.)

PR #1901, opened 2020-10-31
Misc. updates

jorgeorpinel:
- Misc. copy edits e.g. per https://github.com/iterative/dvc.org/pull/1890#pullrequestreview-521891889
- Term reviews around "layers" and "components" (DVC features)
- Fixes to https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory


content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -332,16 +330,13 @@ $ tree .dvc/cache
...
-A typical `.dir` cache file looks like this:
-
-```dvc
-$ cat .dvc/cache/19/6a322c107c2572335158503c64bfba.dir
-[{"md5": "dff70c0392d7d386c39a23c64fcc0376", "relpath": "cat.jpeg"},
-{"md5": "29a6c8271c0c8fbf75d3b97aecee589f", "relpath": "index.jpeg"}]
+```json
+[{ "md5": "dff70c0392d7d386c39a23c64fcc0376", "relpath": "cat.jpeg" },
+{ "md5": "29a6c8271c0c8fbf75d3b97aecee589f", "relpath": "index.jpeg" }]
 ```
```
jorgeorpinel, line 340: Restyled changes this (see https://github.com/iterative/dvc.org/pull/1902#pullrequestreview-521161503). We can just ignore Restyled and merge this anyway right?

> It may be a autoformatted later in another PR though...

shcheklein: No, I don't think so. If it's important enough removing `json` prefix should help.

content/docs/user-guide/related-technologies.md
```diff
@@ -82,8 +82,8 @@ _Luigi_, etc.
 
 ## Experiment management software
 
-- DVC uses Git as the underlying layer for data, pipelines, an experiment
-  versioning, instead of a custom web application.
+- DVC integrates with Git, so it remains the underlying version control tool for
+  data, pipelines, an experiments, instead of reinventing the wheel.
```
shcheklein, line 86: this gives it a quite different meaning

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -332,16 +330,13 @@ $ tree .dvc/cache
...
-The cache file with `.dir` extension is a special text file that contains the
-mapping of files in the `data/` directory (as a JSON array), along with their
-hash values. The other two cache files are the files inside `data/`.
+This `.dir` file contains the mapping of files in `data/images` (as a JSON
+array), including their hash values. That's how DVC knows that the other two
+cached files belong in the directory:
 
-A typical `.dir` cache file looks like this:
-
-```dvc
```
shcheklein, line 341: No reason to change this

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -308,21 +317,10 @@ data/images/
 $ dvc add data/images
 ```
 
-When running `dvc add` on this directory of images, a `data/images.dvc`
```
shcheklein, line 311: why this change? it looks more or less good for me

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -282,22 +282,31 @@ Full <abbr>parameters</abbr> (key and value) are listed separately under
 
 ## Structure of the cache directory
 
+The DVC cache is a
+[content-addressable storage](https://en.wikipedia.org/wiki/Content-addressable_storage)
+(CAS), which adds a layer of indirection between code and data. It's hidden from
+the user (and from Git), and typically local. However, it can be synchronized
```
shcheklein, line 288: sounds like it's a typical scenario to have _only_ local cache and remote is kinda optional

content/docs/start/index.md
```diff
@@ -49,13 +49,13 @@ Changes to be committed:
...
-DVC features can be grouped into layers. We'll explore them one by one in the
-next few sections:
+DVC features can be grouped into functional components. We'll explore them one
+by one in the next few sections:
 
-- [**Data versioning**](/doc/start/data-versioning) is the core part of DVC for
-  large files, datasets, machine learning models versioning and efficient
-  sharing. We'll show how to use a regular Git workflow, without storing large
-  files with Git. Think "Git for data".
+- [**Data versioning**](/doc/start/data-versioning) is the base layer of DVC. It
```
shcheklein, line 55: Again, I don't understand the reason for this change. Not obvious, sounds repetitive.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Hmmm rather keep `json` so there's highlighting. OK I'll let it format the block.

shcheklein: But it can be misleading then ....

shcheklein: We can keep the command that we had `cat ...`

content/docs/user-guide/related-technologies.md
jorgeorpinel: Yes for 2 reasons.
1. A correction: DVC doesn't use Git, it integrates with Git.
2. "instead of a custom web application"? It's not clear why we're talking about web apps here IMO. I guess in reference to MLFlow's interface? Any other examples? Maybe we should mention them directly like we mention git-lfs for ex.
  In any case, "uses Git for versioning instead of a web app" still doesn't make much sense. Are web apps used for versioning? But maybe I'm missing something or not understanding the original intention of this text.

jorgeorpinel: For now I'm simplifying it to:
```md
- DVC integrates with Git as the underlying version control tool for data,
  pipelines, an experiments.
```

shcheklein: yes, I was referring to the `instead of reinventing the wheel.` part mostly (integrates or uses - not that important, uses sounds friendlier though)


`a custom web application` is not the best either, agreed. What it was trying to say that `DVC saves info into Git vs some external app or external database or external API`

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: My main purpose here is to insert the CAS and indirection concepts.

I removed the notes about how it's hidden, local and about remote storage.

jorgeorpinel: It's a bit tricky because the definition of the cache is spread in a few places... Here, in the `cache` cmd index, and in the tooltip...

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I removed this chunk because it was explaining `dvc add`/`.dvc` files, not the cache structure.

I can put it back if you prefer.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: OK, I reverted to `$ cat` but I think it looks better with the JSON formatting.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: The sentence was unnecessary as it's already mentioned in the previous paragraph. The code block I have rolled back now.

content/docs/start/index.md
jorgeorpinel: The main change here is "core part" -> "base layer". I was reviewing the usage of terms around "feature". We used "layers" before but the concept of "components" (Composable principle) is more appropriate (this came from reading Hashicorp's Tao).

jorgeorpinel: It was more repetitive before: "Data versioning is the core part of DVC for large file versioning..."

But let me re-work this to fix the language with a minimal change... ⌛

jorgeorpinel: OK changed it again. I separated the part about "efficient sharing" because it was making the first sentence too complicated. PTAL

> p.s. I think the repetitiveness was inserted recently by a SEO PR now that I think about it... Jeremy wanted to boost the term "ML model versioning". I guess it's too forced to have that in this doc.

content/docs/user-guide/related-technologies.md
jorgeorpinel: OK changed back to "uses" although it's slightly incorrect... But you're right, easier to read/ friendlier. I also updated the other part of the explanation. PTAL

content/docs/use-cases/versioned-storage.md
```diff
@@ -0,0 +1,13 @@
+# Versioned storage
```
shcheklein, line 1: ??

shcheklein (approved):
looks good, except some extra doc

content/docs/use-cases/versioned-storage.md
jorgeorpinel: Oops. Doc removed!

PR #1903, opened 2020-11-01
cmd: some example updates as per 1.x

imhardikj:
- [x] `fetch` example
- [x] `get` example
- [x] `pull` example

Partially addresses #1824

### DIFF
--- a/content/docs/command-reference/fetch.md
+++ b/content/docs/command-reference/fetch.md
@@ -163,8 +163,9 @@ $ tree .dvc/cache
 .dvc/cache
 ├── 20
 │   └── b786b6e6f80e2b3fcf17827ad18597.dir
-├── 32
-│   └── b715ef0d71ff4c9e61f55b09c15e75
+├── c8
+│    ├── d307aa005d6974a8525550956d5fb3
+│    └── ...
 ...
 ```
 
@@ -180,7 +181,7 @@ Note that the `.dvc/cache` directory was created and populated.
 Used without arguments (as above), `dvc fetch` downloads all files and
 directories needed by all `dvc.yaml` and `.dvc` files in the current branch. For
 example, the hash values `20b786b...` and `c8d307a...` correspond to the
-`model.pkl` file and `data/features/` directory, respectively.
+`data/features/` directory and `model.pkl` file, respectively.
 
 Let's now link files from the cache to the workspace with:
 
@@ -240,8 +241,8 @@ $ dvc status -c
 ```
 
 One could do a simple `dvc fetch` to get all the data, but what if you only want
-to retrieve the data up to our third stage `train`? We can use the `--with-deps`
-(or `-d`) option:
+to retrieve the data up to our third stage, `train`? We can use the
+`--with-deps` (or `-d`) option:
 
 ```dvc
 $ dvc fetch --with-deps train
@@ -250,6 +251,9 @@ $ tree .dvc/cache
 .dvc/cache
 ├── 20
 │   └── b786b6e6f80e2b3fcf17827ad18597.dir
+├── c8
+│   ├── 43577f9da31eab5ddd3a2cf1465f9b
+│   └── d307aa005d6974a8525550956d5fb3
 ├── 32
 │   └── b715ef0d71ff4c9e61f55b09c15e75
 ├── 54
@@ -258,11 +262,8 @@ $ tree .dvc/cache
 │   └── 597d341ceb7d8fbbe88859a892ef81
 ├── a1
 │   └── 414b22382ffbb76a153ab1f0d69241.dir
-├── a3
-│   └── 04afb96060aad90176268345e10355
-└── c8
-    ├── 43577f9da31eab5ddd3a2cf1465f9b
-    └── d307aa005d6974a8525550956d5fb3
+└── a3
+    └── 04afb96060aad90176268345e10355
 ```
 
 Fetching using `--with-deps` starts with the target stage (`train`) and searches
--- a/content/docs/command-reference/fetch.md
+++ b/content/docs/command-reference/fetch.md
@@ -161,10 +161,10 @@ $ dvc fetch
 
 $ tree .dvc/cache
 .dvc/cache
-├── 38
-│   └── 63d0e317dee0a55c4e59d2ec0eef33
-├── 42
-│   └── c7025fc0edeb174069280d17add2d4.dir
+├── 20
+│   └── b786b6e6f80e2b3fcf17827ad18597.dir
+├── 32
+│   └── b715ef0d71ff4c9e61f55b09c15e75
 ...
 ```
 
@@ -234,10 +234,9 @@ dependencies/outputs of other pipeline stages are still missing from the cache:
 ```dvc
 $ dvc status -c
 ...
-    deleted:            model.pkl
     deleted:            data/features/test.pkl
     deleted:            data/features/train.pkl
-    deleted:            data/data.xml
+    deleted:            model.pkl
 ```
 
 One could do a simple `dvc fetch` to get all the data, but what if you only want
@@ -249,22 +248,21 @@ $ dvc fetch --with-deps train
 
 $ tree .dvc/cache
 .dvc/cache
-├── 38
-│   └── 63d0e317dee0a55c4e59d2ec0eef33
-├── 42
-│   └── c7025fc0edeb174069280d17add2d4.dir
-├── 58
-│   └── 245acfdc65b519c44e37f7cce12931
-├── 68
-│   └── 36f797f3924fb46fcfd6b9f6aa6416.dir
-├── 9d
-│   └── 603888ec04a6e75a560df8678317fb
+├── 20
+│   └── b786b6e6f80e2b3fcf17827ad18597.dir
+├── 32
+│   └── b715ef0d71ff4c9e61f55b09c15e75
+├── 54
+│   └── c0f3ef1f379563e0b9ba4accae6807
+├── 6f
+│   └── 597d341ceb7d8fbbe88859a892ef81
+├── a1
+│   └── 414b22382ffbb76a153ab1f0d69241.dir
 ├── a3
-│   └── 04afb96060aad90176268345e10355
-├── aa
-│   └── 35101ce881d04b41d5b4ff3593b423
-└── dc
-    └── a9c512fda11293cfee7617b66648dc
+│   └── 04afb96060aad90176268345e10355
+└── c8
+    ├── 43577f9da31eab5ddd3a2cf1465f9b
+    └── d307aa005d6974a8525550956d5fb3
 ```
 
 Fetching using `--with-deps` starts with the target stage (`train`) and searches
@@ -277,4 +275,4 @@ files needed to reproduce this pipeline up to the third stage into the workspace
 > Note that in this example project, the last stage `evaluate` doesn't add any
 > more data files than those form previous stages, so at this point all of the
 > data for this pipeline is cached and `dvc status -c` would output
-> `Cache and remote 'myremote' are in sync.`
+> `Cache and remote 'storage' are in sync.`
--- a/content/docs/command-reference/fetch.md
+++ b/content/docs/command-reference/fetch.md
@@ -228,7 +228,7 @@ the cache entries for the `data/features` directory, and
 ## Example: With dependencies
 
 After following the previous example (**Specific stages**), only the files
-associated with the `prepare` stage have been fetched. Several
+associated with the `prepare` stage has been fetched. Several
 dependencies/outputs of other pipeline stages are still missing from the cache:
 
 ```dvc
@@ -241,7 +241,7 @@ $ dvc status -c
 ```
 
 One could do a simple `dvc fetch` to get all the data, but what if you only want
-to retrieve the data up to our third stage, `train`? We can use the
+to retrieve the data up to our third stage (`train`)? We can use the
 `--with-deps` (or `-d`) option:
 
 ```dvc
--- a/content/docs/command-reference/fetch.md
+++ b/content/docs/command-reference/fetch.md
@@ -179,8 +179,8 @@ Note that the `.dvc/cache` directory was created and populated.
 
 Used without arguments (as above), `dvc fetch` downloads all files and
 directories needed by all `dvc.yaml` and `.dvc` files in the current branch. For
-example, the hash values `3863d0e...` and `42c7025...` correspond to the
-`model.pkl` file and `data/features/` directory, respectively.
+example, the hash value `20b786b...` corresponds to the `data/prepared/`
+directory.
 
 Let's now link files from the cache to the workspace with:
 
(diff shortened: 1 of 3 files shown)


jorgeorpinel (changes requested):
Revieweing fetch only for now:

@imhardikj the updates so far seem correct but we are missing to read the whole section from the beginning and make sure it still makes sense. Specifically, the **Click and expand to setup the project** expandable section already runs `dvc fetch` so after doing that, many or all of the examples are repetitive or confusing (since everything is already pre-fetch. We shuold not have that fetch there and make sure subsequent examples are progressive in case someone follows them one by one, so the reader can notice a change when trying each one. Or add a note/instruction to delete the cache before or after each example.

Please apply the same principle to the other refs. too before I review them. Thanks

jorgeorpinel:
p.s. and actually run every command and update the output of `tree` if you have to, even if it's just a bunch of md5 hashes. Thanks

imhardikj:
> Click and expand to setup the project expandable section already runs dvc fetch

@jorgeorpinel `dvc fetch` is actually after the expandable section.
Also in 2nd example there's already a note to delete cache.
For 3rd example its's not required since it's continuation of second.

In 2nd example of `pull` also there's already a  note to delete cache. It isn't require for the 3rd one.

`Get` doesn't require a note.

jorgeorpinel:
I see, seems like I was confused then. OK let me review...

content/docs/command-reference/fetch.md
```diff
@@ -161,10 +161,10 @@ $ dvc fetch
...
 $ tree .dvc/cache
 .dvc/cache
-├── 38
-│   └── 63d0e317dee0a55c4e59d2ec0eef33
-├── 42
-│   └── c7025fc0edeb174069280d17add2d4.dir
+├── 20
+│   └── b786b6e6f80e2b3fcf17827ad18597.dir
+├── 32
+│   └── b715ef0d71ff4c9e61f55b09c15e75
```
jorgeorpinel, line 167: OK but make sure all the mentions of these values match. There are some in line 182, for example.

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
 ## Example: With dependencies
 
 After following the previous example (**Specific stages**), only the files
-associated with the `prepare.dvc` stage file have been fetched. Several
+associated with the `prepare` stage has been fetched. Several
```
jorgeorpinel, line 231: have

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
...
 dependencies/outputs of other pipeline stages are still missing from the cache:
 
 ```dvc
 $ dvc status -c
 ...
-    deleted:            model.pkl
     deleted:            data/features/test.pkl
     deleted:            data/features/train.pkl
-    deleted:            data/data.xml
```
jorgeorpinel, line 240: Are you sure `data/data.xml` isn't printed here?

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
...
+    deleted:            model.pkl
 ```
 
 One could do a simple `dvc fetch` to get all the data, but what if you only want
-to retrieve the data up to our third stage, `train.dvc`? We can use the
+to retrieve the data up to our third stage (`train`)? We can use the
```
jorgeorpinel, line 243: OK to remove ".dvc" but why did we change comma for parentheses?

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
...
+backwards through its pipeline for data to download into the project's cache.
+All the data for the second and third stages (`featurize` and `train`) has now
+been downloaded to the cache. We could now use `dvc checkout` to get the data
+files needed to reproduce this pipeline up to the third stage into the workspace
+(with `dvc repro train`).
+
+> Note that in this example project, the last stage `evaluate` doesn't add any
+> more data files than those form previous stages, so at this point all of the
+> data for this pipeline is cached and `dvc status -c` would output
+> `Cache and remote 'storage' are in sync.`
```
jorgeorpinel, line 278: > storage

Good catch! This is why we should always actually run all the commands when reviewing examples 🙂

content/docs/command-reference/fetch.md
imhardikj: `32b71..` is not related to any file present in workspace, not there in `dvc.lock` either. So I have kept only `20b7...` in line 182.

content/docs/command-reference/fetch.md
imhardikj: Yes. I double checked this. Initially I thought it isn't printed because `data.xml` is already because of first example. But trying this example itself from starting gave same output.

content/docs/command-reference/fetch.md
imhardikj: Changed.

content/docs/command-reference/fetch.md
imhardikj: Thanks.

content/docs/command-reference/fetch.md
jorgeorpinel: Hmm so why is `32b715e in the cache above? It's supposed to be the `model.pkl` file. Please don't just remove the other file from line 182, both should be mentioned to explain the cache tree above. Just update the correct hash values...

content/docs/command-reference/fetch.md
```diff
@@ -179,8 +179,8 @@ Note that the `.dvc/cache` directory was created and populated.
 
 Used without arguments (as above), `dvc fetch` downloads all files and
 directories needed by all `dvc.yaml` and `.dvc` files in the current branch. For
-example, the hash values `3863d0e...` and `42c7025...` correspond to the
-`model.pkl` file and `data/features/` directory, respectively.
+example, the hash value `20b786b...` corresponds to the `data/prepared/`
+directory.
```
jorgeorpinel, line 183: Both hash values should be kept (for the model file and for the data/prep dir)

content/docs/command-reference/fetch.md
jorgeorpinel: > it isn't printed because data.xml is already because of first example

Makes sense.

content/docs/command-reference/get.md
```diff
@@ -177,11 +177,11 @@ The `model.monograms.pkl` file now contains the older version of the model. To
 get the most recent one, we use a similar command, but with
 `-o model.bigrams.pkl` and `--rev bigrams-experiment` (or even without `--rev`
 since that tag has the latest model version anyway). In fact, in this case using
-`dvc pull` with the corresponding `.dvc` files should suffice, downloading the
+`dvc pull` with the corresponding stage `target` should suffice, downloading the
```
jorgeorpinel, line 180:
```suggestion
`dvc pull` with the corresponding stage as target should suffice, downloading the
```
May need formatting

content/docs/command-reference/fetch.md
imhardikj: Hash value for `model.pkl` is `c8d30..`. Since it wasn't visible in `tree` displayed above so I removed it. Have updated now.

content/docs/command-reference/fetch.md
```diff
@@ -179,7 +179,7 @@ Note that the `.dvc/cache` directory was created and populated.
 
 Used without arguments (as above), `dvc fetch` downloads all files and
 directories needed by all `dvc.yaml` and `.dvc` files in the current branch. For
-example, the hash values `3863d0e...` and `42c7025...` correspond to the
+example, the hash values `20b786b...` and `c8d307a...` correspond to the
```
jorgeorpinel, line 182: Where is this `c8d307a` coming from? Why doesn't it match the cache tree above? It should be `32b715e`

jorgeorpinel: according to the cache tree shown above.

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
...
     deleted:            data/features/train.pkl
-    deleted:            data/data.xml
+    deleted:            model.pkl
 ```
 
 One could do a simple `dvc fetch` to get all the data, but what if you only want
-to retrieve the data up to our third stage, `train.dvc`? We can use the
-`--with-deps` (or `-d`) option:
+to retrieve the data up to our third stage `train`? We can use the `--with-deps`
+(or `-d`) option:
```
jorgeorpinel, line 244: Why did you remove the comma?

content/docs/command-reference/fetch.md
jorgeorpinel: Now it's missing the comma... Please let's not overcomplicate these simple replacing tasks unless there's a strong reason to change existing syntax. 3 rounds of feedback on a single replace shouldn't be needed. Thanks

content/docs/command-reference/pull.md
```diff
@@ -168,14 +171,13 @@ example-get-started/
 │   ├── data.xml
 │   ├── data.xml.dvc
 ...
-├── model.pkl
-└── train.dvc
+└── model.pkl
```
jorgeorpinel, line 174: Did you try these commands? `$ tree example-get-started/` doesn't seem to be correct here if we're already inside that dir.

content/docs/command-reference/fetch.md
imhardikj: `tree` in line 163 above is not displayed full. `c8d307a..` is at bottom of tree (complete `tree` is displayed on line 251).
Hash value of `model.pkl` can be confirmed from [lock file](https://github.com/iterative/example-get-started/blob/master/dvc.lock). 
I change put `c8d307a` in the `tree` in line 163 instead of `3863d0e`, but that will we wrong order(tree is printed in different order).

jorgeorpinel: > tree in line 163 above is not displayed full

O I get that but the point of displaying those 2 cache entries originally is that they would match these 2 files in this paragraph. So just change the entries displayed in that tree so they match please.

content/docs/command-reference/fetch.md
```diff
@@ -228,53 +228,51 @@ the cache entries for the `data/features` directory, and
...
 ├── a3
-│   └── 04afb96060aad90176268345e10355
-├── aa
-│   └── 35101ce881d04b41d5b4ff3593b423
-└── dc
-    └── a9c512fda11293cfee7617b66648dc
+│   └── 04afb96060aad90176268345e10355
+└── c8
+    ├── 43577f9da31eab5ddd3a2cf1465f9b
+    └── d307aa005d6974a8525550956d5fb3
```
jorgeorpinel, line 265: > complete tree is displayed on line 251

BTW, do we need to display the whole thing? I'm not looking at the whole context, but please decide whether it's useful to have the full cache tree here or if it can be shortened and why. Thanks

imhardikj: I think the complete tree is displayed here because in previous example about prepare stage, the tree displayed only 3 values. But in this  example we are fetching remaining stages. So it shows that more files are downloaded and hence cache contents increased. We can display it like this also:
```
.dvc/cache
...
├── a1
│   └── 414b22382ffbb76a153ab1f0d69241.dir
├── a3
│   └── 04afb96060aad90176268345e10355
└── c8
    ├── 43577f9da31eab5ddd3a2cf1465f9b
    └── d307aa005d6974a8525550956d5fb3
```
Only the last 3 files.

content/docs/command-reference/get.md
imhardikj: Done.

content/docs/command-reference/fetch.md
imhardikj: Updates.

content/docs/command-reference/pull.md
imhardikj: Yes this wouldn't actually work. It should only be `$ tree`. Thanks.

content/docs/command-reference/fetch.md
jorgeorpinel: Thanks

content/docs/command-reference/get.md
```diff
@@ -129,12 +129,12 @@ stored:
...
 $ dvc get --show-url \
           https://github.com/iterative/example-get-started model.pkl
-https://remote.dvc.org/get-started/66/2eb7f64216d9c2c1088d0a5e2c6951
+https://remote.dvc.org/get-started/c8/d307aa005d6974a8525550956d5fb3
 ```
 
 `remote.dvc.org/get-started` is an HTTP
 [DVC remote](/doc/command-reference/remote), whereas
-`662eb7f64216d9c2c1088d0a5e2c6951` is the file hash.
+`c8d307aa005d6974a8525550956d5fb3` is the file hash.
```
jorgeorpinel, line 137: Let's use short SHAs in paragraphs 🙂 
```suggestion
`c8d307...` is the file hash.
```

jorgeorpinel: Committing since this file is ready otherwise.

content/docs/command-reference/pull.md
jorgeorpinel: Cool

content/docs/command-reference/pull.md
```diff
@@ -162,20 +165,19 @@ We can now just run `dvc pull` to download the most recent `data/data.xml`,
 ```dvc
 $ dvc pull
 
-$ tree example-get-started/
+$ tree
 example-get-started/
```
jorgeorpinel, line 169: Wouldn't it output this though?
```suggestion
$ tree
.
```

content/docs/command-reference/fetch.md
jorgeorpinel: Yeah, looks better but why only those last 3 though? Are those the only values actually used throughout the examples? If so, I agree.

content/docs/command-reference/pull.md
imhardikj: Yeah this will be the output. Thanks

content/docs/command-reference/fetch.md
imhardikj: Only `c8....` and one more file `20b7...` is used in first example.
I think it's better to display whole tree as the purpose of the tree was to display that more files are downloaded and hence cache contents are increased.
Else we can display these two files above and one more random file like this:
```
....
├── a1
│   └── 414b22382ffbb76a153ab1f0d69241.dir
├── 20
│   └── b786b6e6f80e2b3fcf17827ad18597.dir
└── c8
    ├── 43577f9da31eab5ddd3a2cf1465f9b
    └── d307aa005d6974a8525550956d5fb3
```

jorgeorpinel: OK

jorgeorpinel (changes requested):
LGTM. Finally getting this done! Thanks to @imhardikj 👍

jorgeorpinel (approved):
Meant to approve 🙄

PR #1911, opened 2020-11-05
Remove CSS line break on em in p following highlights

### LINKED ISSUE #1910: formatting for italic text is broken
https://dvc.org/doc/command-reference/remote/modify#available-parameters-for-all-remotes

`Or a local remote (a directory in the file system):`

got formatted into:

Screen Shot 2020-11-05 at 9 30 26 AM

rogermparent:
This fixes #1910, but I'm not sure what this rule was created for so it may have caused an issue with something else. That said, we have a selector specifically for images so we can be sure that isn't the case.

### DIFF
--- a/src/components/Documentation/Markdown/styles.module.css
+++ b/src/components/Documentation/Markdown/styles.module.css
@@ -44,8 +44,7 @@
     }
 
     img + em,
-    :global(.gatsby-resp-image-wrapper) + em,
-    .gatsby-highlight + p > em:only-child {
+    :global(.gatsby-resp-image-wrapper) + em {
       color: #6a737d;
       font-size: 0.9em;
       display: block;


rogermparent:
The line I removed [seems quite old](https://github.com/iterative/dvc.org/commit/0c9f49df5ce2a061a539f3d6d1e52148dc42f3a6#diff-e31c2196d724cf7a0b54b9eec9e49ea1afbca6941b9a14d73209d4e6e081af38), and doesn't have a particular purpose attributed to it, so I think it's safe to keep this change and be vigilant about things that may have broken.

Existing content that would be affected are special elements in the docs that need italics line-broken after them. Maybe things like code examples or code blocks. If any of these are broken, I can try adding a rule that more specifically targets those cases or adapt the source formatting to something more distinguishable from regular italics.

shcheklein:
thanks @rogermparent !

PR #1913, opened 2020-11-08
How to doc - Adding dependency to stage

### LINKED ISSUE #460: cmd ref: improve `run` and `commit` / how to add outs/deps without re-running stage?

imhardikj:
Fixes #460 
 Adding dependency to a stage without re-running it

Partially addresses #899 (How-to sub section)


imhardikj:
@jorgeorpinel I have merged both adding output and dependency in the same doc currently. Should this be a separate doc or both topics in one doc is correct?

jorgeorpinel:
@imhardikj should be the same doc.

But please merge my branch `how-to` to this and close #1914 first. I reordered the how-tos and made some quick copy edits.

content/docs/sidebar.json
```diff
@@ -100,7 +100,7 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
+          "add-dependency-or-output-to-stage",
```
jorgeorpinel, line 103: Once merged the slug should be add-deps-or-outs-to-a-stage and the doc title should be **Add Dependencies or Outputs to a Stage** but if that's too long for the nav label, let's use something else like "Add Deps/Outs to a Stage".

jorgeorpinel: \* Once merged with my changes

content/docs/user-guide/how-to/add-dependency-or-output-to-stage.md
```diff
@@ -0,0 +1,50 @@
+# Add Dependency or Output to Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the dependencies, or the output files/directories it creates, which
+are already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`deps`
```
jorgeorpinel, line 6: just adding "dependencies or" everywhere is not going to be enough. Please think the changes through and request my review then. Thanks @imhardikj

content/docs/sidebar.json
imhardikj: @jorgeorpinel I merged the branch `how-to`. If you check "file changed" in this PR, it shows 12 files changed. But here (https://github.com/iterative/dvc.org/compare/guide/how-to) it shows 6 files changed (in `guide/how-to` branch) . Is there some problem/conflict? Because the changes in those extra 5-6 files (shown under this PR) have already been merged to master. Please help.

content/docs/user-guide/how-to/add-dependency-or-output-to-stage.md
imhardikj: Pushed new update.

content/docs/sidebar.json
jorgeorpinel: Sorry, I don't know what you did in commits 6ecc645 and da629e9. You're gonna have to abandon this branch and just take over PR #1914. Please update the title and description and switch to that branch to apply again all of your changes here @imhardikj.

jorgeorpinel:
Closing in favor of #1914

p.s. @imhardikj please use descriptive messages in your commits. "Updates" is not informative. You can see my commits to get an idea. Try to keep commits short and focused. Thanks!

Issue #625, opened 2019-09-17
how to: "undo" 'dvc add' + other "how to undo mistakes" recipes


gwerbin:
*This issue is 1 of 3 that I was asked to post here, following a conversation on Discord with mroutis and ivan*

I encountered this problem within my first 15 minutes of using DVC: "how do I undo `dvc add`?"

The answer turns out to be:
1. `unprotect` the file, just in case
2. Delete its corresponding .dvc file
3. Optionally run `gc` or look up the MD5 in `data.dvc` and delete it from `.dvc/cache` by hand.

It would be nice if this recipe were included somewhere in the docs.

Part of the reason this procedure was so non-obvious is that I never thought to look in https://dvc.org/doc/commands-reference/add. This turns out to have a complete technical explanation of linking, and what "protect" means in DVC. This document should be referenced early and often, maybe with a sentence like "For a technical explanation of the caching and linking process, see [here](https://dvc.org/doc/commands-reference/add)".

In the meantime, hopefully my [Stackoverflow self-answer](https://stackoverflow.com/a/57967291) will help other users, as well as get the DVC ball rolling on Stackoverflow.

DVC, like Git, admits the possibility of [potentially-destructive beginner mistakes](https://github.com/iterative/dvc/issues/1524). I think it would be good practice to try and anticipate and provide recipes for undoing mistakes, and link to them **prominently** in the getting started documentation.

dashohoxha:
@gwerbin There is an example of how to undo `dvc add` here: https://katacoda.com/dvc/courses/basics/data-management
However this is still unfinished work and not yet advertised on the website of DVC.

dashohoxha:
> In the meantime, hopefully my Stackoverflow self-answer will help other users, as well as get the DVC ball rolling on Stackoverflow.

There is also a stackexchange about DS: https://datascience.stackexchange.com

Suor:
I feel that there should be an easier way)

shcheklein:
@Suor do you meant that there _is_ an easier way already, or rather that it's a good sign to implement one? like `dvc reset` or something?

Suor:
@shcheklein I don't see one, but there probably should be. I am surprised `dvc remove` does something arcane instead of this.

Suor:
BTW, this recipe is also dangerous considering how `dvc gc` operates.

shcheklein:
[...]

PR #1914, opened 2020-11-09
guide: reorder how-tos and copy edits; expand on adding existing dependencies to a stage

### LINKED ISSUE #460: cmd ref: improve `run` and `commit` / how to add outs/deps without re-running stage?

jorgeorpinel:
~~Let's incorporate these changes to #1913 and close this PR.~~

This PR will now also incorporate https://github.com/iterative/dvc.org/pull/1913#pullrequestreview-525917714

---

Fixes #460 (How to add dependencies to a stage without re-running it)


content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,56 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the build requirements are missing from `dvc.yaml`:
+
+- Files or directories in the <abbr>workspace</abbr> that are dependencies of
+  the stage, are missing from `deps` field.
```
imhardikj, line 8: @jorgeorpinel I have made some changes to add dependencies part. Also updated to run/commit.

jorgeorpinel: Sure, reviewing https://github.com/iterative/dvc.org/pull/1913#pullrequestreview-525917714 now. Resolving this. Thanks

content/docs/sidebar.json
```diff
@@ -100,8 +100,8 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
           "undo-adding-data",
+          "add-deps-or-outs-to-a-stage",
```
jorgeorpinel, line 104: Not sure about the label. Let's try:
```suggestion
        {
          "label": "Add Tracked Data to a Stage",
          "slug": "add-deps-or-outs-to-a-stage"
        },
```

jorgeorpinel:
```suggestion
        {
          "label": "Add Tracked Data to a Stage",
          "slug": "add-deps-or-outs-to-a-stage"
        },
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,56 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the build requirements are missing from `dvc.yaml`:
```
jorgeorpinel, line 5: Good idea. Maybe something like is better though:
```suggestion
that some of its data requirements are missing from `dvc.yaml`. Namely, one of:
```
May need formatting

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Files or directories in the <abbr>workspace</abbr> that are dependencies of
+  the stage, are missing from `deps` field.
+
```
jorgeorpinel, line 9:
```suggestion
  the stage are missing from `deps` field.
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of the build requirements are missing from `dvc.yaml`:
+
+- Files or directories in the <abbr>workspace</abbr> that are dependencies of
+  the stage are missing from `deps` field.
+- Output files or directories that the stage creates, which are already in the
+  workspace, are missing from `outs` field.
```
jorgeorpinel, line 10: "from deps" or "from THE deps field"
Same for "outs"

jorgeorpinel: Let's just just say `deps`/`outs` (no "field").

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+
+Follow the steps below to add existing files/directories as
+<abbr>dependencies</abbr> or <abbr>outputs</abbr> to a stage without
+re-executing it again, which can be expensive/time-consuming, and is
```
jorgeorpinel, line 14: "re" + "again" is redundant:
```suggestion
executing it again, which can be expensive/time-consuming, and is
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+unnecessary.
+
+We start with an example `prepare`, which has a single dependency and output. To
```
jorgeorpinel, line 17: Example `prepare` what?

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+add a missing dependency `data.csv`, and output `data/validate` to this stage,
```
jorgeorpinel, line 18: I'm starting to notice your use of commas is sometimes unnecessary. Not a big deal and sometimes stylistic, but please keep an eye on that @imhardikj. Thanks:
```suggestion
add a missing dependency `data.csv` and output `data/validate` to this stage,
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+we can edit `dvc.yaml` like this:
+
+```git
+ stages:
+   prepare:
+     cmd: python src/prepare.py
+     deps:
++    - data.csv
```
jorgeorpinel, line 26: Should probably be a path inside `data/`?
```suggestion
+    - data/raw.csv
```

jorgeorpinel: Applies in a few other places.

content/docs/command-reference/commit.md
```diff
@@ -42,10 +42,10 @@ scenarios are further detailed below.
 
 - In cases where we have previously executed a stage (either by writing
   `dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later
-  notice that some of the output files or directories it creates, which are
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  notice that some of the existing dependencies, or output files/directories it
+  creates, which are already in the <abbr>workspace</abbr>, are missing from
```
jorgeorpinel, line 46: "dependencies it creates" ? Again, let's think the changes through please.
Also, no comma was needed here.

content/docs/command-reference/run.md
```diff
@@ -107,9 +107,10 @@ Relevant notes:
   defined as outputs every time its executed by DVC.
 
 - In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  existing dependencies, or output files/directories it creates, which are
```
jorgeorpinel, line 110: Same, deps are not created by stages (and unnecessary comma).

jorgeorpinel (commented):
A bunch of copy edits above, and a couple more relevant comments ☝️ 

Thanks

content/docs/command-reference/commit.md
imhardikj: Pushed new update to clarify this.

imhardikj: Also should title of the doc be same as label? https://github.com/iterative/dvc.org/pull/1914#discussion_r521781518

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage (either by writing
+`dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later notice
+that some of its data requirements are missing from `dvc.yaml`. Namely, one of:
+
+- Files or directories in the <abbr>workspace</abbr> that are dependencies of
+  the stage are missing from `deps`.
+
```
jorgeorpinel, line 9: I thought I removed this line before. Is the formatting inserting it, or is that you @imhardikj? (Please see https://dvc.org/doc/user-guide/contributing/docs#doc-style-guidelines-javascript-and-markdown about bullet lists.) 
```suggestion
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+- Output files or directories that the stage creates, which are already in the
+  workspace are missing from `outs`.
+
+Follow the steps below to add existing files/directories as
+<abbr>dependencies</abbr> or <abbr>outputs</abbr> to a stage without executing
+it again, which can be expensive/time-consuming, and is unnecessary.
+
+We start with an example stage `prepare`, which has a single dependency and
```
jorgeorpinel, line 17: "exapmle stage prepare"? Sounds strange.
```suggestion
We start with an example `prepare` stage , which has a single dependency and
```

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+output. To add a missing dependency `data/raw.csv` and output `data/validate` to
```
jorgeorpinel, line 18: Let's be clearer, since this part got complicated:
```suggestion
output. To add a missing dependency (`data/raw.csv`) as well as a missing output (`data/validate`) to
```

content/docs/command-reference/commit.md
```diff
@@ -42,10 +42,10 @@ scenarios are further detailed below.
 
 - In cases where we have previously executed a stage (either by writing
   `dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later
-  notice that some of the output files or directories it creates, which are
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  notice that some of the existing dependencies of the stage or output
+  files/directories it creates, which are already in the <abbr>workspace</abbr>,
```
jorgeorpinel, line 46: Need to talk about files/dirs first, deps/outs later. Something like "...notice that some of the existing files/dirs used by the stage as dependencies, or created as outputs..."

content/docs/command-reference/commit.md
jorgeorpinel: What doc? What label? Sorry, didn't get that Q.

content/docs/command-reference/run.md
```diff
@@ -107,9 +107,10 @@ Relevant notes:
   defined as outputs every time its executed by DVC.
 
 - In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  existing dependencies of the stage or output files/directories it creates,
```
jorgeorpinel, line 110: Same. Deps are also files/dirs. Like this is sounds as if only outs are files/dirs.

jorgeorpinel (commented):
A few more minor comments ☝️  but we're almost there, thanks @imhardikj

content/docs/command-reference/commit.md
imhardikj: I meant that title of how-to doc should be same as label "Add Tracked Data to a Stage"? Currently the title is "Add Dependencies or Outputs to a Stage".

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: Actually it's my mistake. I but didn't notice that you removed the empty line in f578ddd06f32b6b660a5ff5917658d771ca76474. Updated now.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: Updated

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: Done

content/docs/command-reference/commit.md
imhardikj: Done

content/docs/command-reference/run.md
imhardikj: Also done

content/docs/command-reference/commit.md
```diff
@@ -42,10 +42,11 @@ scenarios are further detailed below.
 
 - In cases where we have previously executed a stage (either by writing
   `dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later
-  notice that some of the output files or directories it creates, which are
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  notice that some of the existing files/directories used by the stage as
+  dependencies, or created as outputs, which are already in the
+  <abbr>workspace</abbr>, are missing from `dvc.yaml` (`deps` and `outs` field
```
jorgeorpinel, line 47: > existing... which are already in

Redundant. Probably just remove the first "existing". Please double check, re-read it. Thanks

content/docs/command-reference/commit.md
jorgeorpinel: I think we discussed that before. The label is shorter so it fits in the nav. Please always check the docs you're writing on the dev environment, BTW. You will notice important details.

If you have a better suggestion for title or label, please open a separate comment on the corresponding lines of code. Resolcing this. Thanks

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: No problem. Thanks

content/docs/command-reference/run.md
```diff
@@ -107,9 +107,10 @@ Relevant notes:
   defined as outputs every time its executed by DVC.
 
 - In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  existing files/directories used by the stage as dependencies, or created as
+  outputs, which are already in the workspace, are missing from `dvc.yaml`
```
jorgeorpinel, line 111: Same

jorgeorpinel (commented):
~~Pretty much done! Just a couple minor comments above ☝️~~

Nvmd: see https://github.com/iterative/dvc.org/pull/1914#pullrequestreview-534019642 + below.

content/docs/command-reference/commit.md
imhardikj: Done.

content/docs/command-reference/run.md
imhardikj: Done

content/docs/command-reference/commit.md
```diff
@@ -42,10 +42,10 @@ scenarios are further detailed below.
 
 - In cases where we have previously executed a stage (either by writing
   `dvc.yaml` manually and using `dvc repro`, or with `dvc run`), but later
-  notice that some of the output files or directories it creates, which are
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  notice that some of the files/directories used by the stage as dependencies,
+  or created as outputs, which are already in the <abbr>workspace</abbr>, are
```
jorgeorpinel, line 46: > which are

"which could be" would be more precise but OK, since this whole part is in a `, which (something,` form that is equivalent to parentheses (meaning that it doesn't change the result of the sentence).

jorgeorpinel: However these sentences are really long and complicated. Not easy to read...

jorgeorpinel (commented):
~~@imhardikj please approve this PR since I can't, because I opened it. I think it's ready to merge.~~ Actually let's wait for Ivan to double check this.

> I did leave one last comment on some complicated sentences in case you want to try improving that text.

content/docs/command-reference/commit.md
jorgeorpinel: (same in run)

content/docs/sidebar.json
```diff
@@ -100,8 +100,11 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
           "undo-adding-data",
+        {
+          "label": "Add Tracked Data to a Stage",
+          "slug": "add-deps-or-outs-to-a-stage"
+        },
           "update-tracked-files"
```
jorgeorpinel, line 108: Also, there's a formatting error here according to https://app.circleci.com/pipelines/github/iterative/dvc.org/6615/workflows/b705d2d8-9861-41a9-8a2b-e18222cd66b1/jobs/6671

Probably indentation?

jorgeorpinel: Please remember to follow https://dvc.org/doc/user-guide/contributing/docs#doc-style-guidelines-javascript-and-markdown

shcheklein:
CI checks failed?

content/docs/command-reference/commit.md
shcheklein: To be honest I'm not sure that we need anything at all after the `[add missing outputs to an existing stage]` link. That's the whole point of this item, right? Everything else doesn't add much information unless I'm missing something.

Agree with @jorgeorpinel - it feels that it should be simlpler.

content/docs/command-reference/run.md
```diff
@@ -107,9 +107,10 @@ Relevant notes:
   defined as outputs every time its executed by DVC.
 
 - In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+  files/directories used by the stage as dependencies, or created as outputs,
```
shcheklein, line 110: same here - too many details, hard to read - sacrifice some:

`which are already in workspace`
`deps and outs field`
?

content/docs/sidebar.json
```diff
@@ -100,8 +100,11 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
           "undo-adding-data",
+        {
+          "label": "Add Tracked Data to a Stage",
```
shcheklein, line 105: is it only about tracked data?

it can become tracked probably after this operation

should we use the slug and label?

content/docs/user-guide/how-to/undo-adding-data.md
```diff
@@ -3,41 +3,33 @@
 There are situations where you want to stop tracking data added previously.
 Follow the steps listed here to undo `dvc add`.
 
-Let's first add a data file into an example <abbr>project</abbr> using
-`dvc add`, which creates a `.dvc` file to track the data:
+Let's first add a data file into an example <abbr>project</abbr>, which creates
```
shcheklein, line 6: I would eventually expect this how-to to become "Remove data" or something ... inverse of `dvc add` sounds a bit too specific?

shcheklein: though we can keep creating more and more small recipes ... and link them after

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,54 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage (either by writing
```
shcheklein, line 3: In this case title sounds too general for the `dvc commit` scenario. It sounds like all we need to describe in this section is `open dvc.yaml - edit` or execute `dvc run` again. If it happened that files are already in place use --no-exec + `dvc commit`

content/docs/sidebar.json
imhardikj: Done.

content/docs/command-reference/commit.md
imhardikj: Removed these points (https://github.com/iterative/dvc.org/pull/1914#discussion_r526539027):
`either by ..... with dvc run`, `which are already in the workspace`, `deps and outs field respectively`
 Now I think it's much simpler. Also we already have a link to `how-to` which already has all the points I removed,

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: In both cases (either editing `dvc.yaml` or using `--no-exec` with `run`), `dvc commit` is required. So that's why it's a separate scenario in `dvc commit`.

jorgeorpinel:
> CI checks failed?

Yes @shcheklein, thanks for noticing! I did report this in https://github.com/iterative/dvc.org/pull/1914#pullrequestreview-533885828 and it has been now fixed by Hardik 👍

content/docs/command-reference/commit.md
```diff
@@ -40,12 +40,10 @@ scenarios are further detailed below.
...
-  notice that some of the output files or directories it creates, which are
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+- In cases where we have previously executed a stage and later notice that some
+  of the files/directories used by the stage as dependencies, or created as
+  outputs, are missing from `dvc.yaml`. We can
+  [add missing dependencies/outputs to an existing stage](/docs/user-guide/how-to/add-deps-or-outs-to-a-stage)
   without having to execute it again. Use `dvc commit` to update the `dvc.lock`
   file and save outputs to the cache.
```
jorgeorpinel, line 48: Per https://github.com/iterative/dvc.org/pull/1914#pullrequestreview-533883376:
```suggestion
- In some cases, we have previously executed a stage, and later notice that some
  of the files/directories used by the stage as dependencies or created as
  outputs are missing from `dvc.yaml`. It is possible to
  [add missing data to an existing stage](/docs/user-guide/how-to/add-deps-or-outs-to-a-stage).
  Use `dvc commit` to update the `dvc.lock` file and save outputs to the cache.
```
Please double check the text and format it if needed.

content/docs/command-reference/commit.md
jorgeorpinel: The latest version had some errors. I left a separate comment with a specific suggestion so we can finalize this. Resolving here

content/docs/command-reference/run.md
jorgeorpinel: Please apply my suggestion from commit here too @imhardikj.

jorgeorpinel: > deps and outs field

This is a good idea. Something like:

```suggestion
  files/directories are missing from the `deps` or `outs` fields of `dvc.yaml`.
```

It assumes the user will know what it means to be in those fields but I guess you can always click on `dvc.yaml`, find the fields, and see the "dependency" and "output" tooltips from there.

content/docs/sidebar.json
jorgeorpinel: The doc title is too long for the nav bar "Add Dependencies or Outputs to a Stage" even if we shorten "Deps and Outs", I think.

But Tracked Data = Dependencies and Outputs, right? Technically speaking. The meaning/context is different, but what alternative is there? Maybe "Add Existing Data to a Stage" 💡

content/docs/sidebar.json
```diff
@@ -100,8 +100,11 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
           "undo-adding-data",
+          {
+            "label": "Add Tracked Data to a Stage",
```
jorgeorpinel, line 105:
```suggestion
            "label": "Add Deps or Outs to a Stage",
```

content/docs/sidebar.json
jorgeorpinel: Actually, "Add Deps or Outs to a Stage" seems to fit:

image

I've updated it now. Resolving

content/docs/sidebar.json
```diff
@@ -100,8 +100,11 @@
         "slug": "how-to",
         "source": false,
         "children": [
-          "add-output-to-stage",
           "undo-adding-data",
+          {
+            "label": "Add Deps or Outs to a Stage",
+            "slug": "add-deps-or-outs-to-a-stage"
+          },
```
jorgeorpinel, line 107: Actually, I guess we can just use the slug then, like Ivan suggested:
```suggestion
          "add-deps-or-outs-to-a-stage",
```

jorgeorpinel: Committing...

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: > too general for the dvc commit scenario... all we need to describe in this section is...

@shcheklein do you mean in https://github.com/iterative/dvc.org/pull/1914/files#diff-b8f371cb91c101e733f37e95de023afe203b77b8a0443b3d453a76739b675852R43-R48 above?

The scenario in `commit` is "we have previously executed a stage and later notice that some files/directories are missing from `deps` or `outs` in `dvc.yaml`" which seems to me exactly like what this doc addresses.

> Cc @imhardikj that would be the whole text in it's most compressed form BTW.

Or are you saying that the long text and example in this how-to is unnecessary and could be compressed into a single paragraph (thus maybe not even needing a whole how-to page)?

Thanks

content/docs/user-guide/how-to/undo-adding-data.md
jorgeorpinel: Good ideas. Should we write this in #899? We probably don't want to expand the scope of this PR (even further 😅).

jorgeorpinel: Added

- Explore changing the existing how-to "Undo Adding Data" to "Remove data" — what other ways to remove data other than un-doing `add` are there?

to the how-tos epic. Resolving here 🙂

content/docs/command-reference/commit.md
imhardikj: Done

content/docs/command-reference/run.md
imhardikj: Updated this

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
shcheklein: My point is that when I read `Add Dependencies or Outputs to a Stage` I would expect to a much simpler scenario first - `Open your file, add it manually, save file. Run repro or commit in some cases.` - that kind of a thing. Not one of the very specific sub-scenarios from the `dvc commit`.

content/docs/command-reference/commit.md
shcheklein: This is still too complicated to my mind. Very hard to read.

`... , but later  .... user by, which ... are (more explanation here)` - let's try to avoid this.

Let's try to keep sentence structure simple. But in this case you probably don't need even that level of details (like `outs` field?)

content/docs/command-reference/run.md
shcheklein: It's better, @imhardikj ! It's always better to go first with a simple statement sentence - what it does. Then explanation if it's needed. This still a very complicates sentence structure and wrong order of sentences within a paragraph to my mind.

content/docs/command-reference/commit.md
```diff
@@ -40,14 +40,11 @@ scenarios are further detailed below.
...
-  already in the <abbr>workspace</abbr>, are missing from `dvc.yaml` (`outs`
-  field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
-  without having to execute it again. Use `dvc commit` to update the `dvc.lock`
-  file and save outputs to the cache.
+- In some cases, we have previously executed a stage, and later notice that some
+  of the files/directories used by the stage as dependencies or created as
+  outputs are missing from `dvc.yaml`. It is possible to
+  [add missing data to an existing stage](/docs/user-guide/how-to/add-deps-or-outs-to-a-stage).
+  Use `dvc commit` to update the `dvc.lock` file and save outputs to the cache.
```
shcheklein, line 47: It is actually better and good enough! 

I would may try still to change the order though. It is possible to add ... 

but it's fine as is I think.

content/docs/command-reference/run.md
```diff
@@ -106,10 +106,10 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
-- In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+- In some situations, we have previously executed a stage, and later notice that
+  some of the files/directories are missing from the `deps` or `outs` fields of
```
shcheklein, line 110: do we need to mention `outs` and `deps` fields here? 

also consider reverse order of the sentences.

content/docs/command-reference/run.md
shcheklein: outdate comment ^^, disregrad

content/docs/command-reference/commit.md
shcheklein: also an outdated comment, disregard please ^^

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: Gotcha Ivan. You're right. We should simplify the intro to this doc and separate that from the example below @imhardikj. I'll resolve this conversation and reopen in the exact places...

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,54 @@
...
+that some of its data requirements are missing from `dvc.yaml`. Namely, one of:
+
+- Files or directories in the <abbr>workspace</abbr> that are dependencies of
+  the stage are missing from `deps`.
+- Output files or directories that the stage creates, which are already in the
+  workspace are missing from `outs`.
+
+Follow the steps below to add existing files/directories as
+<abbr>dependencies</abbr> or <abbr>outputs</abbr> to a stage without executing
+it again, which can be expensive/time-consuming, and is unnecessary.
```
jorgeorpinel, line 14: Ok so this can be simplified quite a bit probably, as mentioned in https://github.com/iterative/dvc.org/pull/1914#pullrequestreview-534022299. E.g.:

> all we need to describe in this section is open dvc.yaml - edit or execute dvc run again. If it happened that files are already in place use --no-exec + dvc commit

> much simpler scenario and recommendation first - Open your file, add it manually, save file. Run repro or commit in some cases.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,54 @@
...
+
+We start with an example `prepare` stage, which has a single dependency and
+output. To add a missing dependency (`data/raw.csv`) as well as a missing output
+(`data/validate`) to this stage, we can edit `dvc.yaml` like this:
```
jorgeorpinel, line 18: And starting somewhere around here it can all be in a separate `## Example` section.

content/docs/command-reference/commit.md
jorgeorpinel: Agree on the order of the ideas here, let me think... ⌛

content/docs/command-reference/run.md
jorgeorpinel: mentioning deps/outs fields was just a shorter alternative to "files/directories used by the stage as dependencies or created as outputs are missing from". BTW @imhardikj how we have different text in commit vs here in run. We have to pick one 😅

jorgeorpinel: BTW @imhardikj how we have different text in commit vs here in run. We have to pick one 😅

content/docs/command-reference/commit.md
jorgeorpinel, line 47: The right order of ideas here would be something like:

> It is possible to [add missing data to existing stages], and `dvc commit` can then be used to save the outputs to the cache (and update `dvc.yaml`).

content/docs/command-reference/run.md
imhardikj: Updated. Similar now.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: Added `## Example` section.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: Simplified this into two small paragraphs.

jorgeorpinel: Very good, thanks.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,52 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage, but later notice that some
+of the files/directories used by the stage as dependencies or created as outputs
+are missing from `dvc.yaml`.
+
+To add existing files/directories as <abbr>dependencies</abbr> or
+<abbr>outputs</abbr> to a stage, either edit the `dvc.yaml` file or use
+`dvc run` with the `--no-exec` option. Then use `dvc commit` to save the
+output(s) to the <abbr>cache</abbr> (and update `dvc.lock`).
```
jorgeorpinel, line 10: WDYT @shcheklein ? It's still specific to existing dep/out files or dirs. (which was our original intention). Do you think we need to generalize it even more and leave the `commit` part as an optional step?

Either way I say this is ready to be merged for now, but since I opened the PR, I can't approve it. If you agree, please do.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,52 @@
+# Add Dependencies or Outputs to a Stage
+
+There are situations where we have executed a stage, but later notice that some
+of the files/directories used by the stage as dependencies or created as outputs
+are missing from `dvc.yaml`.
+
+To add existing files/directories as <abbr>dependencies</abbr> or
+<abbr>outputs</abbr> to a stage, either edit the `dvc.yaml` file or use
```
jorgeorpinel, line 8: to a stage "without executing it again" ?

@imhardikj

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
shcheklein: > It's still specific to existing dep/out files or dirs (which was our original intention)

then name should be different?

the point is- name is general, content is not. it makes a bit misleading. Like people might think that this the only way to add, or that it's the most general way when add is needed, etc, etc

content/docs/command-reference/run.md
```diff
@@ -106,10 +106,10 @@ Relevant notes:
   also means that the stage command needs to recreate any directory structures
   defined as outputs every time its executed by DVC.
 
-- In some situations we have executed a stage and later notice that some of the
-  output files or directories it creates, which are already in the workspace,
-  are missing from `dvc.yaml` (`outs` field). We can
-  [add missing outputs to an existing stage](/docs/user-guide/how-to/add-output-to-stage)
+- In some situations, we have previously executed a stage, and later notice that
```
shcheklein, line 109: situations , ... , and later , ... that something/something used by, or .... 

and only after that 

are missing 

this is very hard to read still. to many clauses per sentence.

in most cases in this PR it should be significantly simplified still.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: OK so @imhardikj, to generalize this we need to:
- Invert the order of the paragraphs above (don't talk about existing files in the first one)
- The `dvc commit` part should be in the 2nd one
- Edit text accordingly so it makes sense after the change ^
- The example will also need to emphasize that the last part is needed only for existing deps/outs files or dirs ("Finally, we need...")

content/docs/command-reference/run.md
jorgeorpinel: It should be easier to read after it gets generalized per https://github.com/iterative/dvc.org/pull/1914#discussion_r528391826 above.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
imhardikj: @jorgeorpinel updated with changes as per your suggestion.

jorgeorpinel: Hmmm yeah it didn't make much sense. Bullets 3 and 4 weren't really addressed. I'm going to take over this part... Thanks, resolving.

content/docs/command-reference/run.md
jorgeorpinel: Just rewrote this and did other updates to how-tos per some recent discussions around `remove` on Discord. PTAL.

content/docs/sidebar.json
```diff
@@ -100,9 +100,12 @@
...
         "children": [
-          "add-output-to-stage",
-          "undo-adding-data",
-          "update-tracked-files"
+          {
+            "label": "Un-track Data",
+            "slug": "untrack-data"
+          },
+          "update-tracked-files",
+          "add-deps-or-outs-to-a-stage"
```
jorgeorpinel, line 108: I reordered them and updated the slugs of the new ones.

content/docs/user-guide/how-to/untrack-data.md
```diff
@@ -0,0 +1,45 @@
+# How to Un-track Data
```
jorgeorpinel, line 1: I added "How to" to the titles for SEO purposes.

content/docs/user-guide/how-to/untrack-data.md
```diff
@@ -0,0 +1,45 @@
...
+
+There are situations where you may want to stop tracking data added previously.
+Let's see how it can be done using an example `data.csv` file.
```
jorgeorpinel, line 4: I generalized this one (it's still basically "undo `dvc add`) but expressed in more generic terms that we can expand on later e.g. with pipelines.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
+# How to Add Dependencies or Outputs
+
+We have executed a stage, but later notice that some of the files/directories it
+uses as dependencies or creates as outputs are missing from `dvc.yaml`...
+
+To add files or directories as <abbr>dependencies</abbr> or <abbr>outputs</abbr>
+to an existing stage without re-executing it (which can be expensive and is
+unnecessary) you can either edit the `dvc.yaml` file directly, or use `dvc run`
+with the `-f` and `--no-exec` options to the same end.
```
jorgeorpinel, line 9: I simplified and generalized this new one as we agreed recently.

content/docs/user-guide/how-to/update-tracked-files.md
```diff
@@ -1,76 +1,71 @@
-# Updating Tracked Files
+# How to Update Tracked Files
```
jorgeorpinel, line 1: This old one just has a bunch of copy edits and I reordered the H2 sections below.

jorgeorpinel: > p.s. due to our md formatting it's pretty hard to tell what's going on from the git diff. PTAL at https://dvc-landing-how-to-cb6eiv3d2sr.herokuapp.com/doc/user-guide/how-to/update-tracked-files (vs the existing https://dvc.org/doc/user-guide/how-to/update-tracked-files).

content/docs/user-guide/how-to/update-tracked-files.md
```diff
@@ -1,76 +1,71 @@
...
+$ git commit -m "modify train data"
 ```
 
-## Modifying content
+## Replacing files
 
-"Unlink" the file with `dvc unprotect`. This will make `train.tsv` safe to edit:
+If you want to replace the file, you can take the following steps.
+
+First, [un-track](/doc/user-guide/how-to/untrack-data) the file with
```
jorgeorpinel, line 52: Also, it now links to **How to Un-track Data** how-to from the **Replacing files** section. There's definitely some overlap there (around `remove`).

We could consider merging this section and that doc, not sure.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+
+After the update, use `dvc commit` to save any output files that already exist
```
shcheklein, line 11: This should go as a very specific note if it's a general case. In general case you edit, and run `dvc repro`. `dvc commit` should be considered as a hack in most cases. Users should understand what are they doing when they use it (potentially break reproducibility).

content/docs/user-guide/how-to/untrack-data.md
shcheklein, line 1: you should use git mv or something like that to move the file? otherwise there is no way to review what has changed in the file. It looks like a new one

shcheklein: Un-track feels and looks strange a bit ...

content/docs/user-guide/how-to/untrack-data.md
shcheklein: it's not undo `dvc add` though ... files is still in cache in this case. We should run `gc` or remove file manually from cache to make undo `dvc add`. Or at least mention about this.

shcheklein: okay, I see that it is mentioned

content/docs/user-guide/how-to/untrack-data.md
```diff
@@ -0,0 +1,45 @@
...
+data.csv    data.csv.dvc
+$ cat .gitignore
+/data.csv
+```
+
+</details>
+
+> Note that if you are using `symlink` or `hardlink` as
+> [link type](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache),
+> you will have to `dvc unprotect` the tracked file first.
```
shcheklein, line 26: you don't need `dvc unprotect` if you use `dvc remove` I hope

content/docs/user-guide/how-to/untrack-data.md
jorgeorpinel: Yes. How about "stop tracking"? Updated.

content/docs/user-guide/how-to/untrack-data.md
jorgeorpinel: True, I said I'd double check on this. `remove`'s help output does say it unprotects it so I'll remove this anyway (if it doesn't, then it's a bug)... ⌛

jorgeorpinel: Done. Clarified about remove unlinking and about gc along the way.

content/docs/user-guide/how-to/untrack-data.md
jorgeorpinel: The .dvc file? It will be deleted by `dvc remove`. So I guess the output of `git status` would be different if the .dvc file was previously committed, but probably no need to focus much on Git here?

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: OK, I started with a conditional now, but I don't imagine this is such a rare edge case. May even be the norm that you realize the stage needs an update only after haven forgotten to define all the outs and executing it that way.

I don't know if the generic case of just continuing to edit dvc.yaml before having reproed the pipeline even needs a doc. It will be pretty obvious for those who create stages by hand by editing dvc.yaml, no? People who may find this useful will have already done a repro or used run from the beginning, I assume.

content/docs/user-guide/how-to/stop-tracking-data.md
```diff
@@ -0,0 +1,41 @@
+# How to Stop Tracking Data
```
shcheklein, line 1: Hmm ... still don't like the name :(


How would people search for this? What would be the query if we try think as an end user? Are there analogs for Git - how do they phrase this?

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
+# How to Add Dependencies or Outputs
+
+We have executed a stage, but later notice that some of the files/directories it
```
shcheklein, line 3: Is it needed? the whole paragraph?

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
...
+uses as dependencies or creates as outputs are missing from `dvc.yaml`...
+
+To add files or directories as <abbr>dependencies</abbr> or <abbr>outputs</abbr>
+to an existing stage without re-executing it (which can be expensive and is
```
shcheklein, line 7: I would still simplify it:

To add files or directories ... edit the `dvc.yaml` (alternatively do `dvc run --no-exec` if you still have this command around).

Now, to update the `dvc.lock` either run `dvc repro` or `dvc commit` to "force" it in case outputs already exist and you don't want to run anything again.

-----

It's better to push people editing file manually. `dvc run` is complicated, people don't have it around sometimes, etc.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
```diff
@@ -0,0 +1,55 @@
+# How to Add Dependencies or Outputs
```
shcheklein, line 1: add to something?

content/docs/user-guide/how-to/stop-tracking-data.md
jorgeorpinel: The most obvious would be "remove data" but that collides with `dvc remove`... Maybe we don't really need this doc and rather an example there? The only thing that makes this how-to special is the usage of `gc`.

> This all cames from #625 BTW (now a check box in the how-tos epic).

jorgeorpinel: OR like that issue suggests, this How To could be "How to Undo/Fix (Common) Mistakes" and keep the undoing add recipe as a first one. But only if we can identify other common mistakes to add...

jorgeorpinel: Oh and in fact `remove` already has a similar example which we could just enhance with `git status` and `dvc gc`: https://dvc.org/doc/command-reference/remove#example-remove-a-tracked-file-or-directory — I'm inclining to do this and remove this whole how-to for now...

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: It's a single sentence on motivation.

Like I mentioned, I realized that this how-to is really only useful for people who have already reproed their dvc.yaml or used run from the beginning. For people that write dvc.yaml manually and haven't reproed it, it's obvious how to add more deps/outs.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: > to update the dvc.lock either run dvc repro

But the scenario is "add deps/outs to an existing stage **without re-executing it**". I simplified it along the lines of your suggestion for now, but skipped the parts that contradict that.

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: Unfortunately the title is too long so I felt like we needed to compromise. See nav vs. title here:

image

I've added a SEO `title` now that is more correct though. Should help

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: > this how-to is really only useful for people who have already reproed their dvc.yaml or used run from the beginning

Let me think this through thouogh... ⌛

content/docs/user-guide/how-to/stop-tracking-data.md
jorgeorpinel: OK I changed this to "How to Reverse Common Mistakes" although there's only one H2 section in there for now (Stop tracking data).

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: OK I went your way on this. PTAL

content/docs/user-guide/how-to/add-deps-or-outs-to-a-stage.md
jorgeorpinel: p.s. on this, I think it's OK since "Dependencies or Outputs" can only apply to stages anyway.

content/docs/user-guide/how-to/reverse-common-mistakes.md
```diff
@@ -0,0 +1,45 @@
+# How to Reverse Common Mistakes
```
shcheklein, line 1: hmm ... I get the idea, but feel to early for such generalization

I think even Stop tracking is better if we can't find anything else ... let's do that one

content/docs/user-guide/how-to/update-tracked-files.md
```diff
@@ -1,76 +1,72 @@
-# Updating Tracked Files
+# How to Update Tracked Files
```
shcheklein, line 1: Files or Data - in some case we use Data, in some cases Files. is it okay?

shcheklein (approved):
Looks great to me, just a few comments left - I would simplify the title.

content/docs/user-guide/how-to/update-tracked-files.md
jorgeorpinel: Data is better but this is a pre-existing use case and renaming it was outside of the scope here esp. since it requires a redirect.

TBH I want to move other guides into this section: Merge Conflicts and Setup a Google Drive Remote for starters. So should we do all that in a separate PR?

content/docs/user-guide/how-to/reverse-common-mistakes.md
jorgeorpinel: K, done.

content/docs/user-guide/how-to/update-tracked-files.md
shcheklein: > So should we do all that in a separate PR? 

yes, def not this PR!

> Setup a Google Drive Remote

this should part of a bigger section that has a page per remote type. It's not How to

> Merge Conflicts

It can be probably part of the `How to` indeed.

jorgeorpinel: > Setup a Google Drive Remote...
should part of a bigger section that has a page per remote type. It's not How to

I think it's definitely "how-to" material, even if we put that remote config section somewhere else. I'll let it be for now then.

shcheklein: You can put pretty much anything into How to this way ... How to manage external data, How to contribute, How to optimize deal with large data, How to run on Windows, etc ...


I think there should be something else that defines this section - closer to FAQ? some specific workflow questions?

In case of Google Drive - we'll have one page per remote. Even size in this case makes a bad fit. Eventually it might become bigger than everything else.

jorgeorpinel: Continued in #1976

PR #1916, opened 2020-11-09
remote: Document support for WebHDFS

### LINKED ISSUE #1899: remote: Document support for WebHDFS
Support for WebHDFS has been discussed in https://github.com/iterative/dvc/issues/4687 and is being implemented in https://github.com/iterative/dvc/pull/4747.

Documentation should be updated to reference WebHDFS as an alternative to HDFS: anywhere HDFS is used, WebHDFS can be used instead. It should be mentioned that the advantage of using WebHDFS over pure HDFS is it eliminates the requirement of having `libhdfs` installed.

tomasfarias:
Closes #1899 (see iterative/dvc/pull/4747)

This PR adds documentation for WebHDFS support, as it's being implemented in the related PR: https://github.com/iterative/dvc/pull/4747. For the most part, this PR adds documentation in `content/docs/`, I'm a bit unsure whether something is needed in `user-guide/`.


content/docs/command-reference/remote/modify.md
```diff
@@ -674,6 +674,74 @@ more information.
 
 <details>
 
+### Click for WebHDFS
```
shcheklein, line 677: @tomasfarias great PR, thanks 🙏 . I think my first question reading the docs would be - which one of HDFS remotes should I use. Would be great to summarize some distinctions, I think, and why do we have both, and when should be one used or another.

tomasfarias: Thanks for the review. I added some commentary that seeks to address the questions you raised. However, I'm not entirely sure if this would be the right place to add said commentary. Would like some input on that, I can move it around if needed.

jorgeorpinel: @tomasfarias I think this new section should come after the HDFS section, and then start the section with a short paragraph explaining why there's 2, when to use each, etc. Mainly in the `remote add` command reference BTW. It's probably not needed here in `remote modify` (or just a very short paragraph/note).

jorgeorpinel:
> this PR adds documentation in content/docs/, I'm a bit unsure whether something is needed in user-guide/.

Not sure what you meant @tomasfarias, since user-guide/ is in content/docs/ 🙂

> anywhere HDFS is used, WebHDFS can be used instead
From https://github.com/iterative/dvc.org/issues/1899#issue-733766438

What I would do is to search for "hdfs" (case-insensitive) in the whole repo. There are 42 results in 13 files... Not sure all are important, let me check ⌛

jorgeorpinel (commented):
List of files where WebHDFS info. should be added (and what exactly):

- content\docs\command-reference\config.md (new config option)
- content\docs\command-reference\get-url.md and content\docs\command-reference\import-url.md (add URL tables row)
- content\docs\command-reference\remote\add.md and content\docs\command-reference\remote\modify.md (as discussed above)
- content\docs\user-guide\external-dependencies.md and content\docs\user-guide\managing-external-data.md ? Not sure, does it support external deps/outs?

Other files that mention HDFS in a way that we could append WebHDFS, but I'm not sure we want to do so since they're essentially the same thing (prob not):

- content\docs\start\data-versioning.md
- content\docs\user-guide\dvc-files-and-directories.md

tomasfarias:
@jorgeorpinel Thanks for the pointers 🙌 

* \command-reference\get-url.md and \command-reference\import-url.md already had WebHDFS added to the URL table.
* As suggested, moved the explanation/comparison of hdfs and webhdfs to command-reference\remote\add.md from command-reference\remote\modify.md.
* I lean towards not adding WebHDFS to start\data-versioning.md since, as suggested, WebHDFS uses the same storage as HDFS, which is already mentioned. However, for user-guide\dvc-files-and-directories.md I do mention WebHDFS next to HDFS since the checksum hash value also applies to it.
* External deps/outs should be supported in a similar fashion as how they are for HDFS.

tomasfarias:
Maybe should also mention the relevant dvc PR was recently merged: https://github.com/iterative/dvc/pull/4747

content/docs/command-reference/remote/add.md
```diff
@@ -334,6 +334,33 @@ $ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
 
 <details>
 
+### Click for WebHDFS
+
+**HDFS and WebHDFS:**
```
shcheklein, line 339: Looks great @tomasfarias ! One last thing from me - let's add into `dvc remote add` HDFS, `dvc remote modify` HDFS/WebHDFS sections a link/note/something that mentions about two options and that details could be found here.

tomasfarias:
restyled bot took more or less 10 minutes to pick up my changes and by that time I had already run `prettier` locally myself. Machine uprising is still far away 🤖

content/docs/command-reference/remote/add.md
tomasfarias: Added a note in `dvc remote add` HDFS and two links + note in `dvc remote modify` HDFS/WebHDFS as you suggested 👍 . Let me know if they are clear enough.

content/docs/command-reference/remote/modify.md
```diff
@@ -683,6 +687,78 @@ more information.
...
+
+  ```dvc
+  $ dvc remote modify --local myremote webhdfs_alias myalias
+  ```
+
+> The username, token, webhdfs_alias, and hdfscli_config may contain sensitive
+> user info. Therefore, it's safer to add it with the `--local` option, so it's
+> written to a Git-ignored config file.
+
+💡 WebHDFS serves as an alternative for using the same remote storage supported
```
shcheklein, line 754: Looks great, but let's move to the top of these expandable sections?

shcheklein (approved):

shcheklein:
@tomasfarias this is perfect 👌 thanks 🙏

content/docs/command-reference/remote/add.md
```diff
@@ -334,6 +337,33 @@ $ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
...
+
+Both remotes, HDFS and WebHDFS, allow using a Hadoop cluster as a remote
+repository. However, HDFS relies on `pyarrow` which in turn requires `libhdfs`,
+an interface to the Java Hadoop client, that must be installed separately.
```
jorgeorpinel, line 346: Looks like this long text could be boiled down a lot (less implementation details).

content/docs/command-reference/remote/add.md
```diff
@@ -334,6 +337,33 @@ $ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
...
+Meanwhile, WebHDFS has no need for this requirement as it communicates with the
+Hadoop cluster via a HTTP REST API using the Python libraries `HdfsCLI` and
+`requests`. The latter remote should be preferred by users who seek easier and
+more portable setups, at the expense of performance due to the added overhead of
+HTTP.
```
jorgeorpinel, line 351: Same. Great info but too much detail. Will people read it?

content/docs/command-reference/remote/add.md
```diff
@@ -334,6 +337,33 @@ $ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
...
+
+One last note: WebHDFS does require enabling the HTTP REST API in the cluster by
+setting the configuration property `dfs.webhdfs.enabled` to `true` in
+`hdfs-site.xml`.
```
jorgeorpinel, line 355: ~~One last note:~~ WebHDFS ~~does~~ requires...

content/docs/command-reference/remote/add.md
```diff
@@ -334,6 +337,33 @@ $ dvc remote add -d myremote hdfs://user@example.com/path/to/dir
 
 <details>
 
+### Click for WebHDFS
+
+**HDFS and WebHDFS:**
+
+Both remotes, HDFS and WebHDFS, allow using a Hadoop cluster as a remote
```
jorgeorpinel, line 344: In fact I'd probably move this whole text to after the `remote add` example.

content/docs/command-reference/remote/modify.md
```diff
@@ -574,6 +574,10 @@ more information.
 
 ### Click for HDFS
 
+💡 Using a HDFS cluster as remote storage is also supported via the WebHDFS API.
+Read more about by expanding the WebHDFS section in
+[`dvc remote add`](/doc/command-reference/remote/add#supported-storage-types).
+
```
jorgeorpinel, line 580: Prob not needed in this doc. since you come here after already decided which one to use.

content/docs/command-reference/remote/modify.md
```diff
@@ -683,6 +687,78 @@ more information.
 
 <details>
 
+### Click for WebHDFS
+
+💡 WebHDFS serves as an alternative for using the same remote storage supported
+by HDFS. Read more about by expanding the WebHDFS section in
+[`dvc remote add`](/doc/command-reference/remote/add#supported-storage-types).
```
jorgeorpinel, line 694: Again, this would prob. belong in `remote add` (maybe as the intro instead of the "regular vs web HDFS" text).

content/docs/command-reference/remote/modify.md
```diff
@@ -683,6 +687,78 @@ more information.
...
+  or if using a `HdfsCLI` cfg file. May only be used when Hadoop security is
+  off. Defaults to current user as determined by `whoami`.
+
+  ```dvc
+  $ dvc remote modify --local myremote user myuser
+  ```
+
+- `token` - Hadoop delegation token for WebHDFS, can be empty in case of using
+  `user` or if using a `HdfsCLI` cfg file. May be used when Hadoop security is
+  on.
```
jorgeorpinel, line 712: "May only" ?

content/docs/command-reference/remote/modify.md
```diff
@@ -683,6 +687,78 @@ more information.
...
+
+  ```dvc
+  $ dvc remote modify --local myremote token mytoken
+  ```
+
+- `hdfscli_config` - path to a `HdfsCLI` cfg file. WebHDFS access depends on
+  `HdfsCLI`, which allows the usage of a configuration file by default located
+  in `~/.hdfscli.cfg`. In the file, multiple aliases can be set with their own
+  connection parameters, like `url` or `user`. If using a cfg file,
+  `webhdfs_alias` can be set to specify which alias to use.
```
jorgeorpinel, line 722: Could simplify a lot:
```suggestion
- `hdfscli_config` - path to a `HdfsCLI` cfg file, by default located
  in `~/.hdfscli.cfg`. If using a cfg file,
```

And ideally link somewhere instead of explaining what's inside .cfg files, maybe https://hdfscli.readthedocs.io/en/latest/quickstart.html? (`webhdfs_alias` referring back to this should be enough, I think).

content/docs/command-reference/remote/modify.md
```diff
@@ -683,6 +687,78 @@ more information.
...
+  url = http://example.com/path/to/dir
+  user = myuser
+
+  [production.alias]
+  url = http://prodexample.com/path/to/dir
+  user = produser
+  ```
+
+  See more information in the `HdfsCLI`
+  [docs](https://hdfscli.readthedocs.io/en/latest/quickstart.html#configuration).
```
jorgeorpinel, line 744: Again, not sure we want to document WebHDFS. I see the link is there already so I'd say we should just link to that casually.

content/docs/user-guide/managing-external-data.md
```diff
@@ -161,6 +161,25 @@ it. So systems like Hadoop, Hive, and HBase are supported!
 
 <details>
 
+### Click for WebHDFS
+
+```dvc
```
jorgeorpinel, line 166: Just missing adding WebHDFS to the bullet list earlier in this doc., maybe in the same line as HDFS.

jorgeorpinel: Also missing similar updates to the ext deps page.

jorgeorpinel (commented):
Sorry, very late post-merge review from my part. This was a great contribution BTW, very well written and explained.

Here's a list of stuff I'd like to address later, when possible:

PR #1921, opened 2020-11-10
Fix s3 remote commands in docs (remove url text placeholder)

brbickel:
Very minor typo fix.


shcheklein (approved):
Good catch @brbickel ! thanks!

PR #1922, opened 2020-11-11
November HB draft

elleobrien:
I need to create an image still for the header. Maybe a new picture of DeeVee is in order

Also will update comment link after review


shcheklein (approved):
Looks great, @elleobrien ! 🎉  Feel free to merge when image and other meta (if needed) is fixed.

Issue #1924, opened 2020-11-11
development server is slow to build


elleobrien:
Working from a local clone to write new blogs, building a dev server (`yarn develop`) is getting pretty slow (>1 minute). Is there a way to make a smaller clone; i.e., maybe only blogposts after a certain date (`%today-30`) or not checkout the docs?

rogermparent:
I'm on it! There's a few features I'm aware of that we can try to curb this issue.

1. [Conditional Rebuilds] - This has been in for a while, and is a basic way to only rebuild pages that have been edited. While it doesn't work on the Heroku builds, it has worked for me locally and may speed up writing.
2. [Query-on-demand alpha] - This is a new alpha feature that I was notified of in an email- it essentially makes Gatsby dev mode only run queries as you go to the individual page.

I'll try both in a branch.

rogermparent:
The prior two features have showstopping issues that prevent them from local use, however:

> Is there a way to make a smaller clone; i.e., maybe only blogposts after a certain date (%today-30) or not checkout the docs?

We can adapt the `LIMIT_BLOG_PAGES` flag I have in already and bind it to an npm script to make invoking it easier. This way, we can have a separate limited mode for blog authoring that's quicker to build while maintaining the functionality of a full-site `develop`.

elleobrien:
@rogermparent i love that idea!

rogermparent:
I've made a new `yarn author` command that does just that at #1932, and it cuts development server spin-up time to 25s on a hot cache by building only the latest blog post and skipping Docs entirely.

I didn't time it on a cold cache yet, but the difference should be *much* bigger in that case as the images of skipped pages are also skipped.

jorgeorpinel:
> yarn author command ... building only the latest blog post and skipping Docs entirely

What about the opposite? A `yarn docs` command that only builds /docs and skips the whole blog? Would that be easy to include @rogermparent? Thanks

jorgeorpinel:
(Assuming it's significantly faster than `yarn develop`)

jorgeorpinel:
Taking this to an extreme (prob not worth it) `yarn develop` could take arguments to specify which modules to work with (`all` by default):
- `home` (could be always included)
- `land(ing)` (Features and Support, could be part of `home`)
- `docs`
- `blog` or `author`
- `com(munity)`

rogermparent:
> > yarn author command ... building only the latest blog post and skipping Docs entirely
> 
[...]

PR #1932, opened 2020-11-11
Add blog author script with reduced build time

rogermparent:
This PR's main goal is to add a new npm/yarn script, `author`, which runs `gatsby develop` on a subset of content to increase build speed.

In particular, it limits to only the most recent blog page and skips docs entirely.

To use it, just use `yarn author` over `yarn develop` and make sure the blog page you're working on is the latest.

The change uses and improves the LIMIT_BLOG_PAGES env var, which now takes a number of blog pages to limit to. All blog posts are still read into schema, but pages (and images) will only be generated for the pages that are actually made.

It also adds a similar SKIP_DOCS boolean environment variable that disables both node and page creation for docs. Schema is still created so as not to break other types and queries that refer to docs.

On the machine I'm on now, with a warm cache, `yarn author` takes 25 seconds to get to a usable server. This is a big improvement over `develop` in the same context, and should be *much* moreso with no cache.

Addresses #1924


shcheklein:
👍 
let's update contrib docs as well please?

rogermparent:
Additional docs mentioning this script are now [on the deploy](https://dvc-landing-author-mode-su68mw.herokuapp.com/doc/user-guide/contributing/blog)!

shcheklein (approved):

Issue #1127, opened 2020-04-08
term: ambiguous use of "external" and "workspace"


woodshop:
The term "workspace" is defined in the glossary as "Directory containing all your project files..." However advanced dvc usage allows for remote storage of external data and external dependencies. It is not clear from the docs whether the "workspace" includes external data & dependencies. Furthermore, the docs for various commands assume different definitions of workspace. For instance, `dvc push` does not interact with external data, but `dvc checkout` does. This should be clarified in the docs.

UPDATE: Go to https://github.com/iterative/dvc.org/issues/1127#issuecomment-783060111

woodshop:
In general, I find that several definitions could be tightened in the docs along with workspace.

E.g. "external dependencies" and "external data" seem to refer, respectively, to stage inputs and outputs that don't reside in the "workspace" but that are nevertheless tracked by dvc. I think it would be really helpful to clarify this in the docs. At the same time, it would be useful to refer to non-external dependencies/data as "internal" to make it easier to discriminate from external. I suppose "internal" data is located in a directory that shares the same root as ".dvc" but I might be wrong about that.

shcheklein:
More context - https://discordapp.com/channels/485586884165107732/563406153334128681/697459286455877732

jorgeorpinel:
Hi, sorry for the delay. Some feedback:

> advanced dvc usage allows for remote storage of external data and external dependencies. It is not clear from the docs whether the "workspace" includes external data & dependencies

External dependencies, outputs, and cache are the exception to the rule. Here are some of the explanation from docs:
![image](https://user-images.githubusercontent.com/1477535/79404772-3d44f380-7f58-11ea-8840-4db1a3ad5922.png)
![image](https://user-images.githubusercontent.com/1477535/79404801-53eb4a80-7f58-11ea-9134-981bdd310f67.png)
![image](https://user-images.githubusercontent.com/1477535/79404857-78472700-7f58-11ea-9af6-fb4a030459a5.png)

In any case, I don't think we really consider them part of the workspace. They are an extension of it, like Andy correctly interpreted. They are part of the project as a whole. Any suggestion on how/where best we should clarify this?

> refer to non-external dependencies/data as "internal"

Thanks for the idea but I think it would be a little cumbersome to have to call deps and outs "internal" in let's say 95% of the cases just to be bring attention to the exceptions.

> docs for various commands assume different definitions of workspace. For instance, dvc push does not interact with external data, but dvc checkout does

[...]

Issue #53, opened 2018-07-16
user-guide: cache/remote/workspace relationships


efiop:
Describe how these three relate and what they are.

shcheklein:
In the user guide the part about basic DVC concepts is missing. Like described in #144 the User Guide should be starting with some introduction to those concepts with some explanation and how they relate to each other. This introduction should server as a reference when we use these concepts in other documents (e.g. command references, and/or examples, tutorials).

Basic concepts should include:
- working space
- cache
- remote
- dvc file
- etc

vibhor98:
Hi @shcheklein, I'm Vibhor and I'm new to DVC community. I have been contributing to the open source projects since more than a year now. I'm GSoD' 19 aspirant. I would like to take up this issue to create user guide to cover up all the remaining concepts. Shall I go ahead? Thanks!

PS: I have already read DVC documentation and found it really helpful in getting started.

shcheklein:
@vibhor98 hi! welcome to the community. I'm happy to see that you are interested in collaborating with us. Sure, would be great if can make some progress with this. Feel free to do a PR. I would recommend though to start with something simpler just to learn our requirements, learn the workflow, learn DVC, etc. After that jump to this one. Let me know what your thoughts are and if you need our help to pick the issue to start with.

vibhor98:
@shcheklein, thanks for quick response! I completely agree! I would like to start from something simpler and later on take up this one. Can you please help me in picking up simpler issues? Thanks!

shcheklein:
@vibhor98 how about this one - https://github.com/iterative/dvc.org/issues/404? And please, join the chat dvc.org/chat and use #dev-docs channel to ask any question re the DVC.

shcheklein:
@vibhor98 if #404 sounds good, please leave a comment that you are taking it. Thanks!

jorgeorpinel:
We should probably also clarify that DVC remotes are not very much like Git remotes. They are cache backups and not distributed copies of the entire DVC repo. Per https://github.com/iterative/dvc.org/issues/1087#issuecomment-607670846

jorgeorpinel:
Guys, can we close this old one in favor of https://github.com/iterative/dvc.org/issues/550#issuecomment-520541603 ? I feel like the community is mostly aware of these differences/relationships now but adding the concepts section will help too.

shcheklein:
[...]

PR #1944, opened 2020-11-14
[WIP] concepts: move from previous docs to new section, etc.

jeremydesroches:
UPDATE: Latest status at https://github.com/iterative/dvc.org/pull/1944#issuecomment-739065804 (at the time of writing)

Per #550, creates a new **Basic Concepts** section in the user guide.

> Initially an outline of extracted descriptions from requested topics/tooltips for scoping: single document vs. multiple and depth of description. 

Summary of goals and plans from https://github.com/iterative/dvc.org/issues/550#issuecomment-673585050: 
 
> * [x]  There should be a page or section (maybe called **Basic Concepts** in UG, see example in [this closed PR](https://github.com/iterative/dvc.org/pull/1655/files?file-filters%5B%5D=#diff-cfffe134a2be45efa832d561bda45797)) — properly written on a user-level language (closer to UC) — explaining some concepts and DVC philosophy
>   > concepts like Cache, Remote, Wokspace, etc, and their relationships (with diagrams if needed)
> * [x] Possibly extract concept explanations from cmd refs (e.g. cache, dag, metrics/plots, remote)
  > * [x]  Figure out a connection to glossary tooltips (currently special md files with Frontmatter in `content/docs/user-guide/basic-concepts/` — see #1395): Just link to the Concepts? Use a similar engine (with Frontmatter)? Merge completely?
  > * [ ] Find other issues that would be included in solving this (e.g. #53, #1127).
> * ~~Consider also a related DVC Principles section (more philosophical, maybe in the Concepts section index)? See [#550 (comment)](https://github.com/iterative/dvc.org/issues/550#issuecomment-544106096) above.~~ Too much for this PR

**Concepts to be included**

- [x] Workspace
- (Versioning through) Codification ?
- [x] DVC Metafiles (may be part of ^)
- [x] DVC Cache
- [x] Data Pipelines
- [x] Remote Storage
- [x] Metrics and Plots
- [x] Parameters

---

**Concepts without main place in existing docs**? (for a future PR)

- DVC Project (glossary entry exists, may be combined with workspace)

Other notes:

- [ ] rename DVC Dirs&Files guide and DVC metafiles concept page later per https://github.com/iterative/dvc.org/pull/1944#discussion_r533822230
  > But don't extract detailed spec from guide (just the concept explanation)! It will be a separate task to figure that part out.


jeremydesroches:
@jorgeorpinel created this to address #550. Just an initial outline so far but feel free to comment on progress as I go. I'll request your review when decisions required or closer to merge, thanks.

shcheklein:
@jeremydesroches @jorgeorpinel qq- how will it replace/augment already existing basic concepts? I see the intention (one single page vs many pages), but we'll have to duplicate pretty much the information and support it later.

jeremydesroches:
> how will it replace/augment already existing basic concepts? I see the intention (one single page vs many pages), but we'll have to duplicate pretty much the information and support it later.

I think that the existing basic concepts, which are set up like a glossary, could remain and Basic Concepts would expand on select concepts. 

The glossary approach is quite useful and it would be odd to exclude the "concepts". Instead, I suggest linking to the concept from the glossary items that have them. 

For example, the dvc-cache.md glossary/tooltip links to DVC Files and Directories, so we might extract that to the Basic Concepts doc and apply a parallel structure to the other concepts.

With the above approach you could continue building out the glossary while also expanding on the more conceptual items. It avoids having to decide if a given term is best in the glossary or concepts — it can be both. 

In practice all "concepts" should probably have a glossary entry, but not all glossary terms are concepts.

jorgeorpinel:
> qq- how will it replace/augment already existing basic concepts?

Meaning the glossary/tooltips @shcheklein ? Yes. That is one of the questions:

> - [ ] Figure out a connection to glossary tooltips (currently special md files with Frontmatter in `content/docs/user-guide/basic-concepts/`

Not sure... But the tooltips are means as very short summaries so we may need to maintain both? The longer explanations can grow considerably, e.g. DVC remotes as currently explained in https://dvc.org/doc/command-reference/remote.

> one single page vs many pages

In fact we're not very sure of this format. May need one page per concept both because some concepts will be long to explain, and for SEO purposes (see https://github.com/iterative/dvc.org/issues/550#issuecomment-726434615).

jorgeorpinel:
> existing basic concepts, which are set up like a glossary, could remain

@jeremydesroches I would rename that folder to `content/docs/user-guide/glossary/` though, BTW — but there's an issue about that: #1395, we may just want to reactivate it (let's see what Ivan thinks).

Agree about linking from tooltips to concepts.

> not all glossary terms are concepts

➕

content/docs/user-guide/basic-concepts.md
```diff
@@ -0,0 +1,129 @@
+# Basic Concepts
+
+Intro and DVC philosophy...possible diagram of cache/remote/workspace
```
jorgeorpinel, line 3: Well, the most basic concept in DVC is that of versioning through codification (currently being worked in #1747). So we could mention that and link to the Versioning use case as an intro.

content/docs/user-guide/basic-concepts.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Cache
+
+_From `dvc cache`_
```
jorgeorpinel, line 7: Thanks for these notes. Could be `<!-- HTML comments -->` for now if you want.

jorgeorpinel: Or actually, if you can list in the description of the PR the list of initial concepts you intend to put here and where they exist currently, that would be idea.

jorgeorpinel: p.s. If there are any that you think should also be included but don't have a main place in existing docs let's list them too but we can address them in a future PR. This one should be mostly about concentrating existing content to the new doc/section, and link whatever is needed (for SEO purposes).

content/docs/user-guide/basic-concepts.md
```diff
@@ -0,0 +1,129 @@
...
+the <abbr>workspace</abbr> are links\* to (or copies of) the ones in cache.
+Learn more about it's
+[structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
+
+_from tooltip_
+
+The DVC cache is a hidden storage (by default located in the `.dvc/cache`
+directory) for files that are tracked by DVC, and their different versions.
+Learn more about it's
+[structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
```
jorgeorpinel, line 20: OK I take it it's an intermediate step to put this here right? Because we don't want to duplicate these texts exactly, as the tooltips will (most likely) also remain.

shcheklein:
Okay, to share a bit of context they way I see it. Existing "basic concepts" framework was made in way to accommodate longer texts (that's the actual basic concepts thingy) + glossary (that comes from the short description that is provided via frontmatter) + it can drive SEO (again via frontmatter). 

Make take from this discussion is that we should be doing something else only if we want to have everything on the single page.

> In practice all "concepts" should probably have a glossary entry, but not all glossary terms are concepts.

this is true, but it doesn't break the existing framework in any way, right?

content/docs/user-guide/basic-concepts.md
jeremydesroches: > OK I take it it's an intermediate step to put this here right?

Yes, I placed it as a mock-up for discussion purposes.

jeremydesroches:
> Existing "basic concepts" framework was made in way to accommodate longer texts (that's the actual basic concepts thingy) + glossary (that comes from the short description that is provided via frontmatter) + it can drive SEO (again via frontmatter).

Thanks @shcheklein, this context helps a lot. 

> Make take from this discussion is that we should be doing something else only if we want to have everything on the single page.

Given the above context thats how it seems to me, too. To me this is a win-win — I also like the frontmatter approach to create a tooltip, while still having underlying (expanded) content for clickthrough or to drive SEO. 

That would mean we're taking the multiple document approach... sound OK @jorgeorpinel ?

> > In practice all "concepts" should probably have a glossary entry, but not all glossary terms are concepts.
> 
> this is true, but it doesn't break the existing framework in any way, right?

Right, we could add the concepts pages (or page) using the existing framework.

jorgeorpinel:
> Existing "basic concepts" framework was made to accommodate longer texts + glossary (short description via frontmatter).

Oh, I didn't realize md files in `content/docs/user-guide/basic-concepts/` support a description in Frontmatter. Currently they only use the `name` and `match` fields. If that's the case then let's definitely just edit them there, but we'll need a separate ticket to implement displaying them as pages of a special Basic Concepts section with the engine.

> we're taking the multiple document approach... sound OK?

Yes @jeremydesroches. But the most important thing you can provide for starters is the content, whatever the final structure is. Please focus on that in parallel with this discussion. I feel like we still need to catch up on progress on that front.

Thanks

jeremydesroches:
> Oh, I didn't realize md files in `content/docs/user-guide/basic-concepts/` support a description in Frontmatter. Currently they only use the `name` and `match` fields. If that's the case then let's definitely just edit them there, but we'll need a separate ticket to implement displaying them as pages of a special Basic Concepts section with the engine.

I might be missing something but I don't follow how the description frontmatter would work and what the advantage is. That is, using the current glossary/tooltip form to link to Basic Concepts, either as a single document with anchors or multiple documents within a section. I tried adding a `description:` block to test and saw no change to the tooltip behaviour.

In any case, I'll work on content for the concepts in parallel so we can place them in whatever structure is decided.

shcheklein:
> I tried adding a description: block to test and saw no change to the tooltip behaviour.

that can be fixed I think. @rogermparent can we make the engine take the text for tooltips from the frontmatter's description in this case?

@jeremydesroches the question is- should we use the same description in this case for SEO, or should we create two different fields? like `glossary` and `description`?

jeremydesroches:
> @jeremydesroches the question is- should we use the same description in this case for SEO, or should we create two different fields? like `glossary` and `description`?

Good question. The constraints for the meta description (less than 160 characters, doesn't like markup) would make it difficult to address both with the same content. For example, the existing descriptions/definitions have links and are longer than 160 characters.

Your suggestion would work. Even if there was no unique meta description, good answers to the query in the body copy will get picked up automatically.

rogermparent:
> that can be fixed I think. @rogermparent can we make the engine take the text for tooltips from the frontmatter's description in this case?

Yes, we can specifically parse a field as Markdown at build time and use its HTML output to provide glossary text.
If we decide to co-locate pure glossary entries and full Basic Concepts pages in `basic-concepts`, we can have glossary entries fall back to the body if that field is not present.

rogermparent:
I just made PR #1954 against this branch, it includes the ability to override glossary entries with a new frontmatter entry named `tooltip` that also supports Markdown like the current glossary entries.

jeremydesroches:
Hey @rogermparent, the tooltip markdown works great but it seems that the glossary pages are being skipped for rendering. So if I add them to the navigation with sidebar.json or link to them directly the slug is there but the page is rendered blank.

To mock up what (I think) we are after here, I added a new "Concepts" folder and broke out the topics on separate pages. This isn't the latest content but I wanted to check on the direction. 

My understanding of the desired behaviour is that glossary items could be added to the nav and rendered, appearing like these Concepts entries. The frontmatter on each would define the tooltip and the markdown would get rendered as a doc page. This would leave us with a single-source for each concept and/or glossary definition.

Does that sound right @jorgeorpinel and @shcheklein ?

shcheklein:
> My understanding of the desired behaviour is that glossary items could be added to the nav and rendered, appearing like these Concepts entries. The frontmatter on each would define the tooltip and the markdown would get rendered as a doc page. This would leave us with a single-source for each concept and/or glossary definition.

👍 yep, that's my understanding as well.

content/docs/user-guide/concepts/dvc-files.md
```diff
@@ -0,0 +1,26 @@
+# DVC Files
```
jorgeorpinel, line 1: Not sure we need this in basic concepts. There's a guide for these. Also, the term is "DVC Metafiles" (referring to .dvc and dvc.yaml/lock files). But even extracting those, there's still DVC internal files & dirs that should stay in the existing guide.

So the question is should we/ is it worth it to move the metafiles out of the existing guide into here?

content/docs/sidebar.json
```diff
@@ -87,6 +87,18 @@
...
+      {
+        "slug": "concepts",
+        "source": false,
+        "children": [
+          "cache",
+          "dvc-files",
+          "metrics-plots",
+          "pipelines",
+          "remote",
+          "workspace"
```
jorgeorpinel, line 99: Ordered from more to less basic:
```suggestion
          "workspace",
          "dvc-metafiles",
          "cache",
          "pipelines",
          "metrics-plots",
          "remote"
```

content/docs/user-guide/concepts/pipelines.md
```diff
@@ -0,0 +1,20 @@
+# Pipelines
```
jorgeorpinel, line 1:
```suggestion
# Data Pipelines
```

content/docs/user-guide/concepts/remote.md
```diff
@@ -0,0 +1,17 @@
+# Remote
```
jorgeorpinel, line 1:
```suggestion
# Remote Storage
```

content/docs/user-guide/glossary/dvc-project.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+Initialized by running `dvc init` in the **workspace** (typically a Git
+repository). It will contain the
+[`.dvc/` directory](/doc/user-guide/dvc-files-and-directories), as well as
+`dvc.yaml` and `.dvc` files created with commands such as `dvc add` or
+`dvc run`.
+
+This page extends for a _while_, elaborating on what a DVC Project is in a way
+where it doesn't make sense to put **all** this stuff in a tooltip.
```
jorgeorpinel, line 28: Where can we see this page in the review app (link pls)? Thanks

content/docs/user-guide/glossary/workspace.md
```diff
@@ -6,3 +6,6 @@ match: [workspace]
 Directory containing all your project files e.g. raw datasets, source code, ML
 models, etc. Typically, it's also a Git repository. It will contain your DVC
 project.
+
+Learn more about the [workspace concept](/doc/user-guide/concepts/workspace) in
+DVC.
```
jorgeorpinel, line 11: Try to stick to a single paragraph for tooltips, check out how they look in the review app. That last sentence for the link to basic concepts can probably be shorter too.

jorgeorpinel: And I guess we're adding similar links to all tooltip texts?

jorgeorpinel: But wait... I'm confused. I thought we were keeping a single file per concept with both the `glossary` text in Frontmatter, and the md content for the actual docs?

jorgeorpinel:
> > should we use the same description in this case for SEO, or should we create two different fields?
>
> difficult to address both with the same content

@shcheklein I agree with @jeremydesroches on this. The SEO description is (yet) another different text (with different goals and key words) we'd need to maintain, which won't usually be very similar to the tooltip text.

Question: do docs in basic-concepts md files support the `description` field in Frontmatter already?

I'm thinking maybe another alternative (instead of the current `glossary` field) is to only keep `description` (for SEO) and for the tooltips, extract just the first paragraph of the basic concept docs. WDYGT? (G=guys) cc @rogermparent

rogermparent:
> Hey @rogermparent, the tooltip markdown works great but it seems that the glossary pages are being skipped for rendering. So if I add them to the navigation with sidebar.json or link to them directly the slug is there but the page is rendered blank.

You're correct, pages in that glossary folder are skipped in doc page creation currently. The removal of one line will make them created just like regular docs pages again. I assumed we would want to do this at some point, so the line is even marked with a comment.

> To mock up what (I think) we are after here, I added a new "Concepts" folder and broke out the topics on separate pages. This isn't the latest content but I wanted to check on the direction.
> 
> My understanding of the desired behaviour is that glossary items could be added to the nav and rendered, appearing like these Concepts entries. The frontmatter on each would define the tooltip and the markdown would get rendered as a doc page. This would leave us with a single-source for each concept and/or glossary definition.

Totally doable! May I add a commit to re-enable glossary page creation?

jeremydesroches:
> Totally doable! May I add a commit to re-enable glossary page creation?

Yes please, @rogermparent. That way we can see exactly how this approach could work. Thanks!

rogermparent:
I've pushed some commits that that re-enable blog page creation and change sidebar.json to point to the glossary directory to show off docs pages in glossary. I'm certain this is a kludgy representation that will need to be changed, but it gets the idea across.

[Check it out here!](https://dvc-landing-ug-basic-co-kn0p3l.herokuapp.com/doc/user-guide/glossary/dvc-project)

jeremydesroches:
> Question: do docs in basic-concepts md files support the `description` field in Frontmatter already?

Yes @jorgeorpinel , I've verified that both meta title and meta description work when added to the glossary item frontmatter after @rogermparent's latest changes. So there isn't a problem with trying to address both with the same content, and we won't need to extract the first paragraph either.

jeremydesroches:
> I've pushed some commits that that re-enable blog page creation and change sidebar.json to point to the glossary directory to show off docs pages in glossary. I'm certain this is a kludgy representation that will need to be changed, but it gets the idea across.

This is great @rogermparent, thanks! I've merged the concepts and glossary in the latest push. I think this is going to work well... a lot of flexibility between the frontmatter and navigation so we can choose what's needed for each concept.

content/docs/user-guide/basic-concepts.md
jorgeorpinel: I see this page no longer exists so will resolve. We could consider to use an index page here to address:

> - Consider also a related DVC Principles section (more philosophical, maybe in the Concepts section index)? See https://github.com/iterative/dvc.org/issues/550#issuecomment-544106096 above.

But for now I've scratched out that point from the goals of this PR to improve the scope.

content/docs/sidebar.json
```diff
@@ -87,6 +87,19 @@
         "slug": "what-is-dvc",
         "source": "what-is-dvc.md"
       },
+      {
+        "label": "Concepts",
```
jorgeorpinel, line 91: Will commit this for now:
```suggestion
        "label": "Basic Concepts",
```

content/docs/user-guide/basic-concepts.md
jorgeorpinel: Cool

content/docs/user-guide/glossary/workspace.md
jeremydesroches: Agreed on single paragraph. For the links I think it will come down to which items have their own page (basic concepts) and which ones remain as glossary items (same as previous behaviour). Yes, a single file per concept.

content/docs/user-guide/glossary/dvc-project.md
jeremydesroches: This was example copy that Roger added to show up the new functionality, that was merged into my branch. DVC Project is a glossary item but not in our list for Basic Concepts so it's not in the nav and I'll revert it to the previous form.

content/docs/sidebar.json
jeremydesroches: Thanks for the notes on order... I'll update these to follow the labels in the issue description.

content/docs/user-guide/concepts/dvc-files.md
jeremydesroches: > So the question is should we/ is it worth it to move the metafiles out of the existing guide into here?

Let's try to answer this after seeing how the other concepts shape up. Working on it.

content/docs/sidebar.json
jorgeorpinel: K. And actually remote is probably more basic than metrics/plots.

content/docs/user-guide/glossary/dvc-project.md
jorgeorpinel: Yep. This is no longer relevant. Resolving, thanks

content/docs/sidebar.json
```diff
@@ -87,6 +87,19 @@
...
+        "label": "Basic Concepts",
+        "slug": "glossary",
+        "source": false,
+        "children": [
+          "workspace",
+          "dvc-metafiles",
+          "dvc-cache",
+          "data-pipelines",
+          "metrics-and-plots",
+          "remote-storage"
```
jorgeorpinel, line 100:
```suggestion
          "remote-storage",
          "metrics-and-plots"
```

content/docs/user-guide/glossary/dvc-cache.md
```diff
@@ -0,0 +1,22 @@
+---
+name: 'DVC Cache'
+match: ['DVC cache', 'cache', 'caches', 'cached']
+tooltip: >-
+  The DVC cache is a hidden storage (by default located in the `.dvc/cache`
+  directory) for files that are tracked by DVC, and their different versions.
+  Learn more about it's
+  [structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).Learn
+  more about the [concept of cache](/doc/user-guide/glossary/dvc-cache) in DVC.
```
jorgeorpinel, line 9: > Learn more about the concept of cache

Prob no need to use the term "concept". These links could even be more casual e.g. link "DVC cache" at the beginning of the p (this way you also avoid repeating "Learn more about").

content/docs/user-guide/glossary/workspace.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Workspace
+match: [workspace]
+tooltip: >-
+  Directory containing all your project files e.g. raw datasets, source code, ML
+  models, etc. Typically, it's also a Git repository. It will contain your DVC
+  project. Learn more about the [workspace
+  concept](/doc/user-guide/glossary/workspace) in DVC.
```
jorgeorpinel, line 8: Note that links should be the same line, I think. At least for regular md content, Prettier enforces that (but it doesn't understand that this is also md so ignores it).

jorgeorpinel: > For the links I think it will come down to which items have their own page (basic concepts) and which ones remain as glossary items

Agree, BTW. But don't they all have concepts pages as of now?

jorgeorpinel:
> both meta title and meta description work...

I see we're not yet using the `description` field in any doc though @jeremydesroches. Planning to? We can (probably should) definitely leave that for a separate PR, although that's kind a major reason for you to have taken this issue on, I think 😬. So if you want to focus more on that, I can definitely help with finalizing the longer concept texts that you've already centralized from their previous locations.

jorgeorpinel:
> and we won't need to extract the first paragraph

But if I understand the current approach correctly, we will now end up with 3 sources of text for the same concepts: 1 SEO `description` field, 2 `tooltip` field, and 3 long-form md content for the actual rendered doc. I can see 1 being different enough from the others, but 2 and 3 will overlap a lot. So the idea of extracting the 1st paragraph instead of the `tooltip` field is still appealing to me — if I'm understanding this correctly.

Cc @rogermparent sorry to keep picking on you: do you think that ^ would be doable (instead of the `tooltip` field)? Hacky? Just to know whether it's even an option for now. Thanks

> we can have glossary entries fall back to the body if that field is not present

BTW (also Roger) I'm not sure. Maybe we want to have concepts pages that aren't also tooltips. For now all the concepts will have a `tooltip` text, so it's not a big deal anyway.

content/docs/user-guide/glossary/dvc-cache.md
jeremydesroches: OK, updated this and also tried something similar for workspace.

content/docs/user-guide/glossary/dvc-metafiles.md
```diff
@@ -0,0 +1,32 @@
+---
+name: 'DVC Metafiles'
+match: ['DVC files', 'files', 'directories']
+tooltip: 'DVC metafiles tooltip...'
+---
+
+# DVC Metafiles
```
jorgeorpinel, line 7: Yeah now that I see this I think it's best to keep this here and only leave DVC internals in https://dvc.org/doc/user-guide/dvc-files-and-directories which has gotten too long. But I need @shcheklein to confirm on this plan please, since that guide is probably the most important one we have so far.

p.s. if we keep it we can also mention/reiterate on the (versioning through) codification principle here, which will be explained in the new Versioning use case (#1747) — so no need for that concept stand-alone.

content/docs/user-guide/glossary/dvc-cache.md
jorgeorpinel: Thanks.

content/docs/user-guide/glossary/workspace.md
jeremydesroches: > it doesn't understand that this is also md so ignores it

Thanks for pointing this out. Tried the shorter version of linking like you suggested with cache here.

> But don't they all have concepts pages as of now?

Technically all the items in the /glossary/ folder have pages and can be linked to. What I meant is some of those items are listed under Basic Concepts in nav and some still only appear as tooltips (and won't have links to the concept page unless we add it).

jorgeorpinel: > some of those items are listed under Basic Concepts in nav and some still only appear as tooltips

Oh like "dependency", yeah OK gotcha. So yes, let's link from the tooltip to the page of all the ones we are actually editing in this PR. Thanks!

rogermparent:
> But if I understand the current approach correctly, we will now end up with 3 sources of text for the same concepts: 1 SEO `description` field, 2 `tooltip` field, and 3 long-form md content for the actual rendered doc. I can see 1 being different enough from the others, but 2 and 3 will overlap a lot. So the idea of extracting the 1st paragraph instead of the `tooltip` field is still appealing to me — if I'm understanding this correctly.
>
> Cc @rogermparent sorry to keep picking on you: do you think that ^ would be doable (instead of the `tooltip` field)? Hacky? Just to know whether it's even an option for now. Thanks

No worries! These kinds of changes are what I'm specialized for.

While it's theoretically doable, I think it would end up a little hacky and the mechanics behind it would cause issues that would require many tooltips to be overridden to be usable anyway. The three fields in question: **SEO description**, **Tooltip**, and **body** (or rather its first paragraph) are only near identical right now because two of the fields didn't exist before. If I understand the SEO discussion from where the description field was added, the SEO description field will often not even be populated to leverage search engines' context-driven automatic descriptions.

Considering Tooltips are very likely to be small, keeping them as a parallel text to the body isn't much more overhead and avoids the considerations of "what is a paragraph" and possible oddities of using an arbitrary slice of body text as a tooltip.

> BTW (also Roger) I'm not sure. Maybe we want to have concepts pages that aren't also tooltips. For now all the concepts will have a `tooltip` text, so it's not a big deal anyway.

I think this is a fine idea, if nothing else to allow for the option.

All this considered, the route I'd suggest is: 

- We keep the `tooltip` and `match` fields as the only source for tooltip text (maybe changed into a sort of `{tooltip: {body, match}}` structure in the frontmatter)
- Trim tooltips to the minimal necessary text to be useful while delegating all detail to the page's body. 
- Add a link to the source page (if it exists) into the tooltip component such that any user who needs more info than a concept's tooltip provides can easily go to its main page.
- Disable tooltip creation for all pages with `tooltip` and/or `match` specified.

As always, I'm open to discussion or suggestions on this plan.

content/docs/user-guide/glossary/workspace.md
jeremydesroches: > let's link from the tooltip to the page of all the ones we are actually editing here

OK, I'll use that pattern. Added placeholder links to each for now.

content/docs/user-guide/glossary/dvc-metafiles.md
jeremydesroches: > I think it's best to keep this here and only leave DVC internals

Agree on that. Will skip over metafiles for now while we wait for confirmation. In latest deploys you can see how it's looking with cache information extracted already.

> if we keep it we can also mention/reiterate on the (versioning through) codification principle here ... no need for that concept stand-alone

Makes sense. A single source keeps things simple.

jeremydesroches:
> I see we're not yet using the `description` field in any doc though @jeremydesroches. Planning to? 

Yes, I plan to add meta descriptions. This could be a good place to use a pattern for in the meta title too. For example, "Concept: Data Pipelines" and the full title would become "Concept: Data Pipelines | Data Version Control · DVC".

> I can definitely help with finalizing the longer concept texts that you've already centralized from their previous locations.

That would be great. I started with `cache` because the information to extract was already pretty well-defined. I'll continue with the rest like that and then work on refining and adding to the extracted parts. We'll also have quite a few links to fix/check.

jorgeorpinel:
Ah, sorry for the delay here. Somehow missed all the updates until now...

@rogermparent 

> it would end up a little hacky
> only near identical right now because two of the fields didn't exist before

OK I will think we will have considerable duplicity to manage but let's iterate on this later if it's a problem, since using the 1st p is not ideal implementation-wise.

> the possibility of allowing tooltip on any docs page is doable

Interesting... Will keep that in mind, thanks.

> Disable tooltip creation for all pages with tooltip and/or match specified.

I didn't get this part. Why would we disable all tooltips? Did you mean "without"? 🙂 

@jeremydesroches 

> could be a good place to use a pattern for in the meta title too. For example, "Concept: Data Pipelines

No need for that pattern for concepts, I think. Doesn't seem to add anything IMO.

And OK, I'll take over the content, please finalize the SEO titles, descriptions, and also help me by listing which key words we're going for (so I review them and incorporate them to the content). If you can also check all the links and possibly redirects that would also be great.

content/docs/command-reference/cache/dir.md
```diff
@@ -18,7 +18,7 @@ positional arguments:
 ## Description
 
 Helper to set the `cache.dir` configuration option. (See
-[cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).)
+[cache directory](/doc/user-guide/glossary/dvc-cache#structure-of-the-cache-directory).)
```
jorgeorpinel, line 21: Can anchored links have their own search history with Google @jeremydesroches ?

I'm thinking whether and how to setup a redirect for this change... Seems tricky since the old content will be split into a new page.

content/docs/command-reference/cache/index.md
```diff
@@ -15,11 +15,10 @@ positional arguments:
...
 
-The DVC Cache is where your data files, models, etc. (anything you want to
-version with DVC) are actually stored. The data files and directories visible in
-the <abbr>workspace</abbr> are links\* to (or copies of) the ones in cache.
-Learn more about it's
-[structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
+The <abbr>DVC cache</abbr> is where your data files, models, etc. (anything you
+want to version with DVC) are actually stored. The data files and directories
+visible in the <abbr>workspace</abbr> are links\* to (or copies of) the ones in
+cache.
```
jorgeorpinel, line 21: OK, since we moved the concept out of the cmd refs, let's just focus on what the `cache` command (group) is for here please. And link to the concept page

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -239,7 +239,7 @@ Full <abbr>parameters</abbr> (key and value) are listed separately under
   hand or with the command `dvc config --local`.
 
 - `.dvc/cache`: The <abbr>cache</abbr> directory will store your data in a
-  special [structure](#structure-of-the-cache-directory). The data files and
+  special [structure](/doc/user-guide/glossary/dvc-cache). The data files and
```
jorgeorpinel, line 253: Good! There's another 10+ instances of this link/anchor throughout docs though... Let's grep em all

content/docs/user-guide/glossary/dvc-metafiles.md
jorgeorpinel: > I think it's best to keep this here and only leave DVC internals in https://dvc.org/doc/user-guide/dvc-files-and-directories which has gotten too long

Ping @shcheklein WDYT?

content/docs/command-reference/config.md
```diff
@@ -127,8 +127,8 @@ remote. See `dvc remote` for more information.
 A DVC project <abbr>cache</abbr> is the hidden storage (by default located in
 the `.dvc/cache` directory) for files that are tracked by DVC, and their
 different versions. (See `dvc cache` and
-[DVC Files and Directories](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory)
-for more details.) This section contains the following options:
+[DVC Files and Directories](/doc/user-guide/dvc-files-and-directories) for more
```
jorgeorpinel, line 130: Hmmm, I think this should link to /doc/user-guide/glossary/dvc-cache#structure-of-the-cache-directory now, no?

content/docs/sidebar.json
```diff
@@ -87,6 +87,19 @@
         "slug": "what-is-dvc",
         "source": "what-is-dvc.md"
       },
+      {
+        "label": "Basic Concepts",
+        "slug": "glossary",
```
jorgeorpinel, line 92: BTW. How did we end up with a mix of `glossary` in the URLs but "Basic Concepts" in the nav and title? 😋 

Let's use `concepts` in the slug and URL paths too. We just know this is where we have our glossary (tooltips) internally 🙂 

> Sry, will imply lots of link changes again, prob.

content/docs/command-reference/cache/index.md
jorgeorpinel: > let's just focus on what the cache command (group) is for

p.s. like it was done with `dag` 👍

content/docs/user-guide/glossary/data-pipelines.md
```diff
@@ -0,0 +1,33 @@
+---
+name: 'Data Pipelines'
+match: ['data pipeline', 'pipeline', 'pipelines']
+tooltip: >-
+  In DVC, [data pipeline](/doc/user-guide/glossary/data-pipelines) stages and
+  commands, inputs, outputs, interdependencies, and results (intermediate or
+  final) are specified in `dvc.yaml`, which can be written manually or built
+  using the helper command `dvc run`. This allows DVC to restore one or more
+  pipelines later (see `dvc repro`).
```
jorgeorpinel, line 9: Too long/complicated, I think. Probably better to switch with the current 1st paragraph of the body for this one and work it from there. It's best to actually run the dev env and see what the tooltip looks like:

image

content/docs/command-reference/dag.md
```diff
@@ -15,25 +15,8 @@ positional arguments:
...
-written manually or built using the helper command `dvc run`. This allows DVC to
-restore one or more pipelines later (see `dvc repro`).
-
-> DVC builds a dependency graph
-> ([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)) to do this.
-
-`dvc dag` command displays the stages of a pipeline up to the target stage. If
-`target` is omitted, it will show the full project DAG.
+The `dvc dag` command displays the stages of a <abbr>data pipeline</abbr> up to
+the target stage. If `target` is omitted, it will show the full project DAG.
```
jorgeorpinel, line 19: We tend to go straight to the verb in these intros:
```suggestion
Displays the stages of a <abbr>data pipeline</abbr> up to the target stage. If
`target` is omitted, it will show the full project DAG.
```
(committing)

jorgeorpinel:
I just realized we're missing the Parameters concept, which shouold be extracted from the cmd ref index currently in https://dvc.org/doc/command-reference/params. Added a check box to the PR desc.

content/docs/command-reference/cache/index.md
jorgeorpinel: ACTUALLY. We may be able to skip the `cache` index completely and have it jump directly to `cache dir` when you click on this...

jorgeorpinel: UPDATE: I've deleted this file and changed the source to `false` in sidebar.json so the nav entry skips to the first child (`dir`). @jeremydesroches please do the same for Parameters once extracted into its concept page.

content/docs/command-reference/cache/index.md
```diff
@@ -1,37 +0,0 @@
-# cache
-
-Contains a helper command to set the <abbr>cache</abbr> directory location:
```
jorgeorpinel, line 3: BTW another kind of link to review is those done automatically with the `` `dvc cache` `` text in other md files. Those used to link here for an explanation of what the cache is, but should be regular links to the concept now, instead (same for `dvc params`). Thanks

jorgeorpinel: \* same for SOME INSTANCES of `dvc params` — need to check the context.

jorgeorpinel:
Also, this PR is too big 😢 . We should try to address the comments so far and then break it up somehow... Maybe a first one enabling the concepts feature, and then a PR for each concept.

content/docs/command-reference/metrics/index.md
```diff
@@ -63,10 +33,8 @@ stages:
...
 ```
 
-> `cache: false` above specifies that `summary.json` is not tracked or
-> <abbr>cached</abbr> by DVC (`-M` option of `dvc run`). These metrics files are
-> normally committed with Git instead. See `dvc.yaml` for more information on
-> the file format above.
+Note that metrics files are normally committed with Git (that's what
+`cache: false` above is for). See `dvc.yaml` for more information.
 
 ### Supported file formats
```
jorgeorpinel, line 40: This should be moved to the concepts page.

Cmd ref indices are all going to end up being pretty small after this change. Maybe even remove them all (but for now this one can stay).

content/docs/command-reference/plots/index.md
```diff
@@ -17,37 +18,17 @@ positional arguments:
...
-confusion matrices, among others.
+...
 
-This type of metrics files are created by users, or generated by user data
-processing code, and can be defined in `dvc.yaml` (`plots` field) for tracking
-(optional).
+### Supported file formats
 
 DVC generates plots as HTML files that can be open with a web browser. These
 HTML files use [Vega-Lite](https://vega.github.io/vega-lite/). Vega is a
```
jorgeorpinel, line 28: Same

content/docs/command-reference/plots/show.md
```diff
@@ -174,10 +174,60 @@ file:///Users/usr/src/plots/logs.csv.json
...
-    "data": {
-        "values": [
-    {
-        "accuracy": "0.9418667",
-    ...
+  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
+  "data": {
+    "values": [{ "accuracy": "0.9418667", ... }]
+  }
+}
```
jorgeorpinel, line 181: No need to change this.

content/docs/command-reference/plots/show.md
```diff
@@ -174,10 +174,60 @@ file:///Users/usr/src/plots/logs.csv.json
...
+```
+
+## Example: Smooth plot
+
+In some cases we would like to smooth our plot. In this example we will use a
+plot with 1000 data points:
+
+```dvc
```
jorgeorpinel, line 189: Hmmmm. Where does this come from @jeremydesroches? Seems useful but totally out of scope for this PR?

jorgeorpinel: Ooooh I see it was moved here from the index. Yeah that's not necessary, we have some mixed examples in cmd indices some times. Let's put all those back please 🙂

content/docs/command-reference/plots/index.md
```diff
@@ -154,106 +159,3 @@ header (first row) are equivalent to field names.
...
-
-- `-h`, `--help` - prints the usage/help message, and exit.
-
-- `-q`, `--quiet` - do not write anything to standard output.
-
-- `-v`, `--verbose` - displays detailed tracing information.
-
-## Example: Tabular data
-
-We'll use tabular metrics file `logs.csv` for this example:
```
jorgeorpinel, line 168: Same. No reason to remove these.

content/docs/command-reference/plots/index.md
```diff
@@ -154,106 +159,3 @@ header (first row) are equivalent to field names.
...
-```dvc
-$ dvc plots show -t smooth data.csv
-file:///Users/usr/src/plots/plots.html
-```
-
-![](/img/plots_show_smooth.svg)
-
-## Example: Confusion matrix
-
-We'll use `classes.csv` for this example:
```
jorgeorpinel, line 233: Same

jorgeorpinel (changes requested):
I've reviewed everything in detail except the contents of the concept pages. Let's address all the feedback so far and we'll take it from there. As mentioned recently, this PR will have to be split into several after that. Thanks

content/docs/sidebar.json
jeremydesroches: Haha, we went back and forth on it. Fixing this first to minimize link changes on the rest.

content/docs/command-reference/cache/dir.md
jeremydesroches: > Can anchored links have their own search history with Google @jeremydesroches ?

No, they can't. Depending on content and query an #id anchor and subtext might be used for a snippet, but that's true for all structured content.

So we don't need a redirect on the #id anchors. Only for any base URLs that we remove or change. (Plus it doesn't really work like that... would redirect all `.../files-and-directories` links).

content/docs/command-reference/config.md
jeremydesroches: Yep, fixed.

content/docs/user-guide/glossary/data-pipelines.md
jeremydesroches: > Too long/complicated, I think. Probably better to switch with the current 1st paragraph of the body for this one and work it from there.

OK, agreed. Replaced with p1 edited to remove extra links. Also derived the meta description from that (removed markup too).

> It's best to actually run the dev env and see what the tooltip looks like

Yep, I've been doing that. Is this "cleaner" style with only a single link a good pattern to follow?

content/docs/command-reference/plots/show.md
jeremydesroches: OK, reverted.

content/docs/command-reference/plots/index.md
jeremydesroches: OK, reverted.

content/docs/command-reference/plots/index.md
jeremydesroches: OK, reverted.

content/docs/command-reference/plots/show.md
jeremydesroches: OK, reverted. (Was only to fix the sea of red in editor from the open bracket, but now the examples that followed are reverted too.)

content/docs/command-reference/metrics/index.md
jeremydesroches: > This should be moved to the concepts page.

OK, done.

> Maybe even remove them all.

Yes, would also help with SEO on two fronts: 1. removing duplicate/similar content, 2. expanded/improved content on remaining pages

content/docs/command-reference/plots/index.md
jeremydesroches: Done

content/docs/user-guide/basic-concepts.md
jeremydesroches: I browsed and thought on this for a bit and didn't find any obvious candidates. Listed Project as a possible opportunity. `params` is a good one, but it's already in the glossary (and now as a concept.)

content/docs/command-reference/cache/index.md
jeremydesroches: > please do the same for Parameters once extracted into its concept page.

OK, done for Parameters. I also pulled the example from the index there because it was extensive and unique (I think). Good question for the other concepts too: do examples belong or should they remain in cmd ref?

content/docs/user-guide/dvc-files-and-directories.md
jeremydesroches: OK, done. Where it was possible, tried to preserve style by retaining the anchor link.

content/docs/command-reference/cache/index.md
jeremydesroches: Nice... OK, fixed one `dvc cache` link in the docs. Should I fix the one in this blog post too?:

> /blog/2020-04-16-april-20-community-gems.md

jeremydesroches: > * same for SOME INSTANCES of dvc params — need to check the context.

OK, got these too. Didn't change the `dvc params diff` instances.

jeremydesroches:
> I've reviewed everything in detail except the contents of the concept pages. Let's address all the feedback so far and we'll take it from there. As mentioned recently, this PR will have to be split into several after that.

Thanks for the detailed review @jorgeorpinel. I believe I've addressed all of the feedback so far. I've just added meta descriptions for the concepts as well. Will follow-up tomorrow with the keywords for each. Also will add a meta description and keyword for `dvc-metafiles` once that approach is confirmed.

> Also, this PR is too big 😢 . We should try to address the comments so far and then break it up somehow... Maybe a first one enabling the concepts feature, and then a PR for each concept.

Yep, makes sense to enable with parity to existing glossary, then enable each concept as completed. What's the best way (technically) to do that from here with all the WIP? Hoping for a GitHub pro-tip... 😉

content/docs/user-guide/glossary/dvc-metafiles.md
jorgeorpinel: Hi! Update on this one: I did chat with Ivan and we (sort of) agreed to go this way, but will probably change the name to just **DVC Files** here (it may become a section with several pages in the future even), and **DVC Internals** for the guide. No need to do that in this PR; this is just meant as a confirmation.

content/docs/command-reference/cache/dir.md
jorgeorpinel: > So we don't need a redirect

jorgeorpinel: > it doesn't really work like that

True...

> So we don't need a redirect

OK! But how do we best dodge losing traffic for removing half of the content from there? I guess adding a prominent link from the guide to the new concepts page? Could be removed later, when someone remembers.

content/docs/command-reference/cache/index.md
jorgeorpinel: > pulled the example from the (params) index there because it was extensive and unique

Good thinking. Always best to err on the side of caution. I'll finalize the concept pages contents when the time comes.

> Good question for the other concepts too: do examples belong or should they remain in cmd ref?

We want to keep examples in cmd refs. Please don't move those (general rule), just conceptual explanations. Thanks

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: All previous links to the anchor should still use the anchor in the new location. Just linking to the cache concept page could be redundant if the doc already has a cache tooltip, for example. Let's assume link to `#structure-of-the-cache-directory` for a good reason, thanks.

content/docs/command-reference/config.md
```diff
@@ -126,8 +126,8 @@ remote. See `dvc remote` for more information.
 
 A DVC project <abbr>cache</abbr> is the hidden storage (by default located in
 the `.dvc/cache` directory) for files that are tracked by DVC, and their
-different versions. (See `dvc cache` and
-[DVC Files and Directories](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory)
+different versions. (See
+[DVC cache](/doc/user-guide/concepts/dvc-cache#structure-of-the-cache-directory)
 for more details.) This section contains the following options:
```
jorgeorpinel, line 131: Actually given the context in this case no anchor is needed:
```suggestion
[DVC cache](/doc/user-guide/concepts/dvc-cache) for more details.) This section
contains the following options:
```

content/docs/user-guide/dvc-files-and-directories.md
jeremydesroches: > All previous links to the anchor should still use the anchor in the new location.

OK, revisited and checked these again... all with original anchor should have it now.

content/docs/command-reference/cache/dir.md
jeremydesroches: > how do we best dodge losing ranking/traffic for moving half of the content out of the guide?

If the content were totally removed there could be an issue. In this case, content updates are encouraged and moving it to another location on the site is not likely to be penalized. The search engine will notice and index the new content. It may even notice the similarity (a black box to us, though). 

When it is reindexed at the new location it should retain its original "value" as a starting point (same content). Assuming our reorganization and addition of concepts is an improvement (i.e., better content structure and engagement) then it will be able to grow on that basis.

>  I guess adding a prominent link from there to the new concepts page?

Yes, exactly, and any link will do in terms of making sure the search engine finds it and reindexes. There will be multiple links, it's in the nav, and it will show up in the sitemap, which is indexed often for DVC.org.

content/docs/user-guide/glossary/dvc-metafiles.md
jorgeorpinel: And another update on this: actually it's uncertain whether moving the DVC Files specs from the current guide to the concepts section will be an improvement. And since we want concept pages to be brief, after all this is probably not a good idea. If it has already been done I'll take care of figuring it out, otherwise let's not do this now. Thanks

content/docs/command-reference/cache/dir.md
jorgeorpinel: Sounds good. In any case , per my recent https://github.com/iterative/dvc.org/pull/1944#discussion_r536392345, let's not move the details about the file formats to the concepts section (if it hasn't been done).

jorgeorpinel: > let's not move the details about the file formats to the concepts section

I see we haven't yet. OK. Resolving this.

content/docs/user-guide/glossary/data-pipelines.md
jorgeorpinel: Thanks.

> Is this "cleaner" style with only a single link a good pattern to follow?

Yes.

content/docs/command-reference/cache/index.md
jorgeorpinel: > Should I fix the one in this blog post too?:
/blog/2020-04-16-april-20-community-gems.md

I don't see any `dvc cache` link in https://dvc.org/blog/april-20-community-gems

jorgeorpinel: Links will be tricky. This is also why we need to split this PR into many so we can take a look at the details concept by concept. Resolvin this one for nwo.

content/docs/command-reference/metrics/index.md
jorgeorpinel: Looks good, thanks. And yes, I'll prob remove the indices as we work on each concept in it's own PR.

content/docs/command-reference/plots/show.md
jorgeorpinel: Sounds like an editor issue.

content/docs/command-reference/plots/show.md
jorgeorpinel: Thanks

jorgeorpinel:
> enable with parity to existing glossary, then enable each concept as completed. What's the best way (technically) to do that from here

1. First I'd merge the latest master to this branch in case there are any merge conflicts (not the case rn).
2. Create a new branch from (latest) master, move there, and `git checkout ug-basic-concepts -- [file] [file] ...` the files for the first PR and open a PR from that branch.
3. Same for the other PRs one by one, basically. I guess I can finish up the PR split but at least having the base one and a first concept from you would be nice.

Needless to say this PR would be closed.

For now I'll review the SEP meta/desc fields here.

jorgeorpinel:
p.s. perhaps the first concept PR (after the PR that enables this functionality) can be for the DVC Cache, so that most of the cmd refs are included there, as they mostly have the structure-of-cache link change.

content/docs/command-reference/metrics/index.md
```diff
@@ -63,33 +33,8 @@ stages:
           cache: false
 ```
 
-> `cache: false` above specifies that `summary.json` is not tracked or
-> <abbr>cached</abbr> by DVC (`-M` option of `dvc run`). These metrics files are
-> normally committed with Git instead. See `dvc.yaml` for more information on
-> the file format above.
```
jorgeorpinel, line 69: Looks like this note should stay? It refers to the code block above, which is left in the ref.

jorgeorpinel: Ah never mind I see we rewrote this as a regular p.

content/docs/command-reference/params/diff.md
```diff
@@ -1,7 +1,7 @@
 # params diff
 
-Show changes in [parameter dependencies](/doc/command-reference/params) between
-commits in the <abbr>DVC repository</abbr>, or between a commit and the
+Show changes in [parameter dependencies](/doc/user-guide/concepts/parameters)
```
jorgeorpinel, line 3: We should use tooltips instead of links to `/doc/user-guide/concepts/parameters` (in most places, depends on the context) now.

content/docs/command-reference/params/diff.md
```diff
@@ -23,7 +23,7 @@ in the repository history. Requires that Git is being used to version the
 project params.
 
 > Parameter dependencies are defined with the `-p` option in `dvc run`. See also
-> `dvc params`.
+> [parameters](/doc/user-guide/concepts/parameters).
```
jorgeorpinel, line 26: Same

content/docs/command-reference/params/diff.md
```diff
@@ -52,7 +52,7 @@ itself does not ascribe any specific meaning for these values.
 ## Examples
 
 Let's create a simple YAML parameters file named `params.yaml` (default params
-file name, see `dvc params` to learn more):
+file name, see [parameters](/doc/user-guide/concepts/parameters) to learn more):
```
jorgeorpinel, line 55: Same

content/docs/command-reference/push.md
```diff
@@ -185,7 +185,7 @@ Finally, we used `dvc status` to double check that all data had been uploaded.
 ## Example: What happens in the cache?
 
 Let's take a detailed look at what happens to the
-[cache directory](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory)
+[cache directory](/doc/user-guide/concepts/dvc-cache#structure-of-the-cache-directory)
```
jorgeorpinel, line 188: Kind of unrelated to the PR but we probably don't need this link since it's also in a block quote by the end of the example. But maybe could be a tooltip? If there aren't other tooltips to this concept nearby already.

content/docs/command-reference/run.md
```diff
@@ -193,8 +193,8 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
   on, from a parameters file. This is done by sending a comma separated list as
   argument, e.g. `-p learning_rate,epochs`. The default parameters file name is
   `params.yaml`, but this can be redefined with a prefix in the argument sent to
-  this option, e.g. `-p parse_params.yaml:threshold`. See `dvc params` to learn
-  more about parameters.
+  this option, e.g. `-p parse_params.yaml:threshold`. See
+  [parameters](/doc/user-guide/concepts/parameters) to learn more.
```
jorgeorpinel, line 197: > Right, sometimes an auto-link to the cmd ref may be most appropriate.

jorgeorpinel: But actually in this case the default params file name info has been moved to the concepts page so a link to concepts page would be better here... (for now, not sure where I'll keep all that detailed params behavior info).

jorgeorpinel: And there's another 2 instances of this change down this dic.

jorgeorpinel: doc

content/docs/user-guide/concepts/data-pipelines.md
```diff
@@ -0,0 +1,37 @@
+---
+name: 'Data Pipelines'
+match: ['data pipeline', 'pipeline', 'pipelines']
+tooltip: >-
+  In DVC, a [data pipeline](/doc/user-guide/concepts/data-pipelines) is a series
+  of data processing stages (for example, console commands that take an input
+  and produce an output). A pipeline may produce intermediate data, and has a
+  final result.
+description: >-
```
jorgeorpinel, line 9: What about SEO titles?

content/docs/user-guide/concepts/data-pipelines.md
```diff
@@ -0,0 +1,37 @@
+---
+name: 'Data Pipelines'
+match: ['data pipeline', 'pipeline', 'pipelines']
+tooltip: >-
+  In DVC, a [data pipeline](/doc/user-guide/concepts/data-pipelines) is a series
+  of data processing stages (for example, console commands that take an input
+  and produce an output). A pipeline may produce intermediate data, and has a
+  final result.
```
jorgeorpinel, line 8:
```suggestion
  A [data pipeline](/doc/user-guide/concepts/data-pipelines) is a series of data
  processing stages, chained by their outputs and inputs. They use some initial
  data, may produce intermediate artifacts, and reach a final result.
```

content/docs/user-guide/concepts/data-pipelines.md
```diff
@@ -0,0 +1,36 @@
+---
+name: 'Data Pipelines'
+match: ['data pipeline', 'pipeline', 'pipelines']
+tooltip: >-
+  A [data pipeline](/doc/user-guide/concepts/data-pipelines) is a series of data
+  processing stages, chained by their outputs and inputs. They use some initial
+  data, may produce intermediate artifacts, and reach a final result.
+description: >-
+  In DVC, a data pipeline is a series of data processing stages. A pipeline may
+  produce intermediate data, and has a final result.
```
jorgeorpinel, line 10: - No need for "In DVC" I think
- Maybe use "They use some initial data, may produce intermediate artifacts, and reach a final result" (text I used in the tooltip)
- What key terms are we emphasizing?
- Has there been a SEO research on existing terms (per the GSoD project proposal)? E.g. I suspect "workfow" could be a good term to incorporate here (not just in the SEO desc but in the content).

jorgeorpinel (changes requested):
Looks like SEO metadata is missing from the dependency, dvc-cache, dvc-metafiles, output concepts (there are no changes to those files). All the frontmatter in fact (inc. draft tooltip)

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -250,7 +250,7 @@ Full <abbr>parameters</abbr> (key and value) are listed separately under
   hand or with the command `dvc config --local`.
 
 - `.dvc/cache`: The <abbr>cache</abbr> directory will store your data in a
-  special [structure](#structure-of-the-cache-directory). The data files and
+  special [structure](/doc/user-guide/concepts/dvc-cache). The data files and
```
jorgeorpinel, line 253: Missing anchor 🙂

content/docs/user-guide/large-dataset-optimization.md
```diff
@@ -4,7 +4,7 @@ In order to track the data files and directories added with `dvc add` or
 `dvc run`, DVC moves all these files to the <abbr>cache</abbr>. A
 <abbr>project</abbr>'s cache is the hidden storage (by default located in
 `.dvc/cache`) for files that are tracked by DVC, and their different versions.
-(See `dvc cache` and
+(See [DVC cache](/doc/user-guide/concepts/dvc-cache) and
```
jorgeorpinel, line 7:
```suggestion
(See [DVC Cache](/doc/user-guide/concepts/dvc-cache) and
```

jorgeorpinel:
> > What's the best way (technically) to do that from here
>
> First I'd merge the latest master...

You know what @jeremydesroches I can split this PR when ready. Let's focus on addressing feedback from your changes so far and especially on the SEO stuff. Ad each part seems ready I'll open parallel PRs with the actual intention to close them and ask Ivan to review those. Thanks

content/docs/user-guide/concepts/dvc-cache.md
```diff
@@ -0,0 +1,87 @@
+---
+name: 'DVC Cache'
+match: ['DVC cache', 'cache', 'caches', 'cached']
+tooltip: >-
+  The [DVC cache](/doc/user-guide/concepts/dvc-cache) is a hidden storage (by
+  default located in the `.dvc/cache` directory) for files that are tracked by
+  DVC, and their different versions.
+description: >-
+  The DVC cache adds a layer of indirection between code and data to efficiently
+  version large datasets, data science features, and machine learning models.
```
jorgeorpinel, line 10: Same general Qs as with data-pipelines: what terms did the SEO/search history research throw for this one?

So far all the descriptions seem like excerpts from previous content we had. I was looking for some new stuff based on SEO research.

Can we list the terms/goals for each concept's SEO desc? Thanks

content/docs/user-guide/concepts/dvc-metafiles.md
```diff
@@ -0,0 +1,37 @@
+---
+name: 'DVC Metafiles'
+match: ['DVC files', 'files', 'directories']
+tooltip: >-
+  'DVC [metafiles](/doc/user-guide/concepts/dvc-metafiles) tooltip...'
+---
```
jorgeorpinel, line 6: p.s. You can list them as comments in each file change if you'd like @jeremydesroches, seems like the easiest approach to me.

p.p.s. This one (dvc-metafiles) doesn't have SEO metadata yet.

content/docs/user-guide/concepts/dvc-cache.md
jeremydesroches: > Looks like SEO metadata is missing from the dependency, dvc-cache, dvc-metafiles, output concepts (there are no changes to those files). All the frontmatter in fact (inc. draft tooltip)

Not following you here... dependency and output are glossary terms but aren't in the list of concepts. For dvc-cache, it is a full concept but it does have the frontmatter in the latest. 🤔  So I was only going to add frontmatter to dvc-metafiles. Is that different from what you were thinking @jorgeorpinel ?

jeremydesroches:
> Looks like SEO metadata is missing from the dependency, dvc-cache, dvc-metafiles, output concepts (there are no changes to those files). All the frontmatter in fact (inc. draft tooltip)

Not following you here... dependency and output are glossary terms but aren't in the list of concepts. For dvc-cache, it is a full concept but it does have the frontmatter in the latest. 🤔  So I was only going to add frontmatter to dvc-metafiles. Is that different from what you were thinking @jorgeorpinel ?

content/docs/user-guide/concepts/data-pipelines.md
jeremydesroches: > What about SEO titles?

In our previous discussion we discussed adding the "Concept:" prefix and the decision was to leave it out. That would be the only change I'd make so otherwise the generated titles are OK.

content/docs/user-guide/dvc-files-and-directories.md
jeremydesroches: Fixed. Thanks!

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: > In our previous discussion we discussed adding the "Concept:" prefix and the decision was to leave it out. 

Yes. But I thought we add `titles` everywhere as a "strong suggestion" to Google on what to use as title, since it tends to decide the title on it's own anyway 😆 — you call.

content/docs/user-guide/concepts/dvc-cache.md
jorgeorpinel: Ping @jeremydesroches

content/docs/user-guide/concepts/dvc-metafiles.md
jorgeorpinel: Ping @jeremydesroches

jorgeorpinel:
P.s. I pinged you in the few SEO-related questions left here, Jeremy. I can take care of the rest after that.

jorgeorpinel:
Ping @jeremydesroches if you no longer have the capacity to finish the SEO-related parts (see conversations above) just lmk please. Thanks.

jeremydesroches:
> Ping @jeremydesroches if you no longer have the capacity to finish the SEO-related parts (see conversations above) just lmk please. Thanks.

Sorry for the delay @jorgeorpinel. I'll get back to you on these items and most of the others today.

content/docs/user-guide/concepts/data-pipelines.md
jeremydesroches: Our discussions on metadata have mostly been about the `description`, which does need to be specified because otherwise it defaults to the same copy for every URL.

For the `title` though, if we don't add it in frontmatter it's generated from the label in `sidebar.json`. So we are giving Google a "strong suggestion" for the `title` on each doc URL already. 

My thinking was to only add the title in frontmatter if it needs to be different from the navigation label (to add a prefix, for example). Along the lines of "don't repeat yourself" — but I don't have a problem with being explicit about it either. WDYT?

content/docs/command-reference/params/diff.md
jeremydesroches: OK, good idea. Updated these ones (even though I know we'll be splitting this up).

content/docs/command-reference/params/diff.md
jeremydesroches: Done

content/docs/command-reference/params/diff.md
jeremydesroches: Done

content/docs/command-reference/push.md
jeremydesroches: Agreed. Added the tooltip here and added `cache directory` to the available matches in the concept.

There is another `cache` tooltip in the next paragraph. Too many?

content/docs/command-reference/run.md
jeremydesroches: Replaced these three with tooltips, plus another one in the description. (There are also similar references for Metrics and Plots that may need similar treatment.)

content/docs/user-guide/concepts/data-pipelines.md
jeremydesroches: Updated the description with your suggestions. 

The existing terms to focus on are listed in the HTML comment, and I've added some more around "workflow" that I verified in search console (good idea).

Terms in the description have no direct impact on ranking, but the clickthrough ratio does. So for the description we should be doing our best to improve clickthrough (looking at terms with most potential). Based on DVC traffic, it's for sure "data pipelines", and the ML, data science, *ops terms listed are speculative based on related terms and searches.

Terms in the content do have a direct impact on the ranking, so that's where working in the keywords is most worthwhile.

jeremydesroches: > So for the description we should be doing our best to improve clickthrough (looking at terms with most potential). 

An idea for how to improve CTR: Since this is a concept the first sentence of the description is good (definition of a pipeline). For the second sentence (~100 characters), we could use one or two key terms to describe uses for pipelines in DVC. Something like "Use DVC to build a machine learning pipeline or data science workflow."

> Terms in the content do have a direct impact on the ranking, so that's where working in the keywords is most worthwhile.

To further highlight the difference in approach: "workflow" comes up in a few different cases, so to improve rank beyond adding keywords, we might adjust or create an example to describe "building a machine learning (data science) workflow using pipelines in DVC". The better we can answer the query the more likely it will be rewarded. (Keeping in mind that we have to choose which queries are worth the effort.)

content/docs/user-guide/concepts/dvc-cache.md
jeremydesroches: > I was looking for some new stuff based on SEO research.

> Can we list the terms/goals for each concept's SEO desc?

I added the last half of the description based on the terms, which are listed in the HTML comments.

content/docs/user-guide/concepts/dvc-metafiles.md
jeremydesroches: Based on your [comment](https://github.com/iterative/dvc.org/pull/1944#discussion_r536392345) way up there ☝️ it seems like you are saying you don't want to move the Files content or create this item at all. Is that right @jorgeorpinel? If I'm mistaken on this, not a problem to research and create the metadata.

Otherwise, I think I've addressed the remaining feedback so we can start splitting up this PR to work on the pieces in order as you suggested.

content/docs/command-reference/params/diff.md
jorgeorpinel: > Updated these ones (even though I know we'll be splitting this up

Any progress here will be preserved when I can finally split this so thanks.

content/docs/command-reference/push.md
jorgeorpinel: I'd just use `<abbr>cache</abbr> directory` and avoid the extra glossary match entry.

> There is another cache tooltip in the next paragraph. Too many?

Yep, I'd remove that later one. Ugh managing `<abbr>` tags can be painful.

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: > if we don't add it in frontmatter it's generated from the label in sidebar.json

Ah OK then, I didn't remember that. Thanks

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: Awesome, thanks for the tips!

content/docs/user-guide/concepts/dvc-cache.md
jorgeorpinel: I see. Alright.

content/docs/user-guide/concepts/dvc-metafiles.md
jorgeorpinel: You're right but now we just need to revert moving content from https://dvc.org/doc/user-guide/dvc-files-and-directories to here. I guess I can just leave it out when I split this PR, so 👍

jorgeorpinel:
No worries @jeremydesroches. I'm going to have low capacity for a few weeks anyway but I'll come back to splitting this ASAP. It looks like so far you've completed your contributions so thanks a bunch! unfortunately I have to close this PR but your commits will be preserved in following PRs which I'll link below for the record (later). Closing for now.

PR #1945, opened 2020-11-14
Misc. updates

jorgeorpinel:
Copy edits stashed from recent PRs


content/docs/command-reference/add.md
```diff
@@ -144,9 +143,12 @@ directory symlinks.
 - `--external` - allow `targets` that are outside of the DVC repository. See
   [Managing External Data](/doc/user-guide/managing-external-data).
 
-- `--glob` - allows adding files and directories that match the specified
-  pattern as specified by `target`. Shell-style wildcards are supported: `*`,
-  `?`, `[seq]`, `[!seq]`, and `**`.
+  > Note that this option implies `--no-commit`, as external outputs are never
```
shcheklein, line 146: hmm, not sure this is true/relevant here

content/docs/command-reference/add.md
```diff
@@ -144,9 +143,12 @@ directory symlinks.
...
+  > pushed or pulled from/to remote storage. See link above for more details.
+
+- `--glob` - allows adding files and directories that match the
```
shcheklein, line 149: also, looks like `--glob` is out of order with the other options?

content/docs/command-reference/add.md
jorgeorpinel: This was added after noticing people don't realize --external implies --no-commit but I don't have the links to those conversations right now. I only remember the last one (somewhere on Discord): basically a user asking how to backup external outputs.

So when(/if?) the import --store option gets merged, this will be revisited anyway. Up to you. I think that if it's not wrong, we should err on the side of TMI (too much info) in cmd refs.

content/docs/command-reference/add.md
jorgeorpinel: I guess it can go before --external? I'm not sure which one we expect people to use more. I moved it up one place for now.

content/docs/command-reference/add.md
shcheklein: It looks wrong to me that the biggest issue. Unless I'm missing something.

content/docs/command-reference/add.md
shcheklein: it should be in the same order as DVC CLI has it in, right?

jorgeorpinel: Not necessarily. We use an arbitrary order here. The core team tends to always append options at the end. There's no consistency so far 🙁.

There's a discussion about this in iterative/dvc/issues/3642.

content/docs/command-reference/add.md
jorgeorpinel: It's probably not a big issue. Thus it's a block quote note. So, should we remove it?

content/docs/command-reference/add.md
shcheklein: kk

content/docs/command-reference/add.md
shcheklein: It's kinda an issue if this information is wrong. I'm fine with both - remove and/or clarify and confirm.

jorgeorpinel: I guess you're right, external outputs can be cached, but externally. Is that what you meant? Sorry, just getting it now... Updated to:

  > Note that external outputs require an external cache setup (unless
  > `--no-commit` is used), and are never pushed or pulled from/to remote
  > storage. See link above for more details.

content/docs/command-reference/add.md
```diff
@@ -141,12 +140,17 @@ directory symlinks.
...
+  Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`, and `**`
+
 - `--external` - allow `targets` that are outside of the DVC repository. See
   [Managing External Data](/doc/user-guide/managing-external-data).
 
-- `--glob` - allows adding files and directories that match the specified
-  pattern as specified by `target`. Shell-style wildcards are supported: `*`,
-  `?`, `[seq]`, `[!seq]`, and `**`.
+  > Note that external outputs require an external cache setup (unless
+  > `--no-commit` is used), and are never pushed or pulled from/to
```
shcheklein, line 151: is there a reasonable workflow with `--no-commit` + `--external`?

jorgeorpinel: I think so, especially since we mention "For cached external outputs..." in https://dvc.org/doc/user-guide/managing-external-data

For example if you want to avoid the external data from being moved and linked back to its position for other processes that also use it which may not support handling links.

content/docs/command-reference/gc.md
```diff
@@ -93,8 +93,8 @@ The default remote is cleaned (see `dvc config core.remote`) unless the
   from remote storage. This only applies when the `--cloud` option is used, or a
   `--remote` is given. The default value is `4 * cpu_count()`. For SSH remotes,
   the default is `4`. Note that the default value can be set using the `jobs`
-  config option with `dvc remote modify`. Using more jobs may improve the
-  overall connection speed.
+  config option with `dvc remote modify`. Using more jobs may speed up the
+  operation.
```
jorgeorpinel, line 97: Here

content/docs/command-reference/status.md
```diff
@@ -143,8 +143,8 @@ that.
   information from remote storage. This only applies when the `--cloud` option
   is used, or a `--remote` is given. The default value is `4 * cpu_count()`. For
   SSH remotes, the default is `4`. Note that the default value can be set using
-  the `jobs` config option with `dvc remote modify`. Using more jobs may improve
-  the overall connection speed.
+  the `jobs` config option with `dvc remote modify`. Using more jobs may speed
+  up the operation.
```
jorgeorpinel, line 147: And here

jorgeorpinel (commented):
Note changes per https://github.com/iterative/dvc.org/pull/1941#discussion_r524541298:

content/docs/command-reference/add.md
shcheklein: okay, we don't move data, but we also don't calculate any hashes in this case, right? what would the next step for this `.dvc` file?

jorgeorpinel: I just tried this and it does calculate the hash. The data is just not moved to the cache. (At least using an external data file in the local system)

jorgeorpinel: And fortunately the --no-commit option description doesn't state that the has isn't calculated.

Unrelated: It does however mention the state file (fixed the path in 2ab35cc2) which is probably completely out of context/unnecessary to mention here but oh well...

shcheklein: `I just tried this and it does calculate the hash` - hmm looks like a bug to me. Local `--no-commit` avoids calculating hash for outputs as far as I remember. Let me check.

jorgeorpinel: It's calculating it even for regular outputs in the workspace for me (DVC 1.9.1 on Win).

shcheklein: okya, you are right on this (my usual confusion since --no-commit is not an opposite to `dvc commit` unfortunately)

it calculates hashes

```
√ Projects/test-dvc % dvc add --external --no-commit s3://dvc-temp/ivan/images
Adding...
ERROR: Current operation was unsuccessful because 's3://dvc-temp/ivan/images' requires existing cache on 's3' remote. See <https://man.dvc.org/config#cache> for information on how to set up remote cache.
```

it is still not correct it seems.

To be honest I doubt that this note could be understood by anyone :) I would keep this information to the Managing external data section.

shcheklein: > It does however mention the state file

it is irrelevant, agreed

jorgeorpinel: > I would keep this information to the Managing external data section
> it (state file) is irrelevant, agreed

Ok, updated both things in 7bad093e. Just left the note about the external cache, which still seems pretty relevant IMO.

jorgeorpinel: PTAL

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,13 +34,17 @@ Currently, the following types (protocols) of external outputs (and
...
 
-> Note that these are a subset of the remote storage types supported by
+> Note that these are only a subset of the remote storage types supported by
 > `dvc remote`.
 
-> Avoid using the same [DVC remote](/doc/command-reference/remote) (used for
-> `dvc push`, `dvc pull`, etc.) for external outputs, because it may cause file
-> hash overlaps: the hash of an external output could collide with a hash
-> generated locally for another file with different content.
+> 💡 Note that external outputs are never pushed or pulled from/to
```
shcheklein, line 40: Number of notes is getting out of hands. If we consider this important, let's improve the explanation how they work?

content/docs/user-guide/managing-external-data.md
```diff
@@ -19,12 +19,13 @@ outputs for `dvc.yaml` files (only `outs` field, not metrics or plots).
 External outputs are considered part of the (extended) DVC project: DVC will
 track changes in them, and reflect this in `dvc status` reports, for example.
 
-For cached external outputs (e.g. `dvc add`, `dvc run -o`), you will need to
-[setup an external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
-in the same external/remote file system first.
+Normally, external outputs (e.g. `dvc add`, `dvc run -o`), require an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+in the same external/remote file. The exception is when they are not
+<abbr>cached</abbr> (`--no-commit` option).
```
shcheklein, line 25: not cached and no-commit are different things (like -O and --no-commit for regular outputs) Also, as I checked `--no-commit` still requires cache to be set.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: I removed the first one (also from the external deps doc for consistency) and made the 2nd one not a block quote. So now there's just one block quote left at the bottom.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: OK... I removed the last sentence for now. It's kind of confusing/hard to explain right. Maybe we should review the behavior in the core repo first?

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,13 +34,13 @@ Currently, the following types (protocols) of external outputs (and
 - HDFS
 - Local files and directories outside the <abbr>workspace</abbr>
 
-> Note that these are a subset of the remote storage types supported by
-> `dvc remote`.
+💡 Note that external outputs are never pushed or pulled from/to
```
shcheklein, line 36: it's not about note literally. It's about text becoming a list of disconnected paragraphs. Question- why do we have this paragraph here, after the list of supported types? Can we make it organically part of the text above?

content/docs/user-guide/managing-external-data.md
shcheklein: `--no-commit` workflow is not about caching/not caching. Usually it is about deliberately _delaying_ the caching for some reason.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Hmmm, I'd prefer not to mix the previous paragraph with the concept of "remote storage". That will add confusion. I think that's why these mentions of remote storage are after the bullet list. I could possibly mix them into a more narrative paragraph (instead of descriptive notes)... ⌛

content/docs/user-guide/managing-external-data.md
jorgeorpinel: OK. Is that relevant to mention especially in this guide? Seems like a general thing for add/run (and it's already mentioned in those refs).

The reason it was mentioned here is that it is the exception to the "Normally" at the beginning of the paragraph.

content/docs/user-guide/managing-external-data.md
shcheklein: it's not about the previous paragraph literally. It's about making it part of the some text above. Probably the one about `cache` works good. It can explain why does it happen. And even note about hash collision (I hope it won't be relevant) should go probably before the list of supported remote types.

content/docs/user-guide/managing-external-data.md
shcheklein: Yep, I would avoid changing it to `Normally, ` and then we don't need mentioning this.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Yes, you didn't even mention the previous paragraph. I was also trying to refer to all the text before the list. I think mentioning remote storage in most places above may be confusing but let me try ⌛

content/docs/user-guide/managing-external-data.md
jorgeorpinel: OK, I'm removing "Normally", but then this isn't always true. But maybe it will be fixed by the changes per our other discussion here ⌛

shcheklein: They do always require the external cache as we checked, right? Also, I don't see a good use case w/o cache.

jorgeorpinel: Yes they do. Which seems like a bug to me, but if there's no use case for the combination anyway then I guess this is not an issue. Resolving!

content/docs/user-guide/managing-external-data.md
jorgeorpinel: OK I mostly managed to do this, while also improving the explanations a little, I think. But I did have to leave one block quote note under the list of remote types (in the ext deps doc). PTAL @shcheklein

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,25 +29,26 @@ Currently, the following types (protocols) of external outputs (and
...
 
-> Note that these are a subset of the remote storage types supported by
-> `dvc remote`.
+> Note [remote storage](/doc/command-reference/remote) is a separate feature,
+> and that external outputs are not pushed or pulled from/to DVC remotes.
 
-> Avoid using the same [DVC remote](/doc/command-reference/remote) (used for
-> `dvc push`, `dvc pull`, etc.) for external outputs, because it may cause file
-> hash overlaps: the hash of an external output could collide with a hash
-> generated locally for another file with different content.
```
jorgeorpinel, line 42: Note that I totally removed this note. How likely is this scenario really? And if we're worried about it, should DVC core handle it as an error, at least?

shcheklein: It's hard for DVC to handle this. It will change a bit in DVC 2.0, so it might be safe to remove it. On the other hand, it's a nature of clouds- they generate some hash that might differ from the one we generate with DVC. And there is no any guarantees. The best we can do is to have per cloud instruction- is it safe or not mixing those things.

content/docs/user-guide/managing-external-data.md
```diff
@@ -13,18 +13,14 @@ track data outside of the <abbr>project</abbr>.
 ## How external outputs work
 
 DVC can track existing files or directories on an external location with
-`dvc add` (`out` field). It can also create external files or directories as
-outputs for `dvc.yaml` files (only `outs` field, not metrics or plots).
+`dvc add`. It can also define external outputs for `dvc.yaml` stages to create.
 
 External outputs are considered part of the (extended) DVC project: DVC will
-track changes in them, and reflect this in `dvc status` reports, for example.
+detect when they change, reporting this in `dvc status` for example.
```
shcheklein, line 19: it's not only about detecting changes, it's about versioning them (that's where we use "track")

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,25 +29,26 @@ Currently, the following types (protocols) of external outputs (and
...
-> Note that these are a subset of the remote storage types supported by
-> `dvc remote`.
+> Note [remote storage](/doc/command-reference/remote) is a separate feature,
+> and that external outputs are not pushed or pulled from/to DVC remotes.
 
-> Avoid using the same [DVC remote](/doc/command-reference/remote) (used for
-> `dvc push`, `dvc pull`, etc.) for external outputs, because it may cause file
-> hash overlaps: the hash of an external output could collide with a hash
-> generated locally for another file with different content.
+Importantly, external outputs require an
```
shcheklein, line 35: I'm a bit lost here ... I think the previous version was pretty precise, why change it?

content/docs/user-guide/managing-external-data.md
efiop: I would keep it for now. Don't see a good reason to remove it.

jorgeorpinel: > Don't see a good reason to remove it.

Doing so per https://github.com/iterative/dvc.org/pull/1945#pullrequestreview-531982134 and https://github.com/iterative/dvc.org/pull/1945#pullrequestreview-531992838 above. And since it seems to me like an extremely rare edge case. And if it will change in 2.0 we can think about what to say about it then.

jorgeorpinel: > It's hard for DVC to handle this.

BTW if the problem is "hash overlaps: the hash of an external output could collide with a hash generated locally for another file" that sounds like a very precise condition that could be caught/handled. But I'm not sure.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: > why change it?

The previous version was:

> For cached external outputs (e.g. dvc add, dvc run -o), you will need to setup an external cache in the same external/remote file system first.

We decided to not mentioned "cached outs" here since there's no real use case for non cached external outs. The mention of `add` and `run -o` also don't seem that important. So most of that sentence had to go anyway. Then we decided to incorporate the info involving remotes to the paragraphs before the bullet list so pretty much all the text changed. And it was a good thing, I think, because it explains what ext outs are first, emphasizes in dvc.yaml, and how their URLs are formed (same as remotes).

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Good point. But "track when they change" doesn't sound right... I went for "track them for [versioning], thus detecting when they change".

content/docs/user-guide/managing-external-data.md
jorgeorpinel: I removed "Importantly" now, and moved this paragraph before the note about remotes being a different feature (right after the bullet list). I think the structure of this section makes sense: 1 what are ext outs, 2 how to define them, 3 note about external cache.

content/docs/user-guide/managing-external-data.md
efiop: @jorgeorpinel There is no guarantee it will go away in 2.0. This note discourages people from misusing our incomplete ext feature, so better to not remove it for now.

jorgeorpinel: OK sure, I rolled this back (mostly) in defd99d8 @efiop.

What did you mean by "incomplete ext feature" thouogh? ext = external data? Just curious

content/docs/user-guide/external-dependencies.md
```diff
@@ -1,22 +1,23 @@
 # External Dependencies
 
 There are cases when data is so large, or its processing is organized in a way
-such that you would like to avoid moving it out of its external/remote location.
-For example from a network attached storage (NAS), processing data on HDFS,
-running [Dask](https://dask.org/) via SSH, or having a script that streams data
-from S3 to process it.
+such that its preferable to avoid moving it from its original location, even if
+it's external or remote to the project. For example: data on a network attached
```
shcheklein, line 5: `even if it's external or remote to the project` - no need for this. It complicates it.

content/docs/user-guide/external-dependencies.md
```diff
@@ -1,22 +1,23 @@
...
 data outside of the <abbr>project</abbr>.
 
-## How it works
+## How external dependencies work
 
 You can specify external files or directories as dependencies for your pipeline
-stages. DVC will track changes in them and reflect this in the output of
-`dvc status`.
+[stages](/doc/command-reference/run). DVC will track changes in them and reflect
+this in the output of `dvc status`.
```
shcheklein, line 17: to be honest mentioning `dvc status` is strange here. The most important thing that it will for example detect this change on `dvc repro`.

content/docs/user-guide/external-dependencies.md
```diff
@@ -27,20 +28,19 @@ supported:
 - HTTP
 - Local files and directories outside the <abbr>workspace</abbr>
 
-> Note that these are a subset of the remote storage types supported by
-> `dvc remote`.
-
-In order to specify an external <abbr>dependency</abbr> for your stage, use the
-usual `-d` option in `dvc run` with the external path or URL to your desired
-file or directory.
+> Note [remote storage](/doc/command-reference/remote) is a separate feature.
```
shcheklein, line 31: why is it needed?

content/docs/user-guide/external-dependencies.md
```diff
@@ -27,20 +28,19 @@ supported:
...
 
 ## Examples
 
-Let's take a look at a `download_file` [stage](/doc/command-reference/run) that
-simply downloads a file from an external location.
+To define an external <abbr>dependency</abbr>, add the external URL or path to
```
shcheklein, line 35: let's simplify this ... reading two paragraphs of text as intro to an example looks wrong

content/docs/user-guide/external-dependencies.md
```diff
@@ -27,20 +28,19 @@ supported:
...
+the `deps` field of `dvc.yaml`. For example, with the usual `-d` option in
+`dvc run`, giving it the external URL/path to your desired file or directory.
+
+Let's take a look at defining and running a `download_file` stage that simply
+downloads a file from an external location, on all the supported location types.
 
-> Note that some of these commands use the `/home/shared` directory, typical in
-> Linux distributions.
+> Note that some of the example commands below use the `/home/shared` directory,
```
shcheklein, line 42: this note is also not needed here I think

content/docs/user-guide/external-dependencies.md
shcheklein: I think it was fine to keep in the previous paragraph?

content/docs/user-guide/managing-external-data.md
```diff
@@ -1,10 +1,10 @@
 # Managing External Data
 
 There are cases when data is so large, or its processing is organized in a way
-such that its preferable to avoid moving it from its external/remote location.
-For example data on a network attached storage (NAS), processing data on HDFS,
-running [Dask](https://dask.org/) via SSH, or having a script that streams data
-from S3 to process it.
+such that its preferable to avoid moving it from its original location, even if
+it's external or remote to the project. For example: data on a network attached
```
shcheklein, line 5: same here

content/docs/user-guide/managing-external-data.md
```diff
@@ -13,18 +13,16 @@ track data outside of the <abbr>project</abbr>.
 ## How external outputs work
 
 DVC can track existing files or directories on an external location with
-`dvc add` (`out` field). It can also create external files or directories as
-outputs for `dvc.yaml` files (only `outs` field, not metrics or plots).
+`dvc add`. It can also define external outputs for `dvc.yaml` stages to create.
```
shcheklein, line 16: DVC pipelines stages?

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,25 +31,32 @@ Currently, the following types (protocols) of external outputs (and
 - HDFS
 - Local files and directories outside the <abbr>workspace</abbr>
 
-> Note that these are a subset of the remote storage types supported by
-> `dvc remote`.
+External outputs require an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+in the same external/remote file.
+
+> Note that [remote storage](/doc/command-reference/remote) is a separate
```
shcheklein, line 38: same here, why is it needed?

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,25 +31,32 @@ Currently, the following types (protocols) of external outputs (and
...
 
 ## Examples
 
-Let's take a look at:
+Let's take a look at the following operations on all the supported location
+types:
 
-1. Adding a `dvc remote` to use as cache for data in the external location, and
+1. Adding a `dvc remote` in the same location as the desired outputs, and
    configure it as external <abbr>cache</abbr> with `dvc config`.
```
shcheklein, line 53: as an external cache?

content/docs/user-guide/managing-external-data.md
```diff
@@ -33,25 +31,32 @@ Currently, the following types (protocols) of external outputs (and
...
-2. Tracking existing data on an external location with `dvc add` (this doesn't
-   download it). This produces a `.dvc` file with an external output.
-3. Creating a simple [stage](/doc/command-reference/run) that moves a local file
-   to the external location. This produces a stage with another external output
-   in `dvc.yaml`.
+2. Tracking existing data on the location using `dvc add` (`--external` option
```
shcheklein, line 54: on the location - sounds strange a bit

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Removed in 8a6892dd.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Good point. I changed these texts in dd1501bb for both external docs since describing what deps/outs are isn't a goal here.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Because we mention remotes before so it may be confusing. Is a short note by the end of the section too much?

content/docs/user-guide/external-dependencies.md
jorgeorpinel: > it was fine to keep in the previous paragraph

Yes, going back to that (almost) now that the mechanism is explained a little more in the previous section.

Done in de67413d

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Gone.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Not seeing need to mention pipelines in these docs, we're already going over deps/outs and stages. Too many concepts otherwise, probably? You can click on "stage" to get to pipelines.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: No "an" needed. Maybe "the" but it's OK like this too, I think.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Yes. Going for "on the external location". Thanks

content/docs/user-guide/external-dependencies.md
```diff
@@ -1,23 +1,27 @@
...
-running [Dask](https://dask.org/) via SSH, or having a script that streams data
+There are cases when data is so large, or its processing is organized in such a
+way, that its preferable to avoid moving it from its original location. For
+example data on a network attached storage (NAS), processing data on HDFS,
+running [Dask](https://dask.org/) via SSH, or for a script that streams data
 from S3 to process it.
 
-External <abbr>dependencies</abbr> and
+External dependencies and
 [external outputs](/doc/user-guide/managing-external-data) provide ways to track
```
shcheklein, line 10: track and version?

content/docs/user-guide/external-dependencies.md
```diff
@@ -1,23 +1,27 @@
...
-You can specify external files or directories as dependencies for your pipeline
-stages. DVC will track changes in them and reflect this in the output of
-`dvc status`.
+External <abbr>dependencies</abbr> are considered part of the (extended) DVC
+project: DVC will track them, detecting when they change (triggering stage
+executions on `dvc repro`, for example).
 
-Currently, the following types (protocols) of external dependencies are
-supported:
+DVC can track files or directories on an external location as
```
shcheklein, line 19: sounds like it repeats a lot of the previous paragraph?

content/docs/user-guide/external-dependencies.md
jorgeorpinel: Do we version dependencies? Only when they're outputs of a previous stage, I think?

jorgeorpinel: I guess we do keep track of their versions in any case, but don't really control those versions if they are external.

shcheklein (approved):
looks good to me, a few minor questions

content/docs/user-guide/external-dependencies.md
shcheklein: This sentence is about `External dependencies and [external outputs]`

content/docs/user-guide/external-dependencies.md
jorgeorpinel: True 🤦. Updated in cfe6e07a.

jorgeorpinel:
Merging this! (I have lots of staged regular updates 😅) but please follow up on any discussions above @shcheklein, and I'll address them in a coming up PR.

content/docs/user-guide/external-dependencies.md
jorgeorpinel: True. Updating in 60e8055d.

PR #1382, opened 2020-05-31
Regular updates & plots 1.0 update

jorgeorpinel:
- Misc. updates
- Improve structure of `dvc run` cmd ref.
- Copy edits
- finish updating `dvc metrics/plots` for 1.0 release, started in #1360
- "working tree" -> "workspace" per iterative/dvc/pull/3914

UPDATE: Jump to ~~https://github.com/iterative/dvc.org/pull/1382#issuecomment-639989366~~ https://github.com/iterative/dvc.org/pull/1382#issuecomment-640435301 for pending questions/items.

PR #1946, opened 2020-11-15
cmd: example on modifying an outputs into a plots

imhardikj:
Partially Addresses #1414

- [x] term: "column index" vs "column number"? See #1382 (comment) and #1382 (review)
- [x] modify: how to make a regular output into a plot? See iterative/dvc#3994 (review)

### DIFF
--- a/content/docs/command-reference/plots/diff.md
+++ b/content/docs/command-reference/plots/diff.md
@@ -77,12 +77,13 @@ all the current plots, without comparisons.
 - `-x <field>` - field name from which the X axis data comes from. An
   auto-generated `index` field is used by default. See
   [Custom templates](/doc/command-reference/plots#custom-templates) for more
-  information on this `index` field. Column names or numbers are expected for
-  tabular metrics files.
+  information on this `index` field. Column names or indices (when `--no-header`
+  option is specified) are expected for tabular metrics files.
 
 - `-y <field>` - field name from which the Y axis data comes from. The last
-  field found in the `--targets` is used by default. Column names or numbers are
-  expected for tabular metrics files.
+  field found in the `--targets` is used by default. Column names or indices
+  (when `--no-header` option is specified) are expected for tabular metrics
+  files.
 
 - `--x-label <text>` - X axis label. The X field name is the default.
 
--- a/content/docs/command-reference/plots/diff.md
+++ b/content/docs/command-reference/plots/diff.md
@@ -77,13 +77,12 @@ all the current plots, without comparisons.
 - `-x <field>` - field name from which the X axis data comes from. An
   auto-generated `index` field is used by default. See
   [Custom templates](/doc/command-reference/plots#custom-templates) for more
-  information on this `index` field. Column names or indices (when `--no-header`
-  option is specified) are expected for tabular metrics files.
+  information on this `index` field. Column names or numbers are expected for
+  tabular metrics files.
 
 - `-y <field>` - field name from which the Y axis data comes from. The last
-  field found in the `--targets` is used by default. Column names or indices
-  (when `--no-header` option is specified) are expected for tabular metrics
-  files.
+  field found in the `--targets` is used by default. Column names or numbers are
+  expected for tabular metrics files.
 
 - `--x-label <text>` - X axis label. The X field name is the default.
 
--- a/content/docs/command-reference/plots/modify.md
+++ b/content/docs/command-reference/plots/modify.md
@@ -26,6 +26,10 @@ The path to the metrics file `target` is required. It must be listed in a
 `dvc.yaml` file (see the `--plots` option of `dvc run`). `dvc plots modify` adds
 the display properties to `dvc.yaml`.
 
+Convert any
+[output or simple metrics file into a plot](#example-convert-any-output-into-a-plot)
+by passing it as a `target` to `dvc plot modify`.
+
 Property names are passed as [options](#options) to this command (prefixed with
 `--`). These are based on the full
 [Vega specification](https://vega.github.io/vega/docs/specification/).
@@ -35,10 +39,10 @@ Property names are passed as [options](#options) to this command (prefixed with
 - `-t <name_or_path>, --template <name_or_path>` - set a default
   [plot template](/doc/command-reference/plots#plot-templates).
 
-- `-x <field>` - set a default field or column name (or index) from which the X
+- `-x <field>` - set a default field or column name (or number) from which the X
   axis data comes from.
 
-- `-y <field>` - set a default field or column name (or index) from which the Y
+- `-y <field>` - set a default field or column name (or number) from which the Y
   axis data comes from.
 
 - `--x-label <text>` - set a default title for the X axis.
@@ -123,29 +127,6 @@ plots:
       title: Accuracy
 ```
 
-## Example: Make a regular output into a plot
-
-We can use `dvc plots modify` to make an <abbr>output</abbr> or metrics file of
-a stage into a plot. Lets take an example stage `evaluate` which has an
-<abbr>output</abbr> `logs.csv`:
-
-```yaml
-evaluate:
-  cmd: python src/evaluate.py
-    deps:
-    - src/evaluate.py
-    outs:
-    - logs.csv
-    plots:
-    - scores.json
-```
-
-To change the output file `logs.csv` into a plot run:
-
-```dvc
-$ dvc plots modify logs.csv
-```
-
 ## Example: Template change
 
 _dvc run --plots file.csv ..._ command assign the default template that needs to
@@ -153,4 +134,28 @@ be changed in many cases. A simple command changes the template:
 
 ```dvc
 $ dvc plots modify classes.csv --template confusion
+```
+
+## Example: Convert any output into a plot
+
+We can use `dvc plots modify` to make an <abbr>output</abbr> or simple metrics
+file of a stage into a plot. Lets take an example `evaluate` stage which has
+`logs.csv` as output. To change the output file `logs.csv` into a plot run:
+
+```dvc
+$ dvc plots modify logs.csv
+```
+
+Changes in `dvc.yaml` shows `logs.csv` is now a plot:
+
+```git
+evaluate:
+  cmd: python src/evaluate.py
+    deps:
+    - src/evaluate.py
+-   outs:
+-   - logs.csv
+    plots:
+    - scores.json
++   - logs.csv
 ```
--- a/content/docs/command-reference/plots/modify.md
+++ b/content/docs/command-reference/plots/modify.md
@@ -31,7 +31,7 @@ Property names are passed as [options](#options) to this command (prefixed with
 [Vega specification](https://vega.github.io/vega/docs/specification/).
 
 Note that a secondary use of this command is to convert output or simple
-[metrics file](/doc/command-reference/metrics) into a plot (see
+`dvc metrics` file into a plots file (see
 [example](#example-convert-any-output-into-a-plot)).
 
 ## Options
@@ -138,9 +138,9 @@ $ dvc plots modify classes.csv --template confusion
 
 ## Example: Convert any output into a plot
 
-Let's take an example `evaluate` stage which has `logs.csv` as an output. We use
-`dvc plots modify` to convert the `logs.csv` output file into a plot, and then
-observe changes that happened in `dvc.yaml`:
+Let's take an example `evaluate` stage which has `logs.csv` as an output. We can
+use `dvc plots modify` to convert the `logs.csv` output file into a plots file,
+and then confirm the changes that happened in `dvc.yaml`:
 
 ```dvc
 $ dvc plots modify logs.csv
--- a/content/docs/command-reference/plots/modify.md
+++ b/content/docs/command-reference/plots/modify.md
@@ -26,13 +26,13 @@ The path to the metrics file `target` is required. It must be listed in a
 `dvc.yaml` file (see the `--plots` option of `dvc run`). `dvc plots modify` adds
 the display properties to `dvc.yaml`.
 
-Convert any
-[output or simple metrics file into a plot](#example-convert-any-output-into-a-plot)
-by passing it as a `target` to `dvc plot modify`.
-
 Property names are passed as [options](#options) to this command (prefixed with
 `--`). These are based on the full
 [Vega specification](https://vega.github.io/vega/docs/specification/).
+
+Note that a secondary use of this command is to convert output or simple
+[metrics file](/doc/command-reference/metrics) into a plot (see
+[example](#example-convert-any-output-into-a-plot)).
 
 ## Options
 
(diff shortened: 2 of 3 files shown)


imhardikj:
From the description of #1414 
> Apply changes from iterative/dvc/pull/3963 once that is merged.

These changes are already applied I think(in #1382).
> Add YAML and TSV examples to plot show? Or maybe just translate JSON example to YAML and CSV to TSV.

Is this really required? 
`maybe just translate JSON example to YAML and CSV to TSV`  But this will result in same plot. Basically there will be same plot twice.
Instead can we can change file extension (in command) in  https://dvc.org/doc/command-reference/plots/show#headerless-tables to `tsv` and https://dvc.org/doc/command-reference/plots/show#example-vega-specification-file to `yaml`.
> Per #1382 (review) double check about "All plots are used" in plot show -h usage (targets arg).

I checked this with examples which has 2 and 3 plot files, `plots show` generated image with 2 and 3 plots respectively. This validates  `All plots are used` I believe.
> modify: how to make a regular output into a plot? See iterative/dvc#3994 (review)

Made a new example in this PR. https://github.com/iterative/dvc.org/pull/1946/commits/6caa73c83eed9d3cc7d0e91493aca6f6e854da68

> term: "column index" vs "column number"? See #1382 (comment) and #1382 (review)

Changes this (in this PR https://github.com/iterative/dvc.org/pull/1946/commits/71e19ea6987cfb7a8a638cf26ad6d14890137929)
> For JSON plot files we say "every object should have the same structure" but actually I'm not sure what the exact restriction is. Ask/test/check. See #1382 (review)

I checked this. For example if we have 100 rows in a JSON like this:
```
{
  "train": [
    { "accuracy": 0.96658, "loss": 0.10757 },
    { "accuracy": 0.97641, "loss": 0.07324 },
.......
```
The last column in first row is `loss`. `plots show` will generate a plot with all 100 rows irrespective of extra (or less) key-value pairs in any row from 2 to 100 BUT `loss` should be there in each row.
Now let's say if loss is missing from 56th row only. The plot generated will show continuous line from1st to 55th row and 57th and to 100th row. Somethinh like this:
download
So `every object should have the same structure` doesn't hold true.
> Double check how to apply all Vega spec props in plots: modify subcmd only has a limited no. of options.

I passed properties like this:
```dvc
$ dvc plots modify logs.csv --width 1000
```
This didn't work for any property. The template was default only. (Have to check it for custom template, might work on it.)

jorgeorpinel:
@imhardikj why are we discussing #1414 here? Let's discuss it in #1414.

content/docs/command-reference/plots/diff.md
```diff
@@ -77,12 +77,13 @@ all the current plots, without comparisons.
...
-  information on this `index` field. Column names or numbers are expected for
-  tabular metrics files.
+  information on this `index` field. Column names or indices (when `--no-header`
+  option is specified) are expected for tabular metrics files.
 
 - `-y <field>` - field name from which the Y axis data comes from. The last
-  field found in the `--targets` is used by default. Column names or numbers are
-  expected for tabular metrics files.
+  field found in the `--targets` is used by default. Column names or indices
+  (when `--no-header` option is specified) are expected for tabular metrics
```
jorgeorpinel, line 85: The problem is that the `index` field is explained right next to the mention of "column indices" so even with that parenthesis, it's easily confusing. Also, the parenthesis about `--no-header` adds to the confusion as that option is described after, not before this. Not sure what to do about this but I think "column number" is a perfectly fine term.

jorgeorpinel: Yeah per https://trends.google.com/trends/explore?geo=US&q=column%20number,column%20index I will say that we should just keep "number" here. Cc @shcheklein in case you want to override my decision here (context: https://github.com/iterative/dvc.org/pull/1382#pullrequestreview-427729775). Thanks

image

content/docs/command-reference/plots/modify.md
```diff
@@ -35,10 +35,10 @@ Property names are passed as [options](#options) to this command (prefixed with
 - `-t <name_or_path>, --template <name_or_path>` - set a default
   [plot template](/doc/command-reference/plots#plot-templates).
 
-- `-x <field>` - set a default field or column name (or number) from which the X
+- `-x <field>` - set a default field or column name (or index) from which the X
   axis data comes from.
 
-- `-y <field>` - set a default field or column name (or number) from which the Y
+- `-y <field>` - set a default field or column name (or index) from which the Y
```
jorgeorpinel, line 41: In this case "index" is nowhere else in the doc but for consistency we should also keep "number" here, probably.

content/docs/command-reference/plots/show.md
```diff
@@ -48,12 +48,12 @@ please see `dvc plots`.
...
-  information on this `index` field. Column names or numbers are expected for
-  tabular metrics files.
+  information on this `index` field. Column names or indices (when `--no-header`
+  option is specified) are expected for tabular metrics files.
 
 - `-y <field>` - field name from which the Y axis data comes from. The last
-  field found in the `targets` is used by default. Column names or numbers are
-  expected for tabular metrics files.
+  field found in the `targets` is used by default. Column names or indices (when
+  `--no-header` option is specified) are expected for tabular metrics files.
```
jorgeorpinel, line 56: Same

content/docs/command-reference/plots/modify.md
```diff
@@ -123,6 +123,29 @@ plots:
       title: Accuracy
 ```
 
+## Example: Make a regular output into a plot
```
jorgeorpinel, line 126:
```suggestion
## Example: Convert any output into a plot
```

But I'm not sure an example is the best place for this. It seems more like a subsection of the description.

jorgeorpinel: Or add a paragraph to the description mentioning this possible, use, and linking to this example. But move the example to last (after the existing one). Whatever seems to make more sense for you — remember to look at your work-in-progress in the browser with a dev environment.

content/docs/command-reference/plots/modify.md
```diff
@@ -123,6 +123,29 @@ plots:
...
+
+We can use `dvc plots modify` to make an <abbr>output</abbr> or metrics file of
```
jorgeorpinel, line 128: metrics file > simple metrics file

and link to https://dvc.org/doc/command-reference/metrics

content/docs/command-reference/plots/modify.md
```diff
@@ -123,6 +123,29 @@ plots:
...
+a stage into a plot. Lets take an example stage `evaluate` which has an
+<abbr>output</abbr> `logs.csv`:
```
jorgeorpinel, line 130: Names before nouns:
```suggestion
a stage into a plot. Lets take an example `evaluate` stage, which has an
`logs.csv` output:
```

jorgeorpinel: (also, note there's no need for 2 tooltips of the same term in the same paragraph, again, please always look at your work in browser to detect these issues)

content/docs/command-reference/plots/diff.md
shcheklein: Context is important here:

https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.index.html

`number` is not used,  I guess, since it's hard to distinguish with data type "number". May be there are some other reason. Bottom line- we need to slice this search results by the audience to make these search results relevant.

content/docs/command-reference/plots/modify.md
```diff
@@ -123,6 +123,29 @@ plots:
...
+```yaml
+evaluate:
+  cmd: python src/evaluate.py
+    deps:
+    - src/evaluate.py
+    outs:
+    - logs.csv
+    plots:
+    - scores.json
+```
```
jorgeorpinel, line 141: Let's mention somewhere that this is a `dvc.yaml` file (so its linked to the DVC files page)

content/docs/command-reference/plots/modify.md
```diff
@@ -123,6 +123,29 @@ plots:
...
+
+To change the output file `logs.csv` into a plot run:
+
+```dvc
+$ dvc plots modify logs.csv
+```
```
jorgeorpinel, line 147: Maybe say what the operation is going to be and then use a ` ```git ` (instead of YAML) with `-` and `+` rows showing the `out` being moved to `plots`.

content/docs/command-reference/plots/modify.md
imhardikj: Added a paragraph in description.

content/docs/command-reference/plots/modify.md
imhardikj: > `-` and `+` rows showing the file being moved from outs to plots

Updated the example with this.

content/docs/command-reference/plots/diff.md
imhardikj: Remove these changes for now. If they are correct I'll make a new PR for the same.

jorgeorpinel: I agree that "index" would better than "number" if we didn't use the term for another concept in the same paragraph @shcheklein. That's always been the reason to use "number" here. The trends I presented are just to show that it's an OK alternative. I'm aware that those trends don't filter by field.

shcheklein: @jorgeorpinel yep, got it! I guess I didn't understand all the nuances of the question here.

jorgeorpinel: Yep, Sorry! This came out of nowhere haha

content/docs/command-reference/plots/modify.md
```diff
@@ -26,6 +26,10 @@ The path to the metrics file `target` is required. It must be listed in a
 `dvc.yaml` file (see the `--plots` option of `dvc run`). `dvc plots modify` adds
 the display properties to `dvc.yaml`.
 
+Convert any
+[output or simple metrics file into a plot](#example-convert-any-output-into-a-plot)
+by passing it as a `target` to `dvc plot modify`.
+
```
jorgeorpinel, line 32: - Strange choice of which words are linked, please review.
- Let's be a bit more descriptive (less prescriptive) in cmd descriptions. Something like "Note that a secondary use of this command is to convert..."
- Should probably be the last paragraph in the desc.

content/docs/command-reference/plots/modify.md
jorgeorpinel: Missing link part.

content/docs/command-reference/plots/modify.md
jorgeorpinel: Thanks, looks great in https://dvc-landing-plotscmd-kdw3xfowb.herokuapp.com/doc/command-reference/plots/modify#example-convert-any-output-into-a-plot!

content/docs/command-reference/plots/modify.md
```diff
@@ -131,3 +135,27 @@ be changed in many cases. A simple command changes the template:
...
+file of a stage into a plot. Lets take an example `evaluate` stage which has
+`logs.csv` as output. To change the output file `logs.csv` into a plot run:
+
+```dvc
+$ dvc plots modify logs.csv
+```
+
+Changes in `dvc.yaml` shows `logs.csv` is now a plot:
+
+```git
```
jorgeorpinel, line 151: This is great but now that I read it vs. the paragraph in the description (which has all the info you need, really) it looks like overkill to have an entire example for this... Not sure.

I guess it doesn't hurt? But let's try to condense it more: 1 short paragraph (2 sentences preferrably), the `dvc` code block, and the `git` code block. There's a challenge for you @imhardikj !

content/docs/command-reference/plots/modify.md
jorgeorpinel: BTW does this also apply to `metrics modify`? Let's check, please. And if so include equivalent changes there.

jorgeorpinel (commented):
So far so good! A few more comments added ☝️

content/docs/command-reference/plots/modify.md
imhardikj: Sorry. Missed the link. Added now.

content/docs/command-reference/plots/modify.md
imhardikj: Updated the paragraph.
> apply to `metrics modify`

Metrics doesn't have `modify` subcommand

content/docs/command-reference/plots/modify.md
imhardikj: > let's try to condense paragraph more

Updated will much simpler info

content/docs/command-reference/plots/modify.md
```diff
@@ -30,6 +30,10 @@ Property names are passed as [options](#options) to this command (prefixed with
 `--`). These are based on the full
 [Vega specification](https://vega.github.io/vega/docs/specification/).
 
+Note that a secondary use of this command is to convert output or simple
+[metrics file](/doc/command-reference/metrics) into a plot (see
+[example](#example-convert-any-output-into-a-plot)).
```
jorgeorpinel, line 35:
```suggestion
Note that a secondary use of this command is to convert output or
[simple metrics](/doc/command-reference/metrics) into a plot (see
file [example](#example-convert-any-output-into-a-plot)).
```

jorgeorpinel: Or we could use "or simple `dvc metrics` file" perhaps...

content/docs/command-reference/plots/modify.md
jorgeorpinel: Ah yeah 😅 OK then!

content/docs/command-reference/plots/modify.md
jorgeorpinel: Yep, almost perfect! (missing a "the")

content/docs/command-reference/plots/modify.md
```diff
@@ -131,3 +135,25 @@ be changed in many cases. A simple command changes the template:
 ```dvc
 $ dvc plots modify classes.csv --template confusion
 ```
+
+## Example: Convert any output into a plot
+
+Let's take an example `evaluate` stage which has `logs.csv` as an output. We use
+`dvc plots modify` to convert the `logs.csv` output file into a plot, and then
+observe changes that happened in `dvc.yaml`:
```
jorgeorpinel, line 143: - We **can** use
- observe **the** changes

jorgeorpinel: I'm not sure "observe" is the right term though, how about "check" or "confirm"?

jorgeorpinel (approved):
A couple details left before we can merge this, but it looks ready! Thanks @imhardikj

content/docs/command-reference/plots/modify.md
imhardikj: "confirm" looks better

content/docs/command-reference/plots/modify.md
```diff
@@ -30,6 +30,10 @@ Property names are passed as [options](#options) to this command (prefixed with
 `--`). These are based on the full
 [Vega specification](https://vega.github.io/vega/docs/specification/).
 
+Note that a secondary use of this command is to convert output or simple
+`dvc metrics` file into a plots file (see
```
jorgeorpinel, line 34: > `dvc metrics` file

I like this, we should probably apply in other places where manual links to "plots/metrics file(s)" exist. (In another PR).

jorgeorpinel: But don't worry about it @imhardikj, I'll take care of that.

jorgeorpinel, line 34: Missing an article though:
```suggestion
`dvc metrics` file into a plots file (see an
```

PR #1951, opened 2020-11-17
docs: add desc field

efiop:
Per https://github.com/iterative/dvc/pull/4901

### DIFF
--- a/content/docs/command-reference/add.md
+++ b/content/docs/command-reference/add.md
@@ -7,7 +7,7 @@ file.
 
 ```usage
 usage: dvc add [-h] [-q | -v] [-R] [--no-commit] [--external]
-               [--file <filename>] [--glob]
+               [--file <filename>] [--glob] [--desc <text>]
                targets [targets ...]
 
 positional arguments:
@@ -147,6 +147,9 @@ directory symlinks.
 - `--glob` - allows adding files and directories that match the specified
   pattern as specified by `target`. Shell-style wildcards are supported: `*`,
   `?`, `[seq]`, `[!seq]`, and `**`.
+
+- `--desc <text>` - user description of the data (optional). This doesn't affect
+  any DVC operations.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/run.md
+++ b/content/docs/command-reference/run.md
@@ -12,7 +12,7 @@ usage: dvc run [-h] [-q | -v] -n <name> [-d <path>] [-o <path>]
                [-w <path>] [--no-exec] [-f]
                [--no-run-cache] [--no-commit]
                [--outs-persist <path>] [--outs-persist-no-cache <path>]
-               [--always-changed] [--external]
+               [--always-changed] [--external] [--desc <text>]
                command
 
 positional arguments:
@@ -262,6 +262,9 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
 
 - `--external` - allow outputs that are outside of the DVC repository. See
   [Managing External Data](/doc/user-guide/managing-external-data).
+
+- `--desc <text>` - user description of the stage (optional). This doesn't  
+  affect any DVC operations.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/user-guide/dvc-files-and-directories.md
+++ b/content/docs/user-guide/dvc-files-and-directories.md
@@ -37,6 +37,7 @@ edit, or even created manually. Here is a sample:
 outs:
   - md5: a304afb96060aad90176268345e10355
     path: data.xml
+    desc: cats and dogs dataset
 
 # Comments and user metadata are supported.
 meta:
@@ -74,6 +75,8 @@ An _output entry_ (`outs`) consists of these fields:
 - `nfiles`: If a directory, number of files inside.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `desc`: User description for this output. This doesn't affect any DVC
+  operations.
 
 A _dependency entry_ (`deps`) consists of these fields:
 
@@ -128,12 +131,14 @@ stages:
     metrics:
       - performance.json
   training:
+    desc: train your model
     cmd: python train.py
     deps:
       - train.py
       - features
     outs:
-      - model.pkl
+      - model.pkl:
+          desc: my model description
     plots:
       - logs.csv:
           x: epoch
@@ -173,6 +178,8 @@ the possible following fields:
 - `meta` (optional): Arbitrary metadata can be added manually with this field.
   Any YAML contents is supported. `meta` contents are ignored by DVC, but they
   can be meaningful for user processes that read or write `.dvc` files directly.
+- `desc` (optional): User description for this stage. This doesn't affect any
+  DVC operations.
 
 `dvc.yaml` files also support `# comments`.
 
--- a/content/docs/command-reference/import-url.md
+++ b/content/docs/command-reference/import-url.md
@@ -11,6 +11,7 @@ changes in the remote data source. Creates a `.dvc` file.
 
 ```usage
 usage: dvc import-url [-h] [-q | -v] [--file <filename>] [--no-exec]
+                      [--desc <text>]
                       url [out]
 
 positional arguments:
@@ -128,6 +129,9 @@ source.
   The data hash is not calculated by this, only the metadata is saved into the
   `.dvc` file. You can use `dvc commit <out>.dvc` if you need the hashes in the
   new `.dvc` file and save existing data to the cache.
+
+- `--desc <text>` - user description of the data (optional). This doesn't  
+  affect any DVC operations.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
 
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -12,7 +12,7 @@ the import.
 ```usage
 usage: dvc import [-h] [-q | -v]
                   [-o <path>] [--file <filename>]
-                  [--rev <commit>] [--no-exec]
+                  [--rev <commit>] [--no-exec] [--desc <text>]
                   url path
 
 positional arguments:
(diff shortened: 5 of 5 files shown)


content/docs/command-reference/add.md
```diff
@@ -148,6 +148,8 @@ directory symlinks.
   pattern as specified by `target`. Shell-style wildcards are supported: `*`,
   `?`, `[seq]`, `[!seq]`, and `**`.
 
+- `--desc` - arbitrary string description for generated `.dvc` file.
```
shcheklein, line 151: for generated .dvc file -> for data artifact ... or something similar, it does not describe `.dvc` file.

I wonder if we should provide a short-cut (`-d`)? also, would `-m` be a better option? to remind `git commit -m` ?

shcheklein: it should be also `--desc <description>` I think?

shcheklein (changes requested):
A few comments.

How about desc for imports/import-url?

Also, for outputs in the pipeline?

content/docs/command-reference/add.md
efiop: > for generated .dvc file -> for data artifact ... or something similar, it does not describe .dvc file.

Depends on how to look at it. It is not really for specific `out` either, from the internal perspective. In 2.0 we'll likely simplify the internal structure for `dvc add`-generated file, and then your suggestion will be 100% correct.

> I wonder if we should provide a short-cut (-d)? also, would -m be a better option? to remind git commit -m ?

Both -d and -m are taken. Don't want to take up short options for it just yet. Also analogy with `git commit -m` is not that intuitive for this, IMO.

> it should be also --desc <description> I think?

Added. Thanks.

shcheklein: > Depends on how to look at it. 

it's important to understand what is the purpose of this from the end user perspective and describe it in a way that would resonates with that in the first place. To my mind it's about providing descriptions to data unless I'm missing something. Now I read as I describe the file - why would I even want to do that?

> Don't want to take up short options for it just yet.

yep, fine 👍

efiop:
> How about desc for imports/import-url?

Will add for symmetry. :+1: 

> Also, for outputs in the pipeline?

I don't think it was part of the plan. But let me clarify...

content/docs/command-reference/add.md
jorgeorpinel: > To my mind it's about providing descriptions to data

Agree. There's already the `meta` field for general user info about the .dvc file itself.

> would -m be a better option? to remind git commit -m

Good idea but I also think it's not intuitive: It could be confusing since this is `add`, not `commit`.

> Both -d and -m are taken.

Not for `add`.

jorgeorpinel: My suggestion:
```suggestion
- `--desc <text>` - arbitrary description for the data `targets`.
```
> Please confirm metavar per https://github.com/iterative/dvc/pull/4901#pullrequestreview-532809188)

BTW does the same `desc` get applied to multiple targets (if several are provided to `dvc add`)? Thanks

content/docs/command-reference/run.md
```diff
@@ -263,6 +263,8 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
 - `--external` - allow outputs that are outside of the DVC repository. See
   [Managing External Data](/doc/user-guide/managing-external-data).
 
+- `--desc <description>` - arbitrary string description for this stage.
```
jorgeorpinel, line 266:
```suggestion
- `--desc <text>` - arbitrary description for the stage outputs. This doesn't
  affect any DVC operations.
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -60,6 +61,7 @@ meta:
 - `meta` (optional): Arbitrary metadata can be added manually with this field.
   Any YAML contents is supported. `meta` contents are ignored by DVC, but they
   can be meaningful for user processes that read `.dvc` files.
+- `desc` (optional): Arbitrary string description for this `.dvc` file.
```
jorgeorpinel, line 64: Wait, so this field is for the whole .dvc file? So what's the difference with `meta`? I thought this would be per output (so, in the next bullet list).

efiop: Yes, both. `meta` is not used by us at all, it is a completely arbitrary metadata. `desc` is required to be a string/text and will be used by Viewer and maybe even dvc at some point.

content/docs/command-reference/add.md
efiop: Thanks, updated the doc and the help in the original PR.

> BTW does the same desc get applied to multiple targets (if several are provided to dvc add)? Thanks

Yes.

efiop:
Combined your suggestions in both PRs @jorgeorpinel . Thanks!

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: I see, thanks for the explanation @efiop. Would it be better to keep it hidden (undocumented) for now? I see no significative difference between `meta` and `desc` from the DVC user perspective. Possible source of confusion

> p.s. I'd even suggest the Viewed uses a special format in `meta` values instead.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -74,6 +75,8 @@ An _output entry_ (`outs`) consists of these fields:
 - `nfiles`: If a directory, number of files inside.
 - `cache`: Whether or not this file or directory is <abbr>cached</abbr> (`true`
   by default, if not present). See the `--no-commit` option of `dvc add`.
+- `desc`: User description for this output. This doesn't affect any DVC
+  operations.
```
jorgeorpinel, line 79: Also, is it correct that for .dvc files `desc` is a subfield in `outs`? Seems odd since it's a top-level field in dvc.yaml files.

efiop: It is both.

content/docs/user-guide/dvc-files-and-directories.md
fabiosantoscode: Hey, I didn't know about this but it sounds like a good idea. Regarding screen real estate we can only really display it when you hover or click a file version, where we show some more long form information like the full hash, delta and change dates.

@shcheklein what do you think?

efiop: We've talked about `meta` before. `meta` is for arbitrary data that is not for us at all. Not even for viewer. `desc` is something that we recognize and something that viewer will be using very soon. We will be using it in some dvc commands in the future as well.

shcheklein: @fabiosantoscode yes, we should def read and show this information if it's present. I hoped it was already part of the tooltip in the viewer?

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Sorry, didn't get that. I guess you're saying that yes, this is correct: `desc` is at the `outs` level in .dvc files, but at the top-level for dvc.yaml. This "inconsistency" seems unintuitive to me.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > desc is something that we recognize

But https://dvc.org/doc/user-guide/dvc-files-and-directories reads "This doesn't affect any DVC operations." so it's only the viewer that recognizes it (and to my mind it could do the same with a special format in `meta` but up to you guys. I suspect at some point the Viewer may also display `meta` for the user, since it's their info after all — so the distinction becomes blurry again.

My main objection is to have both these options documented since they look like the same from the user perspective (and desc is a little inconsistent as mentioned in https://github.com/iterative/dvc.org/pull/1951#pullrequestreview-540675703 below).

efiop: @jorgeorpinel It doesn't affect any dvc operations right now, but we still recognize it as a stage/output description string. Meta is also recognized by us but as an arbitrary object. Viewer will likely show it, but dvc - likely not. Still, those are separate things. We've considered using meta for it and decided not to.

content/docs/user-guide/dvc-files-and-directories.md
efiop: desc can be for stage or for specific output

efiop: @jorgeorpinel Let's not focus on this too much, this is minor for docs.

content/docs/user-guide/dvc-files-and-directories.md
fabiosantoscode: Not at all! I didn't know about this and it wasn't present in the designs

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: `desc` may be minor but 1. that doesn't mean its OK to have an error in docs if we see it (not the case here but I wasn't sure until now) and 2. this is making me realize the way we present outputs vs dvc.yaml in https://dvc.org/doc/user-guide/dvc-files-and-directories is confusing because outputs are only expanded on in the .dvc file section... which is kind of a major. I'll address that elsewhere.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: > but we still recognize it as a stage/output description string

What does it mean to "recognize it"? Does DVC actually read/validate/process it in some way it doesn't do with `meta`? I do realize add/run have the --desc flag that writes it (except for stage outputs) so if that's what we mean then OK.

> We've considered using meta for it and decided not to.

Sure, I'm not insisting. I'm just detecting possible inconsistencies and confusions. Maybe I'll just extract this to an issue so we can officially burry it as a nice-to-have 😋

Thanks for all the answers!

jorgeorpinel:
One last comment from me here: The dvc.yaml schema doesn't seem to have the field yet cc @skshetry thanks

See https://github.com/iterative/dvcyaml-schema/blob/master/schema.json

skshetry:
Thanks @jorgeorpinel. I'm planning to add them during 2.0 releases. I'm hoping if I can generate it automatically.

content/docs/user-guide/dvc-files-and-directories.md
efiop: > What does it mean to "recognize it"? Does DVC actually read/validate/process it in some way it doesn't do with meta? I do realize add/run have the --desc flag that writes it (except for stage outputs) so if that's what we mean then OK.

Yes, it does read/validate it as str. Meta is arbitrary, we don't validate it ever.

PR #1953, opened 2020-11-18
dag: explanation of what pipelines are

### LINKED ISSUE #1500: dag: rewrite explanation of what pipelines are
Extracted from https://github.com/iterative/dvc.org/pull/1496#pullrequestreview-437828132:

> all this text is pretty bad :( let's ... rewrite it
(referring to the description of what pipelines are [in the `dag` ref.](https://dvc.org/doc/command-reference/dag#directed-acyclic-graph))

Also, note from https://github.com/iterative/dvc.org/issues/608, that we have to:

> - [x] Decide whether to use term pipeline, DAG, dependency graph, and when, and apply. See #1383 (comment)

UPDATE: Let's use "pipeline" as much as possible (per https://github.com/iterative/dvc.org/issues/1500#issuecomment-649895248).

imhardikj:
Fixes #1500

### DIFF
--- a/content/docs/command-reference/dag.md
+++ b/content/docs/command-reference/dag.md
@@ -15,14 +15,17 @@ positional arguments:
 
 ## Description
 
-A data pipeline, in general, is a series of data processing
+Data science and machine learning pipelines are used to help automate the
+workflow. A data pipeline, in general, is a series of data processing
 [stages](/doc/command-reference/run) (for example, console commands that take an
 input and produce an <abbr>output</abbr>). A pipeline may produce intermediate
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Thanks for the attempt @imhardikj but a) I'm not sure why we're working on this in the context of your GSoD project (is it a 1.0 update or a how-to?) and b) the current changes aren't very meaningful, probably not more correct than before either.

I realize it's a draft so maybe you had other intentions but as GSoD is coming to an end soon, there'll be no time to help you with this so I'm closing it for now. Thanks again

PR #1961, opened 2020-11-20
cmd: Updating description of <DVC_METRIC_Y> in plots

imhardikj:
Per https://github.com/iterative/dvc.org/issues/1414#issuecomment-730093511:

> Update the description of <DVC_METRIC_Y> to specify only the first row matters
> Update the text in the plots index description 

Per https://github.com/iterative/dvc.org/issues/1414#issuecomment-731027131
> let's just double check that all other metrics and plots usage blocks correspond to the -h help output of the commands on the latest DVC version (1.9.1)

### DIFF
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -64,8 +64,11 @@ epoch, AUC, loss
 37, 0.92302, 0.0299015
 ```
 
-Hierarchical file formats such as JSON and YAML consists of an array of objects
-where each object should contain all fields that are required for the plot.
+Hierarchical file formats such as JSON and YAML consists of an array of
+consistent objects (sharing a common structure): All objects should contain the
+fields used for the X and Y axis of the plot (see
+[DVC template anchors](#dvc-template-anchors)); Extra elements will be ignored
+silently.
 
 `dvc plots` subcommands can produce plots for a specified field or a set of
 them, from the array's objects. For example, `val_loss` is one of the field
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -64,11 +64,8 @@ epoch, AUC, loss
 37, 0.92302, 0.0299015
 ```
 
-In hierarchical file formats (JSON or YAML), an array of consistent objects is
-expected: every object should have the same structure. However, if an object has
-extra elements, it won't affect the plot as long as
-[fields specified](#dvc-template-anchors) for X and Y axis are present in all
-the objects.
+Hierarchical file formats such as JSON and YAML consists of an array of objects
+where each object should contain all fields that are required for the plot.
 
 `dvc plots` subcommands can produce plots for a specified field or a set of
 them, from the array's objects. For example, `val_loss` is one of the field
@@ -151,8 +148,8 @@ header (first row) are equivalent to field names.
 
 - `<DVC_METRIC_Y>` (optional) - field name of the data for the Y axis. It can be
   defined with the `-y` option of the `dvc plot` subcommands. The default is the
-  last one found in the first row of the metrics file: the last column for
-  CSV/TSV, or the last field for JSON/YAML.
+  last header of the metrics file: the last column for CSV/TSV, or the last
+  field for JSON/YAML.
 
 - `<DVC_METRIC_X_TITLE>` (optional) - field name to display as the X axis label
 
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -65,7 +65,10 @@ epoch, AUC, loss
 ```
 
 In hierarchical file formats (JSON or YAML), an array of consistent objects is
-expected: every object should have the same structure.
+expected: every object should have the same structure. However, if an object has
+extra elements, it won't effect the plot as long as
+[fields specified](#dvc-template-anchors) for X and Y axis are present in all
+the objects.
 
 `dvc plots` subcommands can produce plots for a specified field or a set of
 them, from the array's objects. For example, `val_loss` is one of the field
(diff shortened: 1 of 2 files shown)


jorgeorpinel:
Hi @imhardikj how about also including

> 2. Update the text in the plots index to something like "array of consistent objects is expected: " (also not sure "expected" is the right term, feel free to rewrite the whole paragraph if needed).

here? It's from the same paragraph in the comment linked to this issue.

content/docs/command-reference/plots/index.md
```diff
@@ -65,7 +65,10 @@ epoch, AUC, loss
 ```
 
 In hierarchical file formats (JSON or YAML), an array of consistent objects is
-expected: every object should have the same structure.
+expected: every object should have the same structure. However, if an object has
+extra elements, it won't affect the plot as long as
+[fields specified](#dvc-template-anchors) for X and Y axis are present in all
+the objects.
```
jorgeorpinel, line 71: > Update the text in the plots index to something like "array of consistent objects is expected: "

OK I see you did try that too...

jorgeorpinel:
Would be nice to throw in https://github.com/iterative/dvc.org/issues/1414#issuecomment-731027131 as well, to finish addressing all those comments...

content/docs/command-reference/plots/index.md
jorgeorpinel, line 71: OK we have all the info. in but this got too long and complex. Instead of just appending a bunch of conditions to the previous text ("array of consistent ... same structure"), which is incorrect, can we just rewrite it?

content/docs/command-reference/plots/index.md
```diff
@@ -148,8 +151,8 @@ header (first row) are equivalent to field names.
 
 - `<DVC_METRIC_Y>` (optional) - field name of the data for the Y axis. It can be
   defined with the `-y` option of the `dvc plot` subcommands. The default is the
```
jorgeorpinel, line 153:
```suggestion
  defined with the `-y` option of the `dvc plot` subcommands. t defaults to the
```

content/docs/command-reference/plots/index.md
```diff
@@ -147,9 +150,9 @@ header (first row) are equivalent to field names.
   `index` field (explained above) is the default.
 
 - `<DVC_METRIC_Y>` (optional) - field name of the data for the Y axis. It can be
-  defined with the `-y` option of the `dvc plot` subcommands. The default is the
-  last one found in the metrics file: the last column for CSV/TSV, or the last
-  field for JSON/YAML.
+  defined with the `-y` option of the `dvc plot` subcommands. t defaults to the
```
jorgeorpinel, line 153:
```suggestion
  defined with the `-y` option of the `dvc plot` subcommands. It defaults to the
```

content/docs/command-reference/plots/index.md
```diff
@@ -147,9 +150,9 @@ header (first row) are equivalent to field names.
   `index` field (explained above) is the default.
 
 - `<DVC_METRIC_Y>` (optional) - field name of the data for the Y axis. It can be
-  defined with the `-y` option of the `dvc plot` subcommands. The default is the
-  last one found in the metrics file: the last column for CSV/TSV, or the last
-  field for JSON/YAML.
+  defined with the `-y` option of the `dvc plot` subcommands. It defaults to the
+  last one found in the first row of the metrics file: the last column for
+  CSV/TSV, or the last field for JSON/YAML.
```
jorgeorpinel, line 155: > It defaults to the last one found in the first row of the metrics file: the last column for CSV/TSV

Questions:

- What about CSV/TSV files that have a headers row though? Is it the last header from that first row, or some un-named column in case the first data row is longer than headers?
- What about CSV/TSV files without a headers row? See `dvc plots show --no-header`.

jorgeorpinel: In any case,

> last one found in the first row of the metrics file

is too hard to read. Let's find a way to simplify that is correct even if it doesn't have all the implementation details, e.g. `last header found in the metrics file`.

content/docs/command-reference/plots/index.md
imhardikj: Pushed new updates.

content/docs/command-reference/plots/index.md
imhardikj: > What about CSV/TSV files that have a headers row though? Is it the last header from that first row, or some un-named column in case the first data row is longer than headers?

It will be last header from first row.

> What about CSV/TSV files without a headers row? See dvc plots show --no-header

Again it will be last value from first row. Suppose first row has 6 values. So in each row every 6th value will be used to generate the plot.

content/docs/command-reference/plots/index.md
```diff
@@ -64,8 +64,8 @@ epoch, AUC, loss
 37, 0.92302, 0.0299015
 ```
 
-In hierarchical file formats (JSON or YAML), an array of consistent objects is
-expected: every object should have the same structure.
+Hierarchical file formats such as JSON and YAML consists of an array of objects
+where each object should contain all fields that are required for the plot.
```
jorgeorpinel, line 68: OK it's good that this is short again but this text doesn't seem to explain what we're trying to cover here (how to put plots data in hierarchical formats). In fact it sounds like we're suddenly describing what JSON and YAML files are in general.

For example, "contain all fields that are required for the plot" is implied, what we need to explain is how.

jorgeorpinel: Things to summarize (from https://github.com/iterative/dvc.org/pull/1961#pullrequestreview-540688242):

- json/yaml files shouodl contain an array of objects
- every object should have the same base structure
- X and Y fields should be part of that base structure
- extra elements will be ignored (not an error though)

content/docs/command-reference/plots/index.md
jorgeorpinel: Sounds good, resolving this one.

content/docs/command-reference/plots/index.md
jorgeorpinel, line 68:
```suggestion
Hierarchical file formats such as JSON and YAML consists of an array of
consistent objects (sharing a common structure): All objects should contain the
fields used for the X and Y axis of the plot (see
[DVC template anchors](#dvc-template-anchors)); Extra elements will be ignored
silently.
```

jorgeorpinel: Finishing this up in https://github.com/iterative/dvc.org/pull/1961#pullrequestreview-555140849 below.

jorgeorpinel (approved):
LGTM!

PR #1968, opened 2020-11-24
Update remove.md

Erotemic:
Based on a conversation I had with ivan on discord, running `dvc remove` on a `.dvc` file does *not* remove the tracked files from the workspace. I also modified the wording to be slightly more clear. 

I also verified that this is the case in a test.

```bash

mkdir -p $HOME/tmp/dvc-test
rm -rf $HOME/tmp/dvc-test
mkdir -p $HOME/tmp/dvc-test

cd $HOME/tmp/dvc-test
mkdir test_repo
cd test_repo

git init

echo "hello world" > README.md

git add README.md
git commit -am "initial commit"

dvc init
git commit -am "initialize dvc"

mkdir -p my_data/images
echo "a" > my_data/images/img1.jpg
echo "b" > my_data/images/img2.jpg
echo "c" > my_data/images/img3.jpg

echo "e" > my_data/annotations.json

# Add the top level directory as a single dvc file
dvc add my_data

git add my_data.dvc .gitignore
git commit -am "Use dvc to track the entire data folder"

dvc remove my_data.dvc

dvc add my_data/images
dvc add my_data/annotations.json

git add my_data/annotations.json.dvc my_data/images.dvc my_data/.gitignore

git commit -am "Subdivide monolithic dvc file into smaller parts"
```

PR #1971, opened 2020-11-25
cmd: corrections and copy edits, etc. in remove 

jorgeorpinel:
- `remove` updates per 1.0 changes and #1968
- Clarify `gc -w`


content/docs/command-reference/remove.md
```diff
@@ -1,6 +1,7 @@
 # remove
 
-Remove stage, `.gitignore` entry, and unprotect outputs.
+Remove stages or `.dvc` files, unprotect or delete their <abbr>outputs</abbr>,
+and erase the `.gitignore` entries.
```
jorgeorpinel, line 4: This should have a matching core repo update, probably.

content/docs/command-reference/remove.md
```diff
@@ -13,35 +14,32 @@ positional arguments:
 
 ## Description
 
-This command safely removes stages from
-[dvc.yaml](/doc/user-guide/dvc-files-and-directories#dvcyaml-file), their
-`.gitignore` entries, and optionally removes from the <abbr>workspace</abbr>
-files or directories that are tracked by DVC. It takes one or more stage names
-(see `-n` option of `dvc run`) or `.dvc` files as target, removes it, and
-optionally removes all of its outputs (`outs` field).
+This command can safely remove stages from `dvc.yaml` or `.dvc` files. This
```
shcheklein, line 17: `sages from .dvc files` - this doesn't sound correct?

content/docs/command-reference/remove.md
```diff
@@ -13,35 +14,32 @@ positional arguments:
...
-`dvc remove` does not remove files from the DVC cache or remote storage (see
-`dvc gc`). However, remember to run `dvc push` to save the files you actually
-want to use or share in the future.
+Note that the actual <abbr>output</abbr> files or directories tracked by the
+stage (`outs` field) are not removed by this command, unless the `--outs` option
+is used.
 
-Refer to [Updating Tracked Files](/doc/user-guide/how-to/update-tracked-files)
-to see how it can be used to replace or modify files that are tracked by DVC.
+> `dvc remove` doesn't remove files from the DVC <abbr>cache</abbr> or
```
shcheklein, line 31: should be at the very top I think - very important for people to understand that the can always revert changes in Git and do `dvc checkout` to restore their data

shcheklein: if data is cached of course

content/docs/command-reference/remove.md
jorgeorpinel: Just noticed this but it was fixed before. Thanks

content/docs/command-reference/remove.md
jorgeorpinel: OK I'm moving the note near the top, but I didn't understand the relationship between the text in the note and what you're describing... Let me 🤔

jorgeorpinel: > can always revert changes in Git and do dvc checkout to restore their data

Yeah sorry I'm not seeing the connection to `remove` here. But the note is up now.

shcheklein:
Merge master to fix CI?

content/docs/command-reference/gc.md
```diff
@@ -51,8 +51,8 @@ The default remote is cleaned (see `dvc config core.remote`) unless the
 
 ## Options
 
-- `-w`, `--workspace` - keep files and directories _only_ referenced in the
-  current workspace This option is enabled automatically if `--all-tags`,
+- `-w`, `--workspace` - keep _only_ files and directories referenced in the
+  workspace. This behavior is included automatically with `--all-tags`,
   `--all-branches`, or `--all-commits` are used.
```
shcheklein, line 56: `are used` should be removed now

why do we change this at all? looks very cosmetic - takes time to review

content/docs/command-reference/remove.md
shcheklein: It's like a complete algorithm here, but what does this command do from the user perpective?

content/docs/command-reference/remove.md
```diff
@@ -13,35 +14,32 @@ positional arguments:
...
-`dvc remove` does not remove files from the DVC cache or remote storage (see
-`dvc gc`). However, remember to run `dvc push` to save the files you actually
-want to use or share in the future.
+If there are no stages left in `dvc.yaml` after the removal, then both
+`dvc.yaml` and `dvc.lock` are deleted.
 
-Refer to [Updating Tracked Files](/doc/user-guide/how-to/update-tracked-files)
-to see how it can be used to replace or modify files that are tracked by DVC.
+Note that the actual <abbr>output</abbr> files or directories tracked by the
+stage (`outs` field) are not removed by this command, unless the `--outs` option
```
shcheklein, line 32: `.dvc` is not tracked by the stage

content/docs/command-reference/remove.md
shcheklein: we mention unprotect here but not in the Description - also doesn't sound right

content/docs/command-reference/remove.md
```diff
@@ -13,35 +14,32 @@ positional arguments:
 
 ## Description
 
-This command safely removes stages from
-[dvc.yaml](/doc/user-guide/dvc-files-and-directories#dvcyaml-file), their
-`.gitignore` entries, and optionally removes from the <abbr>workspace</abbr>
-files or directories that are tracked by DVC. It takes one or more stage names
-(see `-n` option of `dvc run`) or `.dvc` files as target, removes it, and
-optionally removes all of its outputs (`outs` field).
+Safely removes `.dvc` files or stages from `dvc.yaml`. This includes deleting
```
shcheklein, line 17: I like the previous version better (even though it was not exactly correct with regards to `.dvc` files). We need to find a better language -  something that unites stages and .dvc - a higher level abstraction to convey the essence of this command.

content/docs/command-reference/gc.md
jorgeorpinel: Removed "are used".

Changed because "_only_ referenced" means something quite different to "keep _only_". The previous phrasing was incorrect, as they can be references in multiple versions but that doesn't matter nor it it the point. A typo basically

jorgeorpinel: I guess the 2nd sentence I didn't need to change. "Included" feels more correct that "enabled" too me but no strong opinion, I can roll that back if you'd like.

jorgeorpinel: p.s. I suppose "implied" is even better.

content/docs/command-reference/remove.md
jorgeorpinel: > hat does this command do from the user perpective?

The opposite of `add` or `run/repro` per stage output... ⌛ 

> we mention unprotect

True, I guess this doesn't unprotect outputs, it unlinks them rather (no matter which link type). Updating...

jorgeorpinel: Changed it to just "Stop tracking files or directories, and optionally delete them."

BTW more and more https://dvc.org/doc/user-guide/how-to/stop-tracking-data seems redundant... Should I just remove that how-to and include `dvc gc` in an example here in the remove ref?

jorgeorpinel: > Sent core update to iterative/dvc/pull/4999

jorgeorpinel: More updates here... Actually, I guess it's more than stopping tracking data in the case of stages: it removes the entire stage... So this command basically does different things depending on the targets you send it: for .dvc files or outputs it stops tracking the data; for stages it removes them form the pipeline.

Updated again, to "Remove stages from `dvc.yaml` and/or stop tracking files or directories (and optionally delete them)."

content/docs/command-reference/remove.md
jorgeorpinel: > need to find a better language - something that unites stages and .dvc

We have metafiles for now (DVC Files soon) but it doesn't really apply here since the command's behavior is different for one or the other. I think that we need to specify each one.

content/docs/command-reference/gc.md
shcheklein: 👍

content/docs/command-reference/remove.md
shcheklein: > Should I just remove that how-to and include dvc gc in an example here in the remove ref?

don't have a strong opinion, but yes, they overlap a lot

how to might be useful since it's easier to find to my mind and a lot of people asking about it

shcheklein (approved):

jorgeorpinel:
Merging to master and v1

PR #1972, opened 2020-11-26
Nov gems

elleobrien:
Draft is here- depending on how fast it is to get out, we can try to get out tonight or Monday, when I'm back in the office after Thanksgiving. Monday is still November :) 

Note that I'll update the comment link when we're ready to post.


content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+  - CML
+  - Cache
+  - Bitbucket
+---
+
+## DVC questions
+
+### [Q: If I make a new Git branch, how can I synchronize it with DVC?](https://discord.com/channels/485586884165107732/485596304961962003/773498570795778058)
+
+Generally, yes. When you make a new Git branch in your project:
```
shcheklein, line 27: `Generally, yes.` reads a bit wrong here? (question is `how`)

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+  - Discord
+  - Gems
+  - CML
+  - Cache
+  - Bitbucket
+---
+
+## DVC questions
+
+### [Q: If I make a new Git branch, how can I synchronize it with DVC?](https://discord.com/channels/485586884165107732/485596304961962003/773498570795778058)
```
shcheklein, line 25: to be honest, question is a bit strange as well ... if it's a new branch - we don't need to synchronize anything, right?

elleobrien: i had that thought too.... i thought maybe i was misunderstanding. checkout hook is more for when you switch to a branch that has a different workspace, which by definition would only happen when the branch isn't brand new. Maybe we can change question to

`If i checkout a different Git branch, how do I synchronize with DVC?` Not exactly what was in the channel but more useful, maybe

shcheklein: > checkout hook is more for when you switch to a branch that has a different workspace, which by definition would only happen when the branch isn't brand new

yep!

> Not exactly what was in the channel but more useful, maybe

yep, sounds def better!

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+  repo="https://github.com/<your-repo>")
+)
+```
+
+This code means the API will return the URL for a file that ends in `.dir`. The
+`.dir` file contains a JSON-formatted table of the hashes and relative paths for
+all the files inside `<top-level-directory>`. You could then parse that file to
+tget the locations of the full directory contents.
+
+Why use the API? It's a good way to access DVC-tracked files without downloading
```
shcheklein, line 65: so, in this case it doesn't provide much of a help - you have to parse JSON, and you are pretty much on your own- have to use specific cloud library to stream/downloads.

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+them to your workspace! No downloads means nothing in your cache. Read more
```
shcheklein, line 66: > No downloads means nothing in your cache.

not sure I understood this, tbh

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+resource_url = dvc.api.get_url(
+  "<top-level-directory>",
+  repo="https://github.com/<your-repo>")
+)
+```
+
+This code means the API will return the URL for a file that ends in `.dir`. The
+`.dir` file contains a JSON-formatted table of the hashes and relative paths for
+all the files inside `<top-level-directory>`. You could then parse that file to
+tget the locations of the full directory contents.
```
shcheklein, line 63: tget -> get

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+By default, DVC pipelines records all your stages (and their inputs and outputs)
+in a single file, `dvc.yaml`. Per directory, you can have one `dvc.yaml` file.
+If you want to run pipelines in a different folder than your project root, you
+could create another `dvc.yaml` in a subdirectory.
+
+However, `dvc.yaml` is intended to be the only file you need to record and
+reproduce pipelines per directory. Pipelines are designed to have all stages
+stored in the same place, and there's currently no method to rename `dvc.yaml`.
+
+If you really, really, _really_ want to put different stages in different files,
```
shcheklein, line 84: let's focus on multiple `dvc.yaml`s only. Renaming with `--single-stage` is a hack and is deprecated. Also, author was looking for a solution to split, not rename.

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+
+Why use the API? It's a good way to access DVC-tracked files without downloading
+them to your workspace! No downloads means nothing in your cache. Read more
+about our
+[Python API and the `get_url` function](https://dvc.org/doc/api-reference/get_url#dvcapiget_url)
+in our docs.
+
+### [Q: Can I have more than one `dvc.yaml` file in my project?](https://discord.com/channels/485586884165107732/563406153334128681/777946398250893333)
+
+_Sort of._ Here's what we mean.
```
shcheklein, line 73: I guess if we focus on multiple dvc.yaml per directory, then we can remove this?

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+If you really, really, _really_ want to put different stages in different files,
+there is one workaround: you can use our hidden option,
+`dvc run --single-stage`, to do this. Your stage will be stored in its own
+`.dvc` file, which can be named whatever you choose. This is a remnant from the
+pre-DVC 1.0 era. But we hope you'll embrace 1.0 and the simplicity of
+`dvc.yaml`!
+
+### [Q: How can I untrack a file that's being tracked by DVC? I want to remove it from remote storage and my local cache, too.](https://discord.com/channels/485586884165107732/563406153334128681/773277514717462548)
+
+We don't have a specific command for this, but you can do it manually:
```
shcheklein, line 93: we can mention this link first - https://dvc.org/doc/user-guide/how-to/stop-tracking-data

then say, that alternative is to find the file manually (doesn't work that easy for a directory, btw), mention some link that describes how cache is organized - https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory ?

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+cache- [read up here](https://dvc.org/doc/command-reference/gc). This function
+lets you get granular about what you're keeping; for example, you can instruct
+`dvc gc` to preserve cache files that are currently used your local worksapce,
+tips of Git branches, tagged Git commits or all Git commits. Everything else
+will be removed.
+
+One word of caution: make sure that when you collect garbage from your cache,
+you don't delete any files that you haven't yet pushed to a remote. If this
+happens, you'll delete them permanently. To be safe, it never hurts to
+`dvc push` before cleaning.
```
shcheklein, line 115: `dvc push` with appropriate options ? ... (by default it will push only current workspace)

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+1. Like GitLab, Bitbucket Cloud requires you to create a token for authorizing
+   CML to write comments. Make sure you don't forget this step (it's in the
+   docs!) or you'll surely hit a permissions error.
+
+2. Bitbucket Cloud uses Bitbucket Pipelines for continuous integration
+   workflows, which currently doesn't support self-hosted runners. That means
+   bringing your own GPUs is not supported. Sorry! But you can still have all
+   the other CML benefits of plots, tables and text in your Pull Request.
+
+3. Bitbucket Server support (with Jenkins and Bamboo) is under active
```
shcheklein, line 135: is there a link to the ticket?

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+project repo,
+[so check them out](https://github.com/iterative/cml/wiki/CML-with-Bitbucket-Cloud)
+to get started. A few quick notes to keep in mind:
+
+1. Like GitLab, Bitbucket Cloud requires you to create a token for authorizing
+   CML to write comments. Make sure you don't forget this step (it's in the
+   docs!) or you'll surely hit a permissions error.
+
+2. Bitbucket Cloud uses Bitbucket Pipelines for continuous integration
+   workflows, which currently doesn't support self-hosted runners. That means
```
shcheklein, line 131: same here, good to link a ticket if we have one (this gems will become outdated, but we'll keep seeing this)

content/blog/2020-11-25-november-20-community-gems.md
```diff
@@ -0,0 +1,195 @@
...
+By the way... stay tuned for a big update here. We're currently replacing the
+Docker Machine approach with a method based on TerraForm, and we can't wait to
+unveil it. It should make deploying cloud instances on AWS, GCP and Azure work
+with less code than ever.
+
+### Q: What did DeeVee do for Thanksgiving?
+
+She stayed home and made mashed potatoes.
+
+![](/uploads/images/2020-11-25/deevee_n_taters.png)
```
shcheklein, line 192: :) 🦃

shcheklein (changes requested):
I hope @elleobrien you had a great thanksgiving 🦃 

PR looks great as usual, some comments questions to address + images look huge (13MB, 1MB)

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: Noted, changed to "Here's what we recommend:"

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: I was trying to say that if you don't download a file, it won't be in your cache. But I think it is not necessary to add this detail (previous sentence explains that the file won't be in your workspace if you use the API to stream it). So I will remove for clarity.

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: Done, I'll remove the `--single-stage` method

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: yep

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: good idea

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: yep

elleobrien: well, they're tickets on Atlassian forums. but that should do

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: yeah, added "`dvc push` your files of interest"

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: yeah, i'll mention the `dvc remove` function. Looks like you'd follow up with `dvc gc -w` according to our docs. But to get rid of the file in remote storage, like the user asked, should they use `dvc gc -w --cloud`?

content/blog/2020-11-25-november-20-community-gems.md
elleobrien: i agree that you get very little help for parsing the data, but I think there's something important about using the API to avoid downloading the huge directory.

Do you think this person's question could be better answered via `dvc list <myremote>`, and then appending the paths of those files to the remote path in your `.dvc/config`?

shcheklein: > but I think there's something important about using the API

it is, but in this case it's not DVC API that will be used (DVC actually even makes it harder - you need to collect file names in the cloud to access this data).

the answer is fine, the only part about `Why use the API?` to my mind is misleading a bit. In this case the answer to that question is that we have to use it to get the list of file since DVC stores files in the content-addressable way, not because it gives us a way to access something w/o downloading.

Better probably give an example here, like Peter did later in the discussion how does path is created

content/blog/2020-11-25-november-20-community-gems.md
shcheklein: yes, it should `--cloud`. But be very careful with `-w` - it will remove a lot of data. I would just say that all `gc` commands should run with some appropriate options after that.

... or alternatively use the method you described.

elleobrien: great, i've worked this in

shcheklein (approved):
Good stuff!

PR #1978, opened 2020-11-28
Misc. updates

jorgeorpinel:
- From my stash of changes done while, but unrelated to, recent PRs
- ~~Update commit and --no-commit options~~ Extracted to 1989

Note for self:
- [x] Finalize the fetch part of iterative/dvc/pull/4999 when this is done.


content/docs/command-reference/fetch.md
```diff
@@ -1,7 +1,8 @@
 # fetch
 
-Get tracked files or directories from
-[remote storage](/doc/command-reference/remote) into the <abbr>cache</abbr>.
+Download <abbr>cache</abbr> files or directories from
+[remote storage](/doc/command-reference/remote), based on the current `dvc.yaml`
+and `.dvc` files.
```
jorgeorpinel, line 5: - [ ] Needs matching core update

content/docs/command-reference/commit.md
```diff
@@ -21,7 +21,7 @@ The `dvc commit` command is useful for several scenarios, when data already
 tracked by DVC changes: when a [stage](/doc/command-reference/run) or
 [pipeline](/doc/command-reference/dag) is in development/experimentation; to
 force-update the `dvc.lock` or `.dvc` files without reproducing stages or
-pipelines; or to mark existing files/dirs as stage <abbr>outputs</abbr>. These
+pipelines; or when modifying stage <abbr>outputs</abbr> without DVC. These
```
shcheklein, line 24: All this paragraph is too complicated and sometimes wrong: e.g. `when data already
tracked by DVC changes` contradicts to the last point. 

it's hard to get the idea behind the last point (not sure the previous version is better)

content/docs/command-reference/fetch.md
```diff
@@ -1,7 +1,8 @@
 # fetch
 
-Get tracked files or directories from
-[remote storage](/doc/command-reference/remote) into the <abbr>cache</abbr>.
+Download <abbr>cache</abbr> files or directories from
+[remote storage](/doc/command-reference/remote), based on the current `dvc.yaml`
```
shcheklein, line 4: dvc.lock ... also I would try to avoid reference to such level of details here ... also, why do we change this, btw?

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +18,16 @@ positional arguments:
 
 ## Description
 
-Downloads DVC-tracked files from remote storage into the cache of the project
-(without placing them in the <abbr>workspace</abbr>, like `dvc pull` would).
-This makes them available for linking (or copying) into the workspace (refer to
-[`dvc config cache.type`](/doc/command-reference/config#cache)).
+`dvc fetch` downloads tracked data from remote storage into the
```
shcheklein, line 21: It's better to avoid starting command reference `dvc fetch` with `dvc fetch`

content/docs/command-reference/fetch.md
shcheklein: also, it's not exactly true - there are options to take into account history

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,15 @@ source.
...
   the imported data (`out`).
 
-- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated when this option is used, only the import
-  metadata is saved to the `.dvc` file. `dvc commit <out>.dvc` can be used if
-  the data hashes are needed in the `.dvc` file, and to save existing data to
-  the cache.
+- `--no-exec` - create the import `.dvc` file without actually downloading
+  `url`. No connection to the source location is made, so the data hash is not
```
shcheklein, line 129: `No connection to the source location is made, so the data hash is not calculated.`

I'm not sure I get this tbh

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,15 @@ source.
...
+  calculated. Only the import operation is recorded in the `.dvc` file. Use
+  `dvc commit` to save any existing dep/out files to the cache and record their
+  hashes to the `.dvc` file.
+
+  This can be useful to skip the download if the file or directory already
```
shcheklein, line 134: Seems it got too complicated ... I'm not sure I follow all the logic in these two paragraphs now

content/docs/user-guide/how-to/merge-conflicts.md
```diff
@@ -108,8 +108,7 @@ branches.
 ### Imported data
 
 To resolve conflicted `.dvc` files generated by `dvc import` or
-`dvc import-url`, remove the conflicted hashes (as well as `size` and, possibly,
-`nfiles`) altogether:
+`dvc import-url`, remove the conflicted value altogether:
```
shcheklein, line 111: value -> values?

content/docs/command-reference/commit.md
jorgeorpinel: > `when data already changes`... when data already

That is the previous version 🙂 I didn't change that part.

I agree this sentence is too complex. The small change here is a small simplification but I can def. work on it more. Keeping in mind this is the cmd ref though so some complexity should be OK.

jorgeorpinel: > when data already changes

BTW, I removed that bit.

content/docs/command-reference/fetch.md
jorgeorpinel: Good call. I moved the detail about dvc.yaml/lock files from the first line down to the description.

> why do we change this

Most of these regular updates come from recent(ish) conversations and discussions in different places e.g. other PRs. They tend to add up in my Git stash and I lose the exact references, sorry.

content/docs/command-reference/fetch.md
jorgeorpinel: Agree, starting at "Downloads" directly.

content/docs/command-reference/import-url.md
jorgeorpinel: I simplified this in 2785d594. PTAL

content/docs/command-reference/import-url.md
jorgeorpinel: I simplified it now as I just mentioned but it's still a little long. I don't see that as a big problem, but if you prefer I can remove the paragraph about motivation (added because you mentioned you liked the phrase "cook the imports before downloading altogether" or something like that).

content/docs/user-guide/how-to/merge-conflicts.md
jorgeorpinel: Yes, values! Fixed, thanks.

content/docs/command-reference/commit.md
```diff
@@ -17,12 +17,11 @@ positional arguments:
...
-tracked by DVC changes: when a [stage](/doc/command-reference/run) or
-[pipeline](/doc/command-reference/dag) is in development/experimentation; to
-force-update the `dvc.lock` or `.dvc` files without reproducing stages or
-pipelines; or to mark existing files/dirs as stage <abbr>outputs</abbr>. These
-scenarios are further detailed below.
+The `dvc commit` command is useful for several scenarios: when a
+[stage](/doc/command-reference/run) or [pipeline](/doc/command-reference/dag) is
+in development/experimentation; to force-update the `dvc.lock` or `.dvc` files
+without reproducing stages or pipelines; or when modifying stage
+<abbr>outputs</abbr> without DVC. These scenarios are further detailed below.
```
shcheklein, line 24: `without DVC` - it won't be clear what does it mean. Better to specify _with_ (how) vs _without_ - it is more explicit.

content/docs/command-reference/commit.md
```diff
@@ -69,13 +67,14 @@ There are many cases where the last step is not desirable (for example rapid
...
+`dvc commit` comes into play: It performs that last step when needed.
 
 Note that it's best to avoid the last three scenarios. They essentially
-force-update the `dvc.lock` or `.dvc` files and save data to cache. They are
-still useful, but keep in mind that DVC can't guarantee reproducibility in those
-cases.
+force-update the metafiles and cache. They are still useful, but keep in mind
+that DVC can't guarantee reproducibility in those cases.
+
+> Note that committing data does not affect code, `dvc.yaml`, or `.dvc` files.
```
shcheklein, line 76: committing data affects .dvc though ... affects is probably not the right word here

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,17 @@ positional arguments:
...
+Downloads tracked data from remote storage into the <abbr>cache</abbr> of the
+project (without placing it in the <abbr>workspace</abbr> like `dvc pull`). This
+makes them available for linking (or copying) into the workspace (refer to
+`dvc config cache.type`).
 
-Without arguments, `dvc fetch` ensures that the files specified in all
-`dvc.lock` and `.dvc` files in the workspace exist in the cache. The
-`--all-branches`, `--all-tags`, and `--all-commits` options enable fetching data
-for multiple Git commits.
+> Note that fetching data does not affect code or
```
shcheklein, line 25: Is it actually needed? If yes, let's simplify it ... affect is also not the right word here to my mind

content/docs/command-reference/commit.md
jorgeorpinel: Working on this I realized there are errors in the cmd ref so I'm rewriting it again (based on previous text but much shorter) avoiding repeating implementation details... ⌛

content/docs/command-reference/freeze.md
```diff
@@ -24,7 +24,7 @@ Freezing a stage is useful to avoid syncing data from the top of its
 (non-frozen) stages only.
 
 Note that <abbr>import stages</abbr> are frozen by default. Use `dvc update` to
-update the imported data from the external data source.
+bring the import up to date from the data source.
```
shcheklein, line 27: with the data source?

content/docs/command-reference/fetch.md
```diff
@@ -1,7 +1,7 @@
 # fetch
 
-Get tracked files or directories from
-[remote storage](/doc/command-reference/remote) into the <abbr>cache</abbr>.
+Download <abbr>cached</abbr> files or directories from
```
shcheklein, line 3: hmm .. not sure I like it this way .. previous version was easy enough. Why do we change it this way.

I'm fine to change Get to Download ... cached files - we usually use `tracked` everywhere else

content/docs/command-reference/fetch.md
```diff
@@ -51,6 +46,15 @@ project's cache                  ++ | dvc pull |
  workspace
 ```
 
+Without arguments, it downloads all files and directories missing from the
+project, found as <abbr>outputs</abbr> in all `dvc.yaml` and `dvc.lock` pairs,
```
shcheklein, line 50: pairs -> files? (too complicated, introduces new concept)

if it relies on `dvc.yaml` then it's fine to omit `dvc.lock` here -it's an implementation detail (can be described if describe how it works). Or mentioned as a separate sentence.

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,17 @@ source.
...
   the imported data (`out`).
 
-- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated when this option is used, only the import
-  metadata is saved to the `.dvc` file. `dvc commit <out>.dvc` can be used if
-  the data hashes are needed in the `.dvc` file, and to save existing data to
-  the cache.
+- `--no-exec` - create the import `.dvc` file without actually downloading
+  `url`. No connection to the data source is made, so only the import operation
```
shcheklein, line 129: `No connection to the data source is made` I don't quite understand this :(

We have 3 paragraphs for a quite simple option - we should simplify it significantly. it's very repetitive now.

shcheklein: same in other case, other commands in this PR

content/docs/command-reference/commit.md
jorgeorpinel: OK I corrected and simplified (also reorged the text) of all of this, included add/run/repo --no-commit options in ae902867. PTAL

content/docs/command-reference/commit.md
jorgeorpinel: Simplified.

I would even like to remove this note but I guess `commit` can be confused between DVC and Git and they're really nothing like each other...

content/docs/command-reference/commit.md
jorgeorpinel: BTW please also check this core request I made: iterative/dvc/issues/5000

content/docs/command-reference/fetch.md
jorgeorpinel: This is another clarification that "DVC is not Git"... I don't think we should need these but who knows... I guess it's there just in case.

jorgeorpinel: Fixed it for now.

content/docs/command-reference/freeze.md
jorgeorpinel: Hmmmm I see what you mean (using the data source) but "bring up to date" + "from <source>" seems better to me.

content/docs/command-reference/fetch.md
jorgeorpinel: > cached files - we usually use tracked everywhere else

But we usually refer to files in the workspace so that wouldn't apply to fetch. I'm just trying to state what this command does in the simplest and most direct way: it downloads files to the cache.

jorgeorpinel: Changed it to

> Download files or directories from
> [remote storage](/doc/command-reference/remote) to the <abbr>cached</abbr>.

without saying "tracked", to avoid confusion.

content/docs/command-reference/fetch.md
jorgeorpinel: - [x] pairs -> files
- [x] fine to omit dvc.lock here

jorgeorpinel: > fine to omit dvc.lock here -it's an implementation detail (can be described if describe how it works

Yeah this is a widespread problem in some cmd refs. Too many or too few mentions of dvc.lock and possibly in the wrong ways...

content/docs/command-reference/import-url.md
jorgeorpinel: > No connection to the data source is made

It was meant originally to explain why we can't even add hash values with this option. But OK, I just removed that now.

I also removed the 2nd p about scenarios but left the note about not knowing whether the data source is valid until attempting the download.

content/docs/command-reference/add.md
```diff
@@ -128,10 +128,9 @@ not.
   among the `targets`, this option is ignored. For each file found, a new `.dvc`
   file is created using the process described in this command's description.
 
-- `--no-commit` - do not save outputs to cache. A `.dvc` file is created, while
-  nothing is added to the cache. (`dvc status` will report that the file is
-  `not in cache`.) Use `dvc commit` when ready to commit outputs with DVC. This
-  is analogous to using `git add` before `git commit`.
+- `--no-commit` - do not cache outputs. A `.dvc` file is created, while nothing
+  is stored in the cache. Use `dvc commit` to finish the operation. This is
```
jorgeorpinel: @shcheklein 

> `save something to cache` - way more explicit
> repetitive with is stored in the cache

I shortened it because it was repeated/explained in the next sentence indeed. But I've reverted and removed the 2nd mention now (in run/repro too).

> def not outputs here

Well, it's in the `outs` field of the .dvc file but I guess that's an implementation detail and we don't want to mix this up with stages. But, so should I remove all the mentions of "output" in add? There's another 4 (even uses abbr tags).

> Use dvc commit to finish the operation. -> can be used to save ...

I don't think we should explain `dvc commit` here. The user can click and learn about it in it's ref? That way we simplify maintaining the commit explanation too.

shcheklein: > But, so should I remove all the mentions of "output" in add?

no, please. This PR is already unmanageable :(

shcheklein: > I don't think we should explain dvc commit here.

I think it's fine to mention. Again not sure why do we change it all to be honest - looks very cosmetic and debatable.

content/docs/command-reference/commit.md
```diff
@@ -1,7 +1,7 @@
 # commit
 
-Record changes to DVC-tracked files in the <abbr>project</abbr>, by saving them
-to the <abbr>cache</abbr> and updating the `dvc.lock` or `.dvc` files.
+<abbr>Cache</abbr> changed files and directories tracked by DVC, and update the
+corresponding `dvc.lock` or `.dvc` files if needed.
```
jorgeorpinel: > it was more explicit and human friendly

"Record changes to DVC-tracked files in the project" ? That's a kind of complicated way to say what commit does: store files in the cache.

Changed to "Stores changed changed files and directories tracked by DVC in the cache" for now.

content/docs/command-reference/commit.md
```diff
@@ -17,65 +17,52 @@ positional arguments:
 
 ## Description
 
-The `dvc commit` command is useful for several scenarios, when data already
-tracked by DVC changes: when a [stage](/doc/command-reference/run) or
-[pipeline](/doc/command-reference/dag) is in development/experimentation; to
-force-update the `dvc.lock` or `.dvc` files without reproducing stages or
```
jorgeorpinel, line 23: > Are there any significant important fixes/changes here?

Yes @shcheklein this became a full revision... I'll extract this to its own PR at this point ⌛

content/docs/command-reference/add.md
jorgeorpinel: > not sure why do we change it all to be honest

A contributor changed this part recently and it became evident that it was a mess, as we are discovering more and more now. I've extracted the commit-related stuff to #1989 so we can try to merge this one first and then come back to this without any rush.

content/docs/command-reference/commit.md
jorgeorpinel: Moved to https://github.com/iterative/dvc.org/pull/1989/files#diff-b8f371cb91c101e733f37e95de023afe203b77b8a0443b3d453a76739b675852L3-R4

content/docs/command-reference/commit.md
jorgeorpinel: Moved to https://github.com/iterative/dvc.org/pull/1989#pullrequestreview-541500924

content/docs/command-reference/add.md
jorgeorpinel: BTW, on

> > def not outputs here
>
> so should I remove all the mentions of "output" in add?

Actually, while I agree that .dvc file outputs aren't real "outputs" (from a stage) it woudl be pretty hard to remove this connection in other refs. e.g. `dvc remove` states "removes `.dvc` files... and optionally the actual output files they track (see `--outs` option)" (being reviewed in #1971).

This connection between .dvc files and the "output" term is probably pervasive, even in the field name of course (`outs`) so I think it's better to treat outputs as a more general concept, not only for stages or pipelines.

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,17 @@ positional arguments:
...
 ## Description
 
-Downloads DVC-tracked files from remote storage into the cache of the project
-(without placing them in the <abbr>workspace</abbr>, like `dvc pull` would).
-This makes them available for linking (or copying) into the workspace (refer to
-[`dvc config cache.type`](/doc/command-reference/config#cache)).
+Downloads tracked data from remote storage into the <abbr>cache</abbr> of the
+project (without placing it in the <abbr>workspace</abbr> like `dvc pull`). This
+makes them available for linking (or copying) into the workspace (refer to
+`dvc config cache.type`).
```
shcheklein, line 23: where will it link it?

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,17 @@ positional arguments:
...
 
-Without arguments, `dvc fetch` ensures that the files specified in all
```
shcheklein, line 25: I still don't see any significant improvements. We removed this paragraph and now we have to put `Note that this does not download [metafiles](/doc/user-guide/dvc-files-and-directories). That can be done with "git fetch" (if using Git).` note, which is very confusing the it is written now.

If you say something like `Note that this does not download [metafiles](/doc/user-guide/dvc-files-and-directories)` you expect user to think "Probably this command downloads metafiles" or asking that question. It's hard for me to imagine it happening that way. What unasked question we are trying to resolve with this note?

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,17 @@ positional arguments:
...
+> [metafiles](/doc/user-guide/dvc-files-and-directories). That can be done with
+> `git fetch` (if using Git).
 
-The `targets` given to this command (if any) limit what to fetch. It accepts
-paths to tracked files or directories (including paths inside tracked
-directories), `.dvc` files, and stage names (found in `dvc.yaml`).
-
-Fetching is performed automatically by `dvc pull` (when the data is not already
-in the <abbr>cache</abbr>), along with `dvc checkout`:
+Fetching is performed automatically if needed by `dvc pull`, along with
```
shcheklein, line 29: `data not in cache` is better than `if needed` (explicit vs some thing that is not clear what does it exactly mean)

content/docs/command-reference/fetch.md
```diff
@@ -51,6 +46,15 @@ project's cache                  ++ | dvc pull |
  workspace
 ```
 
+Without arguments, it downloads all files and directories missing from the
+project, found as <abbr>outputs</abbr> in all `dvc.yaml` and `.dvc` files
```
shcheklein, line 50: Without arguments, it downloads files referenced in the current workspace (reads `.dvc`, `dvc.yaml`, `dvc.lock` to find <abbr>outputs</abbr> and DVC-tracked data). ...

content/docs/command-reference/import-url.md
```diff
@@ -50,8 +50,8 @@ contains the corresponding local path in the <abbr>workspace</abbr>. It records
 enough metadata about the imported data to enable DVC efficiently determining
 whether the local copy is out of date.
 
-Note that `dvc repro` doesn't check or update import `.dvc` files, use
-`dvc update` on them to bring the import up to date from the external data
+Note that `dvc repro` doesn't check or update import `.dvc` files (see
+`dvc freeze`), use `dvc update` to bring the import up to date from the data
```
shcheklein, line 54: no need to reference `dvc freeze` here I think - it complicates this

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,13 @@ source.
   default file name: `<file>.dvc`, where `<file>` is the desired file name of
   the imported data (`out`).
 
-- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated when this option is used, only the import
-  metadata is saved to the `.dvc` file. `dvc commit <out>.dvc` can be used if
-  the data hashes are needed in the `.dvc` file, and to save existing data to
-  the cache.
+- `--no-exec` - create the import `.dvc` file but don't download `url`. The No
```
shcheklein, line 128: The No

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,13 @@ source.
...
+  hash values are calculated or saved to the `.dvc` file. Use `dvc update` to
```
shcheklein, line 129: what is the use case with `dvc update`?

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +125,13 @@ source.
...
+  finish the operation later, or `dvc commit` if the data has already been
+  downloaded manually.
+
+  > Note that there's no guarantee that the data source is valid until the
```
shcheklein, line 133: can we embed this ... like keep three - four words and embed into the paragraph

content/docs/command-reference/import-url.md
shcheklein: Still think that `No hash values are calculated or saved to the `.dvc` file.` will be very hard to understand. We don't even mention anywhere else about any hash calculations.

Can we describe this "no hash" using some human friendly message - like .dvc file is half-baked, or .dvc file is detached ... or .dvc file is a placeholder - URL is not checked, data is not downloaded, thus we don't save any information about it.

content/docs/command-reference/fetch.md
jorgeorpinel: Not sure I understand the Q. DVC will link them into the workspace as mentioned (that text didn't change in this PR BTW).

jorgeorpinel: I guess instead of referring to `dvc config` we can use `dvc checkout` here... Updated.

content/docs/command-reference/fetch.md
jorgeorpinel: OK. The "Without arguments" wasn't removed, just moved to after the diagram. The `git fetch` note is a different change (nor meant as a replacement).

jorgeorpinel: > What unasked question we are trying to resolve with this note?

At some point we decided to clarify that "DVC is not Git" in the analogous commands, which is why I added it. But now that it's here (even if I fixed it) I agree it seems unnecessary. At least it should be unnecessary... So I removed it for now.

content/docs/command-reference/fetch.md
jorgeorpinel: Updated.

content/docs/command-reference/fetch.md
jorgeorpinel: Like it. Updated here and in pull, push.

content/docs/command-reference/import-url.md
jorgeorpinel: OK, removed freeze ref.

content/docs/command-reference/import-url.md
jorgeorpinel: > We don't even mention anywhere else about any hash calculations.

Good point. I guess we can skip this note completely (implementation detail). Updated this and related files where the same problem existed.

> some human friendly message - like .dvc file is half-baked

I think that the next sentence, "Use `dvc update` to finish the operation later", implies that it's half-baked.

content/docs/command-reference/import-url.md
jorgeorpinel: Do you mean the use case for `import --no-exec` + `update`? To define a (series of) import instructions to be downloaded later/elsewhere.

This was mentioned in a 2nd paragraph before, but we decided it was too long.

content/docs/command-reference/import-url.md
jorgeorpinel: Embedded.

content/docs/command-reference/run.md
```diff
@@ -228,13 +228,12 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
   `command`.
 
 - `--no-exec` - create a stage file, but do not execute the `command` defined in
```
shcheklein, line 230: create a stage file -> create a stage

content/docs/command-reference/run.md
```diff
@@ -228,13 +228,12 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  it, nor cache dependencies or outputs (like with `--no-commit`, explained
-  below). DVC will also add your outputs to `.gitignore`, same as it would do
-  without `--no-exec`. Use `dvc commit` to force committing existing output file
-  versions to cache.
+  it. DVC will still add the outputs to `.gitignore`. Implies `--no-commit`
+  (explained below). If the outputs already exist in the workspace, use
+  `dvc commit` to force caching them.
```
shcheklein, line 233: it's not about force caching only

content/docs/command-reference/run.md
```diff
@@ -228,13 +228,12 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
   `command`.
 
 - `--no-exec` - create a stage file, but do not execute the `command` defined in
-  it, nor cache dependencies or outputs (like with `--no-commit`, explained
-  below). DVC will also add your outputs to `.gitignore`, same as it would do
-  without `--no-exec`. Use `dvc commit` to force committing existing output file
-  versions to cache.
+  it. DVC will still add the outputs to `.gitignore`. Implies `--no-commit`
```
shcheklein, line 231: It's too complicated - to go and read a reference ... then next sentence that you would get only if read the reference

content/docs/command-reference/push.md
```diff
@@ -33,14 +33,12 @@ use cases for these commands.
...
-remote.
+option is used. See `dvc remote` for more information on how to configure them.
 
-Without arguments, it uploads all files and directories missing from remote
-storage, found as <abbr>outputs</abbr> of the
-[stages](/doc/command-reference/run) or `.dvc` files present in the workspace.
-The `--all-branches`, `--all-tags`, and `--all-commits` options enable pushing
-multiple Git commits.
+Without arguments, it uploads all files and directories referenced in the
+current workspace (found in `dvc.yaml` and `.dvc` files) that are missing from
```
shcheklein, line 39: it sounds like _all_ files and directories, which is not true

content/docs/command-reference/import.md
```diff
@@ -103,12 +102,10 @@ repo at `url`) are not supported.
...
   > example below).
 
-- `--no-exec` - create the import `.dvc` file without actually downloading the
-  file or directory. E.g. if the file or directory already exists, this can be
-  used to skip the download. The data hash is not calculated when this option is
-  used, only the import metadata is saved to the `.dvc` file.
-  `dvc commit <out>.dvc` can be used if the data hashes are needed in the `.dvc`
-  file, and to save existing data to the cache.
+- `--no-exec` - create the import `.dvc` file but don't download the target data
+  (assumes that the data source is valid). Use `dvc update` to finish the
```
shcheklein, line 106: doesn't check if source is valid

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,12 @@ positional arguments:
 
 ## Description
 
-Downloads DVC-tracked files from remote storage into the cache of the project
-(without placing them in the <abbr>workspace</abbr>, like `dvc pull` would).
-This makes them available for linking (or copying) into the workspace (refer to
-[`dvc config cache.type`](/doc/command-reference/config#cache)).
+Downloads tracked data from remote storage into the <abbr>cache</abbr> (without
```
shcheklein, line 20: it's better to be neutral (data almost equals to datasets) - files and directories is better

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,12 @@ positional arguments:
...
+placing it in the <abbr>workspace</abbr> like `dvc pull`). This makes them
```
shcheklein, line 21: placing -> changing the workspace?

content/docs/command-reference/run.md
jorgeorpinel: True, it also creates/updates the lock file. Fixed (using general language). PTAL

jorgeorpinel: Although there might be a bug related to this... See iterative/dvc/issues/5034

content/docs/command-reference/run.md
jorgeorpinel: I'm not sure I follow this. By references here do you mean the --no-exec and --no-commit options?

jorgeorpinel: The alternative is to have duplicate explanation of --no-commit. Or maybe --no-commit should go before --no-exec — it's probably less rare actually.

jorgeorpinel: I removed the "Implies `--no-commit`" part for now, as it is, well... implied. No need to mention it, I guess (obviously since the stage didn't run no outputs were generated, etc.).

content/docs/command-reference/push.md
jorgeorpinel: Updated to "uploads the files and directories referenced in the
current workspace (found in all `dvc.yaml`..."

content/docs/command-reference/import.md
jorgeorpinel: Applied 👍

content/docs/command-reference/fetch.md
jorgeorpinel: Changed 👍

content/docs/command-reference/fetch.md
jorgeorpinel: Not sure. This parenthesis is just in case the user didn't catch the "into the cache" part so if they didn't catch that, "without changing the workspace" will be confusing. Safer to be explicit?

content/docs/command-reference/run.md
```diff
@@ -227,14 +227,13 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  versions to cache.
-
-  This is useful if, for example, you need to build a pipeline quickly first,
-  and run it all at once later.
+- `--no-exec` - create a stage, but do not execute the `command` defined in it.
+  DVC will still add the outputs to `.gitignore`. If the outputs already exist
+  in the workspace, use `dvc commit` to force DVC to track them (as if the stage
+  had run).
+
+  This is useful, for example, to define a full `dvc.yaml` pipeline first, and
```
shcheklein, line 235: This in this context reads as it belongs to the existing outputs ... it's confusing a bit

content/docs/command-reference/run.md
```diff
@@ -227,14 +227,13 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  it, nor cache dependencies or outputs (like with `--no-commit`, explained
-  below). DVC will also add your outputs to `.gitignore`, same as it would do
-  without `--no-exec`. Use `dvc commit` to force committing existing output file
-  versions to cache.
-
-  This is useful if, for example, you need to build a pipeline quickly first,
-  and run it all at once later.
+- `--no-exec` - create a stage, but do not execute the `command` defined in it.
+  DVC will still add the outputs to `.gitignore`. If the outputs already exist
+  in the workspace, use `dvc commit` to force DVC to track them (as if the stage
```
shcheklein, line 232: you _can_ use ... otherwise it feels that it is the most general case about this option.

content/docs/command-reference/fetch.md
```diff
@@ -17,22 +17,12 @@ positional arguments:
 
 ## Description
 
-Downloads DVC-tracked files from remote storage into the cache of the project
-(without placing them in the <abbr>workspace</abbr>, like `dvc pull` would).
-This makes them available for linking (or copying) into the workspace (refer to
-[`dvc config cache.type`](/doc/command-reference/config#cache)).
+Downloads tracked files and directories from remote storage into the
+<abbr>cache</abbr> (without placing them in the <abbr>workspace</abbr> like
+`dvc pull`). This makes them available for linking (or copying) into the
```
shcheklein, line 22: if we change this doc, we can now mention that it also makes these objects available to commands like `metrics show`, plots and others that read different commits

I would split `This makes them available for linking (or copying) into the` into a separate paragraph.

content/docs/command-reference/import-url.md
```diff
@@ -125,12 +124,9 @@ source.
...
   the imported data (`out`).
 
-- `--no-exec` - create `.dvc` file without actually downloading `url`. E.g. if
-  the file or directory already exists, this can be used to skip the download.
-  The data hash is not calculated when this option is used, only the import
-  metadata is saved to the `.dvc` file. `dvc commit <out>.dvc` can be used if
-  the data hashes are needed in the `.dvc` file, and to save existing data to
-  the cache.
+- `--no-exec` - create the import `.dvc` file but don't download `url` (assumes
+  that the data source is valid). Use `dvc update` to finish the operation
```
shcheklein, line 128: update it here as well

content/docs/command-reference/pull.md
```diff
@@ -57,11 +57,10 @@ The default remote is used (see `dvc remote default`) unless the `--remote`
...
 
-Without arguments, it downloads all files and directories missing from the
-project, found as <abbr>outputs</abbr> of the
-[stages](/doc/command-reference/run) or `.dvc` files present in the workspace.
-The `--all-branches`, `--all-tags`, and `--all-commits` options enable pulling
-multiple Git commits.
+Without arguments, it downloads all files and directories referenced in the
+current workspace (found in `dvc.yaml` and `.dvc` files) that are missing from
+the workspace. The `--all-branches`, `--all-tags`, and `--all-commits` options
+enable pulling multiple Git commits.
```
shcheklein, line 63: pulling multiple Git commits sounds like we pull actual Git commits

content/docs/command-reference/fetch.md
jorgeorpinel: > mention that it also makes these objects available to commands like metrics show, plots and others

Good idea. I checked `diff` and it doesn't need the cache so it's only `metrics/plots show/diff`. Added a note because once I wrote it it seemed sort of secondary (too specific to metrics and plots).

> I would split This

Since the new text is a note, I didn't split that for now.

jorgeorpinel: p.s. I opened iterative/dvc/issues/5039

content/docs/command-reference/run.md
jorgeorpinel: Applied.

content/docs/command-reference/run.md
jorgeorpinel: Hmmmm it's a separate paragraph so it shouldn't read like that. But I added "This option" at the beginning just in case.

content/docs/command-reference/fetch.md
jorgeorpinel: > also makes these objects available to commands like metrics show

Actually, it was already mentioned: "Here are some scenarios... To use comparison commands across different Git commits, for example dvc metrics" so I just added a `dvc plots diff` link there to make the bullet more complete (rolled back the note previously added).

content/docs/command-reference/pull.md
jorgeorpinel: Goof point and this also affects fetch, push, and status. Updated them and reordered these sentences a little in 564797d1. PTAL

content/docs/command-reference/fetch.md
```diff
@@ -56,14 +46,25 @@ Here are some scenarios in which `dvc fetch` is useful, instead of pulling:
...
   reason.
 
+Without arguments, it downloads all files and directories referenced in the
+current workspace (found in `dvc.yaml` and `.dvc` files) that are missing from
+the workspace. Any `targets` given to this command limit what to fetch. It
+accepts paths to tracked files or directories (including paths inside tracked
+directories), `.dvc` files, and stage names (found in `dvc.yaml`).
+
+The `--all-branches`, `--all-tags`, and `--all-commits` options enable fetching
+files/dirs referenced in multiple Git commits.
```
shcheklein, line 61: dirs -> directories?

content/docs/command-reference/fetch.md
```diff
@@ -56,14 +46,25 @@ Here are some scenarios in which `dvc fetch` is useful, instead of pulling:
...
+
 The default remote is used (see
-[`dvc config core.remote`](/doc/command-reference/config#core)) unless the
-`--remote` option is used.
+[`dvc config core.remote`](/doc/command-reference/config#core)) unless a
+specific one is given with `--remote`.
+
+> See `dvc gc` for ways to remove cached files.
```
shcheklein, line 67: it's a strange reference, why `gc`, why not checkout, not pull, other commands ... or remove?

content/docs/command-reference/run.md
```diff
@@ -227,14 +227,13 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  versions to cache.
-
-  This is useful if, for example, you need to build a pipeline quickly first,
-  and run it all at once later.
+- `--no-exec` - create a stage, but do not execute the `command` defined in it.
+  DVC will still add the outputs to `.gitignore`. If the outputs already exist
+  in the workspace, you can use `dvc commit` to force DVC to track them (as if
+  the stage had run).
+
+  This option is useful, for example, to define a full `dvc.yaml` pipeline
```
shcheklein, line 235: It all still feels strange. Sequence is not correct here to my mind. 

`If the outputs already exist in the workspace, you can use `dvc commit` to force DVC to track them (as if the stage had run).` - seems to be an essential part of the explanation of how this command operates. But in fact, it's just a similar use case as the that `This option is useful` describes

jorgeorpinel:
Oops I forgot about this PR! Addressing now...

content/docs/command-reference/fetch.md
jorgeorpinel: I've started being less formal with "directories". We have "dir(s)" here and there in places where the full word has already been spelled out a few times. WDYT?

shcheklein: Up to you on this one 👍

content/docs/command-reference/fetch.md
jorgeorpinel: checkout and pull are referenced near the top of the description. This note is the very last thing in the description.

shcheklein: I understand, but it feels very artificial to me ... probably because we don't do this in other place (like `dvc pull`?)

content/docs/command-reference/run.md
jorgeorpinel: OK I changed "If the outputs already exist..." for "You can use `dvc commit` to finish the operation." (like in the other commands and options that relate to commit per our other recent PR on commit-related texts).

shcheklein: but then you are loosing the point about existing outputs ...

(aslo, is it about other commands we put `This option is useful` into?)

content/docs/command-reference/fetch.md
jorgeorpinel: Yeah. It's not very necessary. Removed.

content/docs/command-reference/run.md
jorgeorpinel: The point about existing outputs is quite an edge case no? Not sure we need to go over it in docs. We similarly removed it from the `import` refs.

In the case of run/repro is not AS edge as with imports, but we are already mentioning it in https://dvc.org/doc/command-reference/commit#example-executing-stage-commands-without-dvc and in https://dvc.org/doc/user-guide/how-to/add-deps-or-outs-to-a-stage ("If the `data/raw.csv` or `data/validate` files already exist, we can use `dvc commit`". It's more about commit than about run/repro I think.

But if you prefer I can mention it here and in repro again.

jorgeorpinel: > is it about other commands we put This option is useful into?

Oops, didn't understand the question, sorry. We do have a "This is useful, for example..." paragraph here still though, if that's what you were referring to.

content/docs/command-reference/import-url.md
```diff
@@ -125,10 +124,9 @@ source.
   default file name: `<file>.dvc`, where `<file>` is the desired file name of
   the imported data (`out`).
 
-- `--no-exec` - create the import `.dvc` file without actually downloading
-  `url`. The data hash is not calculated when this option is used, only the
-  import metadata is saved to the `.dvc` file. You can use `dvc commit` to
-  finish the operation.
+- `--no-exec` - create the import `.dvc` file but don't download `url` (assumes
+  that the data source is valid). You can use `dvc commit` to finish the
```
shcheklein, line 128: If we remove mention of existing outputs, `dvc commit` looks artificial. It's not clear why would you run it.

I still think we can fit into one paragraph, mention existing files (avoid downloads, avoid run in case of `dvc run`

jorgeorpinel: Wait yeah good point, `commit` doesn't finish the operation (that would be `update` to download it). So I would argue to remove the reference to `commit` here. Again, is import + commit an actual case we want to document or just a side-effect?

jorgeorpinel: Separately, I wonder why repro doesn't have --no-exec and why import* don't have --no-commit. Is it worth asking on the core repo?

jorgeorpinel: > I still think we can fit into one paragraph

Done, PTAL.

content/docs/command-reference/commit.md
```diff
@@ -25,8 +25,8 @@ next.
 💡 For convenience, a pre-commit Git hook is available to remind you to
 `dvc commit` when needed. See `dvc install` for more info.
 
-Mainly, `dvc commit` provides a way to complete DVC commands that track data
-(`dvc add`, `dvc repro`, `dvc import`, etc.), when they have been used with the
+`dvc commit` provides a way to complete DVC commands that track data (`dvc add`,
```
shcheklein, line 28:
```
Sometimes after executing a stage, we realize that not all of its dependencies
  or outputs are defined in `dvc.yaml`. It is possible to
  [add the missing deps/outs](/docs/user-guide/how-to/add-deps-or-outs-to-a-stage),
  and `dvc commit` may be needed to finalize the remedy (see link).
```

contradicts this. (and probably the previous version as well)

content/docs/command-reference/run.md
```diff
@@ -228,15 +228,15 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-
-  This is useful, for example, if you need to define a pipeline quickly (and
-  perhaps share it with others) before executing anything, and run all its
-  stages at once later (with `dvc repro`).
+- `--no-exec` - write the stage to `dvc.yaml`, but do not execute the `command`.
+  DVC will still add the outputs to `.gitignore`, but they won't be cached or
+  recorded in `dvc.lock` (like with `--no-commit` below). You can use
+  `dvc repro` to finish the operation.
+
+  This is useful if you need to define a pipeline quickly before executing
```
shcheklein, line 236: what about `dvc commit` here? if outputs already exist?

shcheklein: also, I still think we are complicating this ... when I was saying one paragraph, I meant literally one paragraph for the whole description, including DVC-fying and existing ouputs case, dvc repro and dvc commit.

content/docs/command-reference/commit.md
jorgeorpinel: I don't see the contradiction, but maybe the way this is written gives that impression? Here's what each use case means:
1. finish add/run/repro --no-commit/exec
2. register and cache missing outputs that you forgot to add to dvc.yaml before (but now fixed)

jorgeorpinel: Clarified a little in ffff0a43 but it's not a big change.

content/docs/command-reference/run.md
jorgeorpinel: > what about dvc commit here?

Added link.

> I meant literally one paragraph for the whole description

OK, I merged the Ps.

content/docs/command-reference/commit.md
shcheklein: Contradiction here is that you we state "Mainly" and then clarify "Specifically", but point is completely different from the one in the "Mainly" section. To my mind either introductory section should be completely removed, or generalized if we keep it this way. wdyt?

content/docs/command-reference/run.md
```diff
@@ -230,15 +230,14 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  `dvc commit` to save any existing dep/out files to the cache and record their
-  hashes to the lock file.
-
-  This is useful, for example, if you need to define a pipeline quickly (and
-  perhaps share it with others) before executing anything, and run all its
-  stages at once later (with `dvc repro`).
+- `--no-exec` - write the stage to `dvc.yaml`, but do not execute the `command`.
+  DVC will still add the outputs to `.gitignore`, but they won't be cached or
+  recorded in `dvc.lock` (like with `--no-commit` below). You can use
+  `dvc repro` to finish the operation. This is useful if you need to define a
```
shcheklein, line 236: I think we can omit now ` You can use  dvc repro to finish the operation.`

I wonder if we also simplify the long sentence further somehow ...

content/docs/command-reference/commit.md
jorgeorpinel: Just to clarify, the the "Mainly" word was deleted in the lines above 🙂

> either introductory section should be completely removed, or generalized

The intro is "Stores the current contents..." which is already a general explanation. That p ends in "We explore the scenarios next" and the desc goes on to the main scenario ("provides a way to complete add/repro --no-commit etc."). Then we have a list of other scenarios but we wanted to use the word "Specifically" there.

So I'm a bit confused, why delete the main scenario? Do you mean move it down to the list? How about reinstating "Other" instead of "Specific"? That was my originally intended structure for this Description (again: intro, main scenario, list of other scenarios).

content/docs/command-reference/run.md
jorgeorpinel: > we can omit now You can use dvc repro

True, it was repeated. Removed. And I did find a way to simplify the text (dce4cc8b) so it's 20-30% shorter now.

TBH I would completely remove the mentions of --no-cache and `dvc commit` here too. It may be unnecessary (implied/obvious).

shcheklein (approved):

PR #1980, opened 2020-11-29
Update the cmd entry of the dvc.yaml file to add cmd as list option

ClementWalter:
Update the dvc.yaml documentation with the list possibility for the `cmd` entry.

Core PR [#4934](https://github.com/iterative/dvc/pull/4934)


jorgeorpinel:
> Fix #4373

I don't think this docs PR fixes a core issue 🙂 please link to the core PR this matches instead (don't link them). Thanks

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -151,7 +151,10 @@ stages:
 by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
 the possible following fields:
 
-- `cmd` (always present): Executable command defined in this stage
+- `cmd` (always present): Executable commands defined in this stage. This could be
+  either a single command or a list of commands that should be executed sequentially.
+  The failure of one of these commandes will result in the failure of the stage.
+  Note that each command is passed as is to the shell.
```
jorgeorpinel, line 157: Cool.

> The failure of one of these commandes will result in the failure of the stage.

Typo: commandes 

How does that work exactly? I assume it stops and doesn't execute the remaining commands but we should be explicit. Also, what happens to the outputs of previous successful commands?

jorgeorpinel: p.s. Feel free to use multiple paragraphs or a block quote (note) inside this bullet to make it easier to read since it's getting long.

jorgeorpinel (changes requested):
Left a suggestion ^ Also:

- The dvc.yaml example should show multiple commands, probably. Any other dvc.yaml examples throughout docs that could feature this?
- And what about the cmd refs? At least https://dvc.org/doc/command-reference/run needs an update, I think. Maybe `repro`

Thanks

jorgeorpinel:
Oh and BTW the checks are failing, please follow https://dvc.org/doc/user-guide/contributing/docs (we can also merge the restyled PR #1981 if needed but preferably, it shouldn't be necessary).

ClementWalter:
> Left a suggestion ^ Also:
> 
> * And what about the cmd refs? At least https://dvc.org/doc/command-reference/run needs an update, I think. Maybe `repro`
> 
> Thanks

Do you think that the `dvc run` should handle a command, with, for instance, `&&` so as to make it eventually a list in the `dvc.yaml`? I may be nice, but I am not sure it is the desired behavior because it requires some parsing of the user command, which, indeed, does not seem to be a preferred choice. @efiop, an idea about that?

efiop:
> Do you think that the dvc run should handle a command, with, for instance, && so as to make it eventually a list in the dvc.yaml? I may be nice, but I am not sure it is the desired behavior because it requires some parsing of the user command, which, indeed, does not seem to be a preferred choice. @efiop, an idea about that?

@ClementWalter I agree, that is not a desired behavior. List of commands could be only specified by hand by editing dvcfile, `dvc run` doesn't need to support it.

jorgeorpinel:
> dvc run doesn't need to support it.

Sounds good. But at the very least this will need checking the usage of the term "command" in repro (e.g. see -q option) and maybe other refs so it goes from singular to plural. Also it seems like the output of `dvc repro` will change slightly? There are a few examples that it would be ideal to update per this. Thanks

ClementWalter:
@jorgeorpinel I try to rephrase, hopefully it is clearer. I have added a use case for a list of commands, hopefully relevant as well. What do you think?

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -132,7 +132,9 @@ stages:
       - performance.json
   training:
     desc: train your model
-    cmd: python train.py
+    cmd:
+      - test -z "$(git status -s)"
+      - python train.py
```
jorgeorpinel, line 137: Thanks for the update! BTW quick Q. Will the old format also work still? (single command string, not an array)

jorgeorpinel, line 137: > test -z "$(git status -s)"

I'd maybe use something much simpler so no one gets confused by this. Maybe

```suggestion
      - echo "Train my model!"
      - python train.py
```

content/docs/command-reference/repro.md
```diff
@@ -253,8 +253,8 @@ If we now run `dvc repro`, we should see this:
 ```dvc
 $ dvc repro
 Stage 'filter' didn't change, skipping
-Running stage 'count' with command:
-        python process.py numbers.txt > count.txt
+Running stage 'count':
+$ python process.py numbers.txt > count.txt
```
jorgeorpinel, line 257: This is the actual output? Including a `$` character?

content/docs/user-guide/dvc-files-and-directories.md
ClementWalter: yes it does

content/docs/command-reference/repro.md
```diff
@@ -315,5 +315,5 @@ $ dvc dag
 ```
 
 > Note that using `dvc repro` without `--downstream` in the above example
-> results in the execution of the target (`count`), and the preceeding stages
+> results in the execution of the target (`count`), and the preceding stages
```
jorgeorpinel, line 318: Thanks for the typo fix 🙂

content/docs/user-guide/dvc-files-and-directories.md
ClementWalter: indeed I had to look at the command on google! But I was looking for a relevant use case, echoing while the command is printer as well seems a bit pointless. Maybe `git pull` ?

content/docs/command-reference/repro.md
ClementWalter: yes, it has changed from the previous "with command:  \t {command}" to this with line beginning with `$`

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: And would a list with a single item also work the same as the old single-value format?

ClementWalter: yes as well

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -151,7 +153,14 @@ stages:
 by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
 the possible following fields:
 
-- `cmd` (always present): Executable command defined in this stage
+- `cmd` (always present): Executable commands defined in this stage. Each
+  command is passed as is to the shell. The command can be either a single
+  command or a list of commands. In that latter case:
```
jorgeorpinel, line 158: > Each command is passed as is to the shell.

This suggestion removed that part because it seems obvious unless I'm not understanding a special meaning of this.

```suggestion
- `cmd` (always present): One or more commands executed by the stage
  (may contain either a single value, or a list).
```
> May need formatting

I'm not sure about the "may contain either a single value, or a list" part though, see Q in https://github.com/iterative/dvc.org/pull/1980#pullrequestreview-544313656 above.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -151,7 +153,14 @@ stages:
...
+  - the commands are run one after the other
+  - the failure of one given command will stop the run of the stage and raises
+    an error. Thus that all the commands before failure are well executed.
+    > in the `dvc.yaml` example above, the `python train.py` command will be
+    > executed only if the command before succeeds, _i.e_ if git is clear
```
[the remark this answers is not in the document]
jorgeorpinel: - [ ] But the details of how commands are executed should go somewhere in the `dvc repro` command reference instead of here. This is just a description of the field in dvc.yaml 🙂

jorgeorpinel (commented):
Thanks for all the details! Let's try to summarize this last bit about the `cmd` field though. And no worries, I can finish the writing at some point but I do have some questions about it before I can do that. My suggestions for now:

content/docs/user-guide/dvc-files-and-directories.md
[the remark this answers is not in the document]
ClementWalter: I like your suggestion, it is more concise while clearer

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Got it 👍

jorgeorpinel: Yeah I know my example goes too far into the useless side ha, sorry. `git pull` is probably not something you would do inside DVC stages, as Git manages the dvc.yaml file in the first place. `echo` is fine here if we can't come up with a better (simple) idea.

jorgeorpinel: > if we can't come up with a better (simple) idea

Are there examples somewhere in the core repo where this feature was originally proposed perhaps? I mean, what's the main use case for multiple commands in the first place?

content/docs/command-reference/repro.md
jorgeorpinel: So repro becomes some sort of background process that actually enters commands into the user's shell? Or we specifically print the `$` character + command string?

content/docs/user-guide/dvc-files-and-directories.md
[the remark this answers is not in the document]
jorgeorpinel: Thanks! This was marked resolved though? I don't see any changes yet. Unresolving...

content/docs/user-guide/dvc-files-and-directories.md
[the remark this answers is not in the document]
jorgeorpinel: Unresolving this too. I don't see any new changes above or in the repro ref.

content/docs/command-reference/repro.md
ClementWalter: we print it. Indeed it went from previous repeated "running command", to "> {command}" suggested by @skshetry to "$ {command}" because the ">" were confusing with possible ">" inside a command (like echo > file)

content/docs/user-guide/dvc-files-and-directories.md
ClementWalter: well, nothing mentionne in the issue at least

content/docs/user-guide/dvc-files-and-directories.md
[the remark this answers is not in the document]
ClementWalter: sorry i am used to resolve as soon I have taken them into account locally because the "outdated" from github should eventually show if something actually changed

content/docs/user-guide/dvc-files-and-directories.md
ClementWalter: maybe just the `git status` then?

content/docs/command-reference/repro.md
shcheklein: @ClementWalter @skshetry @efiop this can create a mess across all docs where we use `$` as a starting symbol for all shell commands. We'll need some mechanism to escape it, and even if we do we'll end up with confusing text anyway - mix of actual commands and outputs. Let's reconsider this please. Also out of curiosity how exactly is it related to the PR (multiple commands), I mean this particular change in UI/UX? what was the reason to change it?

shcheklein: Also, it's better to make it in a way that you can copy the whole line (command) by clicking on it. In docs engine made in way to ignore $, in case or suggested output in repro $ will go into command - it might be annoying.

skshetry: @ClementWalter, could you please rollback `$ command` to be something like the following:
```console
Running stage 'build':
  echo hello
  echo world
```

We can discuss a better UI later (`>` vs indent, etc.), but the `$` sign seems to be not possible in the docs and will make users confuse with the shell. 

Also could just do either of https://github.com/iterative/dvc/pull/4934#discussion_r533048696 as it's a 2.0 change. WDYT @shcheklein?

shcheklein: @skshetry I'm totally fine with `>` for now. Don't see any other good options to be honest how to keep it "copyable" and at them moment look different then output. 

A few possible options can be: 

- use delimiters (like testing systems do). Can be noisy
- use colors? Highlight commands somehow?

jorgeorpinel: Maybe similar as before, saying "with commands:" and using tabs? Other commands use tabs (e.g. `status` so that would be consistent). Up to you, just lmk please. so I can finish up the docs.

content/docs/user-guide/dvc-files-and-directories.md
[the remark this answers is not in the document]
jorgeorpinel: No problem, I guess we were both online at the same time. It happens.

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -132,7 +132,9 @@ stages:
       - performance.json
   training:
     desc: train your model
-    cmd: python train.py
+    cmd:
+      - echo "Train my model!"
+      - python train.py
```
jorgeorpinel, line 137: @skshetry can you think of a better example? I added an echo command here for now but a good example would show the usefulness of having several `cmd`s. TBH I'm not sure what the original motivation for this change was so I can't think of something off the top of my head, hoping you are more familiar. Thanks

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -151,7 +153,10 @@ stages:
 by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
 the possible following fields:
 
-- `cmd` (always present): Executable command defined in this stage
+- `cmd` (always present): One or more commands executed by the stage (may
+  contain either a single value, or a list). Commands are executed sequentially
+  until all are finished or until one of them fails (see
+  [`dvc repro`](/doc/command-reference/repro))
```
jorgeorpinel, line 159:
```suggestion
  [`dvc repro`](/doc/command-reference/repro) for details).
```

content/docs/user-guide/dvc-files-and-directories.md
```diff
@@ -151,7 +153,10 @@ stages:
 by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
 the possible following fields:
 
-- `cmd` (always present): Executable command defined in this stage
+- `cmd` (always present): One or more commands executed by the stage (may
+  contain either a single value, or a list). Commands are executed sequentially
+  until all are finished or until one of them fails (see
+  [`dvc repro`](/doc/command-reference/repro) for details).
```
jorgeorpinel, line 159: Note for self:

- [ ] The details of how commands are executed should go somewhere in the `repro` command reference.

content/docs/user-guide/dvc-files-and-directories.md
skshetry: What about this?
```suggestion
    cmd:
      - pip install -r requirements.txt
      - python train.py
```

content/docs/command-reference/repro.md
jorgeorpinel: p.s. not sure `>` is great since it can be part of the command, as above. Updated this with a tab for now.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: LGTM!

content/docs/command-reference/repro.md
```diff
@@ -254,7 +254,7 @@ If we now run `dvc repro`, we should see this:
 $ dvc repro
 Stage 'filter' didn't change, skipping
 Running stage 'count' with command:
-        python process.py numbers.txt > count.txt
+	python process.py numbers.txt > count.txt
```
jorgeorpinel, line 257: I wonder if we need an example with multiple commands in repro...

content/docs/command-reference/repro.md
skshetry: Implemented this on https://github.com/iterative/dvc/pull/5041.

> not sure > is great since it can be part of the command

For now, we add `>` instead of `$`. Let's discuss this on a different issue.

skshetry: > Let's discuss this on a different issue.

Created https://github.com/iterative/dvc/issues/5043 to discuss this further.

content/docs/user-guide/dvc-files-and-directories.md
jorgeorpinel: Done in 56252e34

content/docs/command-reference/repro.md
```diff
@@ -291,7 +296,7 @@ of only the target (`count`) and following stages (none in this case):
 ```dvc
 $ dvc repro --downstream count
 Running stage 'count' with command:
-        python process.py numbers.txt > count.txt
+	python process.py numbers.txt > count.txt
 Updating lock file 'dvc.lock'
```
skshetry, line 300: I didn't add "with command" back as it's unnecessary. And, the indent is replaced with ">" in the output.

skshetry (approved):
Rest looks good, just one issue with the output format.

ClementWalter:
I am not sure about the remaining changes to be brought to this PR.

content/docs/command-reference/repro.md
jorgeorpinel: > For now, we add > instead of $.

Actually right now it's a tab, why change it if we're not sure? I'd vote to roll back that change and yes to the discussion.

jorgeorpinel: Nvmd. Per https://github.com/iterative/dvc/discussions/5043#discussioncomment-198313 I have no strong opinion on this.

content/docs/command-reference/repro.md
jorgeorpinel: Id if "with command(s)" is unnecessary. Better to be explicit? What if someone just uses DVC for the first time and `pull` + `repro` are their first commands? They don't necessarily realize that a stage contains terminal commands. I say keep it (but make it plural when there's >1 cmd).

jorgeorpinel:
Hey no worries @ClementWalter I've assigned myself to finish up this one. Thanks!

content/docs/command-reference/repro.md
jorgeorpinel: Oh I guess you've already merged it your way in iterative/dvc/pull/4934 for now. OK

content/docs/command-reference/repro.md
```diff
@@ -2,7 +2,7 @@
 
 Reproduce complete or partial [pipelines](/doc/command-reference/dag) by
 executing commands defined in their [stages](/doc/command-reference/run) in the
-correct order. The commands to be executed are determined by recursively
+correct order. The stage commands to be executed are determined by recursively
```
jorgeorpinel, line 5:
```suggestion
correct order. The commands to be executed are determined by recursively
```

content/docs/command-reference/repro.md
```diff
@@ -245,7 +250,7 @@ If we now run `dvc repro`, we should see this:
 $ dvc repro
 Stage 'filter' didn't change, skipping
 Running stage 'count' with command:
```
jorgeorpinel, line 252:
```suggestion
Running stage 'count':
```

content/docs/command-reference/repro.md
```diff
@@ -282,7 +287,7 @@ of only the target (`count`) and following stages (none in this case):
 ```dvc
 $ dvc repro --downstream count
 Running stage 'count' with command:
```
jorgeorpinel, line 289:
```suggestion
Running stage 'count':
```

content/docs/command-reference/repro.md
```diff
@@ -244,8 +249,8 @@ If we now run `dvc repro`, we should see this:
 ```dvc
 $ dvc repro
 Stage 'filter' didn't change, skipping
-Running stage 'count' with command:
-        python process.py numbers.txt > count.txt
+Running stage 'count':
```
jorgeorpinel, line 252: O I removed "with command" from this file but there's another 4 instances/files with this text in the repo.

content/docs/command-reference/repro.md
```diff
@@ -244,8 +249,8 @@ If we now run `dvc repro`, we should see this:
...
+	python process.py numbers.txt > count.txt
```
jorgeorpinel, line 253:
```suggestion
> python process.py numbers.txt > count.txt
```

content/docs/command-reference/repro.md
```diff
@@ -281,8 +286,8 @@ of only the target (`count`) and following stages (none in this case):
 
 ```dvc
 $ dvc repro --downstream count
-Running stage 'count' with command:
-        python process.py numbers.txt > count.txt
+Running stage 'count':
+	python process.py numbers.txt > count.txt
```
jorgeorpinel, line 290:
```suggestion
> python process.py numbers.txt > count.txt
```

jorgeorpinel (approved):
I think this is mergeable pending https://github.com/iterative/dvc.org/pull/1980#pullrequestreview-550659656 and https://github.com/iterative/dvc.org/pull/1980#pullrequestreview-545845361 above, for a follow-up PR.

content/docs/command-reference/repro.md
jorgeorpinel: Actually I'm back to having an opinion 😆 — see https://github.com/iterative/dvc/discussions/5043#discussioncomment-198360

shcheklein (commented):
Thanks @ClementWalter , @jorgeorpinel , and @skshetry !

content/docs/command-reference/repro.md
jorgeorpinel: Moved to #2045

jorgeorpinel:
Looks like this is a 2.0 feature so I removed it from the `v1` branch in 365b7680.

PR #1982, opened 2020-11-29
cmd: finish list 1.0 update and clarify about local url

### LINKED ISSUE #1963: ref: example for `list .` (local repo as target)
The current examples in https://dvc.org/doc/command-reference/list only show how to view the contents of a tracked Git repository.

I think it's also worth denoting examples of how to view remote storages, like:

```
$ ls *.dvc
data.dvc

$ dvc list . data/some/dirs
data_1.csv
data_2.csv
data_3.csv
```

jorgeorpinel:
Closes #1963, I think.

### DIFF
--- a/content/docs/command-reference/get.md
+++ b/content/docs/command-reference/get.md
@@ -32,7 +32,8 @@ directory. (Analogous to `wget`, but for repos.)
 
 The `url` argument specifies the address of the DVC or Git repository containing
 the data source. Both HTTP and SSH protocols are supported (e.g.
-`[user@]server:project.git`). `url` can also be a local file system path.
+`[user@]server:project.git`). `url` can also be a local file system path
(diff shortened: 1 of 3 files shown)


shcheklein (approved):

shcheklein:
Looks good, not sure it's enough to close the ticket though (no example after all). But let's not prioritize this please. It's not even a 1.0 thingy.

jorgeorpinel:
Just wanted to get this closed since it already existed for some time. And while it doesn't add an example, is `list .` an intended/main use case? If so we can def. reopen #1963 (for later).

shcheklein:
> And while it doesn't add an example, is list . an intended/main use case. 

it can be, yes. Before you do a partial pull for example. Or listing a specific revision.

Issue #1975, opened 2020-11-27
repro: document `--glob` flag


skshetry:
`repro` got a yet another flag `--glob` to allow globbing of the stages. https://github.com/iterative/dvc/pull/4976 
Need to document it. :)

PR #1983, opened 2020-11-30
docs: doc repro --glob and update targets arg info.

skshetry:
Fixes https://github.com/iterative/dvc.org/issues/1975. The feature was introduced in https://github.com/iterative/dvc/pull/4976 which was released in 1.11.0.


content/docs/command-reference/repro.md
```diff
@@ -10,12 +10,23 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
...
 usage: dvc repro [-h] [-q | -v] [-f] [-s] [-c <path>] [-m] [--dry] [-i]
                  [-p] [-P] [-R] [--no-run-cache] [--force-downstream]
-                 [--no-commit] [--downstream] [--pull]
+                 [--no-commit] [--downstream] [--pull] [--glob]
                  [targets [targets ...]]
 
 positional arguments:
-  targets        Stage or path to dvc.yaml or .dvc file to reproduce. Using -R,
-                 directories to search for stages can also be given.
+  targets        Stage or path to dvc.yaml or .dvc file to reproduce.
```
skshetry, line 17: This does not look good. Can we move this content to a separate section, for example?
> #### The `target` option

But, I'd like to keep information about the ":" addressing though.

shcheklein: Yes! This is good, @skshetry - my 2cs - it deserves a section similar to the `### Parallel execution` below (and it should be before) ... and there should be a clear link to it in the first paragraph of the description ... or even in the synopsis?

shcheklein (approved):
This looks great @skshetry ... let's just find a place to put it.

For now section will do, but we can consider doing something standard here ... any good ideas that will make it easier to navigate for users?

content/docs/command-reference/repro.md
skshetry: I think it would be better to shift **Parallel execution** to the **Examples** section or further below and add a one-liner linking to it in the **Description**. 

For now, we could just add it to the Synopsis, but in the long run, we could move this `targets` section in the `Options` itself as done on [`git add`](https://git-scm.com/docs/git-add#Documentation/git-add.txt-ltpathspecgt82308203).

shcheklein: > I think it would be better to shift Parallel execution to the Examples

it might be, but let's not touch it in this PR - it's not related at the moment to my mind.

> For now, we could just add it to the Synopsis

you mean link or the whole text?

skshetry: > you mean link or the whole text?

I meant all of the text in Synposis, but I could also just add a link on the synopsis (it's not possible to link in the `usage`, is it?), and move the text to the **Options**. What do you think?

shcheklein: > it's not possible to link in the usage, is it

probably not at the moment :(

I would for now create a section in the description like you suggested. Options works fine as well. And let's create a ticket to figure out a better way to link things from the synopsis to their actual location.

content/docs/command-reference/repro.md
```diff
@@ -93,6 +96,25 @@ up-to-date and only execute the final stage.
 
 ## Options
 
+- `<target>` - Specify the stages to reproduce.
+
+  Target can be a name of the stage in the dvc.yaml file or a path to a .dvc or
```
shcheklein, line 101: here we can now put `dvc.yaml`. `.dvc`, etc into `inline code`

content/docs/command-reference/repro.md
```diff
@@ -93,6 +96,25 @@ up-to-date and only execute the final stage.
...
+  a dvc.yaml file. In case of a file, it will reproduce all of the stages
+  present in that file.
+
+  With `-R`, the target can be a directory to search for stages.
+
+  With `--glob`, the targets are used as a pattern to match stages in the
+  dvc.yaml file (example: `train-*`).
+
+  A stage from a dvc.yaml in a different directory can be specified using a path
```
shcheklein, line 110: I wonder if in this case it's easier to just show 5 or more examples with their meaning (instead of explaining `:` explicitly)?

skshetry: Yeah, it might be too much to read if you just need to get it done. But, I think it's good to also layout the _spec_ of it, so I added an example for each scenario.

shcheklein:
CI fail is unrelated, should be fixed in master

content/docs/command-reference/repro.md
shcheklein: @skshetry please, address this ^^

content/docs/command-reference/repro.md
shcheklein: yeah, spec is good, and I saw your examples (that's why I'm actually suggesting to do emphasis on them) the question is - should the spec be part of the examples explanation? this way most people with get the idea from what they see, almost immediately, and they can still read if needed.

content/docs/command-reference/repro.md
skshetry: Done.

content/docs/command-reference/repro.md
skshetry: That sounds good. Will make changes later in the day. Thanks.

content/docs/command-reference/repro.md
```diff
@@ -10,14 +10,17 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
...
-                 [targets [targets ...]]
+                 [--no-commit] [--downstream] [--pull] [--glob]
+                 [targets [<target> ...]]
 
 positional arguments:
-  targets        Stage or path to dvc.yaml or .dvc file to reproduce. Using -R,
-                 directories to search for stages can also be given.
+  targets        Stage or path to dvc.yaml or .dvc file to reproduce.
+                 If no targets are provided, it is assumed to be the
+                 dvc.yaml present in the current working directory.
```
jorgeorpinel, line 19: I don't see this change in https://github.com/iterative/dvc/pull/4976/files. Seems unrelated to the PR?

jorgeorpinel:
> CI fail is unrelated, should be fixed in master

This has been fixed. Please merge `master`, @skshetry.

content/docs/command-reference/repro.md
```diff
@@ -10,14 +10,17 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
...
 ```
 
+See [\<target\>](#options) for more details.
+
```
jorgeorpinel, line 23: Hmmm `targets` (plural BTW) is not an option right? It's an argument.

I think we should make a new H3 (`###`) section in the description instead.

jorgeorpinel: p.s. Probably before **Parallel stage execution**.

content/docs/command-reference/repro.md
```diff
@@ -93,6 +96,25 @@ up-to-date and only execute the final stage.
...
+  With `--glob`, the targets are used as a pattern to match stages in the
+  dvc.yaml file (example: `train-*`).
+
+  A stage from a dvc.yaml in a different directory can be specified using a path
+  to the dvc.yaml, followed by a colon `:`, followed by the name of the stage
+  (`models/dvc.yaml:prepare` for example).
+
+  Similarly, to pattern match the stages on a dvc.yaml in a different directory,
+  the pattern could follow after the path and the colon `:`
+  (`models/dvc.yaml:train-*` for example).
```
jorgeorpinel, line 116:
```suggestion
  With `--glob`, the targets are used as a pattern to match stages in the
  dvc.yaml file (example: `train-*`).

  To pattern match the stages on a dvc.yaml in a different directory, the pattern
  could follow after the path and colon `:` (`models/dvc.yaml:train-*` for
  example).
```

jorgeorpinel (commented):
I'd reorder these paragraphs a little so that everything --glob is together, like this:

content/docs/command-reference/repro.md
jorgeorpinel: > > easier to just show 5 or more examples with their meaning (instead of explaining :
>
> it might be too much to read if you just need to get it done

It's good to have the spec thoough. I can probably help summarize this text when it's ready. Consider instead adding an actual **Example** about wildcards in addition to this (and link to it from here).

jorgeorpinel (commented):
Thanks @skshetry this is really good.

~~Side-note guys: I'm a bit concerned about the `--glob` name for UX reasons. See https://github.com/iterative/dvc/pull/4976#issuecomment-736701953.~~ Moved to https://github.com/iterative/dvc/issues/4816#issuecomment-738217500

content/docs/command-reference/repro.md
skshetry: Sorry about this. This one was missing and as I already had a big change, so I thought why not I add a one-liner for this.

content/docs/command-reference/repro.md
skshetry: Now, `<target>` is a spec, whereas `targets` is a list of `<target>` here. 

I'd suggest to do similar to [git-add](https://git-scm.com/docs/git-add#Documentation/git-add.txt-ltpathspecgt82308203) and keep it in Options, but could do as you suggested. Let me know what you think.

content/docs/command-reference/repro.md
```diff
@@ -93,6 +96,25 @@ up-to-date and only execute the final stage.
...
+
+  Target can be a name of the stage in the `dvc.yaml` file or a path to a `.dvc`
+  or a `dvc.yaml` file. In case of a file, it will reproduce all of the stages
+  present in that file.
+
+  With `-R`, the target can be a directory to search for stages.
+
+  With `--glob`, the targets are used as a pattern to match stages in the
+  dvc.yaml file (example: `train-*`).
+
```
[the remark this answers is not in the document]
skshetry: I ordered them on the basis of what's more important. Running a different file is less important, but it's there. What do you think?

content/docs/command-reference/repro.md
jorgeorpinel: Fine by me as long as it's actually part of the help output now 🙂 

It's a strange change though, since `targets` is an optional arg I'm not sure its necessary to say what the default behavior is there, and probably inconsistent with other commands that also use `targets`.

jorgeorpinel: p.s. currently on DVC 1.10.1 I get `Stages to reproduce. 'dvc.yaml' by default.`

jorgeorpinel: Same for the master branch.

jorgeorpinel: Could we go for this?
```suggestion
  targets        Stage name or path to dvc.yaml or .dvc file to reproduce.
                 ./dvc.yaml by default
```

content/docs/command-reference/repro.md
jorgeorpinel: OK. But I don't see `<target>` (with `<` `>`) in the help output or any earlier indication that it's some sort of structure with a defined format (a spec). In the Git ref you link to, `<pathspec>` is in the Synopsis. In any case Git cmd refs aren't great inspiration 😋

I think that for now we should stick to a section in the Description @skshetry, but is this a new thing in 2.0? (Having certain specs) Do they have a new/special implementation? Do you know if there's a list of all our specs somewhere in the core repo perhaps? Thanks

content/docs/command-reference/repro.md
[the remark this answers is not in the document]
jorgeorpinel: You know what, it's OK. I think I'm going to try to combine several of these paragraphs to make it shorter when it's time to take this over so np. Thanks for adding all the details though, will be very helpful.

jorgeorpinel:
OK so it looks like we have all the info we need for this doc. The Q is whether you can help move stuff around and address some of the feedback above @skshetry please lmk so we can plan to finish this up when ready. Thanks

content/docs/command-reference/repro.md
skshetry: > but is this a new thing in 2.0? (Having certain specs)

except for the glob, this was there since 1.0. The spec is not just applicable to repro, but most of the commands like commit, checkout, push, pull, etc. Some of them also accept output file path in addition to that spec. But, adding them in all other places is not that feasible and/or important, so only added on `repro`.

> But I don't see <target> (with < >) in the help output

It's hard to add these details on the help output and is something that should be in the manpage.

skshetry:
@jorgeorpinel, this should be ready, I don't have anything to add. 
I was trying to show examples before the spec explanation like how Ivan mentioned above, but I am not sure how to do that, so I'm skipping it.

shcheklein:
@jorgeorpinel if we need something extra here - can we do it quick (w/o updating a lot of other stuff along the way :) ) and merge?

jorgeorpinel:
There are several reviews/comments open that haven't been addressed above. The example is not critical at all but the structuring is not ready (as described in the reviews). Lmk if I should take this over @shcheklein. Thanks

content/docs/command-reference/repro.md
jorgeorpinel: Hmmm.. I'm not sure what you mean by specs, but I still think we should stick to a section in the Description like we already do if nothing changed specially form 1.0 to 2.0.

shcheklein: @jorgeorpinel I think @skshetry has a point here. A section in description is not convenient/enough. There should be some simple description for all options. Ideally we should "teach" engine to link to them from the synopsis. I think that's what @skshetry was tryin to imitate, it's just engine is not sophisticated enough to support that yet.

shcheklein:
@jorgeorpinel I think we can def take it over. I suggest we don't do to many changes though (e.g. let's not do a separate section in the Description -it doesn't feel 100% right in this case). Rather let's create a ticket to improve the engine itself.

Same with examples - copy edits are fine, otherwise it's good enough and I doubt we need to spend too much time on further iterations.

content/docs/command-reference/repro.md
jorgeorpinel: Hmmm a different section for **Arguments** then? I guess that could be clearer than explaining them in the description... Not sure, I can try and then we can decide. But it seems like that's not quite in scope for this PR. I'd prefer to use the existing format for now and open a separate issue.

jorgeorpinel:
OK, taking this over.

> I suggest we don't do to many changes though (e.g. let's not do a separate section in the Description

From my perspective, keeping that new section is the big change here (kind of out of scope for the main purpose of this PR) as I understand it.

> let's create a ticket to improve the engine itself

Agree but is it an engine question? Or just an improvement of our cmd. ref. template?

shcheklein:
> keeping that new section is the big change here

not sure I follow, what is the new section? my point was `e.g. let's not do a separate section in the Description` - _not_ do a separate section

> Agree but is it an engine question? Or just an improvement of our cmd. ref. template?

it's a mix - we need a place to put those arguments (and I'm fine to put them into the options for now), we need a way to link with synopsis, etc

content/docs/command-reference/repro.md
```diff
@@ -10,14 +10,17 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
 ```usage
 usage: dvc repro [-h] [-q | -v] [-f] [-s] [-m] [--dry] [-i]
                  [-p] [-P] [-R] [--no-run-cache] [--force-downstream]
-                 [--no-commit] [--downstream] [--pull]
-                 [targets [targets ...]]
+                 [--no-commit] [--downstream] [--pull] [--glob]
+                 [targets [<target> ...]]
```
jorgeorpinel, line 14: I don't see this `<targets>` in the help output on the latest master branch @skshetry was that something you added or is it how it will look in 2.0? Thanks
```suggestion
                 [targets [targets ...]]
```

content/docs/command-reference/repro.md
```diff
@@ -10,14 +10,17 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
...
 
 positional arguments:
-  targets        Stage or path to dvc.yaml or .dvc file to reproduce. Using -R,
-                 directories to search for stages can also be given.
+  targets        Stage or path to dvc.yaml or .dvc file to reproduce.
+                 If no targets are provided, it is assumed to be the
+                 dvc.yaml present in the current working directory.
```
jorgeorpinel, line 19: Similar question @skshetry. I don't see this in master so I will change it to what's actually in the core repo unless you have more details. Thanks

content/docs/command-reference/repro.md
jorgeorpinel: Continuing discussion in https://github.com/iterative/dvc.org/pull/1983#pullrequestreview-550965610

content/docs/command-reference/repro.md
jorgeorpinel: In fact I think I'm going to go for my earlier suggestion (https://github.com/iterative/dvc.org/pull/1983#discussion_r535430806):
```suggestion
  targets        Stage name or path to dvc.yaml or .dvc file to reproduce
                 (./dvc.yaml by default)
```
(and send a corresponding update to core)

jorgeorpinel: Hmmmm. Or "Limit command scope to these stage names." for consistency with all the other cmd refs that have a `targets` argument.

shcheklein:
That's what I think @skshetry was trying to achieve - https://git-scm.com/docs/git-commit#Documentation/git-commit.txt-ltpathspecgt82308203 .. also they don't use a separate section for this ... probably because it's too much to have a section for one argument (in most cases it's one)

Also, it's very that they have an anchor for each option.

Unfortunately, they don't have links in the synopsis.


Again, point here - it's fine to it quick here, within one section and create a ticket for all these improvements.

content/docs/command-reference/repro.md
```diff
@@ -91,6 +91,27 @@ same time (e.g. in separate terminals). After both finish successfully, you can
 then run `dvc repro train`: DVC will know that both branches are already
 up-to-date and only execute the final stage.
 
+## Arguments
+
+- `targets` - Stage name(s) to reproduce (`./dvc.yaml` by default)
+
+  Target can be a name of the stage in the `dvc.yaml` file or a path to a `.dvc`
+  or a `dvc.yaml` file. In case of a file, it will reproduce all of the stages
+  present in that file.
```
jorgeorpinel, line 100: > or a path to a `.dvc` (file)
> In case of a file, it will reproduce all of the stages present in that file.

what about .dvc files? They don't contain stages. What's reproduced in that case? I guess it depends on the contents (e.g. import .dvc files are always frozen last time I checked). I will double check all this this but any notes are appreciates @skshetry.

content/docs/command-reference/repro.md
skshetry: `target` is a spec, so either you emphasize it with `<target>` as:
```
                 [targets [<target> ...]]
```

or, just with the `target`. `targets [targets ...]` don't make sense.

> I don't see this <targets> in the help output on the latest master branch

I don't have any problem with changing this on the DVC side. The only reason I made `<target>` here is to emphasize it's more of a pattern/syntax. Also, note that this kind of syntax (except glob) is supported by `push/pull/checkout` commands, but as we are moving away from using stage name in those commands, I only added them in the `repro` where it makes the most sense.

> is it how it will look in 2.0?

This has been the case since 1.0 (except glob of course).

content/docs/command-reference/repro.md
skshetry: > I don't see this in master so I will change it to what's actually in the core repo unless you have more details.

The previous `Using -R` wasn't in the DVC help output. I am not sure if they even should be, as the command reference and the help output has different objectives (check `git add -h` output). Like I said above, this is more suited for docs, rather than a help output.

content/docs/command-reference/repro.md
skshetry: > What about .dvc files?

We still support `.dvc` files as a stage (and, internally every `.dvc` file including `add/import`-ed ones are a stage). So, this must have been generalized. 

> They don't contain stages

Each `.dvc` file is a stage on itself, so it contains one stage.

> What's reproduced in that case?

The non-frozen imports are re-imported/updated. By default, they are frozen, but you can remove the `frozen` keyword.
For the `dvc add`-ed files, they are just checked out.
For the old `.dvc` files with commands, they are run. If we have a run cache, they are checked out without running.
Same in the case of stages in `dvc.yaml` files.

skshetry: I am not sure what information should be or not be in the docs. But, that's how the `dvc repro` works.

content/docs/command-reference/repro.md
jorgeorpinel: This part (the Synopsis) is a copy-paste of the help output (minus some details that would be repetitive). Sometimes it gets outdated, which is probably why the previous version also didn't match core.

> And again, Git output and docs aren't great. We do not try to copy the Git UI or git-scm.com at all.

> this is more suited for docs, rather than a help output

I disagree. At some point the UI and docs should intersect, and I think it's here. If it's clear enough for docs, it should be used in the UI as well. It also helps with consistency.

jorgeorpinel:
@shcheklein 

> > keeping that new section is the big change here
>
> not sure I follow, what is the new section?

Sorry I meant adding the long text for the one argument in the options section. That in itself is a big change.

> > is it an engine question?
>
> we need a place to put those arguments... link with synopsis... anchor for each option

OK, got it. I created #2024 for that, please prioritize as needed.

> That's what I think **skshetry** was trying to achieve...

Yes, he mentioned Git UI and references a few times. But I really don't think we want to emulate Git much, their UI and docs are pretty bad.

> it's fine to do it quick here, within one section and create a ticket

I do agree it's a good idea to have some special place for arguments. I'm not sure we should figure it out in this PR, but we can.

Working on it...

jorgeorpinel:
Should I also address #2023 here along the way? That report came out of this PR.

content/docs/command-reference/repro.md
jorgeorpinel: > internally every .dvc file including add/import-ed ones are a stage
> Each .dvc file is a stage on itself, so it contains one stage.

This was kind of true in DVC 0.x but in actuality .dvc files are not stages right? Cc @shcheklein 

> p.s. whether they're called `stage` internally somewhere in the code is probably not an indication that that's what they are.

jorgeorpinel: > non-frozen imports are re-imported/updated. By default, they are frozen, but you can remove the frozen

Right I forgot you could do that, thanks for the info.

> dvc add-ed files, they are just checked out

Hmmm I think that if they have changes they are re-added by repro.

shcheklein: > Each .dvc file is a stage on itself, so it contains one stage.

my take on this - this is an implementation detail. We should be going away from this. Even imports are not stages to my mind.

I would argue that if there are other commands to "reproduce" .dvc files we should deprecate and do not complicate repro with them.

shcheklein:
> Should I also address #2023 here along the way? That report came out of this PR.

please no, it's taking too long already

content/docs/command-reference/repro.md
jorgeorpinel: Agree but that's a disccusion for core. For now I'll keep .dvc files and stages separate comceptually, and try to include only the minimal info about using repro for .dvc files.

shcheklein: @jorgeorpinel my point is that we can start deprecating it. It means we should be paying too much attention or include into docs them. Unless I'm missing something and we indeed have a case for `dvc repro .dvc`

jorgeorpinel: > dvc add-ed files, they are just checked out

@skshetry I checked this and it doesn't seem that's the case. It gives me this error:

```
ERROR: failed to reproduce 'data.dvc': missing data 'source': data
```
> The file contents are cached.

The docs now read "`dvc repro` does not run `dvc fetch`, `dvc pull` or `dvc checkout`" which seems accurate so I'll leave that as is.

content/docs/command-reference/repro.md
```diff
@@ -1,32 +1,37 @@
 # repro
 
-Reproduce complete or partial [pipelines](/doc/command-reference/dag) by
-executing commands defined in their [stages](/doc/command-reference/run) in the
-correct order. The commands to be executed are determined by recursively
-analyzing dependencies and <abbr>outputs</abbr> of the target stages.
+Reproduce complete or partial <abbr>pipelines</abbr> by executing commands
```
shcheklein, line 3: why do we change this? is it relevant? can we do it as part of another PR?

jorgeorpinel:
OK I pushed my updates. `targets` is now a section just before the Options. It's not in options because a) it's not an option (listed under "positional arguments" in the Synopsys) and b) the text is too long and wouldn't be well fit for a singe bullet in the options list — in fact it contains its own bullet list. Did some other changes to the description and options that I thought were related and to try to make everything more compact.

content/docs/command-reference/repro.md
```diff
@@ -1,32 +1,37 @@
...
+> See [`targets`](#targets-argument-optional) for more details.
+
 ## Description
 
-`dvc repro` provides a way to regenerate data pipeline results, by restoring the
-dependency graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))
-implicitly defined by the stages listed in `dvc.yaml`. The commands defined in
-these stages can then be executed in the correct order, reproducing pipeline
-results.
+Provides a way to regenerate data pipeline results, by restoring the dependency
```
shcheklein, line 25: this doesn't look relevant, let's remove this change from this PR please

content/docs/command-reference/repro.md
jorgeorpinel: Just changed a link to `<abbr>`.

jorgeorpinel: But OK rolling back.

content/docs/command-reference/repro.md
```diff
@@ -1,32 +1,37 @@
...
+graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)) implicitly
+defined by the stages listed in `dvc.yaml`. The commands defined in these stages
+are then be executed in the correct order.
+
+For stages with multiple commands (having a list in the `cmd` field), commands
+are run one after the other in the order they are defined. The failure of any
+command will halt the remaining stage execution, and raises an error.
 
-> Pipeline stages are defined in a `dvc.yaml` file (either manually or by using
```
shcheklein, line 29: does not look relevant, please let's scope this PR to the `--glob` thingy

content/docs/command-reference/repro.md
jorgeorpinel: Done.

content/docs/command-reference/repro.md
```diff
@@ -95,6 +91,39 @@ same time (e.g. in separate terminals). After both finish successfully, you can
 then run `dvc repro train`: DVC will know that both branches are already
 up-to-date and only execute the final stage.
 
+### `targets` argument (optional)
```
shcheklein, line 94: we complicated it way too much, let's simplify instead and make it part of the options sections like we agreed please

shcheklein (changes requested):
sorry, @jorgeorpinel , we need to scope this back to the `--glob` and simplify it.

content/docs/command-reference/repro.md
jorgeorpinel: Done.

content/docs/command-reference/repro.md
jorgeorpinel: Hmmm. This is (losely) related to the `targets` argument, which is what this PR is actually on at least since I took it over.

jorgeorpinel: But OK, rolled back.

content/docs/command-reference/repro.md
jorgeorpinel: As you prefer. I put it in Options now but I don't think it makes sense.

shcheklein: it doesn't make sense if we make it very long, putting it into the arguments/options force you to keep it short more or less and this is good to my mind. It's better to have a few simple examples and a few words about them then asking user to read a long spec (that might exist, but should be complimentary)

shcheklein: also, I doubt that user will care much about it's specific location (like "it looks bad because it's technically not an option " - I doubt that someone will think like this). 

I agree, though that it's better to revisit it, that's why we created a ticket - to support them properly

jorgeorpinel: OK, I simplified `targets` and `--glob` (which I put right after targets in the Options list).

content/docs/command-reference/repro.md
```diff
@@ -10,21 +10,26 @@ analyzing dependencies and <abbr>outputs</abbr> of the target stages.
...
 ## Description
 
 `dvc repro` provides a way to regenerate data pipeline results, by restoring the
 dependency graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))
 implicitly defined by the stages listed in `dvc.yaml`. The commands defined in
-these stages can then be executed in the correct order, reproducing pipeline
-results.
+these stages are then be executed in the correct order.
+
+For stages with multiple commands (having a list in the `cmd` field), commands
```
shcheklein, line 30: not relevant?

content/docs/command-reference/repro.md
```diff
@@ -37,30 +42,21 @@ and <abbr>caches</abbr> the pipeline's <abbr>outputs</abbr> along the way.
 💡 For convenience, a Git hook is available to remind you to `dvc repro` when
 needed after a `git commit`. See `dvc install` for more details.
 
+[Stage](/doc/command-reference/run) outputs are deleted from the
+<abbr>workspace</abbr> before executing the stage commands that produce them.
 `dvc repro` does not run `dvc fetch`, `dvc pull` or `dvc checkout` to get data
 files, intermediate or final results (except if the `--pull` option is used).
 
-By default, this command checks all [pipeline](/doc/command-reference/dag)
```
shcheklein, line 43: all not relevant?

content/docs/command-reference/repro.md
```diff
@@ -105,15 +126,21 @@ up-to-date and only execute the final stage.
   recursive search for changed dependencies. Multiple stages are executed
   (non-recursively) if multiple stage names are given as `targets`.
 
-- `-R`, `--recursive` - determines the stages to reproduce by searching each
-  target directory and its subdirectories for stages (in `dvc.yaml`) to inspect.
-  If there are no directories among the targets, this option is ignored.
+- `-R`, `--recursive` - looks for `dvc.yaml` files to reproduce in any
```
shcheklein, line 129: not relevant?

content/docs/command-reference/repro.md
```diff
@@ -105,15 +126,21 @@ up-to-date and only execute the final stage.
...
+  directories given as `targets`, and in their subdirectories. If there are no
+  directories among the targets, this option has no effect.
 
 - `--no-commit` - do not store the outputs of this execution in the cache
   (`dvc.yaml` and `dvc.lock` are still created or updated); useful to avoid
   caching unnecessary data when exploring different data or stages. Use
   `dvc commit` to finish the operation.
 
+- `-p`, `--pipeline` - reproduce the entire pipelines that the `targets` belong
```
shcheklein, line 138: not relevant? (I know it might be painful, but otherwise we iterate too long on all these changes). Especially when team is involved, let's keep it to the point and iterate fast please

content/docs/command-reference/repro.md
```diff
@@ -97,6 +93,31 @@ up-to-date and only execute the final stage.
...
+  - For **`dvc.yaml` files**, their [pipeline(s)](/doc/command-reference/dag)
+    are checked for changes, and reproduced as needed (explained in the command
+    [description](#description) above). E.g. `dvc repro pipes/linear/dvc.yaml`
+
+  - **Stage names** must be defined in `./dvc.yaml`. E.g.
+    `dvc repro train-vision`. Stages in other `dvc.yaml` files can be given
+    using by using a colon `:` following the path to that file. E.g.
+    `models/dvc.yaml:prepare`
+
+  - Files and directories tracked by **`.dvc` files** given as `targets` are
```
shcheklein, line 109: let's skip this I think?

content/docs/command-reference/repro.md
```diff
@@ -97,6 +93,31 @@ up-to-date and only execute the final stage.
 
 ## Options
 
+- `targets` (optional argument) - one or more file or directory paths (to `.dvc`
```
shcheklein, line 96: -R is not covered anymore?

shcheklein: --glob is not covered anymore

content/docs/command-reference/repro.md
jorgeorpinel: This is not being introduced here, it comes from a recent merge to master. Given some changes to the description this paragraph seems like it was moved which is why this part is green.

jorgeorpinel: p.s. there were merge conflicts I had to resolve which also cause some of these apparent changes + as you know our 80-line formatting tends to exaggerate the amount of changes.

content/docs/command-reference/repro.md
jorgeorpinel: This paragraph is redundant (both with the intro as well as with the new targets text) and slightly wrong (but that will be corrected for #2024). Removing it aims to make the Description shorter so that the whole doc doesn't grow too much as we introduced the targets text.

jorgeorpinel: But I can easily stash the change if you prefer and leave it for later, lmk.

content/docs/command-reference/repro.md
jorgeorpinel: So the reason some options are changed here is because they involve the `targets` argument.

TBH I'm confused with the idea that this Pr is about glob when it was always about targets even more than about --glob. But would you like me to split into 2 PRs?

content/docs/command-reference/repro.md
jorgeorpinel: Yes! I would like that too but are you sure? The help output (and Synopsis above) mention .dvc files so if we don't explain it here, it'd be up to the user to assume what will happen.

content/docs/command-reference/repro.md
shcheklein: hmm, sorry, I still don't get it. If it comes from master, could you rebase/merge master? Otherwise this change adds a paragraph.

content/docs/command-reference/repro.md
```diff
@@ -37,30 +42,21 @@ and <abbr>caches</abbr> the pipeline's <abbr>outputs</abbr> along the way.
 💡 For convenience, a Git hook is available to remind you to `dvc repro` when
 needed after a `git commit`. See `dvc install` for more details.
 
+[Stage](/doc/command-reference/run) outputs are deleted from the
```
shcheklein, line 45: same here - it's not related?

content/docs/command-reference/repro.md
shcheklein: Yes, let's make PRs focused if possible.

shcheklein: I expect some changes here, but it probably should be only the paragraph about targets?

content/docs/command-reference/repro.md
```diff
@@ -37,30 +42,21 @@ and <abbr>caches</abbr> the pipeline's <abbr>outputs</abbr> along the way.
...
-specifying stages as `targets`, or by using `--single-item`, among other
-options.
+specifying specific reproduction [`targets`](#options), or by using certain
+command [options](#options), such as `--single-item`.
 
 > Note that stages without dependencies are considered _always changed_, so
 > `dvc repro` always executes them.
 
-It stores all the data files, intermediate or final results in the
+`dvc repro` saves all the data files, intermediate or final results into the
```
shcheklein, line 57: Not related?

content/docs/command-reference/repro.md
shcheklein: Synopsis is a copy/paste from DVC which should probably change, so, let's not complicate. The point is not cover everything to the very small detail - the point is to focus on important stuff and convey it in the simplest possible format (e.g. people don't read long texts, unless they want to dive in)- thus short self-explainable examples are better

content/docs/command-reference/repro.md
shcheklein: but you do at least a few changes at once - you move two options, you also introduce some changes to one of them. Also the change is about `dvc dag`, not even dvc repro in this case

content/docs/command-reference/repro.md
jorgeorpinel: Are we talking about the "For stages with multiple commands" paragraph? It was already there, look at old line 49: https://github.com/iterative/dvc.org/pull/1983/files#diff-6c1f3192f09e2722ba169e9fa219b3b5158bbafa470b382c2d6135db7aa1e20dL49

content/docs/command-reference/repro.md
jorgeorpinel: OK, I reverted this. I'm afraid that if we just keep adding paragraphs without reviewing the context we'll end up with lots of redundancy and a text that may be unnecessary long.

content/docs/command-reference/repro.md
jorgeorpinel: This is directly related to `repro targets`. It used to read "determines the stages ... by searching ... for stages". Isn't this a good opportunity to fix that? TBH I already have a mountain of stashed changes so anything I can't get into existing PRs will probably get lost until we happen to notice it again 🙁

content/docs/command-reference/repro.md
jorgeorpinel: OK here I was confused indeed, not about `targets` per se. Reverted move.

content/docs/command-reference/repro.md
jorgeorpinel: But TBH I don't think I intended to move this one so I assume it has something to do with a merge conflict I had to solve during this PR. This is not about targets, it's about multiple commands.

jorgeorpinel: Reverted this move for now. Resolving

content/docs/command-reference/repro.md
jorgeorpinel: Agree. Removed info about .dvc files from this ref.

content/docs/command-reference/repro.md
jorgeorpinel: Hi! OK I rewrote this as we agreed privately. PTAL

content/docs/command-reference/repro.md
jorgeorpinel: Reverted.

content/docs/command-reference/repro.md
```diff
@@ -128,7 +153,8 @@ up-to-date and only execute the final stage.
   to. Use `dvc dag <target>` to show the parent pipeline of a target.
 
 - `-P`, `--all-pipelines` - reproduce all pipelines for all `dvc.yaml` files
-  present in the DVC project.
+  present in the DVC project. There's no need to specify `targets` when using
+  this option.
```
jorgeorpinel, line 157: I'm just not 100% sure about this note. Will double check

jorgeorpinel: OK I checked and adjusted this.

content/docs/command-reference/repro.md
```diff
@@ -97,17 +99,40 @@ up-to-date and only execute the final stage.
...
+  - With `-R` you can provide directory paths to search for `dvc.yaml` files in,
+    recursively.
+  - With `--glob`, you can use special patterns (using wildcards) to match
+    groups of stage names.
+
+- `-R`, `--recursive` - looks for `dvc.yaml` files to reproduce in any
+  directories given as `targets`, and in their subdirectories. If there are no
+  directories among the targets, this option has no effect.
+
+- `--glob` - causes the `targets` to be interpreted as wildcard
```
jorgeorpinel, line 122: Added -R and --glob to `targets` but also moved those options up right next to targets, so I still think it seems a bit redundant.

shcheklein: Suggestion:

Keep only

```
Different things can be provided as targets depending on the flags used (more
  details in each option), namely:
```

then a list of examples:

- `dvc reporo -R pipeline` - will look recursively into ....
- `dvc reporo --glob ...` - will execute stages that match pattern
- `dvc repro` - will find dvc.yaml(s) and execute all stages

it's way simpler to grasp the idea, no need to read long long text (if some details are need let's try very hard to squeeze them into each item above)

jorgeorpinel: `reporo`? 😆 Sounds funny

OK good idea, applied! PTAL.

Note that issue #1614 is pretty evident this way, should we prioritize it?

image

shcheklein (approved):

content/docs/command-reference/repro.md
```diff
@@ -97,17 +99,35 @@ up-to-date and only execute the final stage.
...
 
-- `-f`, `--force` - reproduce a pipeline, regenerating its results, even if no
-  changes were found. This executes all of the stages by default, but it can be
-  limited with the `targets` argument, or the `-s`, `-p` options.
+- `targets` (optional command argument) - what to reproduce (the pipeline(s) in
+  `./dvc.yaml` by default). Different things can be provided as targets
+  depending on the flags used (more details in each option). Examples:
+
+  - `dvc repro linear/dvc.yaml`: Specific `dvc.yaml` file to reproduce
+  - `dvc repro -R pipelines/`: Directory path to explore recursively for for
```
skshetry, line 107:
```suggestion
  - `dvc repro -R pipelines/`: Directory path to explore recursively for
```

jorgeorpinel: Oops thanks! Want to send a quick typo fix PR since this is merged?

PR #1989, opened 2020-12-01
cmd: update commit-related info.

### LINKED ISSUE #186: cmd ref: add a few lines about cache and no-commit in `dvc run`
`--no-commit` + `dvc commit` introduced a new way of iterating, it's an important to put a paragraph in the description or even a separate section.

jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/pull/1978#commitcomment-44682818

- Includes some misc. updates carried from #1978
- Happens to fix #186

> Also rel. iterative/dvc/issues/5000

Note for self:
- [ ] Finalize the commit part of iterative/dvc/pull/4999 when this is done.


content/docs/command-reference/commit.md
```diff
@@ -17,65 +17,52 @@ positional arguments:
...
-still useful, but keep in mind that DVC can't guarantee reproducibility in those
-cases.
+This command can be useful for several scenarios, when
+[stages](/doc/command-reference/run) are under development, or to force-update
+the cache without having to execute stage commands again (avoiding `dvc repro`).
+
+> Note that this does not register project versions. That can be done with
+> `git commit` (if using Git).
+
+Normally, `dvc repro` and `dvc run` finish up with the same steps as `dvc add`,
```
jorgeorpinel, line 27: > Are there any significant important fixes/changes here?

@shcheklein yes. The whole structure changed to emphasize the main use (development process i.e. a companion for add/run/repro --no-commit) in the first half. The 2nd half of the desc. goes over the other scenarios where this might be useful.

All the explanations were simplified and implementation details were concentrated in a single paragraph or 2.

content/docs/command-reference/commit.md
```diff
@@ -17,65 +17,52 @@ positional arguments:
...
+- It's also possible to execute stage commands by hand (without `dvc repro`), or
+  to manually modify their output files or directories. Use `dvc commit` to
+  register the changes with DVC once you're done.
+
+  > Note that `dvc unprotect` (or removing the outputs) is usually required
+  > before rewriting files/dirs tracked by DVC.
+
+Note that it's best to try avoiding these scenarios, where the cache,
+`dvc.lock`, and `.dvc` files are force-updated. DVC can't guarantee
+reproducibility in those cases.
```
jorgeorpinel, line 65: > The 2nd half of the desc. goes over the other scenarios where this might be useful.

BTW... If we don't recommended, why mention them? 😅

content/docs/command-reference/commit.md
```diff
@@ -1,7 +1,7 @@
 # commit
 
-Record changes to DVC-tracked files in the <abbr>project</abbr>, by saving them
-to the <abbr>cache</abbr> and updating the `dvc.lock` or `.dvc` files.
+Stores changed changed files and directories tracked by DVC in the
```
shcheklein, line 3: changed changed

content/docs/command-reference/commit.md
```diff
@@ -1,7 +1,7 @@
...
+<abbr>cache</abbr>, and update the `dvc.lock` or `.dvc` files if needed.
```
shcheklein, line 4: store ... update
stores ... updates

content/docs/command-reference/commit.md
shcheklein: they are advanced, and users should understand the consequences

content/docs/command-reference/commit.md
shcheklein, line 3: This way we just put an algorithm here instead of providing some high level summary .. Previous was not ideal, but better to my mind

content/docs/command-reference/commit.md
```diff
@@ -17,65 +17,52 @@ positional arguments:
...
-the `dvc commit` command comes into play: It performs that last step when
-needed.
-
-Note that it's best to avoid the last three scenarios. They essentially
-force-update the `dvc.lock` or `.dvc` files and save data to cache. They are
-still useful, but keep in mind that DVC can't guarantee reproducibility in those
-cases.
+This command can be useful for several scenarios, when
+[stages](/doc/command-reference/run) are under development, or to force-update
+the cache without having to execute stage commands again (avoiding `dvc repro`).
```
shcheklein, line 22: it's not only about updating the cache

content/docs/command-reference/commit.md
```diff
@@ -17,65 +17,52 @@ positional arguments:
...
+
+> Note that this does not register project versions. That can be done with
```
shcheklein, line 24: Hmm ... I don't quite understand this

content/docs/command-reference/commit.md
shcheklein: it's a very large change ... what was a specific problem that we are solving? can we do a fix instead of rewriting the whole doc?

jorgeorpinel: There were 2 problems: (1) to focus on the main use case of `commit` (putting advanced uses lower in the description); and (2) fix the details about what commit does. I rewrote (2) since it had errors (the part about "Normally, `dvc repro` and `dvc run` finish up with the same steps as `dvc add`"), the remaining changes consist of moving paragraphs the bottom of the description (1).

content/docs/command-reference/commit.md
jorgeorpinel: > changed changed

Fixing

content/docs/command-reference/commit.md
jorgeorpinel: > store ... update

Will fix (may no longer be relevant when I address your next https://github.com/iterative/dvc.org/pull/1989#pullrequestreview-542464568).

content/docs/command-reference/commit.md
jorgeorpinel: Right. OK sort of rolled back to

> Record changes to files or directories tracked by DVC, by storing the current
versions in the \<abbr\>cache\</abbr\>, and updating `dvc.lock` or `.dvc` files as
needed.

WDYT?

content/docs/command-reference/commit.md
jorgeorpinel: True. I changed it to "or to force DVC to accept changed data in the workspace"

It's a bit abstract now but expanded on in the following paragraphs.

content/docs/command-reference/commit.md
jorgeorpinel: Again, it was a "DVC is not Git" note. Removed...

content/docs/command-reference/commit.md
shcheklein: just keep `Record changes to files or directories tracked by DVC`  ? (see git commit)

content/docs/command-reference/commit.md
shcheklein: still, it's not about stages only

content/docs/command-reference/commit.md
shcheklein: still here?

content/docs/command-reference/commit.md
```diff
@@ -17,65 +18,52 @@ positional arguments:
...
+
+Normally, `dvc repro` and `dvc run` finish up with the same steps as `dvc add`,
+for each <abbr>output</abbr> involved. In summary:
+
+- Compute the hash value of the file or directory and save it in the `dvc.lock`
+  or `.dvc` file.
+- If using Git, append the file/directory path to `.gitignore`.
+- Store the data in question in the <abbr>cache</abbr>.
+
+For stages under development, the last step can be skipped with the
```
shcheklein, line 36: this is not about stages obky

content/docs/command-reference/commit.md
jorgeorpinel: > still here?

My `git push` failed and I didn't realize until now...

content/docs/command-reference/commit.md
jorgeorpinel: Sure, I guess its best to keep the cmd ref intros as short as possible. I moved the other details down to the desc.

content/docs/command-reference/commit.md
jorgeorpinel: Hmmm also true. But do we want to emphasize `commit` as an alternative to `add`? I changed it a bit in 9b06323d, PTAL.

content/docs/command-reference/commit.md
jorgeorpinel: OK. I generalized this more.

content/docs/command-reference/commit.md
shcheklein: > But do we want to emphasize commit as an alternative to add

it's about imports as well, right? it can be useful for `add` since it might serve as a `dvc add *`.

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,54 @@ positional arguments:
...
-needed.
-
-Note that it's best to avoid the last three scenarios. They essentially
-force-update the `dvc.lock` or `.dvc` files and save data to cache. They are
-still useful, but keep in mind that DVC can't guarantee reproducibility in those
-cases.
+<abbr>Caches</abbr> the current contents of files and directories tracked by
+DVC, and updates `dvc.lock` or `.dvc` files as needed.
+
+This can be useful for several scenarios, when the project is under development,
```
shcheklein, line 22: projects are always under development :)

need to find a better phrase I think

content/docs/command-reference/run.md
```diff
@@ -244,14 +244,9 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-- `--no-commit` - do not save outputs to cache. A stage created, while nothing
-  is added to the cache. In the stage file, the file hash values will be empty;
-  They will be populated the next time this stage is actually executed, or
-  `dvc commit` can be used to force committing existing output file versions to
-  cache.
-
-  This is useful to avoid caching unnecessary data repeatedly when running
-  multiple experiments.
+- `--no-commit` - do not store outputs in the cache (`dvc.yaml` and `dvc.lock`
+  are still created or updated); useful to avoid caching unnecessary data when
```
shcheklein, line 248: `useful to avoid caching unnecessary data`

this sounds like `-O` options for an output

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,54 @@ positional arguments:
...
+or to force DVC to accept changed data (avoiding `dvc add` or `dvc repro`).
+We'll expand on these uses below.
+
+Normally, `dvc repro` and `dvc run` finish up with the same steps as `dvc add`,
```
shcheklein, line 26: also, import? import-url?

generalize this?

shcheklein: okay, actually they don not cache I guess (by default at least) ... I would try to generalize this somehow 

`Normally, commands that involve data tracking (dvc run, dvc repro, dvc add, etc) ...` ?

content/docs/command-reference/commit.md
jorgeorpinel: The use with `import --no-exec` is very edge, I'd say. And it's mentioned in that ref., which should be enough...?

But true about `add *` (`*` = everything already tracked by DVC). Updated per that.

content/docs/command-reference/commit.md
jorgeorpinel: How about "when experimenting during the development of a project" ? (Updated that way in  for now.)

shcheklein: GIt, DVC are always about development, so it is still confusing ... :(

jorgeorpinel: "when developing data or stages" ?

content/docs/command-reference/run.md
jorgeorpinel: True. But I guess its the -O/M options that need some clarification... Updated in , PTAL.

jorgeorpinel: * updated in 3711dc81

content/docs/command-reference/commit.md
jorgeorpinel: > actually [import commands] don not cache I guess (by default

They never cached, I think, even if the `store: true` option gets merged (which BTW would change the concept of "remote storage" a little as it could contain files that the cache can never contain... But I digress).

This also made me realize that the explanation of `import/get --no-exec` was wrong with regards to caching — prob. because it was copied from `run`. Updated that in ce8ccad2

jorgeorpinel: > would try to generalize this... commands that involve data tracking...

But it's only run and repro, I think. And `add` only does the last part of what run/repro do: save the file hash and cache the file. So if it's just those 3, why say "etc." ?

jorgeorpinel: > But it's only run and repro

p.s. meaning, the ones that have a `--no-commit` option to skip that last "adding" part.

shcheklein: `dvc commit <out>.dvc can be used if the data hashes are needed in the .dvc file, and to save existing data to the cache.` - this comes from the `dvc import`

cache != remote, import/import-url cache artifacts as far as I remember, after all

> But it's only run and repro, I think. And add only does the last part of what run/repro do: save the file hash and cache the file. So if it's just those 3, why say "etc." ?

hmm, what is the first part?

content/docs/command-reference/commit.md
shcheklein: still not good ... try to rewrite the phrase completely, use simpler words, be more explicit?

shcheklein: or skip it?

content/docs/command-reference/run.md
shcheklein: okay, looks good

content/docs/command-reference/commit.md
jorgeorpinel: > import/import-url cache artifacts as far as I remember

Yeah I may be confused. Let em double check on that... ⌛ 

> what is the first part?

"the last part" refers to the part of repro and run that has to do with the pipeline or stage structure and command (write or read dvc.yaml, determine what to execute, run commands, etc.). But I guess the whole "Normally, dvc repro and dvc run finish up..." paragraph and bullet list bellow can be simplified to avoid explaining too much about those commands here... ⌛

jorgeorpinel: > import/import-url cache artifacts... Let me double check

Yeah I was confused 🤦 — I reverted those changes (and noticed other problems in those refs but will leave that for later.)

content/docs/command-reference/commit.md
jorgeorpinel: Hmmm OK, I rephrased the first desc. paragrph so this isn't mentioned there. But note that it's still mentioned a couple paragraphs below as "for example when testing/experimenting with data or stages" — hopefully that one is OK since the reader has much more context at that point.

content/docs/command-reference/commit.md
jorgeorpinel: > the whole "Normally, dvc repro and dvc run finish up..." paragraph and bullet list bellow can be simplified

OK I generalized and simplified this without completely removing the bullet list (which I think makes the reading easier/more interesting).

jorgeorpinel:
PTAL @shcheklein

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,56 @@ positional arguments:
...
-commands where it's available). The file hash is still computed and added to the
-`dvc.lock` or `.dvc` file, but the actual data is not cached. And this is where
-the `dvc commit` command comes into play: It performs that last step when
+Stores the current contents of files and directories tracked by DVC in the
+<abbr>cache</abbr>, and updates `dvc.lock` or `.dvc` files as needed. This
+forces DVC to accept the contents of tracked data currently in the
+<abbr>workspace</abbr>, even if they have changed. We explore the scenarios in
+which this can be useful next.
+
+DVC commands that track data (`dvc add`, `dvc repro`, `dvc run`) do the
```
shcheklein, line 25: , etc ? (import, import-url)?

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,56 @@ positional arguments:
...
+following for each file or directory in question:
+
+- Save the hash value of the file(s) in the `dvc.lock` or `.dvc` file.
+- Store the file contents in the cache.
+
+The last step can be skipped with the `--no-commit` option of those commands,
+for example when testing/experimenting with data or
+[stages](/doc/command-reference/run). This avoids caching unfinished data. And
+that's where `dvc commit` comes into play: It performs that last step when
 needed.
```
shcheklein, line 35: I like it!

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,56 @@ positional arguments:
...
 
-Note that it's best to avoid the last three scenarios. They essentially
-force-update the `dvc.lock` or `.dvc` files and save data to cache. They are
-still useful, but keep in mind that DVC can't guarantee reproducibility in those
-cases.
+💡 For convenience, a pre-commit Git hook is available to remind you to
+`dvc commit` when needed. See `dvc install` for more info.
+
+Other scenarios include:
```
shcheklein, line 40: Not sure `Other` applies here? it's more like a detailed explanation?

content/docs/command-reference/run.md
```diff
@@ -178,15 +178,16 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
   `dvc add`).
 
 - `-O <path>`, `--outs-no-cache <path>` - the same as `-o` except that outputs
-  are not tracked by DVC. It means that they are not cached, and it's up to a
-  user to manage them separately. This is useful if the outputs are small enough
-  to be tracked by Git directly, or if these files are not of future interest.
+  are not tracked by DVC. This means that they are never cached, so it's up to
+  the user to manage them separately. This is useful if the outputs are small
+  enough to be tracked by Git directly, or if these files are not of future
```
shcheklein, line 183: or too large and can be reproduced if neeed ...

content/docs/command-reference/run.md
```diff
@@ -244,14 +244,10 @@ $ dvc run -n my_stage './my_script.sh $MYENVVAR'
...
-  is added to the cache. In the stage file, the file hash values will be empty;
-  They will be populated the next time this stage is actually executed, or
-  `dvc commit` can be used to force committing existing output file versions to
-  cache.
-
-  This is useful to avoid caching unnecessary data repeatedly when running
-  multiple experiments.
+- `--no-commit` - do not store the outputs of this execution in the cache
+  (`dvc.yaml` and `dvc.lock` are still created or updated); useful to avoid
+  caching unnecessary data when executing tests or experiments. Use `dvc commit`
```
shcheklein, line 249: tests and experiments ...

both are overloaded ... 

may be something like `when iterations on code and data happen fast`? 

also in this specific case - it's about creating the pipeline first ... (then run it as a whole)

shcheklein (changes requested):
Overall it's great, just a few minor comments to iterate on

content/docs/command-reference/commit.md
jorgeorpinel: No, it's only those because "The last step can be skipped with the `--no-commit` option of those commands" (in the next paragraph).

content/docs/command-reference/commit.md
jorgeorpinel: Hmmm they're meant as different scenarios. The part before this is about `add/run/repro --no-commit` + `dvc commit`. The bullets in this list are
- `add *`
- force-accepting cosmetic changes to dependencies
- adding missing deps/outs
- executing commands manually (this one I guess is pretty similar to the main case)

content/docs/command-reference/commit.md
shcheklein: so, the good question how does it relate to `--no-exec` then? (since we also mention `dvc commit` in those, right?)

content/docs/command-reference/commit.md
shcheklein: some of them (all of them?) are part of the _main_ use case in this terminology.

`add/run/repro --no-commit` is not even a use case by itself, right? it doesn't explain a specific need.

content/docs/command-reference/run.md
jorgeorpinel: OK, updated this bit in 97be1b91.

content/docs/command-reference/commit.md
jorgeorpinel: Hmmm, OK true. I thought we were trying to avoid that complication (mentioning `--no-exec`) but I did put it in pretty easily in 144749f5.

content/docs/command-reference/commit.md
jorgeorpinel: I thought the main reason `commit `exists as a stand-alone command is to complement the `--no-commit` options of add/run/repro?

In terms of the "story", its explained as "forces DVC to accept the contents of tracked data currently in the workspace" in the first p of the description. So do you mean they're all different flavors of that explanation and that there should be a single bullet list (including `add/run/repro --no-commit`)?

shcheklein: > I thought the main reason commit exists as a stand-alone command is to complement the --no-commit options of add/run/repro?

probably initially it was the case, now we reference it in other places? (like `--no-exec`?) So we might need to revisit,  generalize? (not 100% sure, just asking)

> So do you mean they're all different flavors of that explanation 

it seems so (not sure about _all_)

> that there should be a single bullet list 

ah, not necessarily. Just removing Other might help? or rephrasing it a bit if the first paragraph is already general.

again, not very constructive feedback here - just highlighting stuff as I read it, a think that seemed strange

content/docs/command-reference/run.md
jorgeorpinel: Yeah in the `commit` ref. this is now explained as "when exploring different data or stages". Let me think 🧠

jorgeorpinel: > iterations on code and data happen fast

But why do people iterate fast? It's during an exploration/testing/experimenting phase of the development process. I guess "exploring" is the only non-overloaded term there. Other phrases: spatial search, brainstorm, play around

jorgeorpinel: > it's about creating the pipeline first

That applies more to --no-exec I think, let me check it's mentioned clearly... ⌛

jorgeorpinel: > applies more to --no-exec I think, let me check it's mentioned clearly

Wait no, we already discussed this in a recent PR over the import refs. and decided to remove the motivation paragraph because the option description was too long, among other reasons...

jorgeorpinel: It is, however, mentioned in `run` already, as "This is useful if, for example, you need to build a pipeline quickly first, and run it all at once later (with `dvc repro`)." so... I guess we should add it to imports? Added for now! (in

content/docs/command-reference/import-url.md
```diff
@@ -125,11 +125,15 @@ source.
...
-  directory already exists locally, for example, along with `dvc commit` to
-  store it in the cache and record its hash value in the `.dvc` file.
+- `--no-exec` - create the import `.dvc` file without actually downloading
+  `url`. The data hash is not calculated when this option is used, only the
+  import metadata is saved to the `.dvc` file. It can be useful to skip the
+  download if the file/directory already exists locally, for example, along with
+  `dvc commit` to store it in the cache and record its hash value in the `.dvc`
+  file.
+
+  This is useful if, for example, you need to build a project that will use
```
shcheklein, line 135: you have repetition useful ... useful

content/docs/command-reference/commit.md
```diff
@@ -17,65 +16,56 @@ positional arguments:
...
+these step(s) of the tracking process:
+
+- Save the hash value of the file(s) in the `dvc.lock` or `.dvc` file.
+- Store the file(s) contents in the cache.
+
+> Skipping these steps is typically done to avoid caching unfinished data, for
+> example when exploring different data or
+> [stages](/doc/command-reference/run)).
+
+Other scenarios for `dvc commit` include:
```
shcheklein, line 40: it still reads that they are different from the Note above ... which is not exactly true..

also, do we need it to be a Note? why is it a Note?

content/docs/command-reference/import-url.md
jorgeorpinel: Hm yeah this sentence wasn't great, sorry... Updated in d3ed5536.

content/docs/command-reference/commit.md
jorgeorpinel: > still reads that they are different from the Note above

I don't see this so I'm not sure how to fix that but happy to make any specific changes you have in mind.

> do we need it to be a Note? why is it a Note?

Maybe we don't need it (no strong opinion, please lmk) but it explains why people would skip steps during add/repro/import, etc., since that's what we're explaining as the main use for `commit`. It's a note because it's about those other commands, not about `commit`.

content/docs/command-reference/import-url.md
shcheklein: it's repetitive with the last sentence of the previous paragraph, essentially it reads:

```
something. It is useful when A.

This is useful when B.
```

is it intentional?

and overall it got complicated again. Is there a way to merge A and B?

content/docs/command-reference/commit.md
shcheklein: > but happy to make any specific changes you have in mind.

Other -> Specifically, scenarios for ... 

It looks like that is the meaning (at least from the content)

content/docs/command-reference/import-url.md
jorgeorpinel: True. I guess this is why we didn't have so much motivation in the --no-exec options before... I tried to merge it now but it's still a little long.

Maybe we should remove the part about "if the file/directory already exists locally"? It seems like a very edge case to download something manually and THEN register it with `import`.

content/docs/command-reference/commit.md
jorgeorpinel: Changed!

content/docs/command-reference/import-url.md
jorgeorpinel: > Maybe we should remove the part about "if the file/directory already exists locally"

Yeah, I removed that to simplify.

jorgeorpinel:
I suggest we don't obsess over the `--no-exec` text for now and instead open/reactivate core tickets to make that the default behavior, at least for run (which would need to be renamed probably) — since it's supposed to be a helper for writing dvc.yaml files.

It's also an interesting question what to do with the `import` commands. I think that .dvc files could be phased and all tracked data could be included in a section of dvc.yaml (before `stages`, see https://github.com/iterative/dvc/issues/3936#issuecomment-692470880) so in that case `--no-exec` should also be the default behavior.

shcheklein (approved):

PR #1994, opened 2020-12-02
add more helps to s3-compatible remote

johnnychen94:

### DIFF
--- a/content/docs/command-reference/remote/add.md
+++ b/content/docs/command-reference/remote/add.md
@@ -143,19 +143,7 @@ $ dvc remote modify myremote endpointurl \
 > See `dvc remote modify`(the "Amazon S3" section) for a full list of S3 API
 > parameters.
 
-To save and reuse key and secret for all your dvc project, you could add them to
-the global dvc config file:
-
-```dvc
-$ dvc remote modify --global myremote access_key_id <my-access-key>
-$ dvc remote modify --global myremote secret_access_key <my-secret-key>
-```
-
-You should use this with caution and make sure others don't have access to your
-credential information. `--global` is required here because you should never
-commit your credential information to git history.
-
-If temporary usage is need, you can also configure it via environment variables:
+S3 remotes can also be configured entirely via environment variables:
 
 ```dvc
 $ export AWS_ACCESS_KEY_ID="<my-access-key>"
@@ -495,3 +483,21 @@ $ dvc remote list
 
 myremote	s3://mybucket/another-path
 ```
+
+The credential information can be set via environment variables, for example:
+
+```dvc
+$ export AWS_ACCESS_KEY_ID="<my-access-key>"
+$ export AWS_SECRET_ACCESS_KEY="<my-secret-key>"
+```
+
+or via the dvc (global) config:
+
+```dvc
+$ dvc remote modify --global myremote access_key_id <my-access-key>
+$ dvc remote modify --global myremote secret_access_key <my-secret-key>
+```
+
+You should use this with cautiqon and make sure others don't have access to your
+credential information. For security reason, `--global` is required here because
+you should never commit your credential information to git history.
--- a/content/docs/command-reference/remote/add.md
+++ b/content/docs/command-reference/remote/add.md
@@ -140,9 +140,22 @@ $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-> See `dvc remote modify` for a full list of S3 API parameters.
+> See `dvc remote modify`(the "Amazon S3" section) for a full list of S3 API
+> parameters.
 
-S3 remotes can also be configured entirely via environment variables:
+To save and reuse key and secret for all your dvc project, you could add them to
+the global dvc config file:
+
+```dvc
+$ dvc remote modify --global myremote access_key_id <my-access-key>
+$ dvc remote modify --global myremote secret_access_key <my-secret-key>
+```
+
+You should use this with caution and make sure others don't have access to your
+credential information. `--global` is required here because you should never
+commit your credential information to git history.
+
+If temporary usage is need, you can also configure it via environment variables:
 
 ```dvc
 $ export AWS_ACCESS_KEY_ID="<my-access-key>"
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -23,7 +23,7 @@ positional arguments:
 
 Remote `name` and `option` name are required. Config option names are specific
 to the remote type. See `dvc remote add` and
-[Available settings](#available-settings-per-storage-type) below for a list of
+[Available settings](#available-parameters-per-storage-type) below for a list of
 remote storage types.
 
 This command modifies a `remote` section in the project's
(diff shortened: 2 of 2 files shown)


content/docs/command-reference/remote/add.md
```diff
@@ -140,9 +140,19 @@ $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-> See `dvc remote modify` for a full list of S3 API parameters.
```
johnnychen94, line 143: This doesn't give useful information, so I've removed this.

```console
ERROR: the following arguments are required: name, option
usage: dvc remote modify [-h] [--global | --system | --local] [-q | -v] [-u]
                         name option [value]

Modify the configuration of a data remote.
Documentation: <https://man.dvc.org/remote/modify>

positional arguments:
  name           Name of the remote
  option         Name of the option to modify.
  value          (optional) Value of the option.

optional arguments:
  -h, --help     show this help message and exit
  --global       Use global config.
  --system       Use system config.
  --local        Use local config.
  -q, --quiet    Be quiet.
  -v, --verbose  Be verbose.
  -u, --unset    Unset option.
```

content/docs/command-reference/remote/add.md
```diff
@@ -140,9 +140,19 @@ $ dvc remote modify myremote endpointurl \
...
+To save and reuse key and secret for all your dvc project, you could add them to
+the global dvc config file:
 
-S3 remotes can also be configured entirely via environment variables:
+```dvc
+$ dvc remote modify --global myremote access_key_id <my-access-key>
+$ dvc remote modify --global myremote secret_access_key <my-secret-key>
```
johnnychen94, line 148: I'm a bit lost where the original source of these two keywords, but they are what work instead of `aws_access_key_id`/`aws_secret_access_key`

content/docs/command-reference/remote/add.md
shcheklein: `dvc remote modify` is a link to the command in this case (it's a regular way to link commands in docs)

content/docs/command-reference/remote/add.md
shcheklein: this is general thing for any S3 ... the recommended way to achieve what you are trying to do (the way I understand this) is to setup AWS cli util or manually create `~/.aws/credentials`. `boto` (a lib we use internally will pick it up) automatically. Rarely you wound need to override keys/secrets via DVC config this way.

content/docs/command-reference/remote/add.md
johnnychen94: The point is that `dvc remote modify` does not provide anything related to S3 API parameters.

content/docs/command-reference/remote/add.md
shcheklein: > I'm a bit lost where the original source of these two keywords, but they are what work instead of

they come from the `remote modify` docs - https://dvc.org/doc/command-reference/remote/modify (S3 section)

shcheklein: > but they are what work instead of `aws_access_key_id`/`aws_secret_access_key`

@efiop it's mighta  good point to actually use prefixes like `aws_` - we can name them the same way as they are named for a specific provider then. To be specific - if provider has prefixes we should probably use the same name. It's indeed convenient to have aws_access_key_id, AWS_ACCESS_KEY_ID, and the same name in DVC config.

content/docs/command-reference/remote/add.md
shcheklein: It does. 

Screen Shot 2020-12-01 at 9 35 13 PM

johnnychen94: Oh, I misunderstood the intention here; didn't realize that it was used as a URL placeholder.

shcheklein:
@johnnychen94 I see the problem you are trying to solve, but let's think a bit about alternative approaches. The problem is very general and `--global` could be used in a lot of cases (not only s3-compatible, and not only even s3)

should it be made clear in the the `dvc remote modify` docs that `--global` could be used in a way to override/set certain values?

johnnychen94:
A good suggestion, I've moved it to the example part https://dvc.org/doc/command-reference/remote/add#example-customize-an-s3-remote. Might be something obvious for oss users but at least it's worth giving a warning.

content/docs/command-reference/remote/add.md
```diff
@@ -140,7 +140,8 @@ $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-> See `dvc remote modify` for a full list of S3 API parameters.
+> See `dvc remote modify`(the "Amazon S3" section) for a full list of S3 API
+> parameters.
```
jorgeorpinel, line 144: We can leave this as it was. https://dvc.org/doc/command-reference/remote/modify#available-parameters-per-storage-type has a section called "Click for S3 API compatible storage", matching the one here. The user can decide which one to check (or both).

content/docs/command-reference/remote/add.md
```diff
@@ -482,3 +483,21 @@ $ dvc remote list
 
 myremote	s3://mybucket/another-path
 ```
+
+The credential information can be set via environment variables, for example:
+
+```dvc
+$ export AWS_ACCESS_KEY_ID="<my-access-key>"
```
jorgeorpinel, line 490: This example is about customizing the remote (in .dvc/config) so this new text would maybe need to be it's own separate example?

jorgeorpinel: But the info about setting up credentials is already in the expandable sections. Not sure we need an example here. Is this trying to solve a specific scenario?

content/docs/command-reference/remote/add.md
```diff
@@ -482,3 +483,21 @@ $ dvc remote list
...
+or via the dvc (global) config:
+
+```dvc
+$ dvc remote modify --global myremote access_key_id <my-access-key>
+$ dvc remote modify --global myremote secret_access_key <my-secret-key>
+```
+
+You should use this with cautiqon and make sure others don't have access to your
+credential information. For security reason, `--global` is required here because
+you should never commit your credential information to git history.
```
jorgeorpinel, line 503: Why --global and not --local? I'm still not getting what the scenario we're targeting here is. If someone can explain that to me, then I can prob. give a better review. Thanks

content/docs/command-reference/remote/modify.md
```diff
@@ -277,6 +277,9 @@ $ dvc remote modify myremote endpointurl \
                              https://nyc3.digitaloceanspaces.com
 ```
 
+Please refer to the "Amazon S3" section for a list of complete S3 variables;
+whether the variable is useful depends on your actual S3 backend.
+
```
jorgeorpinel, line 282: Good note! But this is already mentioned in the previous paragraph: "Besides that, any settings that are available for Amazon S3 (see previous section) may be available for S3 compatible storage."
```suggestion
```

content/docs/command-reference/remote/add.md
johnnychen94: IIUC with `--local` the credential information will be written into `<repo>/.dvc/config`, which will be tracked by git. So if there's anyone who doesn't take this carefully, the password might get leaked.

Or perhaps it should leave `dvc` to throw a warning when the user tries to do this?

jorgeorpinel: No, with --local another config file will be used, which is put in .gitignore. See https://dvc.org/doc/command-reference/config#command-options-flags. Thanks

johnnychen94: Oh right, I always thought the default behavior is equivalent to `--local`. Thanks for pointing this out.

shcheklein:
Closing this as stale. @johnnychen94 feel free to reopen it when you have a bit more time to address the comments.

PR #2016, opened 2020-12-09
Add simple Typeform component and /enterprise page using it

rogermparent:
A quick solution to adding a Typeform to the url `/enterprise`, much like [the viewer beta form](https://viewer.iterative.ai/beta-form).

Currently, it blends into the site UI and specified a min-height that ensures a nested vertical scroll doesn't happen.

If we want any changes to the Enterprise page specifically, they can simply be added to the page component.

I haven't added a link to any menus yet, but that can also be done in either this or another PR.


shcheklein:
@rogermparent looks awesome, could we disable the banner on that page? The idea that people will go from the banner link to it.

Also, can we update the banner itself as part of the PR - `:shield: DVC for Enterprise - data access control & compliance!`

shcheklein:
Thank @rogermparent !

PR #2017, opened 2020-12-10
Update checks-action to 1.1.1

rogermparent:
the [checks action](https://github.com/LouisBrunner/checks-action/blob/master/README.md) we use has updated again recently, and possibly with an issue that solves the `deployment_status` issue which has been the main hurdle of Link Check since GitHub Actions were introduced to it.


rogermparent:
I opened this PR to test, and it seems a simple version change doesn't fix the issue. The new `sha` field may be an avenue we can take for fixing the Checks issue, however.
For now, I'll close this PR so as not to leave this dangling while I focus on other things. I'll either reopen this or make a new issue in the future.

PR #2025, opened 2020-12-15
heartbeat draft

elleobrien:
i'm looking for a good cover image, so that will be replaced before we publish


shcheklein:
https://www.zazzle.com/t_shirt-235920696568133954  - this is fun @elleobrien  :)) 

(but let's add it to the links exclusion list please, otherwise CI fails  ^^ - it doesn't all us to check it - throws HTTP 403)

shcheklein:
how about btw take that image as a header image :)

content/blog/2020-12-14-december-20-dvc-heartbeat.md
```diff
@@ -0,0 +1,171 @@
+---
+title: December '20 Heartbeat
+date: 2020-12-14
+description: |
+  Monthly updates are here- read all about our brand new video docs, the DVC Udemy course, open jobs with our team, and essential reading about Git-flow with DVC.
```
shcheklein, line 5: 💅 please keep it under 80

content/blog/2020-12-14-december-20-dvc-heartbeat.md
```diff
@@ -0,0 +1,171 @@
...
+to full-stack machine learning. The event is free and virtual, so consider
+stopping by!
+
+![](/uploads/images/2020-12-14/pydata_mtl.jpg)
+
+## From the community
+
+There's no shortage of cool things to report from the community:
+
+### The DVC Udemy Course
```
shcheklein, line 101: Mikhail had a special singup link for the singup - let's ask him and put it. It's important for him as I know.

shcheklein (approved):
a few things to care of, and the image if you decide to change it (should we start doing them on upwork btw? to save us time)

elleobrien:
@shcheklein thanks for comments! as for using upwork... right now, the header images take <10 minute because i made a template a while ago (on Adobe Spark). So less time than hiring someone. But if we want to go any more complicated designs in the future, contractor is the way to go.

content/blog/2020-12-14-december-20-dvc-heartbeat.md
elleobrien: yes, good idea- asked on slack

shcheklein:
> So less time than hiring someone

yep, I was thinking about someone who can do on a permanent basis - similar to the vide editing? (so that we don't create a new job just for one image every time)

elleobrien:
@shcheklein may be worth it in the long run. will start looking around for someone when we release next Community Gems (could have Gems cover and HB covers).

elleobrien:
@rogermparent the link checker seems to be breaking on twitter URLs, which are usually no problem. can you take a look?

rogermparent:
> @rogermparent the link checker seems to be breaking on twitter URLs, which are usually no problem. can you take a look?

`curl` fails too, returning this in the body: 
```
<h1>This browser is no longer supported.</h1>
<p>
  Please switch to a supported browser to continue using twitter.com. 
  You can see a list of supported browsers in our Help Center.
</p>
```
 it looks like Twitter returns 400 with the basic agent-less requests we send. My assumption is that they check against a whitelist of user agents, and a true fix involves confirming this and adding the capability to mock user agents to the link check script- possibly on a per-site basis, but I have a feeling that every site would support something like the latest google chrome user agent.

rogermparent:
Confirmed that `curl` the user agent `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.89 Safari/537.36` gets a 200 from Twitter.

shcheklein:
@rogermparent the question here is why does it fail even after it was added to the exclusion list, I guess

rogermparent:
> @rogermparent the question here is why does it fail even after it was added to the exclusion list, I guess

I must have forgot that Twitter was on the exclusion list, I'll have to check that out.

shcheklein:
But I thought this was the link that failed https://www.zazzle.com/t_shirt-235920696568133954, no?

rogermparent:
> But I thought this was the link that failed https://www.zazzle.com/t_shirt-235920696568133954, no?

No, it's all Twitter links that are failing on the report. I think Twitter may have recently just started filtering user agents.

shcheklein:
Yep, I see .. yesterday it was all about that one ... https://github.com/iterative/dvc.org/runs/1554043079

shcheklein:
We can update the user agent, I guess.

rogermparent:
> Yep, I see .. yesterday it was all about that one ... https://github.com/iterative/dvc.org/runs/1554043079

`curl`ing this gives the following in the body:

> <p> Access to this page has been denied because we believe you are using automation tools to browse the website. </p> 
> <p> This may happen as a result of the following: </p> <ul> <li> Javascript is disabled or blocked by an extension (ad blockers for example) </li> <li> Your browser does not support cookies </li> </ul> <p> Please make sure that Javascript and cookies are enabled on your browser and that you are not blocking them from loading. </p>

And this happens even with the user agent that fixes twitter, so it's a good example of sites that are more strict about non-browser requests.

elleobrien:
I'm wondering if it just failed on the zazzle link first and that's why we
didn't see twitter errors yesterday

rogermparent:
> I'm wondering if it just failed on the zazzle link first and that's why we didn't see twitter errors yesterday

While anything is possible, I don't think this is the case. I'm pretty sure the link check script was written specifically not to stop early other than Exceptions that would prevent any check output at all.

It sounds a little silly even to me to blame the appearance of this issue on Twitter changing their behavior on user agents overnight, but I can't think of any other scenario where it wouldn't have been an issue already since we do a *lot* of link checks and additions of Twitter links. That said, this anti-scraper behavior is very commonplace and just something this script needs to adapt to on a per-site basis, because they all behave a little differently.

PR #2044, opened 2020-12-22
Misc. updates

jorgeorpinel:
This is some vanity stuff, sorry. No priority at all but it also shouldn't be very hard to check? It's well scoped too 😅 

I just need to clear out my stash periodically. There will be a constant stream of these (some more relevant, some less) so again, no pressure of course.


shcheklein (approved):

jorgeorpinel:
BTW this applies to v1 but for simplicity, I won't backport.

Issue #1663, opened 2020-08-06
term: simplify multiple mentions about dvc.yaml + dvc.lock + .dvc files


jorgeorpinel:
UPDATE: One thing left here, see https://github.com/iterative/dvc.org/issues/1663#issuecomment-749333023 below.

> Continues #1431

> Extracted from https://github.com/iterative/dvc.org/pull/1384#pullrequestreview-461182909 et al.

To achieve this, we will probably need to introduce a new glossary term such as "Special DVC files".

shcheklein:
Not sure about _Special_ though. They are not very special I guess.

jorgeorpinel:
"DVC metafiles"? "DVC-specific files"? Can't use just "DVC files" because it looks almost like ".dvc files".

shcheklein:
DVC files with <abbr> if needed sounds good to me. I don't how to generalize it better. `.dvc`, `dvc.lock`, `dvc.yaml` - they all DVC files, right?

jorgeorpinel:
People have always called `.dvc` files "DVC files", which makes sense and it's one of the reasons we wanted to stop using the term "stage file" (even before 1.0) because we really had to force it and still was not adopted. Throwing in dvc.yaml (and dvc.lock) sounds like making a similar mistake, where we'll have to force people to get that DVC files are not just .dvc files but also other files.

Also, technically theres plenty other "DVC files" such as `.dvc/config` or plots templates, but we wouldn't include those in the term.

Maybe "Stage files" wasn't such a bad term after all...

jorgeorpinel:
OK, so I vote for "DVC metafiles" (would mostly be used in plural). WDYT @shcheklein? Should we call in a vote from everyone via Slack or mentioning them here perhaps?

jorgeorpinel:
Actually... a term to group the files may not solve our problems. We also have to repeat the phrase "stages and .dvc files" a lot, and that could not be replaced by a term that refers to files.

We may need a broader concept, kind of like how we used to consider .dvc files "(orphan) stages" before, except I would use that term anymore since it would be overloaded... So maybe "DVC nodes", "DVC blocks"? Idk.

This could be in addition to "DVC (meta)files"

- **DVC node**: it can have deps and outs (a possible link in a DAG chain). Includes stages (found in dvc.yaml) and the data structure in .dvc files (the latter has no command thus doesn't represent a process, thus is not a stage).
[...]

PR #2050, opened 2020-12-22
term: remove "DVC-file" (except from old blogs)

### LINKED ISSUE #1366: term: stop using "DVC-file" and "stage file"

jorgeorpinel:
> Continues #2047 and #2048.
> - [x] Change base here to `master` once those are merged.

Nice-to-have before 2.0 dvc.yaml changes

- Reviews usage of all remaining instances of "DVC-file"
- Some related/nearby copy edits

Closes #1366


jorgeorpinel:
Ready for review

content/docs/command-reference/checkout.md
```diff
@@ -18,9 +18,9 @@ positional arguments:
 ## Description
 
 This command is usually needed after `git checkout`, `git clone`, or any other
-operation that changes the current `dvc.lock` or `.dvc` files. It restores the
-corresponding versions of the DVC-tracked files and directories from the
-<abbr>cache</abbr> to the workspace.
+operation that changes the current [DVC files](/doc/user-guide/dvc-files). It
+restores the corresponding versions of the DVC-tracked files and directories
+from the <abbr>cache</abbr> to the workspace.
```
jorgeorpinel, line 23: This is the only generalization done in this PR. Reviewing all the instances of these repetitive mentions will require a separate PR and is related to https://github.com/iterative/dvc.org/issues/1663#issuecomment-749333023

shcheklein (approved):

PR #2052, opened 2020-12-23
2.0 Pipelines (parametrization)

jorgeorpinel:
> Based on https://github.com/iterative/dvc/wiki/Parametrization

- DVC Files guide - `dvc.yaml` section
  + [x] General vars info
  + [x] Parameters files
  + [x] Global/Local vars 
  + [x] Multi-stages (foreach "loops")
  + [x] Specific/multiple/partial params files in `vars`
  + [x] Other notes e.g. that "parametrized variables are also displayed by `params diff`" and about `wdir`.
- [x] `params` cmd ref.
- [x] term "parametrize" -> "parametErize" + intro "templat(e|ing)"

- Follow this up with #2059 immediately after.


content/docs/command-reference/params/index.md
```diff
@@ -144,6 +147,48 @@ $ dvc run -n train -d logs/ -o users.csv \
           python train.py
 ```
 
+## Examples: Print all parameters
+
+Following the previous example, we can use `dvc params diff` to list all of the
+param values available in the <abbr>workspace</abbr>:
```
jorgeorpinel, line 153: This example has no changes but I moved the one about Python param files to the bottom so it looks like this is added.

content/docs/command-reference/params/index.md
```diff
@@ -144,6 +147,48 @@ $ dvc run -n train -d logs/ -o users.csv \
...
+
+This command shows the difference in parameters between the workspace and the
+last committed version of the `params.yaml` file. In our example there's no
+previous version, which is why all `Old` values are `—`.
+
+## Example: Parametrized stages in `dvc.yaml`
+
+Let's say we have a `params.yaml` file with the following contents:
+
+```yaml
```
jorgeorpinel, line 173: This is the example I added here.

content/docs/command-reference/params/index.md
```diff
@@ -28,6 +28,9 @@ YAML 1.2, JSON, TOML, or [Python](#examples-python-parameters-file) _parameters
 file_. Multiple parameter dependencies can be specified from one or more
 parameters files.
 
+💡 Parameters can also be used as variables in `dvc.yaml` itself. See the
+[Parametrizes stages](#example-parametrized-stages-in-dvcyaml) example below.
```
jorgeorpinel, line 32: I find this feature pretty confusing. Params were originally meant as values for your code, which you can track granularly with DVC. Now we're also enabling their use as variables inside dvc.yaml itself, where they can also be define as params for the code. You can even do both things at once. Meta and kind of confusing.

In fact I'm avoiding calling this "parametrization" (of stages/dvc.yaml) in docs to lessen the confusion.

This has probably been discussed internally so for now I'm just commenting here cc @shcheklein @skshetry thoughts?

For now I'll try to write the docs encouraging the use of `vars` sections mainly, and just mentioning this as an alternative.

content/docs/user-guide/dvc-files.md
```diff
@@ -129,7 +129,14 @@ can be written manually or generated by user code.
 Here's a comprehensive `dvc.yaml` example:
 
 ```yaml
-stages:
+vars: # Optional section
+  - train:
+      script: train.py
```
jorgeorpinel, line 134: - [ ] This intro example to dvc.yaml is probably getting too complicated...

jorgeorpinel: Should we work on a dvc.yaml writing guide along with or in parallel to this @shcheklein ? That way I can leave only the basic essentials here for now and leave all the possibilities for that separate guide or how-to.

content/docs/user-guide/dvc-files.md
```diff
@@ -251,3 +266,59 @@ directory tracked by DVC. Specifically: `md5`, `etag`, or `checksum` (same as in
 
 Full <abbr>parameters</abbr> (key and value) are listed separately under
 `params`, grouped by parameters file.
+
+### Variables in `dvc.yaml`
+
+To avoid repetition, `dvc.yaml` supports variables with the following syntax.
```
jorgeorpinel, line 272: I'm also thinking we will need to split dvc.yaml from .dvc files in the guide (as a 3rd level in the main nav) so these H3s become H2s and are shown in the right-hand nav... Do you agree @shcheklein ? And if so should I do so in this same PR? Thanks

jorgeorpinel: p.s. my plan is to also add a "Stage groups" section (for loops) soon, so the doc will get pretty long... Another good reason to split from .dvc files.

p.p.s we'll just need to repeat the Outputs and Dependencies sections in both pages (.dvc file outs/deps are a little different from those in dvc.yaml anyway so that would be more correct even if there's overlap).

content/docs/command-reference/params/index.md
jorgeorpinel: p.s. an obvious proposal is to move this functionality to a file named `vars.yaml` (by default, configurable) to clearly separate code params from dvc.yaml vars.

content/docs/user-guide/dvc-files.md
skshetry: @jorgeorpinel, I think it'd be better to not show parametrization right away, and put the note at the end. 

> Should we work on a dvc.yaml writing guide along with?

That'd be great, but even if not, maybe we could separate the dvc.yaml format explanation?
Like `dvc.yaml` short reference, then a section about "stage definition" explaining what key/value it takes, then "variables/parametriation" and then introduce loops. This way, concepts will get built as we go.

content/docs/command-reference/params/index.md
skshetry: > You can even do both things at once (with the same param)

This is one of the problems that it tries to solve. Is it a bad thing?

> In fact I'm avoiding calling this "parametrization" (of stages/dvc.yaml) in docs to lessen the confusion.

I like to call the values "variables" (though they are static 😕), but the stage definition does have parameters to be passed (like in functions), so parametrization?

Another term could be "templating".

shcheklein: Not sure about the naming - vars, etc.

>You can even do both things at once (with the same param)

I also don't see a problem with this. Consider your pipeline definition part of the project/your code. That's it, you parametrize all things this way. `Parameters can also be used as variables` - this phrase sounds a bit confusing indeed though. I would need to read more to recommend something better.

content/docs/user-guide/dvc-files.md
shcheklein: it's already part of the user guide, right? may be just split it into subsections - one about regular simple pointers .dvc, second about dvc.yaml and start writing a proper guide in it, start building proper structure for it. We can take a look at similar docs (spacy, ansible, etc, @skshetry have you seen some examples that you like?)

content/docs/command-reference/params/index.md
jorgeorpinel: > This is one of the problems that it tries to solve.
> Consider your pipeline definition part of the project/your code.

OK so you have `var: 'value'` in params.yaml and use that value in your code.py, but you also use it as part of the definition of the stage that execs code.py? Resulting in something like

```yaml
stages:
  my-thing:
    cmd: python code.py # reads params.yaml, uses var
    params:
      - var # dvc status/repro will check this from params.yaml 
    outs:
      - ${var}.txt
```

I'm just not seeing the logic for this use case. But maybe an example would help me understand and explain this @skshetry 🙂

jorgeorpinel: In any casem that's not my main concern. That would be that DVC reads params.yaml twice, for separate reasons:

1. to substitute it's values in the dvc.yaml definition itself; and
2. (status, repro) to determine if the stage is outdated

But it never passes the values to code.py BTW, which has to read it a third time.

It seems to me like params files are being overloaded with features which may not be easy to grasp by most users.

content/docs/user-guide/dvc-files.md
jorgeorpinel: It's already in the user guide indeed. I meant a how-to, I suppose.

content/docs/command-reference/params/index.md
shcheklein: ` # reads params.yaml, uses var` -  why does it have to use it?

skshetry: > OK so you have `var: 'value'` in params.yaml and use that value in your code.py, but you also use it as part of the definition of the stage that execs code.py?

The main use case of this is to pass values in `cmd` itself (this way it will be tracked in `params diff`, so we don't need to add it in `params` section).

```yaml
stages:
 my-thing:
   cmd: python code.py --thresh ${thresh}
```

Better to start with some simple examples.

jorgeorpinel: > why does it have to use it?

@shcheklein it doesn't have to but it could and that's the scenario I'm presenting.

> Main use: pass values in cmd itself (this way it will be tracked in params diff

Hmm... interesting. I didn't know `params` subcommands would track values if they're not in the `params` field. Let me try it out ⌛

jorgeorpinel: Oh I see what you mean: everything in params.yaml is used by `params diff` anyway (whether or not is used at all in dvc.yaml). OK, that makes sense... I still think that the feature is overloaded in a possibly confusing way but I'll leave it up to you guys.

As for the case of using the same param both in the stage def (in dvc.yaml) AND in the stage code, I'll just regard that as a side-effect which we don't need to emphasize (maybe not even mention directly) in docs for now.

> Consider your pipeline definition part of the project/your code. That's it, you parametrize all things this way.

I'll try to adopt this perspective as well to keep and generalize the "parametrization" concept/term...

shcheklein: > As for the case of using the same param both in the stage def (in dvc.yaml) AND in the stage code, I'll just regard that as a side-effect which we don't need to emphasize (maybe not even mention directly) in docs for now...

Hmm, I still don't quite understand the concern. Why should we limit it any way? Params "parametrize" _all_ the code. `dvc.yaml` is part of your code. Thus params can be used to parametrize it. It's up to users to use params anywhere they want to my mind.

jorgeorpinel: > Not sure about the naming - vars, etc.
> `Parameters can also be used as variables` - this phrase sounds a bit confusing indeed

@shcheklein I've gone for "parameteriz(e|ation)" and "templat(e|ing)" for now. I'm avoiding "variables" and also opened iterative/dvc/issues/5164 about the naming of the currently `vars` section cc @skshetry.

> Why should we limit it any way?

I'm not limiting it. I'm just not planning to complicate the docs explaining this particular case (using the same param in dvc.yaml as in the code it executes) for now. Maybe if this was a cmd ref or when/if we do extract this part of the guide into it's own (true) guide, we can then add examples and cover that specifically.

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK so on this: I've removed the `vars` section in the first comprehensive example and only left a simple example of `${}` in the 2nd stage (and a note under the code block about that, with a link to the appropriate section).

> dvc.yaml short reference, then a section about "stage definition"...

I like this idea as a first step towards a proper "guide" but I'm not sure it's within this PR's scope.

> just split it into subsections - one about regular simple pointers .dvc, second about dvc.yaml and start writing a proper guide in it

I think for a proper guide we need to write it from scratch because the current doc from the beginning is much more of a reference (without a story). In any case I've extracted this into a separate issue (which is not 2.0 specific but we can still prioritize) to not complicate this PR too much: #2059

content/docs/user-guide/dvc-files.md
jorgeorpinel: Extracted to https://github.com/iterative/dvc.org/issues/2059#issuecomment-751423317

content/docs/command-reference/params/index.md
skshetry: > Oh I see what you mean @skshetry: everything in params.yaml is used by params diff anyway

So, any files specified in the `params` section and `params.yaml` by default are shown in the `params diff`. This includes every entry, not just the specified "param" keys. That's just how `params diff` work (which might change in the future). 🤷🏽 

In addition to that, parametrized variables are also displayed on `params diff`. This will be clear if you specify vars as follows for example (and not include them on `params` section):
```
vars:
  - test_params.yaml
```

skshetry: @jorgeorpinel, please don't confuse `params` with `vars`. They are different things and for different purposes. Trying to mix them will only make it more confusing.

`dvc.yaml` always defaults to `params.yaml` for `vars` and `params`. The defaults might be making you see them as same/similar thing, but they are not. As an example:

```yaml
vars:
  - test_params.yaml
stages:
  stage1:
    cmd: "${script} --thresh ${thresh}"
    params:
      - script
      - thresh
```

Here, the `script` and `thresh` are from `params.yaml` in `params` section whereas `script` and `thresh` in the `cmd` might be coming from `test_params.yaml` or `params.yaml` (default). This is because `vars` is actually recursively merged together.

skshetry: `params` section defaults to `params.yaml` is also incorrect though. It always defaults to the `params.yaml` of the `wdir` (which defaults to `dvc.yaml`'s directory).

This is the same for `vars` as well in the `stages` though. So, that means `wdir` is always resolved from global `vars` section as local `vars` section cannot be loaded before `wdir`.

jorgeorpinel:
OK only the caveats and notes are left to integrate in terms of content here, so please feel free to start reviewing @shcheklein. One thing I can tell you already is that it's way too long... But maybe that's OK as a first step, and we can restructure for #2059?

content/docs/command-reference/params/index.md
jorgeorpinel: > In addition to that, parametrized variables are also displayed on params diff...

Good to know. A bunch of notes/clarifications are still missing from here (last check box left unmarked in the description of this PR). For now let's focus on getting the main cases clearly explained.

> params and vars are different things and for different purposes. Trying to mix them will only make it more confusing

I think they overlap a lot i.e. they have the same purpose in the end, which is "parameterizing" dvc.yaml. They are already mixed in the way this feature was designed so if that is confusing I think we should discuss separately (e.g. in https://github.com/iterative/dvc/issues/5164). For now I'm following Ivan's suggestion to just consider everything "parameterization" which I do agree now is the simplest path for this iteration of docs.

> params section ... always defaults to the params.yaml of the wdir
> the same for vars as well in the stages... wdir is always resolved from global vars section

Yep, again one of the many notes still missing, but thanks for the reminder.

.prettierrc
```diff
@@ -6,5 +6,5 @@
   "printWidth": 80,
   "tabWidth": 2,
   "useTabs": false,
-  "proseWrap": "always"
+  "proseWrap": "preserve"
```
shcheklein, line 9: not related to this PR, please let's remove it from here and have a separate PR/discussion on this.

content/docs/command-reference/params/index.md
```diff
@@ -144,6 +147,51 @@ $ dvc run -n train -d logs/ -o users.csv \
...
+
+```yaml
+stages:
+  build-us:
+    cmd: python script.py --out ${models.us.filename}
+    outs:
+      - ${models.us.filename}
+```
+
+> Note that nothing prevent you from using the same parameters in your stages
```
shcheklein, line 191: prevents

shcheklein: not sure I understand why do we need this note tbh

shcheklein: a better question would be - if I don't specify it in `params` and still use in `dvc.yaml` does it affect anything at all?  E.g. what does `dvc status` show if I change `models.us.filename` in the params file?

content/docs/command-reference/plots/index.md
```diff
@@ -125,8 +125,8 @@ All metrics files given to `dvc plots show` and `dvc plots diff` as input are
 combined together into a single data array for injection into a template file.
 There are two important fields that DVC adds to the plot data:
 
-- `index` - self-incrementing, zero-based counter for the data rows/values. In
-  many cases it corresponds to a machine learning training epoch or step number.
+- `index` - zero-based counter for the data rows/values. In many cases it
```
shcheklein, line 128: not related? please let's not include it in the PR

also, not sure even about the reason of changing this

content/docs/user-guide/dvc-files.md
```diff
@@ -143,13 +143,15 @@ stages:
     metrics:
       - performance.json
   training:
-    desc: Training stage description
-    cmd: python train.py
+    desc: Train model with Python
+    cmd:
+      - pip install -r requirements.txt
```
shcheklein, line 148: is it related? `requirements.txt` should be a dependency. It's not clear why do we make it a single stage. It's probably better to make it a separate stage

shcheklein: (I would not include and complicate it here at all)

content/docs/user-guide/dvc-files.md
```diff
@@ -159,6 +161,9 @@ stages:
     # User metadata and comments are supported.
 ```
 
+💡 You can parameterize this file with the special expression `${}`. See
```
shcheklein, line 164: should it go before the example?

content/docs/user-guide/dvc-files.md
```diff
@@ -199,23 +204,23 @@ the following fields:
...
 
 ### `dvc.lock` file
 
 For every `dvc.yaml` file, a matching `dvc.lock` (YAML) file usually exists.
 It's created or updated by DVC commands such as `dvc run` and `dvc repro`.
 `dvc.lock` describes the latest pipeline state. It has several purposes:
 
-- Tracking of intermediate and final results of a pipeline — similar to
-  [`.dvc` files](#dvc-files).
+- Tracking of intermediate and final <abbr>outputs</abbr> of a pipeline —
```
shcheklein, line 215: not related to this PR?

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+`dvc.lock`, but with an additional field with a hash of their last known
+contents. Specifically: `md5`, `etag`, or `checksum` are used (same as in `deps`
+and `outs` entries of `.dvc` files).
+
+Full <abbr>parameters</abbr> (key and value) are listed too (under `params`),
+grouped by parameters file. And in the case of templated/parameterized
+`dvc.yaml` files (see next section), their actual values are substituted into
+the `dvc.lock` YAML structure.
+
+### Parameterize `dvc.yaml`
```
shcheklein, line 259: Is there a ticket to split this properly?

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+
+`dvc.yaml` supports a special templating format that enables reusing simple
```
shcheklein, line 261: get rid of `special`

get rid of `simple`

`from different sources` - what does it mean?

`that enables reusing` - can we simplify the language?

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+values from different sources in the YAML structure itself.
+
+Global parameters (file scope) can be defined either in a
```
shcheklein, line 264: what is `file scope`? what is `global`?

Better start with "Let's say .." - give an example, explain things after

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+entries as a way to include their contents for substituting in `dvc.yaml`, e.g.:
+
+```yaml
+vars:
+  - params.json
+  - myconfig.yaml
+  - myvar: value
+```
+
+> ⚠️ Note that the default `params.yaml` is always included (completely) first.
```
shcheklein, line 318: get rid of compeletely or simplify - `all values ...`

.prettierrc
jorgeorpinel: Without this the changes to dvc.yaml samples in this PR fail checks though, so I consider it related. Is it OK if I have to skip git hooks and then merge this PR with failing checks?

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+
+It's also possible to narrow down what to include from each params file by using
```
shcheklein, line 320: is it possible for params.yaml?

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+```yaml
+${param} # Simple
+${param.key} # Nested values through . (period)
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the same expression literally in `dvc.yaml`, escape it with a
+> backslash, e.g. `\${...`.
+
+### Stage groups (loops)
```
shcheklein, line 368: groups or loops? or something that would user expects?

looks like we are not sure on the terminology

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,224 @@ stages:
...
+
+You can create loop structures in `dvc.yaml` stage entries to define more than
```
shcheklein, line 370: `loop structures` sounds too low level .. better go from the user perpsective

.prettierrc
shcheklein: What are the changes that fail?

jorgeorpinel: I've reverted this. The changes that fail are [here](https://github.com/iterative/dvc.org/pull/2052/files#diff-584bff5a435baee84112ffda33bc2d9dc94e1512e7306322f43655d5e7593876R283-R287) (see new restyled version: https://github.com/iterative/dvc.org/pull/2061/files).

jorgeorpinel: p.s. we can discuss there: https://github.com/iterative/dvc.org/pull/2061#pullrequestreview-558975989. Resolving here 🙂

content/docs/command-reference/plots/index.md
jorgeorpinel: For consistency, to match the wording in the text about indices in complex foreach groups found [here](https://github.com/iterative/dvc.org/pull/2052/files#diff-584bff5a435baee84112ffda33bc2d9dc94e1512e7306322f43655d5e7593876R401-R402)

content/docs/user-guide/dvc-files.md
jorgeorpinel: TBH I'm not sure why this isn't already in `master`. It should be (since we introduced multiple commands in dvc.yaml). Maybe a merge error somewhere.

I've extracted this one to #2062 (will roll back here before requesting another review).

content/docs/user-guide/dvc-files.md
jorgeorpinel: Extracted to #2062. Will roll back here

content/docs/command-reference/plots/index.md
shcheklein: hmm, why should the match at all? those are two completely different features?

`self-incrementing` was an important part here to my mind, btw

content/docs/user-guide/dvc-files.md
shcheklein: sounds good (but example is pretty bad to my mind, btw)

content/docs/user-guide/dvc-files.md
jorgeorpinel: > Is there a ticket to split this properly?

Yes: #2059

content/docs/command-reference/plots/index.md
jorgeorpinel: It's just nice that they match. But I don't have a strong opinion on this. Reverted.

jorgeorpinel: > self-incrementing was an important part

Actually I'm not sure that's what self-incrementing means though. Let's discuss in https://github.com/iterative/dvc.org/pull/2062#pullrequestreview-558981991?

content/docs/user-guide/dvc-files.md
jorgeorpinel: > example is pretty bad to my mind

Agree, we can discuss in https://github.com/iterative/dvc.org/pull/2062#pullrequestreview-558982138 now 🙂

content/docs/user-guide/dvc-files.md
skshetry: Better to just say `foreach` here.

content/docs/command-reference/params/index.md
```diff
@@ -28,14 +28,17 @@ YAML 1.2, JSON, TOML, or [Python](#examples-python-parameters-file) _parameters
...
+
 The default parameters file name is `params.yaml`. Parameters should be
 organized as a tree hierarchy inside, as DVC will locate param names by their
 tree path. Parameters files have to be manually written, or generated, and these
 can be versioned directly with Git.
 
 Supported parameter _value_ types are: string, integer, float, and arrays. DVC
 itself does not ascribe any specific meaning for these values. They are
-user-defined, and serve as a way to generalize and parametrize an machine
+user-defined, and serve as a way to generalize and parameterize an machine
```
skshetry, line 41: Is `parameterize` preferred compared to `parametrize`?

content/docs/command-reference/params/index.md
jorgeorpinel: > why do we need this note

We don't need it. I guess I forgot to delete it from this example after we discussed that topic. Removed

> if I don't specify it in params and still use in dvc.yaml does it affect anything

`dvc params diff` shows everything in the params file whether you use it in dvc.yaml in any way or not. It's not directly related to 2.0 dvc.yaml parameterization but we can review in the misc updates PR... Will mention you there if I change anything.

jorgeorpinel: > if I don't specify it in params and still use in dvc.yaml

If you meant values from `vars` then yes there are changes to how params diff works but I haven't finished all those notes in this PR. It's the last open check box in the description of this PR.

content/docs/user-guide/dvc-files.md
jorgeorpinel: I don't think so. The idea in this section is to introduce all the features of dvc.yaml by example first, and explain them later. The simple `${model_file}` substitution example aims to be self-explanatory (which BTW was missing from the 2nd stage's `cmd` but I just fixed that).

content/docs/user-guide/dvc-files.md
jorgeorpinel: All applied in 4aa410a6

> from different sources - what does it mean?

multiple/partial params files, global vars, local vars... Everything that is explained after this intro p. Added "(detailed below)" here.

content/docs/user-guide/dvc-files.md
jorgeorpinel: Done 🙂

content/docs/user-guide/dvc-files.md
jorgeorpinel: This has already been removed 👍

content/docs/user-guide/dvc-files.md
jorgeorpinel: > narrow down what to include from each params file
> possible for params.yaml

No. This has been clarified already 👍

content/docs/user-guide/dvc-files.md
jorgeorpinel: But what's the point of these "loops"? To generate stage groups. They get expanded in dvc.lock. I think that is the more relevant concept/term, but the "loop" part is an important detail on how to use them.

I kept both in the section title for SEO purposes but I don't have a strong opinion, I'm happy to remove "loops" from there, lmk.

jorgeorpinel: To clarify my position, here's an alternative (equally valid to my mind) way in which the syntax could've been implemented without the "foreach" term (which people may associate with "loops"):

```yaml
stages:
  mygroup:
    set:
      - foo
      - bar
    cmd: echo ${foo}
    ...
```

In the above example is clearer that the looping part is secondary, I think. Could be called "iterating over the set" instead.

> p.s. something related to this is mentioned in iterative/dvc/issues/5165

content/docs/user-guide/dvc-files.md
jorgeorpinel: Agree. OK per this other comment I'm removing the term "loop" completely for now. Going for "stage groups", "define more than one stage at a time", etc. — user perspective.

content/docs/command-reference/params/index.md
jorgeorpinel: I did a quick trends check on this and decided for that, yes. But there's no clear consensus out there.

I personally like "parameterize" better: parameter + iz(e|ation). Without the extra 'e' it seems closer to "parametric" which is slightly different.

content/docs/user-guide/dvc-files.md
shcheklein: > Better to just say foreach or foreach loops here.

it depends on what to we expect from this doc - is it a language reference? then we should name sections with keywords. But we'll need a proper document which will connect dots and explain these new concepts from the user perspective.

We've been implementing this based on a lot of tickets - how did users ask about this? what terminology they were using?

> To generate stage groups

we don't have a term `group` defined. When you introduce it if might feel that we have some special meaning to it, while it's not (there is no special concept of a group on the dvc.lock file level)

word `generate` might be a good bu itself though

jorgeorpinel: > is it a language reference? then we should name sections with keywords

BTW the `foreach` and `do` sections are named of course.

> we don't have a term group defined. When you introduce it if might feel that we have some special meaning

It's a very common word. A "group of stages" seems self-explanatory to me...?

Good idea about "generate". Introduced in 70b4b9a6.

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +240,227 @@ stages:
...
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the expression literally in `dvc.yaml`, escape it with a backslash,
+> e.g. `\${...`.
+
+### Stage groups
+
+You can generate more than one stage from a single `dvc.yaml` entry with the
+following syntax. A `foreach` element accepts a list or dictionary with values
```
jorgeorpinel, line 372: ### Terminology discussion

Summary (from https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-558972873, https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-558972942, iterative/dvc/issues/5165):

> Better to just say foreach or foreach loops?
> we'll need a proper document which will connect dots and explain these new concepts from the user perspective... implementing this based on a lot of tickets - how did users ask about this? what terminology
> loop structures sounds too low level
> They are not a group in any way
> A group is "more than one of the same thing, regarded as a unit"
> in our case it's not a group, these stages are not different from any other stages
> On the dvc.yaml level it's indeed looks like a single "entity"

Takeaways:

- we don't like group, because group is a very specific term in math/languages
- totally fine to use "group" in some informal way. The misleading would be to make it formal (use in title
- @skshetry can you maybe help me find these original tickets or requests from users to check on that? Thanks

content/docs/user-guide/dvc-files.md
jorgeorpinel: Continued in https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-559754434. Resolving this one.

content/docs/user-guide/dvc-files.md
jorgeorpinel: For now I'm using "Defining multiple stages at once" as title and introduced the term "multi-stage".

content/docs/command-reference/params/index.md
skshetry: > Without the extra 'e', the root seems closer to "parametric", which is slightly different.

Which is what is it. Anyway, it [looks like `parameterize` is dominant form in US-English](https://en.wiktionary.org/wiki/parameterize).

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +240,233 @@ stages:
...
+stages:
+  build-us:
+    desc: ${desc}
+    cmd: python train.py --thresh ${models.us.threshold}
+```
+
+⚠️ DVC merges external params and `vars` as long as there are no leaf node
+collisions (so the two examples above can't be used simultaneously). For
+example, `{"grp": {"a": 1}}` can be merged with `{"grp": {"b": 2}}`, but not
+with `{"grp": {"a": 7}}`.
```
jorgeorpinel, line 310: This note is a little complicated but I'm not sure how to explain this in easier terms.

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +240,233 @@ stages:
...
+        - ${item.prop2}
+```
 
-Full <abbr>parameters</abbr> (key and value) are listed separately under
-`params`, grouped by parameters file.
+> ⚠️ Remember that DVC only guarantees stage execution order based on
+> [DAGs](/doc/command-reference/dag). This means that if the resulting stages
+> are not connected by their <abbr>outputs</abbr> and <abbr>dependencies</abbr>,
+> they may not be [reproed](/doc/command-reference/repro) in the order in which
+> they are defined.
```
jorgeorpinel, line 472: Should we also insert this note in `repro` or somewhere else? (separate to this PR) It's not really mentioned anywhere.

> Related to iterative/dvc/issues/5181

skshetry: I don't find this information necessary.

content/docs/user-guide/dvc-files.md
```diff
@@ -237,14 +242,245 @@ stages:
...
+    foreach: # List of simple values
+      - foo
+      - bar
+      - baz qux
+    do:
+      cmd: echo "${item}"
+```
+
+Upon `dvc repro`, each item in the list is expanded into its own stage by
+substituting is value in expression `${item}`. The item's value is appended to
```
skshetry, line 402:
```suggestion
substituting its value in expression `${item}`. The item's value is appended to
```

jorgeorpinel: Thanks!

jorgeorpinel:
@skshetry thanks for reviewing this and the couple comments you left. So, do you approve of it?

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK. I think it's a relevant warning to set expectations but we would need to mention it in other places too (out of scope for this PR). Let's see what Ivan thinks.

content/docs/user-guide/dvc-files.md
```diff
@@ -118,14 +118,14 @@ and `dvc commit` commands, but not when a `.dvc` file is overwritten by
 
 ## `dvc.yaml` file
 
-`dvc.yaml` files describe data science or machine learning pipelines, similar to
+`dvc.yaml` files describe data science or machine learning pipelines (similar to
 how [Makefiles](https://www.gnu.org/software/make/manual/make.html#Introduction)
-work for building software. Its YAML structure contains a list of stages which
+work for building software). Its YAML structure contains a list of stages which
```
jorgeorpinel, line 123: @shcheklein FYI some of the first few changes in this file I missed moving to the misc. updates PR we just merged, sorry! They're just small copy edits...

content/docs/user-guide/dvc-files.md
shcheklein: I would not complicate this Guide with it. It's more about `repro` if we decide to mention it all. Let's do it separately please.

content/docs/command-reference/params/index.md
```diff
@@ -36,6 +36,9 @@ stages:
           - batch_size
 ```
 
+💡 Parameters can also be used directly in `dvc.yaml` itself. See the
```
shcheklein, line 39: It's too abrupt here, in this context  `can also be used directly in `dvc.yaml` itself` is very hard to read properly.

content/docs/command-reference/params/index.md
```diff
@@ -165,6 +168,50 @@ $ dvc run -n train -d logs/ -o users.csv \
...
+params.yaml  process.thresh  —      0.98
+params.yaml  train.epochs    —      70
+params.yaml  train.layers    —      9
+```
+
+This command shows the difference in parameters between the workspace and the
+last committed version of the `params.yaml` file. In our example there's no
+previous version, which is why all `Old` values are `—`.
+
+## Example: Parameterize stages in `dvc.yaml`
```
shcheklein, line 190: To be honest, the whole `dvc params` is more about parameter dependencies, not sure this example is relevant at all.

content/docs/user-guide/dvc-files.md
```diff
@@ -118,14 +118,14 @@ and `dvc commit` commands, but not when a `.dvc` file is overwritten by
 
 ## `dvc.yaml` file
```
shcheklein, line 119: just to reiterate, the whole document is overloaded, a lot of notes, a lot of ad-hoc structure - let's split this into a separate page right after we merge this

content/docs/user-guide/dvc-files.md
```diff
@@ -161,22 +161,27 @@ stages:
     # User metadata and comments are supported.
 ```
 
-`dvc.yaml` files consists of a group of `stages` with names provided explicitly
-by the user with the `--name` (`-n`) option of `dvc run`. Each stage can contain
+💡 You can [parameterize this file](#parameterize-dvcyaml) with the special
```
shcheklein, line 164: we should have this references, there should be some general description, then basic layout, then other syntax like templating, loops, etc. Again, let's refer to some good examples that describe languages. (Btw, templating is a way better term in our case than parametrization to my mind)

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
-`dvc.lock`, but with an additional field to store the hash value of each file or
-directory tracked by DVC. Specifically: `md5`, `etag`, or `checksum` (same as in
-`deps` and `outs` entries of `.dvc` files).
+`dvc.lock`, but with an additional field with a hash of their last known
+contents. Specifically: `md5`, `etag`, or `checksum` are used (same as in `deps`
+and `outs` entries of `.dvc` files).
+
+Full <abbr>parameter dependencies</abbr> (key and value) are listed too (under
+`params`), grouped by parameters file. And in the case of
+templated/parameterized `dvc.yaml` files (see next section), their actual values
```
shcheklein, line 262: let's agree on using one of those consistently. Looking though the docs (and considering the semantics of these constructs), it feels like templating is a better option. Also it doesn't overlap with parameters that much then.

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+are substituted into the `dvc.lock` YAML structure.
+
+### Parameterize `dvc.yaml`
+
+`dvc.yaml` supports a templating format to reuse values from different sources
```
shcheklein, line 267: here you use `templating` for example

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+${param.key} # Nested values through . (period)
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the expression literally in `dvc.yaml`, escape it with a backslash,
+> e.g. `\${...`.
+
+### Defining multiple stages at once
+
+You can generate more than one stage from a single `dvc.yaml` entry with the
```
shcheklein, line 391: if you use `generate` here, why don't you use in the title?

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+`params`), grouped by parameters file. And in the case of
+templated/parameterized `dvc.yaml` files (see next section), their actual values
+are substituted into the `dvc.lock` YAML structure.
+
+### Parameterize `dvc.yaml`
+
+`dvc.yaml` supports a templating format to reuse values from different sources
+in the YAML structure itself. The sources can be external
+[parameters files](/doc/command-reference/params), or internal `vars`, as
+detailed below.
```
shcheklein, line 270: `reuse values` - why `reuse`? I'm not sure I understand this ...

`The sources can be external
[parameters files](/doc/command-reference/params), or internal `vars`, as
detailed below.` 

this explanation is too heavy, implicit, abrupt ... start from simple ideas and example, provide details later

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+    desc: ${desc}
+    cmd: python train.py --thresh ${models.us.threshold}
+```
+
+> DVC merges external params and `vars` as long as there are no leaf node
+> collisions (so the two examples above can't be used simultaneously). For
+> example, `{"grp": {"a": 1}}` can be merged with `{"grp": {"b": 2}}`, but not
+> with `{"grp": {"a": 7}}`.
+
+⚠️ Note that doesn't track values from `vars` like parameters (ignored by
```
shcheklein, line 323: it's not needed I think

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+        threshold: 10
+  - desc: 'Reusable description'
+
+stages:
+  build-us:
+    desc: ${desc}
+    cmd: python train.py --thresh ${models.us.threshold}
+```
+
+> DVC merges external params and `vars` as long as there are no leaf node
```
shcheklein, line 318: what are `external params`?

shcheklein: there should be definitely a way to write it simpler ...

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+> example, `{"grp": {"a": 1}}` can be merged with `{"grp": {"b": 2}}`, but not
+> with `{"grp": {"a": 7}}`.
+
+⚠️ Note that doesn't track values from `vars` like parameters (ignored by
+`dvc params diff`).
+
+To load additional params files, list them in the top `vars`, in the desired
+order, e.g.:
+
+> Their paths will be evaluated based on `wdir`, if one given.
```
shcheklein, line 329: ? where that `wdir` is specified?

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+
+```yaml
+vars:
+  - params.json
+  - myvar: 'value'
+  - config/myapp.yaml
+```
+
+(ℹ️) Note that the default `params.yaml` file is always loaded first, if
+present.
```
shcheklein, line 339: if present where?

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+Stage-specific values are also supported, with inner `vars`. You may also load
+additional params files locally. For example:
+
+```yaml
+stages:
+  build-us:
+    vars:
+      - params.json:build
+      - model:
+        filename: 'model-us.hdf5'
```
shcheklein, line 366: indent?

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+    cmd: python train.py ${build.epochs} --out ${model.filename}
+    outs:
+      - ${model.filename}
+```
+
+⚠️ Important: Limitations of local `vars`:
```
shcheklein, line 372: How important are they to attract so much attention? I would say they are just "Known limitations .."

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+
+- `wdir` cannot use values from local `vars`, as DVC uses the working directory
+  first (to load any values from params files listed in `vars`).
+- `foreach` is also incompatible with local `vars` at the moment.
+
+The substitution expression supports these forms:
```
shcheklein, line 378: just to check, how correct the `substitution expression` term is?

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+      python train.py
+      --thresh ${models.us.threshold}
+      --out ${models.us.filename}
+    outs:
+      - ${models.us.filename}:
+        cache: true
+```
+<!-- prettier-ignore-end -->
+
+DVC will track simple param values (numbers, strings, etc.) used in `${}` (they
```
shcheklein, line 299: ?

content/docs/user-guide/dvc-files.md
shcheklein: it also feels that we should define this way before we jumped into different options for `vars`

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+```yaml
+${param} # Simple
+${param.key} # Nested values through . (period)
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the expression literally in `dvc.yaml`, escape it with a backslash,
+> e.g. `\${...`.
+
+### Defining multiple stages at once
```
shcheklein, line 389: how about `foreach`/`do` - using loops to generate stages

content/docs/user-guide/dvc-files.md
```diff
@@ -240,14 +248,245 @@ stages:
...
+You can generate more than one stage from a single `dvc.yaml` entry with the
+following syntax. A `foreach` element accepts a list or dictionary with values
+to iterate on, while `do` contains the regular stage fields (`cmd`, `outs`,
+etc.). Here's a simple example:
+
+```yaml
+stages:
+  echo: # Multi-stage
+    foreach: # List of simple values
+      - foo
```
shcheklein, line 400: can we something more realistic (keep it simple though)?

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK sure, will extract!

content/docs/user-guide/dvc-files.md
jorgeorpinel: Example removed. The note in the description now links to the section in dvc.yaml

content/docs/command-reference/params/index.md
jorgeorpinel: Example removed. The note in the description now links to the section in dvc.yaml

content/docs/user-guide/dvc-files.md
jorgeorpinel: Agree. I mentioned this in https://github.com/iterative/dvc.org/pull/2052#issuecomment-751533344. I'm just not sure if the split can focus just on this new content or if we need to rethink the whole dvc.yaml ref vs. guide thing (#2059)...

jorgeorpinel: OK so after reading some other comments I started splitting this doc into several pages...

content/docs/user-guide/dvc-files.md
jorgeorpinel: Well for now this is all still in guides so it's still not "right" that we have references here. We could have the dvc.yaml "langauge" ref. as its own section (before Command Reference, for example but that change I would definitely try to avoid putting here (let's just prioritize #2059 to right after this?)

I still need to do that research on other tech refs and guides ⌛

content/docs/user-guide/dvc-files.md
jorgeorpinel: Agree. I'm using "template" primarily now.

content/docs/user-guide/dvc-files.md
jorgeorpinel: Normally I avoid being repetitive in these cases. Similar to not using a word to define that word. Also having more that one word helps make sure people get the right idea. It's just a general writing practice I guess — That's to answer the question.

But yes, perhaps "Generating" is better for the title. Applied (but I moved "define" to the paragraph).

content/docs/user-guide/dvc-files.md
jorgeorpinel: BTW I removed those 💡 notes from the main dvc.yaml page (just added a gray note linking to the Advanced dvc.yaml page). We can keep that `${model_file}` in the example right? No strong opinion but it could spark people's curiosity (to find the advanced page).

content/docs/user-guide/dvc-files.md
jorgeorpinel: > why reuse?

You don't have to type the same value over and over.

> explanation is too heavy, implicit, abrupt

2 sentences is too much for an intro? This is how it looks:

image

I would just be removing the last sentence and put that info somewhere later. It seems mostly a stylistic decision to me.

jorgeorpinel: re `reuse` - I changed it to "insert" now.

content/docs/user-guide/dvc-files.md
jorgeorpinel: @skshetry mentions when values are tracked and when they're not in https://github.com/iterative/dvc/wiki/Parametrization so I assumed it was important. I personally think it can be relevant to (be able to) know this, but no strong opinion (b/c on the other hand it's kind of expected/intuitive). Up to you, please lmk 🙂

content/docs/user-guide/dvc-files.md
jorgeorpinel: "external params" alluded to the contrast with "internal" `vars`, which was established in the intro of the section. But I've removed these terms now (c3d82fdc).

jorgeorpinel: > way to write it simpler

Do you mean just "external"/"internal"? (fixed) or the whole paragraph (as mentioned in https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-560721648) ?

jorgeorpinel: Anyway, I simplified the note now (c3eefbac).

content/docs/user-guide/dvc-files.md
jorgeorpinel: > where that wdir is specified?

In the bullet list of dvc.yaml fields, which used to be in the same (super long) page but now it's [in another page](https://dvc-landing-2-0-pipelin-enhjqc.herokuapp.com/doc/user-guide/dvc-files/dvc.yaml#accepted-fields) so this may be a problem 🤔

jorgeorpinel: OK I linked `wdir` (2 instances in this page) to the field reference `/doc/user-guide/dvc-files/dvc.yaml#accepted-fields` where that is described (for now at least).

content/docs/user-guide/dvc-files.md
jorgeorpinel: Just present. If it exists.

content/docs/user-guide/dvc-files.md
jorgeorpinel: Yes, thanks. Indented!

content/docs/user-guide/dvc-files.md
jorgeorpinel: True. Changed for "Known limitations of local `vars`".

content/docs/user-guide/dvc-files.md
jorgeorpinel: > how correct the substitution expression term is?

Will research ⌛ 

> should define this way before

We do intro the term early on before using `${}` in any example. Do you mean putting "substitution expression supports these forms..." before dvc.yaml examples?

content/docs/user-guide/dvc-files.md
jorgeorpinel: > `... doesn't track values from `vars` like parameters`

p.s. for now I made it into a grayed note.

content/docs/user-guide/dvc-files.md
jorgeorpinel: Same answer as https://github.com/iterative/dvc.org/pull/2052#discussion_r553765332, I think (not sure what the question is).

content/docs/user-guide/dvc-files.md
jorgeorpinel: I'm not sure we want to introduce the concept of "loops" (even when its implied by the `foreach` analogy). Per https://github.com/iterative/dvc/issues/5181#issuecomment-752326990 the items given to `foreach` are not ordered so it doesn't seem like an actual loop to me.

jorgeorpinel: \* not considered ordered

jorgeorpinel: (This title is **Generating multiple stages** now BTW.)

content/docs/user-guide/dvc-files.md
jorgeorpinel: Changes in 228da879.

shcheklein:
Okay, it's impossible to iterate on this PR further (GH breaks at this stage, it's impossible to follow all the comments, etc). Let's merge and work separately on the next version. The biggest concern is still structure:

Screen Shot 2021-01-08 at 1 58 52 PM

All these titles do not look good to me. They are fine for Basic Concepts I guess, but not as a top level section in the User Guide. They should higher level.

After that we define the expected structure we can start iterating on the actual content. 

- `advanced-dvc.yaml` - is too heavy and a mix of a reference (precise spec) and guide (titles are high level, no structure - story like)
- do we need two seprate `dvc.yaml` sections? why do we consider basic templating and advanced feature?
- `.dvc file` - is clearly a reference thing. Where do we put it?

shcheklein (approved):

skshetry:
> All these titles do not look good to me. They are fine for Basic Concepts I guess, but not as a top level section in the User Guide. They should higher level.

I don't understand why we don't use right-side TOC instead. I find traversing through those pages difficult than if they were in a single page (and the left-side navbar is deeply nested making it even harder).

Why is it neglected so much? 

It feels like we are splitting pages arbitrarily (just look at [installation instructions](https://dvc.org/doc/install)).

shcheklein:
@skshetry I agree that structure is not ideal, we'll iterate on it next thing. Let's merge it and do it in the next PR, this is one is not manageable.

Re the split/non-split - to the large extent if feels that it's a matter of preferences and we won't find a solution that fits all. I think the current approach is to keep pages reasonably short, to a single point (e.g. it's bad to have .dvc and the whole dvc.yaml - they are too different concepts. `Can dvc.yaml` fit a single page - good question-let's see in the next PR. I really hope so, I'm though worried that about some points I describe below, e.g. complexity of the right hand bar).

Usual reasons to split (besides personal preferences, which should not serve as an argument unless we do a focus group testing):

- have a clear link to separate page that describes one thing (good for SEO, since term becomes part of the URL, page has tailored meta section, etc)
- when you have too much on a single page- you won't have any navigation at all (or we'll have to complicate right hand side bar)

> Why is it neglected so much?

Nothing is neglected (e.g. we changed Get started when we had time from 10+ pages to 4). But `I find traversing through those pages difficult` is not an argument that we can use as a guideline to my mind. If you have some good structure or guideline in mind, please suggest it.

jorgeorpinel:
Merged and moved discussion to https://github.com/iterative/dvc.org/issues/2059#issuecomment-758510934.

jorgeorpinel:
> It feels like we are splitting pages arbitrarily (just look at installation instructions).

@skshetry I don't see how Install pages are spit arbitrarily. But if you'd like to elaborate or have better examples feel free to open a separate issue. Thanks

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK @shcheklein, on the question of the "substitution expression" I here some the alternatives:

- We call it "`${}` expression" and use "interpolation" to explain it in the original [wiki page](https://github.com/iterative/dvc/wiki/Parametrization), but that latter term seemed too mathematical to me.
- [Helm](https://helm.sh/docs/chart_template_guide/getting_started/) (K8 YAML templating) calls it a "directive" (e.g. `{{ .Something }}`) and uses words "value", "generated", and "render/inject/insert" for the explanations.
  > BTW apparently there are some [strong opinions](http://leebriggs.co.uk/blog/2019/02/07/why-are-we-templating-yaml.html) out there against templating YAML 😆
- [Kustomize](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/) (YAML) uses a different approach (template-free) with terms like "generate", "variants" and "overlays" — not directly comparable but listing here because it's an interesting alternative, maybe we can consider adopting something like that in the future.
- [Handlebars](https://handlebarsjs.com/guide/) (HTML/JS) calls them "handlebars expressions" (e.g. `{{prop}}`); uses "generate/render", "property/value", "replace"
- [Twig](https://twig.symfony.com/doc/3.x/api.html)/[Django](https://docs.djangoproject.com/en/3.1/topics/templates/) (HTML/PHP): "tags/filters" (e.g. `{% var %}`, `{{ function }}`); "render", "generate", "values"
- [Mustache](https://mustache.github.io/mustache.5.html) (Ruby, JS, etc.): "tags" (e.g. `{{key}}`); "render", "replace", "values".
- [Go](https://golang.org/pkg/text/template/) (text): "actions" (e.g. `{{ "string" }}`); "values", "evaluation", "generate"
- [Bash](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html) calls `$(command)` a "command substitution"

Conclusions:
- "Substitution" is not very used out there, I guess the `$` in our syntax reminded me more of Bash than of modern templating engines.
- For `${}`, just "expression" is OK. Most templating formats have a special term though, usually "tag" for value injections. 
- For related terms, I think in our case "replace" and "generate" are the most appropriate.

PR #2053, opened 2020-12-24
Misc updates (1.0)

jorgeorpinel:
Old stashed changes that apply to both v1 and thus master (2.0)

- [ ] This can be safely backported to v1 up to 79ec70b943eb780d6aa04a57b9f6d9080c8dbbfd.

PR #2062, opened 2020-12-28
Misc. updates (2.0ish)

jorgeorpinel:
- Some copy edits loosely realated to the work in #2052
- ~~Review~~Reorg/rewrite of params cmd refs.


content/docs/command-reference/plots/index.md
```diff
@@ -125,8 +125,8 @@ All metrics files given to `dvc plots show` and `dvc plots diff` as input are
 combined together into a single data array for injection into a template file.
 There are two important fields that DVC adds to the plot data:
 
-- `index` - self-incrementing, zero-based counter for the data rows/values. In
-  many cases it corresponds to a machine learning training epoch or step number.
+- `index` - zero-based counter for the data rows/values. In many cases it
```
jorgeorpinel, line 128: In my experience "self-incrementing" (more commonly "auto-increment" means something else: when you insert records (e.g. to a DB table) and the index field gets an automatic value. Numbers may be skipped if records are deleted too.

In this case all the data is already there and we're just assigning a linear integer scale to the automatic index field.

content/docs/user-guide/dvc-files.md
```diff
@@ -143,9 +143,12 @@ stages:
...
   training:
-    desc: Training stage description
-    cmd: python train.py
+    desc: Train model with Python
+    cmd:
+      - pip install -r requirements.txt
+      - python train.py --out ${model_file}
     deps:
+      - requirements.txt
       - train.py
```
jorgeorpinel, line 152: This example is not great...

jorgeorpinel: This was introduced in https://github.com/iterative/dvc.org/pull/1980/files#diff-101f4b5763f1f071bfa5c855cf2ed36fa5be72531acaa2c0beb2245da215c6d6 so I'm not sure why it's not already in master. Probably a bad merge somewhere...

content/docs/user-guide/dvc-files.md
```diff
@@ -143,9 +143,12 @@ stages:
...
       - performance.json
   training:
-    desc: Training stage description
-    cmd: python train.py
+    desc: Train model with Python
+    cmd:
+      - pip install -r requirements.txt
+      - python train.py --out ${model_file}
     deps:
+      - requirements.txt
```
jorgeorpinel, line 151: This is being reintroduced after rolling it back accidentally in https://github.com/iterative/dvc.org/pull/2053/files#diff-584bff5a435baee84112ffda33bc2d9dc94e1512e7306322f43655d5e7593876

jorgeorpinel: In any case, this isn't a great example. What is the main use case for multiple commands in a single stage (sharing deps and outs)? For now I've asked in https://github.com/iterative/dvc/issues/4373#issuecomment-751830121

content/docs/command-reference/params/diff.md
```diff
@@ -19,21 +19,24 @@ positional arguments:
...
+repository history. Requires that Git is being used to version the project
+params.
 
 > Parameter dependencies are defined with the `-p` option in `dvc run`. See also
 > `dvc params`.
 
 Without arguments, this command compares parameters currently present in the
 <abbr>workspace</abbr> (uncommitted changes) with the latest committed version.
+This includes everything in the default parameters file (`params.yaml`) as well
+as `params` found in `dvc.yaml` files.
```
jorgeorpinel, line 32: @shcheklein tagging you here as promised. Here's the clarification to your question in https://github.com/iterative/dvc.org/pull/2052#discussion_r549186721. I ended up reviewing all of the description here as well as in the index of params...

content/blog/2020-04-06-april-20-dvc-heartbeat.md
```diff
@@ -15,12 +15,11 @@ descriptionLong: |
...
-  Cobre](https://en.wikipedia.org/wiki/Copper_Canyon), shot by Jorge Orpinel
-  Pérez. Jorge has mastered the art of working on DVC remotely.
-author: elle_obrien
-commentsUrl: https://discuss.dvc.org/t/april-20-heartbeat/347
-tags:
+  A view from
+  [Barrancas del Cobre](https://en.wikipedia.org/wiki/Copper_Canyon), shot by
+  Jorge Orpinel Pérez. Jorge has mastered the art of working on DVC remotely.
+  author: elle_obrien commentsUrl:
+  https://discuss.dvc.org/t/april-20-heartbeat/347 tags:
```
shcheklein, line 22: something funny is happening here with tags

content/docs/command-reference/params/diff.md
shcheklein: `as `params` found in `dvc.yaml` files.` what does it exactly mean?

content/docs/command-reference/params/diff.md
```diff
@@ -19,21 +19,24 @@ positional arguments:
...
+
+> Note that by default, `dvc params diff` only shows parameters that were
```
shcheklein, line 34: this what would expect from _diff_ , not sure we need this note

shcheklein: or we can make it part of the first paragraph

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,44 @@ positional arguments:
 
 In order to track parameters and hyperparameters associated to machine learning
 experiments in <abbr>DVC projects</abbr>, DVC provides a different type of
-dependencies: _parameters_. Parameters are defined using the the `-p`
-(`--params`) option of `dvc run`, using simple names like `epochs`,
-`learning-rate`, `batch_size`, etc.
+dependencies: _parameters_. Params are defined in the `params` field of
```
shcheklein, line 21: `or using ...` should be `corresponds to`

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,44 @@ positional arguments:
...
+`dvc.yaml` or using the the `-p` (`--params`) option of `dvc run`. They usually
+have simple names like `epochs`, `learning-rate`, `batch_size`, etc.
```
shcheklein, line 23: better to move this sentence up

content/docs/command-reference/params/index.md
shcheklein, line 21: Params -> they (to make it clear that this defines dependency, params are defined in the params.yaml)

content/docs/command-reference/plots/index.md
shcheklein: kk

content/docs/user-guide/dvc-files.md
shcheklein: let's not complicate the base example then for now

content/blog/2020-04-06-april-20-dvc-heartbeat.md
jorgeorpinel: Indeed. My bad (VSCode paragraph wrap). Fixed.

content/blog/2020-04-06-april-20-dvc-heartbeat.md
```diff
@@ -14,10 +14,11 @@ descriptionLong: |
   projects by our users and big ideas about best practices in ML and data
   science.
 picture: 2020-04-06/april_header.png
+# prettier-ignore
 pictureComment:
-  A view from [Barrancas del
-  Cobre](https://en.wikipedia.org/wiki/Copper_Canyon), shot by Jorge Orpinel
-  Pérez. Jorge has mastered the art of working on DVC remotely.
+  A view from
+  [Barrancas del Cobre](https://en.wikipedia.org/wiki/Copper_Canyon), shot by
```
jorgeorpinel, line 20: BTW funny that we need to prettier-ignore inside frontmatter so that Pretty doesn't break the same format it enforces in regular md haha. I guess it has no way to know that this string in YAML is supposed to be md.

content/docs/command-reference/params/diff.md
jorgeorpinel: "`params` found in `dvc.yaml` files" means the parameters tracked by DVC which may or may not match what's in the params file(s) in the workspace.

I added some missing context around this in ad461309.

content/docs/command-reference/params/diff.md
jorgeorpinel: > by default, `dvc params diff` only shows parameters that were changed

I think it's a relevant note. I shortened it and moved it to the paragraph.

content/docs/command-reference/params/index.md
jorgeorpinel: Agree. This has already been updated: the `run` part is a parenthesis now.

content/docs/command-reference/params/index.md
jorgeorpinel: Do you mean "They usually have simple names ..."? That looks like the least relevant sentence, it's a list of examples.

No strong opinion but what do you mean by move up? Up one-over? Start the paragraph with it?

content/blog/2020-04-06-april-20-dvc-heartbeat.md
shcheklein: hmm ... that's strange - I though that prettier doesn't touch frontmatter (it def doesn't wrap lines for example)

content/docs/command-reference/params/diff.md
```diff
@@ -19,21 +19,22 @@ positional arguments:
...
-This command provides a quick way to compare parameter values among experiments
-in the repository history. Requires that Git is being used to version the
-project params.
+Provides a quick way to compare parameter values among experiments in the
+repository history. Requires that Git is being used to version the project
+params.
 
-> Parameter dependencies are defined with the `-p` option in `dvc run`. See also
-> `dvc params`.
+> Parameters are defined in the `params` field of `dvc.yaml` (e.g. with the the
```
shcheklein, line 26: so, here is another example - I think `Parameter dependencies` was the right form

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,44 @@ positional arguments:
 
 In order to track parameters and hyperparameters associated to machine learning
 experiments in <abbr>DVC projects</abbr>, DVC provides a different type of
-dependencies: _parameters_. Parameters are defined using the the `-p`
-(`--params`) option of `dvc run`, using simple names like `epochs`,
-`learning-rate`, `batch_size`, etc.
+dependencies: _parameters_. They are defined in the `params` field of `dvc.yaml`
+(e.g. with the the `-p` (`--params`) option of `dvc run`). Params usually have
```
shcheklein, line 22: I guess my point here is that this sentence is way more valuable than anything else around- it is something that immediately make it easier to understand what Parameters we talk about.

content/docs/command-reference/params/index.md
jorgeorpinel: ✔️  Params->They, The-> Params

content/blog/2020-04-06-april-20-dvc-heartbeat.md
jorgeorpinel: It's wrapping them (no changes to config), and since it doesn't know it's md, it will break the link.

jorgeorpinel: I mean "break" it into 2 lines (the link still works I think).

content/docs/command-reference/params/diff.md
jorgeorpinel: I was trying to simplify the language but we do use "param dep" a lot already so I reverted this to not mess with that for now.

Another example of? Sorry, didn't get that part.

content/docs/command-reference/params/index.md
jorgeorpinel: OK. This made me realize we should separate the part about how they are defined in dvc.yaml from this first p so I started doing that and ended up moving and rewriting a lot of the text again... I also introduced some structure and inner anchor links to avoid repetition. PTAL

content/docs/user-guide/dvc-files.md
```diff
@@ -163,15 +166,19 @@ stages:
...
   file's location). If this field is not present explicitly, it defaults to `.`
   (the file's location).
 - `deps`: List of <abbr>dependency</abbr> file or directory paths of this stage
   (relative to `wdir` which defaults to the file's location). See
   [Dependency entries](#dependency-entries) above for more details.
-- `params`: List of <abbr>parameter</abbr> dependency keys (field names) that
-  are read from a YAML, JSON, TOML, or Python file (`params.yaml` by default)
+- `params`: List of <abbr>parameter</abbr> dependency keys (field names) to
+  track in `params.yaml`. The list may also contain other YAML, JSON, TOML, or
+  Python file names, with a sub-list of the param names to track in them.
```
jorgeorpinel, line 181: Note that I also expanded on this dvc.yaml field since we're encouraging users to avoid dvc run now (so we should betterexplain all the possible ways to write this manually or generate it).

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK I'm all about letting this be for now. But just FYI, there's some discussion on this question in https://github.com/iterative/dvc/issues/4373#issuecomment-751996212.

My conclusions are that:
- This is mainly syntax sugar as an alt. to `&&` (which could be quite nice in some cases like setting up a bunch of env vars at once).
- Another good use could be to add "parallel" stages in a single definition: while not the cleanest dvc.yaml code, it would save extra lines.
- If neither of those examples seem like a strong/real user need for this, it could be un-documented for simplicity.

content/blog/2020-04-06-april-20-dvc-heartbeat.md
shcheklein: okay, try to use `pictureComment: |` (Block Style Indicator)

content/docs/user-guide/dvc-files.md
shcheklein: so, there were some good points and example in that discussion (e.g. `cd`) - should we take a real example?

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,61 @@ positional arguments:
...
 In contrast to a regular <abbr>dependency</abbr>, a parameter is not a file (or
-directory). Instead, it consists of a _parameter name_ (or key) to find inside a
-YAML 1.2, JSON, TOML, or [Python](#examples-python-parameters-file) _parameters
-file_. Multiple parameter dependencies can be specified from one or more
-parameters files.
+directory). Instead, it consists of a _parameter name_ (or key) in a
+[parameters file](#parameters-file), where the _parameter value_ should be
+found. This allows you to define [stage](/doc/command-reference/run)
+dependencies more granularly: changes to other parts of the params file will not
+affect the stage. Parameters also prevent situations where several stages share
```
shcheklein, line 29: Parameter dependencies ?

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,61 @@ positional arguments:
...
+a common dependency (e.g. a config file), and any change in it invalidates all
+these stages, causing unnecessary re-executions upon `dvc repro`.
+
+To start tracking parameters from the default `params.yaml` file, list them
```
shcheklein, line 33: I would put this paragraph and example above the previous long explanation. Especially this short example.

shcheklein: To start tracking parameters ... (let's not specify in this way that they are from params.yaml - you put too much focus on a spec detail). I would split this:

To start tracking parameters, specify under the `params` field of `dvc.yaml` stages (manually or with the the
`-p`/`--params` option of `dvc run`):

```
example
```

all other long explanations here ....

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,61 @@ positional arguments:
...
 
-DVC saves the param names and their latest values in the `dvc.yaml` file. These
-values will be compared to the ones in the params files to determine if the
-stage is invalidated upon pipeline [reproduction](/doc/command-reference/repro).
+### Parameters files
 
-> Note that DVC does not pass the parameter values to stage commands. The
-> associated command executed by `dvc run` or `dvc repro` will have to open and
-> parse the parameters file by itself, and use the params specified with `-p`.
+The default params file name is `params.yaml`, but any other YAML 1.2, JSON,
```
shcheklein, line 54: it feels that we mention this default thingy in every paragraph

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,61 @@ positional arguments:
...
+generated) and they can be versioned directly with Git.
 
-The parameters concept helps to define [stage](/doc/command-reference/run)
-dependencies more granularly. A particular parameter or set of parameters will
-be required for the stage invalidation (see `dvc status` and `dvc repro`).
-Changes to other parts of the dependency file will not affect the stage. This
-prevents situations where several stages share a (configuration) file as a
-common dependency, and any change in this dependency invalidates all these
-stages and causes their reproduction unnecessarily.
+Param values should be organized in tree-like hierarchies (dictionaries) inside
```
shcheklein, line 59: you have `Param values` here and then immediately `### Parameter values` as a title - looks strange a bit

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,61 @@ positional arguments:
...
-itself does not ascribe any specific meaning for these values. They are
-user-defined, and serve as a way to generalize and parametrize an machine
-learning algorithms or data processing code.
+The `dvc params diff` command is available to show parameter changes, displaying
+their current and previous [values](#parameter-values).
 
-DVC saves the param names and their latest values in the `dvc.yaml` file. These
-values will be compared to the ones in the params files to determine if the
-stage is invalidated upon pipeline [reproduction](/doc/command-reference/repro).
+### Parameters files
```
shcheklein, line 52: This content should be simple enough to avoid additional structure to my mind. Keep it simpler, remove repetitions, move from high level explanation (an example) to the details or/and advanced cases (custom file name).

Also keep in mind, that `dvc params diff` can actually operate on any file now, not even registered in `dvc.yaml`.

content/docs/user-guide/dvc-files.md
jorgeorpinel: OK looking in https://github.com/search?q=filename%3Advc.yaml+%26%26&type=code ⌛

content/blog/2020-04-06-april-20-dvc-heartbeat.md
jorgeorpinel: 🤦 yeah I didn't think about that for some reason. That works 👍

content/docs/command-reference/params/index.md
jorgeorpinel: Hmmm OK, changes.

But I'm not sure about the phrase "parameter dependency" in general. It would be simpler to only say "parameter" most of the time. I guess since we already use the full phrase everywhere in `params diff` and especially since there will be dvc.yaml parameters (`${param}`) now (which are not stage dependencies) maybe it's a good idea to keep the phrase in these refs.

jorgeorpinel: > especially since there will be dvc.yaml parameters (${param}) now

OK, I changed some instances to just "params" and other to "param deps" depending on that. See 7165723e.

content/docs/command-reference/params/index.md
jorgeorpinel: Good idea, I started with the example first now.

Keep in mind that a lot of this content will be extracted to the basic concepts section as soon as I have time to take that project again so let's not focus too much on the exact order of ideas here? But I did change it like you suggested 🙂

content/docs/command-reference/params/index.md
jorgeorpinel: > we mention this default thingy in every paragraph

It's only mentioned once in a note after the desc. example now, and in this **Params file** section.

content/docs/command-reference/params/index.md
jorgeorpinel: > content should be simple enough to avoid additional structure

Yes that would be ideal although I'm not sure that sections hurt in this instance (there's several relevant aspects of params so maybe sections are appropriate for a reference doc.)

But 1. all of this content was already there (in fact this PR already makes the text shorter compared to https://dvc.org/doc/command-reference/params and 2. again, since we will move a lot of this info to basic concepts soonish, should we let it be for now?

> keep in mind, that dvc params diff can actually operate on any file now

Good point. It will also operate on parameters insterted to dvc.yaml from `vars` soon... Will address this ⌛

content/docs/command-reference/params/index.md
jorgeorpinel: Moved this into the Params values section for now.

content/docs/command-reference/params/diff.md
```diff
@@ -19,35 +19,29 @@ positional arguments:
...
 
 ## Options
 
-- `--targets <paths>` - limit command scope to these params files. When
-  specifying arguments for `--targets` before `revisions`, you should use `--`
-  after this option's arguments, e.g.:
+- `--targets <paths>` - limit or expand the command's scope to these params
+  files.
 
-  ```dvc
```
jorgeorpinel, line 43: > dvc params diff can actually operate on any file now

On that I did the change aboveto the `--targets` option. Is that what you meant? I should probably test this because I'm not sure it makes sense: if the param is was never registered in dvc.lock can it still diff it to previous versions? Will check...

jorgeorpinel: I tested this and found some possible bugs: iterative/dvc/issues/5186 so I reverted the part about "or expand" for now. Will review the text again once I have an answer and better understand `params diff --targets` which as of now doesn't seem to do much...

content/docs/user-guide/dvc-files.md
shcheklein: any updates here?

content/docs/command-reference/params/index.md
shcheklein: so, when we talk about `dvc.yaml` or `dvc.lock`, or how can put it there via `dvc run` it is parameter dependency. When we talk about a key/value in the `params.yaml` it is a parameter, that your code could read for example.

content/docs/command-reference/params/index.md
shcheklein: I guess I don't see any value in those sections, they have a lot of repetition with the previous introductory section. They don't have any motivation behind them (not clear why do they exist, how do they connect to the other text).

> Will address this ⌛

is is addressed already?

content/docs/command-reference/params/diff.md
shcheklein: > if the param is was never registered in dvc.lock can it still diff it vs. previous versions? Will check... ⌛

it exactly how it works. You don't have to register anything with `dvc.yaml` now for params and metrics. You don't even need it to be a DVC project.

content/docs/command-reference/diff.md
```diff
@@ -44,18 +44,13 @@ for example when `dvc init` was used with the `--no-scm` option.
 
 ## Options
 
-- `--targets <paths>` - limit command scope to these paths. When specifying
-  arguments for `--targets` before `a_rev`/`b_rev`, you should use `--` after
```
shcheklein, line 48: why removing this?

content/docs/command-reference/diff.md
```diff
@@ -44,18 +44,13 @@ for example when `dvc init` was used with the `--no-scm` option.
...
-  this option's arguments, e.g.:
+- `--targets <paths>` - limit command scope to these paths.
 
-  ```dvc
```
shcheklein, line 51: same here - why removing this? We are complicating users life this way for the sake of preserving "perfect" and clean structure to my mind.

content/docs/command-reference/params/diff.md
```diff
@@ -19,35 +19,28 @@ positional arguments:
...
 
 Without arguments, this command compares parameters currently present in the
 <abbr>workspace</abbr> (uncommitted changes) with the latest committed version.
-
-Supported parameter _value_ types are: string, integer, float, and arrays. DVC
-itself does not ascribe any specific meaning for these values.
-
-❗ By default it only shows parameters that were changed.
+This includes everything in the default parameters file (`params.yaml`) as well
+as all `params` found in [DVC files](/doc/user-guide/dvc-files). Only params
```
shcheklein, line 32: "as well as all `params` found in" - this is not very clear to me

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,64 @@ positional arguments:
...
-The default parameters file name is `params.yaml`. Parameters should be
-organized as a tree hierarchy inside, as DVC will locate param names by their
-tree path. Parameters files have to be manually written, or generated, and these
-can be versioned directly with Git.
+```yaml
+stages:
+  mystage:
+    cmd: ./myscript.sh
+    params:
+      - foo
```
shcheklein, line 33: let's make some realistic names please instead of `foo`, `bar`

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,64 @@ positional arguments:
...
+      - myparams.toml:
+          - qux
+```
 
-Supported parameter _value_ types are: string, integer, float, and arrays. DVC
-itself does not ascribe any specific meaning for these values. They are
-user-defined, and serve as a way to generalize and parametrize an machine
-learning algorithms or data processing code.
+> By default, parameters are read from `params.yaml`. Other params files can be
+> listed too, with sub-lists of the params found in them (as shown above).
```
shcheklein, line 40: reference file names, param names from the example

content/docs/user-guide/dvc-files.md
jorgeorpinel: I'm not prioritizing this part for the current PR (per https://github.com/iterative/dvc.org/pull/2062#discussion_r549473711), but will continue searching for an actual real life example ASAP.

content/docs/command-reference/params/index.md
jorgeorpinel: Agreed! That's basically the context I was referring to 🙂 (except sometimes for 2.0 `dvc.yaml` its sometimes not dependency but that's another PR).

content/docs/user-guide/dvc-files.md
shcheklein: let's not touch it then make it a separate issue?

content/docs/command-reference/params/index.md
jorgeorpinel: > sections, they have a lot of repetition

Maybe in past iterations? I'm not seeing the repetition right now (maybe just 2 mentions of `repro`?).

jorgeorpinel: > motivation behind them (not clear why do they exist... how do they connect

They are the references for parameters files and for parameter values, which are not detailed in the Description, but they are linked as part of the explanation.

shcheklein: I see a few concerns (overall comes down to complicating and making things less clear):

- `Parameters files section` repeats the note to my mind (expands a bit, but it's hard to justify both to my mind)
- It's hard to justify the whole section for a single sentence content as well. 
- `Parameter values` starting from the `DVC saves ...` has some general content about the params not only values.
- If deal with that ^^ we'll one-paragraph content again.
- Most important - both sections appear w/o any motivation, w/o any reference to them. See the last comment `They don't have any motivation behind them (not clear why do they exist, how do they connect to the other text).` When I read it I don't understand this structure, why does it exist.

content/docs/command-reference/params/diff.md
jorgeorpinel: >  You don't have to register anything with dvc.yaml now for params

Gotcha, so it's purely based of Git commits. OK, updated! PTAL.

jorgeorpinel: > and metrics

Updated that ref. too. What about `plots diff`?

shcheklein: > What about plots diff?

Most likely not, it should be a ticket on GH to my mind to cover that.

jorgeorpinel: I confirmed that I also applies to plots. All updated!

content/docs/command-reference/metrics/diff.md
```diff
@@ -36,19 +36,15 @@ lists all the current metrics, without comparisons.
 
 ## Options
 
-- `--targets <paths>` - limit command scope to these metrics files. Using `-R`,
-  directories to search metrics files in can also be given. When specifying
-  arguments for `--targets` before `revisions`, you should use `--` after this
-  option's arguments, e.g.:
+- `--targets <paths>` - specify the command's scope to these metrics files. It
+  accepts `paths` to any valid metrics file, regardless of whether it's used by
```
shcheklein: I think `not used by DVC` is not very clear, can we rephrase it somehow?

content/docs/command-reference/plots/diff.md
```diff
@@ -6,7 +6,8 @@ overlaying them in a single image. This allows to compare them easily.
 ## Synopsis
 
 ```usage
-usage: dvc plots diff [-h] [-q | -v] [--targets [<path> [<path> ...]]]
+usage: dvc plots diff [-h] [-q | -v]
+                      [--targets [<paths> [<paths> ...]]]
```
jorgeorpinel, line 10: - [ ] Corresponding change required in core (also check in `diff` and `params diff`)

content/docs/command-reference/params/diff.md
shcheklein: where does it take Vega templates if DVC is not initialized?

content/docs/user-guide/dvc-files.md
jorgeorpinel: Oh I thought that by "base example" you meant the multi-cmd one with requirements.txt, which was already merged to master but I accidentally rolled it back at some point. But yes, I can extract this... ⌛

shcheklein: I see,  `but I accidentally rolled it back at some point.` - that's where all this confusion comes from. Up to you then 👍

jorgeorpinel: Already extracted to #2070 🙂 resolving here

content/docs/command-reference/params/index.md
jorgeorpinel: Any repetition I should definitely address (⌛), its kind of a separate issue (it could be there with or without sections).

> hard to justify the whole section for a single sentence
> DVC saves ... has some general content

Good points. OK, I'll removed the headers ⌛ 

> both sections appear w/o any motivation, w/o any reference to them

Not sure what you mean by w/o ref. since they're linked to from the main Desc. but indeed there's no intro paragraph giving them some motivation, that's because these are reference docs so we don't have a story in each one. Sections can be pretty useful in this kinds of docs so I'm not sure why we're trying to avoid them in general (your specific points did convince me this time though). For example what about https://dvc.org/doc/command-reference/metrics#supported-file-formats? We have many of those.

jorgeorpinel: OK I removed the H3s and addressed the other specific feedback.

I also just realized that the **Examples** had some problems so I threw that in. PTAL

content/docs/command-reference/params/index.md
jorgeorpinel: OK done but PTAL and lmk if I over-did it. I can just reuse some of the ones in **Examples**.

content/docs/command-reference/params/index.md
jorgeorpinel: This note is completely gone now as it was repeated later by the params file paragraph.

content/docs/command-reference/params/diff.md
jorgeorpinel: Updated to "as well everything listed in the `params` field of `dvc.yaml`/`dvc.lock`"

p.s. had to expand "DVC files" since we're talking about a specific field, and it felt necessary keep "dvc.lock" since it's not mentioned anywhere else in the doc.

content/docs/command-reference/metrics/diff.md
jorgeorpinel: OK changed to "regardless of whether `dvc.yaml` is currently tracking any params in them". WDYT?

content/docs/command-reference/diff.md
```diff
@@ -44,18 +44,13 @@ for example when `dvc init` was used with the `--no-scm` option.
...
-  ```
-
-  Alternatively, you can also run the above statement as:
+  When specifying arguments for `--targets` before `a_rev`/`b_rev`, you should
+  use `--` after this option's arguments (POSIX terminals), e.g.:
 
   ```dvc
-  $ dvc diff HEAD v1 --targets t1.json t2.json
+  $ dvc diff --targets t1.json t2.yaml -- HEAD v1
   ```
```
jorgeorpinel, line 54: @shcheklein I have a couple emails about this change that I can't find in the PR. Do you still think this change complicates users' life? I was trying to simplify this actually, but I can revert, if you prefer.

jorgeorpinel: p.s. Recently you've asked me to keep other cmd ref. options short so I thought you would approve of this.

content/docs/command-reference/params/diff.md
jorgeorpinel: > if DVC is not initialized

Right, I didn't test without DVC, and I didn't mention that either. But it works for plots files that dvc.yaml isn't tracking — that's what I meant.

jorgeorpinel: > I didn't mention that either in any ref.

Should I? It seems like an edge case/ side-effect to me. Not sure it's worth documenting but maybe. Something like "this command doesn't require an existing DVC project to run in." (from [get](https://dvc.org/doc/command-reference/get)).

jorgeorpinel: > should be a ticket on GH to my mind to cover that

BTW yes there's iterative/dvc/issues/4446 so we can also decide on what to doc as that issue gets' implemented and docs PRs are created for those changes.

shcheklein: > It seems like an edge case/ side-effect to me. 

it's not side effect - it's intentionally made that you don't need DVC project if you want capturing metrics, params, plots

content/docs/command-reference/diff.md
shcheklein: I think I deleted them :) so we are good, I was confused by reading diff

content/docs/command-reference/metrics/diff.md
shcheklein: it's not even about `dvc.yaml` at this point, even DVC repo might not be initialized. I see that the whole change for that ticket might be out of scope since it's a much bigger change

content/docs/command-reference/diff.md
shcheklein: sorry for the confusion, I think it was fine as it was before, previous iteration. Up to you to revert.

content/docs/command-reference/params/diff.md
```diff
@@ -19,35 +19,30 @@ positional arguments:
...
 
 Without arguments, this command compares parameters currently present in the
 <abbr>workspace</abbr> (uncommitted changes) with the latest committed version.
-
-Supported parameter _value_ types are: string, integer, float, and arrays. DVC
-itself does not ascribe any specific meaning for these values.
-
-❗ By default it only shows parameters that were changed.
+This includes everything in `params.yaml` (default parameters file) as well
+everything listed in the `params` field of `dvc.yaml`/`dvc.lock`. Only params
```
shcheklein, line 32: still don't understand what does `as well everything listed in the `params` field of `dvc.yaml`/`dvc.lock`` mean

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,56 @@ positional arguments:
...
 
-Supported parameter _value_ types are: string, integer, float, and arrays. DVC
-itself does not ascribe any specific meaning for these values. They are
-user-defined, and serve as a way to generalize and parametrize an machine
-learning algorithms or data processing code.
+```yaml
+stages:
+  preprocess:
+    cmd: ./prepare.py
+    params:
```
shcheklein, line 32: we use `epochs, learning-rate, batch_siz` above, let's do the same in the example?

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,56 @@ positional arguments:
...
+      - html_tags
+      - cleanup.stemming
+      - myparams.toml:
+          - nlp.lemmatize
+```
 
-DVC saves the param names and their latest values in the `dvc.yaml` file. These
-values will be compared to the ones in the params files to determine if the
-stage is invalidated upon pipeline [reproduction](/doc/command-reference/repro).
+In contrast to a regular <abbr>dependency</abbr>, a parameter is not a file or
```
shcheklein, line 39: a parameter dependency ...

content/docs/command-reference/params/index.md
```diff
@@ -18,44 +18,56 @@ positional arguments:
...
+be generated) and they can be versioned directly with Git.
+
+**Parameter values** should be organized in tree-like hierarchies (dictionaries)
+inside param files (see [Examples](#examples)). DVC will interpret param names
+as the tree path to find those values. Supported types are: string, integer,
+float, and arrays (groups of params). Note that DVC does not ascribe any
+specific meaning to these values.
+
+DVC saves parameter names and values in the project's
+[DVC files](/doc/user-guide/dvc-files) in order to track them over time. They
```
shcheklein, line 61: in this case we can be specific? `dvc.lock`?

shcheklein (approved):
Some minor things left. I suggest to address the whole story related to non DVC repos for metrics/plots/etc separately. It's a bigger problem that we can handle for Misc updates.

content/docs/command-reference/params/diff.md
jorgeorpinel: Okdk. Added a note to metrics diff and params diff about this (no DVC project required when using --targets, only a Git repo).

content/docs/command-reference/metrics/diff.md
jorgeorpinel: You're right but yes, since iterative/dvc/issues/4446 is still open and the core team hasn't really sent many docs updates per their work on that... We could leave for later.

jorgeorpinel: Then again since I just added that note about not needing a DVC project... This text in --targets is kind of contradicting again... I'll simplify it a little for now (less specific).

jorgeorpinel: Actually I'm not sure how to generalize since we didn't want to say "not used by DVC". Sry

jorgeorpinel: p.s. the current text is not incorrect so... I think it should do for now 🙂

content/docs/command-reference/diff.md
jorgeorpinel: personally prefer the shorter version. The 2nd example is overkill for an option bullet and it's implied already.

content/docs/command-reference/params/diff.md
jorgeorpinel: `params diff` withuot arguments shows everything changed in params.yaml as well as all params defined in dvc.yaml. And soon also even params used to templetize dvc.yaml (but that's not in scope, more for #2052).

I simplified to "...as well all the `params` used in `dvc.yaml`." and separated the mention of dvc.lock into a following sentence.

content/docs/command-reference/params/index.md
jorgeorpinel: Yes, that is way better 😅. Done!

content/docs/command-reference/params/index.md
jorgeorpinel: Updated.

content/docs/command-reference/params/index.md
jorgeorpinel: Agree. Specified.

jorgeorpinel:
> address the whole story related to non DVC repos for metrics/plots/etc separately

Agree. I'm hoping to do so as docs changes from core PRs related to iterative/dvc/issues/4446 continue to arrive.

Merging!

PR #2067, opened 2020-12-31
Dec gems patch

elleobrien:
patch to fix bad comment link


shcheklein:
@elleobrien doesn't look like a simple fix - it has all the changes from the previous PR?

elleobrien:
@shcheklein you're right i made branch off branch, not `main`. will close and do again

PR #2073, opened 2021-01-05
feat: sidebar dropdown toggle with subpages #1427

### LINKED ISSUE #1427: sidebar dropdown toggle
It would be good if the user is able to toggle the dropdowns containing sub-pages.
For example, 
Screenshot from 2020-06-12 09-07-17
Let's say I need to go back to the `Tutorials` section to view another page.
I tried clicking back on the `Command Reference` again to collapse it rather than scrolling all the way up in the sidebar.
But it just displays the home page of `Command Reference`. Maybe if clicking on the arrow next to it could collapse it and then display `Command Reference`'s page, the sidebar would be more intuitive.

It's not a really big issue I guess, and not many users might navigate the way I do perhaps, just thought I should let you guys know, though.
That being said, the sidebar and docs page is definitely already quite intuitive. Just a feature request :v:

ashiishme:
I've fixed the drop down issue with sub-pages in documentation (closes #1427).

This changes improves the user experience when user tries to navigate quickly. Currently, the user has to scroll all the way down. 

I have used React state to handle the toggle based on the `activePaths`. The demo is attached below.

dvc-org-drop-down-toggle-demo

### DIFF
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -51,10 +51,14 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   type
 }) => {
   const isActive = activePaths && includes(activePaths, path)
+  const [dropDownToggle, setDropDownToggle] = useState(isActive)
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
   const isLeafItem = children === undefined || children.length === 0
-  const currentLevelOnClick = (): void => onClick(isLeafItem)
+  const currentLevelOnClick = (): void => {
+    setDropDownToggle(!dropDownToggle)
+    onClick(isLeafItem)
+  }
 
   // Fetch a special icon if one is defined
   const IconComponent = icon && ICONS[icon]
@@ -64,7 +68,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
 
   const className = cn(
     styles.sectionLink,
-    isActive && styles.active,
+    dropDownToggle && styles.active,
     isRootParent && 'docSearch-lvl0',
     'link-with-focus',
     style ? styles[style] : styles.sidebarDefault,
@@ -101,8 +105,8 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
     <>
       {parentElement}
       {children && (
-        <span hidden={!isActive}>
-          <Collapse isOpened={!!isActive}>
+        <span hidden={!dropDownToggle}>
+          <Collapse isOpened={!!dropDownToggle}>
             {children.map(item => (
               <SidebarMenuItem
                 key={item.path}
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -50,8 +50,6 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   icon,
   type
 }) => {
-  const isActive = activePaths && includes(activePaths, path)
-
   const [isExpanded, setIsExpanded] = useState(
     activePaths && includes(activePaths, path)
   )
@@ -68,7 +66,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   const currentLevelOnClick = (
     event: SyntheticEvent<HTMLAnchorElement>
   ): void => {
-    if (isActive) {
+    if (event.currentTarget.getAttribute('aria-current') === 'page') {
       event.preventDefault()
       setIsExpanded(!isExpanded)
     }
@@ -112,7 +110,11 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
         onClick={currentLevelOnClick}
         target="_blank"
       >
-        {iconElement}
+        {iconElement ? (
+          iconElement
+        ) : (
+          <span className={bulletIconClassName}></span>
+        )}
         {label} <ExternalLinkIcon />
       </Link>
     ) : (
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -50,12 +50,14 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   icon,
   type
 }) => {
-  const [isActive, setIsActive] = useState(
+  const isActive = activePaths && includes(activePaths, path)
+
+  const [isExpanded, setIsExpanded] = useState(
     activePaths && includes(activePaths, path)
   )
 
   useEffect(() => {
-    setIsActive(activePaths && includes(activePaths, path))
+    setIsExpanded(activePaths && includes(activePaths, path))
   }, [activePaths])
 
   const isRootParent =
@@ -64,16 +66,16 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   const currentLevelOnClick = (
     event: SyntheticEvent<HTMLAnchorElement>
   ): void => {
-    if (event.currentTarget.getAttribute('aria-current') === 'page') {
+    if (isActive) {
       event.preventDefault()
-      setIsActive(!isActive)
+      setIsExpanded(!isExpanded)
     }
     onClick(isLeafItem)
   }
 
   const bulletIconClick = (event: SyntheticEvent<HTMLSpanElement>): void => {
     event.preventDefault()
-    setIsActive(!isActive)
+    setIsExpanded(!isExpanded)
   }
 
   // Fetch a special icon if one is defined
@@ -84,7 +86,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
 
   const className = cn(
     styles.sectionLink,
-    isActive && styles.active,
+    isExpanded && styles.active,
     isRootParent && 'docSearch-lvl0',
     'link-with-focus',
     style ? styles[style] : styles.sidebarDefault,
@@ -129,8 +131,8 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
     <>
       {parentElement}
       {children && (
-        <span hidden={!isActive}>
-          <Collapse isOpened={!!isActive}>
+        <span hidden={!isExpanded}>
+          <Collapse isOpened={!!isExpanded}>
             {children.map(item => (
               <SidebarMenuItem
                 key={item.path}
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -71,6 +71,11 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
     onClick(isLeafItem)
   }
 
+  const bulletIconClick = (event: SyntheticEvent<HTMLSpanElement>): void => {
+    event.preventDefault()
+    setIsActive(false)
+  }
+
   // Fetch a special icon if one is defined
   const IconComponent = icon && ICONS[icon]
   const iconElement = IconComponent ? (
@@ -107,6 +112,13 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
         className={className}
         onClick={currentLevelOnClick}
       >
+        <span
+          className={styles.absBeforeIconOverlay}
+          onClick={bulletIconClick}
+          onKeyDown={bulletIconClick}
+          role="button"
+          tabIndex={-1}
+        ></span>
         {iconElement}
         {label}
       </Link>
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -62,7 +62,9 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
 
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
+
   const isLeafItem = children === undefined || children.length === 0
+
   const currentLevelOnClick = (
     event: SyntheticEvent<HTMLAnchorElement>
   ): void => {
@@ -95,6 +97,12 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
     icon ? undefined : styles.withDefaultBullet
   )
 
+  const bulletIconClassName = cn(
+    styles.sidebarDefaultBullet,
+    isExpanded && styles.active,
+    isLeafItem && styles.sidebarLeafBullet
+  )
+
   const parentElement =
     type === 'external' ? (
       <Link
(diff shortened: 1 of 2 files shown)


shcheklein:
@ashiishme I've deployed it here https://dvc-landing-fix-dropdow-hdyck8.herokuapp.com/doc/install.

Something funny is happening with the active state:

Screen Shot 2021-01-05 at 6 22 11 PM

that what I've managed to do after clicking around

shcheklein:
Or like this:

Screen Shot 2021-01-05 at 6 23 45 PM

shcheklein:
Also, a few other concerns:

- Looks like now sections stay in the expanded mode when I jump between them
- May be it's fine, but we need to have a click on the arrow support I think, that doesn't take you to the page, but only collapeses/expands the section. We should make them active on hover (bigger?)

shcheklein (changes requested):
Good attempt! Thanks for the PR, but there are some bugs and some other changes that are needed here.

ashiishme:
> Or like this:
> 
> Screen Shot 2021-01-05 at 6 23 45 PM

😄 Didn't notice the active style, I should have tested in more detail.

ashiishme:
> Good attempt! Thanks for the PR, but there are some bugs and some other changes that are needed here.

Thank you, I will fix the bug and update as you have requested.

jorgeorpinel:
> Looks like now sections stay in the expanded mode when I jump between them

I think we should collapse the previously expanded section if you click on another one for now. Just having that ability is already an improvement and would also take care of the bug where you click on an open section head, it collapses it, but takes you to that page.

ashiishme:
@jorgeorpinel  @shcheklein 

I am able to fix the bug but need your suggestion as well.

Now, I am using the arrow to only collapse the menu, it will not expand it as well as not take user to any page. The expand functionality is as same as it now in production. This fixes the both active style & page jump issue.

But, I am not sure if it is okay when the user is inside the nested page, and rather collapse the parent item, the style will be remove but user remains at the same page. This is the expected behavior if we need to collapse but let me know about your thoughts. You can see the example in video.


https://user-images.githubusercontent.com/18111862/104105723-30cbac80-52d8-11eb-853d-02b131f37834.mp4

ashiishme:
I have increased the size of arrow button by 2px on active. I am using `span` tag so for accessibility purpose, the outline doesn't look that good and the icon too when it is large. I am not sure what to do with it?

shcheklein:
> the outline doesn't look that good

yep, look pretty ugly :( - better do some glow effect? 

Rearding the dynamics - it's better now. A few improvements that come to my mind:

- when you click the expanded section main title we can collapse it
- when you click arrow it should serve as collapse/expand - I'm not sure how to combine it with auto-collapse. For now we can collapse everything I think (except the page you open)

jorgeorpinel:
Agree with Ivan, no need for special arrow behavior (it's kind of unintuitive).

jorgeorpinel:
but thanks for addressing the previous issues!

ashiishme:
> * when you click the expanded section main title we can collapse it

- Same as before right when we click the title adding the toggle effect?

> * when you click arrow it should serve as collapse/expand - I'm not sure how to combine it with auto-collapse. For now we can collapse everything I think (except the page you open)

- It will allow us to expand & collapse from both title or the arrow icon isn't it?

ashiishme:
@shcheklein  I think we might not need arrow toggle as mentioned by @jorgeorpinel  because with the menu title toggle, it feels completely unnecessary.

Now, we can expand / collapse menu item also the previous bugs are addressed.

This is how it works now:

If we are inside the child pages:
- Parent
- Child ( active )
- Child

The user can navigate back to Parent page without collapsing.

If we are in parent page:
- Parent ( active )
- Child
- Child

The user can collapse by clicking the parent menu title.

Please check the video:

https://user-images.githubusercontent.com/18111862/104121308-391df900-5365-11eb-9f1d-911c9fdec6d9.mp4

shcheklein:
@ashiishme it looks smooth and great!

I still think that click to arrows to expand/collapse w/o activating the page itself makes sense. E.g. I'm on the https://dvc-landing-fix-dropdow-hdyck8.herokuapp.com/doc/user-guide/dvc-internals and I click on the User Guide to collapse it to see other side bar items - no way I can do it now, w/o switching to the page. And I have to click twice actually.

ashiishme:
> @ashiishme it looks smooth and great!
> 
> I still think that click to arrows to expand/collapse w/o activating the page itself makes sense. E.g. I'm on the https://dvc-landing-fix-dropdow-hdyck8.herokuapp.com/doc/user-guide/dvc-internals and I click on the User Guide to collapse it to see other side bar items - no way I can do it now, w/o switching to the page. And I have to click twice actually.

Yeah, now it make sense to me. I will enable the arrow toggle.

ashiishme:
@shcheklein  @jorgeorpinel I have updated with arrow toggle now.

shcheklein:
@ashiishme great! we are getting there! 

I see two strange issues right now:

1. Open https://dvc-landing-fix-dropdow-hdyck8.herokuapp.com/doc/api-reference/open. Click arrow to collapse. Arrow doesn't work to expand anymore
2. Arrows do not work to expand on other (inactive) sections

ashiishme:
> @ashiishme great! we are getting there!
> 
> I see two strange issues right now:
> 
> 1. Open https://dvc-landing-fix-dropdow-hdyck8.herokuapp.com/doc/api-reference/open. Click arrow to collapse. Arrow doesn't work to expand anymore
> 2. Arrows do not work to expand on other (inactive) sections

Looks like the state issue when directly visiting the specific page. Will check this :smile:

ashiishme:
@shcheklein It seems not possible or maybe I am not able to figure it out to auto collapse with arrow because of the re-render when the path matches the `activePaths` lists. So, what I did is, when I am on page `A` and I want to use arrow for collapse / expand, the first click on arrow will work same as the page navigation when clicking on the title which means, the user will be taken to that page, but after that, the arrow will only work to collapse / expand and this seems to work as expected without creating any further issues.

Can you please review now.

shcheklein:
> with arrow because of the re-render when the path matches the activePaths lists

but expanding/collapsing on the arrow should not be changing activePath, right?

> which means, the user will be taken to that page

it breaks to some extent the "exploration" flow - I'd like to see the pages, may be I don't even want to go the next page, just see the full list.

ashiishme:
> but expanding/collapsing on the arrow should not be changing activePath, right?

Yes, it shouldn't as per our discussion and it does work but for the first click, it is necessary to re-render to collapse others. If I am at Page `A` which is expanded & I want to see the content of Page `B`. When I click the arrow, it will take me to the page and after that the arrow will just expand & collapse only. The way the sidebar menu is configured, it requires re-render to work well with Collapse component.

> it breaks to some extent the "exploration" flow - I'd like to see the pages, may be I don't even want to go the next page, just see the full list.

It does break the exploration flow. I think we have two option now, the one that I have applied recently or the one where user can collapse / expand without visiting the page ( keeps the exploration flow without any issue ) but it will not collapse other when using arrow button. Only toggles when user interacts ( no auto collapse ).

What do you say?

shcheklein:
@ashiishme 
> but it will not collapse other when using arrow button. Only toggles when user interacts ( no auto collapse ).

could you clarify, not sure I understand exactly.

> it is necessary to re-render to collapse others.

I'm not sure I understand why do we have to re-render anything to collapse/expand anything.

ashiishme:
> could you clarify, not sure I understand exactly.

When I expand `dropdown 1` and use arrow button to expand another dropdown, the `dropdown 1` will be in expand state together with another dropdown. It will only collapse / expand when the same arrow button is toggle by user. Currenlty with the pages title click, when I am at `page 1` which is expanded, and I click on `page 2` title, the `page 1` auto collapse and the `page 2` is expand. Let me know if you understand it now, or I will attach a demo today evening if you like.

> I'm not sure I understand why do we have to re-render anything to collapse/expand anything.

The child components wrapped by `Collapse` component renders based on the `activePath` isn't it? So, to collapse or expand we will need to render again into view or remove it again from the view.

shcheklein:
@ashiishme 

> The child components wrapped by Collapse component renders based on the activePath isn't it?

it might be, I don't know. From the product perspective I don't see why would re-render is needed. If changes to the way components are done needed - so be it, right? Or is there some fundamental reason? 

What I'm trying here is to clarify/specify the best behavior from the end user perspective first. Then think how can it be implemented. 

> the one where user can collapse / expand without visiting the page ( keeps the exploration flow without any issue ) but it will not collapse other when using arrow button. Only toggles when user interacts ( no auto collapse ).
> Let me know if you understand it now, or I will attach a demo today evening if you like.

Still not sure exactly. Demo would be helpful :)

shcheklein:
@rogermparent it looks good to from the product perspective, could you review the PR from the JS/TS perspective, please?

shcheklein:
A few things I noticed playing around with this:

- sometimes mouse pointer goes a bit below the arrow symbol and it trigger the page to be activated
- i think since arrows now can be clicked we should highlight them on hover - so that users understand that they can be clicked

ashiishme:
> * sometimes mouse pointer goes a bit below the arrow symbol and it trigger the page to be activated

The click listener for arrow depends on the `span` inside `anchor tag` so, outside the arrow will trigger the click of `anchor tag`. This also creates the accessibility issue because a click listener inside `anchor tag`. I will try to find a better way to cover both issues after reviewed by **Roger**.

> * i think since arrows now can be clicked we should highlight them on hover - so that users understand that they can be clicked

Sure, will do this.

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useEffect, useRef, useState, SyntheticEvent } from 'react'
```
rogermparent, line 1: This is a useful Type that I haven't used yet!

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -50,11 +50,31 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
   icon,
   type
 }) => {
-  const isActive = activePaths && includes(activePaths, path)
+  const [isActive, setIsActive] = useState(
+    activePaths && includes(activePaths, path)
+  )
```
rogermparent, line 55: My main contention with this is that it steals the former `isActive`, which is useful as a value distinct from this feature. I would make this `useState` separate and called `isExpanded`/`setIsExpanded`.

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -50,11 +50,31 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
...
+  }, [activePaths])
+
   const isRootParent =
     activePaths && activePaths.length > 1 && activePaths[0] === path
   const isLeafItem = children === undefined || children.length === 0
-  const currentLevelOnClick = (): void => onClick(isLeafItem)
+  const currentLevelOnClick = (
+    event: SyntheticEvent<HTMLAnchorElement>
+  ): void => {
+    if (event.currentTarget.getAttribute('aria-current') === 'page') {
```
rogermparent, line 67: A separate `isActive` with the old behavior could be useful here.

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -92,6 +112,14 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
...
         onClick={currentLevelOnClick}
       >
+        <span
+          className={styles.absBeforeIconOverlay}
+          onClick={bulletIconClick}
+          onKeyDown={bulletIconClick}
+          data-path={path}
+          role="button"
+          tabIndex={0}
+        ></span>
```
rogermparent, line 122: I think it would be preferable not to add an overlay element when we have full control over the arrow already.  
Maybe we could apply the styles currently on the `::before` to this element?

rogermparent (changes requested):
I can't deny the functionality, but I think the implementation would benefit from a refactor that integrates into the current code better.

src/components/Documentation/Layout/SidebarMenu/index.tsx
ashiishme: @rogermparent  What about the former isActive variable? It will not be used anywhere so should I just rename the variable names?

ashiishme: Okay, it has been useful now based on the below comment, thank you. :D

src/components/Documentation/Layout/SidebarMenu/index.tsx
ashiishme: Awesome, didn't thought about that, thank you.

src/components/Documentation/Layout/SidebarMenu/index.tsx
ashiishme: @rogermparent Can you suggest any work around to detect the click for the arrow?

shcheklein:
@ashiishme hey, sorry for the delay here! What is the status now for this PR?

shcheklein:
Closing this as stale. @ashiishme please, feel free to get back to it and finalize when/if you have time. Thanks 🙏

ashiishme:
@shcheklein Hi Ivan, I have made some changes based on Roger's review, can you please reopen this PR?

rogermparent:
Just on first view I can see you totally addressed the feedback! I'll pull this down to make sure it runs well as well.

shcheklein:
It got deployed, I still see a few issue:

- when section that has an index page (e.g. User Guide) is opened, one of the pages inside I can't open the index page anymore. Open https://dvc-org-fix-dropdown-to-qdss4m.herokuapp.com/doc/user-guide/dvc-internals -> click User Guide
- chanegelog icon (dot) disappeared in the main menu
- we still need to add now some effect on hover on the arrows so that people understand there is some functionality to them

rogermparent:
Some small nitpicks alongside what @shcheklein already pointed out:

Both arrow animations and slide-out don't work on the way back.
I think this one can be fixed by simply tweaking the CSS animation `transition` rules.

The "slide-out" animation stops working both ways after the first expansion.
This issue looks a little different from the former, as it seems to involve the state

rogermparent (changes requested):
The animations issue should probably be solved, but it's small enough to possibly warrant another PR instead of blocking this one

ashiishme:
> It got deployed, I still see a few issue:
> 
> * when section that has an index page (e.g. User Guide) is opened, one of the pages inside I can't open the index page anymore. Open https://dvc-org-fix-dropdown-to-qdss4m.herokuapp.com/doc/user-guide/dvc-internals -> click User Guide
> * chanegelog icon (dot) disappeared in the main menu
> * we still need to add now some effect on hover on the arrows so that people understand there is some functionality to them

@rogermparent  I have checked why the the index page doesn't open after visiting it's child ( point 1 of Ivan's review), I found out that after replacing the current page condition with `isActive` it doesn't seems to work. I am not sure why because we need the condition to be true and `isActive` does returns true but the menu is being collapsed. So, I have reverted it with the old condition the solves the point 1.

ashiishme:
@shcheklein  I have addressed all the issues mentioned by you. I have added scale to the icon when hover. What do you say?

Also, should I fix the animation in this PR or address it on new issue?

shcheklein:
@ashiishme no, not a blocker for sure, thank you! LGTM, I will let @rogermparent do the final review and merge.

shcheklein (approved):

rogermparent (approved):
Great refactor, looks good to me!
I agree that animations can be saved for the future, but it's worth making a lower-priority Issue for them now while we're focused on this feature.

PR #2074, opened 2021-01-06
dvc: still pending 1.x updates (1)

jorgeorpinel:
Per #2026

- [x] Review all .dvc file examples throughout docs (YAML code blocks)
- [x] ^ Also push/pull examples
- [x] Review DVC cache definitions (found in `dvc cache`, DVC Cache tooltip, Internal Files guide, `dvc config` cache section, and Large Dataset Optimization guide, some in `add` and possibly other refs...) - rel. #550.
- [ ] plots refs (closes #1414)
- [ ] other refs? (closes #1824)

---

- [ ] Port this to `master` (2.0) after merge.

Issue #2026, opened 2020-12-16
2.0 docs


jorgeorpinel:
> See release plan in iterative/dvc/issues/5367.

Finalize pending 1.x-related updates (extracted from #1255) - #2074

- [x] Review all .dvc file examples throughout docs (in YAML code blocks) — note we require YAML 1.2 (se iterative/dvc/pull/4415 and #1764) - mostly done #1608. 
- [x] ^ Also push/pull examples
- [x] Review DVC cache info everywhere, especially in the internal files/dirs guide.
- [x] `plots` refs per #1414
- [x] other refs? per #1824

> Would be nice to backport those to `v1` when done, if possible
---

- [x] "Parametrized" pipelines (dvc.yaml changes) - @jorgeorpinel + @skshetry (https://github.com/iterative/dvc/wiki/Parametrization) - #2052
- [x] New/changed Get Started - @dberenbaum + @jorgeorpinel 
  - New example project and story plan (including new pipelines and experiments)
  - Rewrite https://dvc.org/doc/start
- [x] "DVC Experiments" (sample project: https://github.com/dberenbaum/dvc-exp-mnist) - @jorgeorpinel + @pmrowla 
  - Doc run-cache? (#1289) — #2137
  - Prepare docs structure conceptually for regular (persistent) and new (transient) experiments.
  - Experiment Management guide (#816) — #2146
  - Document `dvc exp` refs — ~~#1926~~ #2199, others
  - Ideas: https://replicate.ai/docs/tutorial#experiments-and-checkpoints
- [x] Live metrics - @pared +? — #2227
  - Write README in that repo?
  - ~~Add refs to https://dvc.org/doc/api-reference~~
- `dvc stage` (https://github.com/iterative/dvc.org/issues/2134#issuecomment-771214046) — #2208

---

- [ ] Archived v1 docs - #2006 / #1167 + https://github.com/iterative/dvc/pull/5315
- [x] Update [Katacoda](https://github.com/iterative/dvc.org/issues?q=is%3Aissue+is%3Aopen+katacoda) scenarios - iterative/katacoda-scenarios/issues/20
- [x] `config/remote {} --repo` (#2064) - @efiop + @jorgeorpinel

pared:
@jorgeorpinel we can skip API references for Dvclive, after iterative/dvc#5466 this method will be removed.

jorgeorpinel:
Sounds good, thanks.

jorgeorpinel:
[...]

PR #2075, opened 2021-01-06
dvc: still pending 1.x updates (2)

### LINKED ISSUE #1414: cmd: pending plots changes by 1.0 release

### LINKED ISSUE #1521: list: document usage for data export/archive
For projects created with `dvc init --no-scm`, since there's no Git repo to version all the files NOT tracked by DVC (code, DVC-files), it could be useful to have a `dvc export <external-location>` command to easily create a lightweight copy of the project (for backup). It's "lightweight" because it wouldn't include any of the data tracked by DVC.

Similar to [`git archive`](https://git-scm.com/docs/git-archive). Could even include an `--archive` flag to make a tar/zip bundle of the export.

> In the future, similar bundling/compressing functionality for actual data sets could be reused.

Just a random idea! (It came from reading some conversations about non-Git projects on Discord.)

UPDATE: To see the latest discussion go to https://github.com/iterative/dvc.org/issues/1521#issuecomment-558726115, but in summary:

Don't need a new command for now, just document this to archive a snapshot:

```bash
git archive -o code.zip HEAD
dvc list . -R --dvc-only | zip -@ data.zip  # if `zip` available
dvc list . -R --dvc-only | xargs python -m zipfile -c data.zip  # alternative for windows (assuming `xargs` available)
```

### LINKED ISSUE #1641: doc: mention that multiple dvc.yaml files are supported
> Extracted from https://github.com/iterative/dvc.org/pull/1615#discussion_r463373449

Note needed probably in
- https://dvc.org/doc/user-guide/dvc-files-and-directories,
- https://dvc.org/doc/command-reference/run, and
- maybe https://dvc.org/doc/command-reference/repro

Anywhere else?

### LINKED ISSUE #1700: docs: simplify `run` examples, leaving bare min. usage
Extracted from https://peckofpickledpojos.wordpress.com/2020/08/09/learning-dvc-by-trial-and-error:

> the way stages are run assumes that the script src/prepare.py takes in command line arguments [incorrect but all of our examples look this way]... took me a few seconds before I could parse out what was happening

For example the first `run` example in https://dvc.org/doc/start/data-pipelines could just use `python src/prepare.py` as command and avoid repeating the dependency to avoid confusion. Probably other simplifications can be done to show the gist of `run` there and in other very-first examples of `run` users might find.

### LINKED ISSUE #1824: cmd refs: Updates per 1.x

jorgeorpinel:
> Continues #2074

Per #2026

- [x] Review DVC cache definitions (found in `dvc cache`, DVC Cache tooltip, Internal Files guide, `dvc config` cache section, and Large Dataset Optimization guide, some in `add` and possibly other refs...) - rel. #550.
- [x] plots refs (closes #1414) — only https://github.com/iterative/dvc.org/issues/1414#issuecomment-731027131 is done.
- [x] other refs? (closes #1824, closes #1521, closes #1700)
- [x] mention that there can be more than 1 dvc.yaml file (closes #1641)

---

- [x] Port this branch to `master` (2.0) after merge. See #2077

### DIFF
--- a/content/docs/command-reference/run.md
+++ b/content/docs/command-reference/run.md
@@ -341,7 +341,7 @@ $ cd more_stages/
 $ dvc run -n process_data \
           -d data.in \
           -o result.out \
-          ./my_script.sh --in data.in --out result.out
+          ./my_script.sh data.in result.out
 $ tree ..
 .
 ├── dvc.yaml
@@ -379,7 +379,7 @@ Execute an R script that parses the XML file:
 $ dvc run -n parse \
           -d parsingxml.R -d data/Posts.xml \
           -o data/Posts.csv \
-          Rscript parsingxml.R --in data/Posts.xml --out data/Posts.csv
+          Rscript parsingxml.R data/Posts.xml data/Posts.csv
 ```
 
 To visualize how these stages are connected into a pipeline (given their outputs
--- a/content/docs/command-reference/run.md
+++ b/content/docs/command-reference/run.md
@@ -73,7 +73,7 @@ so on (see `dvc dag`). This graph can be restored by DVC later to modify or
 
 ```dvc
 $ dvc run -n printer -d write.sh -o pages ./write.sh
-$ dvc run -n scanner -d read.sh -d pages -o signed.pdf ./read.sh
+$ dvc run -n scanner -d read.sh -d pages -o signed.pdf ./read.sh pages
 ```
 
 Stage dependencies can be any file or directory, either untracked, or more
@@ -151,7 +151,7 @@ variables in it that should be evaluated dynamically. Examples:
 
 ```dvc
 $ dvc run -n my_stage "./my_script.sh > /dev/null 2>&1"
-$ dvc run -n my_stage './my_script.sh $MYENVVAR'
+$ dvc run -n my_stage -f './my_script.sh $MYENVVAR'
 ```
 
 ## Options
@@ -317,17 +317,17 @@ dataset (`20180226` is a seed value):
 
 ```dvc
 $ dvc run -n train \
-          -d matrix-train.p -d train_model.py \
-          -o model.p \
-          python train_model.py matrix-train.p 20180226 model.p
+          -d train_model.py -d matrix-train.p -o model.p \
+          python train_model.py 20180226 model.p
 ```
 
 To update a stage that is already defined, the `-f` (`--force`) option is
 needed. Let's update the seed for the `train` stage:
 
 ```dvc
-$ dvc run -n train -f -d matrix-train.p -d train_model.py -o model.p \
-          python train_model.py matrix-train.p 18494003 model.p
+$ dvc run -n train --force \
+          -d train_model.p -d matrix-train.p -o model.p \
+          python train_model.py 18494003 model.p
 ```
 
 ## Example: Separate stages in a subdirectory
@@ -341,7 +341,7 @@ $ cd more_stages/
 $ dvc run -n process_data \
           -d data.in \
           -o result.out \
-          ./my_script.sh data.in result.out
+          ./my_script.sh --in data.in --out result.out
 $ tree ..
 .
 ├── dvc.yaml
@@ -379,7 +379,7 @@ Execute an R script that parses the XML file:
 $ dvc run -n parse \
           -d parsingxml.R -d data/Posts.xml \
           -o data/Posts.csv \
-          Rscript parsingxml.R data/Posts.xml data/Posts.csv
+          Rscript parsingxml.R --in data/Posts.xml --out data/Posts.csv
 ```
 
 To visualize how these stages are connected into a pipeline (given their outputs
@@ -421,9 +421,9 @@ Define a stage with both regular dependencies as well as parameter dependencies:
 
 ```dvc
 $ dvc run -n train \
-          -d matrix-train.p -d train_model.py -o model.p \
+          -d train_model.py -d matrix-train.p  -o model.p \
           -p seed,train.lr,train.epochs
-          python train_model.py matrix-train.p model.p
+          python train_model.py 20200105 model.p
 ```
 
 `train_model.py` will include some code to open and parse the parameters:
--- a/content/docs/command-reference/run.md
+++ b/content/docs/command-reference/run.md
@@ -150,8 +150,8 @@ like `|` (pipe) or `<`, `>` (redirection), otherwise they would apply to
 variables in it that should be evaluated dynamically. Examples:
 
 ```dvc
-$ dvc run -n my_stage "./my_script.sh > /dev/null 2>&1"
-$ dvc run -n my_stage -f './my_script.sh $MYENVVAR'
+$ dvc run -n first_stage "./a_script.sh > /dev/null 2>&1"
+$ dvc run -n second_stage './another_script.sh $MYENVVAR'
 ```
 
 ## Options
--- a/content/docs/user-guide/basic-concepts/dvc-cache.md
+++ b/content/docs/user-guide/basic-concepts/dvc-cache.md
@@ -4,6 +4,6 @@ match: ['DVC cache', cache, caches, cached, 'cache directory']
 ---
 
 The DVC cache is a hidden storage (by default in `.dvc/cache`) for files and
-directories tracked by DVC, and their different versions. Data is cached into a
-special flattened
-[structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
+directories tracked by DVC, and their different versions. Learn more about it's
+structure
+[here](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
--- a/content/docs/user-guide/basic-concepts/dvc-cache.md
+++ b/content/docs/user-guide/basic-concepts/dvc-cache.md
@@ -3,7 +3,7 @@ name: 'DVC Cache'
 match: ['DVC cache', cache, caches, cached, 'cache directory']
 ---
 
-The DVC cache is a hidden storage (by default located in the `.dvc/cache`
-directory) for files that are tracked by DVC, and their different versions.
-Learn more about it's
+The DVC cache is a hidden storage (by default in `.dvc/cache`) for files and
+directories tracked by DVC, and their different versions. Data is cached into a
+special flattened
 [structure](/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).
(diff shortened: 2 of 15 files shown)


content/docs/command-reference/run.md
```diff
@@ -341,7 +341,7 @@ $ cd more_stages/
 $ dvc run -n process_data \
           -d data.in \
           -o result.out \
-          ./my_script.sh data.in result.out
+          ./my_script.sh --in data.in --out result.out
```
shcheklein, line 344: why do we need `--in`, `--out`? Overall looks like a lot of very cosmetic changes to the (valid) commands in this PR

content/docs/command-reference/run.md
```diff
@@ -421,9 +421,9 @@ Define a stage with both regular dependencies as well as parameter dependencies:
 
 ```dvc
 $ dvc run -n train \
-          -d matrix-train.p -d train_model.py -o model.p \
+          -d train_model.py -d matrix-train.p  -o model.p \
           -p seed,train.lr,train.epochs
-          python train_model.py matrix-train.p model.p
+          python train_model.py 20200105 model.p
```
shcheklein, line 426: where does `20200105` come from?

content/docs/command-reference/run.md
```diff
@@ -151,7 +151,7 @@ variables in it that should be evaluated dynamically. Examples:
 
 ```dvc
 $ dvc run -n my_stage "./my_script.sh > /dev/null 2>&1"
-$ dvc run -n my_stage './my_script.sh $MYENVVAR'
+$ dvc run -n my_stage -f './my_script.sh $MYENVVAR'
```
shcheklein, line 154: let's use `--force` (other commands have `-f` for `file`, it's better to be explicit)

shcheklein: also, I thinks it's better to use a different name for the stage in this case

content/docs/user-guide/basic-concepts/dvc-cache.md
```diff
@@ -3,7 +3,7 @@ name: 'DVC Cache'
 match: ['DVC cache', cache, caches, cached, 'cache directory']
 ---
 
-The DVC cache is a hidden storage (by default located in the `.dvc/cache`
-directory) for files that are tracked by DVC, and their different versions.
-Learn more about it's
+The DVC cache is a hidden storage (by default in `.dvc/cache`) for files and
+directories tracked by DVC, and their different versions. Data is cached into a
```
shcheklein, line 7: it's sound ambiguous and not clear why it's needed

content/docs/command-reference/run.md
jorgeorpinel: The idea behind these `-d` reviews is to try keeping the dep file name in the `cmd` only when it's obvious why it's being passed to the script/program. Or to make it obvious, thus --in and --out. This is per #1700

But I don't have a strong opinion at all.

content/docs/command-reference/run.md
jorgeorpinel: It's yesterdays' date. Added for consistency because in the previous uses of this same sample command (in https://dvc.org/doc/command-reference/run#example-overwrite-an-existing-stage) a date is passed.

jorgeorpinel: p.s. now that I see it I realize maybe it's not a date but anyway, it's just an 8-digit number.

content/docs/command-reference/run.md
jorgeorpinel: There's no `-f` short flag for `--file` (`add` or `import*`). I think because I nagged the core team about flag naming consistency at some point last year 😬 

But sure, it makes the example self-documenting. Will do ⌛

jorgeorpinel: Oh nvmd I see you suggested to use a different name. Agree. OK, done!

content/docs/user-guide/basic-concepts/dvc-cache.md
jorgeorpinel: Oh. I just thought that an sentence that actually gives some info (causes curiosity) AND links to the structure doc was better than "Learn more about it's [structure]."

Why ambiguous?

content/docs/command-reference/run.md
shcheklein: > The idea behind these -d reviews 

it doesn't sound like a priority for now. Also. I would say it's rare to have explicit arguments like `--in` for your scripts. It might even be implicit completely, file internally reads it w/o getting it from CLI. Let's minimize these changes please.

content/docs/user-guide/basic-concepts/dvc-cache.md
jorgeorpinel: BTW what I'm trying to let the user know is that the cache doesn't preserve the directory structure in the workspace. I guess I could say it like that, is it clearer?

shcheklein: `into a special flattened structure` - it can be interpreted in so many ways. If saw this, I would start thinking that it's something very special, complicated, etc.

 > doesn't preserve the directory structure in the workspace.

it's an implementation detail - today it doesn't, tomorrow it will. I would not emphasize it here tbh.

content/docs/command-reference/run.md
jorgeorpinel: I guess the .in and .out file extensions were enough to make it self-documenting in this case... But what about the other example in line 382? Curr. `Rscript parsingxml.R --in data/Posts.xml --out data/Posts.csv`

jorgeorpinel: > doesn't sound like a priority for now

I hear you. I'm following the plan layed out in https://github.com/iterative/dvc.org/issues/2026 which starts by wrapping up pendine 1.0 updates.

jorgeorpinel: > Let's minimize these changes please.

OK sure, rolled back the 2 lines with --in/out.

content/docs/user-guide/basic-concepts/dvc-cache.md
jorgeorpinel: > I would not emphasize it here tbh.

OK, rolled back.

Or did you mean remove that sentence completely? I could call it "internal" instead of "hidden" and link that word to the doc where the structure is explained instead.

content/docs/command-reference/run.md
shcheklein: push changes?

content/docs/user-guide/basic-concepts/dvc-cache.md
shcheklein: I guess, my biggest problem is `special flattened structure` (not hidden or not)

content/docs/command-reference/run.md
jorgeorpinel: `Everything up-to-date` these changes are already gone from https://github.com/iterative/dvc.org/pull/2075/files#diff-f562129dd43887e87c126dd583630aeb0e838bf0993599f4a6a2520982a87048

content/docs/user-guide/basic-concepts/dvc-cache.md
jorgeorpinel: OK np that term is already gone 🙂

content/docs/command-reference/run.md
jorgeorpinel: PTAL

content/docs/command-reference/run.md
shcheklein: 👍

content/docs/user-guide/basic-concepts/dvc-cache.md
shcheklein: 👍

shcheklein (approved):

PR #2077, opened 2021-01-06
dvc: still pending 1.x updates (3)

jorgeorpinel:
This PR is basically for solving conflicts vs. `master` after #2075 is merged. No more changes are being added.


jorgeorpinel:
@shcheklein note that this has the exact same changes as #2074 and #2075 combined. Those were PRs into `v1` (in case you didn't notice). This one would actually get deployed when merged. Thanks

> p.s. don't worry, that's the only time ever we do this double review as there are no more pending 1.x updates to work in that way.

shcheklein (approved):

shcheklein:
Good to go, make sure that it has exactly same changes though :) I trust you on this.

jorgeorpinel:
I did double check vs the changes in those PRs while merging conflicts. Thanks 👍

PR #2085, opened 2021-01-08
Update pre-commit install commands

amin-nejad:
The `pre-commit install` command has been modified to also install the `pre-commit` hook. `pre-commit` is the default hook which gets installed but if `hook-types` are specified, _all_ relevant `hook-types` must be specified, including `pre-commit`.


jorgeorpinel:
Thanks @amin-nejad ! Looks right.

@efiop/@pared can someone on your side please double check this? (it's a small change)

efiop (approved):

Issue #2059, opened 2020-12-27
guide: make proper DVC File guide + separate reference?


jorgeorpinel:
Some/parts of the docs in https://dvc.org/doc/user-guide/dvc-files feel more like a reference than a guide. While it's arguable whether that is a problem (e.g. other docs in that section have the same situation), it seems we agree that an actual guide for writing dvc.yaml would be great to have, since it's such a major component of DVC projects. Some ideas (from https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-558335370):

> we could separate the dvc.yaml format explanation?
Like dvc.yaml short reference, then a section about "stage definition" explaining what key/value it takes, then "variables/parametriation" and then introduce loops. This way, concepts will get built (up on)

> split it into subsections - one about regular simple pointers .dvc, second about dvc.yaml and start writing a proper guide in it, start building proper structure

> similar docs (spacy, ansible, etc...
> I like [docker-compose](https://docs.docker.com/compose/compose-file/compose-file-v2/). Examples with one or two lines of text
> to write intro and help people write the yaml themselves, that fusion in [spacy](https://nightly.spacy.io/usage/projects) projects is nicer.
> we could also take a look at [pypyr](https://pypyr.io/docs/): They have separated the yaml definitions nicely into sections/concepts

And from https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-562269533

> we should have this references, there should be some general description, then basic layout, then other syntax like templating, loops, etc. Again, let's refer to some good examples that describe languages

And https://github.com/iterative/dvc.org/pull/2052#issuecomment-757025327

>  The biggest concern is still structure... these titles (inside [DVC Files](https://dvc.org/doc/user-guide/dvc-files)) do not look good to me... They should higher level.
> - .dvc file - is clearly a reference thing. Where do we put it?
> - do we need two seprate dvc.yaml sections?
> - advanced-dvc.yaml - is too heavy and a mix of a reference (precise spec) and guide (story-like)

jorgeorpinel:
Oh, and:

> I'm also thinking we will need to split dvc.yaml from .dvc files in the guide (as a 3rd level in the main nav) so these H3s become H2s and are shown in the right-hand nav
> we'll just need to repeat the Outputs and Dependencies sections in both pages (.dvc file outs/deps are a little different from those in dvc.yaml anyway so that would be more correct even if there's overlap).
From https://github.com/iterative/dvc.org/pull/2052#pullrequestreview-558335854

[...]

PR #2090, opened 2021-01-12
hotfix: fix breaking guide URLs

jorgeorpinel:
per https://github.com/iterative/dvc.org/issues/2059#issuecomment-758512781

URLs shouldn't end in .yaml (although .dvc works in https://dvc.org/doc/user-guide/dvc-files/.dvc — prob. no auto mime/type assigned by Node server)


content/docs/sidebar.json
```diff
@@ -102,11 +102,11 @@
           },
           {
             "label": "dvc.yaml & dvc.lock",
-            "slug": "dvc.yaml"
+            "slug": "dvc-yaml"
```
jorgeorpinel, line 105: Hi @rogermparent what do you think of this little bug? 🙂 

Expected I guess. Or is it something we should be able to support to have URLs ending in .anything and have them work when loaded directly? (In this case https://dvc.org/doc/user-guide/dvc-files/dvc.yaml doesn't load directly, but works once the interface is loaded on another URL e.g. from https://dvc.org/doc/user-guide)

Mainly curious, for now we'll just avoid these unconventional URLs.

shcheklein: @jorgeorpinel please, create a ticket, let's not focus on this at the moment

shcheklein (approved):

PR #2093, opened 2021-01-13
Update repro.md

akanz1:
Fixed "for for" -> "for"

Aside, i've noticed that all instances of `dvc.yaml` (and not just in repro.md) link to https://dvc.org/doc/user-guide/dvc-files/dvc.yaml which returns 404, file not found.

Instead I believe these requests should be redirected to https://dvc.org/doc/user-guide/dvc-files/dvc-yaml


shcheklein (approved):

shcheklein:
thanks @akanz1 ! 🙏

PR #2098, opened 2021-01-16
DVC File guide reorg/rewrite

### LINKED ISSUE #2059: guide: make proper DVC File guide + separate reference?

jorgeorpinel:
> Closes #2059

Summarizing goals and principles here:

1. Come up with a consistent structure for all DVC file refs, try to separate the guides from the refs.

2. Write a proper guide for (basic) writing dvc.yaml (good story). Possible structure: dvc.yaml intro guide -> stage definition (ref+guide) -> "parametrization" -> "loops". But what about [output entries](https://dvc.org/doc/user-guide/dvc-files/dvc-yaml#output-entries) and dvc.lock?

3. Structure: From the research above, looks like a single long page that starts as mostly guide and ends as mostly ref, with as many sections as needed (long right-side ToC) is the most effective. We can always separate the ref. into a different schema ref. section later if needed (criteria: if we consider the dvc.yaml "language" central to DVC). Separately, we should try to improve the UI/design for a boost. See https://github.com/iterative/dvc.org/issues/2059#issuecomment-760639766.

- [ ] Double check all redirects and links are finished before merging this.


shcheklein:
@jorgeorpinel deployment is broken for me.

jorgeorpinel:
@shcheklein maybe it was deploying for the first time when you checked? It's live now: https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/dvc-files 🙂

jorgeorpinel:
BTW I wouldn't even look at the diff in this PR, except maybe the list of files (all collapsed).

shcheklein:
@jorgeorpinel it was broken If I was trying to open these pages. Let's wait ...

shcheklein:
I still don't the low level titles in the User Guide

`DVC File` is too generic and doesn't give any sense what is it about (also like you mentioned can include a lot of other stuff)

`dvc.yaml` - is too level, also doesn't mean anything. Except if we consider this a ref. Then it should be very dry, very formal, can be indeed in some place with all other files. Then we'll need a proper guide that is friendly.

Why don't we make `dvc.yaml` a proper section "Pipelines"? or something like that? For now it can be a single page, after if it becomes long, or we see other sections (e.g. Quick overview, Creating and updating, Parametrizing (Templating), Language Reference, etc)

There can be a section in user guide - `Data Management` (which might include `.dvc`) ,etc ..


It's one of the ideas of the structure that will be user friendly, will repeat Get Started to some extent which should make it easier to navigate. I guess there could be other approaches to this.

shcheklein:
That's what I meant by broken:

Screen Shot 2021-01-15 at 8 22 36 PM

jorgeorpinel:
Yeah those 2 docs are just empty (which causes that rendering bug). WIP

jorgeorpinel:
> Why don't we make dvc.yaml a proper section "Pipelines"? For now it can be a single page

Good idea @shcheklein, maybe "Creating Pipelines"? We'll need a similar name for the .dvc file guide, perhaps "Tracking Existing Data"?

One problem is where to redirect https://dvc.org/doc/user-guide/dvc-files to though, since that currently has an index that lists the kinds of DVC Files. I'll also have to review 32 existing links to `/doc/user-guide/dvc-files` throughout docs... (I'll redirect for now in this PR)

Also, so we won't have the "DVC FIle" concept around anymore? Or should it still be the term to unify .dvc, dvc.yaml, and dvc.lock (in the Basic Concepts section — not worked on in this PR).

> Except if we consider this a ref. Then it should be very dry, very formal

I'm trying to have another stand-alone ref. per the conclusions from https://github.com/iterative/dvc.org/issues/2059#issuecomment-760639766.

jorgeorpinel:
> I'm trying to have another stand-alone ref

\* trying to **avoid** having ...

content/docs/user-guide/tracking-existing-data.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+> See [How to Merge Conflicts](/doc/user-guide/how-to/merge-conflicts) for tips
+> on managing DVC files.
+
+## .dvc YAML specification
+
+| Field  | Description                                                                                                                                                                                                                   |
+| ------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `outs` | (Required) list of [output entries](#output-entries) (details below) that represent the files or directories tracked with DVC. Typically there is only one (but several can be added or combined manually).                   |
```
jorgeorpinel, line 49: Possible problem: while these lists [look ](https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/tracking-existing-data#dvc-yaml-specification)like a much better structured ref. they're painful to deal with in MD (our formatter makes all the lines of the same length).

Should we consider a specially designed element for these so they're easy to write/manage in md and yet look like well-structured tables? Example in https://nightly.spacy.io/usage/projects#project-yml

shcheklein:
Good ideas here (with tables, with files structure, etc), but it seems that top level UG structure and titles are still messy and do not reflect the content well.

E.g. `Tracking Existing Data ...` - the whole section is about `.dvc`, also, why Existing?

Creating Pipelines - also, not sure, why not just Pipelines? (it's not only about Creating probably)? `DVC YAML files` - can it  be `Pipeline file - dvc.yaml``? 

why DVC Internals go before those two sections, which are way more important to my mind?

(and the overall UG structure is still a total mess - a bunch of random titles :( )

shcheklein:
Can it be something like:

UG:

--> Project Structure:

-----> Data files (.dvc)
-----> Project file (dvc.yaml)
-----> Internal files (.dvc directory)
-----> .dvcignore

jorgeorpinel:
> Tracking Existing Data ... - the whole section is about .dvc, also, why Existing?

Because just "tracking data" would be a different doc which would need to include info about caching, checkout, .dvc and dvc.lock. An in fact we probably need such a doc, but it's not in the scope of this PR. Since in this PR we wanted to move the dvc.yaml guide to the root of UG, we also had to move .dvc outside (it was the only page left in **DVC Files**) and thus give it a guide-like name. But we can definitely rethink the title, maybe "Tracking Data with .dvc Files"?

jorgeorpinel:
or "Understanding .dvc Files"

jorgeorpinel:
> Creating Pipelines - also, not sure, why not just Pipelines? Pipelines? (it's not only about Creating probably

That's a good question. What is this guide's purpose exactly? I thought it was to show people how to write dvc.yaml files, basically (instead of emphasizing dvc run). Please also notice related Q: https://github.com/iterative/dvc.org/issues/2059#issuecomment-761712677 (regarding the Pipelines GS page).

shcheklein:
> Because just "Tracking Data" would be a different doc which would need to include info about caching, checkout, .dvc and dvc.lock. An in fact we probably need such a doc, but it's not in the scope of this PR.

I see. The problem that even for me it was not obvious (especially Existing), so I doubt users can understand this. In this case simpler is better. I think also, that there was no intention creating a top level doc for `.dvc`?

jorgeorpinel:
Sure we can revert or change the name as needed.

> no intention creating a top level doc for .dvc?

But where do we put it for now? Currently it's in https://dvc.org/doc/user-guide/dvc-files/.dvc and the other 2 pages there will go away in this PR.

shcheklein:
> What is this guide's purpose exactly?

The whole user guide purpose is not to be tutorial, it's a "Use Manual"-like structure. May be we should rename it if you see that it creates confusion. I open to discuss the meaning of it as well.

shcheklein:
> But where do we put it for now? 

how about this for example https://github.com/iterative/dvc.org/pull/2098#issuecomment-761726514 ?

jorgeorpinel:
> Can it be something like:
>
> UG: --> Project Structure:
...

I applied this structure, updated titles and redirects, and completed the dvc.yaml guide internal structure @shcheklein

jorgeorpinel:
See https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/project-structure/data-files

jorgeorpinel:
@shcheklein do you think we need the Project Structure index? Or should it jump directly to dvc.yaml? See https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/project-structure

redirects-list.json
```diff
@@ -28,11 +28,17 @@
...
-  "^/doc/user-guide/dvc-file-format$                                                      /doc/user-guide/dvc-files-and-directories",
-  "^/doc/user-guide/dvc-files-and-directories$                                            /doc/user-guide/dvc-files",
+  "^/doc/user-guide/dvc-file-format$                                                      /doc/user-guide/project-structure",
+  "^/doc/user-guide/dvc-files-and-directories$                                            /doc/user-guide/project-structure",
+  "^/doc/user-guide/dvc-files$                                                            /doc/user-guide/project-structure",
   "^/doc/user-guide/updating-tracked-files$                                               /doc/user-guide/how-to/update-tracked-data",
   "^/doc/user-guide/how-to/update-tracked-files$                                          /doc/user-guide/how-to/update-tracked-data",
   "^/doc/user-guide/merge-conflicts$                                                      /doc/user-guide/how-to/merge-conflicts",
+  "^/doc/user-guide/dvc-files/dvc-yaml$                                                   /doc/user-guide/project-structure/pipelines-files",
```
shcheklein, line 37: no need for this redirect - we've released it not long ago, let's not complicate this

content/docs/user-guide/project-structure/index.md
```diff
@@ -0,0 +1,18 @@
+# Project Structure
+
+Using `dvc init` in your <abbr>workspace</abbr> will start a <abbr>DVC
+project</abbr>, including the default internal
+[`.dvc/`](/doc/user-guide/project-structure/internal-files) directory. From
+there on, you will create and manage different DVC files and populate the
+<abbr>cache</abbr> as you use DVC and work on your data science experiments.
+
+- `dvc.yaml` files define stages that form the pipeline(s) of a project.
```
shcheklein, line 9: let's highlight the same structure as in the sidebar here. May be even visualize it somehow?

content/docs/user-guide/project-structure/index.md
```diff
@@ -0,0 +1,18 @@
...
+
+- Files ending with the `.dvc` extension are placeholders to track data files
+  and directories.
+
+- `.dvcfiles` contain a list of paths for DVC to ignore, which can dramatically
```
shcheklein, line 14: .dvcignore files

content/docs/user-guide/project-structure/index.md
shcheklein, line 9: I would mention here also that they define metrics, parameters, etc

content/docs/user-guide/project-structure/dvc-files.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+These files use the [YAML 1.2](https://yaml.org/) file format, and a
+human-friendly schema described below. We encourage you to get familiar with it
+so you may modify, write, or generate `.dvc` files on your own.
+
+> See also
+> [How to Merge Conflicts](/doc/user-guide/how-to/merge-conflicts#dvc-files).
+
+## Specification
```
shcheklein, line 35: let's make it General Spec, and other parts make H2 so they are visible in the left sidebar

shcheklein:
Should we include "available in 2.0" note into some parts of the dvc.yaml doc?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+```yaml
+${param} # Simple
+${param.key} # Nested values through . (period)
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the expression in `dvc.yaml`, for example for the shell to interpret a
+> command substitution, escape it with a backslash, e.g. `\${...`.
+
+## Generating multiple stages
```
shcheklein, line 225: still don't like the title ... let's brainstorm a bit ...

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+
+> `cache: false` is typical here, since they're small enough for Git to version
+> directly.
+
+The commands in `dvc metrics` and `dvc plots` help you display and compare
+metrics and plots.
+
+## Templating
+
+`dvc.yaml` supports a templating format to insert values from different sources
```
shcheklein, line 100: for SEO purposes and so that users can find it-  probably makes sense to include parametrization here

shcheklein:
Looks like a good iteration - some minor comments, @skshetry @dberenbaum could you review please? what's your take on this? (here is the link https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/project-structure )

content/docs/user-guide/project-structure/pipelines-files.md
dberenbaum: +1. Throwing out other title ideas:

* Looping over stages
* `foreach` stages

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+  build@uk:
+    cmd: python train.py 'uk' 3 10
+    outs:
+      - model-uk.hdfs
+  build@us: ...
+```
+
+Importantly, dictionaries [from parameters](#templating) files can be used in
+`foreach` multi-stages as well:
+
```
dberenbaum, line 315: Could an example `params.yaml` be included here? I'm not totally clear how this works. I'm assuming `params.yaml` would look like:

```yaml
myobject:
  us:
    prop1: 3
    prop2: 10
  uk:
    prop1: 10
    prop2: 15
```

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+| `cmd`            | (Required) One or more commands executed by the stage (may contain either a single value or a list). Commands are executed sequentially until all are finished or until one of them fails (see `dvc repro`).                                                                              |
+| `wdir`           | Working directory for the stage command to run in (relative to the file's location). Any paths in other fields are also based on this. It defaults to `.` (the file's location).                                                                                                          |
+| `deps`           | List of <abbr>dependencies</abbr> paths of this stage (relative to `wdir`).                                                                                                                                                                                                               |
```
dberenbaum, line 335: Should `dependencies` be singular (`dependency`)?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+| `outs`           | List of <abbr>outputs</abbr> paths of this stage (relative to `wdir`). See [Output entries](#output-entries) for more details.                                                                                                                                                            |
```
dberenbaum, line 336: Should `outputs` be singular (output)?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+| `frozen`         | Whether or not this stage is frozen from reproduction                                                                                                                                                                                                                                     |
+| `always_changed` | Whether or not this stage is considered as changed by commands such as `dvc status` and `dvc repro`. `false` by default                                                                                                                                                                   |
+| `meta`           | (Optional) arbitrary metadata can be added manually with this field. Any YAML contents is supported. `meta` contents are ignored by DVC, but they can be meaningful for user processes that read or write `.dvc` files directly.                                                          |
```
dberenbaum, line 342: "contents" should be singular ("content").

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Output entries
+
+| Field     | Description                                                                                                                                |
+| --------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| `cache`   | Whether or not this file or directory is <abbr>cached</abbr> (`true` by default). See the `--no-commit` option of `dvc add`.               |
+| `persist` | Whether the output file/dir should remain in place while `dvc repro` runs (`false` by default: outputs are deleted when `dvc repro` starts |
+| `desc`    | (Optional) user description for this output. This doesn't affect any DVC operations.                                                       |
+
+## dvc.lock file
```
dberenbaum, line 364: As mentioned earlier, should `dvc.lock` have its own page? It is a separate file type, and this page is long even without it.

Also, should there be some explanation of the relationship between `dvc.lock` and `.dvc` files? Since they contain similar info, I'm not sure the differences between them are clear.

dberenbaum (commented):
I might have gone overboard with the review comments :) Since I don't have as much context here, many of the comments might not make sense or warrant changes, but it's helpful for me to learn. Overall, I think it looks great, and it's because it's so detailed and particular already that I'm able to digest it and provide feedback.

redirects-list.json
jorgeorpinel: Actually this is probably necessary at least temporarily, because https://dvc.org/doc/user-guide/dvc-files already received the rank and traffic from the previous URLs (currently [indexed](https://search.google.com/u/1/search-console/inspect?resource_id=sc-domain%3Advc.org&id=X4AehNlF5NqXR8XBNrJVUA) by Google):

image

We want the traffic and rank to transfer to the new URL ([/doc/user-guide/project-structure](https://dvc-landing-2-0-pipelin-rttezd.herokuapp.com/doc/user-guide/project-structure)) I think.

jorgeorpinel: But what we can do now is remove the previous 2 redirect lines above (those URLs are no longer indexed on Google) to simplify the redirects list.

> It could break a few links from old web pages or blog posts outside our domain though.

jorgeorpinel:
> Should we include "available in 2.0" note into some parts of the dvc.yaml doc?

@shcheklein depends on when this gets merged. I'd rather keep it out for now and add it last minute if needed.

shcheklein:
@jorgeorpinel I thought we wanted to merge and iterate fast (and thus avoiding redirects and stuff). There are not major problems in this iteration from what I see.

content/docs/user-guide/project-structure/index.md
jorgeorpinel: OK, I updated the bullet list to match and look like the nav. I'm not sure what you mean by "visualize it" though.

jorgeorpinel: Added.

content/docs/user-guide/project-structure/index.md
```diff
@@ -0,0 +1,18 @@
...
+  increase its operational performance.
+
```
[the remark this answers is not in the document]
jorgeorpinel: This was a typo @dberenbaum, should be .dvcignore. Fixed now. Will get to your other comments when I'm done with Ivan's 😬

content/docs/user-guide/project-structure/dvc-files.md
jorgeorpinel: Good call, done. Not sure about the title though. I went for **Specification (file-level)** for now. WDYT?

shcheklein: `Specification (file-level) ` there should be a term for this - something like File schema or File syntax

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: I've been opposed to the "loop" analogy from the beginning here haha. But it has been discussed (see iterative/dvc/issues/5312) so at this point if that's the consensus, I have no strong opinion against. "Foreach-stages" or maybe "Stage loops" do sound user-friendly.

content/docs/user-guide/project-structure/dvc-files.md
jorgeorpinel: What I'm trying to indicate with "file-level" is that the section focuses on the root fields, subfields are explained in further sections. For now I went back to just **Specification** and added a note about what the section covers.

I also think keeping the word "Specification" is desirable (same as in the dvc.yaml guide) but no strong opinion. "File schema" (in both places) can also work (we do already use the term "schema" in the text).

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: That's what this section is. Or do you mean the term "parameterization"?

jorgeorpinel: p.s. we don't have that term anywhere in docs anymore BTW.

jorgeorpinel: Yeah I guess you mean the term. Added to the note for now.

jorgeorpinel:
OK @shcheklein, added the 2.0 warning (and finished addressing your other comments). Does that mean you approve this? 😏

shcheklein (approved):

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
+# Pipelines Files (`dvc.yaml`)
+
+You can construct data science or machine learning pipelines by defining
+individual [stages](/doc/command-reference/run) in one or more `dvc.yaml` files
+(or _pipelines files_). Stages form a pipeline when they connect with each other
+(forming a _dependency graph_, see `dvc dag`). Refer to
+[Data Pipelines](/doc/start/data-pipelines).
+
+> Note that a helper command, `dvc run`, is available to create (and execute)
```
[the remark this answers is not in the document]
jorgeorpinel: > OK Dave, I'm back with you.

We're actually trying to encourage people to write dvc.yaml files manually but still need to mention `dvc run` since it's there.

Is your comment based on your previous experience with DVC where `run` was more prevalent, or do you think the text here doesn't convey the intentions I just mentioned?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,420 @@
...
+> `{"grp": {"a": 1}}` merges with `{"grp": {"b": 2}}`, but not with
+> `{"grp": {"a": 7}}`.
+
+> Note that values from `vars` are not tracked like parameters.
+
+To load additional params files, list them in the top `vars`, in the desired
+order, e.g.:
+
+> Their paths will be evaluated based on
+> [`wdir`](/doc/user-guide/dvc-files/dvc.yaml#accepted-fields), if one given.
```
skshetry, line 166: This note should be moved on "Stage-specific values" below.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,420 @@
...
+<abbr>dependencies</abbr> (`deps`). DVC will check whether they have changed to
+decide whether the stage requires re-execution (see `dvc status`).
+
+If the command writes files or directories, they can be defined as
+<abbr>outputs</abbr> (`outs`). DVC will track them going forward (similar to
+using `dvc add`).
+
+### Parameter dependencies
+
+[Parameters](/doc/command-reference/params) are a special type of stage
```
skshetry, line 50: There's no non-default case? Was it documented before?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,420 @@
...
+    vars:
+      - params.json:build
+      - model:
+          filename: 'model-us.hdf5'
+    cmd: python train.py ${build.epochs} --out ${model.filename}
+    outs:
+      - ${model.filename}
+```
+
+⚠️ Known limitations of local `vars`:
```
skshetry, line 209: I don't think this warning is useful as both of them are expected.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,420 @@
...
+to iterate on, while `do` contains the regular stage fields (`cmd`, `outs`,
+etc.). Here's a simple example:
+
+```yaml
+stages:
+  cleanups: # Multi-stage
+    foreach: # List of simple values
+      - raw1
+      - labels1
+      - raw2 labels2
```
skshetry, line 242: Maybe just `raw2` or `labels2`? This might confuse users when they look into dvc.lock file.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Stages
+
+`stages` are named by the user and have a flexible structure. Here's a simple
+example:
+
+```yaml
+stages:
+  print2cols:
+    cmd: awk '{print $1 "\t" $2}' lines.txt > 2columns.txt
```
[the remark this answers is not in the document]
jorgeorpinel: Good catch. This didn't use to be the first example in a previous iteration. I changed it to something more conceptual now while avoiding unrealistic echo/cat.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,420 @@
...
+```yaml
+stages:
+  build-us:
+    cmd: >-
+      python train.py
+      --thresh ${models.us.threshold}
+      --out ${models.us.filename}
+    outs:
+      - ${models.us.filename}:
+        cache: true
```
skshetry, line 133:
```suggestion
          cache: true
```

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+      - lines.txt
+    outs:
+      - 2columns.txt
+```
+
+The most important part of a stage it's the terminal command(s) it executes
+(`cmd` field). This is what DVC runs when the stage is reproduced (see
+`dvc repro`).
+
+If a program executed by the command read input files, they can be defined as
```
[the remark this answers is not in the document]
jorgeorpinel: "read" was a typo (fixed). Removed "a program executed by" although it was more correct. I guess it's too verbose. Did other simplifications too. Not sure about "reads directories" though, I phrased that differently, PTAL.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+human-friendly schema explained below. We encourage you to get familiar with it
+so you may modify, write, or generate stages and pipelines on your own.
+
+> Note that we use [GNU/Linux](https://www.gnu.org/software/software.html) in
+> most of our examples.
+
+## Stages
+
+`stages` are named by the user and have a flexible structure. Here's a simple
+example:
```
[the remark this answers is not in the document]
jorgeorpinel: Good idea. Updated.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+stages:
+  build-us:
+    desc: ${desc}
+    cmd: python train.py --thresh ${models.us.threshold}
+```
+
+> DVC merges values from params files and `vars` when possible. For example,
+> `{"grp": {"a": 1}}` merges with `{"grp": {"b": 2}}`, but not with
+> `{"grp": {"a": 7}}`.
+
```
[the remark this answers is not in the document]
jorgeorpinel: > If params.yaml has {"grp": {"a": 1}} and vars has {"grp": {"a": 7}}, what will be the value of grp.a? What if the value is defined in another params file?

These can't be merged, and error will happen. That's what the main point (1.) in iterative/dvc/issues/5312 is about.

> the order is params.yaml followed by anything in vars in the order listed (including additional params files). Is that right?

Yes.

> leaving out merging until after introducing each source

Yeah maybe. I moved it all the way down for now. The sentence is confusing anyway but I'm waiting to see if the behavior is changed before trying to rewrite it. Any suggestions are welcome of course.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -0,0 +1,416 @@
...
+The substitution expression supports these forms:
+
+```yaml
+${param} # Simple
+${param.key} # Nested values through . (period)
+${param.list[0]} # List elements via index in [] (square brackets)
+```
+
+> To use the expression in `dvc.yaml`, for example for the shell to interpret a
+> command substitution, escape it with a backslash, e.g. `\${...`.
```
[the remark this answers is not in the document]
jorgeorpinel: Oops this was missing "literally":

> To use the expression literally in `dvc.yaml`...

I also took the opportunity to introduce the term "replace" here, PTAL.

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: I fear it'll be too many code blocks... For now I changed the link to go to https://dvc.org/doc/command-reference/params instead. At some point (hopefully soon) those links will go to a new Concepts section with the explanation and better examples of each thing (in this case parameters files) — see #550. WDYT?

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Nope, it's a list 🙂

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Negative

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Fixed.

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: It has it's own page in some iteration and it seemed too little for a whole page. Plus it's kind of part of dvc.yaml (meaningless without it).

And as long as we have a decent right-hand ToC nav, length is probably not a big concern. See https://github.com/iterative/dvc.org/issues/2059#issuecomment-760639766

> should there be some explanation of the relationship between dvc.lock and .dvc files? ... not sure the differences between them are clear

There's no direct relationship, they share one of their purposes (mentioned in the 2nd bullet). Please refer to iterative/dvc/issues/3936 for a deeper dive on this situation.

Thanks for all the feedback!

content/docs/user-guide/project-structure/pipelines-files.md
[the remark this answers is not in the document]
dberenbaum: Yes, I think it makes sense in the context of wanting to discourage `dvc run`. I'm not sure it's clear from the note that `dvc run` adds the stage to `dvc.yaml`, but maybe that's unnecessary if we don't really want users relying on it.

content/docs/user-guide/project-structure/pipelines-files.md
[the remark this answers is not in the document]
dberenbaum: Yup, correct grammar but maybe just too verbose.

content/docs/user-guide/project-structure/pipelines-files.md
dberenbaum: Sounds good. Once a Concepts section exists, it would be great to have the templating example use the same params that are shown in the example `params.yaml`.

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: > There's no non-default [params] case?

Sorry, Idk what you mean @skshetry.

Params was never explained in the dvc.yaml file before (only mentioned the `params` fields).

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Hmmm. Why would users expect or assume any of this? The one about `wdir` is a logical consequence of the loading order but that's a complex thing to grasp (for example it wasn't obvious to me at first and I had to ask). The one about `foreach` has some internal reason (I don't even remember it now) pretty much imposible for a user to know.

Plus we have all sorts of usage warnings throughout docs. I see this in line with that.

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: YAML blocks are auto-formatted in docs. A lot of them have strange indentation which I'm not sure would actually work 😆 Does it actually matter to DVC?

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: > [note on wdir] should be moved on "Stage-specific values

You mean also mentioned in that part? Because it's important here too (and this part comes first). For now I made it more general so people can assume it applies to local vars too.

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Yeah I think this was a typo, fixed. Thanks!

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Wait no, you were right. It should be singular because "paths" is plural 🤦 

Fixed (also in outs)!

content/docs/user-guide/project-structure/pipelines-files.md
dberenbaum: Same here as above ("paths" is plural so "outputs" should be singular).

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Never mind, this change did pass tests 👍

jorgeorpinel:
OK merging this since it's approved and I've gotten no further replies. But feel free to comment more: I can address anything left in another PR.

PR #2091, opened 2021-01-12
Add docs regaring --to-remote option for add/import-url

isidentical:
Bound to iterative/dvc#5198

PR #2104, opened 2021-01-19
docs: external: add a disclaimer

efiop:
People find this article quite often and tend to misuse it. Added a simple disclaimer to try to redirect them.

### DIFF
--- a/content/docs/user-guide/managing-external-data.md
+++ b/content/docs/user-guide/managing-external-data.md
@@ -1,6 +1,7 @@
 # Managing External Data
 
-> ⚠️ Artifacts added with --external are not affected by
+> ⚠️ This is an advanced feature that we don't recommend using unless you really
+> know what you are doing. Artifacts added with --external are not affected by
 > `dvc push/pull/status -c`. You are likely looking for
 > [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520)
 > functionality or `dvc import-url`
--- a/content/docs/user-guide/managing-external-data.md
+++ b/content/docs/user-guide/managing-external-data.md
@@ -1,7 +1,7 @@
 # Managing External Data
 
-> ⚠️ Artifacts added with --external are not affected by dvc push/pull/status
-> -c. You are likely looking for
+> ⚠️ Artifacts added with --external are not affected by
+> `dvc push/pull/status -c`. You are likely looking for
 > [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520)
 > functionality or `dvc import-url`
 
--- a/content/docs/user-guide/managing-external-data.md
+++ b/content/docs/user-guide/managing-external-data.md
@@ -1,5 +1,8 @@
 # Managing External Data
 
+> ⚠️ Artifacts added with --external are not affected by dvc push/pull/status -c.
+> You are likely looking for [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520) functionality or `dvc import-url`
+
 There are cases when data is so large, or its processing is organized in such a
 way, that its preferable to avoid moving it from its original location. For
 example data on a network attached storage (NAS), processing data on HDFS,
--- a/content/docs/user-guide/managing-external-data.md
+++ b/content/docs/user-guide/managing-external-data.md
@@ -1,7 +1,9 @@
 # Managing External Data
 
-> ⚠️ Artifacts added with --external are not affected by dvc push/pull/status -c.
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/managing-external-data.md
```diff
@@ -1,5 +1,10 @@
 # Managing External Data
 
+> ⚠️ Artifacts added with --external are not affected by dvc push/pull/status
+> -c. You are likely looking for
+> [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520)
+> functionality or `dvc import-url`
```
jorgeorpinel, line 6:
```suggestion
> ⚠️ Artifacts added with --external are not affected by
> `dvc push/pull/status -c`. You are likely looking for
> [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520)
> functionality or `dvc import-url`
```

content/docs/user-guide/managing-external-data.md
```diff
@@ -1,5 +1,10 @@
 # Managing External Data
 
+> ⚠️ Artifacts added with --external are not affected by
+> `dvc push/pull/status -c`. You are likely looking for
+> [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520)
+> functionality or `dvc import-url`
```
jorgeorpinel, line 6: Good note! If it's not urgent should we make this part of https://github.com/iterative/dvc.org/pull/2091#pullrequestreview-569642813 ?

And prob. link to /doc/command-reference/add#transferring-data-directly-to-the-remote instead.

jorgeorpinel: Cc @isidentical

content/docs/user-guide/managing-external-data.md
```diff
@@ -1,7 +1,7 @@
 # Managing External Data
 
-> ⚠️ Artifacts added with --external are not affected by dvc push/pull/status
-> -c. You are likely looking for
+> ⚠️ Artifacts added with --external are not affected by
```
shcheklein, line 3: `are not affected by "dvc push/pull/status -c"` - @efiop  what would be the explicit way to explain this to someone less technical/DVC proficient?

efiop: @shcheklein Not sure what you mean.

content/docs/user-guide/managing-external-data.md
efiop: It will have to wait for 2.0 anyways, and even then people might try to use this. Better to have this note anyways.

content/docs/user-guide/managing-external-data.md
shcheklein: I guess my point here is that this note is so specific and technical that I would not expect it being understood by users exploring different ways to deal with data with DVC.

Could you explain here in a comment, not using any DVC terms, what do we want to convey with this disclaimer? I'll try to help to find a better way to phrase this.

shcheklein: I mean that this note should be probably written like this (if I understand its meaning correctly):

Note, the way to handle data described below can be considered an advanced alternative to an actual `dvc add`, `dvc push` workflow. In some cases you do need it indeed, in other cases it might be you have a better, simpler solution. If, for example, you are looking for a way to add data to DVC without moving it to the machine first, you might check this option [straight-to-remote/cache](https://github.com/iterative/dvc/issues/4520), or `dvc import-url` ...

Otherwise my point that it's too low level, it will be very hard for folks exploring different ways of managing data with DVC to understand what implications does `are not affected by dvc push/pull/status -c` have, why is that a problem.

They way it is phrased now, it feels very abrupt be at the very beginning - no context, no explanation.

efiop: The aim was to put a short and quick disclaimer to address the typical questions we get. I might need to get back to this later for a verbose explanation.

shcheklein: Yep, I think we are on the same page on this. I see all this questions and understand what you are trying to achieve. My only concern that they way it's written it won't move a needle since people won't understand this. People don't have enough context when they come here. For example, they might easily think that "advanced" is for them (why not, right?). If we don't know how to explain it better, I'm fine - let's merge it as is, may be it'll change it a bit indeed.

shcheklein (approved):

content/docs/user-guide/managing-external-data.md
efiop: Leaving as is to save time for now.

PR #2130, opened 2021-02-01
remote: terminology and sample standardization

jorgeorpinel:
From my older git stash (finally close to clearing it! No rush on reviewing BTW)

> "Forwardport" of 997f6667 from `v1`

- [ ] UPDATE: After this, (make 1+ PRs) from #2132


shcheklein (approved):
If it has been reviewed in v1, let's merge it ... why do we do fowardports btw? it should be other way around ideally

jorgeorpinel:
OK! I think I merged the one to `v1` after only reviewing it myself? Anyway, I don't expect any more of these (my pre-2.0 git stash is clear now!)

> why do we do fowardports btw?

These old changes don't carry > 2.0 code (from master). I apply them to `v1` first because otherwise in the process of merging to master (if there are conflicts) the branch could get "contaminated" with >2.0 code, and it would be harder to backport.

PR #2135, opened 2021-02-01
Preload fonts

mvshmakov:
This PR adds the fonts preloading to every page of the `dvc.org`. The key aim here is to get rid of font fallback problem can be seen on each website reload (or [FOUT](https://css-tricks.com/fout-foit-foft/)):

Feb-01-2021 17-51-41

This PR changes it to this:

Feb-01-2021 18-03-42

It improves CLS (cumulative layout shift metric) which implies UX improvements. There is a downside of this approach though - every new page added requires a run of `yarn preload-fonts` (even a new blog post) to grab the fonts required for it. Another option is to manually add all fonts that are used in the application in `<link rel="preload" />`, but this brings the possibility of unused on the page but still required to download fonts which potentially can make the page download less fact.


jorgeorpinel:
Can't `gatsby-preload-fonts` just be included in `yarn build`?

And does this affect `yarn develop`?

Thanks

shcheklein:
@iAdramelk @rogermparent I remember we had a discussion about this and made a decision not do font preload?

I guess the major reason was "time-to-render" - if font is not loaded you still can start reading, otherwise you see a blank page longer

did we have any other concerns?

rogermparent:
> Can't `gatsby-preload-fonts` just be included in `yarn build`?

I haven't run it yet, but I would assume the `preload-fonts` script takes an abnormally long time because it crawls the whole site to generate font preload data- that, and there isn't really any other reason for it to be separate despite the wide criteria where it must be re-run.

The ideal solution may be to keep the scripts separate in dev, but always run before deploy.

rogermparent:
> @iAdramelk @rogermparent I remember we had a discussion about this and made a decision not do font preload?

I don't remember personally, either I forgot or this discussion happened before I was working on this project.

mvshmakov:
> Can't gatsby-preload-fonts just be included in yarn build?

Not, unfortunately. It requires a running development server to crawl all the web pages that are in the app.

> And does this affect yarn develop

Not at all. If you are interested if these fonts are preloaded on development - sure they are.

> I guess the major reason was "time-to-render" - if the font is not loaded you still can start reading, otherwise you see a blank page longer

@ivan this might be an issue only on the initial load with cold caches. They are well-optimized in terms of size and requests are made in parallel, so it is equivalent of downloading ~30kb. When fonts are cached it is worth a non-blocking roundtrip to receive 304 from the server and grab them from the local memory (almost free). Benefits in terms of UX improvements are much higher for the target users (I assume the most of them are using PC with a stable connection).

> I would assume the preload-fonts script takes an abnormally long time because it crawls the whole site to generate font preload data

@rogermparent and you would be absolutely right, unfortunately. It takes ~250 sec on my PC to crawl all pages. Again, the alternative solution might be to just use manually-added `<link rel="preload" />` with all fonts from the single `font-face` defined in the app.

> The ideal solution may be to keep the scripts separate in dev, but always run before deploy.

Great idea!

shcheklein:
> `@shcheklein this might be an issue only on the initial load with cold caches`

for me it's every time I open the page in the new tab

> The ideal solution may be to keep the scripts separate in dev, but always run before deploy.

is it implemented in this PR?

shcheklein:
Guys, I'm closing this for now. Too much stuff in other places and it's not a priority. Let's get back to this discussion when we have bit more time.

iAdramelk:
@shcheklein sorry for the late answer. 

There are different strategies for font preload. What we discussed in the past was inlining font file inside css as base64 string. This would fully remove font blinking, but would significantly increase css file size and block page until css is fully parsed as result so we decide not doing it.

@mvshmakov did the different thing here - he added script that automatically appends `<link rel="preload">` with the font file address to the <head> of each page that uses it. It didn't prevent font blinking because there is still separate request, but it should have reduced time before font is loaded – this way font should start loading earlier.

shcheklein:
@mvshmakov so, if we don't have a concern with content being delayed and we have a benefit of blinking taking less time, what else prevents this PR from being merged? Concern with ` It takes ~250 sec on my PC to crawl all pages. ` ? Is it easy to resolve?

@iAdramelk thanks!

iAdramelk:
@mvshmakov my 2 cents. If I remember correctly only custom font that is used on site right now is `BrandonGrotesque` in a few variations and at least some of its variations are used on every page. Even if some variations are not used on some page it make sense to preload them anyway. So I'd replace this plugin with list of manually added `<link preload /> tags. 

It can cause problems in the future if you add new fonts for some pages that are only used on these pages, but it can be addressed then in a few different ways. For now it should be good enough.

mvshmakov:
Folks, sorry for the late response, there is much stuff going on indeed.

> what else prevents this PR from being merged? Concern with It takes ~250 sec on my PC to crawl all pages. ? Is it easy to resolve?

@shcheklein This is the only one, but as @iAdramelk stated, the whole issue might be worth a few lines with manual `<link rel="preload" />` which would sometimes load stuff we don't really need on the page. Using this, no need for crawling the pages is required at all, so this "preventing from merge" issue would not be a thing.

Let me test this thing locally and with high possibility, I'll push it here.

shcheklein:
hey, @mvshmakov , ping :)

jorgeorpinel:
Hi! Curious about this. Is there a plan to finish it, is it blocked, no longer needed? Thanks

rogermparent:
> Hi! Curious about this. Is there a plan to finish it, is it blocked, no longer needed? Thanks

It seems that if we were going to preload fonts, we'd want to do it as @mvshmakov stated: "a few lines with manual <link rel="preload" /> which would sometimes load stuff we don't really need on the page."

Since the implementation we're thinking about is completely different from what's in this PR, I'd say it's closeable. We can always reopen if we have to.

PR #2116, opened 2021-01-26
Januray Gems for review

elleobrien:
Will update discussion link and cover shortly.

Note that two questions address the `-j` flag and I'm not sure if it's redundant to a reader. Please let me know if you think we should remove one?

PR #2138, opened 2021-02-02
clarify use of env vars for self-signed certs in http remotes

skshetry:
removed the ssl_verify flag mention as well

From https://github.com/iterative/dvc.org/pull/2116#discussion_r567695284


shcheklein (approved):

shcheklein:
Thanks @skshetry !!

PR #2131, opened 2021-02-01
Add schema keywords in lockfile

skshetry:
This is a 2.0 change introduced in https://github.com/iterative/dvc/pull/5222.

Though there wasn't `stages` keyword in `dvc.lock` file in 1.0 version, I found a couple of files that had them. But, `schema` and `stages` were introduced at the same time. `schema` is just set to 2.0 when DVC v2.0 tries to dump into the lockfile. v1.0 lockfiles are migrated at the same time if needed, to v2.0. `stages` just wrap the pre-existing structure of lock entries.

### In 1.0
```yaml
echo-hello:
  cmd: echo hello world
echo-hi:
  cmd: echo hi
```

### In 2.0
```yaml
schema: '2.0'
stages:
  echo-hello:
    cmd: echo hello world
  echo-hi:
    cmd: echo hi
```
> You may disregard these recommendations if you used the **Edit on GitHub** button from dvc.org to improve a doc in place.

❗ Please read the guidelines in the [Contributing to the Documentation](https://dvc.org/doc/user-guide/contributing/docs) list if you make any substantial changes to the documentation or JS engine.

🐛 Please make sure to mention `Fix #issue` (if applicable) in the description of the PR. This causes GitHub to close it automatically when the PR is merged.

Please choose to [allow us](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to edit your branch when creating the PR.

Thank you for the contribution - we'll try to review it as soon as possible. 🙏

PR #2139, opened 2021-02-02
dvc.lock updates

jorgeorpinel:
Continuing #2131


shcheklein (approved):

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -364,7 +364,11 @@ validation and auto-completion.
 > See also
 > [How to Merge Conflicts](/doc/user-guide/how-to/merge-conflicts#dvcyaml).
 
-### Output entries
+### Simple output entries
```
shcheklein, line 367: hmm, I don't like this term "simple". There is nothing specifically simple or complex about those entries. It's totally fine to have them defined differently, with a bit different schema if we very clear in docs about this.

jorgeorpinel: Agree. But the reason I changed the title is that the output (and dependency) entries of dvc.lock files are = those in .dvc files, not these (dvc.yaml output entries). And in the dvc.lock section below I use the terms "out/dep entry", linked to the corresponding section in the .dvc file guide, to avoid repeating those sections in this doc again.

Repeating here is a good option and probably even less confusing in some ways, but we would still need to somehow emphasize that dvc.yaml outs/deps entries are different from those in dvc.lock. Probably we would do that with different section names anyway.

TLDR: How about *dvc.yaml output entries* and **dvc.lock output entries**?

shcheklein: To my mind we are already in the section that describes the dvc.yaml schema and ideally it should be clear from the context that this is about `dvc.yaml`. I can see how additional message can help, but anything that comes to my mind would create even more confusion (people are not aware about other files, different types of outputs, etc).

jorgeorpinel: You know what I looked into this and I think I have an easier solution ⌛

jorgeorpinel: OK I simplified by calling the section in dvc.yaml "Output subfields" (which is what it is), PTAL.

If it's still confusing we could consider calling the sections in .dvc "Output/Dependency **tracking** entries".

jorgeorpinel: p.s. TBH the ideal solution would be to have tables inside tables but that's not doable on md, would need some design and frontend work.

shcheklein: sounds good!

jorgeorpinel:
OK, merging then 🙂

PR #2132, opened 2021-02-01
[v1] last (old) misc updates

jorgeorpinel:
From my older git stash (done clearing!)

> PR into `v1` for visualization purposes for now.

- [ ] (Merge and) port to `master` after #2130 is merged

PR #2141, opened 2021-02-02
Last (old) misc updates (1/3)

jorgeorpinel:
From #2132

1. Use cases copy edits


jorgeorpinel:
Wait something is wrong here... Should only be 2 changed files.

PR #2142, opened 2021-02-02
Last (old) misc updates (1/4) from v1

jorgeorpinel:
1.

- Review `import* --no-exec`
- Review `push/pull/fetch -j` descs
- A copy edit in a link


shcheklein (approved):

jorgeorpinel:
Hmmm the issue with the broken links is also happening here (https://github.com/iterative/dvc.org/pull/2142/checks?check_run_id=1817012043). I see. We removed those redirects so I should update those old links from the blog after all... ⌛

jorgeorpinel:
Hmmm... So I'm not getting what the problem is actually. I can't find the links listed in https://github.com/iterative/dvc.org/runs/1817012043 in the actual blog. Something wrong with the check maybe?

Merging this for now.

PR #2163, opened 2021-02-09
cml: doc titles [SEO] and other fixes

jorgeorpinel:
fix 118858c9582ef4d12865fb319b83396b99b82951 et al.

### DIFF
--- a/content/docs/sidebar.json
+++ b/content/docs/sidebar.json
@@ -127,7 +127,6 @@
           "merge-conflicts"
         ]
       },
-      "experiment-management",
       "setup-google-drive-remote",
       "large-dataset-optimization",
       "external-dependencies",
--- a/content/docs/sidebar.json
+++ b/content/docs/sidebar.json
@@ -127,6 +127,7 @@
           "merge-conflicts"
         ]
       },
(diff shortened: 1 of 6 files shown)


content/docs/sidebar.json
```diff
@@ -412,7 +413,10 @@
     "source": "cml/index.md",
     "icon": "cml",
     "children": [
-      "get-started",
+      {
+        "label": "Get Started (GitHub)",
+        "slug": "start-github"
```
jorgeorpinel, line 418: Took the opportunity to change this URL, which will be needed if we're publishing another Get Started for Gitlab, for example. I just need to check whether a redirect is needed to transfer any search traffic... ⌛

jorgeorpinel: Hmm yeah it's already [indexed](https://search.google.com/u/1/search-console/inspect?resource_id=sc-domain%3Advc.org&id=fDlRfKU-LaU5thDLvisO3g)... But it has almost 0 clicks from search yet, so maybe not worth redirecting

image

shcheklein (approved):

PR #2166, opened 2021-02-09
revert irrelevant changes from --to-remote

isidentical:
This reverts the irrelevant changes done in #2091


jorgeorpinel (approved):
Thanks @isidentical~ I'm just confused as to whether these changes should be in master or not at this point... Merging for now

PR #2177, opened 2021-02-11
Basic Concepts Section for the Documentation

iesahin:


iesahin:
I checked `src/gatsby/models/glossary/index.js` for [failing codeclimate test](https://codeclimate.com/github/iterative/dvc.org/pull/2177) and decided splitting the mentioned `createSchemaCustomization` function actually increases the complexity. Leaving as it is.

shcheklein:
@iesahin could we take the JS part as a separate PR please and describe what it does? (Overall a bit of preference here - it's easier for us to review and merge smaller and focused PRs)

content/docs/sidebar.json
```diff
@@ -91,6 +91,26 @@
...
+          "stage",
+          "dvc-cache",
+          "dvc-metafiles",
+          "remote-storage",
+          "import-stage",
+          "metrics-and-plots",
+          "parameters",
+          "run-cache"
+        ]
+      },
```
jorgeorpinel, line 113: Too many changes in a single PR 🙂  Let's start with the index (if there's one) and 1-3 of the most important concept pages?

That said, I also have a couple comments on (1) the concept naming, and (2) level of abstraction:

1. Let's use `dvc-` everywhere (where it can apply) or nowhere, I think.
2. Not everything that has a tooltip needs a concept page e.g. "external dependency" is a bit specific and falls within the "dependency" concept. Not sure which ones to keep exactly, please suggest.

Thanks

jorgeorpinel: Also on the level of abstraction, we could even consider putting related concepts together e.g. cache and remote, pipeline and stage (they could still each need separate tooltips though). WDYT?

package.json
```diff
@@ -77,6 +77,7 @@
     "react-slick": "^0.25.2",
     "react-use": "^14.0.0",
     "rehype-react": "^5.0.1",
+    "remark-preset-lint-recommended": "^5.0.0",
```
jorgeorpinel, line 80: What was this for again? 🙂

content/docs/user-guide/concepts/data-pipelines.md
```diff
@@ -0,0 +1,34 @@
...
+  outputs and inputs. They use some initial data, may produce intermediate
+  artifacts, and reach a final result.
+  [📖](/doc/user-guide/concepts/data-pipelines)
+description: >-
+  A data pipeline is a series of data processing stages. They use some initial
+  data, may produce intermediate artifacts, and reach a final result.
+---
+
+<!-- keywords: data pipeline, machine learning pipeline, devops for data science, devops for machine learning, "MLops", "what is a data pipeline?", "data pipeline examples", "machine learning (ML) workflow", "data science workflow", "data science pipeline workflow" -->
+
```
jorgeorpinel, line 15: So these are desired terms we want to try to get into the concept text for SEO purposes.

content/docs/user-guide/concepts/data-pipelines.md
```diff
@@ -0,0 +1,34 @@
...
+# Data Pipelines
+
+A data pipeline, in general, is a series of data processing <abbr>stages</abbr>
+like data gathering, data transformation, model training, model testing and
+reporting <abbr>metrics</abbr> and results. A pipeline may produce intermediate
```
jorgeorpinel, line 20: As you may have realized, a lot of the text in this branch comes from existing content, in this case https://dvc.org/doc/command-reference/dag

So the PR should also remove most of those explanations from the cmd ref. — another reason to focus on just one concept at a time per PR probably — not to mention all the link/tooltip updates that some of these concepts will require throughout docs 😱

content/docs/sidebar.json
jorgeorpinel: p.s. please read my comments on Guthub because I've edited some from the original text you get via email.

iesahin:
I fully agree that this should be split into smaller PRs but this is a breaking change. JS part is actually copied from what @rogermparent did in [ug-basic-concepts](https://github.com/iterative/dvc.org/tree/ug-basic-concepts) [here](https://github.com/iterative/dvc.org/commit/fa56570bb66ad3838e9075569abe3963cfa328a3#diff-c8d27d0a004cb7dcff8d1f42e800ce514106da88412e2fda4dd7673fd9acb2a5)

If we don't include JS changes to this, Basic Concepts don't appear in the sidebar and tooltips become large documents. JS part involves showing `tooltip` frontmatter instead of the whole page and actually directly copied from @rogermparent 's work. 

Because of this I had to bundle all pages together. If we apply only the JS part and an `.md` file under `/concepts/` dir doesn't have `tooltip:` frontmatter, it breaks the existing glossary/tooltip functionality. The content is also extended in most of the pages.

I'll be merging `master` into this branch and update the documents as you review but splitting this into smaller pieces will probably do more harm than good. 

Thanks. @shcheklein @jorgeorpinel

package.json
iesahin: It's from @rogermparent 's work [here](https://github.com/iterative/dvc.org/commit/fa56570bb66ad3838e9075569abe3963cfa328a3#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) 

Actually this is my first experience with gatsby or React, so I'm just following his footsteps 😄

content/docs/sidebar.json
iesahin: 1. I agree with this. I tried to make minimal changes to file naming as I think that was a decided matter but we can change all to non-`dvc` versions IMO. 
2. You're also right on this but tooltips in the branch is using frontmatter and there can only be a single frontmatter for each page. We can either have a _dependency_ tooltip/glossary and tell external dependency when the user visits the concept page, or all glosses should have separate pages. IMHO **glossary/tooltips and concepts should be separated** and we need to have 5-10 concept pages and put tooltips/glossary in a single file. 
3. This also needs separation of glossary and basic concepts. Current structure requires single page, short documents for concepts which has tooltips defined at the top.

content/docs/user-guide/concepts/data-pipelines.md
iesahin: Writing a document that puts _all_ these words would be a challange 😄 

Instead there can be a `keywords:` front matter that will be copied to meta tags of pages. Talking to search engine bots and talking to humans are different purposes IMO. These keywords can also be added to other pages.

content/docs/user-guide/concepts/data-pipelines.md
iesahin: I'm not against using duplicate material as soon as it fits to the context. Concepts pages should be revised to contain less details but overall removing material from other parts of the documentation may not be necessary. 

Users will be reading these sporadically and they probably will forget what they read in concepts when they are reading the command ref. Aesthetically it may not be pleasing to repeat ourselves but from POV of a user, these are separate instances of reading. Also we can't be sure the user will ever read the other if they read one of them.

shcheklein:
@iesahin 

> I fully agree that this should be split into smaller PRs but this is a breaking change. JS part is actually copied from what @rogermparent did in ug-basic-concepts here

makes sense! Can we make them as a separate PR still? May be just migrate existing tooltips to the new format? I'm happy to merge it with a content that just copies existing tooltips. Then we can proceed step by step improving it. WDYT?

content/docs/sidebar.json
jorgeorpinel: On 1. We're moving all the files to another directory I think so no problem in renaming them along the way, I think.
On 2. I think we can have empty concept md docs that only serve for tooltip purposes, but not link them in sidebar.js right?
On 3. True. For now we can pick which concept doc is more important and put other concepts in there (even if they have their own concept file that's only used for tooltips. Anyway, no need to worry about this for the first or 2nd PR probably...

package.json
jorgeorpinel: Ah yes np. I remember what this is about: the tooltips will come from the frontmatter instead of the md content.

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: Search engines don't really care about metadata keywords anymore, we need to incorporate the keywords to the content but yeah no need to have them all necessarily, especially in small concept pages that will be impossible.

content/docs/sidebar.json
jorgeorpinel: p.s. the tooltip vs concept page separation part we can worry about when looking into #1395

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: > removing material from other parts of the documentation may not be necessary

That's actually an important goal for this task (see https://github.com/iterative/dvc.org/issues/550#issuecomment-673585050). Some (a little) redundancy is OK but there should be a single source of full/comprehensive explanation for these concepts. Keep in mind that maintaining duplicated content is hard and error-prone 😅 (just like code)

> Users will be reading these sporadically and they probably will forget

TBH I don't expect users to navigate to concepts much it at all (from nav). Like you suggested without context they're pretty hard to care about or understand. They'll probably only open them from links we place in other docs, and from the tooltips of course.

shcheklein: It would be great though for this section to serve as an entry point. May be index page can give an overview of the most important ones? how do hey relate to each other.

jorgeorpinel:
> this is a breaking change
If we don't include JS changes to this, Basic Concepts don't appear in the sidebar and tooltips become large documents

@iesahin you can just leave all the json and js changes EXCEPT for sidebar.js AND move the existing md files (without changes) from `basic-concepts/` to `concepts/`  for the first PR. That will preserve the existing tooltip functionality and enable future adding concept pages to the nav in future PRs (I tried it locally and it seems to work). 👍 

That's what I was referring to with "splitting the PR".

jorgeorpinel:
> move the existing md files (without changes

No sorry that's not accurate. The only change would be to leave the new frontmatter (with `tooltip` fields) but the md content of all those pages would be empty initially.

content/docs/user-guide/concepts/data-pipelines.md
jorgeorpinel: Good idea. I agree about an guiding index page for those who do navigate in here. BUT probably the 1st PR (after this one) will only introduce 1 or a few concepts and the index may not make sense until we have most of them up.

iesahin:
It will be an interesting cherry-picking experience :)

I'll send you another PR by cherry picking the tooltip on the front matter functionality but without the concept pages in sidebar. That's the deal I think.

iesahin:
Also I'll try to fix prettier by changing its config because it's good test.

iesahin:
I split JS parts, didn't touch current text in `master` but moved them to `tooltip:` front matter. Submitted #2192 for this. We can close this.

jorgeorpinel:
Thanks @iesahin !

PR #2188, opened 2021-02-12
Fix Data Access mentions tracking models without clarifying what kind #2096

### LINKED ISSUE #2096: start: Data Access mentions tracking models without clarifying what kind
I was reading the documentation on [Data Access](https://github.com/iterative/dvc.org/blob/master/content/docs/start/data-access.md). It starts by mentioning that 
```quote
Okay, now that we've learned how to track data and models with DVC and how to version them with Git, next question is how can we use these artifacts outside of the project?
```
However, the previous tutorial [Get Started: Data Versioning](https://github.com/iterative/dvc.org/blob/master/content/docs/start/data-versioning.md) only shows how to track data and tracking models are not demonstrated.

iesahin:
Fixes #2096

### DIFF
--- a/content/docs/start/data-access.md
+++ b/content/docs/start/data-access.md
@@ -4,10 +4,13 @@ title: 'Get Started: Data Access'
 
 # Get Started: Data Access
 
-Okay, now that we've learned how to _track_ data and models with DVC and how to
-version them with Git, next question is how can we _use_ these artifacts outside
-of the project? How do I download a model to deploy it? How do I download a
-specific version of a model? How do I reuse datasets across different projects?
+Okay, now that we've learned how to _track_ data files in DVC and how to version
+them with Git. _Models_ in a machine learning project are also files written and
+read by programs and DVC can track and version them similar to data files.
+
+Next question is how can we _use_ these artifacts outside of the project? How do
+I download a model to deploy it? How do I download a specific version of a
+model? How do I reuse datasets across different projects?
 
 > These questions tend to come up when you browse the files that DVC saves to
 > remote storage, e.g.
--- a/.restyled.yaml
+++ b/.restyled.yaml
@@ -1,6 +1,17 @@
 restylers:
   - name: prettier
-    arguments: ['--write']
+    arguments:
+      [
+        '--write',
+        '--trailing-comma=none',
+        '--no-semi',
+        '--arrow-parens=avoid',
+        '--single-quote=true',
+        '--print-width=80',
+        '--tab-width=2',
+        '--use-tabs=false',
+        '--prose-wrap=always',
+      ]
     include:
       - './*.{js,md}'
       - 'pages/**/*.js'
--- a/content/docs/start/data-versioning.md
+++ b/content/docs/start/data-versioning.md
@@ -1,6 +1,6 @@
 ---
-title: 'Get Started: Data Versioning'
-description: 'Get started with data versioning in DVC. Learn how to use a
+title: 'Get Started: Data and Model Versioning'
+description: 'Get started with data and model versioning in DVC. Learn how to use a
 regular Git workflow for datasets and ML models, without storing large files in
 Git.'
 ---
@@ -246,6 +246,16 @@ Yes, DVC is technically not even a version control system! `.dvc` files content
 defines data file versions. Git itself provides the version control. DVC in turn
 creates these `.dvc` files, updates them, and synchronizes DVC-tracked data in
 the <abbr>workspace</abbr> efficiently to match them.
+
+## Model versioning
+
+Apart from data files, DVC eases the way you work with models. Models in a
+project usually change more frequently than data files and they need to be kept
+in sync with changes in other elements of a project. Model files are no
+different than data files when it comes to tracking their versions. DVC also
+provides means to track minor changes in model files without fully checking in
+to underlying VCS. In later sections of this series, you'll see how DVC enables
+to track changes in pipelines consisting of multiple model and data files.
 
 ## Large datasets versioning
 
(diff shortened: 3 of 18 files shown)


shcheklein:
Update: `added Fixes #2096` to the description - GH will close it automatically.

content/docs/start/data-access.md
```diff
@@ -4,10 +4,13 @@ title: 'Get Started: Data Access'
 
 # Get Started: Data Access
 
-Okay, now that we've learned how to _track_ data and models with DVC and how to
-version them with Git, next question is how can we _use_ these artifacts outside
-of the project? How do I download a model to deploy it? How do I download a
-specific version of a model? How do I reuse datasets across different projects?
+Okay, now that we've learned how to _track_ data files in DVC and how to version
+them with Git. _Models_ in a machine learning project are also files written and
```
shcheklein, line 8: I like the change, but it feels a bit misplaced - it splits the previously written paragraph which was written as a single piece. It feels like the whole section will be about models after these two sentences, but then we go with artifacts, and go with `How do I reuse datasets across different projects?`.

Also, I'm not sure it addressed the issue https://github.com/iterative/dvc.org/issues/2096 - it feel that we should put your sentence somewhere earlier? wdyt?

I'm even thinking if it worth creating a separate section Models Versioning ? To be extra explicit, to optimize SEO. Otherwise it'll be a constant battle that people read Data Versioning and do not connect dots (expected).

shcheklein: Btw, section itself can be super simple first. That explains that models are the same as data for us pretty much.

jorgeorpinel: Another alternative is to change https://dvc.org/doc/start/data-versioning instead to avoid the confusion in the first place. I'm thinking specifically around "few commands that you can run along with git to track large files, directories, or ML models".

shcheklein: >  I'm thinking specifically around "few commands that you can run along with git to track large files, directories, or ML models".

Unfortunately, I think it won't be enough. It's hard to beat the "Data Versioning" title.

iesahin: I agree that another document about model versioning between *Data Versioning* and *Data Access* might be better. It can be an overview of changing runtime parameters for models through `params.yaml` or run-cache. These could be a bit complex for a *Get Started* document but a simple *we put these parameters here and run this script and create these different models* kind of introduction can be written I think.

But usually people use already existing models at the beginning. We may need to discuss *access* first and *versioning* later.

Maybe we can just remove models from these documents and focus on data. *Data Versioning ➡️  Data Access ➡️ Model Versioning* can be another path. 

We can also discuss *Model Versioning* as a section at the end of Data Versioning for now. 

I'm fine with any of these solutions. We can continue to discuss this in #2096

BTW mentioning just the issue number (without "fix", "fixes") in the title should be OK for automated closing of issues. (At least that's what VS Code did while sending the PR 😃 ) I'll put "fixes" for clarity though.

iesahin: While reading #2195 I thought a way without much overhaul is adding _Model_ to _Data Versioning_ title and make it _Data and Model Versioning_ and describe that model versioning in a section. What I had in mind is already described in later sections of Getting Started.

jorgeorpinel: My take on this issue is that the term "model" is sprinkled in for SEO purposes but we weren't actually treating models different to a data file. That seemed to cause confusion for a user so we're trying to clarify the language and the easiest path is my suggestion + the page title change for now.

If we want to re-do the GS to have more dedicated model versioning material we should probably open an issue for that and wait until after #2195 which has some overlap (repurposes the **Experiments** page for ML-specific tools, and could also focus on model versioning).

shcheklein: Sounds good to me, @iesahin @jorgeorpinel !

shcheklein: should we also rename Data Access?

jorgeorpinel: "Access Data and Models"? Yes perhaps. Or "Access Artifacts" perhaps

shcheklein: Artifacts - we'll have to define what artifact is then, unfortunately. Feels like we should not be overoptimizing this and do the simplest, the most explicit thing.

iesahin:
Let me revise the Versioning doc and submit to this branch.

iesahin:
**Question**: Can we make URL redirects from `/doc/start/data-versioning` to `/doc/start/data-and-model-versioning` or should we keep URLs intact? @jorgeorpinel @shcheklein

jorgeorpinel:
> Can we make URL redirects from /doc/start/data-versioning to /doc/start/data-and-model-versioning

Good question. Usually we change the URL with the title but there are a few places where they don't match exactly (e.g. /start = Get Started). Maybe in this case we should keep /data-versioning as the new title still sort of matches that and dealing with the redirects can be painful later.

shcheklein:
@jorgeorpinel in this specific case, I think "model" should be part of the URL for SEO? (get started vs start - is not that important). We can remove some old redirects probably at this point if it gets too complicated.

shcheklein:
@iesahin looks like some unrelated commits got there?

jorgeorpinel:
> "model" should be part of the URL for SEO

I think the title and content already make proper use of the term "model" so I doubt the URL matters much. But TBH Idk, if you prefer to be safe then sure, we can add a redirect (please see /redirects-list.json @iesahin).

iesahin:
I think I've messed up with VS Code's GitHub extension to apply some unrelated changes to the branch while merging or whatever. 🤦🏼  Restarting this on a new PR seems to be the easiest.

content/docs/start/data-access.md
iesahin: This should be done in 8dea9639424e3928ec83bb56716ae2dee022d873 (in the new PR #2214 )

Thanks.

iesahin:
This PR is closed in favor of #2214. Will continue to discuss there.

PR #2196, opened 2021-02-15
Fixes #2153: mismatch in default permissions for newly created or downloaded cache files.

### LINKED ISSUE #2153: mismatch in default permissions for newly created or downloaded cache files
In https://github.com/iterative/dvc.org/blame/b2149f1bb09cad948c420733313343eeece4e2ba/content/docs/command-reference/config.md#L176, the octal permissions `0o664` don't match the human-readable permissions `rw-r--r--`. It seems like `0o664` should be change to `0o0644`.

However, it seems like the defaults depend on the user's umask, and the docs assume a `0o022` umask. Would it be better to state something like the default permissions will depend on the umask?

@efiop

iesahin:
This PR removes incorrect statement about default permissions. 

Fixes #2153 

### TODO

- [ ] Resolve whether we need to mention default permissions


content/docs/command-reference/config.md
```diff
@@ -179,11 +179,11 @@ This section contains the following options, which affect the project's
   > set.
 
 - `cache.shared` - permissions for newly created or downloaded cache files and
-  directories. The default is `0o664`(rw-r--r--) for files and `0o755`
-  (rwxr-xr-x) for directories. The only accepted value right now is `group`,
-  which makes DVC use `0o664` (rw-rw-r--) for files and `0o775` (rwxrwxr-x) for
-  directories, which is useful when you are using a a
-  [shared development server](/doc/use-cases/shared-development-server).
+  directories. The only accepted value right now is `group`, which makes DVC use
+  `0o664` (rw-rw-r--) for files and `0o775` (rwxrwxr-x) for directories, which
```
jorgeorpinel, line 183: Do we need the `0o` octal notation? Most people who know about UNIX file permissions will be familiar with plain `664`/`775` probably?

jorgeorpinel: Also I wonder what this does in Windows... @efiop do you know? Thanks

content/docs/command-reference/config.md
```diff
@@ -179,11 +179,11 @@ This section contains the following options, which affect the project's
...
+  is useful when you are using a a
+  [shared development server](/doc/use-cases/shared-development-server). (The
+  default permissions for cache files is system dependent.)
```
jorgeorpinel, line 186: > (The default permissions for cache files is system dependent.)

No need for parentheses I think. But what are the defaults? (`664/755` also seem system-dependent.)

jorgeorpinel:
I see a commit in this branch will close the original issue #2153 when merged, but please link to the PR directly in the description for better visibility. Thanks

content/docs/command-reference/config.md
iesahin: I also don't think we need `0o` at the beginning.

content/docs/command-reference/config.md
iesahin: Other options have parenthesis while mentioning defaults, so I put to this but both ways are fine with me. 

The defaults are determined by `umask` and `acl` in Linux. (`man umask` has a description and example.) I think it's a distraction at that point in the command reference to describe the default behavior in full detail. If DVC doesn't have a custom cache permissions set, we can remove the part even in the parenthesis.

iesahin:
> I see a commit in this branch will close the original issue #2153 when merged, but please link to the PR directly in the description for better visibility. Thanks

Description of the original issue or description of this PR?

content/docs/command-reference/config.md
iesahin: From the [code](https://github.com/iesahin/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/dvc/cache/local.py#L33), it looks the permissions are set by `umask`.

[Here](https://github.com/iesahin/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/dvc/utils/fs.py#L16) it seems that `umask` is set to system default. (What does this mean in Windows, I'm not sure.)

content/docs/command-reference/config.md
efiop: Don't know about windows, python only operates with user read-only bit, so unix perms are almost meaningless. As well as shared development server scenario is probably meaningless, it was never tested on windows. Let's just keep the *nix perms for now.

efiop (approved):

shcheklein:
@iesahin also, some unrelated changes in this PR?

content/docs/command-reference/config.md
jorgeorpinel:
```suggestion
  `664` (`rw-rw-r--`) for files and `775` (`rwxrwxr-x`) for directories, which
```

content/docs/command-reference/config.md
```diff
@@ -179,11 +179,11 @@ This section contains the following options, which affect the project's
...
 
 - `cache.shared` - permissions for newly created or downloaded cache files and
-  directories. The default is `0o664`(rw-r--r--) for files and `0o755`
-  (rwxr-xr-x) for directories. The only accepted value right now is `group`,
-  which makes DVC use `0o664` (rw-rw-r--) for files and `0o775` (rwxrwxr-x) for
-  directories, which is useful when you are using a a
-  [shared development server](/doc/use-cases/shared-development-server).
+  directories. The only accepted value right now is `group`, which makes DVC use
+  `0o664` (rw-rw-r--) for files and `0o775` (rwxrwxr-x) for directories, which
+  is useful when you are using a a
```
jorgeorpinel:
```suggestion
  is useful when using a
```

content/docs/command-reference/config.md
jorgeorpinel: No need for parentheses. It's kind of strange to have a full sentence in parenthesis anyway (I probably wrote that so my bad).

But mostly I'm still confused what the defaults are. Isn't `group` the only accepted value? If you don't set `cache.shared`, what happens? Is it even meaningful to talk about defaults for this config option? Thanks

jorgeorpinel: p.s. `umask = umask = os.umask(0) = 0o000` per https://github.com/iesahin/dvc/blob/08300d3bcf51f7bd9c8509924110416461b11b60/dvc/utils/fs.py#L16

So I guess `0o666 & ~umask = 0o666 & 0x111` always = 0c666? (always 777 for dirs)

iesahin:
> @iesahin also, some unrelated changes in this PR? 

I'm merging changes in `master` to my PRs daily to keep them fresh. :) @shcheklein

content/docs/command-reference/config.md
iesahin: [`umask`](https://docs.python.org/3/library/os.html#os.umask) is a strange function. It returns the current `umask` when you set it to something but there is no other way to learn current `umask.` So the code sets `umask` to 0 to get the current value, then set `umask` to the previously obtained value. 

The code uses it to change permissions in [`chmod`](https://docs.python.org/3/library/os.html#os.chmod) but as @efiop said, Windows doesn't support this so it doesn't matter actually. 

For Linux/Unix, it looks `chmod` is able to modify permissions arbitrarily. e.g. When `umask` was `777`, I was able to set a file's permissions to `r--/---/---` with `os.chmod("file", 0o400)`

content/docs/command-reference/config.md
iesahin: Fixed in 1c49516 and 67c4626. Thanks.

shcheklein:
@iesahin hmm, why do I see these changes when I open https://github.com/iterative/dvc.org/pull/2196/files ?

content/docs/command-reference/config.md
jorgeorpinel: OK. Resolving this long review and reopening a short one in the latest text...

content/docs/command-reference/config.md
```diff
@@ -179,11 +179,11 @@ This section contains the following options, which affect the project's
...
-  directories. The default is `0o664`(rw-r--r--) for files and `0o755`
-  (rwxr-xr-x) for directories. The only accepted value right now is `group`,
-  which makes DVC use `0o664` (rw-rw-r--) for files and `0o775` (rwxrwxr-x) for
-  directories, which is useful when you are using a a
-  [shared development server](/doc/use-cases/shared-development-server).
+  directories. The only accepted value right now is `group`, which makes DVC use
+  `664` (rw-rw-r--) for files and `775` (rwxrwxr-x) for directories. This is
+  useful when using a
+  [shared development server](/doc/use-cases/shared-development-server). The
+  default permissions for cache files is system dependent.
```
jorgeorpinel, line 186: > Continues https://github.com/iterative/dvc.org/pull/2196#pullrequestreview-590249338

So are the `cache.shared` defaults always 666 and 777 in *NIX?

Whatever this is I think we need to be a bit more specific about it if we're going to mention defaults. Otherwise maybe just don't mention them.

iesahin:
I think I should not leave issue / branch tracking to VS Code. I'll do it manually from cli or fugitive now on. A single click changes the history in VS Code. Sorry.

iesahin:
Actually my `git log master..iesahin/issue2153` looks like this. 

Screen Shot 2021-02-18 at 14 36 09

Confused again.

content/docs/command-reference/config.md
iesahin: I think I can remove that sentence about the defaults but don't have a Windows machine to test. 😞 

In current `master`, the lines that change the permissions according to are [here](https://github.com/iterative/dvc/blob/9909ee85a1f61491a959962477208291e38f8f24/dvc/objects/db/local.py#L30)

(It's changed from the previous location and the class name is also changed.)

Screen Shot 2021-02-18 at 14 57 45

From what I see here a default `umask` of `022` will change the _default_ permissions to `644` and `755` respectively. 

When `cache.shared` is set, the permissions are **always** `664` and `775` regardless of the `umask` value. 

If you think it's better I can tell these and also mention that by default Windows doesn't provide much means to change permissions. (as @efiop said above.)

iesahin: Could you please check the changes in 53f43831ab248f1f63ea1fbe377acabb92a38712 ? @jorgeorpinel 

I can revert it and remove the default permissions discussion completely but a detailed explanation seemed better in cmd ref. Thank you.

iesahin: I need your input here @jorgeorpinel . Thank you :)

iesahin: @jorgeorpinel could we merge this now? :)

shcheklein:
@iesahin can we cleanup this PR please? it has now 30 files updated, probably not what it should be?

iesahin:
Cleaned up in a new PR @shcheklein . Closed in favor of #2257.

content/docs/command-reference/config.md
jorgeorpinel: Sorry for the delay, reviewing #2257 now...

PR #2198, opened 2021-02-16
glossary: create "stage" entry

### LINKED ISSUE #1579: glossary: create "stage" entry

iesahin:
This adds a BC document for stage. It also contains `tooltip:` front matter to fix #1579

### TODO

- [x] Resolve whether a stage runs an experiment [here](https://github.com/iterative/dvc.org/pull/2198#discussion_r578206365)
- [x] Resolve whether to mention invalidation [here](https://github.com/iterative/dvc.org/pull/2198#discussion_r578206365)


iesahin:
This is also related with #550

content/docs/user-guide/basic-concepts/stage.md
```diff
@@ -0,0 +1,24 @@
+---
+name: 'Stage'
+match: ['stage', 'stages']
+tooltip: >-
+  A stage is a single unit of execution in a pipeline that produces some
+  (intermediate or final) output. [ð](/doc/user-guide/concepts/stage)
```
shcheklein, line 6: `ð`?

shcheklein: what is the current version of the tooltip? sorry, I'm lost a bit - have we migrated all of the to the new format or it's still in progress?

content/docs/user-guide/basic-concepts/stage.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# Stage
+
+A stage is a single unit of execution in a <abbr>pipeline</abbr> that takes some
```
shcheklein, line 11: Should we go a bit more informal here? To be extreme: "Stage is any command that you would usually run from CLI that usually reads some files and produces other files or models. It can be something as simple as `python train.py " ...

We want here to give people sense about these things, I think. Using regular language, simple examples. Something that would make it very easy to catch the idea ...

iesahin:
Like "a stage is a step in an experiment/pipeline that may import a file, runs an experiment, etc."?

content/docs/user-guide/basic-concepts/stage.md
iesahin: There is tooltip for stage as of now. That sign was a link with a text of 📖 (like [📖 ](https://github.com/iterative/dvc.org/pull/2198) ) but why was it corrupted, I don't get. 

Updating the text and removing icon.

content/docs/user-guide/basic-concepts/stage.md
iesahin: I think 1ea2da3 is better in that regard. Thank you.

shcheklein:
So, this is only for the tooltip right?

iesahin:
This contained a BC document for the stage but I deleted it after your suggestion to have rather an informal language. It doesn't change anything on the sidebar.

content/docs/user-guide/basic-concepts/stage.md
```diff
@@ -0,0 +1,8 @@
+---
+name: 'Stage'
+match: ['stage', 'stages']
+tooltip: >-
+  A single step in a pipeline which may import data, run experiments and produce
+  some (intermediate or final) output. Stages may depend other stages and they
+  can be automatically invalidated when their dependencies change.
```
jorgeorpinel, line 7: Good start. I'd look in https://dvc.org/doc/command-reference/run for the existing definition/explanation. Specific questions on this current text:
- Stages run experiments? I think not exactly
- Probably need to describe "invalidated"

iesahin: A user runs experiments _in stages_ and a stage may be completely about running an experiment, right? Am I missing something? Stages don't run experiment but a stage may only be about running an experiment. 

Will add a "user" there. 

Instead of invalidation, I can tell how `run` or `repro` skips some stages but succinctly using the "invalidation" seemed more appropriate.

jorgeorpinel: Experiments are broader than individual stages (see WIP https://dvc-org-guide-experimen-xigcev.herokuapp.com/doc/user-guide/experiment-management). For now I think there's no nee to complicate this tooltip with the concept of experiments.

> Instead of invalidation, I can tell how run or repro skips some stages

Check the parameter and pipelines file tooltips: they use `invalid (considered outdated)`

iesahin:
It seems `stage` now has a tooltip in `master` and entry in cmd ref. So I'm closing this.

iesahin:
This also fixed #1579

PR #2146, opened 2021-02-03
guide: Experiment Management

jorgeorpinel:
To support the new 2.0 features (`dvc exp`) per #2026. Closes #816

- New guide doc (main file to review): [content/docs/user-guide/experiment-management.md](https://github.com/iterative/dvc.org/pull/2146/files#diff-be6aa503d0cf32633b8a90ecca42b73c0059fd54c83ea20dc55406a9e549aaa8)
- Relevant links to ^ from existing docs
- Intros "experiment" tooltip
- Add `checkpoint` field to dvc.yaml outputs spec
- Update "run-cache" tooltip + some copy edits on that topic

---

Based on https://github.com/iterative/dvc/wiki/Experiments#usage, et al.

> May reuse some stuff from https://github.com/iterative/dvc.org/pull/828/files

PR #1926, opened 2020-11-11
[WIP] cmd: experiment

pmrowla:
See https://github.com/iterative/dvc/wiki/Experiments for ref. (may not be complete)

And https://github.com/iterative/dvc/pulls?q=is%3Apr+label%3Aexperiments+ for related PRs

PR #2195, opened 2021-02-15
Get started tutorial for experiments (with readable diff)

dberenbaum:
Creating a new PR and closing out #2191 due to diff readability issues. You will still need to go to the individual commit to see a clean diff of the old experiments section: https://github.com/iterative/dvc.org/commit/92e66cb204c70245d140e73753edc57f8209f316.

Fix #2189.

PR #2199, opened 2021-02-17
ref: exp cmd index, make_checkpoint() fn

jorgeorpinel:
> Splits and improves #1926

- [x] Change the base to `master` when #2146 is merged.

- Intros `exp` cmd ref (with index, no examples yet).
- Adds `make_checkpoint()` API ref. with fooish example.
- Adds `gc --all-experiments`.
- Uses more "experiments" tooltips.
- Other copy edits around term "experiments"
- Sets up autolink from `dvc experiments` to the /exp cmd ref — per https://github.com/iterative/dvc.org/pull/2199#pullrequestreview-592549850 (by Roger)


content/docs/api-reference/make_checkpoint.md
```diff
@@ -0,0 +1,83 @@
...
+[experiment](/doc/user-guide/experiment-management), you can write your code so
+it registers checkpoints with DVC during runtime. This function should be called
+by the code in stages executes by `dvc exp run` (see `cmd` field of `dvc.yaml`).
+
+> Note that for non-Python code, the alternative is to write a
+> `.dvc/tmp/DVC_CHECKPOINT` signal file.
+
+## Exceptions
+
+None
```
jorgeorpinel, line 32: Hi @pmrowla could you confirm this? Thanks

content/docs/api-reference/make_checkpoint.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Example: Mark a checkpoint every 100 iterations
+
+Let's consider the following `dvc.yaml` file:
+
+```yaml
+stages:
```
jorgeorpinel, line 39: Adapted from https://github.com/iterative/dvc/wiki/Experiments#checkpoint-experiments for now. Need a better example...

config/prismjs/dvc-commands.js
```diff
@@ -42,6 +42,8 @@ module.exports = [
   'gc',
   'freeze',
   'fetch',
+  'exp',
+  'experiments',
   'doctor',
```
jorgeorpinel, line 47: Unrelated to this PR but it looks like just adding these is enough for the engine to also link subcommands?

image
> https://dvc-org-ref-experiments-dv5pbs.herokuapp.com/doc/user-guide/experiment-management

So should we prune subcommands from that list, @rogermparent ? Thanks

rogermparent: Considering the links have the subcommands as fragments, this seems like an intended behavior that just happens to be taken to an extreme here because `dvc exp` uses a lot of subcommands.

Since the current behavior is at worst slightly busy and possibly even useful, the most I'd give to the issue of skipping subcommands is a lower-priority discussion ticket on the rules for auto-links. That logic is a little crufty as it is, so getting some input on a future refactor that unifies that little nest of code could be useful.

jorgeorpinel: OK. Reminds me of #1614 BTW. Maybe we can add feedback in there? Resolving this for now. Thanks

content/docs/api-reference/make_checkpoint.md
pmrowla: This is correct

jorgeorpinel: OK thanks

config/prismjs/dvc-commands.js
```diff
@@ -42,6 +42,8 @@ module.exports = [
   'gc',
   'freeze',
   'fetch',
+  'exp',
+  'experiments',
```
jorgeorpinel, line 46: Yet another Q on the engine @rogermparent: It seems `dvc experiments` isn't being auto-linked.

image
> https://dvc-org-ref-experiments-dv5pbs.herokuapp.com/doc/user-guide/project-structure/pipelines-files#output-subfields

Probably because there's no /doc/command-reference/experiments URL? How do we make it link to /doc/command-reference/exp ?

rogermparent: We can do this with a `simpleLinker` entry in `content/linked-terms.js`, I just tested it out to make sure that works, I'll push up a commit with the change.

jorgeorpinel: Ty!

jorgeorpinel: Works! Woot woot 🎉 Ty

content/linked-terms.js
```diff
@@ -18,5 +18,9 @@ module.exports = [
   {
     matches: '.dvc/',
     url: '/doc/user-guide/project-structure/internal-files'
+  },
+  {
+    matches: 'dvc experiments',
+    url: '/doc/command-reference/exp'
   }
```
jorgeorpinel, line 25: Unrelated: Maybe we should apply something like this to `doctor` and any other aliases to avoid duplicating entries in the cmd ref.

shcheklein (approved):

content/docs/api-reference/make_checkpoint.md
```diff
@@ -0,0 +1,126 @@
...
+            i_ = 0
+
+        with open("int.txt", "w") as fd:
+            fd.write(f"{i_}")
+
+        if i_ % 100 == 0:
+            make_checkpoint()
+
+    except KeyboardInterrupt:
+        exit()
```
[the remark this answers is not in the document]
jorgeorpinel: Good call, committing.

content/docs/api-reference/make_checkpoint.md
```diff
@@ -0,0 +1,126 @@
...
+│ │ ╓ exp-8a3bd │ 02:07 PM     │
+│ │ ╟ 75ff5e0   │ 01:54 PM     │
+│ │ ╟ 6f5009b   │ 01:54 PM     │
+│ ├─╨ d832784   │ 01:54 PM     │
+└───────────────┴──────────────┘
+# Press q to exit this screen.
+```
+
+Now if we use `dvc exp run` again, the process will start from 200. To restart
+from a previous point or even from scratch, you can use use `dvc exp apply`.
```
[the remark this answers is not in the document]
jorgeorpinel: I wasn't even aware of that option. I thought we had scratched exp restart and exp reset in favor for exp apply <first-exp> + exp run again. So should I ONLY mention run --reset? Added it for now but left also exp apply in there.

content/docs/command-reference/exp/index.md
```diff
@@ -0,0 +1,65 @@
...
+    pull          Pull an experiment from a Git remote.
+```
+
+## Description
+
+`dvc exp` subcommands provide specialized ways to create and manage data science
+experiments. The basic workflow goes like this:
+
+- Modify <abbr>dependencies</abbr> (e.g. input data or source code),
+  <abbr>hyperparameters</abbr>, or commands (`cmd` field of `dvc.yaml`) of
```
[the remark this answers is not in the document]
jorgeorpinel: Yes but isn't `dvc exp` pretty much designed for DS/ML?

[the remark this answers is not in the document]
jorgeorpinel: p.s. the tooltip may help:

image

content/docs/api-reference/make_checkpoint.md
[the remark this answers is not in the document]
dberenbaum: It was re-introduced as a shortcut when `resume` was dropped (https://github.com/iterative/dvc/pull/5351). The biggest difference is that `reset` also drops the existing checkpoints. Also, `dvc exp apply` followed by `dvc exp run` will show as a new experiment, while `reset` will replace the existing one.

```
The `--reset` option of that command drops the checkpoints and restarts the experiment from scratch. To start from any previous point, you can use `dvc exp apply`.
```

content/docs/command-reference/exp/index.md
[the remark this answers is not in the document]
dberenbaum: > Yes but isn't `dvc exp` pretty much designed for DS/ML?

Yup, and I understand the preference for `hyperparameters` because of that. I thought `parameters` might be more clear since this is a technical reference, not a user guide, and `parameters` is the term used elsewhere to describe the full scope of what can be modified, rather than expected usage. Also, looking at the getting started example, experiments are used to change the upstream featurization parameters, not only the model hyperparameters.

That being said, the tooltip does show that general description of parameters, so if you think it's more helpful to use `hyperparemeters`, I'm fine with it.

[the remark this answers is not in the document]
jorgeorpinel: Changed, PTAL @dberenbaum

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: I'll improve the example in a future iteration (maybe once @dberenbaum's #2195 is merged).

jorgeorpinel:
The list of subcommands in the index has [broken links](https://github.com/iterative/dvc.org/pull/2199/checks?check_run_id=1955926697) for now, but that should be fixed soon so I'll just commit this now...

PR #2200, opened 2021-02-17
ref: main exp workflow (run,show,diff,apply)

jorgeorpinel:
> Splits and improves #1926

- `exp run/show/diff/apply` Descs and Options
- related updates to the guides and other refs (index and API)

I left out the examples in this PR b/c it's getting too big.


content/docs/sidebar.json
```diff
@@ -217,7 +217,21 @@
       {
         "label": "experiments",
         "slug": "exp",
-        "source": "exp/index.md"
+        "source": "exp/index.md",
+        "children": [
+          {
+            "label": "exp run",
+            "slug": "run"
+          },
```
jorgeorpinel, line 225: Hiii @rogermparent another quick callout... Would it be possible to automatically label subcommands inside `children` as `parent children` to avoid all these `{}` structures (here and in other subcommand groups). Ty

rogermparent: Something like replacing `{ "label": "exp run", "slug": "run" }` with the string `"run"` having the same effect? That's a good idea, and totally doable!

jorgeorpinel: Great! I was just thinking to replace the whole `{}` block with `"run"` in this case (:

jorgeorpinel: Haha yeah Roger you got it exactly right the first time. Didn't read your answer correctly 😬

jorgeorpinel: Any hope we can get that done for this PR @rogermparent? No worries if not but let's make make an issue then? Thanks

content/docs/command-reference/exp/index.md
```diff
@@ -39,16 +41,17 @@ positional arguments:
...
-- Modify <abbr>dependencies</abbr> (e.g. input data or source code),
-  <abbr>parameters</abbr>, or commands (`cmd` field of `dvc.yaml`) of committed
-  stages.
-- Use `dvc exp run` (instead of `repro`) to execute the pipeline. This puts the
-  experiment's results in your <abbr>workspace</abbr>, and tracks it under the
-  hood.
-- Visualize experiment configurations and results with `dvc exp show`. Repeat.
+- Modify stage <abbr>parameters</abbr> or other dependencies (e.g. input data,
+  source code) of committed stages.
+- Use `dvc exp run` (instead of `repro`) to execute the pipeline. The results
```
jorgeorpinel, line 46: This list is repeated from content/docs/user-guide/experiment-management.md which may not be the best idea... For now should work — not sure what else to put here (maybe nothing).

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,97 @@
...
+without polluting it with unnecessary commits, branches, directories, etc.
+
+> `dvc exp run` is equivalent to `dvc repro` for <abbr>experiments</abbr>. It
+> has the same behavior when it comes to stage execution (restores the
+> dependency graph, etc.). It differs from in that experiment dependencies and
+> outputs will always be <abbr>cached</abbr> (so they can be restored later).
+
+Each `dvc exp run` creates a variation based on the last project version (Git
+commit) and stores it internally with an automatic experiment ID (a short SHA-1
+hash). The results of the last experiment can be seen in the
```
jorgeorpinel, line 35: > short SHA-1 hash

Is that accurate?

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,101 @@
...
+committing them to the Git repo. You can use `dvc exp apply` to roll back the
+workspace to a previous experiment first. Unnecessary ones can be removed with
+`dvc exp gc`, or abandoned (and their data with `dvc gc`).
+
+## Parallel execution (multi-threading)
+
+⚠️ Multi-thread runs are experimental and may be unstable. ⚠️
+
+When using `-j` to run experiments in parallel, take caution to only specify a
+number of jobs which can be handled by your local machine or execution
```
jorgeorpinel: What's parallelized into jobs exactly, if stages are not scheduled? Does -j only affect independent stages?

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,39 @@
...
+Provides a quick way to compare `dvc params` and `dvc metrics` between two
+`dvc experiments`. The differences shown by this command include the old and new
+values, and numeric difference (delta) from the previous ones.
+
+> This is similar to combining the reports from `dvc params diff` and
+> `dvc metrics diff` together, for the experiments in question.
+
+Without arguments, this command compares params and metrics currently present in
+the <abbr>workspace</abbr> (last `dvc exp run`) with the latest committed
+versions (required).
```
jorgeorpinel, line 31: Is this correct?

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,39 @@
...
+
+```usage
+usage: dvc experiments diff [-h] [-q | -v] [--all] [--show-json]
+                            [--show-md] [--old] [--no-path]
+                            [--precision <n>] [a_rev] [b_rev]
+
+positional arguments:
+  a_rev            Old experiment to compare (defaults to HEAD)
+  b_rev            New experiment to compare (defaults to the current
+                   workspace)
```
jorgeorpinel, line 17: BTW should these args be called something else? I guess technically they're Git revisions but that's a deep implementation detail...

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,39 @@
...
+> `dvc metrics diff` together, for the experiments in question.
+
+Without arguments, this command compares params and metrics currently present in
+the <abbr>workspace</abbr> (last `dvc exp run`) with the latest committed
+versions (required).
+
+`a_rev` and `b_rev` are optional experiment IDs (printed by `dvc exp run`) to
+compare.
+
+All params and metrics defined in `dvc.yaml` are used by default.
```
jorgeorpinel, line 36: Are params/metrics that only come up in one experiment shown?

jorgeorpinel (commented):
Questions for @pmrowla 🙏 — I'll try to figure these out myself if I can get to it before you answer.

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+# Description
+
+Provides a quick way to compare `dvc params` and `dvc metrics` between two
+`dvc experiments`. The differences shown by this command include the old and new
+values, and numeric difference (delta) from the previous ones.
+
+> This is similar to combining the reports from `dvc params diff` and
+> `dvc metrics diff` together, for the experiments in question.
```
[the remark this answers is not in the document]
jorgeorpinel: That's what https://github.com/iterative/dvc.org/pull/1926/files#diff-03846d2c3cbcbfacc654745454283a35e2adc2e5c3c4195ea2601ef19733363a says but those commands don't show the old value, just the new one + change, I think. Or have they been updated too?

content/docs/command-reference/exp/diff.md
jorgeorpinel: Also does this expect the entire short hashes (7-digit)? Or something like `exp-bfe64`? Or both? What about full SHA-1 hashes?

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,39 @@
...
+
+Without arguments, this command compares params and metrics currently present in
+the <abbr>workspace</abbr> (last `dvc exp run`) with the latest committed
+versions (required).
+
+`a_rev` and `b_rev` are optional experiment IDs (printed by `dvc exp run`) to
+compare.
```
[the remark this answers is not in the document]
jorgeorpinel: update: and it's not an exp hash. Get's ignored and the default behavior happens, I guess?

content/docs/command-reference/exp/diff.md
pmrowla: It accepts git revisions (as either short or full SHA hashes), git branch/tag names, or experiment names (whether it's `exp-1234` or a user specified name via `dvc exp run -n/--name ...`). All of the other diff commands in DVC also accept this same list now (including experiment names) - so `metrics diff`, `params diff`, etc

content/docs/command-reference/exp/diff.md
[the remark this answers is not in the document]
pmrowla: It looks like it's not exactly the same now.

* `metrics diff` currently shows old, new, change
* `params diff` currently shows old, new
* `exp diff` currently shows new, change

It should probably be exactly the same as combining metrics and params diff though (and metrics and params diff should probably also have consistent output as well)

content/docs/command-reference/exp/diff.md
pmrowla: Without arguments it just compares whatever is present in the workspace (which may or may not be an experiment run) against git HEAD, same as metrics/params diff

content/docs/command-reference/exp/diff.md
pmrowla: Not sure I understand this question, it will show all defined params/metrics from both `a_rev` and `b_rev`, the same way that it is handled in metrics/params diff

pmrowla: also, same as above, the target to compare doesn't have to be an experiment name, it can be any git revision or tag/branch

content/docs/command-reference/exp/run.md
pmrowla: Internally experiments are identified by a complete SHA-256 hash (similar to run cache), but I'm not sure we need to document the specifics of it here? The automatic `exp-abc123` naming does come from a shortened version of those hashes, but users can also specify their own names.

content/docs/command-reference/exp/run.md
pmrowla: The entire pipeline is what gets parallelized.

The existing behavior equivalent for `-j 2` would be cloning your git repo twice (into two separate folders/workspaces), and then doing `dvc repro ...` in each one

content/docs/command-reference/exp/diff.md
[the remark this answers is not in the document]
pmrowla: If only `a_rev` is specified, the current workspace is compared to whatever is in `a_rev` (I think this should also be consistent with metrics/params diff?)

Whether or not a_rev is an experiment name or any other git commit doesn't matter

content/docs/command-reference/exp/diff.md
pmrowla: it's probably also worth nothing that accepting experiment names isn't specific to the diff commands either, basically anything in DVC that takes a `--rev` (that previously worked with git branch/tag names) should also now accept experiment names as well, since they are resolved into a git SHA internally exactly the same way that git branch/tag names are resolved.

content/docs/command-reference/exp/run.md
jorgeorpinel: Ah good to know, thanks!

content/docs/command-reference/exp/run.md
jorgeorpinel: > cloning your git repo twice (into two separate folders/workspaces), and then doing dvc repro ... in each one

That sounds like duplicating the work. I'm still not getting this 🤔

content/docs/command-reference/exp/diff.md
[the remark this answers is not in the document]
jorgeorpinel: > should probably be exactly the same

Agreed. For now I'll assume they're all Old, New, Change for now. Extracted to iterative/dvc/issues/5515

content/docs/command-reference/exp/diff.md
jorgeorpinel: OK so this seems correct ✔️

I'm not sure about that behavior though, What about comparing the last experiment with its base by default? Cc @shcheklein 👋

content/docs/command-reference/exp/diff.md
jorgeorpinel: Sounds good. I'd still call them `a_exp` and `b_ext` but it's prob. also alright to leave them like this for consistency with existing `diff` commands.

> anything in DVC that takes a (local) --rev (and already worked with git branch/tag names) should also now accept experiment names as well

Interesting! So the names are Git labels in the `exps`? Good to know but indeed no need to doc that for now prob.

content/docs/command-reference/exp/diff.md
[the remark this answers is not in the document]
jorgeorpinel: Cool. Again, to me it would make more sense to have the `exp` baseline as the default `b_rev` in general.

content/docs/command-reference/exp/diff.md
jorgeorpinel: Sorry, I wasn't clear. I meant if a param/metric is defined in one of the revs but not in the other what happens? I guess it behaves like the each `diffs` where `-` is displayed in the missing one (and no change of course). Prob. no need to document it anyway, resolving!

content/docs/api-reference/make_checkpoint.md
```diff
@@ -20,9 +20,9 @@ while True:
 ## Description
 
 To track successive steps in a longer <abbr>experiment</abbr>, you can write
-your code so it registers checkpoints with DVC during runtime. This function
-should be called by the code in stages executes by `dvc exp run` (see `cmd`
-field of `dvc.yaml`).
+your code so it registers checkpoints with DVC during runtime (similar to a
+logger). This function can be called by the code in stages executes by
```
dberenbaum, line 24:
```suggestion
logger). This function can be called by the code in stages executed by
```

content/docs/api-reference/make_checkpoint.md
```diff
@@ -20,9 +20,9 @@ while True:
...
+`dvc exp run` (see also the `cmd` field of `dvc.yaml`).
```
dberenbaum, line 25: Is this supposed to refer to the `checkpoint` subfield of `output`? I'm unclear how `cmd` is related. Does it make sense to include an example `dvc.yaml`?

content/docs/api-reference/make_checkpoint.md
```diff
@@ -86,8 +86,8 @@ Experiment results have been applied to your workspace.
 
 In this example we kill the process (with Ctrl + C) after 3 checkpoints (at 0,
 100, and 200). The <abbr>cache</abbr> will contain those 3 versions of
-`int.txt`. And DVC applies the last checkpoint to the <abbr>workspace</abbr>
-even when more cycles happened before the interrupt:
+`int.txt`. DVC applies the last checkpoint to the <abbr>workspace</abbr> even
+when more cycles happened before the interrupt:
```
dberenbaum, line 90:
```suggestion
`int.txt`. DVC applies the last checkpoint to the <abbr>workspace</abbr>
regardless of how many cycles happened before the interrupt:
```
I don't understand the current sentence, so I'm not sure if this captures the intended meaning.

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,108 @@
...
+## Description
+
+Provides a way to execute and track `dvc experiments` in your
+<abbr>project</abbr> without polluting it with unnecessary commits, branches,
+directories, etc.
+
+> `dvc exp run` is equivalent to `dvc repro` for <abbr>experiments</abbr>. It
+> has the same behavior when it comes to stage execution (restores the
+> dependency graph, etc.). It differs from in that experiment dependencies and
+> outputs will always be <abbr>cached</abbr> (so they can be restored later).
```
dberenbaum, line 31: I don't understand this sentence. Are `dvc repro` dependencies and outputs not cached?

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,108 @@
...
+
+Each `dvc exp run` creates a variation based on the last project version (Git
+commit) and stores it internally with an automatic experiment name like
```
dberenbaum, line 34: Suggestion to drop "project version" as it might just add confusion.

```suggestion
Each `dvc exp run` creates a variation based on the last Git
commit and stores it internally with an automatic experiment name like
```

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,108 @@
...
+
+> In addition to the following options, `dvc exp run` also accepts the same
+> options as `dvc repro`, with the exception that `--no-commit` has no effect
+> here.
+
+- `--params [<filename>:]<params_list>]` - reproduce an experiment using the
+  specified [parameter](/doc/command-reference/params) values. Accepts a
+  comma-separated list of key-value pairs in the form
+  `[filename:]key1=val1,key2=val2...`. Any specified values will override those
+  from the contents of the relevant params file.
```
dberenbaum, line 80: Feel free to edit the description, but this option name and format changed.

```suggestion
`-S [<filename>:]<param_name>=<param_value>, --set-param [<filename>:]<param_name>=<param_value>` - 
Use the specified param value when reproducing pipelines.
```

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,86 @@
...
+                    [--exclude-params <params_list>]
+                    [--sort-by <metric/param>] [--sort-order {asc,desc}]
+                    [--no-timestamp] [--sha] [--show-json]
+                    [--precision <n>]
+```
+
+## Description
+
+Enters a screen that shows all of the relevant experiments you've run. Typically
+you can exit this screen by typing `q` in your keyboard.
```
dberenbaum, line 23:
```suggestion
Shows all of the relevant experiments you've run. By default, enters a screen that
you can typically exit by typing `q` in your keyboard (or use `--no-pager` to print
output in your terminal without entering this screen).
```
This probably needs some work. The point is that we mention the default option here using very plain language without referring to the `--no-pager` option, and the `--no-pager` description doesn't use similar plain language.

dberenbaum (commented):
Hope you don't mind that I left a few comments here!

content/docs/api-reference/make_checkpoint.md
```diff
@@ -20,9 +20,9 @@ while True:
...
 To track successive steps in a longer <abbr>experiment</abbr>, you can write
-your code so it registers checkpoints with DVC during runtime. This function
-should be called by the code in stages executes by `dvc exp run` (see `cmd`
-field of `dvc.yaml`).
+your code so it registers checkpoints with DVC during runtime (similar to a
+logger). This function can be called by the code in stages executes by
+`dvc exp run` (see also the `cmd` field of `dvc.yaml`).
 
 > Note that for non-Python code, the alternative is to write a
 > `.dvc/tmp/DVC_CHECKPOINT` signal file.
```
jorgeorpinel, line 28: Another QQ @pmrowla: How exactly does this work? You write that file (empty I suppose) and DVC picks it up and deletes it? How can you write one per process for `run -j`? Thanks

jorgeorpinel: Or one per output

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: No, I meant to mention `cmd` but yeah it's not clear, thanks. Updated

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: Changing to "if more cycles happened after that" but this example is kind of a placeholder for now. I'll prob steal/adapt your GS examples for the other docs in a following iteration 😬 Thanks though!

content/docs/command-reference/exp/run.md
jorgeorpinel: Oops "from" wasn't needed there, typo. "It differs in that dependencies and outputs will always be cached". But yeah wait... I need to re-think that (taken [from Peter's original PR](https://github.com/iterative/dvc.org/pull/1926/files#diff-162526d7eb1876fbb25840504db2c51860ae3c1b7af5b5f6431b0828c827189fR35-R36))... ⌛

jorgeorpinel: I think this refers to the fact that there's no `--no-commit` option, like in `repro`. But that's already mentioned in the **Options** section so I'll remove it from here 👍

content/docs/command-reference/exp/run.md
jorgeorpinel: How about "latest project version committed to Git"? I know it's long but the project version concept is more important I think, (motivation), while Git commit part is a technical requirement also worth mentioning.

content/docs/command-reference/exp/run.md
jorgeorpinel: Oh thanks ya I still need to double check all the options 😱

content/docs/command-reference/exp/show.md
jorgeorpinel: Thanks! Def better text, Using something very similar to this now. I'll review all the options text next (hopefully that finishes this iteration).

content/docs/command-reference/exp/run.md
dberenbaum: Sure, I have no strong feelings on this, so feel free to leave as is or make whatever change you like.

content/docs/command-reference/exp/run.md
pmrowla: It would be duplicating the work if the user queued two identical experiments, but the main use case is to queue up multiple experiments with different parameter values (or different versions of code/dependencies/etc) and then run those different experiments simultaneously. If you were going to do this in a single DVC/git clone/workspace with `dvc repro`, you would have to run all of your experiments one at a time sequentially.

pmrowla: Also, this should probably not be documented as `multithreading`. I would consider just using `parallel runs` everywhere in place of multithreading.

The original implementation was to do runs in separate threads, but the current implementation is via multiprocessing - so each run is in its own process, but this is also an internal implementation detail that is subject to change in the future.

content/docs/command-reference/exp/diff.md
pmrowla: exp names are just git refs, same as git branch names or git tags. The only difference is that they are not automatically resolved by CLI git, since CLI git only automatically resolves shortened refs for `refs/heads/<name>` (branches) and `refs/tags/<name>`.

content/docs/api-reference/make_checkpoint.md
pmrowla: Parallel runs with `-j` are run in their own temporary workspaces, so if I have my regular DVC/git workspace at `~/dvc-project`, and then do `exp run -j 2`, we create copies of the project at `~/dvc-project/.dvc/tmp/exp1` and `~/dvc-project/.dvc/tmp/exp2`. These copies are their own independent DVC workspaces (with their own `.dvc/` directories), but they are not full git clones - the temporary workspaces only start with the single experiment commit to reproduce rather than the full git history.

Essentially, the way execution works is that we `cd` into `~/dvc-project/.dvc/tmp/exp1` and then run `dvc repro`. So for generating the checkpoints for experiment one, the full signal path is `~/dvc-project/.dvc/tmp/exp1/.dvc/tmp/DVC_CHECKPOINT` (which is independent of the second experiment's signal path inside `~/dvc-project/.dvc/tmp/exp2/...`)

The user's code (possibly via `dvc.api.make_checkpoint()`) should be generating the signal file by first checking the `$DVC_ROOT` environment variable (so the signal path is `$DVC_ROOT/tmp/DVC_CHECKPOINT`. `DVC_ROOT` points to the location of `.dvc/`. For a temp directory experiment run, it will point to the `.dvc` directory for that specific temporary run, and not the `.dvc` directory for the "main" repo.

pmrowla: Also the checkpoint generation & signaling has nothing to do with individual outputs. When the user's code signals to make a checkpoint, we save the state of the entire repo (including any and all checkpoint outputs).

pmrowla: And again, this is all implementation detail that the user should not care about, all they need to know is that if they aren't using `dvc.api.make_checkpoint`, they just have to create an empty file at `$DVC_ROOT/tmp/DVC_CHECKPOINT` to do the signaling.

pmrowla: I think the signaling process for non-python code does need to be more thoroughly documented somewhere though.
They should be directed to refer to the actual `api.make_checkpoint()` implementation as well, but the way it should work is:

1. User's code does what it needs to and eventually reaches a point where they want to generate a checkpoint commit
2. User checks the `DVC_ROOT` environment variable, and then creates a file `$DVC_ROOT/tmp/DVC_CHECKPOINT` (it can contain anything or be an empty file, DVC only cares that the file exists)
3. User's code should block as long as the file `$DVC_ROOT/tmp/DVC_CHECKPOINT` still exists.
    * Making this a blocking call is important - the user's code should explicitly not do anything that writes to/modifies the repo/workspace state, as DVC will be staging files and creating commits in git during this time
    * The user should also ensure that the file handle used to open the signal file in step 2 has been closed (on windows if the file handle is still open, DVC may not be able to remove the file later, which is the signal for the user's code to continue)
4. When the file `$DVC_ROOT/tmp/DVC_CHECKPOINT` no longer exists (because DVC has removed it), this is the signal that the user's code can resume. Go back to step 1 and repeat checkpoint generation as needed

content/docs/command-reference/exp/run.md
jorgeorpinel: Already updated 👍

content/docs/command-reference/exp/run.md
jorgeorpinel: > -j/--jobs is only applicable for exp run --run-all

Got it. So could it maybe be an optional `<number>` argument to `--run-all` instead of a separate flag?

> this should probably not be documented as multithreading... the current implementation is via multiprocessing

Got it, thanks. I changed the language.

pmrowla: > So could it maybe be an optional <number> argument to --run-all instead of a separate flag?

Maybe? Using `--jobs` seems more consistent with other DVC commands to me.

jorgeorpinel: It does seem that way.

content/docs/command-reference/exp/run.md
jorgeorpinel: Forgot to let you know that all the options have been reviewed now @dberenbaum

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: Interesting @pmrowla! (BTW I ask because I need to know so thanks for the answers). I've put the relevant details in the refs now (see 53f02ab2). Some questions though:

> copies are their own independent DVC workspaces (with their own .dvc/ directories

Including a copy of the cache? (I guess not)

> the checkpoint generation & signaling has nothing to do with individual outputs

Yeah I thought that was strange. But the `checkpoint: true` field is under `outs` in dvc.yaml, apparently. Why not under the stage directly?

Why do you need that field at all? Curious

> the signaling process for non-python code does need to be more thoroughly documented somewhere

- [ ] Agreed

> 2. User checks DVC_ROOT and creates a file $DVC_ROOT/tmp/DVC_CHECKPOINT
> 3. User's code should block as long as the file still exists

Is that what `make_checkpoint()` does internally? (Assumed so for now)

pmrowla: > Including a copy of the cache?

The original repo's cache dir is used as the cache for the temporary repo

> Why not under the stage directly?
> Why do you need that field at all?

Checkpoint outs are treated differently than regular outputs, one of difference is that reproduction of a stage and checkout won't fail if a checkpoint out does not exist. So the flag applies to specific outputs rather than an entire stage.

> Is that what make_checkpoint() does internally? 

Yes

content/docs/command-reference/exp/run.md
```diff
@@ -0,0 +1,141 @@
...
+committing them to the Git repo. Unnecessary ones can be removed with
+`dvc exp gc`, or abandoned (and their data with `dvc gc`).
+
+## Checkpoints
+
+To track successive steps in a longer <abbr>experiment</abbr>, you can register
+checkpoints with DVC during your code or script runtime (similar to a logger).
+
+To do so, mark stage `outs` with `checkpoint: true` in `dvc.yaml`, and in your
+code either call the `dvc.api.make_checkpoint()` function (Python), or write a
```
jorgeorpinel, line 71: > mark stage `outs` with `checkpoint: true` in `dvc.yaml`, and...

@pmrowla so is this required for checkpoints to work or is it just optional? Other than stage repro/checkoiut not failing when a checkpoint output doesn't exist, what are the differences? I'm trying to understand what this is for, still a bit unclear... Thanks

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,69 @@
...
+present in the <abbr>workspace</abbr> (e.g. the last `dvc exp run`) with the
+latest committed versions (required). Only params/metrics that changed are
+listed, by default (show everything with `--all`).
+
+`a_rev` and `b_rev` are optional experiments to compare. They accept experiment
+names or hashes (see `dvc exp run`). Providing a single one results in comparing
+it to the workspace.
+
+All params and metrics defined in the experiments' `dvc.yaml` files are used by
+default.
```
jorgeorpinel, line 39: Why no `--targets` in this command? Could accept both params and metrics.

content/docs/command-reference/exp/diff.md
```diff
@@ -0,0 +1,69 @@
...
+- `--all` - list all parameters and metrics, including those without changes.
+
+- `--show-json` - prints the command's output in JSON format (machine-readable)
+  instead of a human-readable table.
+
+- `--show-md` - prints the command's output in the Markdown table format
+  ([GFM](https://github.github.com/gfm/#tables-extension-)).
+
+- `--old` - include the "Old" value column in addition to the new "Value" (and
+  "Change"). Useful when the values are not numeric, for example
```
jorgeorpinel, line 55: At some point earlier we said "Old" would be printed by default like in `params diff` but it seems this is more like `metrics diff` in general (based on the options), right?

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Description
+
+Shows experiments in a detailed table which includes parameters and metrics.
+Only the experiments derived from the Git `HEAD` are shown by default but all
+experiments can be included with the `---all-commits` option.
+
+`dvc exp show` enters a paginated screen by default, which you can exit by
+typing `Q` (or Esc) in your keyboard. Use `--no-pager` to print the entire
```
jorgeorpinel, line 27: > which you can exit by typing `Q` (or Esc)

Is this correct? I'm only guessing Esc on Windows, perhaps?

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,87 @@
...
+output to terminal instead.
+
+## Options
+
+- `-a`, `--all-branches` - print experiments derived from the tip of all Git
+  branches instead of just those derived from the current workspace. Note that
```
jorgeorpinel, line 33: > instead of just those derived from the current workspace

Did you mean those derived from `HEAD` (last commit)?

jorgeorpinel (commented):
A couple Qs on `exp show`, @pmrowla 👇 Thanks

content/docs/command-reference/exp/show.md
jorgeorpinel: What pager do we use? Looks like `less` but that's not available on Windows, usually.

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,99 @@
...
+Your terminal will enter a paginated screen by default, which you can exit by
+typing `Q` in your keyboard. Use `--no-pager` to print the entire table at once
+instead.
+
+```dvc
+$ dvc exp show --no-pager
+┏━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
+┃ Experiment    ┃ avg_prec ┃ roc_auc ┃ train.n_est┃ train.min_split ┃
+┡━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
+│ workspace     │  0.56191 │ 0.93345 │ 50         │ 2               │
```
jorgeorpinel, line 36: Quick Q. Why are metrics shown first? I feel like params "happen first" in chronological order. Same Q for `exp diff` (expecting same A). Mostly curious

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,99 @@
...
+- `--include-metrics <list>` - include the specified `dvc metrics` in the table
+  (only the specified metrics will be shown). Accepts a comma-separated `list`
+  of metric names.
+
+- `--exclude-metrics <list>` - exclude the specified `dvc metrics` from the
+  table (all param will be shown except for the specified ones). Accepts a
+  comma-separated `list` of metric names.
+
+- `--sort-by <name>` - sort related experiments by the specified metric or param
+  (`name`). Only one sort column (either metric or param) can be specified.
```
jorgeorpinel, line 77: > related experiments

Is that a hint that only experiments that have the given param/metric will be listed?

jorgeorpinel (commented):
Ah wait, I had 2 more Qs on `exp show` 👇

content/docs/command-reference/exp/show.md
pmrowla: It's the same pager behavior that we use for `dvc dag`, so `less` if it's available, otherwise whatever the `DVC_PAGER` env var is set to

content/docs/command-reference/exp/show.md
pmrowla: Yes it should be those derived from `HEAD`

content/docs/command-reference/exp/show.md
pmrowla: To be consistent with the viewer. I'm assuming that the viewer does metrics first since the user likely wants to see the experiment "results" (i.e output metrics) first when comparing a large list of experiments.

content/docs/command-reference/exp/show.md
pmrowla: In this case related means that for a checkpoint run, commits within that checkpoint run will always be grouped together, and that experiments will always be grouped by their original parent commit.

So if you have some experiments that are derived from commit `abc123`, and some experiments that are derived from commit `def456`, the experiments for `abc123` would be sorted and then grouped together, and then the experiments for `def456` would be sorted and grouped together.

content/docs/command-reference/exp/run.md
pmrowla: It is required for checkpoints to work, if a stage does not contain any `checkpoint` outputs, it will be reproduced like a regular standalone stage, and DVC will not monitor for the `make_checkpoint()` signal file (and will not create any checkpoint commits).

pmrowla: One other difference is that `checkpoint` outs are also treated similarly to `persist` outs, and are not removed when reproducing a stage. The difference between `checkpoint` and `persist` outs is that a `checkpoint` out will be reverted to its last committed state at the start of reproducing a stage, and DVC will not touch a regular `persist` out at all at the start of reproducing a stage.

pmrowla: Basically checkpoint outs are circular dependencies, where a truly reproducible stage should have independently tracked input (DVC dependency ) and output (DVC out) versions of the single checkpoint file. In DVC we don't have proper internal support for circular dependencies, but the `checkpoint` out behavior is a good enough workaround for now.

content/docs/command-reference/exp/diff.md
pmrowla: The only reason is because metrics and params also did not accept `--targets` at the time `exp diff` was implemented. It can be added as needed.

content/docs/command-reference/exp/diff.md
dberenbaum: What is the base? How would we define that? It seems like that's going to depend on a user's experiment design, and the current behavior is about as good a guess as we could make about what commit would be the base. It's consistent with all the other dvc commands as @pmrowla mentioned, so my vote would be to leave it as is.

jorgeorpinel: @dberenbaum the base of `dvc exp run` is always the last committed project version at the time of the run AFAU. They are actual Git commits (in the `exps` custom ref) so you can just look at the commit parent to determine the base, I think.

jorgeorpinel: WAIT... 

> What about comparing the last experiment with its base

I guess that's what the current behavior does if we assume the workspace has the last `exp run` 🙂 

The question could be whether to comp the last `exp` (by timestamp) to it's parent but that doesn't seem more inuitive.

OK nvmd! Resolving this one, thanks.

content/docs/command-reference/exp/diff.md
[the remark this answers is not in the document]
jorgeorpinel: Extracted Q to #5530. Resolving here.

content/docs/command-reference/exp/run.md
jorgeorpinel: Thanks for all the details as usual @pmrowla, I have mentioned them in docs now.

> required for checkpoints to work

Just wondering why we want/need this restriction. Couldn't DVC always monitor that signal file regardless? Perhaps it's a significant overhead to do so?

> a checkpoint out will be reverted to its last committed state at the start of reproducing a stage

But only with `exp run` right? Not with `repro`

jorgeorpinel: Also should it be `checkpoints: true`? (plural)

content/docs/command-reference/exp/diff.md
jorgeorpinel: We can discuss in #5530. Thanks

content/docs/command-reference/exp/show.md
jorgeorpinel: I see, got it. Thanks

And will (groups) be sorted by the parent's metric/param? Or chronological as usual

```
exp      sort_val
parent1   1
  child1   2
  child2   3
parent3   2
  child2   1
  child1   7
parent2   4
  child1   2
  child2   4
```

jorgeorpinel:
LGTM @shcheklein @dberenbaum @pmrowla ? Examples will be coming in a separate PR (this is big enough). The remaining commands also have their own PR (#2243).

content/docs/command-reference/exp/run.md
pmrowla: `dvc repro` will abort if a stage contains any `checkpoint` outs, checkpoint outs explicitly are only compatible with `dvc exp run`.

I also don't see why it should be plural, since it is a flag for a singular out? You can have a stage that contains more than one checkpoint out, the same way that you can have a stage containing more than one uncached/persist/etc out, but the flag still only applies per individual out

pmrowla: > Just wondering why we want/need this restriction. Couldn't DVC always monitor that signal file regardless? Perhaps it's a significant overhead to do so?

There is nothing preventing us from doing this, but it was written this way given that this is new behavior, and it's not exactly clear what DVC should be doing for stages that don't contain explicit checkpoint outs. For checkpoint outs, the user's stage should still be reproducible in between checkpoints (given an input checkpoint out state, the next output should always be the same). For "regular" stages, the expectation is that only the full stage is reproducible, rather than arbitrary smaller parts within the stage (so DVC should not create commits for any intermediate state within the stage).

content/docs/command-reference/exp/show.md
pmrowla: the top-level parent commits will always be organized chronologically, sorting is only done for (nested) experiment runs

jorgeorpinel: Got it. Updated info in 2e263859.

content/docs/command-reference/exp/show.md
```diff
@@ -0,0 +1,112 @@
...
+
+By default, the printed experiments table will include columns for all metrics
+and params from the entire project. The `--include-metrics`,
+`--exclude-metrics`, `--include-params`, and `--exclude-params` options can be
+used to choose them.
+
+Experiments in the table are first grouped (by parent commit). They are then
+sorted inside each group, chronologically by default. The `--sort-by` and
+`--sort-order` options can change this ordering, based on any single metric or
+param.
```
jorgeorpinel, line 52: Probably last Q (while we wait for approval) @pmrowla:

Can you sort by a column you're not currently showing (e.g. a param hidden with `--exclude-params`)? Thanks

pmrowla: No, if you try to sort by a column that is excluded, the `dvc exp show` command will error out.

content/docs/command-reference/exp/run.md
jorgeorpinel: > dvc repro will abort if a stage contains any checkpoint outs

Just noticed this answer. Hm, interesting... OK I'll document all this and we can move the discussion if any to the issue I created in core. Thanks!

content/docs/command-reference/exp/show.md
jorgeorpinel: Cool, noted in the ref.

dberenbaum (approved):
LGTM

jorgeorpinel:
OK guys I'm going to merge this. There will be more iterations in case anyone has more comments. Thanks!

PR #2207, opened 2021-02-17
This is a POC script for #2108. Current version doesn't create issues automatically.

### LINKED ISSUE #2108: help: output from i18n-like string file
UPDATE: Jump to https://github.com/iterative/dvc.org/issues/2108#issuecomment-914663833

---

Hello there!

Updating the CLI help output for every command is a bummer, but it can be done automatically.

By adding special HTML comments which are read by a script, it's possible to inline the output of commands into the markdown.

For example, in content/docs/command-reference/params/index.md:

```
## Synopsis

<!-- DVC_HELP "dvc params --help" -->
(output of dvc params --help goes here)
<!-- DVC_HELP_END -->
...
```

By traversing all files and looking for `<!-- DVC_HELP` comments, it's possible to run the specified command and inline its output in the markdown.

This could be generated automatically with each build, or it could be a script that's run manually and updates everything (as long as `dvc` is in the PATH).

@rogermparent thoughts?

iesahin:
Fixes #2108

The script creates `help-text` directory and outputs all `--help` text to different files. It checks `git diff` output to see whether the output changed, and if so asks the user to create an issue with `gh`. Currently formatting of the issue is broken but it should be enough to show what I mean. 

I can update the script to follow a certain branch in the repository and not to pollute the `master`. It can also be modified to use `master` (or some other branch) in DVC core, so that we can update the docs.


iesahin:
It uses `zsh` because of the path operators. I can adapt to `bash` if needed.

iesahin:
I think we can just close this and reopen it as needed. I can continue checking the `dvc --help` outputs and create issues myself if needed.

PR #2225, opened 2021-02-21
upgrades prettier to 2.2.1 to fix the bug causing trailingComma config to be ignored

iesahin:
Previously #2194 was merged to update restylebot configuration. However prettier still seems to behave differently when it comes to trailing commas as can be seen in #2224. I made some tests and prettier version in `package.json` seems to be old and new version adheres the configuration in `.prettierrc`.

### DIFF
--- a/content/docs/user-guide/basic-concepts/dvc-project.md
+++ b/content/docs/user-guide/basic-concepts/dvc-project.md
@@ -9,7 +9,7 @@ match:
     'DVC repository',
     'DVC repositories',
     repository,
-    repositories,
+    repositories
   ]
 tooltip: >-
   Initialized by running `dvc init` in the **workspace** (typically a Git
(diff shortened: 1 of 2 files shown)


shcheklein:
@iesahin CI fails - do we need to update the `dvc-project` doc along with this change?

iesahin:
> @iesahin CI fails - do we need to update the `dvc-project` doc along with this change?

@shcheklein it looks so. I've updated the file. Clean now. Thanks.

shcheklein (approved):

shcheklein:
Thanks @iesahin !

shcheklein:
@iesahin it looks like deployment failed - we need to update the `yarn.lock` as well?

```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
       info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

iesahin:
Yes `yarn install` updates the lock file. Thanks for the merge.

shcheklein:
@iesahin 👍 please create a PR that updates the lock, w/o an update to the lock file deployment is broken

PR #2240, opened 2021-02-25
Add metrics/plots show in get started

dberenbaum:
Following up on https://github.com/iterative/dvc.org/pull/2195#discussion_r581510800. Adds examples of `metrics show` and `plots show` in the getting started guide.


content/docs/start/metrics-parameters-plots.md
```diff
@@ -98,6 +98,34 @@ into `roc.json` for an additional plot.
...
+```
+
+Now let's view the plots:
+
+```dvc
+$ dvc plots show
+file:///Users/dvc/example-get-started/plots.html
+```
+
+![](/img/plots_prc_get_started_show.svg)
```
shcheklein, line 126: can we try to do them side by side? they take quite a lot of space at the moment

shcheklein (approved):

shcheklein:
one small question, otherwise looks good to me!

content/docs/start/metrics-parameters-plots.md
dberenbaum: Agreed. Any advice on how to do that? Putting the two on the same line has no impact, and I'm not sure how to resize given that they are SVGs. I'm assuming it's controlled by the web framework, and I don't know why these are on separate lines when the `diff` plots below fit on one. It looks like the `show` files are bigger than the `diff` SVGs for some reason.

jorgeorpinel: > Any advice on how to do that?

You could hardcode some of the HTML in https://dvc.org/doc that creates columns (on certain screen widths).

Now and then we stumble on the need to make 2-col sections, maybe we can think about introducing an engine customization for this @rogermparent ? Like our custom `<details/abbr>` tags for example (just not sure which tag to customize.

jorgeorpinel (approved):

jorgeorpinel:
Approved but I guess we don't want to merge this yet (into `experimen_reset`) because that will complicate #2195.

content/docs/start/metrics-parameters-plots.md
jorgeorpinel: Wait... Dah there's an easier way: Just combine both plots into one in an img editing app 🙂 I think you're already doing that in #2195?

dberenbaum: No, the other images just happen to fit for some reason despite also being two separate images. How would I combine two SVG files? Is another acceptable format I can convert to?

shcheklein: okay, managed to get them smaller (just edited SVGs directly)

shcheklein:
LGTM!

shcheklein:
@dberenbaum should we merge this first? I think you should have permissions to merge any approved PRs. So, feel free to do so :) (please use `Squash and merge`)

PR #2242, opened 2021-02-26
ref: exp-related updates in other to cmds, etc.

jorgeorpinel:
Matches certain changes in #2200 and #2243 (mainly for consistency in terminology, structure, etc.).

- [x] Change base to `master` before merging (after #2243).


content/docs/command-reference/repro.md
```diff
@@ -182,7 +182,8 @@ up-to-date and only execute the final stage.
 - `--pull` - [pulls](/doc/command-reference/pull) dependencies and outputs
   involved in the stages being reproduced, if they are found in the
   [default remote storage](/doc/command-reference/remote/default). Note that it
-  tries the local run-cache first (unless `--no-run-cache` is also used).
+  includes pulling any available run-cache (unless `--no-run-cache` is also
+  used).
```
jorgeorpinel, line 186: I'm not sure that this new text is correct but the previous one was unnecessary (already explained earlier that `repro` tries the run-cache first, which is not related to `--pull`).

jorgeorpinel: > not sure that this new text is correct

Hi @efiop maybe you can confirm about this please? Thanks

jorgeorpinel:
Need a review pls 🙂

content/docs/command-reference/repro.md
jorgeorpinel: Cc @pmrowla maybe you know? Tx

pmrowla: `repro --pull` pulls outputs for pipeline stages which have been restored via the run-cache. We should probably note that it does not pull data dependencies (see https://github.com/iterative/dvc/issues/4742)

jorgeorpinel: Hmmm ok thanks @pmrowla I'm going to rewrite this whole option desc... Resolving this!

content/docs/command-reference/repro.md
```diff
@@ -179,10 +179,10 @@ up-to-date and only execute the final stage.
...
   has no effect if `targets` are not provided.
 
-- `--pull` - [pulls](/doc/command-reference/pull) dependencies and outputs
-  involved in the stages being reproduced, if they are found in the
-  [default remote storage](/doc/command-reference/remote/default). Note that it
-  tries the local run-cache first (unless `--no-run-cache` is also used).
+- `--pull` - downloads dependencies and outputs in the stages being reproduced
+  from the [default remote storage](/doc/command-reference/remote/default) (see
+  `dvc pull`) based on the run-cache. Note that this doesn't include initial
+  pipeline data sources (never found in the run-cache).
```
jorgeorpinel, line 185: Is this better @pmrowla ? Thanks

jorgeorpinel: p.s. it's a kind of confusing option. Why is `--pull` specifically related to run-cache? 🤷 But anyway, when #4742 is addressed this should be taken care of.

pmrowla: likely just because the original request was just for a method to pull run-cache+outputs for specific (run-cached) stages: https://github.com/iterative/dvc/issues/4223

pmrowla: I think we should probably not use `dependencies` here at all. Maybe just say something like this?
`[pulls](...) outputs for any stages which restored from [run-cache](/doc/user-guide/experiment-management#automatic-log-of-stage-runs-run-cache) during pipeline reproduction. Outputs will be pulled from the [default remote storage](...).`

jorgeorpinel: Good call. Updated 👍

pmrowla (approved):
LGTM

PR #2243, opened 2021-02-26
ref: remaining exp cmds (branch,res,list,push,pull,gc)

jorgeorpinel:
> Continuation of #2200
> Also splits and improves #1926

- [x] Change base to `master` when #2200 is merged.

- `exp branch/list/push/pull/gc` Descs and Options
- Include `exp run --rev/reset`
- related updates to existing exp command refs.

- [ ] Update core repo (UI text) per these PRs. See https://github.com/iterative/dvc.org/pull/2243#pullrequestreview-602019436


content/docs/command-reference/exp/run.md
```diff
@@ -134,6 +141,17 @@ CPU cores).
 - `-j <number>`, `--jobs <number>` - run this `number` of queued experiments in
   parallel. Only applicable when used in conjunction with `--run-all`.
 
+- `-r <commit>`, `--rev <commit>` - continue an experiment from a specific
+  checkpoint name or hash (`commit`). Required for explicitly resuming
+  checkpoints in queued or temp dir runs.
```
jorgeorpinel, line 146: > Required for explicitly resuming checkpoints in queued or temp dir runs.

Got that from the CLI @pmrowla but TBH I don't know what it means 😬. Could you clarify? Thanks

content/docs/command-reference/exp/run.md
```diff
@@ -134,6 +141,17 @@ CPU cores).
...
+
+- `--reset` - restart a checkpoint experiment from scratch (resets existing
+  checkpoints to the initial state).
```
jorgeorpinel, line 149: > resets existing checkpoints

By this do we basically mean delete all checkpoints? A bit confused by the term "reset".

content/docs/command-reference/exp/run.md
```diff
@@ -134,6 +141,17 @@ CPU cores).
...
+
+- `-f`, `--force` - reproduce pipelines even if no changes were found (same as
+  `dvc repro -f`). For `checkpoint` experiments, this restarts form scratch,
+  like `--reset`.
```
jorgeorpinel, line 153: Can you please just confirm this is correct? Ty

content/docs/command-reference/exp/run.md
pmrowla: This is out of date, the current CLI message is 
```
  -r <experiment_rev>, --rev <experiment_rev>
                        Continue the specified checkpoint experiment. (Only
                        required for explicitly resuming checkpoints in queued
                        or temp dir runs.)
```

The old message was from when the default behavior for `exp run` was to always run from scratch (and reset checkpoints). To resume from an existing checkpoint, you would need to specify the revision to resume from.

The current behavior is that `exp run` with no args will resume the last run checkpoint by default (in the workspace), and `--reset` must be used to explicitly restart from scratch.

The `-r/--rev` argument is still required when the user wants to resume from a previous checkpoint  using `--queue` or `--temp`, since it is not obvious what point DVC should resume from in that situation.

content/docs/command-reference/exp/run.md
pmrowla: The current CLI message for this is
```
--reset               Reset existing checkpoints and restart the experiment.
```

If you have previously generated a series of checkpoints (ex. epochs 1, 2, and 3), normally `dvc exp run` will resume from the repo state at the end of `3` (and your next checkpoint would be epoch 4).

Using `--reset` means the repo will be reset to the initial state prior to checkpoint `1` (so your next checkpoint would be epoch 1), and the previous set of checkpoints will cleared/overridden with this new run.

content/docs/command-reference/exp/run.md
pmrowla: This is also outdated, `--reset` must be explicitly used to restart checkpoints from scratch. Also it should be noted that using `--reset` now implicitly sets `-f/--force` so that the checkpoints which were previously generated will be explicitly reproduced again (instead of loaded from run cache).

content/docs/command-reference/exp/branch.md
```diff
@@ -0,0 +1,38 @@
...
+  branch         Git branch name to use
+```
+
+## Description
+
+Makes a Git branch off the last commit (`HEAD`) based on the given `experiment`,
+using the `branch` name provided.
+
+This is equivalent to using `dvc exp apply` (applies the results from any
+experiment to the workspace) followed by Git branching and committing:
```
jorgeorpinel, line 22: Quick Q on `exp branch`: Is this exactly correct @pmrowla? Thanks

content/docs/command-reference/exp/gc.md
```diff
@@ -0,0 +1,59 @@
+# exp gc
+
+Remove `dvc experiments` from the <abbr>project</abbr>.
+
+## Synopsis
+
+```usage
+usage: dvc exp gc [-h] [-q | -v] [-w]
+                  [-a] [-T] [--all-commits] [--queued] [-f]
```
jorgeorpinel, line 9: Why no `--cloud` (or `--git-remote`) option? Since there's `exp push/pull`.

How can you delete experiments from a remote repo? I guess you have to clone it, `exp gc` locally, and `exp push` syncs it up in that case, removing remote exps?

content/docs/command-reference/exp/gc.md
```diff
@@ -0,0 +1,59 @@
...
+
+Note that after removing experiments, the corresponding data which may no longer
+be needed still remains in the <abbr>cache</abbr>. You can use `dvc gc`
+separately to delete it.
+
+## Options
+
+- `-w`, `--workspace` - keep _only_ experiments derived from the current
+  workspace This option is enabled automatically if `--all-tags`,
+  `--all-branches`, or `--all-commits` are used.
```
jorgeorpinel, line 33: When are exps "derived from the workspace"? I thought they can only be derived from a commit (not from unstaged changes i.e. the working tree)

content/docs/command-reference/exp/gc.md
```diff
@@ -0,0 +1,59 @@
...
+  the workspace (implies `-w`). Note that both options can be combined, for
+  example using the `-aT` flag.
+
+- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ Git commits
+  as well as the workspace (implies `-w`). This is mainly only needed when
+  clearing the experiments run queue.
+
+- `--queued` - keep experiments that haven't been run yet (defined via
+  `dvc exp run --queue`) . The experiment run queue will typically be cleared if
+  this option isn't used.
```
jorgeorpinel, line 49: > The experiment run queue will be cleared if this option isn't used.

Even without any scope option (`-w`, etc.)?

content/docs/command-reference/exp/run.md
jorgeorpinel: OK got it (updated).

I'm still seeing ` (Only required for explicitly resuming checkpoints in queued or temp dir runs.)` in DVC `2.0.0a3` (latest `master`) though.

> The -r/--rev argument is still required when the user wants to resume from a previous checkpoint using --queue or --temp, since it is not obvious what point DVC should resume from in that situation.

This part is still confusing to me. Are you saying that if the run is interrupted while there are still queued exps not run, and the run exps registered some checkpoints, DVC doesn't know wha to do? What options do you have as user to send in `rev` in that case? Thanks

content/docs/command-reference/exp/run.md
jorgeorpinel: > the repo will be reset
> the previous set of checkpoints will cleared

Yeah thanks, that's what I was guessing. I was confused by the phase "resets existing experiments". The workspace and experiment are reset. The checkpoints are deleted 👍

jorgeorpinel (commented):
Just 3 Qs about `exp gc` @pmrowla 👇 Thanks in advance.

content/docs/command-reference/exp/run.md
pmrowla: It means that if you want each specific queued run or temp directory run to be a "resume" run, you have to specify what checkpoint commit to resume from, otherwise any queued or temp directory run will always be a "reset" run starting from HEAD by default.

It's independent of whatever else is in the queue. The explicit parameter is required because when using queue+run-all or temp dir runs, the workspace is not modified (so it always just contains `HEAD`).

Let's say I queue 5 different sets of checkpoint experiments (with unique parameters for each set of checkpoints), and then use `--run-all`. Now If I just said `dvc exp run`, DVC has no way of knowing which of the 5 experiments which were just executed should be resumed.

So if I want to continue the second and third sets of checkpoints, I have to explicitly do

```
dvc exp run --queue -r <last_checkpoint_in_2>
dvc exp run --queue -r <last_checkpoint_in_3>
```

Now my next `--run-all` will continue generating two checkpoint runs, one which is resuming from `1`, and one which is resuming from `3`

jorgeorpinel: Oops this comment got mixed with my `exp gc` https://github.com/iterative/dvc.org/pull/2243#pullrequestreview-601464535... Hope it's not too confusing!

content/docs/command-reference/exp/list.md
```diff
@@ -0,0 +1,44 @@
+# exp list
+
+List `dvc experiments` in a <abbr>DVC repository</abbr> (remote or local).
```
jorgeorpinel, line 3: Can `exp list` include checkpoints?

content/docs/command-reference/exp/branch.md
pmrowla: The wording should should probably be `roughly equivalent`.

For standalone experiments, `exp branch` is exactly the same as doing:
```
$ dvc exp apply <experiment>
$ git add .
$ git commit
```

For checkpoints there is a difference, and it is actually equivalent to:
```
$ git merge <checkpoint_experiment>
```

The difference is that when you do `exp branch` on a checkpoint experiment, all of the previous checkpoint commits will also be included in the new branch. So for experiment run which contained 5 generated checkpoints, your new branch would append all 5 of those checkpoint commits to the previous HEAD.

Whereas, if you did `exp apply <checkpoint_experiment>` and then git committed the result, your new branch would only have 1 new commit after the previous HEAD. That one commit would be the squashed end result of all 5 of the checkpoint commits.

content/docs/command-reference/exp/gc.md
pmrowla: There is no way to remove pushed remote experiments at the moment (other than using `git`).

There was some discussion over whether the ability to remove them belongs in `exp gc/remove` (like removing DVC data objects from a DVC remote), or in `exp push` (like removing branches or tags from a git remote).

In git, to remove a remote branch or tag, you have to use `git push <remote_name> --delete <ref_name>`, so for removing DVC experiments from a git remote, it should probably work the same way.

content/docs/command-reference/exp/gc.md
pmrowla: A scope option is still required, `--queued` does not count as a scope option at the moment.

content/docs/command-reference/exp/list.md
pmrowla: It will include the name of a checkpoint run, but not individual commits within the checkpoint run.

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,62 @@
...
+## Synopsis
+
+```usage
+usage: dvc experiments pull [-h] [-q | -v] [-f] [--no-cache]
+                            [-r <name>] [-j <number>] [--run-cache]
+                            git_remote experiment
+
+positional arguments:
+  git_remote          Git remote name or repo URL
+  experiment          Experiment to pull
```
jorgeorpinel, line 15: I noticed the CLI uses `<>` around these names. Can we avoid that for consistency? Currently those are only used for metavalues.

- [ ] I should actually make a full list of core updates to sync up to these docs...

content/docs/user-guide/project-structure/internal-files.md
```diff
@@ -61,8 +61,8 @@ operation.
   dependencies and outputs, to allow safely running multiple DVC commands in
   parallel
 
-- `.dvc/tmp/exp<n>`: workspace copy number _n_ for
-  [parallel experiments](/doc/command-reference/exp/run#queueing-and-parallel-execution).
+- `.dvc/tmp/exp<n>`: _n_<sup>th</sup> workspace copy, used for temporary or
+  parallel <abbr>experiments</abbr> (see `dvc exp run`).
```
jorgeorpinel, line 65: Gm @pmrowla 👋 Is it `exp<n>` for `--temp` runs too? What does _n_ mean in those cases? (For parallel runs it's the job # I think.)

content/docs/command-reference/exp/run.md
```diff
@@ -119,10 +136,8 @@ CPU cores).
   name will generated by default, such as `exp-f80g4` (based on the experiment's
   hash).
 
-- `-f`, `--force` - reproduce an experiment pipeline, regenerating its results,
-  even if no changes were found. For `checkpoint` experiments, `dvc exp run -f`
-  will force regeneration of an existing checkpoint experiment branch from
-  scratch.
+- `--temp` - run this experiment in a separate temporary directory (in
+  `.dvc/tmp/exp<n>`) instead of your workspace. Useful for
```
jorgeorpinel, line 140: > Useful for

Oops I forgot to finish this haha. I guess I'm not sure what the use case is here? If you're already running experiments you're probably OK with changing your workspace? You can always `git reset --hard; dvc checkout`.

content/docs/command-reference/exp/run.md
jorgeorpinel: > means that if you want each specific queued run or temp directory run to be a "resume" run, you have to specify what checkpoint commit to resume from

Here I still wasn't getting how different queued checkpoint exps can share a single starting point if they all have their own checkpoints (different revs)... But I was assuming we're talking about `exp run --run-all` in every case.

> when using queue+run-all or temp dir runs, the workspace is not modified
> If I just tried to do dvc exp run, DVC has no way of knowing which...

This is much clearer, thanks! Updated in docs.

My only remaining Q on this is how to resume *multiple* checkpoint exps. Will `exp run --run-all` continue each one from their last checkpoint? Can you combine `--run-all` and `--rev`? Or perhaps that case is not supported

content/docs/command-reference/exp/branch.md
jorgeorpinel: OK, noted, thanks.

> exp branch is exactly the same as doing...

But when would the the `git branch` command happen? Before committing, I assume.

content/docs/command-reference/exp/gc.md
jorgeorpinel: > whether the ability to remove them belongs in exp gc/remove or in exp push

I vote for `exp gc -c` for consistency with plain `gc`. Maybe the commands could even get merged at some point.

content/docs/command-reference/exp/branch.md
```diff
@@ -0,0 +1,38 @@
+# exp branch
+
+Commit the results from any [experiment](/doc/command-reference/exp) to Git, in
+a new branch (which will become the current <abbr>workspace</abbr>).
```
jorgeorpinel, line 4: Another Q on `exp branch`:

> which will become the current workspace

Is this right? `git checkout <branch>` happens?
And what happens if there's unstaged changes in the working tree, does the command fail with a `git` error message? Are they lost? Thanks

content/docs/command-reference/exp/branch.md
```diff
@@ -0,0 +1,43 @@
+# exp branch
+
+Commit the results from an [experiment](/doc/command-reference/exp) in a new Git
+branch, which will become the <abbr>workspace</abbr>.
```
dberenbaum, line 4: The more we document experiments, the more I think it's better to be precise, even if it exposes the complexity of the implementation, rather than invent terminology to avoid those details (for example, results from an experiment are already committed, which makes this description seem misleading).

Would it be more accurate to say something like, "Create a new branch with the commits for an experiment" or "Persist commits for an experiment in a new branch"?

content/docs/command-reference/exp/gc.md
pmrowla: It should probably be worded more like "derived from the current workspace HEAD"

content/docs/command-reference/exp/branch.md
pmrowla: `git branch` would be before committing yes, the comparisons are just written assuming that you are already in whichever git branch you want the commits to go.

content/docs/user-guide/project-structure/internal-files.md
pmrowla: The directory is temp dir runs, the path should look like `.dvc/tmp/exps/...`. The `<something>` is automatically generated by python and depends on the system/platform, it's just a unique temporary directory name, the actual implementation should not be documented.

The only thing that should be documented here is that for parallel or temporary experiment runs, `.dvc/tmp/exps/` will be used as the default location for temporary workspace directories (rather than the default system `$TMPDIR`).

content/docs/command-reference/exp/run.md
pmrowla: It's just to give the user the option to run individual experiments in the same execution manner as `--queue`, without polluting their workspace.

content/docs/command-reference/exp/branch.md
pmrowla: `exp branch` does not checkout the branch. It just creates it, the same way that `git branch foo` just creates a branch named foo but does not checkout the new branch.

content/docs/command-reference/exp/run.md
pmrowla: I'm not sure that this needs to say anything beyond `run this experiment in a separate temporary directory instead of your workspace.`

There should maybe be a note somewhere else on the `exp run` page documenting where the temp dir for queued/parallel and `--temp` runs will go, but I'm not sure that kind of implementation detail needs to be documented in the usage section?

content/docs/command-reference/exp/run.md
pmrowla: `--run-all` just runs whatever is in your queue. If you want to resume multiple checkpoint exps, then you just have to queue them.

So do `dvc exp run --rev <something to continue> --queue` multiple times, with different exps in `<something to continue>`

pmrowla: Maybe the confusion is about the `continue/resume` wording. Internally, the way that `--rev` works is that DVC just runs an experiment starting from `--rev` instead of HEAD. And if `--rev` is the tip of an existing experiment branch, new checkpoint commits will be appended to that same branch instead of creating a new experiment branch name.

content/docs/command-reference/exp/branch.md
```diff
@@ -0,0 +1,43 @@
...
+`experiment` results to the workspace) followed by Git branching and committing:
+
+```dvc
+$ dvc exp apply experiment
+$ git checkout -b branch
+$ git add . && git commit
+```
+
+For [checkpoints](/doc/command-reference/exp/run#checkpoints), the `experiment`
+(custom Git branch) is merged into the new `branch`.
```
dberenbaum, line 33: If I understand correctly, `exp branch` is equivalent to `git branch <exp_name> <rev>`. This again seems like additional work to explain in other terms what's actually happening. If users understand the experiments already have their own commits (multiple in the cases of checkpoints), it seems easier to understand this command.

content/docs/command-reference/exp/run.md
jorgeorpinel: Aaaah I see. You can queue each one with a given `--rev` dah. Makes sense, good to know (I don't think I'll mention it in docs for now).

content/docs/command-reference/exp/gc.md
jorgeorpinel: Got it. Maybe it should be renamed `--head` (later). Will doc as-is for now and mention that somewhere else, thanks!

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
+# exp pull
+
+Download a single [experiment](/doc/command-reference/exp) to a Git remote, and
+its data to a `dvc remote`.
```
dberenbaum, line 4:
```suggestion
Download a single [experiment](/doc/command-reference/exp) from a Git remote, and
its data from a `dvc remote`.
```

content/docs/command-reference/exp/gc.md
jorgeorpinel: Got it 👍 

Note for self: `gc --queued` vs `remove --queue` (extra `d`).

content/docs/command-reference/exp/list.md
jorgeorpinel: K. `show` does list checkpoints, right? (I'll try it.)

jorgeorpinel: Wait I already know it does! Per https://dvc.org/doc/api-reference/make_checkpoint#example-every-100th-iteration

content/docs/user-guide/project-structure/internal-files.md
jorgeorpinel: Why did you think I intended to document the exact implementation details? I'm confirming about `exp<n>` which I got from somewhere and seems strange for `--temp`. Trying to keep https://dvc.org/doc/user-guide/project-structure/internal-files up to date...

Updated 👍

content/docs/command-reference/exp/run.md
jorgeorpinel: > I'm not sure that this needs to say anything beyond

We explain motivation/uses whenever possible. But if this is just a why-not utility then yes, Idk how to explain it anyway (for now — we'll see how users adopt it later).

> There should maybe be a note somewhere...

There is 🙂
image

> not sure that kind of implementation detail needs to be documented in the usage section

It's a good question. In this case it takes up 3 words in a parenthesis so why-not.

content/docs/command-reference/exp/branch.md
jorgeorpinel: Ooooh OK thanks, I had that wrong. I tried it and noticed that so this is fixed now ✔️

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
+# exp pull
+
+Download a single [experiment](/doc/command-reference/exp) from a Git remote, and
+its data from a `dvc remote`.
```
jorgeorpinel, line 4: Why only a single experiment?

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
...
+  experiment          Experiment to pull
+```
+
+## Description
+
+The `dvc exp push` and `dvc exp pull` commands are the means for sharing
+experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
+
+> Plain `git push` and `git pull` don't work with `dvc experiments` because
+> these are saved under custom Git references. See **How does DVC track
```
jorgeorpinel, line 24: Is there an equivalent `git` command this is a "wrapper" of? Like `git pu* --refs exps` or something like that.

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
...
+## Options
+
+- `-f`, `--force` - rewrite the `experiment` commit if it already exists in the
+  local repo. Equivalent to `git push --force` (rewrites history)
+
+- `--no-cache` - do not pull cached files files associated with this experiment
+  from DVC remote storage.
+
+- `-r <name>`, `--remote <name>` - name of the `dvc remote` to pull cached files
+  to.
```
jorgeorpinel, line 48: Note for self: should we consider renaming here to `--dvc-remote` to disambiguate? Rn we have

`dvc exp pu* <remote> <experiment> --remote <remote>`

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
...
+
+- `--run-cache` - downloads all available history of stage runs to the
+  `dvc remote` (to the cache only, like `dvc fetch --run-cache`).
```
jorgeorpinel, line 51: Why do we need to worry about the run-cache here?

content/docs/command-reference/exp/pull.md
```diff
@@ -0,0 +1,65 @@
...
+
+- `-j <number>`, `--jobs <number>` - parallelism level for DVC to upload data to
+  remote storage. The default value is `4 * cpu_count()`. For SSH remotes, the
```
jorgeorpinel, line 54: Does it also affect the transfer to the Git remote?

jorgeorpinel (commented):
One last round of QAish Qs for @pmrowla 🙏 (on `exp push/pull`) 👇

content/docs/command-reference/exp/branch.md
dberenbaum: I thought I saw a reference to experiments as "hidden branches" somewhere else. If we use that terminology, can we just describe `exp branch` as making the hidden branch visible/persistent?

content/docs/command-reference/exp/branch.md
```diff
@@ -0,0 +1,43 @@
...
+Commit the results from an [experiment](/doc/command-reference/exp) in a new Git
+branch, which will become the <abbr>workspace</abbr>.
+
+## Synopsis
+
+```usage
+usage: dvc exp branch [-h] [-q | -v] experiment branch
+
+positional arguments:
+  experiment     Experiment to be promoted
```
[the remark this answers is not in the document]
jorgeorpinel: Going for `commit/merge` for now bc/ "persist" lacks context here.

content/docs/command-reference/exp/branch.md
pmrowla: This is correct. There's multiple ways you could represent what `exp branch` does using git examples:

```
$ git branch <exp_name> <rev>
```
```
$ git branch <exp_name>
$ git checkout <exp_name>
$ git merge <exp_rev>
$ git checkout <original HEAD>
```
```
$ git checkout -b <exp_name>
$ git merge <exp_rev>
$ git checkout <original HEAD>
```

All of these do exactly the same thing as far as git is concerned. The first example is the simplest, but I'm not sure if that's a git workflow that typical DVC users would recognize without having to go look up the docs for `git branch` to see what it does.

content/docs/command-reference/exp/pull.md
pmrowla: It was just written that way to start with. Internally we can already push/pull multiple experiments at once, but we aren't really sure how users will want to use the experiment sharing via git remotes in the long run. Once we get more user feedback after 2.0 release we can update the CLI commands as needed.

content/docs/command-reference/exp/pull.md
pmrowla: It's just equivalent to `git push` and `git fetch` using the full refnames for experiment refs. The only reason a user can't use `git push` and `git fetch` out of the box is because git will not resolve the shortened experiment name into a full `refs/.../<refname>` path - git only does that automatically for `refs/tags` and `refs/heads` (branches).

pmrowla: If you want to get specific, `exp pull` is equivalent to

```
$ git fetch [-f/--force] <git_remote> refs/exps/<internal_exp_hash>/<exp_name>:refs/exps/<internal_exp_hash/<exp_name>
```

pmrowla: `exp pull` is technically also an explicit `git fetch` rather than `git pull` (git pull is a two part fetch + merge operation that fetches into one reference and then merges it into a second one). `exp pull` does not have a merge step.

content/docs/command-reference/exp/pull.md
pmrowla: I think the current core repo CLI usage explicitly uses `git_remote` everywhere that we want git remotes, and `remote` on its own is always assumed to mean DVC remote.

content/docs/command-reference/exp/pull.md
pmrowla: It's so that the DVC remote pulling step in `exp pull` remains consistent with regular `dvc pull`.

Experiments runs will use run-cached stages if they exist, so it is reasonable to expect that if I am pulling a coworker's experiment and want to reproduce it with `exp run`, I would also want to use the run-cached version of those pipeline stages rather than recomputing them myself. (Otherwise I would run with `-f/--force`)

content/docs/command-reference/exp/pull.md
pmrowla: No, there is nothing to parallelize when pushing a ref to a git remote.

content/docs/command-reference/exp/branch.md
jorgeorpinel: > results from an experiment are already committed

True. But not in a regular Git branch (branches technically are Git refs in `refs/heads`). And we don't really say experiments are commits except in a collapsed details section of `exp run`. We actually say they help you avoid polluting the repo with too many commits/branches.

> saw a reference to experiments as "hidden branches" somewhere

Nope (maybe in a previous iteration). We have "custom Git branch" in the Checkpoints collapsed details section somewhere.

jorgeorpinel: > "Create a new branch with the commits for an experiment
> Persist commits for an experiment in a new branch

"commits for an experiment" lacks context (we don't expect most users to know how exps are implemented internally). And also without enough context, "to persist" means something else (being persistent) but I did link use that term in the description now, linked to the corresponding section of the Exps guide.

p.s. BTW "which will become the workspace" and other phrases were wrong. That has been removed now 👍

content/docs/command-reference/exp/branch.md
jorgeorpinel: > If I understand correctly, exp branch is equivalent to `git branch <exp_name> <rev>`

I think that's only true when your `HEAD` is also the base of the given `experiment`, which may be the most common scenario but not always.

content/docs/command-reference/exp/pull.md
jorgeorpinel: OK cool.

> we aren't really sure how users will want to use the experiment sharing via git remotes

Maybe approximate Git? Push all the experiments derived from the current `HEAD` by default (make `experiment` an optional arg, could even be `commit` (revision) instead so you can specify another base commit, or a specific experiment). Just an idea

content/docs/command-reference/exp/pull.md
jorgeorpinel: > equivalent to git push and git fetch

Hey yeah, why not `exp fetch`? (note for self for now)

Thanks for all the details @pmrowla looks like you're quite the Git expert now!

content/docs/command-reference/exp/pull.md
jorgeorpinel: You're right, this shouldn't be confusing. Resolving

content/docs/command-reference/exp/pull.md
jorgeorpinel: Yes, makes sense. Although pushing/pulling the entire run-cache may be unexpected here since you're specifying an `experiment`. Will extract...

content/docs/command-reference/exp/pull.md
jorgeorpinel: Got it 👍 no change needed here then.

jorgeorpinel:
I think this is mergeable now.

content/docs/command-reference/exp/branch.md
dberenbaum: Thoughts on replacing "Commit" with anything else: "Add," "Keep," etc.? I don't want to hold this up longer, so go with whatever you prefer, but I don't like saying that `exp branch` commits something when that's not true. I worry it's going to cause additional confusion for anyone who starts to look any deeper.

My larger suggestion of being more direct and precise about what experiments are doing would likely require changes across lots of docs at this point, so I'll keep it as a note for now and maybe open an issue. I wish we could do A/B testing on our docs 😄 .

dberenbaum: Last suggestion (I hope): "Creates a new branch with the results from an experiment."

content/docs/command-reference/exp/branch.md
dberenbaum: > I think that's only true when your `HEAD` is also the base of the given `experiment` though? That may be the most common scenario, but not always the case.

For at least the first two commands, it shouldn't matter what your HEAD is, since they do not checkout anything from your current HEAD. For the last two, they would need a first step like `git checkout <base_rev>` if the current HEAD diverged from the experiment.

Also, I forgot that `exp branch` requires a branch name as an argument, so according to the docs, `<exp_name>` should really be `<branch>` in all these examples. Maybe worth considering whether the branch name arg should be optional, especially since users already have the option to name the experiments when creating them, but that's a note for another time.

To get back to the main point, I don't think we should keep:

```
$ dvc exp apply experiment
$ git checkout -b branch
$ git add . && git commit
```

It's inaccurate since `exp branch` is neither applying an experiment (I know this was an earlier confusion that has since been resolved) nor committing anything. My suggestion would be remove this. If we need more description, how about something like:

```
A new branch is created with the history of that experiment.
For [checkpoints](/doc/command-reference/exp/run#checkpoints), 
each checkpoint will have its own commit in the new branch.
```

dberenbaum (commented):
I put some final comments about `exp branch`. I think we are close. I want to make sure we don't have inaccuracies that could cause confusion.

content/docs/command-reference/exp/branch.md
jorgeorpinel: Don't worry we can discuss as much as needed (even after merged).

> replacing "Commit" with anything else: "Add," "Keep," etc... I don't like saying that exp branch commits something when that's not true.

OK I see your concern is about correctness. But I think it is correct when the experiment in question is based on some other branch (not `HEAD`). AFAU the resulting branch here is always based off `HEAD` which implies that its commits can be new (like git-cherry-pick`).

Will need to try this to figure out the best explanation, and circle back (probably in another iteration) ⌛

content/docs/command-reference/exp/branch.md
jorgeorpinel: > For the last two, they would need a first step like git checkout <base_rev> 

This is the part I didn't think was happening (maybe @pmrowla can confirm). I should double check all this anyway ⌛

> worth considering whether the branch name arg should be optional, especially since users already have the option to name the experiments when creating them, but that's a note for another

- Will note that somewhere...

> I don't think we should keep... It's inaccurate

We aren't saying that's exactly how `exp branch` is implemented, we say that it's "Equivalent to" that. Unless I'm (also) getting `exp apply` wrong, that seems correct i.e. the result is the same (BTW those sample commands have since been updated though).

content/docs/command-reference/exp/apply.md
```diff
@@ -1,6 +1,6 @@
 # exp apply
 
-Apply the results from any [experiment](/doc/command-reference/exp) to the
+Apply the results from an [experiment](/doc/command-reference/exp) to the
 <abbr>workspace</abbr>.
```
jorgeorpinel, line 4: https://github.com/iterative/dvc.org/pull/2243#discussion_r586466572 makes me wonder if I got `exp apply` wrong. It does just basically checkout an experiment (which is already a "hidden commit") into the working tree right @pmrowla ?

And if so why do that and expect the user to (re)commit that when we could maybe just `exp commit` directly? Rebasing the refs/exps history onto `HEAD`

AND if so that would be very similar to `exp branch` so these could be merged probably (`exp commit --branch`)

Cc @dberenbaum 

Thanks

shcheklein (approved):

jorgeorpinel:
Some pending discussions and lots of open questions (that I'll turn into QA/prod issues) but this is important for today's release and since @dberenbaum has expressed approval, merging the iteration!

content/docs/command-reference/exp/apply.md
dberenbaum: For the record @jorgeorpinel, I had the exact same confusions as you, which is why I'm so concerned about how we document each of these.

`dvc exp apply` == `git merge --no-commit` (similar to `git stash apply` if we treat the experiment as a stash).

`dvc exp branch` == `git branch <branch> <exp_rev>` (similar to checking out the experiment, creating a new branch from it, then checking out the original branch again, except that it skips the checkouts).

These seem similar on the surface, but they differ in nonobvious ways. For one, `dvc exp apply` can only happen where the experiment is derived from the HEAD. Also, it requires stashing any uncommitted changes in the workspace and then trying to apply them on top of the applied experiment (and these may conflict, in which case `dvc exp apply` fails or ignores the stashed workspace changes depending on whether the `--no-force` flag is used). Finally, `dvc exp apply` ignores any previous checkpoints or other history associated with the experiment and only tries to apply the changes specified in the last experiment commit.

`dvc exp branch` seems more complex on the surface but is a simpler operation: it takes whatever history is tracked by that experiment and puts it into a new branch. It doesn't care what is in HEAD or the workspace since it's doing nothing more than adding <exp_rev> to `refs/heads/<branch>`.

dberenbaum: > And if so why do that and expect the user to (re)commit that when we could maybe just `exp commit` directly? Rebasing the refs/exps history onto `HEAD`

Users need a way to navigate to any experiment without adding it to their branch history, or else it basically defeats the purpose of experiments.

pmrowla: > dvc exp apply == git merge --no-commit (similar to git stash apply if we treat the experiment as a stash).
>
> dvc exp branch == git branch <branch> <exp_rev> (similar to checking out the experiment, creating a new branch from it, then checking out the original branch again, except that it skips the checkouts).

This is correct with one exception:

>  Finally, dvc exp apply ignores any previous checkpoints or other history associated with the experiment and only tries to apply the changes specified in the last experiment commit.

This part is wrong, changes from the previous checkpoints are not ignored, they are also included in the merge. The exact equivalent for `dvc exp apply <rev>` would be `git merge --squash --no-commit <rev>`. All of the checkpoint commits are squashed into a single changeset, and then that is what gets merged into your head (as unstaged/uncommitted changes via `--no-commit`.

pmrowla: Assuming that your current HEAD is the parent/baseline for a given experiment, the important difference between `exp branch ...` and `git checkout -b ...; exp apply ...; git commit` is that for checkpoints, using `exp apply` gives you a single `--squash`'ed commit in the end, whereas `exp branch` would include all of the individual checkpoint commits.

dberenbaum: Ah, sorry about that mistake. Should have left the explanation to Peter.

The point was to stress that there are a few subtle differences (how checkpoint commits are handled, how the current HEAD and workspace impacts the command) that came up as I was playing with experiments, and those differences seem really unintuitive without some understanding of the implementation.

I'll open a new issue to discuss how we should document experiment (and maybe I'll open similar issues for other new features if needed). We can gather feedback from users and decide whether we need to adjust anything there.

dberenbaum: Sorry to keep reviving this thread, but I only now realized that @pmrowla probably was intending `exp apply` and `exp branch` to mirror `git stash apply` and `git stash branch` (didn't realize before there was a `git stash branch`). So borrowing from Git docs is one idea. From https://git-scm.com/docs/git-stash, the justification for having `branch` in addition to `apply` is:

> This is useful if the branch on which you ran git stash push has changed enough that git stash apply fails due to conflicts. Since the stash entry is applied on top of the commit that was HEAD at the time git stash was run, it restores the originally stashed state with no conflicts.

jorgeorpinel: Hey, sorry for the delay here. OK sounds like I should def. review the current docs then... ⌛ 

But 💡 should we also consider letting users use Git commands instead of providing these complicated wrappers? And document recommended `git` operations instead (`merge/rebase --squash (--no-commit)`, `branch <exp-rev>` cherry-pick`, etc.)

> Users need a way to navigate to any experiment

Another example: `git checkout`

> @pmrowla probably was intending exp apply and exp branch to mirror git stash apply and git stash branch

Makes sense. Originally exps were going to piggy-back on the Git stash in fact.

PR #2245, opened 2021-03-01
document stage list

skshetry:
Adding docs for `stage:list`. Implemented on https://github.com/iterative/dvc/pull/5498 and updated on https://github.com/iterative/dvc/pull/5499.

Pending issue: https://github.com/iterative/dvc/issues/5390, unlikely to change in a short term window.

### DIFF
--- a/content/docs/command-reference/stage/list.md
+++ b/content/docs/command-reference/stage/list.md
@@ -1,6 +1,6 @@
 # stage list
 
-List <abbr>stages</abbr> in the `dvc.yaml`.
+List <abbr>stages</abbr> in the project.
 
 ## Synopsis
 
@@ -16,8 +16,8 @@ positional arguments:
 
 ## Description
 
-`dvc stage list` is useful for quick listing of stages present in the dvc.yaml
-file without having to open it.
+`dvc stage list` is useful for quick listing of stages present in the project
+without having to open `dvc.yaml` files.
 
 It provides a list of stages, with a stage name and the description of the
 stage. The description of the stage is read from
@@ -25,15 +25,18 @@ stage. The description of the stage is read from
 or if it does not exist, DVC generates the short description from the other
 metadata it has about the stage.
 
+Without any targets, by default, it only lists the stages from the `dvc.yaml`
+present in the current working directory.
+
 DVC also uses this command to provide autocompletion in `bash` and `zsh`.
 
 ## Options
 
-- `targets` (optional command argument) - Stages to list from (`./dvc.yaml` by
+- `targets` (optional command argument) - stages to list from (`./dvc.yaml` by
   default). Different things can be provided as targets depending on the flags
   used (more details in each option). Examples:
 
-  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to reproduce
+  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to list from
   - `dvc stage list -R pipelines/`: Directory path to explore recursively for
     `dvc.yaml` files
   - `dvc stage list train-model`: Specific stage name or a
@@ -54,7 +57,7 @@ DVC also uses this command to provide autocompletion in `bash` and `zsh`.
   default, those errors are skipped. This option is only in effect if `--all` or
   `-R` is used.
 
-- `--names-only` - Only lists the name of the stages. It can be useful for
+- `--names-only` - only lists the name of the stages. It can be useful for
   scripting purposes. DVC uses it for providing autocompletion on `bash`.
 
 - `-h`, `--help` - prints the usage/help message, and exit.
--- a/content/docs/command-reference/stage/list.md
+++ b/content/docs/command-reference/stage/list.md
@@ -50,8 +50,8 @@ DVC also uses this command to provide autocompletion in `bash` and `zsh`.
   directories given as `targets`, and in their subdirectories. If there are no
   directories among the targets, this option has no effect.
 
-- `--all` - lists all of the stages from all `dvc.yaml` files in the repo, regardless of the
-  targets.
+- `--all` - lists all of the stages from all `dvc.yaml` files in the repo,
+  regardless of the targets.
 
 - `--fail` - fails immediately if any `dvc.yaml` files have syntax errors. By
   default, those errors are skipped. This option is only in effect if `--all` or
--- a/content/docs/command-reference/stage/list.md
+++ b/content/docs/command-reference/stage/list.md
@@ -6,8 +6,7 @@ List <abbr>stages</abbr> in the `dvc.yaml`.
 
 ```usage
 usage: dvc stage list [-h] [-q | -v]
-                      [-R] [--all] [--fail]
-                      [--names-only]
+                      [-R] [--all] [--fail] [--names-only]
                       [targets ...]
 
 positional arguments:
--- a/content/docs/command-reference/stage/list.md
+++ b/content/docs/command-reference/stage/list.md
@@ -50,7 +50,7 @@ DVC also uses this command to provide autocompletion in `bash` and `zsh`.
   directories given as `targets`, and in their subdirectories. If there are no
   directories among the targets, this option has no effect.
 
-- `--all` - lists all of the stages present in the repo, regardless of the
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+    in a specific `dvc.yaml` file
+
+- `-R`, `--recursive` - looks for `dvc.yaml` files to list from in any
+  directories given as `targets`, and in their subdirectories. If there are no
+  directories among the targets, this option has no effect.
+
+- `--all` - lists all of the stages present in the repo, regardless of the
+  targets.
+
+- `--fail` - fails immediately if any `dvc.yaml` files have syntax errors. By
```
dberenbaum, line 54: Does `--fail` check for syntax errors or schema errors?

skshetry: Yes, by default, all of the syntax errors that happens when loading them (yaml error, schema error and foreach resolve errors, etc.) are suppressed. `--fail` will immediately throw on any of those errors.

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+
+DVC also uses this command to provide autocompletion in `bash` and `zsh`.
+
+## Options
+
+- `targets` (optional command argument) - Stages to list from (`./dvc.yaml` by
+  default). Different things can be provided as targets depending on the flags
+  used (more details in each option). Examples:
+
+  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to reproduce
```
shcheklein, line 36: reproduce -> list

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+                      [targets ...]
+
+positional arguments:
+  targets          Show stages from a dvc.yaml file or a directory (with -R).
+                   'dvc.yaml' by default.
+```
+
+## Description
+
+`dvc stage list` is useful for quick listing of stages present in the dvc.yaml
```
shcheklein, line 19: `dvc.yaml` (inline code)

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
+# stage list
+
+List <abbr>stages</abbr> in the `dvc.yaml`.
```
shcheklein, line 3: in the project (since it's not only about _the_ `dvc.yaml`)

content/docs/command-reference/stage/list.md
shcheklein: here also - it's not only about _the_ `dvc.yaml` I think

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+stage. The description of the stage is read from
+[`desc` field in the `dvc.yaml` file](doc/user-guide/project-structure/pipelines-files#stage-entries),
+or if it does not exist, DVC generates the short description from the other
+metadata it has about the stage.
+
+DVC also uses this command to provide autocompletion in `bash` and `zsh`.
+
+## Options
+
+- `targets` (optional command argument) - Stages to list from (`./dvc.yaml` by
```
shcheklein, line 32: Stages -> stages

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+    in `./dvc.yaml`
+  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
+    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
+    in a specific `dvc.yaml` file
+
+- `-R`, `--recursive` - looks for `dvc.yaml` files to list from in any
+  directories given as `targets`, and in their subdirectories. If there are no
+  directories among the targets, this option has no effect.
+
+- `--all` - lists all of the stages present in the repo, regardless of the
```
shcheklein, line 50: does it looks for all `dvc.yaml`s? may be we can be a bit more explicit on this (it is alternative `dvc stage list -R .`?)

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,78 @@
...
+  targets.
+
+- `--fail` - fails immediately if any `dvc.yaml` files have syntax errors. By
+  default, those errors are skipped. This option is only in effect if `--all` or
+  `-R` is used.
+
+- `--names-only` - Only lists the name of the stages. It can be useful for
```
shcheklein, line 57: Only -> only

shcheklein (changes requested):
good stuff, @skshetry ! some quick comments here and there - nothing major

content/docs/command-reference/stage/list.md
skshetry: Yes, it looks for all the `dvc.yaml` files inside the repo, but it's not the same as `dvc stage list -R .` but to the `-R $(dvc root)`.

skshetry: @shcheklein, something like this?

```suggestion
- `--all` - lists all of the stages from all `dvc.yaml` files in the repo, regardless of the
```

shcheklein: sounds good!

shcheklein (approved):

skshetry:
@jorgeorpinel ping.

skshetry:
@shcheklein @jorgeorpinel, ping

shcheklein (approved):

shcheklein:
Merging this, to unblock @skshetry . Let's do a review as a separate ticket. If it's needed.

jorgeorpinel:
Sorry I couldn't get to it earlier but it was in my queue! I'll submit a PR if needed but this looks great, thanks 🙂

content/docs/command-reference/stage/list.md
```diff
@@ -0,0 +1,80 @@
...
+
+  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to list from
+  - `dvc stage list -R pipelines/`: Directory path to explore recursively for
+    `dvc.yaml` files
+  - `dvc stage list train-model`: Specific stage name or a
+    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
+    in `./dvc.yaml`
+  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
+    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
+    in a specific `dvc.yaml` file
```
jorgeorpinel, line 47: Just one Q guys: I thought we didn't want the "group" term for `foreach` stages? BTW I'd be happy to adopt this but requires some changes to the dvc.yaml guide (for context).

dberenbaum: So in the example:

```yaml
stages:
  cleanups:
    foreach: # List of simple values
      - raw1
      - labels1
      - raw2
    do:
      cmd: clean.py "${item}"
      outs:
        - ${item}.cln
```

It looks like both of these are supported:
`dvc stage list cleanups`
`dvc stage list cleanups@raw1`

That's great!

Maybe we could describe `cleanups` as a `foreach key`, or is that too technical? Is this even worth documenting? It seems pretty intuitive to try `dvc stage list cleanups` in this scenario.

jorgeorpinel: "foreach key" seems a bit technical.

Maybe it's not worth being so explicit indeed, and we can just say "Stage(s)" (here and in https://dvc.org/doc/command-reference/repro#options)

> dvc stage list cleanups@raw1

That may need documenting though. We currently say "in dvc.yaml" and `cleanups@raw1` really exists in dvc.lock... ⌛

jorgeorpinel: See #2380.

jorgeorpinel: I mean #2369.

dberenbaum: Good point. It's confusing, but not sure whether it's worth it to worry about the specifics here. Is there even a use case for `dvc stage list cleanups@raw1`?

jorgeorpinel: 🤷 maybe not but keep in mind this also applies to `repro`.

Issue #2121, opened 2021-01-28
cmd: `--to-cache/remote` updates (`add`, `import-url`, `update`)


isidentical:
- [x] add docs for `dvc {add, import-url} --to-remote` (#2091) 
- [x] add docs for straight-to-cache (including `--out` update #2220) (#2246)
- [x] add docs for `dvc update --to-remote` (#2175)
  - [x] mention `update --to-remote` doesn't work with repo imports (only from other external sources i.e. `import-url`ed) (#2237)
- ~~mention about `--to-remote` in https://dvc.org/doc/user-guide/managing-external-data~~
- [x] #2161 (#2172)
- [ ] Differentiate https://dvc.org/doc/command-reference/add#example-transfer-to-remote-storage from https://dvc.org/doc/command-reference/import-url#example-transfer-to-remote-storage per https://github.com/iterative/dvc.org/pull/2172#discussion_r583065469:
   > By default when you do import-url DVC downloads data into the workspace so that it can be saved into cache, and later into the remote storage. That's important to preserve it since we want to keep the project reproducible. In some situations though you might not have enough space on the machine you are running import-url, but you still want this data to be saved into remote storage, you still want this data be accessible through regular commands like dvc pull (e.g. to run the pipeline on another machine that has enough space in cache, or when a large shared cache is being used, etc). In those cases, to "bootstrap" the project it's handy to use --to-remote ....

jorgeorpinel:
> mention about --to-remote in https://dvc.org/doc/user-guide/managing-external-data

No need as we have a big ticket to rewrite that whole thing. But I did add another checkbox per https://github.com/iterative/dvc.org/pull/2175#pullrequestreview-596702408 👍

dberenbaum:
> > mention about --to-remote in https://dvc.org/doc/user-guide/managing-external-data
> 
> No need as we have a big ticket to rewrite that whole thing. But I did add another checkbox per [#2175 (review)](https://github.com/iterative/dvc.org/pull/2175#pullrequestreview-596702408) 👍

So what is the plan with https://dvc.org/doc/user-guide/managing-external-data for now? I think we have agreed that it's not worth re-writing for 2.0, since we need to change the functionality anyway, probably for 3.0. So that leaves us with some suboptimal choices:

- Keep it as is
- Drop it completely
- Find ways to further discourage users from relying on it (remove or replace links to it, find a way to make the warning at the top even more prominent, etc.)

jorgeorpinel:
[...]

PR #2246, opened 2021-03-01
cmd: add to-cache docs

### LINKED ISSUE #2220: ref: add --out for local targets
UPDATE: I see this is actually possible now (2.0). We should document it!
~~Q: Does the same apply to `import-url`?~~

---

Currently `add -o` only applies to "transferred" (external) data — straight-to-cache/remote functionality.

Since we're allowing `add` to move target data paths now, why not enable it for `targets` in the workspace too? E.g.:

```console
$ ls
yowassup
$ dvc add ./yowassup -o nicername
...
$ ls
yowassup  nicername  nicername.dvc
```

Other than UI consistency, there could be actual use cases for this like tracking copies of a same dataset that you know will have variations in different stages/pipelines (may work well with `foreach` stages).

\+ I assume enabling this at this point is as easy as removing the restriction that prevents it.

isidentical:
Part of #2121

Closes #2220

### DIFF
--- a/content/docs/command-reference/add.md
+++ b/content/docs/command-reference/add.md
@@ -151,6 +151,11 @@ not.
   > Note that external outputs typically require an external cache setup. See
   > link above for more details.
 
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace, to
+  [transfer](#example-transfer-to-cache) an external target into cache, and
+  track it in-place. Note that this can be combined with `--to-remote` to avoid
+  storing the data locally, while still adding it to the project.
+
 - `--to-remote` - import an external target, but don't move it into the
   workspace, nor cache it. [Transfer it](#example-transfer-to-remote-storage) it
   directly to remote storage (the default one, unless `-r` is specified)
@@ -159,11 +164,6 @@ not.
 - `-r <name>`, `--remote <name>` - name of the
   [remote storage](/doc/command-reference/remote) to transfer external target to
   (can only be used with `--to-remote`).
-
-- `-o <path>`, `--out <path>` - destination `path` for the transferred data. If
-  used with `--to-remote`, the data will be transferred to the remote storage.
-  Else, it will be transferred [to the cache](#example-transfer-to-cache) and
-  will be linked to the workspace.
 
 - `--desc <text>` - user description of the data (optional). This doesn't affect
   any DVC operations.
@@ -334,6 +334,56 @@ $ tree .dvc/cache
 Only the hash values of the `dir/` directory (with `.dir` file extension) and
 `file2` have been cached.
 
+## Example: Transfer to the cache
+
+When you have a large dataset in an external location, you may want to add it to
+the <abbr>project</abbr> without having to copy it into the workspace. Maybe
+your local disk doesn't even have enough space, but you have setup an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+that could handle it.
+
+The `--out` option lets you add external paths in a way that they are
+<abbr>cached</abbr> first, and then
+[linked](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
+to a given path inside the <abbr>workspace<abbr>. Let's initialize a DVC
+project:
+
+```dvc
+$ mkdir example # workspace
+$ cd example
+$ git init
+$ dvc init
+```
+
+Now we can add a `data.xml` file via HTTP for example, putting it a local path
+in our project:
+
+```
+$ dvc add https://data.dvc.org/get-started/data.xml -o data.xml
+
+To track the changes with git, run:
+
+    git add data.xml.dvc
+
+$ ls
+data.xml data.xml.dvc
+```
+
+The resulting `.dvc` file will save the provided local `path` as if the data was
+always there, while the `md5` hash points to the copy of the data that has now
+been transferred to the cache. Let's check the contents of `data.xml.dvc` in
+this case:
+
+```yaml
+outs:
+  - md5: a304afb96060aad90176268345e10355
+    nfiles: 1
+    path: data.xml
+```
+
+> For a similar operation that actually keeps a connection to the data source,
+> please see `dvc import-url`.
+
 ## Example: Transfer to remote storage
 
 When you have a large dataset in an external location, you may want to track it
@@ -379,50 +429,4 @@ system that can handle it), they can use `dvc pull` as usual:
 
 A       data.xml
 1 file added and 1 file fetched
-```
-
-## Example: Transfer to the cache
-
-When you have a large dataset in an external location, you may want to add it to
-your cache without actually copying it into the workspace first. This might be
-due to cache and workspace are in separate disks, which only the cache can
-handle that size of data. After the data is saved to your cache, we link it to
-your workspace with the
-[preffered links](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache).
-
-Let's initalize a DVC project;
-
-```dvc
-$ mkdir example # workspace
-$ cd example
-$ git init
-$ dvc init
-```
-
-Afterwards, let's setup a shared cache by following
-[this](https://dvc.org/doc/use-cases/shared-development-server#preparation)
-tutorial. When it is ready to go, we can add `data.xml` to our cache directly;
-
-```
-$ dvc add https://data.dvc.org/get-started/data.xml -o data.xml
-```
-
-Depending on the cache type configured on our workspace (can be set using
-`cache.type` config value), the data is either linked up or just copied over.
-For this use case, a reflink or a symlink is suggested.
-
-```
-$ ls
-data.xml data.xml.dvc
-```
-
-As it can be seen, this option doesn't track the source unlike
-[import-url](/doc/command-reference/import-url).
-
-```
- $ cat data.xml.dvc
-outs:
-- md5: a304afb96060aad90176268345e10355
-  nfiles: 1
-  path: data.xml
 ```
--- a/content/docs/command-reference/add.md
+++ b/content/docs/command-reference/add.md
@@ -160,8 +160,10 @@ not.
   [remote storage](/doc/command-reference/remote) to transfer external target to
   (can only be used with `--to-remote`).
 
-- `-o <path>`, `--out <path>` - destination `path` for the transferred data (can
-  only be used with `--to-remote`).
+- `-o <path>`, `--out <path>` - destination `path` for the transferred data. If
+  used with `--to-remote`, the data will be transferred to the remote storage.
+  Else, it will be transferred [to the cache](#example-transfer-to-cache) and
+  will be linked to the workspace.
 
 - `--desc <text>` - user description of the data (optional). This doesn't affect
   any DVC operations.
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/add.md
```diff
@@ -160,8 +160,10 @@ not.
   [remote storage](/doc/command-reference/remote) to transfer external target to
   (can only be used with `--to-remote`).
 
-- `-o <path>`, `--out <path>` - destination `path` for the transferred data (can
-  only be used with `--to-remote`).
+- `-o <path>`, `--out <path>` - destination `path` for the transferred data. If
+  used with `--to-remote`, the data will be transferred to the remote storage.
+  Else, it will be transferred [to the cache](#example-transfer-to-cache) and
+  will be linked to the workspace.
```
jorgeorpinel, line 166: Prob no need to explain to-remote again here. Please consider this text instead (below), and please move this option before `--to-remote`.

```suggestion
- `-o <path>`, `--out <path>` - destination `path` inside the workspace, to
  [transfer](#example-transfer-to-cache) an external target into, and track it
  in-place. Note that this can be combined with `--to-remote` to avoid storing
  the data locally, while still adding it to the project.
```

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
 A       data.xml
 1 file added and 1 file fetched
 ```
+
+## Example: Transfer to cache
+
+When you have a large dataset in an external location, you may want to add it to
```
jorgeorpinel, line 386: Please put this example before the to-remote one.

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
 A       data.xml
 1 file added and 1 file fetched
 ```
+
+## Example: Transfer to cache
```
jorgeorpinel, line 384: Just clarifying since "to cache" can be a verb (but we mean the noun). Committing:
```suggestion
## Example: Transfer to the cache
```

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
...
+## Example: Transfer to the cache
+
+When you have a large dataset in an external location, you may want to add it to
+your cache without actually copying it into the workspace first. This might be
+due to cache and workspace are in separate disks, which only the cache can
+handle that size of data. After the data is saved to your cache, we link it to
+your workspace with the
+[preffered links](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache).
+
+Let's initalize a DVC project;
```
jorgeorpinel, line 393: First iteration on this example intro:
```suggestion
When you have a large dataset in an external location, you may want to add it to
the <abbr>project</abbr> without having to copy it into the workspace. Maybe
your local disk doesn't even have enough space but you have setup an
[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
that could handle it.

The `--out` option lets you add external paths in a way that they are
<abbr>cached</abbr> first, and then
[linked](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
to a given path inside the <abbr>workspace<abbr>. Let's initialize a DVC
project:
```

jorgeorpinel: May still be missing to emphasize which link types are compatible with this scenario (symlink/hardlink)?

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
...
+```dvc
+$ mkdir example # workspace
+$ cd example
+$ git init
+$ dvc init
+```
+
+Afterwards, let's setup a shared cache by following
+[this](https://dvc.org/doc/use-cases/shared-development-server#preparation)
+tutorial. When it is ready to go, we can add `data.xml` to our cache directly;
```
jorgeorpinel, line 404: I don't think the "shared cache" thing is exactly appropriate. We just need the cache to be external, in a location that is large enough to hold the data.

I already linked to https://dvc.org/doc/use-cases/shared-development-server#configure-the-external-shared-cache in the revised intro, maybe that's enough (if so please delete this)? Otherwise let's repeat the whole process steps here please.

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
...
+
+```
+$ dvc add https://data.dvc.org/get-started/data.xml -o data.xml
+```
```
jorgeorpinel, line 408: Yeah for now I think delete the part about shared cache and just leave something like:

````suggestion
Now we can add a `data.xml` file via HTTP for example, putting it a local path
in our project:

```
$ mkdir data
$ dvc add https://data.dvc.org/get-started/data.xml --out data/data.xml
...
```
````

And instead of `...`, maybe show a sample command output here?

Comment (jorgeorpinel) [reply content/docs/command-reference/add.md:408]: Also not sure about the https://data.dvc.org/get-started/data.xml example... We could keep it hypothetical with some abstract path like `/mnt/nfs/raw/data.xml` (or some better idea) or ssh://user@example.com/raw/data.xml

Idk, I guess it's not very important but data.dvc.org is kind of an internal company thing (not a secret but still).

File: content/docs/command-reference/add.md [hunk 7]
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
...
+
+Depending on the cache type configured on our workspace (can be set using
+`cache.type` config value), the data is either linked up or just copied over.
+For this use case, a reflink or a symlink is suggested.
+
+```
+$ ls
+data.xml data.xml.dvc
+```
```
jorgeorpinel, line 417: I think we can merge this code block to the previous one, and skip this paragraph because we have to explain the linking earlier (see previous comments).

> BTW prob. no need to mention "just copied over": That doesn't seem useful in the tp-cache scenario 🙂

content/docs/command-reference/add.md
```diff
@@ -378,3 +380,49 @@ system that can handle it), they can use `dvc pull` as usual:
...
+As it can be seen, this option doesn't track the source unlike
+[import-url](/doc/command-reference/import-url).
+
+```
+ $ cat data.xml.dvc
+outs:
+- md5: a304afb96060aad90176268345e10355
+  nfiles: 1
+  path: data.xml
+```
```
jorgeorpinel, line 428:
````suggestion
The resulting `.dvc` file with save the provided local `path` as if the data was
always there, while the `md5` hash points to the copy of the data that has now
been transferred to the cache. Let's check the contents of `data/data.xml.dvc`
in this case:

```yaml
# data/data.xml.dvc
outs:
- md5: a304afb96060aad90176268345e10355
  nfiles: 1
  path: data.xml
```

> For a similar operation that actually keeps a connection to the data source, please see `dvc import-url`.
````

Comment (jorgeorpinel) [reply content/docs/command-reference/add.md:428]: Maybe also `ls/cat .dvc/cache/a3...` ?

**Review by jorgeorpinel — CHANGES_REQUESTED** [review]
Left a review with specific suggestions.

- [ ] Let's also address #2220 here?

I can take this over if needed after your next iteration @isidentical, thanks.

**Comment by isidentical** [comment]
> Let's also address #2220 here?

How should we document it though? Separate example (Tutorial: copying datasets in the same workspace) or just a note under the `-o` option (like `this option can also be used to copy data within the same project`)?

**Comment by dberenbaum** [comment]
Checking on the status of this @isidentical @jorgeorpinel ?

File: content/docs/command-reference/add.md [hunk 9]
```diff
@@ -151,6 +151,11 @@ not.
   > Note that external outputs typically require an external cache setup. See
   > link above for more details.
 
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace, to
+  [transfer](#example-transfer-to-cache) an external target into cache, and
+  track it in-place. Note that this can be combined with `--to-remote` to avoid
+  storing the data locally, while still adding it to the project.
```
jorgeorpinel, line 157:
```suggestion
- `-o <path>`, `--out <path>` - destination `path` to make a local target copy, or to
  [transfer](#example-transfer-to-cache) an external target into the cache and workspace.
  Note that this can be combined with `--to-remote` to avoid
  storing the data locally, while still adding it to the project.
```

jorgeorpinel:
> just a note under the -o option

That should be enough for now 👍

content/docs/command-reference/add.md
```diff
@@ -332,6 +334,56 @@ $ tree .dvc/cache
 Only the hash values of the `dir/` directory (with `.dir` file extension) and
 `file2` have been cached.
 
+## Example: Transfer to the cache
+
+When you have a large dataset in an external location, you may want to add it to
+the <abbr>project</abbr> without having to copy it into the workspace. Maybe
+your local disk doesn't even have enough space, but you have setup an
```
jorgeorpinel, line 341:
```suggestion
your local disk doesn't have enough space, but you have setup an
```

content/docs/command-reference/add.md
```diff
@@ -332,6 +334,56 @@ $ tree .dvc/cache
...
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+that could handle it.
+
+The `--out` option lets you add external paths in a way that they are
+<abbr>cached</abbr> first, and then
+[linked](/doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache)
+to a given path inside the <abbr>workspace<abbr>. Let's initialize a DVC
+project:
```
jorgeorpinel, line 349:
```suggestion
to a given path inside the <abbr>workspace<abbr>. Let's initialize an
example DVC project to try this:
```

content/docs/command-reference/add.md
```diff
@@ -332,6 +334,56 @@ $ tree .dvc/cache
...
+
+Now we can add a `data.xml` file via HTTP for example, putting it a local path
+in our project:
+
+```
+$ dvc add https://data.dvc.org/get-started/data.xml -o data.xml
+
+To track the changes with git, run:
+
+    git add data.xml.dvc
```
jorgeorpinel, line 366:
```suggestion
```

content/docs/command-reference/add.md
```diff
@@ -332,6 +334,56 @@ $ tree .dvc/cache
...
+
+$ ls
+data.xml data.xml.dvc
+```
+
+The resulting `.dvc` file will save the provided local `path` as if the data was
+always there, while the `md5` hash points to the copy of the data that has now
```
jorgeorpinel, line 373:
```suggestion
already in the workspace, while the `md5` hash points to the copy of the data that has now
```

content/docs/command-reference/add.md
```diff
@@ -332,6 +334,56 @@ $ tree .dvc/cache
...
+$ mkdir example # workspace
+$ cd example
+$ git init
+$ dvc init
+```
+
+Now we can add a `data.xml` file via HTTP for example, putting it a local path
+in our project:
+
+```
```
jorgeorpinel, line 361:
````suggestion
```dvc
````

**Review by jorgeorpinel — COMMENTED** [review]
Copy edits on the example:

**Review by jorgeorpinel — APPROVED** [review]

File: content/docs/command-reference/add.md [hunk 15]
```diff
@@ -332,6 +334,52 @@ $ tree .dvc/cache
...
+$ dvc add https://data.dvc.org/get-started/data.xml -o data.xml
+
+$ ls
+data.xml data.xml.dvc
+```
+
+The resulting `.dvc` file will save the provided local `path` as if the data was
+already in the workspace, while the `md5` hash points to the copy of the data
+that has now been transferred to the cache. Let's check the contents of
+`data.xml.dvc` in this case:
```
jorgeorpinel, line 371:
```suggestion
that has now been transferred to the <abbr>cache</abbr>. Let's check the
contents of `data.xml.dvc` in this case:
```

jorgeorpinel: We should probably also `ls/tree .dvc/cache` here but oh well...

isidentical:
Thanks @jorgeorpinel!

jorgeorpinel:
Np! BTW we forgot to update the link in the note in https://dvc.org/doc/user-guide/managing-external-data (currently goes to https://github.com/iterative/dvc/issues/4520). I'll do it separately.

PR #2262, opened 2021-03-04
cmd: fix params example for using custom params file

iesahin:
The last example in 

https://dvc.org/doc/command-reference/params

has shows parameter tracking for custom filenames. But the supplied `dvc.yaml` and `dvc.lock` files don't contain any custom filename. 

This PR fixes that.

### DIFF
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -240,10 +240,11 @@ stages:
     deps:
       - users.csv
     params:
-      - BOOL
-      - INT
-      - TrainConfig.EPOCHS
-      - TrainConfig.layers
+      - params.py:
+          - BOOL
+          - INT
+          - TrainConfig.EPOCHS
+          - TrainConfig.layers
     outs:
       - model.pkl
 ```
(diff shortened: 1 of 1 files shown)


shcheklein:
hmm, it supposed to work like it is in the docs now 🤔 @skshetry @efiop could you correct me if I'm wrong here? I remember there was a ticket recently on the DVC repo related to this?

skshetry:
@shcheklein, this PR is correct. Default file is `params.yaml`, other file do need to be explicitly specified.

skshetry (dismissed):

content/docs/command-reference/params/index.md
```diff
@@ -257,10 +258,11 @@ stages:
       - path: users.csv
         md5: 23be4307b23dcd740763d5fc67993f11
     params:
-      INT: 5
-      BOOL: true
-      TrainConfig.EPOCHS: 70
-      TrainConfig.layers: 9
+      - params.py:
```
skshetry, line 261: I think this should be a dictionary

shcheklein:
@skshetry thanks, I think I got confused by the name `custom params file` (in the viewer we call _custom_ files that can be connected explicitly)

iesahin:
This is the result of `dvc stage add -n name -p mycustomparams.toml:a,b,c python3 myscript.py` 

Screen Shot 2021-03-09 at 17 10 27

There is an initial `-` for the params file name, if that's what you found to be different than it should be. @skshetry 

What I mean by _custom_ is _not `params.yaml`_ :) @shcheklein

content/docs/command-reference/params/index.md
skshetry: @iesahin, this is a `dvc.lock` file, where `params` are in dictionary.

iesahin: Thank you. I think it should be done now.

shcheklein (approved):

PR #2286, opened 2021-03-10
cmd: document updated `exp run --reset` behavior

pmrowla:

Documents updated `exp run --reset` behavior and clarifies some checkpoints-related issues around `--queue/--run-all`.

See https://github.com/iterative/dvc/issues/5553#issuecomment-793198157 for behavior/change explanation


pmrowla:
core DVC change for this has been merged and is now available in master

content/docs/command-reference/exp/run.md
```diff
@@ -85,9 +85,10 @@ process finishes or gets interrupted (e.g. with Ctrl + `C`), DVC will
 <abbr>workspace</abbr> (overwriting any further changes done by the stage).
 
 `dvc exp run` again will continue from this point (useful for interrupted runs).
-Use `--reset` to roll-back the workspace to `HEAD` and restart the whole
-experiment. Alternatively, you can use `--rev` to continue from a specific
-(previous) checkpoint.
+Use `--reset` to reset (remove) any existing `checkpoint` outputs in your
+workspace before running the experiment. Note that `--reset` overrides any
```
jorgeorpinel, line 89: Sounds like "use reset to reset". We could just keep "to remove" I think. But...

> (remove) any existing `checkpoint` outputs

A little low level. What does `--reset` do from the user perspective. Does it not roll back to `HEAD` anymore?

content/docs/command-reference/exp/run.md
```diff
@@ -85,9 +85,10 @@ process finishes or gets interrupted (e.g. with Ctrl + `C`), DVC will
...
+existing `dvc.lock` entries for `checkpoint` outputs. Alternatively, you can use
```
jorgeorpinel, line 90: > `--reset` overrides any existing `dvc.lock` entries for `checkpoint` outputs

I'm not getting this note, what are we trying to warn about?

content/docs/command-reference/exp/run.md
jorgeorpinel: To be clearer, here are the steps implied in this section (checkpoints workflow with reset):

0. enable checkpoints in the stage ✅
1. update code to write signal files ✅
2. `exo run` ✅ (and possibly interrupt)
3. \> `exp run --reset` < - We are here

Why would we want to reset? Again, to restart the experiment, right? (Go back to step 2.)

content/docs/command-reference/exp/run.md
pmrowla: If your `dvc.lock` file is git-committed and the HEAD version of `dvc.lock` has a hash for the specified `checkpoint` out, that hash will be ignored on `--reset`, and the out will be completely removed before reproducing the pipeline (rather than being checked out to HEAD's hash).

content/docs/command-reference/exp/run.md
pmrowla: For `checkpoint` outs, `reset` removes the specified outs entirely (so that the specified intermediate model can be retrained). It used to roll back to `HEAD`, but this was changed (as of the core PR for these docs).

see https://github.com/iterative/dvc/issues/5553 for the product-side discussion on why this change was made

content/docs/command-reference/exp/run.md
dberenbaum: To put this in user terms, if you save a model at each checkpoint so that its weights get loaded as the starting point for the next training epoch, `--reset` will drop the model so that training will start without any existing model weights.

content/docs/command-reference/exp/run.md
jorgeorpinel: I see, thanks for the context.

> For checkpoint outs

Is there another use case for `--reset`?

jorgeorpinel: How about

```md
Use `--reset` to remove `checkpoint` outputs completely to restart the
experiment from scratch regardless of any cached contents.
```

But should we move `exp run --reset` up in the steps or say it's an alternative to plain `exp run` from the get-go?

0, 1, (same as above)
2a. `exp run` - runs experiment from `HEAD` state
2b. `exp run --reset` - runs experiment from scratch

content/docs/command-reference/exp/run.md
```diff
@@ -85,9 +85,10 @@ process finishes or gets interrupted (e.g. with Ctrl + `C`), DVC will
...
+`--rev` to continue from a specific (previous) checkpoint.
```
jorgeorpinel, line 91: Consolidated suggestion for this part:
```suggestion
Use `--reset` to remove `checkpoint` outputs completely in order to restart an
experiment from scratch, regardless of any cached contents (`dvc.lock` is
disregarded) — useful for re-training ML models, for example. Alternatively, you
can use `--rev` to continue from a specific (previous) checkpoint.
```

jorgeorpinel: But also the earlier sentence (same section) "Having at least a checkpoint output is needed so that the experiment can later restart based on that output's last cached state." should probably be rephrased.

content/docs/command-reference/exp/run.md
```diff
@@ -108,8 +109,11 @@ The `--queue` option lets you create an experiment as usual, except that nothing
...
-Use `dvc exp run --run-all` to process this queue. Note that if the queued
-experiments use checkpoints, `--run-all` implies `--reset` (restarts them).
+Note that for experiments which use checkpoints, queuing an experiment implies
+`--reset` unless an existing checkpoint to resume is explicitly provided via
+`--rev`.
+
+Use `dvc exp run --run-all` to process this queue.
 
 Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
 performance. This creates a temporary workspace copy for each subprocess (in
```
jorgeorpinel, line 119:
```suggestion
> Note that queuing an experiment that uses checkpoints implies `--reset`
> unless `--rev` is used.

Use `dvc exp run --run-all` to process the queue. Adding `-j` (`--jobs`),
experiment queues can be run in parallel for better performance. This creates a
temporary workspace copy for each subprocess (in
```

The 2nd p needs wrapping with the remaining line but I couldn't include it in the suggestion.

content/docs/command-reference/exp/run.md
```diff
@@ -139,11 +143,11 @@ CPU cores).
 
 - `--queue` - place this experiment at the end of a line for future execution,
   but do not actually run it yet. Use `dvc exp run --run-all` to process the
-  queue.
+  queue. For checkpoint experiments, this implies `--reset` unless `--rev` is
+  provided.
```
jorgeorpinel, line 147: Quick copy edit (will commit):
```suggestion
  queue. For checkpoint experiments, this implies `--reset` unless a `--rev` is
  provided.
```

content/docs/command-reference/exp/run.md
```diff
@@ -155,9 +159,8 @@ CPU cores).
   checkpoint name or hash (`commit`). This is needed for example to resume
   experiments from `--queue` or `--temp` runs.
 
-- `--reset` - restart a checkpoint experiment from scratch (resets the workspace
-  and clears existing checkpoints before the run). Implies `--force`, so that
-  cached checkpoint results are regenerated.
+- `--reset` - reset any checkpoint outputs before running this experiment.
+  Implies `--force`, so that cached checkpoint results are regenerated.
```
jorgeorpinel, line 163:
```suggestion
- `--reset` - remove `checkpoint` outputs before running this experiment
  (regardless of `dvc.lock` or their cached contents). Implies `--force`, so
  that cached checkpoint results are regenerated. Useful for ML model
  re-training.
```

jorgeorpinel (changes requested):
Left another round of commends and some specific suggestions ☝️ 

Also, from https://github.com/iterative/dvc/pull/5586, I'm not seeing this currently reflected/emphaiszed:

- "When --reset is used, all other (non-checkpoint) outs in dvc.lock will be unchanged
Previously, --reset would reset the entire dvc.lock file to HEAD"
- "--reset is now mutually exclusive with --rev ... we now explicitly error out"

Should we further clarify on those points?

p.s. I can take this over at some point if needed.

content/docs/command-reference/exp/run.md
pmrowla: If it's a new stage with new `checkpoint` outputs, both `2a` and `2b` are the same thing here (since there would not be any HEAD `dvc.lock` entries for the checkpoint outs).

pmrowla:
> * "When --reset is used, all other (non-checkpoint) outs in dvc.lock will be unchanged
>   Previously, --reset would reset the entire dvc.lock file to HEAD"
> * "--reset is now mutually exclusive with --rev ... we now explicitly error out"
> 
> Should we further clarify on those points?


The new behavior would be the expected default for anyone in ML using checkpoints (at least according to Dave+Dmitry), so I don't think it needs any more clarification.

Also the old behavior wasn't documented anywhere in the first place, and it was changed early enough in the 2.0 release cycle that it is unlikely that anyone will be running dvc while also intentionally expecting the original behavior.

content/docs/command-reference/exp/run.md
jorgeorpinel: OK so the first `exp run` of a checkpoint pipeline is the same as `exp run --reset`, got it. So yeah this whole section needs a bit of work to include all these key points. Please address my other comments/suggestions when you can and lmk to take this over a some final reorg of ideas (I'll try it in the meantime to confirm I got it 100% now).

jorgeorpinel:
"Previously, --reset would reset the entire dvc.lock file to HEAD" is in https://dvc.org/doc/command-reference/exp/run#checkpoints as "Use --reset to roll-back the workspace to HEAD". But it's only been published for a little while indeed.

If this is deemed obvi by the ML guys cc @dberenbaum then sure, one less complication to explain 👍

jorgeorpinel:
@pmrowla I guess at this point I should officially take it over. I'll ping you with more Qs if any (and for approval perhaps), thanks!

jorgeorpinel:
p.s. if possible in the future use branches directly on the upstream for this repo (easier to take over) 🙂

content/docs/command-reference/exp/run.md
jorgeorpinel: One more Q @pmrowla. `exp run --reset` still deletes the checkpoints also, right? Meaning the Git commits (I'm guessing the cache is left alone).

jorgeorpinel:
OK guys do you want to double check the latest text? Thanks

content/docs/command-reference/exp/run.md
pmrowla: The experiment ref for the previously existing checkpoint run is what gets "deleted". In practice, as long as the user's stage is deterministic the old existing is just moved to point to the new run.

Git commits are left alone, git will garbage collect them on its own once they are no longer pointed to by any references.

content/docs/command-reference/exp/run.md
```diff
@@ -69,27 +69,27 @@ checkpoints with DVC during your code or script runtime (similar to a logger).
 
 To do so, first mark stage `outs` with `checkpoint: true` in `dvc.yaml`. Having
 at least a checkpoint <abbr>output</abbr> is needed so that the experiment can
-later restart based on that output's last <abbr>cached</abbr> state.
+later continue from that output's last <abbr>cached</abbr> state.
 
 ⚠️ Using the `checkpoint` field in `dvc.yaml` is only compatibly with
```
pmrowla, line 74:
```suggestion
⚠️ Using the `checkpoint` field in `dvc.yaml` is only compatible with
```

typo here

content/docs/command-reference/exp/run.md
```diff
@@ -69,27 +69,27 @@ checkpoints with DVC during your code or script runtime (similar to a logger).
...
-Once this is setup, you can use `dvc exp run` to begin the experiment. When the
-process finishes or gets interrupted (e.g. with Ctrl + `C`), DVC will
-[apply](/doc/command-reference/exp/apply) the last checkpoint to the
-<abbr>workspace</abbr> (overwriting any further changes done by the stage).
+You can now use `dvc exp run` to begin the experiment. This removes any
+`checkpoint` outputs before running the experiment (regardless of whether they
+have cached versions). When the process finishes or gets interrupted (e.g. with
+Ctrl + `C`), DVC will [apply](/doc/command-reference/exp/apply) the last
+checkpoint to the <abbr>workspace</abbr> (overwriting any further changes done
+by the stage).
```
pmrowla, line 86: I think wording it this way might end up confusing users?

>  This removes any `checkpoint` outputs before running the experiment (regardless of whether they have cached versions).

`checkpoint` outs are only removed before running the experiment if `--reset` is specified. In a normal `dvc exp run` with no other arguments, we continue the experiment using the most recent version of that output.

If there is no previous run at all and the output has never existed before, then yes `dvc exp run` will do the same thing as `dvc exp run --reset`, but only in that specific case.

pmrowla:
@jorgeorpinel one thing that I wasn't sure about, otherwise LGTM

content/docs/command-reference/exp/run.md
```diff
@@ -69,27 +69,27 @@ checkpoints with DVC during your code or script runtime (similar to a logger).
 
 To do so, first mark stage `outs` with `checkpoint: true` in `dvc.yaml`. Having
 at least a checkpoint <abbr>output</abbr> is needed so that the experiment can
```
dberenbaum, line 71:
```suggestion
To do so, first mark stage `outs` with `checkpoint: true` in `dvc.yaml`.
At least one checkpoint <abbr>output</abbr> is needed so that the experiment can
```

dberenbaum: Minor style suggestion

content/docs/command-reference/exp/run.md
dberenbaum: Can this sentence get deleted? There's nothing special about what happens when you begin a checkpoint experiment. It's only different after at least one checkpoint has been generated.

dberenbaum, line 86:
```suggestion
checkpoint to the <abbr>workspace</abbr> (overwriting any further changes done
by the stage if the process was interrupted).
```
Alternatively:
```
When the process finishes, DVC will [apply](/doc/command-reference/exp/apply) the last
checkpoint to the <abbr>workspace</abbr>. If the process gets interrupted  (e.g. with
Ctrl + `C`), DVC will overwrite any changes done by the stage after the last checkpoint.
```
I find "overwriting any further changes" confusing unless it's specifically tied to the process being interrupted. If the process isn't interrupted, I think a final checkpoint is automatically generated to save the final outcome, so it's not overwriting further changes (@pmrowla can you confirm if that's correct?).

content/docs/command-reference/exp/run.md
```diff
@@ -155,9 +156,9 @@ CPU cores).
   checkpoint name or hash (`commit`). This is needed for example to resume
   experiments from `--queue` or `--temp` runs.
 
-- `--reset` - restart a checkpoint experiment from scratch (resets the workspace
-  and clears existing checkpoints before the run). Implies `--force`, so that
-  cached checkpoint results are regenerated.
+- `--reset` - remove `checkpoint` outputs before running this experiment
+  (regardless of `dvc.lock`). Implies `--force`, so that cached checkpoint
+  results are regenerated. Useful for ML model re-training
```
dberenbaum, line 161:
```suggestion
  results are regenerated. Useful for ML model re-training.
```

dberenbaum (commented):
Agree with the comments from @pmrowla and added one about the "overwriting any further changes" language. I already approved, so let me know if you want me to follow up again on anything, or else I'll leave merging to your discretion.

content/docs/command-reference/exp/run.md
pmrowla: Yeah, there's an additional commit generated at the end of the stage (assuming anything changed since the last time the user called `make_checkpoint`), and that is what gets applied. Even if the process is interrupted, we don't overwrite anything in the user's workspace.

dberenbaum: So should `(overwriting any further changes done by the stage)` get dropped?

content/docs/command-reference/exp/run.md
jorgeorpinel: Noted.

content/docs/command-reference/exp/run.md
jorgeorpinel: > There's nothing special about what happens when you begin a checkpoint experiment

What about what Peter just mentioned @dberenbaum? "If there is no previous run at all and the output has never existed before, then yes dvc exp run will do the same thing as dvc exp run --reset" 🙂

jorgeorpinel: > wording it this way might end up confusing users?

@pmrowla good point but maybe the problem is that the behavior can be a bit confusing per-se. I'll see if there's a more explicit way to express this but it already states "use `dvc exp run` **to begin** the experiment. This removes any `checkpoint` outputs" (it specifies "to begin").

pmrowla: That's not really a special case though, and even in that case nothing is being removed (since there is no initial data to even remove in the first place)

jorgeorpinel: Nvmd I think my latest update should make it very clear. It's getting a little long/convoluted but at least it's complete and correct (I think). See 9e1a7e94.

content/docs/command-reference/exp/run.md
```diff
@@ -69,27 +69,30 @@ checkpoints with DVC during your code or script runtime (similar to a logger).
...
-(previous) checkpoint.
+Following uses of `dvc exp run` will continue from this point (using the latest
+cached versions of all outputs). You can add a `--rev` to continue from a
+previous checkpoint instead (list them with `dvc exp show`). Or use `--reset` to
+start over (discards previous checkpoints and deletes `checkpoint` outputs, like
+the first `dvc exp run`) — useful for re-training ML models, for example.
 
-Note that `dvc exp show` displays checkpoints with a special branching format.
+> Note that after resetting checkpoint experiments, you may want to use `dvc gc`
+> to clean up unnecessary data in the cache.
```
pmrowla, line 95: I don't think this note is really necessary, but I suppose it doesn't hurt anything either.

content/docs/command-reference/exp/run.md
jorgeorpinel: Applied!

content/docs/command-reference/exp/run.md
jorgeorpinel: > even in that case nothing is being removed (since there is no initial data to even remove in the first place

What if you just changed regular outputs into checkpoint outputs in the last commit? (to migrate from regular pipeline to a checkpoint experiment)

content/docs/command-reference/exp/run.md
jorgeorpinel: Can't resetting checkpoints several times fill up the cache pretty quickly?

pmrowla: Sure, but that just comes with running a lot of experiments in general, it's not really specific to checkpoints.

content/docs/command-reference/exp/run.md
jorgeorpinel: > there's an additional commit generated at the end of the stage

Ah this I didn't realize. OK I rewrote this based on that and Dave's suggestions (thanks) for now.

> should (overwriting any further changes done by the stage) get dropped?

Woah wo why did I think that? A bit confused on this, so a final checkpoint that reflects the resulting workspace "is applied" (what is there to apply?) but the last checkpoint for interrupted runs is NOT applied?

content/docs/command-reference/exp/run.md
jorgeorpinel: 🤺 you're right. Removed from here but made a slightly more prominent note about this in the general Description.

dberenbaum:
> "Previously, --reset would reset the entire dvc.lock file to HEAD" is in https://dvc.org/doc/command-reference/exp/run#checkpoints as "Use --reset to roll-back the workspace to HEAD" for example. But it's only been public for a little while indeed. If this is deemed obvi by the ML guys cc @dberenbaum then sure, one less detail 👍

I don't know if it's obvious, but it's a sane default and it's only been public for a little while, so I don't think we need to note the change from previous behavior.

content/docs/command-reference/exp/run.md
dberenbaum: So the updated sentence is:

```
On this very first use,
any `checkpoint` outputs are deleted before running the experiment (regardless
of whether they have <abbr>cached</abbr> versions). 
```

> It's getting a little long/convoluted but at least it's complete and correct (I think).

Agreed on all counts 😄 . I think I get your point now that it's unclear whether an initial checkpoint run treats existing outs as checkpoints or not. Ideally, this is called out as an edge case since it might add confusion for the typical case where no outs exist yet, but 🤷 .

dberenbaum: Here's the latest language for reference:

```
If the process gets
interrupted (e.g. with Ctrl + `C`), DVC will
[apply](/doc/command-reference/exp/apply) the last checkpoint to the
<abbr>workspace</abbr> (overwriting any further changes).
```

Pending confirmation from @pmrowla, maybe we should remove any reference to `apply` or the workspace if nothing gets overwritten? Maybe this sentence isn't needed at all? Do we feel it necessary to call out process interruption to make explicit that this is an expected workflow? If so, maybe something like:

```
If the process gets
interrupted (e.g. with Ctrl + `C`), DVC will
keep all checkpoints recorded before interruption.
```

jorgeorpinel: > the typical case where no outs exist yet

OK if its an edge case we can leave it undocumented. Always good to simplify docs. Removed that sentence for now... Not sure I fully understand all the possible scenarios but will try to QA separately.

jorgeorpinel: OK. Removed `apply` and workspace references for now per Peter's comment above but it would def. be great to confirm!

jorgeorpinel:
Seems like the text addresses all the comments although I still have some questions but we can manage them separately. Thanks guys, merging!

PR #2295, opened 2021-03-12
Add target="_blank" rel="noopener noreferrer" to external-link

### LINKED ISSUE #2294: External links on blog not opening in new tab
I noticed that these links: 
image
do not open in a new tab, but direct away from our website.

Spoke with Jorge and tried adding `target="_blank"` inside the link, but not right either.  

Found this:  https://stackoverflow.com/questions/4425198/can-i-create-links-with-target-blank-in-markdown
Shows an answer re: Gatsby JS down the page.  Not sure if this is right, but beyond my knowledge anyway.

For reference this is the kind of link that is produced when published:  

image

rogermparent:
Fixes #2294 

I assume since it's called `external-link`, we want the attrs for external links on all instances of this custom component and there's no need to check for internal links conditionally.

### DIFF
--- a/plugins/external-link-plugin/index.js
+++ b/plugins/external-link-plugin/index.js
@@ -15,7 +15,9 @@ function isCorrectExternalLinkAttr(attrsKeyTagArray) {
 function renderTag(attrs) {
   return `
     <section class="elp-content-holder">
-      <a href="${escape(attrs.href)}" class="external-link-preview">
+      <a href="${escape(
+        attrs.href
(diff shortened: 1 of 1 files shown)


rogermparent:
Looks like it worked! Check out an example [here](https://dvc-org-external-link-t-drubzw.herokuapp.com/blog/january-21-dvc-heartbeat).

shcheklein:
@rogermparent thanks!

Issue #2292, opened 2021-03-11
Document checkpoints without make_checkpoint()


dberenbaum:
Our documentation so far in https://dvc.org/doc/command-reference/exp/run (and maybe elsewhere?) assumes that checkpoints only work with `make_checkpoint()` or a signal file. However, checkpoints still work without `make_checkpoint()` if the stage is designed to make a single checkpoint instead of multiple checkpoints. The final output will still be saved for the following iteration as long as it is marked as `checkpoint: true` in the `dvc.yaml`.

This would be great to document because:
1. It is language agnostic (without resorting to clunky workarounds) and therefore more consistent with the rest of DVC
2. It provides a simpler starting point for using checkpoints

See https://github.com/iterative/dvc-checkpoints-mnist/tree/python_agnostic for an example.

- [ ] Address also specific docs feedback in #2319

@pmrowla @jorgeorpinel @dmpetrov

pmrowla:
This is just using a regular output as a checkpoint. I guess the distinction is that in order for the "resume" feature in `exp run` to work, you still have to specify the `checkpoint` flag, but otherwise this is just doing the regular

```
dvc repro
git add .
git commit
<repeat>
```

workflow (with the results stored in an experiment ref instead of a regular git branch).

Maybe we should just consider having an explicit flag to extend ("resume") an existing experiment branch (that may or may not have `checkpoint` outputs) instead?

dberenbaum:
> This is just using a regular output as a checkpoint.

Exactly, which makes it easier to grasp for existing DVC users and more consistent with the rest of DVC (no need to inject DVC API functions into the user code). Not that `make_checkpoint()` isn't useful, it just seems like we are jumping straight into the advanced use case.

iesahin:
So `make_checkpoint()` is for creating checkpoints manually, but without it `dvc exp run` already saves a checkpoint after running the command. Right?

If `checkpoint: true` in `dvc.yaml`, it's as if running the experiments in a loop, we can 

```
while true ; do 
   dvc exp run
done
```
to run the experiments continuously.

pmrowla:
> So `make_checkpoint()` is for creating checkpoints manually, but without it `dvc exp run` already saves a checkpoint after running the command. Right?

Yes, DVC will always generate a final checkpoint commit after running the command (assuming the workspace state has actually changed and there is actually changes to commit)

jorgeorpinel:
[...]

PR #2300, opened 2021-03-14
ref: general updates to Experiments

jorgeorpinel:
Clarifications, corrections, etc.

- Begins to address #2292
- Reflects iterative/dvc/issues/5664
- Misc. copy edits and fixes

Also related (partially addressed):
- https://github.com/iterative/dvc.org/pull/2243#pullrequestreview-603422799 (and others in that ticket)
- https://github.com/iterative/dvc.org/pull/2286#discussion_r597957922 and https://github.com/iterative/dvc.org/pull/2286#discussion_r596867654


content/docs/api-reference/make_checkpoint.md
```diff
@@ -106,8 +103,7 @@ Experiment results have been applied to your workspace.
 
 In this example we kill the process (with Ctrl + C) after 3 checkpoints (at 0,
 100, and 200). The <abbr>cache</abbr> will contain those 3 versions of
-`int.txt`. DVC applies the last checkpoint to the <abbr>workspace</abbr> even if
-more cycles happened after that:
```
jorgeorpinel, line 110: Removed per https://github.com/iterative/dvc.org/pull/2286#discussion_r596867654 but I'm pretty sure I tried this and it was the case back then. Need to double check now... ⌛

jorgeorpinel: Hi @pmrowla,

So actually the last checkpoint is being applied to the workspace. In this case, int.txt contains `200` (or some other % 100 number). So did we get that wrong in docs after all (see link above) ? I'm on DVC 2.0.6.

pmrowla: I'm on vacation and can't get into testing this, but looking at the example code I'm guessing it is because your "user" code does not cleanly handle the keyboard interrupt (and will exit with an error). If the user stage exits with an error, we don't do the extra step to generate another commit. We will generate the extra commit in situations like:

- User code runs a fixed number of checkpoint iterations and then does some other final step:

```python
for i in range(100):
    # do 100 iterations
    make_checkpoint()

# do something else that modifies the workspace
do_something()
```

- User code runs forever and is intended to be stopped (cleanly) with ctrl-c

```python
try:
    while True:
        # do something that modifies the workspace
        do_something()
        make_checkpoint()
except KeyboardInterrupt:
    pass
# modifications written in do_something() but interrupted before reaching
# make_checkpoint() will now be committed in DVC when this code returns/exits cleanly
```

I'm guessing that if you rewrite the docs example to work like this second case, you will see the non-100 values be written to your text file (and saved/committed in DVC)

pmrowla: Basically, we generate a commit when a repro'd pipeline is reproduced cleanly. The potential "extra" commit at the end of checkpoint runs is just the same commit that is made for regular (non-checkpoint) `exp run`/`repro` runs.

When regular `exp run` or `repro` fails (because the stage/pipeline code exits with non-zero), we don't save anything, since we assume that the user does not want us to save the failed repo state. So if your checkpoint pipeline exits with a non-zero status, we will also not generate the extra/final commit.

content/docs/command-reference/exp/gc.md
```diff
@@ -40,9 +40,10 @@ separately to delete it.
   the workspace (implies `-w`). Note that both options can be combined, for
   example using the `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ Git commits
-  as well as the workspace (implies `-w`). This is mainly only needed when
-  clearing the experiments run queue.
+- `--all-commits` - same as `-a` or `-T` above, but applies to _all_
+  [reachable](https://git-scm.com/docs/gitglossary#def_reachable) Git commits as
+  well as the workspace (implies `-w`). This is mainly only needed when clearing
+  the experiments run queue.
```
pmrowla, line 46: This is incorrect - `--all-commits` just includes all commits which are ancestors of any git tags or branches in the git repo. This is not the same things as "all reachable commits".

All reachable commits would include any git commit object which is reachable from any git reference at all, including things like `git stash` merge commits, `git notes`, and anything referenced in custom git refs (like DVC experiment commits - which are explicitly not included in `--all-commits`).

content/docs/command-reference/gc.md
```diff
@@ -62,7 +62,8 @@ The default remote is cleaned (see `dvc config core.remote`) unless the
   versions of an experiment or project. Note that both options can be combined,
   for example using the `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to all Git commits,
+- `--all-commits` - same as `-a` or `-T` above, but applies to all
+  [reachable](https://git-scm.com/docs/gitglossary#def_reachable) Git commits,
```
pmrowla, line 66: same as above

content/docs/command-reference/exp/remove.md
```diff
@@ -14,9 +14,22 @@ positional arguments:
...
+> Note that all the checkpoints in an experiment are deleted by this command.
+
+<details>
+
+### What happens to removed experiments?
+
+Internally, `dvc exp` uses actual Git commits under custom Git references (found
+in `.git/refs/exps`). `dvc remove` deletes the references only, rendering the
+commits unreachable. `git` should garbage collect them automatically
+[after some time](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt).
```
pmrowla, line 30: Likewise, I don't think this is something that needs to be documented either. It's a low-level implementation detail.

The DVC docs should just say that removing an experiment works the same way as removing a git branch or tag.

content/docs/api-reference/make_checkpoint.md
dberenbaum: @jorgeorpinel I just tested and I think this particular example is causing confusion because `make_checkpoint()` takes way longer than the rest of the loop, making it almost impossible to interrupt at any point other than while `make_checkpoint()` is running. This creates the illusion that a checkpoint is being applied to the workspace. Add an `import time; time.sleep(0.1)` into the loop and you will be able to interrupt between checkpoints and see that the workspace reflects the current iteration, not the last checkpoint.

content/docs/command-reference/exp/run.md
```diff
@@ -74,8 +74,8 @@ To do so, first mark stage `outs` with `checkpoint: true` in `dvc.yaml`. At
 least one checkpoint <abbr>output</abbr> is needed so that the experiment can
 later continue from that output's last cached state.
 
-⚠️ Using the `checkpoint` field in `dvc.yaml` is only compatible with
-`dvc exp run`, `dvc repro` will abort if any stage contains it.
+⚠️ Note that using the `checkpoint` field in `dvc.yaml` is not compatibly with
```
dberenbaum, line 77:
```suggestion
⚠️ Note that using the `checkpoint` field in `dvc.yaml` is not compatible with
```

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -379,8 +379,8 @@ validation and auto-completion.
 | `desc`       | (Optional) user description for this output. This doesn't affect any DVC operations.                                                                                                                                                                                               |
 | `checkpoint` | Set to `true` to let DVC know that this output is associated with [in-code checkpoints](/doc/user-guide/experiment-management#checkpoints-in-source-code). These outputs are reverted to their last cached version at `dvc exp run` and also `persist` during the stage execution. |
 
-⚠️ Using the `checkpoint` field is only compatibly with `dvc exp run`,
-`dvc repro` will abort if any stage contains it.
+⚠️ Note that using the `checkpoint` field in `dvc.yaml` is not compatibly with
```
dberenbaum, line 382:
```suggestion
⚠️ Note that using the `checkpoint` field in `dvc.yaml` is not compatible with
```

dberenbaum (commented):
Since @pmrowla is out this week, do you want me to take review of this, or is there another engineer who might be able to do it? Otherwise, can it wait until next week?

content/docs/command-reference/exp/gc.md
jorgeorpinel: Thanks for the answers and review @pmrowla ! No rush here, please enjoy your vacation 🌴

jorgeorpinel:
OK I guess this has grown a lot already + it's already being reviewed...

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: > your example user code does not cleanly handle the keyboard interrupt
If the user stage exits with an error, we don't do the extra step to generate another commit

The example is simplified. The actual code I run catches the `KeyboardInterrupt`. The note about errors in interesting though, mentioned in eb029ecf

> the same commit that is made for regular (non-checkpoint) exp run/repro

My Q was about the application of the last regular checkpoint though (not the final extra checkpoint).
But again, the info about final checkpoints is def. valuable; I also added some of it (very briefly) to docs.
_Except that `repro` doesn't make commits. (?)_

jorgeorpinel: > this particular example is causing confusion because make_checkpoint() takes way longer than the rest of the loop...  This creates the illusion that a checkpoint is being applied to the workspace.

Ah that makes sense @dberenbaum ! OK that explains it, thanks. So the removal of this sentence is correct 👍 

Should I add that `time` stuff to this example for now? Probably not... I need to totally rewrite it soon anyway (not in this PR).

content/docs/command-reference/exp/gc.md
jorgeorpinel: Good to know. Fortunately the official definition of "reachable" is vague enough that it can mean what we want it to mean, I think (since we don't specify reachable from where).

image
> https://git-scm.com/docs/gitglossary#def_reachable

> --all-commits just includes all commits which are ancestors of any git tags or branches in the git repo.

This makes me wonder though, is there a difference between `-A` and `-aT`? 🤔
_Unrelated to this PR so no rush_

> I don't think we should get into discussing concepts like "reachable" git commits

I think that saying "all commits" is definitely wrong though. Esp. now that we have checkpoints (and we mention that they are implemented as commits).

content/docs/command-reference/gc.md
jorgeorpinel: Included this one since we emphasize from Exp docs that `exp gc` only does half of the job. Keeping for now (see https://github.com/iterative/dvc.org/pull/2300#discussion_r599268970).

content/docs/command-reference/exp/remove.md
jorgeorpinel: You may be right. Then again we have these here and there for the more advanced users. These are in line with the **How does DVC track experiments?** and **How are checkpoints captured by DVC?** sections in [`exp run`](https://dvc.org/doc/command-reference/exp/run).

The reason I decided to include them is that `git`-savvy users may want to manipulate experiments manually or even build integrations (and this is probably hard to figure out from the codebase). WDYT @dberenbaum and/or @shcheklein ? Thanks

p.s. keeping in mind that they are designed as non-intrusive/optional reading:

image

jorgeorpinel:
> can it wait until next week?

@dberenbaum if there are contention points then sure we can postpone but otherwise this is meant as maintenance work and I think we should be able to answer any Qs regardless (but the one I had has been solved).

Checking your review now (ty btw) ⌛

jorgeorpinel:
@dberenbaum PTAL.

And please lmk if you think I should extract the changes to content/docs/api-reference/make_checkpoint.md to another PR (this one won't close #2292 anyway) and then we can merge everything else first. Keeping it here for now since there are valuable discussions about that ref, above.

content/docs/command-reference/gc.md
jorgeorpinel: p.s. on "reachable commits": would "regular commits" work better? Or "your commits"?
This change is also in `exp show -A`, BTW.

jorgeorpinel:
p.p.s. actually it seems that the possible disagreements are more in `exp remove` as well as in `--all-commits` so I could extract **those** into another PR if you'd like (and merge `make_checkpoint` plus the misc. copy edits).

content/docs/api-reference/make_checkpoint.md
dberenbaum: Yeah, probably fine to leave it, especially if you are planning to replace it soon anyway.

dberenbaum:
Looks good, @jorgeorpinel! Agree on your ideas for breaking out the PRs: 

* `make_checkpoint.md` changes can probably stay here, and more will be addressed in a future PR.
* `remove` and `gc` could be extracted so that conversation about reachable commits doesn't block the rest of the PR.

jorgeorpinel:
Rolled back those files @dberenbaum. Pls approve so we can merge. Thanks!

content/docs/api-reference/make_checkpoint.md
```diff
@@ -76,16 +73,18 @@ while True:
...
 
     if i_ % 100 == 0:
         make_checkpoint()
 ```
 
-Using `dvc repro` with a continuous process such as this may not be helpful, as
-you know the output file will keep changing every time. Instead you can execute
-the stage with `dvc exp run` and end the process when you decide:
+Since `checkpoint` outputs in effect implement a circular dependency,
+`dvc repro` does not support running this stage. Let's the stage with
```
dberenbaum, line 86:
```suggestion
`dvc repro` does not support running this stage. Let's execute the stage with
```

dberenbaum (approved):
Thanks, @jorgeorpinel!

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: Thanks 😅

content/docs/command-reference/exp/show.md
```diff
@@ -62,8 +62,9 @@ metric or param.
   the workspace. Note that both options can be combined, for example using the
   `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ Git commits
-  as well as the workspace. This prints all experiments available.
+- `-A`, `--all-commits` - same as `-a` or `-T` above, but applies to _all_
+  reachable Git commits as well as the workspace. This prints all experiments
+  available.
```
jorgeorpinel, line 67: Oops I forgot to roll this one back 🤦

jorgeorpinel: Noted in https://github.com/iterative/dvc.org/pull/2325#discussion_r600038175 so I don't forget to roll back if needed.

content/docs/command-reference/exp/gc.md
jorgeorpinel: Moved "reachable commit" stuff to #2300. PTAL when/if you can.

content/docs/command-reference/exp/remove.md
pmrowla: I think it is fine to explain that "experiments are stored as custom git references, and are comparable to branch heads, removing an experiment does the same thing as removing a git branch" and stop there.

But even "git-savvy" users should not be getting into things like manually running `git gc`, or thinking about individual `git` commit objects, etc. Mentioning this kind of things in the documentation is just going to invite user questions along the lines of "why doesn't DVC manually delete commit objects" (the answer being that no one should ever need to manually invoke `git gc` ever, git is smart enough to know when a repo should be garbage collected on its own).

For DVC, just telling the user that the experiment git ref has been deleted should be enough. The underlying details of how git refs/dangling or reachable commits/garbage collection/etc work should not matter to a DVC user. A "git savvy" user can do their own research into how git refs work.

I'd liken this to the github documentation for deleting a branch: https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-branches#deleting-a-branch

Nowhere in their docs does github mention that they create their own duplicate custom refs for all branches, and that when you "delete" a branch on github, it only deletes the standard git branch head ref. The custom ref still exists, so that those commits will never be garbage collected. Github also permanently disables git garbage collection entirely for their server-side repos.

Github does this so that forks always have access to upstream commits which existed at the time the fork was made, and so that users can "restore" previously deleted branches. However, the low-level implementation details don't matter to the user. All the user cares about is whether or not the branch looks "deleted" from the user's UI (and git CLI) perspective. Whether or not those underlying commits still exist is not relevant at all.

A "git savvy" user can use the CLI to inspect all of the refs in the repo and figure out what github is doing, but github does not need to document it.

jorgeorpinel: I don't think we can decide what users should be doing or what matters to them, even if we wanted. E.g. that was an argument in https://github.com/iterative/dvc/issues/5181#issuecomment-765261870 yet users seem to care (#5644).

I wonder why `git gc` exists if it's never needed. 🤷 Unrelated

A bit confuse about the quotes around "git savvy" too. Don't those users exist?

> not relevant at all

This kind of statement is hard to understand. Not relevant to whom? Under what criteria?

But anyway, you're probably right that about this being unnecessary detail. Moving that to https://github.com/iterative/dvc.org/pull/2325#pullrequestreview-619177754 (these changes were rolled back in this PR per your feedback).

jorgeorpinel: > Not relevant to whom? Under what criteria?

Reading that part again I see this was explained. BTW, very interesting info about Github! Now I know.

jorgeorpinel: Gh is a much higher-level tool though, but point taken.

pmrowla: `git gc` exists because it may be needed in very specific, advanced use-case situations, if a user is doing low level repo operations like creating git commit/tree/etc objects themselves

> Running git gc manually should only be needed when adding objects to a repository without regularly running such porcelain commands, to do a one-off repository optimization, or e.g. to clean up a suboptimal mass-import.

Issue #2032, opened 2020-12-18
config: incorrect system config path


johnnychen94:
# Bug Report

## Description

According to the documentation, the system config path for Linux is `/etc/dvc/config`, but it seems like the actual config path is `/etc/xdg/dvc/config`.

```console
DVC version: 1.11.8 (pip)
---------------------------------
Platform: Python 3.8.3 on Linux-4.15.0-128-generic-x86_64-with-glibc2.10
Supports: All remotes
```

I'm using Ubuntu 18.04, if that helps.

efiop:
Let me compile the correct list of values for these here at least...

shcheklein:
@efiop yep, that would help if you could write a summary of the change that is needed. Even here in the ticket should be enough.

jorgeorpinel:
/etc/xdg looks liek an Arch Linux thing. It probably changes by distro? I guess we could just make a note about that for Linux

efiop:
Ok, so according to https://github.com/ActiveState/appdirs/blob/master/appdirs.py (we use site_config_dir and user_config_dir), in a simplified version:
--global:
```
Mac: ~/Library/Preferences/dvc/config
Linux: ~/.config/dvc/config or $XDG_CONFIG_HOME/dvc/config
```
--system
```
Mac: /Library/Preferences/dvc/config
Linux: /etc/xdg/dvc/config or $XDG_CONFIG_DIRS[0]/dvc/config
```

@jorgeorpinel `/etc/xdg` is not an Arch thing, it can be on any distro and might even differ between different versions. IIRC i've seen different default XDG_CONFIG_DIRS on different versions of ubuntu before. It can be set to `/etc` too, but it is rather uncommon these days, at least from what've experienced (not at all an expert opinion) :slightly_smiling_face:

pmrowla:
So I just noticed that the docs for mac are incorrect still. 

On mac, the global and system paths should be:
```
global: ~/Library/Application Support/dvc/config
system: /Library/Application Support/dvc/config
```

The discrepancy between using `Application Support` and `Preferences` is that appdirs master branch source uses `Preferences`, but the last actual pypi release uses `Application Support`.

@jorgeorpinel

jorgeorpinel:
Thakns @pmrowla updated. But re `appdirs master branch source uses Preferences`, so does `dvc` installed in any other way (not `pip`) use `Preferences/`? Thanks

pmrowla:
[...]

PR #2302, opened 2021-03-15
ref: improve add/import* to-cache/remote info and examples

jorgeorpinel:
> From relatively recent discussions

Related:
- ~~Clo/ses #2032~~
- ~~#2293~~
- Addresses https://github.com/iterative/dvc.org/pull/2172#pullrequestreview-595022222

- [ ] Make #2317 reviewable when this is merged.


content/docs/command-reference/import.md
```diff
@@ -27,20 +27,25 @@ target file or directory (found at `path` in `url`), and tracks it in the local
...
+> download the data).
+
 The imported data is <abbr>cached</abbr>, and linked (or copied) to the current
 working directory with its original file name e.g. `data.txt` (or to a location
 provided with `--out`). An _import `.dvc` file_ is created in the same location
 e.g. `data.txt.dvc` – similar to using `dvc add` after downloading the data.
 
-(ℹ️) DVC won't push or pull imported data to/from
-[remote storage](/doc/command-reference/remote), it will rely on it's original
-source.
```
jorgeorpinel, line 43: This was wrong right? `import-url` states the opposite.

content/docs/command-reference/import.md
```diff
@@ -27,20 +27,25 @@ target file or directory (found at `path` in `url`), and tracks it in the local
...
+💡 Using an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+lets you transfer an import there (and link it in the <abbr>workspace</abbr>),
+without using the local file system.
+
+> Note that imported data can be [pushed](/doc/command-reference/push) and
+> [pulled](/doc/command-reference/pull) to/from
+> [remote storage](/doc/command-reference/remote) normally.
```
jorgeorpinel, line 48: (Replaced with this)

jorgeorpinel (commented):
Just one Q on a fix to the `import` ref. Cc @efiop in case you're around (I'll double check this myself tmr if not so no worries). Thanks

content/docs/command-reference/import-url.md
```diff
@@ -37,25 +36,27 @@ later, if/when the external data source changes. Example scenarios:
...
-update the import later, if the data source has changed (see `dvc update`).
+similar to using `dvc add` after downloading the data. It saves the information
+about the data source, so the import can be updated later if the data source has
+changed (see `dvc update`).
+
+💡 Using an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+or the `--to-remote` option lets you
+[transfer](#example-transfer-to-remote-storage) an import without using the
+local file system.
```
jorgeorpinel, line 55: Note: I realized the to-cache strategy also applies to regular `import` (I think), so I added this note.

content/docs/command-reference/import-url.md
```diff
@@ -360,40 +362,33 @@ Running stage 'prepare' with command:
...
-$ mkdir example # workspace
-$ cd example
-$ git init
-$ dvc init
-$ mkdir /tmp/dvc-storage
-$ dvc remote add myremote /tmp/dvc-storage
-```
+One option is to setup an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+in a location that can handle the data. Another is to use the `--to-remote`
```
jorgeorpinel, line 371: I wonder if we need a full to-cache example in `import-url` (like in https://dvc.org/doc/command-reference/add#example-transfer-to-the-cache).

content/docs/command-reference/import.md
```diff
@@ -27,20 +27,25 @@ target file or directory (found at `path` in `url`), and tracks it in the local
...
 provided with `--out`). An _import `.dvc` file_ is created in the same location
 e.g. `data.txt.dvc` – similar to using `dvc add` after downloading the data.
 
-(ℹ️) DVC won't push or pull imported data to/from
-[remote storage](/doc/command-reference/remote), it will rely on it's original
-source.
+💡 Using an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+lets you transfer an import there (and link it in the <abbr>workspace</abbr>),
+without using the local file system.
```
jorgeorpinel, line 44: Note that I realized the to-cache strategy also applies to regular `import` (I think), so I added this note.

content/docs/command-reference/import.md
pmrowla: `import` and `import-url` have different behavior, the original docs for each command were correct

shcheklein:
@jorgeorpinel please split this into 3-4 PRs. It has too many things at once. Makes it hard to review.

content/docs/command-reference/add.md
```diff
@@ -336,28 +338,21 @@ $ tree .dvc/cache
...
-$ cd example
-$ git init
-$ dvc init
-```
+to a given path inside the <abbr>workspace</abbr>. Combined with an
+[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
+setup, this let's you avoid using your local file system completely.
 
-Now we can add a `data.xml` file via HTTP for example, putting it a local path
+For example, we can add a `data.xml` file via HTTP, outputting it a local path
```
isidentical, line 355: It seems a bit ambiguous, maybe use with "to"
```suggestion
For example, we can add a `data.xml` file via HTTP, outputting it to a local path
```

content/docs/command-reference/import-url.md
isidentical: > I wonder if we need a full to-cache example in import-url

`import-url` doesn't have the `to-cache` ability, but rather work on its own (like you just import an URL with it and it puts it to your cache, but not in a chunked way).

content/docs/command-reference/import.md
jorgeorpinel: Interestinggggg 🤔 OK will roll back, thanks! ⌛

content/docs/command-reference/import.md
jorgeorpinel: Rolled back (with some clarifications) in fa82c89d.

content/docs/command-reference/import-url.md
jorgeorpinel: Ah, good point. But then maybe we should explain about the chunking (transferring) a bit more... I'll review again ⌛

jorgeorpinel:
OK @shcheklein sorry about that. I removed the changes to `config` and `plots` from here for now. Everything else is a single change (updated PR title too).

content/docs/command-reference/import-url.md
jorgeorpinel: But with `--to-cache` the chinked transfer does happen, right @isidentical ? Just double checking. Already updated in c3932123, PTAL.

content/docs/command-reference/import.md
jorgeorpinel: Done. Please note iterative/dvc/issues/5623 when you have a change. Thanks

content/docs/command-reference/add.md
jorgeorpinel: Yep thanks @isidentical that's a typo!

content/docs/command-reference/add.md
```diff
@@ -153,10 +153,12 @@ not.
   > Additionally, this typically requires an external cache setup (see link
   > above).
 
-- `-o <path>`, `--out <path>` - destination `path` to make a local target copy,
-  or to [transfer](#example-transfer-to-cache) an external target into the cache
-  (and link to workspace). Note that this can be combined with `--to-remote` to
-  avoid storing the data locally, while still adding it to the project.
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace to link
+  (or copy) a data target, which will now be tracked by DVC. Note that combining
```
shcheklein, line 157: I'm not sure that this Note is relevant here. If `-o <path>` is generic enough to work in a regular case, we can remove the note.

content/docs/command-reference/add.md
```diff
@@ -153,10 +153,12 @@ not.
...
+  this with an
+  [external cache transfer](#example-transfer-to-an-external-cache), or with the
+  `--to-remote` option, let's you avoid storing an external target locally,
+  while still adding it to the project.
 
 - `--to-remote` - import an external target, but don't move it into the
```
shcheklein, line 163: `import` is not the right term here

shcheklein: I would describe it as `when specified `dvc add` can accept path to an external data (e.g. S3 url, or path on a mounted volume, etc) and it copies data into remote storage w/o caching data into project's cache and w/o bringing it into the workspace. Besides that, it behaves in the same way - it creates a regular `.dvc` file, which can be used with `dvc pull` to get data into workspace when and where it's needed. This options expects (?) `-o` to be provided. It is useful to "bootstrap" the project, when data is too large for the disk the workspace is located on. See example ...`

content/docs/command-reference/add.md
```diff
@@ -153,10 +153,12 @@ not.
...
   workspace, nor cache it. [Transfer it](#example-transfer-to-remote-storage) it
```
shcheklein, line 164: transfer probably is not the right term (we don't remove it, right?)

content/docs/command-reference/add.md
```diff
@@ -153,10 +153,12 @@ not.
...
   > above).
 
-- `-o <path>`, `--out <path>` - destination `path` to make a local target copy,
-  or to [transfer](#example-transfer-to-cache) an external target into the cache
-  (and link to workspace). Note that this can be combined with `--to-remote` to
-  avoid storing the data locally, while still adding it to the project.
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace to link
+  (or copy) a data target, which will now be tracked by DVC. Note that combining
+  this with an
+  [external cache transfer](#example-transfer-to-an-external-cache), or with the
```
shcheklein, line 159: external cache probably is not the best term here - remote storage is better

content/docs/command-reference/add.md
```diff
@@ -336,29 +338,22 @@ $ tree .dvc/cache
...
 
-When you have a large dataset in an external location, you may want to add it to
-the <abbr>project</abbr> without having to copy it into the workspace. Maybe
-your local disk doesn't have enough space, but you have setup an
-[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
-that could handle it.
+Sometimes you may want to add a large dataset currently found in an external
+location, so it becomes local to the project. However, your local file system
+may not have enough space to download it — which is needed to add data in DVC,
+right? Not necessarily!
```
shcheklein, line 346: not sure about `Not necessarily!` ... Not necessarily _what_?

content/docs/command-reference/add.md
shcheklein: should it link to the ` Transfer to remote storage`?

content/docs/command-reference/add.md
```diff
@@ -368,9 +363,10 @@ data.xml data.xml.dvc
 ```
 
 The resulting `.dvc` file will save the provided local `path` as if the data was
-already in the workspace, while the `md5` hash points to the copy of the data
-that has now been transferred to the <abbr>cache</abbr>. Let's check the
-contents of `data.xml.dvc` in this case:
```
shcheklein, line 373: We can clarify that if links are set properly `data.xml` is "virtual" and doesn't take any space

content/docs/command-reference/add.md
```diff
@@ -384,43 +380,37 @@ outs:
 
 ## Example: Transfer to remote storage
 
-When you have a large dataset in an external location, you may want to track it
-as if it was in your project, but without downloading it locally (for now). The
-`--to-remote` option lets you do so, while storing a copy
-[remotely](/doc/command-reference/remote) so it can be
-[pulled](/doc/command-reference/plots) later. Let's initialize a DVC project,
-and setup a remote:
+Similarly to the previous scenario, you may sometimes want to track a large
```
shcheklein, line 383: track -> add

content/docs/command-reference/add.md
```diff
@@ -384,43 +380,37 @@ outs:
 
 ## Example: Transfer to remote storage
```
shcheklein, line 381: Adding large external data (straight to remote) ?

content/docs/command-reference/get.md
```diff
@@ -24,13 +24,13 @@ repository (e.g. source code, small image/other files). `dvc get` copies the
...
 directory. (Analogous to `wget`, but for repos.)
 
+> See `dvc list` for a way to browse repository contents to find files or
+> directories to download.
+
 > Note that unlike `dvc import`, this command does not track the downloaded
 > files (does not create a `.dvc` file). For that reason, it doesn't require an
 > existing DVC project to run in.
 
-> See `dvc list` for a way to browse repository contents to find files or
```
shcheklein, line 31: not related to this PR?

content/docs/user-guide/managing-external-data.md
```diff
@@ -1,13 +1,14 @@
...
-> strategies of `dvc add` and `dvc import-url` are more convenient. **Note**
-> that external outputs are not pushed or pulled from/to
+> recommended except if there's absolutely no other alternative. Note that
+> external outputs are not pushed or pulled from/to
 > [remote storage](/doc/command-reference/remote).
+>
+> In most cases the
+> [to-cache](/doc/command-reference/add#example-transfer-to-an-external-cache)
+> or [to-remote](/doc/command-reference/add#example-transfer-to-remote-storage)
+> strategies of `dvc add` and `dvc import-url` are more convenient.
```
shcheklein, line 11: don't like `are more convenient.` but I guess we'll address it later. It's not about convenience to my mind. It's about different workflows around data. And we need to explain to people that in certain cases (unless you indeed write/read directly from/to remote location from code) you don't need this. If you are looking for a way to add data to DVC in regular way (remote storage, dvc pull, dvc push , whatever - find a way to define it ) there are better options

content/docs/command-reference/import-url.md
```diff
@@ -360,40 +362,30 @@ Running stage 'prepare' with command:
...
-`.dvc` file:
-
-```dvc
 $ ls
 data.xml.dvc
 ```
 
-Whenever anyone wants to actually download the imported data (for example from a
-system that can handle it), they can use `dvc pull` as usual:
+The only change in our local <abbr>workspace</abbr> is a the tiny `.dvc` file
```
shcheklein, line 386: a the

content/docs/command-reference/add.md
jorgeorpinel: OK sure, having the note in the description and the examples should be enough. Simplified (and made consistent with `get/import -o`).

content/docs/command-reference/add.md
jorgeorpinel: Good catch! Updated along those lines in d63b07f3.

> Besides that, it behaves in the same way...

This seems like too many details for an option desc. It's already explained in the main command Description and I think it's assumed that everything still applies with every option except if otherwise noted in each option desc.

> This options expects (?) -o

Not required, it should use `./<basename>` by default.

> useful to "bootstrap" ...

Similarly this concept is described in the example (linked from the option desc.)

content/docs/command-reference/add.md
jorgeorpinel: > transfer probably is not the right term

Yeah I was thinking that too earlier 🤔 but we already use that term a lot and it's kind of nice. Let me think ⌛

content/docs/command-reference/add.md
jorgeorpinel: This whole note was already removed per https://github.com/iterative/dvc.org/pull/2302#pullrequestreview-612504838 above.

content/docs/command-reference/add.md
jorgeorpinel: To download it it (the large dataset) is not "necessarily needed" to add data in DVC — using the exact language from the p above.

But anyway, I rephrased this in simpler terms 👍

content/docs/command-reference/add.md
jorgeorpinel: Good call. Mentioned in the 2nd paragraph and here.

content/docs/command-reference/add.md
jorgeorpinel: I agree that both "Transfer to an external cache" and "Transfer to remote storage" focus on the method rather than the problem/solution but "Adding large external data" can refer to --external or --out or --to-remote. Let me think ⌛

content/docs/command-reference/get.md
jorgeorpinel: I did some general copy edits to the descriptions of all the commands related here (add, get, import, etc.) as I reviewed the -o/-to-remote stuff.

content/docs/user-guide/managing-external-data.md
jorgeorpinel: Agreed. Changing to "better" for now.

content/docs/command-reference/add.md
jorgeorpinel: OK I removed the term in favor of more explicit phrases like "store a copy".

jorgeorpinel: \* the term "transfer"

content/docs/command-reference/add.md
jorgeorpinel: Alright I re-did the titles to focus on the problem being solved, PTAL @shcheklein (at everything) when you can.

content/docs/command-reference/update.md
```diff
@@ -49,11 +49,12 @@ $ dvc update --rev master
...
 
-- `--to-remote` - update a `.dvc` file created with `dvc import-url` and
-  [transfer](/doc/command-reference/import-url#example-import-straight-to-the-remote)
-  the new data directly to remote storage (the default one unless `-r` is used).
-  No changes are done in the <abbr>workspace</abbr>. Use `dvc pull` to get the
-  data locally. This option can't be used with DVC or Git repository imports.
+- `--to-remote` - update a `.dvc` file created with `dvc import-url` and store
+  the latest data directly
+  [on remote storage](/doc/command-reference/import-url#straight-to-remote) (the
+  default one unless `-r` is specified). No changes are done in the
```
shcheklein, line 55: `No changes are done` contradicts to the `update a `.dvc` I think

content/docs/command-reference/update.md
```diff
@@ -49,11 +49,12 @@ $ dvc update --rev master
...
+  <abbr>workspace</abbr>. Use `dvc pull` to get the data locally. This option
+  can't be used with DVC or Git repository imports.
```
shcheklein, line 57: I would be more explicit - with import done with `dvc import` or something like this.

content/docs/command-reference/push.md
```diff
@@ -57,6 +57,8 @@ Note that the `dvc status -c` command can list files tracked by DVC that are new
 in the cache (compared to the default remote.) It can be used to see what files
 `dvc push` would upload.
 
+> Note that `dvc push` doesn't upload `dvc import` data.
```
shcheklein, line 60: not related to this PR

also, seems to be in a place that no one would ever notice this? if we care about search - we can put in the `dvc import`?

content/docs/command-reference/pull.md
```diff
@@ -74,6 +74,9 @@ Note that the command `dvc status -c` can list files referenced in current
 stages (in `dvc.yaml`) or `.dvc` files, but missing from the cache. It can be
 used to see what files `dvc pull` would download.
 
+> Note that in the case of `dvc import` data, `dvc pull` downloads from the
```
shcheklein, line 77: same here as in `dvc push` - not related, probably not the best place

content/docs/command-reference/add.md
```diff
@@ -145,28 +145,31 @@ not.
   [pattern](https://docs.python.org/3/library/glob.html) specified in `targets`.
   Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`, and `**`
 
-- `--external` - allow `targets` that are outside of the DVC repository. See
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace to link
+  (or copy) a data target (instead of using the current working directory).
+  Directories specified in the path will be created by this command. Note that
+  this can be used [with an external cache](#straight-to-cache) to avoid using
+  the disk.
```
shcheklein, line 152: `the disk` - why disk?

shcheklein: `instead of using the current working directory` - not exactly correct. I think it's instead of some default path + default file name?

content/docs/command-reference/add.md
```diff
@@ -145,28 +145,31 @@ not.
...
+
+- `--to-remote` - allow a target outside of the DVC repository (e.g. an S3
+  object, SSH directory URL, file on mounted volume, etc.) but don't move it
+  into the workspace, nor cache it. [Store a copy](#straight-to-remote) on a
+  remote instead (the default one unless `-r` is specified). Use `dvc pull` to
+  get the data locally later.
+
+- `-r <name>`, `--remote <name>` - name of the
+  [remote](/doc/command-reference/remote) to store an external target on (can
```
shcheklein, line 161: can we use data instead of `external target`?

content/docs/command-reference/get.md
shcheklein: it increases surfaces, please please let's try to avoid it as much as we can

content/docs/command-reference/add.md
jorgeorpinel: "disk space", "to disk", etc. are common ways to refer to machine storage in general I think. I guess "the disk" not so much haha. Changed.

> instead of some default path + default file name

It's `./basename`, that's what I mean by "the cwd". Changed the sentence to "to place a data target" so it's clearer.

Thanks

content/docs/command-reference/add.md
jorgeorpinel: Good idea. Updated (also in `import-url`).

content/docs/command-reference/get.md
jorgeorpinel: Sure, will extract. This is a common misunderstanding because I see these changes as related when I make them, as docs have lots of interconnections... Also I don't always have the capacity to remember small details for later and it's it's not always feasible to stash inner file chunks either. So it's tricky, sorry if this happens now and then 😢

content/docs/command-reference/update.md
jorgeorpinel: Changed to "Tracked data is not changed".

content/docs/command-reference/update.md
jorgeorpinel: Updated.

content/docs/command-reference/get.md
jorgeorpinel: Ok, rolled back.

content/docs/command-reference/push.md
jorgeorpinel: Rolled back.

content/docs/command-reference/pull.md
jorgeorpinel: Rolled back too.

content/docs/command-reference/push.md
jorgeorpinel: > we can put in the dvc import

BTW this note is already found in `import`.

jorgeorpinel: Also, that note had some errors which I had fixed in this PR. But I rolled back the fix for now, as it may be unrelated.

content/docs/command-reference/import.md
```diff
@@ -70,19 +70,19 @@ enable DVC efficiently determining whether the local copy is out of date.
...
 repo at `url`) are not supported.
 
 ## Options
 
-- `-o <path>`, `--out <path>` - specify a path to the desired location in the
-  workspace to place the downloaded file or directory (instead of using the
-  current working directory). Directories specified in the path must already
-  exist, otherwise this command will fail.
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace to place
+  the downloaded file or directory (instead of using the current working
```
shcheklein, line 83: is instead of the current directory? or is it instead of a default file name in the current directory? can we phrase it a bit simple, explicit. `(by default it creates the file in the directory you run the command in and it has a name ...)`

content/docs/command-reference/add.md
```diff
@@ -336,91 +339,78 @@ $ tree .dvc/cache
 Only the hash values of the `dir/` directory (with `.dir` file extension) and
 `file2` have been cached.
 
-## Example: Transfer to the cache
+## Example: Caching large data externally {#straight-to-cache}
```
shcheklein, line 342: hmm ... the name is still very confusing to me. have discussed something like `Add large data from an external location`?

content/docs/command-reference/add.md
```diff
@@ -145,28 +145,31 @@ not.
   [pattern](https://docs.python.org/3/library/glob.html) specified in `targets`.
   Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`, and `**`
 
-- `--external` - allow `targets` that are outside of the DVC repository. See
+- `-o <path>`, `--out <path>` - destination `path` inside the workspace to place
+  a data target (instead of using the current working directory). Directories
+  specified in the path will be created by this command. Note that this can be
```
shcheklein, line 150: > Note that this can be  used [with an external cache](#straight-to-cache) to avoid using the local   file system.

it's not very clear ...

This can be especially useful when you use an external cache setup with symlinks. To add large data from an external (to the repo) location w/o taking or requiring enough space ...

shcheklein: do we need to start with `allow a target outside of the DVC repository` as we do in `--to-remote`?

content/docs/command-reference/add.md
```diff
@@ -145,28 +145,31 @@ not.
...
+  used [with an external cache](#straight-to-cache) to avoid using the local
+  file system.
+
+- `--to-remote` - allow a target outside of the DVC repository (e.g. an S3
+  object, SSH directory URL, file on mounted volume, etc.) but don't move it
+  into the workspace, nor cache it. [Store a copy](#straight-to-remote) on a
+  remote instead (the default one unless `-r` is specified). Use `dvc pull` to
+  get the data locally later.
```
shcheklein, line 158: remove locally? it can be on a remote machine?

shcheklein: Similar to `-o`, it can be useful when you need to add large data and don't have enough space on machine/volume you run this command on (e.g. to bootstrap the project, to update some data .... -

content/docs/command-reference/add.md
```diff
@@ -336,91 +339,78 @@ $ tree .dvc/cache
...
-## Example: Transfer to the cache
+## Example: Caching large data externally {#straight-to-cache}
 
-When you have a large dataset in an external location, you may want to add it to
-the <abbr>project</abbr> without having to copy it into the workspace. Maybe
-your local disk doesn't have enough space, but you have setup an
-[external cache](/doc/use-cases/shared-development-server#configure-the-external-shared-cache)
-that could handle it.
+Sometimes you may want to add a large dataset currently found in an external
+location, so it becomes local to the project. But what if there's not enough
```
shcheklein, line 345: `so it becomes local to the project` - it will be hard to understand to be honest? can we make somehow explicit?

content/docs/command-reference/add.md
```diff
@@ -336,91 +339,78 @@ $ tree .dvc/cache
...
 
 $ ls
 data.xml data.xml.dvc
 ```
 
-The resulting `.dvc` file will save the provided local `path` as if the data was
-already in the workspace, while the `md5` hash points to the copy of the data
-that has now been transferred to the <abbr>cache</abbr>. Let's check the
-contents of `data.xml.dvc` in this case:
+The local `data.xml` should be a symlink or hard link to the externally
```
shcheklein, line 367: hardlink?

content/docs/command-reference/add.md
```diff
@@ -336,91 +339,78 @@ $ tree .dvc/cache
...
     nfiles: 1
-    path: data.xml
+    path: raw/data.xml
 ```
 
 > For a similar operation that actually keeps a connection to the data source,
 > please see `dvc import-url`.
 
-## Example: Transfer to remote storage
+## Example: Storing large data remotely {#straight-to-remote}
```
shcheklein, line 382: name is also still confusing, doesn't reflect an actual use case (we store data remotely anyway). This is about adding data ... should be something related to that

content/docs/command-reference/add.md
```diff
@@ -336,91 +339,78 @@ $ tree .dvc/cache
...
-When you have a large dataset in an external location, you may want to track it
-as if it was in your project, but without downloading it locally (for now). The
-`--to-remote` option lets you do so, while storing a copy
-[remotely](/doc/command-reference/remote) so it can be
-[pulled](/doc/command-reference/plots) later. Let's initialize a DVC project,
-and setup a remote:
+Similarly to the previous scenario, you may sometimes want to add a large
+dataset found externally into a regular <abbr>project</abbr> (with a local
+<abbr>cache</abbr>). Can it be done without downloading the data locally (for
+now)? Yes!
```
shcheklein, line 387: `(for now)` - not sure I understand this 🤔

content/docs/command-reference/get.md
```diff
@@ -56,10 +56,9 @@ name.
 
 ## Options
 
-- `-o <path>`, `--out <path>` - specify a path to the desired location in the
-  workspace to place the downloaded file or directory (instead of using the
-  current working directory). Directories specified in the path will be created
-  by this command.
+- `-o <path>`, `--out <path>` - destination `path` to place the downloaded file
+  or directory (instead of using the current working directory). Directories
```
shcheklein, line 60: so, does it expect file name or a directory path?

content/docs/command-reference/import.md
jorgeorpinel: It says "to place" (the data in) but sure, I can rephrase in a clearer way... ⌛

jorgeorpinel: Done.

content/docs/command-reference/get.md
jorgeorpinel: > does it expect file name or a directory path?

Both work. You can give a dir path (ending in `/` if it doesn't yet exist) and `get/import` will create `<path>/<basename>`. Any path not ending in `/` will be the final imported data path.

jorgeorpinel: So I think "destination `path` to place the downloaded file or directory" is correct and not sure we want to make this description very long by adding all those details, which seem intuitive/expected to me.

content/docs/command-reference/add.md
jorgeorpinel: I updated this to "Note that for external targets, this can be combined with an external cache to skip the local file system." when addressing previous feedback. Seems to make enough sense like that?

The link config info seems too detailed and its explained in the link, which people will need to go to anyway for this I think.

> do we need to start with allow a target outside of the DVC repository

jorgeorpinel: > do we need to start with allow a target outside of the DVC repository

--out works for local targets too (renames them).

jorgeorpinel: That said I'm not sure here. Feels like `add -o` is almost totally different with local vs external targets so yeah, I could write a longer explanation here as a 2nd paragraph including info about links and allowing external targets.

content/docs/command-reference/add.md
jorgeorpinel: > Add large data from an external location

The problem with that is that it applies also to straight-to-remote (next example). Need to think... 🧠

content/docs/command-reference/add.md
jorgeorpinel: > we store data remotely anyway

True. Let me think ⌛ 

> This is about adding data

It's the `add` reference so no nee to use the word "add" in every example title, I think.

content/docs/command-reference/add.md
jorgeorpinel: OK, let's go with "Large data from an external location" here since the cache is the base case, and figure something out for the straight-to-remote example title.

content/docs/command-reference/add.md
jorgeorpinel: > we store data remotely anyway

Actually, while true, the context here is not the `push` or `pull` commands, so storing data remotely is not necessarily assumed. Remote storage isn't even mentioned in the main Description.

But anyway, how about "Bootstrap external data tracking with remote storage"? (applied for now) It's a bit long though...

jorgeorpinel:
@shcheklein on the example titles, I changed them again after thinking about it:

`add`:
- **Example: Large external data** (straight-to-cache)
- **Example: External data onto remote storage** (straight-to-remote)

> Word "Adding" left out for consistency with previous examples e.g. "Example: Single file" but no strong opinion. Could also start use "Tracking".

`import`:
- **Example: Importing onto remote storage** (straight-to-remote)

content/docs/command-reference/add.md
jorgeorpinel: Moved discussion to https://github.com/iterative/dvc.org/pull/2302#issuecomment-801442953, resolving here.

content/docs/command-reference/add.md
jorgeorpinel: -> https://github.com/iterative/dvc.org/pull/2302#issuecomment-801442953
Resolving

content/docs/command-reference/add.md
jorgeorpinel: Removed `locally` and extended the motivation with "to skip the local file system". The link has all the details but if you prefer I can expand on this even more (may need a 2nd p).

content/docs/command-reference/add.md
jorgeorpinel: "Hard link" is much more common, it seems (https://trends.google.com/trends/explore?cat=32&geo=US&q=hardlink,hard%20link).

content/docs/command-reference/add.md
jorgeorpinel: Agreed. I removed "so it becomes local to the project". The next sentence ("But what if there's not enough disk space to download") implies that we mean to add locally.

content/docs/command-reference/add.md
jorgeorpinel: Removed `(for now)`.

It was a reference to the fact that later you can `pull` it. But that's explained later in the example so no need for the obscure parenthesis 👍

content/docs/command-reference/add.md
shcheklein: it's about consistency in the docs, and DVC itself names them hardlink, symlink, etc. And I think it's acceptable in a sense that people do use `hardlink` as a term

shcheklein: btw, hardlinks most likely don't work in this scenario (people will have their cache on a different volume)

jorgeorpinel: > it's about consistency in the docs

It looks like we have a mix currently (was planning to change all to "hard link" 😅). We also have a mix of "symlink" and "soft link"...

> hardlinks most likely don't work in this scenario

OK. Since that's the case, I just removed the mention, and we can leave everything else as is.

PR #2311, opened 2021-03-18
Documentation: Update path to system config file

michaelstepner:
The updated path for the system config file matches the path that is listed in https://dvc.org/doc/command-reference/config

### DIFF
--- a/content/docs/command-reference/cache/dir.md
+++ b/content/docs/command-reference/cache/dir.md
@@ -32,8 +32,8 @@ cache directory.
 - `--global` - modify the global config file (e.g. `~/.config/dvc/config`)
   instead of the project's `.dvc/config`.
 
-- `--system` - modify the system config file (e.g. `/etc/xdg/dvc/config`) instead of
-  `.dvc/config`.
+- `--system` - modify the system config file (e.g. `/etc/xdg/dvc/config`) instead
+  of `.dvc/config`.
 
 - `--local` - modify a local [config file](/doc/command-reference/config)
   instead of `.dvc/config`. It is located in `.dvc/config.local` and is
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/cache/dir.md
```diff
@@ -32,7 +32,7 @@ cache directory.
 - `--global` - modify the global config file (e.g. `~/.config/dvc/config`)
   instead of the project's `.dvc/config`.
 
-- `--system` - modify the system config file (e.g. `/etc/dvc/config`) instead of
+- `--system` - modify the system config file (e.g. `/etc/xdg/dvc/config`) instead of
   `.dvc/config`.
```
jorgeorpinel, line 36: This is consistent with what we have in https://dvc.org/doc/command-reference/config 👍

jorgeorpinel (approved):
We should separately update the other 8 instances of this option description...

content/docs/command-reference/cache/dir.md
jorgeorpinel, line 36: Formatting
```suggestion
- `--system` - modify the system config file (e.g. `/etc/xdg/dvc/config`) instead
  of `.dvc/config`.
```

jorgeorpinel:
Thank you @michaelstepner !

content/docs/command-reference/cache/dir.md
jorgeorpinel: Will update everything else in 315e6a5b.

shcheklein:
@jorgeorpinel looks like we broke the formatting in master.

jorgeorpinel:
Yes, my bad. I formatted manually because Restyled didn't kick in this PR for some reason but I didn't do it right.

PR #2312, opened 2021-03-18
cml: document --single, --reuse and --cloud-startup-script

0x2b3bfa0:
This pull request may fix iterative/cml#450 in conjunction with iterative/cml#451

_Please note that all the table rows have been modified because of spacing, but there isn't any significant non-whitespace addition apart from lines 92 93 and 105_


jorgeorpinel:
Hi @0x2b3bfa0, writing "fix iterative/cml#450" in the description above will [cause that issue to auto-close](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) when this PR is merged (in case you didn't want that, please change the text).

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,24 +83,27 @@ you do it your way :)
...
+| `--labels`               | Comma delimited runner labels                                                                                 | Default: `cml`                                                                                                 |
+| `--idle-timeout `        | Time in seconds for the runner to be waiting for jobs before shutting down. 0 waits forever.                  | Default: `300`                                                                                                 |
+| `--name`                 | Name displayed in the repo once registered                                                                    | Default: `cml-` followed by a unique identifier, i.e. `cml-cfwj9rrari`                                         |
+| `--single`               | If specified, exit after running a single job.                                                                | Boolean                                                                                                        |
```
jorgeorpinel, line 92:
```suggestion
| `--single`               | Exit after running a single job.                                                                | Boolean                                                                                                        |
```

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,24 +83,27 @@ you do it your way :)
...
+| `--reuse`                | Don't launch new runners if there is, at least, another one with the same `--name` or containing `--labels`   | Boolean                                                                                                        |
```
jorgeorpinel, line 93:
```suggestion
| `--reuse`                | Don't launch new runners if one already exists with the same name or labels   | Boolean                                                                                                        |
```

jorgeorpinel: But it's not clear what the condition is. The existing runner must have the same name AND exact same labels? Just name or some of the labels? Would be good to clarify the exact behavior.

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,24 +83,27 @@ you do it your way :)
...
+| `--driver`               | If not specified, driver is inferred from environmental variables                                             | Choices: `github`,`gitlab`                                                                                     |
```
jorgeorpinel, line 94: Is "type of driver" the correct way to say this? Or something like "Driver platform" perhaps?

0x2b3bfa0:
🙏  Thanks for the reminder, @jorgeorpinel! I used a closing word purposefully, as the other linked pull request has been approved and will be merged soon.

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,24 +83,27 @@ you do it your way :)
...
+| `--cloud-ssh-private`    | Your private RSA SSH key. If not provided will be generated by the Terraform-provider-Iterative.              | Accepts string                                                                                                 |
+| `--cloud-spot`           | Request a spot instance                                                                                       | Boolean                                                                                                        |
+| `--cloud-spot-price`     | Spot max price. If not specified it takes current spot bidding pricing.                                       | Default: `-1`                                                                                                  |
+| `--cloud-startup-script` | Run the provided script as part of the instance initialization                                                | String (Base64 encoded shell script)                                                                           |
```
jorgeorpinel, line 105:
```suggestion
| `--cloud-startup-script` | Run the script provided in-line during the instance initialization. Expects a Base64-encoded shell script                                                | String                                                                           |
```

jorgeorpinel: > the instance initialization

Is it only one instance always? Or should it be more general in case of multiple runners e.g. "during runner instance(s) initialization?

Also, would it be important to specify when exactly during initialization this is run?

jorgeorpinel: > shell script

Maybe "Linux shell script"? Or can it be other stuff e.g. BAT (Windows). Also if it's specifically Bash, maybe say "Bash script".

jorgeorpinel (changes requested):
Review on the 2 options being introduced. But all my suggestions would need a reformatting again since md tables are aligned... (Restyled will do it automatically if needed).

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: 🙏  Thanks, @jorgeorpinel! As I tried to express in the original text, there are two disjunctive conditions: for a runner to be reused, it has to either (1) have the same name or (2) contain, at least, the specified labels.

Would the following modification convey the behavior in a clearer way? If not, suggestions are highly welcome.
```suggestion
| `--reuse`                | Don't launch new runners if one already exists with the same name or containing the specified labels   | Boolean                                                                                                        |
```

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Please ignore this comment on --driver for now. Didn't mean to post this yet. Resolving!

0x2b3bfa0: The _driver_ concept is a bit of a stretch and describes a platform-specific continuous integration runner [(like this)](https://github.com/actions/runner) that's wrapped by our code. I like the idea of "driver platform" even if the platform part becomes more insightful than the driver part.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Its better, thanks! But I'm still not getting it 100% with regards to labels. So it's either the exact same name or the exact same labels? If so then my suggestion (previous comment, which I just updated) should be OK, I think.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Thanks for the info. It will be useful for a separate review I'm preparing... ⌛

content/docs/cml/self-hosted-runners.md
jorgeorpinel: BTW last question on --reuse:

> at least one already exists

Meaning one runner on the same platform (driver)? Or across all drivers? (This may be obvious but I'm not familiar enough with CML yet)

0x2b3bfa0: @jorgeorpinel, the issue is that, when reusing a runner by labels we don't require an exact match. _Id est:_ if an existing runner [contains](https://en.wikipedia.org/wiki/Subset) all the given labels among many others, we still consider it a match.

0x2b3bfa0: > _Meaning one runner on the same platform (driver)? Or across all drivers? (This may be obvious but I'm not familiar enough with CML yet)_

For users, the so-called drivers are just the name of their continuous integration vendor, like GitHub Actions, GitLab CI/CD, Azure Pipelines, CircleCI... We only support one of these at a time, so «one runner on the same platform» would be the right meaning.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: > _Maybe "Linux shell script"?_

For the _connoisseur_, the term [_shell script_](https://en.wikipedia.org/wiki/Shell_script) doesn't leave any doubt about its more-or-less POSIX-compliant nature, but this is just one of my attrocious nitpicking exercises. 😂  I would be tempted to use Bash as you suggested, but the fun fact is that we're using the good old [Bourne shell](https://en.wikipedia.org/wiki/Bourne_shell).

_Linux shell script_ might be the most helpful term for now, but I'll consider migrating to Bash under the hood and updating the documentation. The good old `sh` shell is not precisely famous for its syntactic sugar.

0x2b3bfa0: > _Is it only one instance always? Or should it be more general in case of multiple runners e.g. "during runner instance(s) initialization?_

Currently we only support launching a single instance per invocation, so the singular would be pretty safe.

0x2b3bfa0: > _Also, would it be important to specify when exactly during initialization this is run?_

The technically inaccurate answer is ["nearly in the middle"](https://github.com/iterative/terraform-provider-iterative/blob/d54e3415ba94fcc7b7c28a9b989b4a21289950dc/iterative/resource_runner.go#L225-L227) between the instance configuration and the runner execution:

1. Install all the base packages and configure the instance operating system if required.
2. Run the custom script and wait until it finishes.
3. Configure and launch the CML self-hosted runner.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Cools, so maybe
```suggestion
| `--reuse`                | Don't launch a new runner if at an existing one has the same name, or overlapping labels   | Boolean                                                                                                        |
```

content/docs/cml/self-hosted-runners.md
jorgeorpinel: OK no need for so much detail in docs involving the script execution order I guess. I edited my suggestion above to include your other answers. Thanks!

jorgeorpinel: Should we link to https://www.base64decode.org/ or similar tool from "Base64-encoded" ? Or perhaps this is explained/linked to from some other CML doc.

jorgeorpinel:
@0x2b3bfa0 feel free to commit the updated suggestions above per our conversation, thanks.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: > _if **at** an_

Is that `at` a residue from the original sentence?

0x2b3bfa0:
```suggestion
| `--reuse`                | Don't launch a new runner if an existing one has the same name, or overlapping labels   | Boolean                                                                                                        |
```

jorgeorpinel: > if at an

Yeah that was a typo sry 😅

jorgeorpinel (approved):
LGTM

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: As far as I know, the CML documentation doesn't include any user-friendly explanation of Base64. Anyhow, I would prefer not to link to any specific on-line tool.

If you consider it neccessary, we can recommend the use of the `base64` command-line tool and link to something neutral, like the pertaining Wikipedia article, but I presume that users searching for this feature will be somewhat familiar with this encoding.

jorgeorpinel:
So is this already released and ready to be published? If so please merge #2315 (approved).

0x2b3bfa0:
Captura de pantalla 2021-03-18 a las 22 03 48

0x2b3bfa0:
Yes, the code was released early today, so we're good to merge the documentation. Unfortunately, I can't apply the restyled.io pull request. 😅

content/docs/cml/self-hosted-runners.md
jorgeorpinel: > CML documentation doesn't include any user-friendly explanation of Base64

Might be something to make an issue to improve on. Probably not a good idea to assume users will know what it even means.

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,24 +83,27 @@ you do it your way :)
...
+| `--version`              | Show version number                                                                                           | Boolean                                                                                                        |
+| `--labels`               | Comma delimited runner labels                                                                                 | Default: `cml`                                                                                                 |
+| `--idle-timeout `        | Time in seconds for the runner to be waiting for jobs before shutting down. 0 waits forever.                  | Default: `300`                                                                                                 |
+| `--name`                 | Name displayed in the repo once registered                                                                    | Default: `cml-` followed by a unique identifier, i.e. `cml-cfwj9rrari`                                         |
```
jorgeorpinel, line 91: I noticed the `Values` column is a bit inconsistent. Sometimes it has a var type e.g. `String` and other times (like here) it specifies the default string instead. Further down there are some defaults in the `Description` too. Would be good to make it consistent one way or another.

Commenting post-merge since it's unrelated to the original intention of the PR.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Agreed! In fact, we probably should start by reviewing if the special character issues we're solving with Base64 are worth the impact from the user experience standpoint.

Note to myself: write a nice issue about this concern on the CML repository.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: I'm inclined to think that the best long-term solution would be replacing this section with automatically generated documentation as per my comment on https://github.com/iterative/cml/issues/361.

In the meanwhile, I'll take a look to the table tomorrow and try to unify the current row style sample book.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Up to you guys @0x2b3bfa0 @DavidGOrtega unfort. currently the docs team has to focus mostly on DVC so can't follow up on this, but hopefully soon we will be able to expand a little and help you more closely 🙂

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved through #2321

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Awesome! Moved to https://github.com/iterative/terraform-provider-iterative/issues/91 in the meanwhile

PR #2317, opened 2021-03-19
ref: get/import/list updates

### LINKED ISSUE #2309: DVC list not working with remote google cloud storage
There are two things here:

1. I expected to be able to just do `dvc list` and it would pick up the default remove.
2. `dvc list gs://...` is not working (cloning is failing)

I can list the content by using `dvc status --remote` so I would like to be able to do the same with listing.

jorgeorpinel:
> Extracted from #2302

Should close #2309

- [x] Change base to `master` when #2302 is merged.


jorgeorpinel:
@shcheklein the changes to `list` are mainly per #2309, the rest are generic copy edits to get/import (but everything is closely related, I think). Thanks

shcheklein (approved):

PR #2321, opened 2021-03-21
cml: improve command documentation

0x2b3bfa0:
This pull request proposes a hopefully consistent structure for the command-line argument table on https://dvc.org/doc/cml/self-hosted-runners#arguments, mainly by splitting the _values_ field into more specific _type_ and _default_ properties.

👉  [**Click here to see the rich text diff**](https://github.com/iterative/dvc.org/pull/2321/files?short_path=f7d2e85#diff-f7d2e8565d1860d11c399a2709d27ffa05588fc58bba366782c13246731e87cd) 👈 

_Note: this is a follow-up of [this comment](https://github.com/iterative/dvc.org/pull/2312#discussion_r597278666) on #2312._


jorgeorpinel:
> 👉 Click here to see the rich text diff 👈

Woah. How did you do that @0x2b3bfa0 ??? Very useful!

jorgeorpinel:
Oh nvmd dah! 

image

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
-| `--cloud-spot-price`     | Spot max price. If not specified it takes current spot bidding pricing.                                         | Default: `-1`                                                                                                  |
-| `--cloud-startup-script` | Run the script provided in-line during the instance initialization. Expects a Base64-encoded Linux shell script | String                                                                                                         |
-| `-h`                     | Show help                                                                                                       | Boolean                                                                                                        |
+| Argument                 | Type                                                                  | Default                                                    | Description                                                                                                                  |
```
jorgeorpinel, line 86: So are these "arguments"? Or flags/options? We call them Options in DVC cmd refs e.g. https://dvc.org/doc/command-reference/add — not saying that's correct though, I'm actually wondering what's the best terminology for this...

For now maybe call this section **Options** and this column **Name** (i.e. option name), avoiding the "argument" term.

jorgeorpinel: > maybe call this section Options and this column Name

Then again it could keep **Arguments** for the title and **Option** for the column, kind of like in https://docs.microsoft.com/en-us/windows/terminal/command-line-arguments?tabs=windows#options-and-commands. Up to you

jorgeorpinel: I guess I'd avoid "argument" for title or column since some options/flags accept arguments of their own.

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
+| ------------------------ | --------------------------------------------------------------------- | ---------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `--version`              | Flag                                                                  | Unset                                                      | Show version number and exit.                                                                                                |
```
jorgeorpinel, line 88: I'd avoif "unset" since it's not clear what that means. Maybe just a `-` or "N/A" is more obvious:

```suggestion
| `--version`              | Flag                                                                  | -                                                      | Show version number and exit (ignoring all other options).                                                                                                |
```

And also, aren't all the args/options "flags"? Some flags just happen to accept an argument themselves. If so I would go back to `Boolean` here for now. I'll lyk if I can think of a better term... ⌛

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
+| `--labels`               | List                                                                  | `cml`                                                      | Comma delimited list of runner labels.                                                                                       |
```
jorgeorpinel, line 89: TBH this whole "Type" column isn't too convincing. All values passed are strings, technically right? May be best to remove it and specify what kind of values are expected as part of the description if needed.

In this case e.g.

```suggestion
| `--labels`               | `cml`                                                      | One or more user-defined labels (delimited with commas).                                                                                       |
```

p.s. Labels are user-defined right? Are spaces allowed between commas?

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
+| `--name`                 | String                                                                | `cml-{identifier}` where `{identifier}` is a random string | Name displayed in the repository once registered.                                                                            |
+| `--single`               | Flag                                                                  | Unset                                                      | Exit after running a single job.                                                                                             |
+| `--reuse`                | Flag                                                                  | Unset                                                      | Don't launch a new runner if an existing one has the same name or overlapping labels.                                        |
+| `--driver`               | Choice: `github` or `gitlab`                                          | Inferred from environment variables                        | Driver for the platform where the repository is hosted.                                                                      |
```
jorgeorpinel, line 94: Another example on how to put type into Description when needed:

```suggestion
| `--driver`               | Inferred from environment variables                        | Platform where the repository is hosted. Accepts `github` or `gitlab`.                                                                      |
```

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
+| `--repo`                 | String                                                                | Inferred from environment variables                        | Git repository address (e.g. https://github.com/iterative/cml)                                                               |
```
jorgeorpinel, line 95:
```suggestion
| `--repo`                 | Inferred from environment variables                        | Git repository URL e.g. https://github.com/iterative/cml                                                               |
```

content/docs/cml/self-hosted-runners.md
```diff
@@ -83,27 +83,27 @@ you do it your way :)
...
+| `--token`                | String                                                                | Inferred from environment variables                        | Personal access token with enough permissions to register a self-hosted runner on the repository.                            |
```
jorgeorpinel, line 96: > Inferred from environment variables

Where are these env vars explained? Just curious 🙂

jorgeorpinel (commented):
Just removing a few unnecessary periods for now (will commit all this, just `git pull` please):

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: As per [POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_01), commands take **options** that may require **option arguments** and anonymous **operands**.  Definitely, it makes sense to rename both the column and the heading.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: I fear being unable to satisfy your curiosity. 😄 

We have them [sprinkled](https://github.com/iterative/cml/blob/5888e49d55662aafd20f62d9acf5ecc2c5478a43/bin/cml-runner.js#L16-L29) around [the code](https://github.com/iterative/cml/blob/81e8a363eb300306c888a77bdeb4d71179f09203/src/cml.js#L49-L55) in order to automatically detect the repository URL, the continuous integration platform and the default token it exposes to every workflow run. They aren't intended for user consumption, but a bit of documentation wouldn't hurt either.

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Yes, they're strings at the shell level, but they get automatically cast to ECMAScript types following the usual logic (?) 🙃  Anyhow, it feels awkward and your proposed solution looks much better.

0x2b3bfa0: Yes, labels are user-defined

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: > _I'd avoid "unset" since it's not clear what that means. Maybe just a `-` or "N/A" is more obvious._

I would prefer to avoid the hyphen because of its `stdin` meaning on many command-line tools. Maybe we can use an [em dash](https://en.wikipedia.org/wiki/Dash#Em_dash) without backticks <kbd>—</kbd> or just leave them blank.

> _And also, aren't all the args/options "flags"? Some flags just happen to accept an argument themselves. If so I would go back to `Boolean` here for now. I'll lyk if I can think of a better term..._ ⌛

Yes, I was thinking of boolean flags (like bit flags), but the word _flag_ is also overloaded to mean _option_ even with arguments. Fortunately enough, we can get rid of the whole **Type** column.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: OK! Can you please update the title to "Options" then? (I can't from GH.) Here are the col renames I rec:

```suggestion
| Name                     | Type                                                                  | Default argument                                           | Description                                                                                                                  |
```

content/docs/cml/self-hosted-runners.md
jorgeorpinel: > I would prefer to avoid the hyphen

Sure. I changed it to "N/A" (does not apply) in my suggestion above.

> Fortunately enough, we can get rid of the whole Type column.

Agreed. I removed that cell in my suggestion ☝️ too.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: Cool.

- [ ] Maybe something else to consider documenting in a single place in the future.

jorgeorpinel:
Hey @0x2b3bfa0 again I think there's no rush here but also wouldn't want it to become stale. Please lmk if you need help wrapping this up on the writing side.

0x2b3bfa0:
Sorry for the delay, @jorgeorpinel, I was quite busy rebasing the Kubernetes support pull request over the most recent changes, and I've this past weekend "chilling out" with baroque from 9 a.m. to 9 p.m.

I'll apply the changes today.

jorgeorpinel:
No problem as long as it's under your radar. Take your time, thanks!

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved with https://github.com/iterative/dvc.org/pull/2321/commits/ca804af226acd6100e23d25430d39b87277ef6f8

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved with https://github.com/iterative/dvc.org/pull/2321/commits/ca804af226acd6100e23d25430d39b87277ef6f8

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved with https://github.com/iterative/dvc.org/pull/2321/commits/ca804af226acd6100e23d25430d39b87277ef6f8

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved with https://github.com/iterative/dvc.org/pull/2321/commits/ca804af226acd6100e23d25430d39b87277ef6f8

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: Resolved with https://github.com/iterative/dvc.org/pull/2321/commits/ca804af226acd6100e23d25430d39b87277ef6f8

0x2b3bfa0:
@jorgeorpinel, I think that I haven't missed anything; please take a look to the new diff.

_Note: I took the liberty of adding [`<pseudo-bnf>`](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) tags to the options that take arguments. Please feel free to revert the change if you don't consider it useful._

0x2b3bfa0:
By the way, I see that you've explicitly removed the final periods on table cells with a single sentence. What would be the consistent way of adapting that policy to multi-sentence cells?

content/docs/cml/self-hosted-runners.md
```diff
@@ -79,31 +79,31 @@ We like the
...
+| `--cloud-region <region>`         | `us-west`                                                  | Region where the instance is deployed. Accepts `us-east`, `us-west`, `eu-west`, `eu-north` or native regions                                          |
+| `--cloud-type <type>`             | N/A                                                        | Instance type. Accepts `m`, `l`, `xl` or native types like `t2.micro`                                                                                 |
+| `--cloud-gpu <type>`              | `nogpu`                                                    | GPU type. Accepts `nogpu`, `k80` or `tesla`.                                                                                                          |
+| `--cloud-hdd-size <size>`         | N/A                                                        | Storage size in GB                                                                                                                                    |
```
jorgeorpinel, line 101: Maybe
```suggestion
| `--cloud-hdd-size <gigs>`         | N/A                                                        | Storage size in GB                                                                                                                                    |
```

content/docs/cml/self-hosted-runners.md
```diff
@@ -79,31 +79,31 @@ We like the
...
+| `--cloud-ssh-private <key>`       | Automatically generated throwaway key                      | Custom private RSA SSH key                                                                                                                            |
+| `--cloud-spot`                    | N/A                                                        | Request a spot instance. If not set, `--cloud-spot-price` won't have any effect                                                                       |
+| `--cloud-spot-price <price>`      | Current spot bidding price                                 | Maximum spot instance bidding price in USD                                                                                                            |
```
jorgeorpinel, line 104: Maybe
```suggestion
| `--cloud-spot-price <usd>`        | Current spot bidding price                                 | Maximum spot instance bidding price in USD                                                                                                            |
```

content/docs/cml/self-hosted-runners.md
```diff
@@ -79,31 +79,31 @@ We like the
...
+| `--cloud-startup-script <script>` | N/A                                                        | Run the provided Base64-encoded Linux shell script during the instance initialization                                                                 |
```
jorgeorpinel, line 105:
```suggestion
| `--cloud-startup-script <text>`   | N/A                                                        | Run the provided Base64-encoded Linux shell script during the instance initialization                                                                 |
```

jorgeorpinel (approved):
Sorry for the delay (Easter week got in the way). LGTM! 🚀 

> adding <pseudo-bnf> placeholders to the options that take arguments

@0x2b3bfa0 Good call 👍 but left a couple suggestions 👇 to rename the pseudo-types so they're not redundant with the option names.

> I see that you've explicitly removed the final periods on table cells with a single sentence.

They weren't sentences. Only full sentences need a period, usually.

content/docs/cml/self-hosted-runners.md
jorgeorpinel: or `<base64>`. Think of users who will only see this part in the `--help` output and need a hint as to what kind of values the option arg accepts.

0x2b3bfa0:
```suggestion
| `--cloud-startup-script <base64>` | N/A                                                        | Run the provided Base64-encoded Linux shell script during the instance initialization                                                                 |
```

I'd go for `base64` as it seems go be higher in the usefulness/byte scale. 😄

content/docs/cml/self-hosted-runners.md
0x2b3bfa0: @jorgeorpinel, what about `gigabytes` without abbreviations?

`<kidding>`
Mostly to avoid triggering my obsessive-compulsive disorder of expanding abbreviations whenever possible and in every context. In my humble opinion, and as far as I know, they only look good to me when the original words are preserved.
`</kidding>`

jorgeorpinel: > what about gigabytes?

Sure
```suggestion
| `--cloud-hdd-size <gigabytes>`    | N/A                                                        | Storage size in GB                                                                                                                                    |
```

jorgeorpinel:
🎉

0x2b3bfa0:
🙏  🚀

0x2b3bfa0:
> _They weren't sentences._

So true. It looks like I'm not ready to pass the Turing test yet. 🤦  🤖  🙃 

> _Only full sentences need a period, usually._

Good to know; thanks! I missed that pattern in my text formatting plugin.

jorgeorpinel:
> Think of users who will only see this part in the --help output and need a hint as to what kind of values the option arg accepts.

BTW! Consider making an update to the actual `cml-runner` core code 🙂

0x2b3bfa0:
Changes will propagate soon through https://github.com/iterative/cml/pull/466. 🙏  Thank you very much!

PR #2325, opened 2021-03-23
exp: review implementation details (run, remove, gc)

jorgeorpinel:
Extracted from https://github.com/iterative/dvc.org/pull/2300#issuecomment-804629685

**Optionally** (or for follow-up PR):

- [ ] Include `exp apply` deets https://github.com/iterative/dvc.org/pull/2243#pullrequestreview-603422799
- [ ] Maybe review first ckp exp run case (see https://github.com/iterative/dvc.org/pull/2286#discussion_r597957922)


content/docs/command-reference/exp/remove.md
```diff
@@ -14,9 +14,22 @@ positional arguments:
...
+> Note that all the checkpoints in an experiment are deleted by this command.
+
+<details>
+
+### What happens to removed experiments?
+
+Internally, `dvc exp` uses actual Git commits under custom Git references (found
+in `.git/refs/exps`). `dvc remove` deletes the references only, rendering the
+commits unreachable. `git` should garbage collect them automatically
+[after some time](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt).
```
jorgeorpinel, line 30: > Continues https://github.com/iterative/dvc.org/pull/2300#pullrequestreview-617332002

This expandable section is in line with the **How does DVC track experiments?** and **How are checkpoints captured by DVC?** sections already published in [`exp run`](https://dvc.org/doc/command-reference/exp/run).

The reason I decided to include them is that `git`-savvy users may want to manipulate experiments manually or even build integrations (and this is probably hard to figure out from the codebase). Note their non-intrusive design:

image

content/docs/command-reference/exp/gc.md
```diff
@@ -40,9 +40,10 @@ separately to delete it.
   the workspace (implies `-w`). Note that both options can be combined, for
   example using the `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ Git commits
-  as well as the workspace (implies `-w`). This is mainly only needed when
-  clearing the experiments run queue.
+- `--all-commits` - same as `-a` or `-T` above, but applies to _all_
+  [reachable](https://git-scm.com/docs/gitglossary#def_reachable) Git commits as
+  well as the workspace (implies `-w`). This is mainly needed when clearing the
+  experiments run queue.
```
jorgeorpinel, line 46: > Continuation of https://github.com/iterative/dvc.org/pull/2300#pullrequestreview-617325868

Linking to "reachable" to match the details added to `exp remove`. Notice it's vague enough that it can mean what we want it to mean, I think (since we don't specify reachable from where):

image

Also, keeping just "all commit" seems incorrect, esp. in the context of experiments which are commits (fact which we explain in docs) but aren't affected by this option.

jorgeorpinel: p.s. on "reachable commits": would "regular commits" work better? Or "your commits"?

Note that this change has already been applied to `exp show -A` (accidentally https://github.com/iterative/dvc.org/pull/2300#pullrequestreview-619179757).

content/docs/command-reference/gc.md
```diff
@@ -62,7 +62,8 @@ The default remote is cleaned (see `dvc config core.remote`) unless the
   versions of an experiment or project. Note that both options can be combined,
   for example using the `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to all Git commits,
+- `--all-commits` - same as `-a` or `-T` above, but applies to all
+  [reachable](https://git-scm.com/docs/gitglossary#def_reachable) Git commits,
```
jorgeorpinel, line 66: Included this one too since we emphasize from Exp docs that `exp gc` only does half of the job.

content/docs/command-reference/exp/gc.md
jorgeorpinel: p.s. And if we keep "regular" or other term, should it be applied to all DVC commands with `--all-commits`? There's another 5-6 I think.

content/docs/command-reference/exp/remove.md
jorgeorpinel: From @pmrowla in https://github.com/iterative/dvc.org/pull/2300#discussion_r600096210:

> I think it is fine to explain that "experiments are stored as custom git references, and are comparable to branch heads, removing an experiment does the same thing as removing a git branch" and stop there.
> I liken this to the github documentation for [deleting a branch](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-branches#deleting-a-branch)... Nowhere in their docs does github mention that they create their own duplicate custom refs.
> A "git savvy" user can use the CLI to inspect all of the refs in the repo and figure out what github is doing

Compelling. Not sure about the exact analogy with GH but still a valid argument. OK, I'll change this PR's scope to simplify the existing implementation detail notes (in `exp run`) instead, and see how to incorporate just a min. of detail here ⌛ 

> Mentioning this kind of things in the documentation is just going to invite user questions along the lines of "why doesn't DVC manually delete commit objects"

Inviting user questions may be a good thing though.

content/docs/command-reference/gc.md
```diff
@@ -62,9 +62,9 @@ The default remote is cleaned (see `dvc config core.remote`) unless the
   versions of an experiment or project. Note that both options can be combined,
   for example using the `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to all Git commits,
-  and in the workspace (implying `-w`). This preserves the cache for all data
-  used in the entire commit history of the project.
+- `--all-commits` - same as `-a` or `-T` above, but applies to all your Git
+  commits, and in the workspace (implying `-w`). This preserves the cache for
+  all data used in the entire commit history of the project.
```
jorgeorpinel, line 67: - [ ] Should be applied to all DVC commands with `--all-commits` (there's another 5 or 6 left) — separate PR.

content/docs/command-reference/exp/run.md
```diff
@@ -35,27 +35,24 @@ data and code updates, or <abbr>hyperparameter</abbr> tuning. You can use the
...
 
 ### How does DVC track experiments?
 
-Internally, `dvc exp` uses actual commits under custom
+Experiments are custom
 [Git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
-(found in `.git/refs/exps`). Each commit has the Git `HEAD` as parent and has
-it's own SHA-256 hash. These are not pushed to the Git remote by default (see
+(found in `.git/refs/exps`) with a commit whose parent is the `HEAD` at the time
+of running. Note that they're not pushed to the Git remote by default (see
```
pmrowla, line 48: Not sure if "at the time of running" needs to be clarified here. Technically it is more like "at the time the experiment is queued". (For regular runs without `--queue/--run-all` it's just that we queue the single exp and execute it immediately afterwards.)

The important thing is that if `HEAD` has moved between `exp run --queue`, and `exp run --run-all`, we still execute each queued experiment based on `HEAD` at the time of `exp run --queue`.

content/docs/command-reference/exp/run.md
```diff
@@ -35,27 +35,24 @@ data and code updates, or <abbr>hyperparameter</abbr> tuning. You can use the
...
 
-> References have a unique signature similar to the
-> [entries in the run-cache](/doc/user-guide/project-structure/internal-files#run-cache).
-
 </details>
 
 The results of the last experiment can be seen in the <abbr>workspace</abbr>. To
 display and compare your experiments, use `dvc exp show` or `dvc exp diff`. Use
-`dvc exp apply` to roll back the workspace to a previous experiment
+`dvc exp apply` to roll back the workspace to a previous experiment.
```
pmrowla, line 55: Maybe need a note here with a reminder that changes from `--queue/--temp` runs are not applied to the workspace.

Also, I think "roll back the workspace..." should maybe be something like "to apply the (git) changes from an experiment to the workspace". To me, using `roll back` sounds like we are reverting to an old changeset, which is not really the case.

content/docs/command-reference/exp/run.md
```diff
@@ -94,12 +91,10 @@ the first `dvc exp run`) — useful for re-training ML models, for example.
...
 
-### How are checkpoints captured by DVC?
+### How are checkpoints captured?
 
-When DVC runs a checkpoint-enabled experiment, a custom Git branch (in
-`.git/refs/exps`) is started off the repo `HEAD`. A new commit is appended each
-time a checkpoint is registered by the code. These are not pushed to the Git
-remote by default (see `dvc exp push`).
+Instead of a single commit, checkpoint experiments have multiple commits under
+the custom Git reference (in `.git/refs/exps`), forming a branch.
```
pmrowla, line 97:
```suggestion
the custom Git reference (in `.git/refs/exps`), similar to a branch.
```

I think we should be explicit that anything in `refs/exps` is *not* a Git branch. For all intents and purposes, exp refs function the same way, but a formal "Git branch" must be a `refs/heads` reference.

The important distinction being that git CLI commands which expect branch names will not work with (shortened) exp ref names.

content/docs/command-reference/exp/run.md
jorgeorpinel: OK yeah we can omit that detail (done). But it's not clear that the exp commit doesn't become HEAD (i.e. doesn't get checked out) ⌛

jorgeorpinel: OK I clarified that too.

> if HEAD has moved between exp run --queue, and exp run --run-all, we still execute each queued experiment based on HEAD

- [ ] I tried testing this but it gets complicated... May require further QA/doc in the future. Is it a use case though? Queueing experiments from different HEADs and then running them all later. Thanks @pmrowla

content/docs/command-reference/exp/run.md
jorgeorpinel: > changes from --queue/--temp runs are not applied to the workspace
> should maybe be something like "to apply the (git) changes from an experiment to the workspace

Updated in 0c7a060d.

content/docs/command-reference/exp/run.md
pmrowla: It's intended to work that way, but I'm not sure how common of a use case it would be for users.

dberenbaum: I would expect my queued experiments to reflect the time they were queued. I'm not sure how common it would be to change HEAD before running experiments in the queue. It might be useful to queue experiments from different branches. It also might work as an alternative to making multiple copies of a repo in cases where someone wants to run different stages simultaneously (`dvc exp run --queue stage1 && dvc exp run --queue stage2 && dvc exp run --run-all -j 2`). I'd be surprised if people use it this way, though.

jorgeorpinel:
Sorry for the delay guys. Please someone approve, so we can merge this. Thanks!

shcheklein (approved):

PR #2333, opened 2021-03-28
dvclive: docs followup

jorgeorpinel:
Continues #2227


content/docs/command-reference/run.md
```diff
@@ -268,6 +268,17 @@ $ dvc run -n second_stage './another_script.sh $MYENVVAR'
 - `--desc <text>` - user description of the stage (optional). This doesn't  
   affect any DVC operations.
 
+- `--live <path>` - specify the directory `path` for
+  [Dvclive](/doc/dvclive/dvclive-with-dvc) to write logs in. Saved in the `live`
+  field of `dvc.yaml`.
+
+- `--live-no-cache <path>` - specify a Dvclive log file <abbr>output</abbr> (not
+  <abbr>cached</abbr>). Turns any `-o <path>` into a `-O`.
```
jorgeorpinel, line 276: Is that what this does? The cli help output isn't very clear

```
$ dvc stage add -h
...
  --live-no-cache <path>
                        Declare output as dvclive (do not put into DVC cache).
```

content/docs/command-reference/stage/add.md
```diff
@@ -246,6 +246,17 @@ data science experiments.
 - `--desc <text>` - user description of the stage (optional). This doesn't  
   affect any DVC operations.
 
+- `--live <path>` - specify the directory `path` for
+  [Dvclive](/doc/dvclive/dvclive-with-dvc) to write logs in. Saved in the `live`
+  field of `dvc.yaml`.
+
+- `--live-no-cache <path>` - specify a Dvclive log file <abbr>output</abbr> (not
+  <abbr>cached</abbr>). Turns any `-o <path>` into a `-O`.
```
jorgeorpinel, line 254: Is that what this does? The cli help output isn't very clear

```
$ dvc stage add -h
...
  --live-no-cache <path>
                        Declare output as dvclive (do not put into DVC cache).
```

content/docs/command-reference/run.md
```diff
@@ -268,6 +268,17 @@ $ dvc run -n second_stage './another_script.sh $MYENVVAR'
 - `--desc <text>` - user description of the stage (optional). This doesn't  
   affect any DVC operations.
 
+- `--live <path>` - specify the directory `path` for
+  [Dvclive](/doc/dvclive/dvclive-with-dvc) to write logs in. Saved in the `live`
+  field of `dvc.yaml`.
```
jorgeorpinel, line 273: - Unrelated Q: Is it worth maintaining `dvc run` going fwd or should we just redirect to `stage add`? I'm not 100% sure they're exactly the same at this point?

content/docs/dvclive/dvclive-with-dvc.md
```diff
@@ -3,17 +3,29 @@
...
+class MetricsCallback(Callback):
+    def on_epoch_end(self, epoch: int, logs: dict = None):
+        logs = logs or {}
+        for metric, value in logs.items():
+            dvclive.log(metric, value)
+        dvclive.next_step()
+
+
 def load_data():
     (x_train, y_train), (x_test, y_test) = mnist.load_data()
```
jorgeorpinel, line 30: TBH I'd even considering replacing both functions and some imports with `...`, or at least their contents (as if collapsed in an IDE) — in the latter case maybe also the class.

content/docs/command-reference/stage/add.md
pared: "Specify a Dvclive log directory" rather than "file". Its the same as `live` but not cached.
I think we might also need to point that in case of `--live` - the `summary` JSON file still has to be versioned by git.

content/docs/dvclive/dvclive-with-dvc.md
pared: I have been including all code, because personally I prefer tutorials that way, so my view on that is skewed. If we collapse imports and methods in other docs, sure.

pared (approved):

content/docs/command-reference/stage/add.md
jorgeorpinel: Updated.

> we might also need to point that in case of --live - the summary JSON file still has to be versioned by git

Yes but not in this ref. as there's 0 context about Dvclive. The explanation only makes sense in the DVClive docs, I think. I'll check if it's mentioned there... ⌛

jorgeorpinel: Updated some more in f326fcb3.

content/docs/dvclive/dvclive-with-dvc.md
jorgeorpinel: OK but it's not really a tutorial, I think. I.e. the specific It's a guide, and it's repeated from https://dvc.org/doc/dvclive/usage which presumably most people will read first.

BTW I'd argue to collapse or at least hide some details in both guides 😬 — not in this PR. And yes we def. shorten code blocks in most docs except in Get Started mainly, which is a step-by-step walkthrough we expect users to actually follow closely.

PR #2336, opened 2021-03-29
Regular updates

jorgeorpinel:
- Misc. copy edits
- Follow up to #2273 (install instructions)


content/docs/command-reference/metrics/index.md
```diff
@@ -102,12 +102,8 @@ to compare and pick the best performing experiment.
 
 ## Examples
 
-> This example is based on the `evaluate` stage of our
-> [Get Started](/doc/start/metrics-parameters-plots), where you can find the
-> actual source code.
-
 First, let's imagine we have a simple [stage](/doc/command-reference/run) that
-produces a `eval.json` metrics file:
+produces an `eval.json` metrics file:
```
jorgeorpinel, line 106: /doc/start/metrics-parameters-plots no longer features an eval.json metrics file.

content/docs/install/linux.md
```diff
@@ -51,13 +51,17 @@ Depending on the type of the [remote storage](/doc/command-reference/remote) you
 plan to use, you might need to install optional dependencies: `dvc-s3`,
 `dvc-azure`, `dvc-gdrive`, `dvc-gs`, `dvc-oss`, `dvc-ssh`.
```
jorgeorpinel, line 52: Hi @efiop. So on this (`conda-forge` paks) what's the default? I'm guessing `dvc` includes everything in this case? Unlike `pip`.

shcheklein (approved):

content/docs/install/linux.md
efiop: It used to before 2.0, but now it is same as `pip` - `dvc` package doesn't include anything except for built-ins like http (same as pip package)

PR #2346, opened 2021-04-01
March comm gems

jendefig:
This makes revisions to the question regarding keeping experiments in different repositories as well as moves the April Meetup info to the end.


shcheklein:
Hmm ... something funny is happening with this PR? it has some old changes and is conflicting with master ...

jendefig:
> Hmm ... something funny is happening with this PR? it has some old changes and is conflicting with master ...

Ok.  What do I need to do.  The changes look how I did them.  Is something else wrong?

shcheklein:
I think you need to create a new PR, from a new branch. It seems that you were trying to keep iterating in the existing branch? But it is merged already, you now need to synchronize your master, do a new branch, do changes there, and create a new PR. That my take from from what I see, may be I'm missing something.

jendefig:
Closing.  Re-did in new-march-comm-gems branch

PR #2364, opened 2021-04-08
ref: rewrite scope options for sync/gc/exp commands (--all-commits, etc.)

jorgeorpinel:
rel https://github.com/iterative/dvc.org/pull/2325#pullrequestreview-620832943

### DIFF
--- a/content/docs/command-reference/fetch.md
+++ b/content/docs/command-reference/fetch.md
@@ -90,19 +90,18 @@ specific one is given with `--remote`.
   config option with `dvc remote modify`. Using more jobs may speed up the
   operation.
 
-- `-a`, `--all-branches` - fetch cache for all Git branches instead of just the
-  current workspace. This means DVC may download files needed to reproduce
-  different versions of a `.dvc` file, not just the ones currently in the
-  workspace. Note that this can be combined with `-T` below, for example using
-  the `-aT` flag.
+- `-a`, `--all-branches` - fetch cache for all Git branches, as well as for the
+  workspace. This means DVC may download files needed to reproduce different
+  versions of a `.dvc` file, not just the ones currently in the workspace. Note
+  that this can be combined with `-T` below, for example using the `-aT` flags.
 
-- `-T`, `--all-tags` - same as `-a` above, but applies to Git tags as well as
-  the workspace. Note that both options can be combined, for example using the
-  `-aT` flag.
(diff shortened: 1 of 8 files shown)


content/docs/command-reference/fetch.md
```diff
@@ -100,9 +100,9 @@ specific one is given with `--remote`.
   the workspace. Note that both options can be combined, for example using the
   `-aT` flag.
 
-- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ Git commits
-  as well as the workspace. This downloads tracked data for the entire commit
-  history of the project.
+- `--all-commits` - same as `-a` or `-T` above, but applies to _all_ your Git
```
shcheklein, line 103: hmm ... it can be interpreted like only my commits (authored by me)?

jorgeorpinel: Ah good point. Idk how to best express this... 🤔

jorgeorpinel: I guess it's not easy to do so and may not be needed since the fact that experiments are Git commits is a hidden implementation detail. Will re-use this PR to roll back this change in exp-related docs instead ⌛

jorgeorpinel: Or how about adding a note like this @shcheklein ?

image

shcheklein: my initial reaction - we are overthinking this a bit. How about "applies to all commits"?

jorgeorpinel: It doesn't apply to experiments (which are commits). May not be much of a confusion issue here in `fetch` but it's relevant in `gc` and `exp show/gc`. All other changes I did for consistency at this point.

shcheklein: > It doesn't apply to experiments (which are commits).

I guess, I doubt anyone reading this will have a thought about experiments commits (that are hidden and far from obvious that they are commits)

It's not ideal, but trying to be too precise can even hurt since we are complicating this (force people to go and read dvc experiments and what does it at all, and if it's important for them, etc)

jorgeorpinel: OK. I removed the note everywhere except in `dvc gc` where there's actual need and some context for it. Did a bunch of other copy edits in the process 😅 PTAL

shcheklein (approved):

PR #2365, opened 2021-04-08
ref: clarify that exp run isn't exactly like repro

jorgeorpinel:
per https://github.com/iterative/dvc/issues/5593#issuecomment-815444772

AND remove some old file we don't need.

### DIFF
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -27,8 +27,9 @@ directories, etc.
 
 > `dvc exp run` is equivalent to `dvc repro` for <abbr>experiments</abbr>. It
 > has the same behavior when it comes to `targets` and stage execution (restores
-> the dependency graph, etc.). See the command [options](#options) for more on
-> the differences.
+> the dependency graph, etc.). An important difference is that changes in the
+> working tree are discarded by `dvc exp run`. See the command
+> [options](#options) for more on the differences.
 
 Before running an experiment, you'll probably want to make modifications such as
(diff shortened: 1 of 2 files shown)


content/docs/command-reference/exp/run.md
```diff
@@ -27,8 +27,9 @@ directories, etc.
 
 > `dvc exp run` is equivalent to `dvc repro` for <abbr>experiments</abbr>. It
 > has the same behavior when it comes to `targets` and stage execution (restores
-> the dependency graph, etc.). See the command [options](#options) for more on
-> the differences.
+> the dependency graph, etc.). An important difference is that changes in the
+> working tree are discarded by `dvc exp run`. See the command
+> [options](#options) for more on the differences.
```
pmrowla, line 32: Changes in the working tree are not supposed to be removed though.

The discussion in #5593 is specifically because changes to DVC data dependencies which have not been `dvc commit`ed to `.dvc` or `dvc.lock` are currently lost. All other changes to the user's working tree are preserved.

dberenbaum: Changes to Git-tracked dependencies (code, parameters) in the working tree are not discarded, so this seems misleading. I think it makes more sense to revisit this after #5593 has been resolved.

jorgeorpinel: OK thanks guys, I reverted these changes, let's revisit along with the core PR that addresses https://github.com/iterative/dvc/issues/5593 then.

jorgeorpinel:
Haha so I guess this PR is empty now.

PR #2368, opened 2021-04-10
ref: clarifications to `exp run`

jorgeorpinel:
- What "based on the latest project version" means.
- `--temp` ~~isn't useful yet~~ is useful **background**-ing long experiments (https://github.com/iterative/dvc/issues/5593#issuecomment-817237581)
- Other clarifications


content/docs/command-reference/exp/run.md
```diff
@@ -52,8 +52,8 @@ out by DVC). Note that these commits are not pushed to the Git remote by default
 
 </details>
 
-The results of the last experiment can be seen in the <abbr>workspace</abbr>. To
-display and compare your experiments, use `dvc exp show` or `dvc exp diff`. Use
+The results of the last experiment can be seen in the workspace. To display and
+compare your experiments, use `dvc exp show` or `dvc exp diff`. Use
```
pmrowla, line 56: Maybe we should document that all of the other DVC diff commands and plots commands can also be used with experiments here

pmrowla (approved):

content/docs/command-reference/exp/run.md
```diff
@@ -52,8 +52,8 @@ out by DVC). Note that these commits are not pushed to the Git remote by default
 
 </details>
 
-The results of the last experiment can be seen in the <abbr>workspace</abbr>. To
-display and compare your experiments, use `dvc exp show` or `dvc exp diff`. Use
+The results of the last experiment can be seen in the workspace. To display and
```
dberenbaum, line 55: Results can only be seen in the workspace if it is not a temp run, right?

dberenbaum: I guess this is clarified below.

content/docs/command-reference/exp/run.md
jorgeorpinel: I'd wait to see if all the `show` and `diff` commands get merged 🙂

pmrowla: Unification of the `show` and `diff` commands is something that probably won't happen for a decent amount of time, and I brought it up because I have seen questions from users in discord asking about how to use plots with experiments.

jorgeorpinel: Oh. I thought you meant `metrics/params diff`, which are mentioned in https://dvc.org/doc/command-reference/exp/diff. I see you also mentioned `plots` now, and I do agree that should be mentioned but maybe also from the `exp diff` ref. for now? ⌛

content/docs/command-reference/exp/run.md
jorgeorpinel: Actually this Q sent me into a QA spiral haha. Will have some results and an answer soon ⌛

jorgeorpinel: > Results can only be seen in the workspace if it is not a temp run

OK I tried this and yes, that is correct (wasn't sure for queued runs). But the first part of the Description intends to on the simplest workflow first, before getting to the other sections like **Checkpoints** and **Queueing**.

> this is clarified below

Yes it is. I changed it a bit anyway (cc6fbaff).

content/docs/command-reference/exp/run.md
jorgeorpinel: > DVC diff commands and plots commands can also be used with experiments here

Wait. But we mean more like show and compare MANY/ALL experiments here. `metrics/params diff` only take 2 and in those cases `exp diff` is more obvious anyway.

This  only really applies to `plots diff` as of now, I think? Mentioned just that in 29004e6b for now.

content/docs/command-reference/exp/run.md
```diff
@@ -100,7 +101,7 @@ the custom Git reference (in `.git/refs/exps`), similar to a branch.
 
 </details>
 
-## Queueing and parallel execution
+## Queueing and background execution
```
jorgeorpinel, line 104: WDYT of this conceptual change?

content/docs/command-reference/exp/run.md
```diff
@@ -137,19 +142,20 @@ CPU cores).
   name will generated by default, such as `exp-f80g4` (based on the experiment's
   hash).
 
+- `--temp` - run this experiment in the background, and outside your workspace
+  (in `.dvc/tmp/exps`). Useful to continue working while a long experiment runs.
```
jorgeorpinel, line 146: My only concern is that "backgound" sounds a bit like `exp run &` which isn't exactly it. with `--temp/queue` you probably need a separate terminal for your "foreground" work. Should probably rephrase... Ideas?

pmrowla: I think it would be better to just say "outside your workspace" instead of "in the background" like it is documented in `--run-all`.

Honestly I think we can just document `--temp` as a shortcut/equivalent to `--queue` + `--run-all` for single experiment runs.

content/docs/command-reference/exp/run.md
pmrowla: I think this should be left alone for the same reasons you noted below. We don't actually run anything in the background, the user can do that themselves if they want using `&` in bash, or by just running multiple terminal instances, but DVC does not behave like any kind of background process/daemon in typical use cases.

content/docs/command-reference/exp/run.md
dberenbaum: > I think it would be better to just say "outside your workspace"

👍 

> I think we can just document `--temp` as a shortcut/equivalent to `--queue` + `--run-all`

Not sure it makes sense to have to read two other options to understand `--temp`, and it's unclear whether `--temp` should refer back to `--queue`/`--run-all` or vice versa.

Simply removing "in the background" seems good enough to me:

```suggestion
- `--temp` - run this experiment outside your workspace
  (in `.dvc/tmp/exps`). Useful to continue working while a long experiment runs.
```

content/docs/command-reference/exp/run.md
```diff
@@ -109,18 +110,22 @@ execution. `dvc exp show` will mark queued experiments with an asterisk `*`.
...
+workspace, like for regular experiments.
 
 ⚠️ Parallel runs are experimental and may be unstable at this time. ⚠️ Make sure
 you're using a number of jobs that your environment can handle (no more than the
 CPU cores).
 
 > Note that each job runs the entire pipeline (or `targets`) serially — DVC
-> makes no attempt to distribute stage commands among jobs.
+> makes no attempt to distribute stage commands among jobs. The order in which
+> they were queues is also not guaranteed.
```
dberenbaum, line 125:
```suggestion
> they were queued is also not guaranteed.
```

content/docs/command-reference/exp/run.md
```diff
@@ -109,18 +110,22 @@ execution. `dvc exp show` will mark queued experiments with an asterisk `*`.
...
+
+To run a single experiment in the background, use `dvc exp run --temp`. This way
+if you can continue to work in your workspace while a long experiment runs.
```
dberenbaum, line 128:
```suggestion
you can continue to work in your workspace while a long experiment runs.
```

dberenbaum (commented):
Agree that "in the background" doesn't seem quite right since temp experiments are blocking once they are run. Otherwise looks good, and I can take one more look once that issue is resolved.

content/docs/command-reference/exp/run.md
jorgeorpinel: Rolled back to "parallel" for now.

content/docs/command-reference/exp/run.md
jorgeorpinel: > better to just say "outside your workspace"... Simply removing "in the background"

Done (7a1e5518).

> just document --temp as a shortcut/equivalent to --queue + --run-all for single experiment runs

Agree with @dberenbaum that it's best to doc `--temp` on its own as it has it's own uses. The fact that it relates closely with `--queue` is more of an implementation detail.

content/docs/command-reference/exp/run.md
```diff
@@ -109,18 +110,21 @@ execution. `dvc exp show` will mark queued experiments with an asterisk `*`.
 > Note that queuing an experiment that uses checkpoints implies `--reset`,
 > unless a `--rev` is provided (refer to the previous section).
 
-Use `dvc exp run --run-all` to process the queue. Adding `-j` (`--jobs`),
-experiment queues can be run in parallel for better performance. This creates a
-temporary workspace copy for each subprocess (in `.dvc/tmp/exps`). See also the
-`--temp` option. Queued (and temporary) experiment runs do not get applied to
-the workspace like regular ones.
+Use `dvc exp run --run-all` to process the queue. This is done outside your
+<abbr>workspace</abbr> (in a temporary dir in `.dvc/tmp/exps`), so you can
```
jorgeorpinel, line 114: Are exp queues always processed outside the workspace, even with `-j 1`? And why

content/docs/command-reference/exp/run.md
```diff
@@ -137,19 +141,21 @@ CPU cores).
...
   provided.
 
-- `--run-all` - run all queued experiments (see `--queue`). Use `-j` to execute
-  them [in parallel](#queueing-and-parallel-execution).
+- `--run-all` - run all queued experiments (see `--queue`) and outside your
+  workspace (in `.dvc/tmp/exps`). Use `-j` to execute them
+  [in parallel](#queueing-and-parallel-execution).
 
 - `-j <number>`, `--jobs <number>` - run this `number` of queued experiments in
-  parallel. Only applicable when used in conjunction with `--run-all`.
```
jorgeorpinel, line 149: What's the `-j` default here, BTW? Also 4 x CPU cores like in other commands?

jorgeorpinel (commented):
Couple more Qs @pmrowla 👇 Thanks!

content/docs/command-reference/exp/run.md
jorgeorpinel: > The fact that it relates closely with --queue is more of an implementation detail.

That said, it is explained in the Queued runs section, so the connection is reflected there.

content/docs/command-reference/exp/run.md
dberenbaum: I think that approach makes sense @jorgeorpinel. Alternatively, we could say something like "any dvc command that accepts Git revisions also accepts experiment names."

content/docs/command-reference/exp/run.md
dberenbaum: To me, "queueing" means setting up an experiment for future execution. To do that, I would expect to add to the queue, then keep making changes to my workspace without impacting the experiment that I already queued. It seems like this workflow would always require processing outside the workspace, right?

content/docs/command-reference/exp/run.md
dberenbaum: Default is 1. See https://github.com/iterative/dvc/pull/5785.

dberenbaum (approved):

content/docs/command-reference/exp/run.md
```diff
@@ -109,18 +110,21 @@ execution. `dvc exp show` will mark queued experiments with an asterisk `*`.
...
-temporary workspace copy for each subprocess (in `.dvc/tmp/exps`). See also the
-`--temp` option. Queued (and temporary) experiment runs do not get applied to
-the workspace like regular ones.
+Use `dvc exp run --run-all` to process the queue. This is done outside your
+<abbr>workspace</abbr> (in a temporary dir in `.dvc/tmp/exps`), so you can
+continue working on the project meanwhile (e.g. on another terminal). You can
+run a single experiment outside the workspace with `dvc exp run --temp`.
+
+Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
+performance (creates a tmp dir for each subprocess).
```
dberenbaum, line 119:
```suggestion
performance (creates a tmp dir for each experiment).
```

content/docs/command-reference/exp/run.md
jorgeorpinel: > Default is 1.

Thanks! Noted in eab608f7.

content/docs/command-reference/exp/run.md
jorgeorpinel: > creates a tmp dir for each experiment

This is always true, even without `--jobs`. Changed to "jobs" instead (c0cbf1b5).

content/docs/command-reference/exp/run.md
jorgeorpinel: > this workflow would always require processing outside the workspace

At `exp run --run-all` it could checkout the workspace locally at each queue time, overwriting whatever you have there and forcing you to wait for it to finish.

dberenbaum: > overwriting whatever you have there

It could do that, but overwriting the workspace isn't ideal, right? I know the current solution isn't perfect either, but avoiding having to overwrite the workspace seems like a big reason why exp queues are always processed outside the workspace (in addition to consistency with parallel execution and possibly remote scenarios in the future).

jorgeorpinel: Ah yeah, I agree! I'm just saying it's not necessarily obvious, so we should be explicit about the behavior I think. I guess you were just answering the Q (with a yes). Updated a bit in d8acecb4 anyway. LGTM!

PR #2369, opened 2021-04-10
ref: update `stage list` refs.

jorgeorpinel:
> Continues https://github.com/iterative/dvc.org/pull/2245#issuecomment-817176199

- Needed update to the `stages` ref index
- Corrections to the **Synopsis**
- Clarifies the **Description**
- Remove term "group" in the context of foreach stages (avoiding that term was decided early on, see https://github.com/iterative/dvc/issues/5165#issuecomment-751642817 and https://github.com/iterative/dvc/issues/5165#issuecomment-752265954)
- Copy edits to the **Example**

Please review #2425 along with or after this.


content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,57 @@ usage: dvc stage list [-h] [-q | -v]
...
-    in `./dvc.yaml`
-  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in a specific `dvc.yaml` file
+    stages
+  - `dvc stage list train-model`: Specific stage in `./dvc.yaml`. Supports
+    [`foreach` stage](doc/user-guide/project-structure/pipelines-files#foreach-stages)
+    names.
+  - `dvc stage list modeling/dvc.yaml:prepare`: Stage in a specific `dvc.yaml`
+    file. Supports `foreach` stage names.
```
jorgeorpinel, line 50: Pending https://github.com/iterative/dvc.org/pull/2245#pullrequestreview-632942351 we could keep the "group" term, or reinstate it in a broader update (including to https://dvc.org/doc/user-guide/project-structure/pipelines-files#foreach-stages).

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,57 @@ usage: dvc stage list [-h] [-q | -v]
                       [targets ...]
 
 positional arguments:
-  targets          Show stages from a dvc.yaml file or a directory (with -R).
-                   'dvc.yaml' by default.
+  targets       Limit command scope to these dvc.yaml files,
+                directories (with -R), or stage names.
+                './dvc.yaml' by default.
```
jorgeorpinel, line 15: Not 100% sure about this, using the "Limit command scope..." for consistency but could be different too. Or I can make a separate PR after this to get rid of that phrase everywhere.

- [ ] Will need core update

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,55 @@ usage: dvc stage list [-h] [-q | -v]
...
 
 ## Options
 
-- `targets` (optional command argument) - stages to list from (`./dvc.yaml` by
-  default). Different things can be provided as targets depending on the flags
-  used (more details in each option). Examples:
+- `targets` (optional command argument) - where to look for stages to list
+  (`./dvc.yaml` by default). Examples:
 
   - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to list from
```
dberenbaum, line 43: Might not need "to list from"? Feel free to ignore.

```suggestion
  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file
```

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,55 @@ usage: dvc stage list [-h] [-q | -v]
...
   - `dvc stage list -R pipelines/`: Directory path to explore recursively for
-    `dvc.yaml` files
-  - `dvc stage list train-model`: Specific stage name or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in `./dvc.yaml`
-  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in a specific `dvc.yaml` file
+    stages
```
dberenbaum, line 45: Why this change? `dvc.yaml` files seemed a more accurate description of `where to look for stages to list`.

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,55 @@ usage: dvc stage list [-h] [-q | -v]
...
 ## Options
 
-- `targets` (optional command argument) - stages to list from (`./dvc.yaml` by
-  default). Different things can be provided as targets depending on the flags
-  used (more details in each option). Examples:
+- `targets` (optional command argument) - where to look for stages to list
+  (`./dvc.yaml` by default). Examples:
 
   - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to list from
   - `dvc stage list -R pipelines/`: Directory path to explore recursively for
```
dberenbaum, line 44: I can see changing the directory from the previous example to anticipate the recursive use case, but I think it's a lot easier to digest if we reuse the examples, and we mention it searches recursively. 

I'd also take out `path` since it's not used in the line above, but that's very minor.

```suggestion
  - `dvc stage list -R linear/`: Directory to explore recursively for
```

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,55 @@ usage: dvc stage list [-h] [-q | -v]
...
-    `dvc.yaml` files
-  - `dvc stage list train-model`: Specific stage name or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in `./dvc.yaml`
-  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in a specific `dvc.yaml` file
+    stages
+  - `dvc stage list train-model`: Specific stage(s) from `./dvc.yaml`
+  - `dvc stage list modeling/dvc.yaml:prepare`: Stage(s) from a specific
```
dberenbaum, line 47: Again, I think it's easier to digest if it's consistent. I'm okay with having different stage names since there are different `dvc.yaml` paths, but IMO using the same stage name is probably better, too, since it's otherwise a bit hard to tell that the part after the colon is a stage name. If you prefer different stage names, I'm fine with it.

```suggestion
  - `dvc stage list linear/dvc.yaml:train-model`: Stage(s) from a specific
```

content/docs/command-reference/stage/list.md
```diff
@@ -67,9 +67,9 @@ DVC also uses this command to provide autocompletion in `bash` and `zsh`.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
-## Example: List stages
+## Examples:
 
-Listing stages from the default `./dvc.yaml` file from the repo.
+The base case is to list stages from `./dvc.yaml`:
```
dberenbaum, line 72: Seems like there should be integration between this section and the `targets` option. At least, they should be part of a common example repo structure with the same `dvc.yaml` files, directories, and stage names. Maybe one should also reference the other so it's clear where to see more in-depth examples?

dberenbaum (commented):
Looks good! A few minor comments on making examples consistent.

content/docs/command-reference/stage/list.md
dberenbaum: Would this be a more appropriate place to include the `foreach` content (https://github.com/iterative/dvc.org/pull/2245#discussion_r614368197)?

content/docs/command-reference/stage/list.md
jorgeorpinel: Removing "path". But I think usually some variation in examples can help people understand things implicitly.

content/docs/command-reference/stage/list.md
jorgeorpinel: Hmmm yeah maybe. I'm not convinced though. The point here is that multiple directories will be searched, and we're searching for stages. The fact that there can only be one dvc.yaml file per dir is a coincidence.

Also for consistency with the `repro` ref.

content/docs/command-reference/stage/list.md
jorgeorpinel: > okay with having different stage names
but IMO using the same stage name is probably better

Did you mean different dirs but same stage name? Or vice versa 🙂 

In any case I like the variation and it should be clear that the part before `:` is a path to a dvc.yaml file.

You may be right but this seems subjective so I'd err on the side of keeping it as-is (I didn't change that in this PR). Or we can ask someone to cast a deciding vote.

content/docs/command-reference/stage/list.md
jorgeorpinel: > integration between this section and the targets option

OK context: Actually `targets` is not an option (it's the one required argument) and I was kind of opposed originally to put in under **Options**, but @skshetry and @shcheklein beat my vote. They were right in that it's gotten so complex that it needs to be explained in detail but we also didn't want to introduce an **Arguments** section just for that.

Your idea of an example (hypothetical I assume) repo structure to demo al the kinds of targets is great but that would be even longer that the current bullet list so it'd have to be a new/improved Example which I don't think could replace the `targets` bullet list ref.

jorgeorpinel: > Would this be a more appropriate place to include the foreach content?

This PR is a follow up to that one actually. I removed the `foreach` terminology here as you suggested. Maybe I'm not getting the Q? 😋

jorgeorpinel: > integration between this section and the targets option

Another idea is to move the main argument ino the Synopsis section but that would make the intro a bit long. Or make a **Targets** section in the description but it would have to be as compressed as the current bullet list as the Desc is already long.

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,54 @@ usage: dvc stage list [-h] [-q | -v]
...
-DVC also uses this command to provide autocompletion in `bash` and `zsh`.
+> (ℹ️) DVC uses this command to provide
+> [shell tab completion](/doc/install/completion).
 
 ## Options
 
-- `targets` (optional command argument) - stages to list from (`./dvc.yaml` by
-  default). Different things can be provided as targets depending on the flags
-  used (more details in each option). Examples:
+- `targets` (optional command argument) - where to look for stages to list
```
shcheklein, line 40: `where to look` is not correct here I believe. Btw, what is the motivation to revisit this? :) Would be great to outline it in the PR description. Copy edits? Fixes? etc

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,54 @@ usage: dvc stage list [-h] [-q | -v]
...
+  - `dvc stage list modeling/dvc.yaml:prepare`: Stage(s) from a specific
+    `dvc.yaml` file
 
-  - `dvc stage list linear/dvc.yaml`: Specific `dvc.yaml` file to list from
-  - `dvc stage list -R pipelines/`: Directory path to explore recursively for
-    `dvc.yaml` files
-  - `dvc stage list train-model`: Specific stage name or a
-    [foreach group](doc/user-guide/project-structure/pipelines-files#foreach-stages)
-    in `./dvc.yaml`
-  - `dvc stage list modeling/dvc.yaml:prepare`: Stage or a
```
shcheklein, line 45: what is the reason removing this?

content/docs/command-reference/stage/list.md
```diff
@@ -67,9 +66,9 @@ DVC also uses this command to provide autocompletion in `bash` and `zsh`.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
-## Example: List stages
+## Examples:
 
-Listing stages from the default `./dvc.yaml` file from the repo.
```
shcheklein, line 72: `Default` is a better term to my mind. Again, what is the reason/motivation for this PR. may be I'm missing the context

content/docs/command-reference/stage/list.md
shcheklein: > OK context: Actually targets is not an option (it's the one required argument) and I was kind of opposed originally to put in under Options, but @skshetry and @shcheklein beat my vote. 

yep, we didn't want to overcomplicate this and fixing two things at a time (potentially needed docs refactoring to accommodate options and arguments) and actually documenting this command. I think it's not a major priority and it's better to first discuss it in the ticket, then apply to all occurrences. WDYT?

content/docs/command-reference/stage/list.md
shcheklein: Looks fine to me

content/docs/command-reference/repro.md
```diff
@@ -105,11 +105,11 @@ up-to-date and only execute the final stage.
   `./dvc.yaml` by default). Different things can be provided as targets
   depending on the flags used (more details in each option). Examples:
 
-  - `dvc repro linear/dvc.yaml`: Specific `dvc.yaml` file to reproduce
-  - `dvc repro -R pipelines/`: Directory path to explore recursively for
-    `dvc.yaml` files
-  - `dvc repro train-model`: Specific stage in `./dvc.yaml`
-  - `dvc repro modeling/dvc.yaml:prepare`: Stage in a specific `dvc.yaml` file
+  - `dvc repro linear/dvc.yaml`: Specific `dvc.yaml` file
```
shcheklein, line 108: let's not mix pls multiple things in one PR

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,54 @@ usage: dvc stage list [-h] [-q | -v]
...
+                './dvc.yaml' by default.
 ```
 
+> See [`targets`](#options) for more details.
+
 ## Description
 
-`dvc stage list` is useful for quick listing of stages present in the project
-without having to open `dvc.yaml` files.
+Prints a summarized list of stages including their names and a one-line
```
shcheklein, line 22: `summarized` - do we need this word?

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,54 @@ usage: dvc stage list [-h] [-q | -v]
...
+the current working directory) by default.
 
-Without any targets, by default, it only lists the stages from the `dvc.yaml`
-present in the current working directory.
+The stage description is read from the `desc` field in `dvc.yaml` (truncated to
+80 characters) if it exists. Otherwise, DVC generates one with the stage's most
+important characteristics (dependencies, outputs, or metrics).
 
-DVC also uses this command to provide autocompletion in `bash` and `zsh`.
+> (ℹ️) DVC uses this command to provide
```
shcheklein, line 35: not sure this Nore is important

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,54 @@ usage: dvc stage list [-h] [-q | -v]
...
 
 positional arguments:
-  targets          Show stages from a dvc.yaml file or a directory (with -R).
-                   'dvc.yaml' by default.
+  targets       Limit command scope to these dvc.yaml files,
+                directories (with -R), or stage names.
+                './dvc.yaml' by default.
 ```
 
+> See [`targets`](#options) for more details.
```
shcheklein, line 18: we don't use anywhere else? let's not complicate it with Notes?

content/docs/command-reference/stage/list.md
dberenbaum: Okay, I don't feel that strongly here.

> dvc.lock files are also involved in the search

Is that true? Can a `dvc.lock` file be a target?

jorgeorpinel: > Can a dvc.lock file be a target?

No. I meant that they have to be checked along with dvc.yaml even if your target is a stage name.

content/docs/command-reference/stage/list.md
jorgeorpinel: I agree, not in the scope of this PR. I'm sure it will come up again if needed.

content/docs/command-reference/stage/list.md
jorgeorpinel: > where to look is not correct here

Changed to "location of the stages to list"

jorgeorpinel: > Btw, what is the motivation to revisit this? :) Would be great to outline it in the PR description. Copy edits? Fixes? etc

I have updated the PR description @shcheklein .

content/docs/command-reference/stage/list.md
jorgeorpinel: > reason removing this?

To avoid the term "foreach group" (explained in the updated PR desc.)

content/docs/command-reference/stage/list.md
jorgeorpinel: OK, but there can only be one `./dvc.yaml` file, so it doesn't make sense to put "default" before that. Changed to

> The default behavior is to list stages from `./dvc.yaml`

content/docs/command-reference/stage/list.md
jorgeorpinel: Removed "summarized".

content/docs/command-reference/stage/list.md
jorgeorpinel: Me neither. Removed note about shell completion.

jorgeorpinel:
> what is the reason/motivation for this PR

@shcheklein basically I didn't have a chance to review #2245 before merge so I did a follow-up PR instead. It includes corrections and copy edits (detailed in the PR desc.)

content/docs/command-reference/stage/list.md
jorgeorpinel: We use it in https://dvc.org/doc/command-reference/repro @shcheklein. I remember we decided to put it there so the user doesn't miss the `targets` section in **Options**. Added here for that same reason and for consistency with the `repro` ref.

content/docs/command-reference/repro.md
jorgeorpinel: To my mind it's basically the same list in `repro`, so I edited both consistently. But sure, I will extract this ⌛

jorgeorpinel: Extracted to #2425

content/docs/command-reference/stage/list.md
```diff
@@ -10,55 +10,50 @@ usage: dvc stage list [-h] [-q | -v]
...
 
-It provides a list of stages, with a stage name and the description of the
-stage. The description of the stage is read from
-[`desc` field in the `dvc.yaml` file](/doc/user-guide/project-structure/pipelines-files#stage-entries),
-or if it does not exist, DVC generates the short description from the other
-metadata it has about the stage.
+Prints a list of stages including their names and a one-line description (which
+can be omitted using `--names-only`). This command is useful for discovering or
+reviewing what stages are present in the project without having to examine
+`dvc.yaml` files manually. It can also be a shorter alternative to `dvc dag`.
```
shcheklein, line 25: shorter sounds like command itself is shorter, also, I doubt this reference is important, let's focus on the list itself?

content/docs/command-reference/stage/list.md
shcheklein: I think it was beneficial to have that reference, it might be fine to use `foreach group` here?

content/docs/command-reference/stage/list.md
shcheklein: kk

shcheklein (approved):
feel free to merge when you feel comments are addressed enough here

content/docs/command-reference/stage/list.md
jorgeorpinel: > I think it was beneficial to have that reference

If we just say "specific stage" (like now) and the official term we use for these "groups" is "foreach stage", then it can be inferred that they can be targets too, as they are a kind of "(specific) stage". It's not an obvious inference but I'm not sure why emphasize foreach stages here. Why `list` or `repro` them in groups, if they're not groups?

> might be fine to use foreach group here

Same Q @shcheklein. I thought they were not meant to be groups in any way (at least originally) i.e. that treating them as groups was a misuse of the feature.

---

To be clear, I've always been of the idea that they are in fact groups as currently implemented (see "wide stage" concept in https://github.com/iterative/dvc/issues/5440). But so far we have specifically avoided documenting that, so it seems contradictory to treat them as such here. I'd be happy to revisit that separately though!

content/docs/command-reference/stage/list.md
jorgeorpinel: Removed `dag` ref.

PR #2377, opened 2021-04-15
guide: clarify auth for `add --external`

jorgeorpinel:
per https://discuss.dvc.org/t/dvc-add-external-s3-mybucket-data-csv-is-failing-with-access-error-even-after-giving-correct-remote-cache-configurations/726/4?u=jorgeorpinel


content/docs/user-guide/managing-external-data.md
```diff
@@ -61,6 +61,11 @@ types:
    moves a local file to the external location. This produces an external output
    in `dvc.yaml`.
 
+> 🔒 Note that for locations that require authentication, it's necessary to
```
shcheklein, line 64: this is not exactly correct, auth can be done via env vars, or regular credentials by the underlying library (like boto3).

jorgeorpinel: True. Updated, PTAL

content/docs/user-guide/managing-external-data.md
```diff
@@ -61,6 +61,12 @@ types:
    moves a local file to the external location. This produces an external output
    in `dvc.yaml`.
 
+> 🔒 Note that for locations that require authentication there may be extra
+> steps required. One way is to configure the remote accordingly after step 1,
```
shcheklein, line 65: Not clear still. " ... there may be extra steps required. One way ..." feels like some explanation of the "extra step" is missing. 

Also, almost all locations require auth. I would say this about some specific auth types.

jorgeorpinel: > feels like some explanation of the "extra step" is missing.

That's what the last sentence, "See `dvc remote modify` for more alternatives," is for.

> almost all locations require auth. I would say this about some specific auth types.

Agree. But this PR is meant as a patch: previously it was only mentioned in SSH and now I'm generalizing the note to apply to every type for now. Since this whole `--external` thing is strongly discouraged and the guide basically needs rewriting too, is it worth adding more complication to it now?

shcheklein: > That's what the last sentence, "See dvc remote modify for more alternatives," is for.

should it go before the "one way" then? Otherwise it looks like it is broken

> 'm generalizing the note to apply to every type for now

I understand. It becomes not a patch though but rather an improvement. The phrase `locations that require authentication` still looks too general. It happens only to certain types of auth. Easy fix is "Depending on the remote storage and authentication type, there may be ... You can use special remote:// ... to provide authentication and other remote settings via DVC config. See ..."

jorgeorpinel: > should it go before the "one way" then?

Moved.

> "locations that require authentication" still looks too general. It happens only to certain types of auth.

That part is general, but "there **may** be extra steps" narrows it down ("may").
But I rephrased the example along the lines of your suggestion (clarified step 1 too).

shcheklein:
@jorgeorpinel it looks like deployment failed? does it run locally for you?

content/docs/user-guide/managing-external-data.md
```diff
@@ -52,15 +52,21 @@ in the same external/remote file.
...
+   `dvc remote`\*) in the same location as the external data, using
+   `dvc config`.
 2. Tracking existing data on the external location using `dvc add` (`--external`
    option needed). This produces a `.dvc` file with an external URL or path in
    its `outs` field.
 3. Creating a simple stage with `dvc run` (`--external` option needed) that
    moves a local file to the external location. This produces an external output
    in `dvc.yaml`.
 
+> \* Note that for certain `dvc remote` authentication methods, extra config
```
shcheklein, line 65: `dvc remote` authentication method - sounds like the `dvc remote` command itself needs some auth

jorgeorpinel: Fixed.

jorgeorpinel:
> does it run locally for you?

It does 🤔 

image

`build` seems to fail at 

```
       [4/4] Building fresh packages...
       error /tmp/build_0e1ac83a/node_modules/sharp: Command failed.
       Exit code: 1
       Command: (node install/libvips && node install/dll-copy && prebuild-install) || (node-gyp rebuild && node install/dll-copy)
```

Cc @rogermparent

shcheklein (approved):

jorgeorpinel:
Should I merge and see what happens when deploying `master`? Or wait for @rogermparent

shcheklein:
@jorgeorpinel I think master also fails (there is a Slack channel that you can follow). Master has been failing for a while I think. So, let's merge it.

jorgeorpinel:
OK, merged. This builds fine in my local machine BTW (Ubuntu 18 on WSL).

PR #2380, opened 2021-04-16
ref: update targets desc in stage list and repro

jorgeorpinel:
rel https://github.com/iterative/dvc.org/pull/2245#pullrequestreview-632942351


jorgeorpinel:
Oh wait I already have a PR related to these refs... Let me move this into #2369...

PR #2390, opened 2021-04-17
ref: updated Azure env vars in remote cmds

### LINKED ISSUE #2376: remote: 2 Qs about Azure params
> Continues https://github.com/iterative/dvc.org/pull/2226#issuecomment-806200723

1. What about `AZURE_STORAGE_CONNECTION_STRING` and `AZURE_STORAGE_CONTAINER_NAME`?
2. What happens first, Azure "default credential" or Microsoft account login?

jorgeorpinel:
Per #2376


content/docs/command-reference/remote/add.md
```diff
@@ -162,36 +162,27 @@ For more information about the variables DVC supports, please visit
 
 ```dvc
 $ dvc remote add -d myremote azure://mycontainer/path
-$ dvc remote modify --local myremote connection_string 'mystring'
+$ dvc remote modify --local myremote connection_string 'mysecret'
```
shcheklein, line 165: is it a recommended/default way still? should we skip the credentials setup here?

jorgeorpinel: Skip credentials because of [default credential](https://docs.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential) ? Good Q, I have the impression we still want to rec `connection_string` but not sure. WDYT @isidentical?

jorgeorpinel:
Oops looks like a recent PR broke some links (checks are failing here but it comes from `master`). Fixing in #2391.

jorgeorpinel:
Oops, accidental request for re-review @shcheklein. Let's wait for an answer for https://github.com/iterative/dvc.org/pull/2390#pullrequestreview-638192264. Thanks

content/docs/command-reference/remote/add.md
isidentical: > I have the impression we still want to rec connection_string but not sure. 

Connection string is definitely easier since it usually contains all the relevant info inside (account_name/account_key/sas_token etc)

shcheklein: > Connection string is definitely easier since it usually contains all the relevant info inside (account_name/account_key/sas_token etc)

I guess my question here was about the most common situation. E.g. with AWS the most natural way is to expect the credentials file and CLI to be set. And everything else is needed only CI for example (to make secrets self-contained) or other automation scenarios.

isidentical: I assume `connection_string` is the most common method, considering that the dvc 2<= recommended it as the first class method for everyone (since it didn't support things like sas_token as a separate option). Outside of dvc, it seems like account_name/account_key combination is also recommended alongside with connection_string method.

jorgeorpinel: I would guess in favor of `connection_string` when considering remote machines (e.g. CI/CD) or other people's envs where not AWS CLI is setup.

jorgeorpinel:
p.s. @isidentical WDYT about the whole PR, please approve if you 👍 it.

content/docs/command-reference/remote/add.md
```diff
@@ -162,36 +162,27 @@ For more information about the variables DVC supports, please visit
...
+> `dvc remote modify` for a full list of Azure parameters.
 
 The Azure Blob Storage remote can also be configured globally via environment
 variables:
 
 ```dvc
 $ export AZURE_STORAGE_CONNECTION_STRING='mysecret'
-$ export AZURE_STORAGE_CONTAINER_NAME='mycontainer'
-$ dvc remote add -d myremote 'azure://'
+$ dvc remote add -d myremote azure://mycontainer/path
```
isidentical, line 179: I'm not sure whether we'd want to document this or keep as an internal behavior but just to note: unlike s3/google storage buckets, when the containers doesn't exist we automatically create them.

isidentical (approved):

content/docs/command-reference/remote/add.md
jorgeorpinel: I added a note in `remote modify`.

jorgeorpinel:
@shcheklein should we merge? There's another update I'd like to make to this topic but don't want to overcomplicate this PR 🙂

shcheklein:
@jorgeorpinel sounds good. Let's create a ticket though to address the `dvc remote add` ... I'm not sure we communicate properly defaults, available options, etc

PR #2392, opened 2021-04-17
Add warning about stale git on Windows

zimka:
Problem details can be found here: https://github.com/iterative/dvc/issues/5792

### DIFF
--- a/content/docs/user-guide/running-dvc-on-windows.md
+++ b/content/docs/user-guide/running-dvc-on-windows.md
@@ -93,3 +93,14 @@ $ choco install less
 `less` can be installed in other ways, just make sure it's available in the
 command line environment where you run `dvc`. (This usually means adding the
 directory where `less` is installed to the `PATH` environment variable.)
+
+## Using private repositories
+
+When a stale enough version of git is installed, you may face issues with 
+several dvc commands (`list`, `get`, `import`) applied to private 
+repositories, which look like infinite loading or "freeze". The issues are 
+caused by git internal auth problems and they are fixed for newer git 
+versions. Please upgrade git to 2.29 at least, where Git Credential Manager 
+Core is included. You can update git using `git update-for-windows` command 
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/running-dvc-on-windows.md
```diff
@@ -93,3 +93,14 @@ $ choco install less
...
+## Using private repositories
+
+When a stale enough version of git is installed, you may face issues with 
+several dvc commands (`list`, `get`, `import`) applied to private 
+repositories, which look like infinite loading or "freeze". The issues are 
+caused by git internal auth problems and they are fixed for newer git 
+versions. Please upgrade git to 2.29 at least, where Git Credential Manager 
+Core is included. You can update git using `git update-for-windows` command 
+either in GitBash or PowerShell. Please install the latest DVC also to ensure 
+the best experience.
```
jorgeorpinel, line 106: Interesting. I think instead we can recommend people to install the latest CLI `git` since we recomend a POSIX-like terminal earlier in this doc. And also we can mention they should check Git4Win is updated if needed. This new section can even be closer to the top of the doc. WDYT @zimka?

zimka: It seems that new users would not actually need this comment - the latest version of Git does not produce the issue. The message is only useful for users who have installed git before Aug2020, and for these users (well, at least for me personally) git update-for-windows CLI command is the simplest way to solve the problem.
AFAIK dvc is planning to replace GitPython with dulwich, and this change would hopefully solve the issue completely.
May be it is worth just to add disclaimer at the top of the page, that one should install the latest dvc and git versions as the first step when faces any issue.

jorgeorpinel: > May be it is worth just to add disclaimer at the top of the page, that one should install the latest dvc and git versions as the first step

Agree @zimka, I'll do so in a separate PR unless you'd like to make that contribution. Closing this one for now, but thanks again!

PR #2396, opened 2021-04-19
install: consistency between pip and conda pak naming

jorgeorpinel:
rel https://github.com/iterative/dvc.org/pull/2336#pullrequestreview-622899490

Ended up being just small copy edits.

### DIFF
--- a/content/docs/install/linux.md
+++ b/content/docs/install/linux.md
@@ -42,13 +42,13 @@ In this case it installs the `boto3` library along with DVC.
 > [Anaconda Distribution](https://docs.anaconda.com/anaconda/).
 
 ```dvc
-$ conda install -c conda-forge mamba # installs much faster than conda
+$ conda install -c conda-forge mamba # Installs much faster than conda
 $ mamba install -c conda-forge dvc
 ```
 
 Depending on the type of the [remote storage](/doc/command-reference/remote) you
-plan to use, you might need to install optional dependencies: `dvc-s3`,
-`dvc-azure`, `dvc-gdrive`, `dvc-gs`, `dvc-oss`, `dvc-ssh`.
+plan to use, you might need to install optional dependencies: `-s3`, `-azure`,
+`-gdrive`, `-gs`, `-oss`, `-ssh`.
 
 <details>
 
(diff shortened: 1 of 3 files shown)


content/docs/install/linux.md
```diff
@@ -42,13 +42,13 @@ In this case it installs the `boto3` library along with DVC.
...
 ```dvc
-$ conda install -c conda-forge mamba # installs much faster than conda
+$ conda install -c conda-forge mamba # Installs much faster than conda
 $ mamba install -c conda-forge dvc
 ```
 
 Depending on the type of the [remote storage](/doc/command-reference/remote) you
-plan to use, you might need to install optional dependencies: `dvc-s3`,
-`dvc-azure`, `dvc-gdrive`, `dvc-gs`, `dvc-oss`, `dvc-ssh`.
+plan to use, you might need to install optional dependencies: `-s3`, `-azure`,
```
shcheklein, line 50: why do we change this?

shcheklein (changes requested):
Not sure I understand the reasoning for this PR. What is the context?

jorgeorpinel:
> What is the context?

@shcheklein It's a left-over from a recent regular updates PR (see https://github.com/iterative/dvc.org/pull/2336#pullrequestreview-622899490). At first I thought the docs were incorrect so I started changing this doc, but when **efiop** replied confirming docs are correct, I just left some style changes in the PR.

Specifically, for `pip` only package suffixes are listed (`[ssh]`, `[s3`], etc.) but for `conda` we have entire package names (`dvc-ssh`, `dvc-s3`, etc.) Alternatively I could change the `pip` paragraph to include the full package names (`dvc[ssh]`, etc.)

Obvi not a very important change, feel free to close it.

shcheklein:
> Specifically: For pip only package suffixes are listed ([ssh], [s3], etc.), but for conda we have entire package names (dvc-ssh, dvc-s3, etc.) Alternatively I could change the pip paragraph to include the full package names (dvc[ssh], etc.)

I think this is because for `conda` those are indeed independent packages, while for `pip` those are extensions to the main `dvc` package. For `pip` it's a well known convention, while for `conda` doing something like `-ssh` feels like we inventing one? Or may be it exists and I'm not familiar

jorgeorpinel:
TBH I'm not sure, I thought they were all technically separate packages (both on `pip` and `conda`). Let me just ask @efiop (👋) and close this for now then. Thanks

Issue #2374, opened 2021-04-14
ref: experiments ignore untracked files


dberenbaum:
Document https://github.com/iterative/dvc/pull/5029.

See also https://github.com/iterative/dvc/issues/5800 and https://github.com/iterative/dvc/issues/5801 for more context. Let's document current behavior, since I think the security issues and other concerns around any proposed changes make it unlikely in the short-term.

PR #2398, opened 2021-04-19
ref: document `exp run` behavior for untracked files

pmrowla:
Document `exp run` behavior for https://github.com/iterative/dvc/pull/5029

Will close https://github.com/iterative/dvc.org/issues/2374

### DIFF
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -118,11 +118,11 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
-> ⚠️ Note that only tracked files and directories will be included in `--queue/temp`
-> experiments. To include untracked files, stage them with
+> ⚠️ Note that only tracked files and directories will be included in
+> `--queue/temp` experiments. To include untracked files, stage them with
 > `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
 > afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
-> the workspace.
+> the workspace to avoid committing secrets into experiments.
 
 Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
 performance (creates a tmp dir for each job).
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -120,7 +120,7 @@ changes between/after queueing runs.
 
 > ⚠️ Note that only tracked files and directories will be included in
 > `--queue/temp` experiments. To include untracked files, stage them with
-> `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
+> `git add` first (before `dvc exp run`). Feel free to `git reset` them
 > afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
 > the workspace to avoid committing unwanted files into experiments.
 
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -122,7 +122,7 @@ changes between/after queueing runs.
 > `--queue/temp` experiments. To include untracked files, stage them with
 > `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
 > afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
-> the workspace to avoid committing secrets into experiments.
+> the workspace to avoid committing unwanted files into experiments.
 
 Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
 performance (creates a tmp dir for each job).
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -119,7 +119,7 @@ changes between/after queueing runs.
 (e.g. on another terminal).
 
 > ⚠️ Note that only tracked files and directories will be included in `--queue/temp`
-> experiments. To include untracked files or directories, stage them with
+> experiments. To include untracked files, stage them with
 > `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
 > afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
 > the workspace.
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -121,8 +121,8 @@ changes between/after queueing runs.
 > ⚠️ Note that only tracked code and data will be included in `--queue/temp`
 > experiments. To include untracked files or directories, stage them with
 > `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
-> afterwards. Git-ignored files/dirs are explicitly excluded from the runs
-> outside the workspace.
+> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
+> the workspace.
 
 Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
 performance (creates a tmp dir for each job).
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -118,6 +118,16 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> Note that only Git-tracked and DVC-tracked files and data will be included in
+> the execution environment when running experiments outside of your workspace.
+> Git-ignored files and directories are explicitly excluded from the execution
+> environment.
+>
+> To include untracked files (such as new code or new DVC pipeline files) in the
+> experiment, you should first stage the file in Git using `git add <file>` or
+> `git add --intent-to-add <file>` and then use `dvc exp run --queue/--temp`
+> once all of your untracked files have been staged.
+
 Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
 performance (creates a tmp dir for each job).
 
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -118,15 +118,11 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
-> Note that only Git-tracked and DVC-tracked files and data will be included in
-> the execution environment when running experiments outside of your workspace.
-> Git-ignored files and directories are explicitly excluded from the execution
-> environment.
->
-> To include untracked files (such as new code or new DVC pipeline files) in the
-> experiment, you should first stage the file in Git using `git add <file>` or
-> `git add --intent-to-add <file>` and then use `dvc exp run --queue/--temp`
-> once all of your untracked files have been staged.
+> ⚠️ Note that only tracked code and data will be included in `--queue/temp`
+> experiments. To include untracked files or directories, stage them with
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,16 @@ changes between/after queueing runs.
...
 
+> Note that only Git-tracked and DVC-tracked files and data will be included in
+> the execution environment when running experiments outside of your workspace.
+> Git-ignored files and directories are explicitly excluded from the execution
+> environment.
+>
+> To include untracked files (such as new code or new DVC pipeline files) in the
+> experiment, you should first stage the file in Git using `git add <file>` or
+> `git add --intent-to-add <file>` and then use `dvc exp run --queue/--temp`
+> once all of your untracked files have been staged.
```
jorgeorpinel, line 129:
```suggestion
> ⚠️ Note that only tracked code and data will be included in `--queue/temp`
> experiments. To include untracked files or directories, stage them with
> `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
> afterwards. Git-ignored files/dirs are explicitly excluded from the runs
> outside the workspace.
```

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked code and data will be included in `--queue/temp`
+> experiments. To include untracked files or directories, stage them with
+> `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
+> afterwards. Git-ignored files/dirs are explicitly excluded from the runs
+> outside the workspace.
```
jorgeorpinel, line 125:
```suggestion
> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
> the workspace.
```

jorgeorpinel (approved):
WDYT @pmrowla ?

pmrowla:
Changes look fine to me

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked code and data will be included in `--queue/temp`
```
dberenbaum, line 121:
```suggestion
> ⚠️ Note that only tracked files and directories will be included in `--queue/temp`
```
Looks okay as is, but seems like `files and directories` is more general (covers params) and consistent with the rest of the paragraph.

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
...
+> experiments. To include untracked files or directories, stage them with
+> `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
+> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
+> the workspace.
```
dberenbaum, line 125:
```suggestion
> the workspace to avoid tracking unnecessary or sensitive information.
```
Wording might not be right, but might be good to include rationale.

dberenbaum (approved):

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,16 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> Note that only Git-tracked and DVC-tracked files and data will be included in
+> the execution environment when running experiments outside of your workspace.
+> Git-ignored files and directories are explicitly excluded from the execution
+> environment.
+>
+> To include untracked files (such as new code or new DVC pipeline files) in the
```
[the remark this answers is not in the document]
jorgeorpinel: Looks like you reviewed an old version somehow @dberenbaum, the "pipelines file" phrase is no longer there.

> What if the user had a new data dependency? Would they first need to dvc add newdata and then git add newdata.dvc?

Good Q. Should it get auto-committed (cached) now as long as it's in dvc.yaml `deps`? Cc @pmrowla

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked code and data will be included in `--queue/temp`
+> experiments. To include untracked files or directories, stage them with
+> `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
```
jorgeorpinel, line 123:
```suggestion
> `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
```

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,16 @@ changes between/after queueing runs.
...
 (e.g. on another terminal).
 
+> Note that only Git-tracked and DVC-tracked files and data will be included in
+> the execution environment when running experiments outside of your workspace.
+> Git-ignored files and directories are explicitly excluded from the execution
+> environment.
+>
+> To include untracked files (such as new code or new DVC pipeline files) in the
+> experiment, you should first stage the file in Git using `git add <file>` or
+> `git add --intent-to-add <file>` and then use `dvc exp run --queue/--temp`
```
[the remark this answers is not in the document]
jorgeorpinel: I was a bit confused by that too. My understanding was that it's a better option but only available for completely new files (so I put it in parenthesis like `git add (-N)` (the snippet above is outdated).

But now that I think about it, we're talking about totally new files here anyway... So removed the parentheses for now.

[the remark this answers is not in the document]
jorgeorpinel: > we're talking about totally new files here

Is that right though? Meaning, are changed files are included automatically (in their latest version)? Sorry, keep forgetting... I think so. Cc @pmrowla

content/docs/command-reference/exp/run.md
jorgeorpinel: True. Committing. But then I think I'll change the next occurrence to just "files" (Git doesn't track dirs per-se anyway).

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked files and directories will be included in `--queue/temp`
+> experiments. To include untracked files or directories, stage them with
```
jorgeorpinel, line 122:
```suggestion
> experiments. To include untracked files, stage them with
```

content/docs/command-reference/exp/run.md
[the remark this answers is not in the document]
dberenbaum: Huh, this is an old comment, not sure why it shows as recent. I'll resolve to avoid confusion.

content/docs/command-reference/exp/run.md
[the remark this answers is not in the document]
dberenbaum: This was also an old comment. I think parentheses are fine, so feel free to add that back.

content/docs/command-reference/exp/run.md
dberenbaum: This comment is new.

jorgeorpinel: I'll add this info.

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked files and directories will be included in `--queue/temp`
+> experiments. To include untracked files, stage them with
+> `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
+> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
+> the workspace.
```
jorgeorpinel, line 125:
```suggestion
> ⚠️ Note that only tracked files and directories will be included in
> `--queue/temp` experiments. To include untracked files, stage them with
> `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
> the workspace to avoid committing secrets into experiments.
```

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
 `dvc exp run --temp`, for example to continue working on the project meanwhile
 (e.g. on another terminal).
 
+> ⚠️ Note that only tracked files and directories will be included in
+> `--queue/temp` experiments. To include untracked files, stage them with
+> `git add -N` first (before `dvc exp run`). Feel free to `git reset` them
```
dberenbaum, line 123:
```suggestion
> `git add (-N)` first (before `dvc exp run`). Feel free to `git reset` them
```
Do you mind adding the parentheses back? Otherwise, it seems like `-N` is required.

content/docs/command-reference/exp/run.md
```diff
@@ -118,6 +118,12 @@ changes between/after queueing runs.
...
+> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
+> the workspace to avoid committing secrets into experiments.
```
dberenbaum, line 125:
```suggestion
> the workspace to avoid committing unwanted files into experiments.
```
I wouldn't restrict it to secrets. It could also be lots of extraneous files, like the venv scenario.

content/docs/command-reference/exp/run.md
jorgeorpinel: I think it's the recommended way. Much safer than without `-N` for this trick, I think!

This is a temporary patch anyway right? Behavior-wise. It may take a while to decide what to do but still...

dberenbaum: > Much safer than without `-N` for this trick

Is it? Looking at https://stackoverflow.com/questions/24329051/what-does-git-add-intent-to-add-or-n-do-and-when-should-it-be-used, my understanding is that `-N` is only useful to diff new files and edit their patches line by line. Once a file is staged/tracked by Git, experiments should track all changes to that file regardless of how it was added.

Since `-N` is not a common option, I worry it will confuse people if we recommend it by default. If you think it's not likely to be confusing regardless, I'm fine letting it go.

jorgeorpinel: I think it's safer because the file contents don't actually get staged to Git, so even if you forget to `git reset` after, you'll only end up with some empty files in your main Git history instead of potentially unwanted contents. Either way it's slightly a hacky trick, hopefully a temporary patch.

I do agree it's uncommon somay be weary about unknown implications of `-N`... Idk. I'd wait on @pmrowla's opinion on whether to keep `-N` or not, but mentioning both probably adds to confusion.

jorgeorpinel: Yeah you know what `-N` seems unnecessary since they'll be committed in the experiment anyway. Merging without it for now but again, we can revisit.

jorgeorpinel, line 123:
```suggestion
> `git add` first (before `dvc exp run`). Feel free to `git reset` them
```

dberenbaum: Yup, you put it better than I did.

PR #2401, opened 2021-04-20
fix Mac OS -> macOS

shcheklein:
Thanks @wbednarski for noticing this

### DIFF
--- a/content/docs/sidebar.json
+++ b/content/docs/sidebar.json
@@ -10,7 +10,7 @@
     "source": "install/index.md",
     "children": [
       {
-        "label": "Mac OS",
+        "label": "macOS",
         "slug": "macos"
       },
       "windows",
--- a/src/components/DownloadButton/index.tsx
+++ b/src/components/DownloadButton/index.tsx
@@ -26,7 +26,7 @@ const itemsByOs = {
     download: false
   },
   [OS.OSX]: {
-    title: 'Mac OS',
+    title: 'macOS',
     url: `/osxpkg/dvc-${VERSION}.pkg`,
     download: true
   },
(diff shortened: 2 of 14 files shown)


content/docs/sidebar.json
```diff
@@ -10,7 +10,7 @@
     "source": "install/index.md",
     "children": [
       {
-        "label": "Mac OS",
+        "label": "macOS",
```
jorgeorpinel, line 13:
```suggestion
        "label": "MacOS",
```

src/components/DownloadButton/index.tsx
```diff
@@ -26,7 +26,7 @@ const itemsByOs = {
     download: false
   },
   [OS.OSX]: {
-    title: 'Mac OS',
+    title: 'macOS',
```
jorgeorpinel, line 29:
```suggestion
    title: 'MacOS',
```

jorgeorpinel (approved):
"MacOS" is also correct, I think. `macOS` is more of a logo. Formally speaking any proper name starts with a capital letter and in lists like "Linux, Windows, MacOS" looks more consistent (reason why we had it that way). But any way is fine probably, as long as we use the same form everywhere.

In case we roll back I'd just make these 2 changes below (removing the space):

content/docs/sidebar.json
shcheklein: macOS is the official name for the OS (like FreeBSD would not be Freebsd for consistency). I think logo being the same is just a coincidence. 

> Formally speaking any proper name starts with a capital letter

brand names are written the way the are written I think. We can refer to the Wiki page - https://en.wikipedia.org/wiki/MacOS - they use macOS everywhere.

jorgeorpinel: Sent from an Iphone 😋

PR #2415, opened 2021-04-24
ref: updates to S3-compat info in `remote modify`

jorgeorpinel:
rel https://discuss.dvc.org/t/allaccessdisabled-error-for-mino-set-as-remote/735/2


jorgeorpinel:
Could you try to review this update @iesahin ? Thanks!

content/docs/command-reference/remote/add.md
```diff
@@ -125,34 +125,40 @@ So, make sure you have the following permissions enabled:
...
+configure the remote's `endpointurl`:
 
 ```dvc
 $ dvc remote add -d myremote s3://mybucket/path
 $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-> See `dvc remote modify` for a full list of S3 API parameters.
+It's also important to setup appropriate authentication with any of the related
```
iesahin, line 142: How about: 

⚠️ It's important to setup appropriate authentication with `dvc remote modify`. Otherwise, DVC uses default AWS credentials which may cause an error.

content/docs/command-reference/remote/modify.md
```diff
@@ -288,23 +288,24 @@ For more information about the variables DVC supports, please visit
...
 ```dvc
 $ dvc remote add -d myremote s3://mybucket/path
 $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-Besides that, any parameters that are available for Amazon S3 (see previous
-section) may be available for S3 compatible storage. For example, let's setup a
-DVC remote using the `example-name`
+It's also important to setup appropriate authentication with any of the related
```
iesahin, line 302: How about: 

⚠️  It's important to setup appropriate authentication with `dvc remote modify`. Otherwise, DVC uses the default AWS credentials which may cause an error.

iesahin (commented):
Thank you 🙏🏼  @jorgeorpinel . I added two comments for more concise explanations.

content/docs/command-reference/remote/add.md
```diff
@@ -125,34 +125,40 @@ So, make sure you have the following permissions enabled:
...
+parameters available for regular S3 storage (see `dvc remote modify`).
+Otherwise, DVC may try to use default AWS credentials (if any), which will
+probably cause an error. Example:
+
+```dvc
+$ dvc remote modify myremote profile 'myprofile'
```
shcheklein, line 148: Let's not push to specific methods in `remote add` (neither in `remote modify`).

I would also simplify the whole message. The is not that much specific about S3-compatible storage. Same happens when you have multiple S3 profiles. Regular credentials are being used and you have to specify auth options to make sure that the correct profile is set.

shcheklein: I would replicate

> By default, DVC expects your AWS CLI is already configured. DVC will be using default AWS credentials file to access S3. To override some of these parameters, use the parameters described in dvc remote modify.

Add one sentence like - make sure that the it uses right credentials, especially if you have multiple S3 storages ...


But again, this problem happens with a regular S3 setup

jorgeorpinel: Agree, will simplify ⌛  But

> The is not that much specific about S3-compatible storage.

I think the special part is that you MUST setup auth with S3-compatible, as default AWS creds are pretty much guaranteed to fail. Maybe DVC shouldn't even try them if the `url` isn't .amazonaws.com ... But that's a separate question.

content/docs/command-reference/remote/add.md
jorgeorpinel: Good rewrite! I'll resubmit it as an actionable GH suggestion and tag you. Resolving here

content/docs/command-reference/remote/add.md
```diff
@@ -125,34 +125,40 @@ So, make sure you have the following permissions enabled:
...
 $ dvc remote add -d myremote s3://mybucket/path
 $ dvc remote modify myremote endpointurl \
                     https://object-storage.example.com
 ```
 
-> See `dvc remote modify` for a full list of S3 API parameters.
+It's also important to setup appropriate authentication with any of the related
+parameters available for regular S3 storage (see `dvc remote modify`).
+Otherwise, DVC may try to use default AWS credentials (if any), which will
+probably cause an error. Example:
```
jorgeorpinel, line 145: Resubmitting https://github.com/iterative/dvc.org/pull/2415#discussion_r619635067:
```suggestion
⚠️ It's important to setup appropriate authentication with `dvc remote modify`.
Otherwise, DVC may use default AWS credentials, which would cause an error.
```
And now I just merge it directly in the PR. Cc @iesahin

content/docs/command-reference/remote/modify.md
```diff
@@ -288,23 +288,24 @@ For more information about the variables DVC supports, please visit
...
+parameters available for regular S3 storage (see previous section). Otherwise,
+DVC may try to use default AWS credentials (if any), which will probably cause
+an error.
```
jorgeorpinel, line 305: Resubmitting https://github.com/iterative/dvc.org/pull/2415#discussion_r619635479:
```suggestion
It's also important to setup appropriate authentication with any of the related
parameters available for regular S3 storage (see previous section). Otherwise,
DVC may try to use default AWS credentials (if any), which will probably cause
an error.
```
Cc @iesahin

content/docs/command-reference/remote/modify.md
jorgeorpinel: Moved to https://github.com/iterative/dvc.org/pull/2415#pullrequestreview-644086184.

content/docs/command-reference/remote/modify.md
jorgeorpinel: BTW no problem if the format isn't perfect, Restyled can fix it later. Committing...

content/docs/command-reference/remote/add.md
jorgeorpinel: OK I moved most info out of the S3-compatible sections. But I didn't replicate the same text as in the S3 section yet because, if we go that route, why have S3-compatible sections at all? If there truly isn't anything special about it, we should be able to merge the sections (just update the title a little). WDYT @shcheklein ?

shcheklein: > If there truly isn't anything special about it

yes, the only special thing is the `endpointurl` + mention their names for SEO/search/visibility

everything else is not special

shcheklein: > DVC shouldn't even try them if the url isn't .amazonaws.com

I would not do this. We don't know in advance the setup, people might use some proxy to access S3. And there is not way to say is it S3 credential or not - that's the whole point- they look the same as far as I understand, they come from the same file, same mechanism to use them, etc, etc

iesahin (approved):
Thank you @jorgeorpinel 💥 . I'm all fine with this if @shcheklein 's comments are also addressed. 

For SEO purposes, we may have some platform-specific pages for each of the well-known S3-compatible options.

content/docs/command-reference/remote/add.md
jorgeorpinel: > everything else is not special

@shcheklein I did also leave a warning about authentication at the bottom which I consider useful. But lmk if you think it should go and I'll follow up. Merging for now as I assume you looked again yet made no more comments + it's approved by @iesahin (ty BTW)

shcheklein: My point about that note that it feels wrong to me. There is nothing wrong with this `DVC will try to use default AWS credentials, which may cause an error.`. And even more - that's what you want to do, you use the same regular mechanisms. I probably understand what you were trying to say here, but it's an edge case and should be written in a different way. Not it looks confusing (why would I even care about this?) or as a bug (that what I actually want).

content/docs/command-reference/remote/add.md
```diff
@@ -104,8 +104,18 @@ By default, DVC authenticates using your AWS CLI
 (if set). This uses the default AWS credentials file. To use a custom
 authentication method, use the parameters described in `dvc remote modify`.
 
-We use the `boto3` library to communicate with AWS. The following API methods
-are performed:
+S3 remotes can also be configured entirely via environment variables, e.g.:
```
shcheklein, line 107: This should not be part of this PR?

I also don't understand this change ... we don't include in `dvc remote add` any specific details ... it should be part of the `dvc remote modify`

content/docs/command-reference/remote/modify.md
```diff
@@ -273,38 +273,54 @@ Operational parameters:
...
 ```dvc
 $ export AWS_ACCESS_KEY_ID='mykey'
 $ export AWS_SECRET_ACCESS_KEY='mysecret'
 $ dvc remote add -d myremote s3://mybucket/path
 ```
 
-For more information about the variables DVC supports, please visit
-[boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variable-configuration)
+For more on the supported env vars, please see the
+[boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variable-configuration)
```
shcheklein, line 285: this actually, can stay as part of the `dvc remote add` - since it's not about auth or config, it's more about remote spec and what to expect

content/docs/command-reference/remote/add.md
jorgeorpinel: Hmmm. Based on a few support cases I thought that it's a bit more common than an edge case. Or do you mean you can use AWS credentials for Minio, DigitalOcean, etc? TBH I've never used those services.

OK @shcheklein, for now I copied the same note from the regular S3 section into here like you suggested in https://github.com/iterative/dvc.org/pull/2415#discussion_r619721921. And try to simplify again... See #2441

content/docs/command-reference/remote/add.md
jorgeorpinel: It was out of scope based on the PR title, sorry. The scope expanded a bit during the review process.

But we do mention env vars in the Azure (and OSS) sections of `remote add`, and those are the only other types that you can config with env vars I think, at least based on `remote modify` info (maybe also GDrive but those details are in it's own guide).

Made all the env var mentions consistent in b2372e7e for now (see #2441)

content/docs/command-reference/remote/modify.md
jorgeorpinel: Oh OK I see you meant the `boto3` permission stuff earlier. You're right about that part being too detailed, removed from `remote add` in 10f177fa.

shcheklein: I was not suggesting to remove it :) It has certain value to it - we specify the list of permissions that admins would need to enable for DVC to operate. We might change language a bit, but this information is good overall.

content/docs/command-reference/remote/add.md
shcheklein: A think a lot of remotes can be configured with env vars.

All this additional config should be part of modify to my mind. Things like

```
export ..
export
dvc remote add
```

do not make much sense overall. 

```
export ...
export ...
```

are important when you run `dvc push`/`dvc pull`. It's a way to provide values to auth flow used in DVC when it tries to access. Other ways - default config files (`.aws/credentials`) or DVC config (the one you modify with `dvc remote modify`)

content/docs/command-reference/remote/add.md
shcheklein: You can use the same credentials _mechanisms_ (same config files, same env vars, etc). It's completely interchangeable with regular S3 in this way. The only thing that you need to provide is the endpointurl to override the default AWS's one.

For example, when you have multiple S3 accounts you will have to distinguish them somehow too (by specifying the `profile` name). Here is the same- if you use S3 _and_ minio you will have multiple profiles most likely.

jorgeorpinel: > You can use the same credentials mechanisms (same config files
when you have multiple S3 accounts you will have to distinguish them somehow

Right. But if you use AWS CLI (not necessarily for S3) \*and\* separately try to setup an S3-compatible remote by only changing the `endpoint` like we say, `push` fails because DVC tries to connect using `~/.aws/config`, which has Amazon-specific credentials. And the error message is about having wrong credentials, instead of a hint to configure the credentials (since none are set, really).

Perhaps it's an edge case (-compatible remotes aren't used often enough), not worth targeting in docs.

jorgeorpinel: p.s. there's an example support case, linked from the PR description.

shcheklein: Same works for two S3 accounts - you specify URL s3://org1-some-bucket but it's using credentials from org2 (because it's set by default in the config). You will have the exact same error message, right?

jorgeorpinel: Yes.

Issue #2397, opened 2021-04-19
ref: clarify `exp apply` behavior


pmrowla:
Noticed this issue while reviewing https://github.com/iterative/dvc.org/pull/2383

The docs for `dvc exp apply` currently state that:

> Restores the results of a given experiment (name or hash are accepted, see dvc exp run for details) into the workspace. This means changing the appropriate metafiles, parameter files, metrics, plots, and corresponding DVC-tracked data.
>
> This is similar to dvc checkout, but for dvc experiments.

- [x] 1. `exp apply` will restore the changes to *any* (git or DVC tracked) file from a given experiment. It is not specific to DVC-related files and data. Rather than a simple `dvc checkout`, it's closer to:

```
git checkout <exp-sha>
dvc checkout
```

(specifically, it's more like a `git stash apply` or `git merge --no-commit` than `git checkout`, but typical DVC users are probably not familiar with those commands)

- [ ] 2. The docs should mention that `exp apply` is potentially destructive, and will overwrite any uncommitted changes to the user's (git) workspace, if they conflict with whatever changes exist in the experiment.
  UPDATE: Jump to https://github.com/iterative/dvc.org/issues/2397#issuecomment-827930004

shcheklein:
> and will overwrite any uncommitted changes 

will it drop changes that do not conflict?

pmrowla:
@shcheklein no, any changes which don't conflict will be left alone. It basically just does a git merge/checkout with `--theirs`, meaning that if there are conflicts the version from the experiment is automatically preferred.

Originally, `exp apply` would error out on any conflicts and had a `-f/--force` option to do the `--theirs` overwrite automatically. But based on feedback from @dmpetrov, the typical use case for `exp apply` meant that users would just always need to run it with `--force`.

shcheklein:
I would clarify this. It's important that changes are not wiped out but rather merged with the experiment. I'm a bit worried that pretty much silently user may end up with a mix of changes here.

Does it makes sense to go even further and do `reset --hard`  ?

jorgeorpinel:
> important that changes are not wiped out but rather merged with the experiment. I'm a bit worried

Should we transfer to the core repo for now? I'll address part 1. for now (in #2417)

pmrowla:
[...]

PR #2417, opened 2021-04-25
ref: corrections to `exp run/apply/branch` +

### LINKED ISSUE #2399: Improve explanation of error handling for interrupting checkpoints
See https://github.com/iterative/dvc-checkpoints-mnist/pull/2 for background.

Here's the scenario:

1. My script includes a checkpoint (using `make_checkpoint()` or similar) but no error handling for keyboard interuptions.
2. I run an experiment and interrupt it with ctrl-c, which returns a non-0 exit code.
3. I resume the experiment.

If there were changes between the last time `make_checkpoint()` was called and when the experiment was interrupted, step 3 will create a new experiment because there was no final cleanup checkpoint applied at the end of step 2 (if the script exited cleanly with a 0 exit code, there would be a final cleanup checkpoint).

I couldn't figure this out until @pmrowla explained it, so IMO it's not going to be clear to users. Since we recommend this ctrl-c workflow in the docs and in our examples, I think we need to better document it. Thoughts?

jorgeorpinel:
- Addresses https://github.com/iterative/dvc.org/pull/2243#pullrequestreview-603422799 and https://github.com/iterative/dvc.org/pull/2286#pullrequestreview-615217332 (finally)
- And 1. from #2397 and close #2399 (inc. `make_checkpoint()` update)
- Other improvements and copy edits

### DIFF
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -67,19 +67,23 @@ import os
 from dvc.api import make_checkpoint
 
 while True:
-    if os.path.exists("int.txt"):
-        with open("int.txt", "r") as fd:
-            i_ = int(fd.read()) + 1
-    else:
-        i_ = 0
+    try:
+        if os.path.exists("int.txt"):
+            with open("int.txt", "r") as fd:
+                i_ = int(fd.read()) + 1
+        else:
+            i_ = 0
 
-    # ... do something meaningful
+        # ... do something meaningful
 
-    with open("int.txt", "w") as fd:
-        fd.write(f"{i_}")
+        with open("int.txt", "w") as fd:
+            fd.write(f"{i_}")
 
-    if i_ % 100 == 0:
-        make_checkpoint()
+        if i_ % 100 == 0:
+            make_checkpoint()
+
+    except KeyboardInterrupt:
+    exit()
 ```
 
 Since `checkpoint` outputs in effect implement a circular dependency,
@@ -102,6 +106,9 @@ Checkpoint experiment iteration '75ff5e0'.
 Reproduced experiment(s): exp-8a3bd
 Experiment results have been applied to your workspace.
 ```
+
+> ⚠️ it's important to handle interruptions or any other errors in your code for
+> DVC checkpoints to behave as expected.
 
 In this example we killed the process (with Ctrl + C) after 3 checkpoints (at 0,
 100, and 200 `i_`). The <abbr>cache</abbr> will contain those 3 versions of
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -83,7 +83,7 @@ while True:
             make_checkpoint()
 
     except KeyboardInterrupt:
-    exit()
+        exit()
 ```
 
 Since `checkpoint` outputs in effect implement a circular dependency,
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -17,17 +17,17 @@ positional arguments:
 Restores the results of a given `experiment` (name or hash are accepted, see
 `dvc exp run` for details) into the workspace. This means changing the
 appropriate [metafiles](/doc/user-guide/project-structure),
-<abbr>parameter</abbr> files, <abbr>metrics</abbr>, <abbr>plots</abbr>, and
+<abbr>parameters</abbr>, <abbr>metrics</abbr>, <abbr>plots</abbr>, and
 corresponding DVC-tracked data.
 
 > This is similar to `dvc checkout`, but for `dvc experiments`.
 
-This is typically used after using `dvc exp show` or `dvc exp diff` to find the
-best experiment, and before committing to Git in order to make it
-[persistent](/doc/user-guide/experiment-management#persistent-experiments).
+This command is typically used after finding a target experiment with
+`dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
+[persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-Note that this command will fail if the target `experiment` was not derived from
-the current Git commit.
+Note that this command may fail if the current Git commit (`HEAD`) has changed
+significantly from the time the `experiment` was run.
 
 ## Options
 
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -17,14 +17,15 @@ positional arguments:
 Restores a given `experiment` (name or hash are accepted, see `dvc exp run` for
 details) into the workspace. This means changing all files (code, data,
 <abbr>parameters</abbr>, <abbr>metrics</abbr>, etc.) needed to reflect the
-experiment conditions and results in the <abbr>workspace</abbr>.
+experiment conditions and results in the workspace.
 
 This command is typically used after finding a target experiment with
 `dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
 [persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-Note that this command may fail if the current Git commit (`HEAD`) has changed
-significantly from the time the `experiment` was run.
+Note that this command will fail if the current Git commit (`HEAD`) has changed
+from the time the `experiment` was run, or if changes in the
+<abbr>workspace</abbr> conflict with the desired experiment files.
 
 ## Options
 
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -19,12 +19,15 @@ details) into the workspace. This means changing all files (code, data,
 <abbr>parameters</abbr>, <abbr>metrics</abbr>, etc.) needed to reflect the
 experiment conditions and results in the workspace.
 
-This command is typically used after finding a target experiment with
+This command is typically used after finding a target `experiment` with
 `dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
 [persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-Note that this command will fail if the current Git commit (`HEAD`) has changed
-from the time the `experiment` was run, or if changes in the
+Note that any [checkpoints](/doc/command-reference/exp/run#checkpoints) found in
+the `experiment` are not preserved when applying and committing it.
+
+⚠️ This command will fail if the current Git commit (`HEAD`) has changed from
+the time the experiment in question was run, or if changes in the
 <abbr>workspace</abbr> conflict with the desired experiment files.
 
 ## Options
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -14,13 +14,10 @@ positional arguments:
 
 ## Description
 
-Restores the results of a given `experiment` (name or hash are accepted, see
-`dvc exp run` for details) into the workspace. This means changing the
-appropriate [metafiles](/doc/user-guide/project-structure),
-<abbr>parameters</abbr>, <abbr>metrics</abbr>, <abbr>plots</abbr>, and
-corresponding DVC-tracked data.
-
-> This is similar to `dvc checkout`, but for `dvc experiments`.
+Restores a given `experiment` (name or hash are accepted, see `dvc exp run` for
+details) into the workspace. This means changing all files (code, data,
+<abbr>parameters</abbr>, <abbr>metrics</abbr>, etc.) needed to reflect the
+experiment conditions and results in the <abbr>workspace</abbr>.
 
 This command is typically used after finding a target experiment with
 `dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -24,7 +24,7 @@ This command is typically used after finding a target `experiment` with
 [persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
 Note that any [checkpoints](/doc/command-reference/exp/run#checkpoints) found in
-the `experiment` are not preserved when applying and committing it.
+the `experiment` are **not** preserved when applying and committing it.
 
 ⚠️ This command will fail if the current Git commit (`HEAD`) has changed from
 the time the experiment in question was run, or if changes in the
--- a/content/docs/command-reference/exp/apply.md
+++ b/content/docs/command-reference/exp/apply.md
@@ -23,8 +23,9 @@ This command is typically used after finding a target `experiment` with
 `dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
 [persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-Note that any [checkpoints](/doc/command-reference/exp/run#checkpoints) found in
-the `experiment` are **not** preserved when applying and committing it.
+Note that the history of
+[checkpoints](/doc/command-reference/exp/run#checkpoints) found in the
+`experiment` are **not** preserved when applying and committing it.
 
 ⚠️ This command will fail if the current Git commit (`HEAD`) has changed from
 the time the experiment in question was run, or if changes in the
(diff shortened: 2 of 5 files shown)


content/docs/command-reference/exp/branch.md
```diff
@@ -15,17 +15,19 @@ positional arguments:
...
-In most cases this is similar to using `dvc exp apply` (applies the `experiment`
-results to the workspace) followed by Git branching and committing, except that
-`dvc exp branch` **does not** switch into the created `branch`. Equivalent to
-this:
+Makes a named Git
+[`branch`](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging)
+containing the target `experiment` (making it
+[persistent](/doc/user-guide/experiment-management#persistent-experiments)). The
+branch is based on the commit that was active (`HEAD`) at the time that the
+target experiment was run.
```
dberenbaum, line 23: I don't think there's anything wrong with this sentence, but I found it a little confusing. Is this necessary? Seems more like a general explanation of experiments than anything specific to `exp branch`, but I might be missing the point you're trying to convey.

content/docs/command-reference/exp/branch.md
```diff
@@ -15,17 +15,19 @@ positional arguments:
...
+
+This command is useful, for example, in cases where `dvc exp apply` fails due to
+significant diversion between the project version at the time of the experiment
+run and the current <abbr>workspace</abbr>.
```
dberenbaum, line 27: I would say there are a few other uses:
1. Wanting to preserve multiple experiments by creating a branch from each.
2. Wanting to preserve experiments without needing to apply them to the workspace (if you're in the middle of experimenting with other things).
3. Wanting to preserve checkpoint commits in your Git history (instead of squashing them like `exp apply` would do).
It's not necessary to be exhaustive, but I actually think that `exp branch` primarily for these reasons rather than as a backup to `exp apply`.

content/docs/command-reference/exp/branch.md
```diff
@@ -15,17 +15,19 @@ positional arguments:
...
+
+Note that `dvc exp branch` **does not** switch into the new branch created. It's
+equivalent to:
```
dberenbaum, line 30: My vote is still to drop this since it's slightly inaccurate, and I don't think it adds much that's not already covered by the description, but I know we discussed it in a previous PR, so 🤷 .

content/docs/command-reference/exp/branch.md
```diff
@@ -34,10 +36,12 @@ this:
 [branch] $ git checkout master
 ```
 
-For [checkpoints](/doc/command-reference/exp/run#checkpoints), the `experiment`
-(custom Git branch with multiple commits) is merged into the new `branch`.
+For [checkpoints](/doc/command-reference/exp/run#checkpoints), the new branch
+will contain multiple commits (the checkpoints).
```
dberenbaum, line 40: 🔥

content/docs/command-reference/exp/apply.md
```diff
@@ -14,20 +14,17 @@ positional arguments:
...
+[persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-This is typically used after using `dvc exp show` or `dvc exp diff` to find the
-best experiment, and before committing to Git in order to make it
-[persistent](/doc/user-guide/experiment-management#persistent-experiments).
-
-Note that this command will fail if the target `experiment` was not derived from
-the current Git commit.
+Note that this command may fail if the current Git commit (`HEAD`) has changed
+significantly from the time the `experiment` was run.
```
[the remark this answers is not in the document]
pmrowla: It fails outright if your current `HEAD` commit does not match the `HEAD` commit from when you ran the experiment.

It may also fail if there are git conflicts between changes you have in your workspace and changes in the experiment you want to apply. We try to automatically resolve the conflicts by preferring the changes in the applied experiment, but sometimes git may still fail to auto-resolve them.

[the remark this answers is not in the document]
jorgeorpinel: Updated in 8762f974. PTAL

> We try to automatically resolve the conflicts by preferring the changes in the applied experiment

Meaning `git merge --theirs`, right @pmrowla ? I think it's not considered a conflict until after Git can't merge it so I went with the phrase "if they conflict", for now.

> Where  they have moved their HEAD ... and then try to exp apply some experiment

Does/should the CLI print a HINT to try `exp branch` instead?

content/docs/command-reference/exp/branch.md
jorgeorpinel: Yeah maybe it doesn't belong in the first paragraph. But I think it's key that people understand that the branch won't be based on their current `HEAD`. BTW the sample `git` commands below are still incorrect, I see... ⌛

jorgeorpinel: Changed a bit 2685e741 for now, but I may change it per your other feedback...

content/docs/command-reference/exp/apply.md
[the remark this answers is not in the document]
pmrowla: No, it will give an error along the lines of "cannot apply `exp-...` onto commit `<user's current HEAD>`. Experiment is derived from `<original experiment HEAD>`"

content/docs/command-reference/exp/branch.md
jorgeorpinel: Yeah, very good summary! I put this info in (with some addendums) in 2510027e, but still keeping the `exp apply` contrasting 😬

content/docs/command-reference/exp/branch.md
jorgeorpinel: Hmmm yeah. I think you're right. ✖️'d

content/docs/command-reference/exp/branch.md
jorgeorpinel: 👍

content/docs/command-reference/exp/apply.md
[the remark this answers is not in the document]
jorgeorpinel: > exp branch is for a different use case than exp apply

OK @pmrowla. Would you mind checking the bullet list in the latest version to confirm that it covers the use case(s)? Thanks

content/docs/command-reference/exp/branch.md
```diff
@@ -15,29 +15,29 @@ positional arguments:
...
+time that the experiment was run). Note that unlike `git branch`, DVC **does
+not** switch into the new `branch`.
+
+`dvc exp branch` is useful for several scenarios, including:
+
+- To make one or more experiments persistent without applying them to the
+  workspace, so they can be
+  [stored and shared](https://dvc.org/doc/use-cases/sharing-data-and-model-files)
+  in a normal Git+DVC workflow;
+- In cases where `dvc exp apply` fails;
```
pmrowla, line 35: This case doesn't make sense to me, `exp branch` is not a replacement for `exp apply`. `exp branch` is for creating a Git branch that contains all of the commits in an experiment run.

`exp apply` is for applying the (merged) contents of a single experiment commit into my current workspace.

content/docs/command-reference/exp/apply.md
```diff
@@ -14,20 +14,21 @@ positional arguments:
 
 ## Description
 
-Restores the results of a given `experiment` (name or hash are accepted, see
-`dvc exp run` for details) into the workspace. This means changing the
-appropriate [metafiles](/doc/user-guide/project-structure),
-<abbr>parameter</abbr> files, <abbr>metrics</abbr>, <abbr>plots</abbr>, and
-corresponding DVC-tracked data.
+Restores a given `experiment` (name or hash are accepted, see `dvc exp run` for
```
iesahin, line 17: Could we make the statement in parentheses a new sentence?

Restores a given `experiment` into the workspace. Experiment can be referenced by a hash or name.

content/docs/command-reference/exp/branch.md
```diff
@@ -15,29 +15,29 @@ positional arguments:
...
+Makes a named Git
+[`branch`](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging)
+containing the target `experiment` (making it
+[persistent](/doc/user-guide/experiment-management#persistent-experiments)). For
+[checkpoint experiments](/doc/command-reference/exp/run#checkpoints), the new
+branch will contain multiple commits (the checkpoints).
+
+The new `branch` will be based on the experiment's parent commit (`HEAD` at the
+time that the experiment was run). Note that unlike `git branch`, DVC **does
+not** switch into the new `branch`.
```
dberenbaum, line 27:
```suggestion
time that the experiment was run).
```
`git branch` also doesn't switch to the new branch, right? Is the point of this paragraph to clarify that the current HEAD/branch has no relation to `exp branch`?

content/docs/command-reference/exp/run.md
```diff
@@ -81,16 +81,20 @@ Then, in your code either call the `dvc.api.make_checkpoint()` function
...
-will be preserved. When a run finishes normally, a final checkpoint will be
-added (if needed) to wrap up the experiment.
+You can now use `dvc exp run` to begin the experiment. All checkpoints
+registered at runtime will be preserved even if the process gets interrupted
+(e.g. with `[Ctrl] C`, or by an error\*). A final checkpoint will be added (if
+needed) to wrap up the experiment.
+
+> \* The stage command(s) should return a non-error exit code (`0`) for the
+> final checkpoint to happen. Otherwise, it will not be possible to continue
+> this experiment as described below (it will restart from scratch).
```
dberenbaum, line 91: It shouldn't start from scratch, it will just potentially start a modified experiment if there are "dirty" changes left over that happened after the last checkpoint was created.

content/docs/command-reference/exp/branch.md
jorgeorpinel: How would you "apply" a past non-checkpoint experiment (once `HEAD` has moved)?

content/docs/command-reference/exp/apply.md
jorgeorpinel: Doesn't seem like a sentence pertaining to that paragraph to me. Would have to be another p just for the argument (and maybe some of the options), which I'm trying to avoid.

content/docs/command-reference/exp/branch.md
jorgeorpinel: Ah yeah I was thinking `git checkout -b` I guess. Fixed

The paragraph clarifies where the branch is created, yes. Since the fact that exps are already commits is hidden from the user, it shouldn't be obvious to them that this just relabels an existing custom Git ref. or what it's tree is.

content/docs/command-reference/exp/run.md
jorgeorpinel: OK, I wasn't sure about this... (removed that last sentence for now). In #2399 you mentioned:

> step 3 will create a new experiment because there was no final cleanup checkpoint

But I didn't test what this looks exactly. What's a "new experiment" here, if it does continue from the last checkpoint?

content/docs/command-reference/exp/branch.md
pmrowla: You can't. The only way you can use `dvc exp apply` in this scenario is with
```
git checkout <old-head-commit>
exp apply <exp>
```

If by "apply" here you really mean that you want to merge or rebase the changes from an experiment onto your new HEAD, you have to do it manually with the appropriate git commands. see also: https://github.com/iterative/dvc/issues/5782

jorgeorpinel: What about something like
```console
$ dvc exp branch exp-12341 exp-branch
$ git merge exp-branch
```
? (half DVC half Git)

pmrowla: yes that will work, but it has the side effect of creating a full Git branch containing your experiment. If you have a ton of experiments you want to test against your new HEAD in this way, this is not ideal, since you probably don't actually want full Git branches for every one of those experiments.

jorgeorpinel: For now maybe
```suggestion
- In cases where it's not possible to `dvc exp apply` an experiment directly;
```

jorgeorpinel: Rephrased in 68892454. WDYT? If it feels to forced, obvious, or indirect we could just remove it indeed. But then there's only 2 items in this bullet list and they kind of relate so would need to rewrite as a single scenario/motivation paragraph.

pmrowla: I still don't really think that bullet point belongs in the list.

Honestly it would probably be better to just combine the last two bullet points into a single statement/paragraph as you suggested.

It should just be something like

> `dvc exp branch` is useful if you would like to an experiment persistent as a Git branch, and for preserving checkpoints as independent commits within that newly created Git branch.

content/docs/command-reference/exp/run.md
dberenbaum: Take a look at https://github.com/iterative/dvc-checkpoints-mnist/pull/2. That should help illustrate what happens, which is the same thing that would happen if someone changed parameters or code before running `exp run` again. 

`(it will restart from scratch)` will be misleading here I think.

content/docs/command-reference/exp/apply.md
```diff
@@ -14,20 +14,21 @@ positional arguments:
...
 
-> This is similar to `dvc checkout`, but for `dvc experiments`.
+This command is typically used after finding a target `experiment` with
+`dvc exp show` or `dvc exp diff`, and before committing it to Git (making it
+[persistent](/doc/user-guide/experiment-management#persistent-experiments)).
 
-This is typically used after using `dvc exp show` or `dvc exp diff` to find the
-best experiment, and before committing to Git in order to make it
-[persistent](/doc/user-guide/experiment-management#persistent-experiments).
+Note that any [checkpoints](/doc/command-reference/exp/run#checkpoints) found in
```
dberenbaum, line 26:
```suggestion
Note that any intermediate [checkpoints](/doc/command-reference/exp/run#checkpoints) found in
```
Wondering if `intermediate` better conveys the message that the whole history is "squashed" into a single merge rather than disregarded. `apply` will still reflect the experiment state at the time of the final checkpoint. Feel free to ignore if you think it's unhelpful.

content/docs/command-reference/exp/branch.md
dberenbaum: IMO the first bullet can be put into the above paragraph and the last two bullets removed. I think the `apply` comparison is a bit too much of a stretch, and the independent checkpoint commits are mentioned already above. There's no reason the use cases need to be exhaustive. Giving the primary usage seems natural to me.

jorgeorpinel: OK I made it into a regular p with only the first item (rephrased slightly).

content/docs/command-reference/exp/apply.md
jorgeorpinel: I agree it would be easier to read though...

content/docs/api-reference/make_checkpoint.md
```diff
@@ -67,19 +67,23 @@ import os
...
+        with open("int.txt", "w") as fd:
+            fd.write(f"{i_}")
 
-    if i_ % 100 == 0:
-        make_checkpoint()
+        if i_ % 100 == 0:
+            make_checkpoint()
+
+    except KeyboardInterrupt:
+    exit()
```
dberenbaum, line 86:
```suggestion
        exit()
```

content/docs/command-reference/exp/apply.md
```diff
@@ -14,20 +14,21 @@ positional arguments:
...
+the `experiment` are **not** preserved when applying and committing it.
```
dberenbaum, line 27: Alternate suggestion (same idea as previous comment adding `intermediate`):
```suggestion
the `experiment` are **not** preserved in separate commits when applying.
```
Again, not a blocker, decide for yourself.

dberenbaum (approved):
Left very minor suggestions to consider, but looks ready to merge.

content/docs/command-reference/exp/run.md
jorgeorpinel: Hm OK I see. I had to intro the concept of "wrap-up checkpoint" to explain this. See 1ceeebef

content/docs/command-reference/exp/apply.md
jorgeorpinel: Yeah it's more correct, but not really a concept we have explained much. I don't know if it's intuitive enough. I went for "the history of checkpoints" for now.

PR #2438, opened 2021-05-03
cases: update Versioning figures

jorgeorpinel:
Thanks for the figures @arcticbear


jorgeorpinel:
I tried using SVGs versions but they don't get centered.

shcheklein (approved):

PR #2441, opened 2021-05-04
ref: update S3 and S3-compatible details in `remote add/modify`

jorgeorpinel:
Followup to #2415


content/docs/command-reference/remote/add.md
```diff
@@ -115,21 +115,7 @@ $ dvc remote add -d myremote s3://mybucket/path
 For more on the supported env vars, please see the
 [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variable-configuration)
 
-The following API methods are performed by the `boto3` library (used by DVC):
```
shcheklein, line 118: hmm .. why do we remove this?

jorgeorpinel: It's a bit too detailed (most deep details are in `remote modify`. I thought that's what you were referring to in https://github.com/iterative/dvc.org/pull/2415#pullrequestreview-650598131 haha. OK, no strong opinion here, will reinstate it ⌛

shcheklein: these details are general, about this type of the remote, they are not about auth - that's what we should keep in mind. Eventually this kind of a spec will go into the UG (similar to the GDrive).

But for now this general remote content probably belong better here, not in the modify. Some details and language can be updated (or keep for now as is if it's tricky)

content/docs/command-reference/remote/add.md
```diff
@@ -104,7 +104,7 @@ By default, DVC authenticates using your AWS CLI
 (if set). This uses the default AWS credentials file. To use a custom
 authentication method, use the parameters described in `dvc remote modify`.
 
-S3 remotes can also be configured entirely via environment variables, e.g.:
+S3 remotes can also be configured via environment variables, for example:
 
 ```dvc
 $ export AWS_ACCESS_KEY_ID='mykey'
```
jorgeorpinel, line 110: > Moving from https://github.com/iterative/dvc.org/pull/2415#discussion_r626199728, from @shcheklein:

A think a lot of remotes can be configured with env vars.

All this additional config should be part of modify to my mind. Things like

```
export ..
export
dvc remote add
```

do not make much sense overall. 

```
export ...
export ...
```

are important when you run `dvc push`/`dvc pull`. It's a way to provide values to auth flow used in DVC when it tries to access. Other ways - default config files (`.aws/credentials`) or DVC config (the one you modify with `dvc remote modify`)

jorgeorpinel: ⌛

content/docs/command-reference/remote/add.md
jorgeorpinel: Got it, makes sense.

I compacted those 2 bullet lists but kept them in both refs for now (dfdaf404), because `remote modify` has some params that refer to the API functions used.

content/docs/command-reference/remote/add.md
jorgeorpinel: > a lot of remotes can be configured with env vars

Good Q. Opened https://github.com/iterative/dvc.org/issues/2445

> export & dvc remote add ... do not make much sense

Wait yeah... `remote add` doesn't need any creds. OK, removed all env var talk from that ref.

jorgeorpinel: > a lot of remotes can be configured with env vars

Good Q @shcheklein . Opened https://github.com/iterative/dvc.org/issues/2445

> export & dvc remote add ... do not make much sense

Wait yeah... `remote add` doesn't need any creds. OK, removed all env var talk from that ref.

> are important when you run dvc push/dvc pull

True. Updated `remote modify`. PTAL

content/docs/command-reference/remote/add.md
```diff
@@ -104,32 +104,12 @@ By default, DVC authenticates using your AWS CLI
 (if set). This uses the default AWS credentials file. To use a custom
 authentication method, use the parameters described in `dvc remote modify`.
 
-S3 remotes can also be configured entirely via environment variables, e.g.:
+The following API methods are performed by DVC: `list_objects_v2` or
+`list_objects`, `head_object`, `download_file`, `upload_file`, `delete_object`,
+`copy`. So make sure you have the following permissions enabled:
```
shcheklein, line 109: let's start with "Make sure " (and then mention API calls) - it's more practical to my mind.

shcheklein (approved):

content/docs/command-reference/remote/add.md
jorgeorpinel: Done!

Issue #1509, opened 2020-06-29
yaml incorrectly rendered


skshetry:
Expand the section `Expand to see what happens under the hood.
` on https://dvc.org/doc/start/data-pipelines#pipeline-stages. The `params.yaml` is incorrectly rendered (indentation that is).

https://github.com/iterative/dvc.org/blob/befb19b5debaf47137d3507e1be919640211cc78/content/docs/start/data-pipelines.md#L76-L80

It seems to be correctly written. Maybe something wrong with rendering it?

shcheklein:
Fixed, had to un-indent the whole block for this to work. Weird. Will keep this as a doc engine bug.

If `yaml` code block is indented to be under a bullet item rendering breaks.

jorgeorpinel:
prettier always re-indents our YAML samples.

jorgeorpinel:
ah but not like this actually. I mean the contents, not the whole block.

jorgeorpinel:
That code block was correctly indented though, it's part of a bullet. Now it's uneven (compare to the next block):

![image](https://user-images.githubusercontent.com/1477535/86285297-f86d4800-bba9-11ea-9850-1219bdd71bcf.png)

shcheklein:
This is a temporary workaround. This engine bug.

jorgeorpinel:
I changed the workaround in #2458 so the block is aligned.

jorgeorpinel:
> Expand the section Expand to see what happens under the hood. on https://dvc.org/doc/start/data-pipelines#pipeline-stages.

BTW this also happens in the `git` block under that YAML block too:

![image](https://user-images.githubusercontent.com/1477535/117699538-964a7f80-b18a-11eb-9b4d-856af14355c7.png)

The code is correct:

````md
  ```git
   .
   ├── data
   │   ├── data.xml
   │   ├── data.xml.dvc
  +│   └── prepared
  +│       ├── test.tsv
  +│       └── train.tsv
...
````

So this probably applies to many kinds of code blocks that are indented themselves.

jorgeorpinel:
Cc @rogermparent @julieg18 here's a docs-engine bug that I think would be the most useful to target next, whenever you guys are planning to look towards dvc.org.

rogermparent:
I believe this has to do with code blocks nested inside custom components not being handled correctly. All the instances I've seen at the top level are fine.

jorgeorpinel:
Yep

[...]

PR #2458, opened 2021-05-10
start: code block indentation workarounds

jorgeorpinel:
rel #1509

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -90,11 +90,11 @@ The command options used above mean the following:
   but the idea is that the stage can depend on field values from a parameters
   file (`params.yaml` by default):
 
-```yaml
-prepare:
-  split: 0.20
-  seed: 20170428
-```
+  ```yaml
+  prepare:
+      split: 0.20
+      seed: 20170428
+  ```
 
 - `-d src/prepare.py` and `-d data/data.xml` mean that the stage depends on
   these files to work. Notice that the source code itself is marked as a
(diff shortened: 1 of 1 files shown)


content/docs/start/data-pipelines.md
```diff
@@ -90,11 +90,11 @@ The command options used above mean the following:
...
-```yaml
-prepare:
-  split: 0.20
-  seed: 20170428
-```
+  ```yaml
+  prepare:
+      split: 0.20
+      seed: 20170428
+  ```
```
jorgeorpinel, line 97: Had to force commit this since Pretty doesn't like it. Formatting check will prob. fail. But this looks correctly for now:

image

content/docs/start/data-pipelines.md
```diff
@@ -105,18 +105,18 @@ prepare:
...
-   .
-   ├── data
-   │   ├── data.xml
-   │   ├── data.xml.dvc
+     .
+     ├── data
+     │   ├── data.xml
+     │   ├── data.xml.dvc
   +│   └── prepared
   +│       ├── test.tsv
```
jorgeorpinel, line 113: Here the workaround so it looks correct really makes the code looks broken. And the formatter doesn't catch this one. So I wonder if it's best to close this PR and focus on fixing #1509 instead.

shcheklein: to be honest looks like a minor problem, I would wait on the engine to fix this, for now broken it's fine to unindent code blocks

PR #2471, opened 2021-05-12
cases: enable file links for shared dev config (external cache)

### LINKED ISSUE #429: cases: update Shared Dev Server with an optimization link

jorgeorpinel:
Closes #429

See https://dvc-org-cases-shared-de-2cu4a8.herokuapp.com/doc/use-cases/shared-development-server#configure-the-external-shared-cache

### DIFF
--- a/content/docs/use-cases/shared-development-server.md
+++ b/content/docs/use-cases/shared-development-server.md
@@ -55,18 +55,20 @@ your <abbr>project</abbr>:
 $ dvc cache dir /home/shared/dvc-cache
 ```
 
-Let's do a few important configurations:
+External cache configuration:
 
 ```dvc
 $ dvc config cache.shared group
 $ dvc config cache.type "reflink,symlink,hardlink,copy"
-$ dvc config cache.protected true
 ```
 
-First we tells DVC to set group permissions on new cache files. Then we enable
-soft and hard links (you limit to one or the other) to avoid copying from the
-external cache to the <abbr>workspace</abbr>. Finally we must "protect" the
-symlinked data in our project to avoid corrupting the cache.
+Above, we first tell DVC to set group permissions on new cache files. Then we
+enable soft and hard links (you limit to one or the other) to avoid having
+copies from the external cache to the <abbr>workspace</abbr>.
+
+⚠️ Note that enabling symlinks/hardlinks causes DVC to protect the linked data
+in the workspace, since editing them in-place would corrupt the cache. See also
+`dvc unprotect`.
 
 > See `dvc cache dir` and `dvc config cache` for more information.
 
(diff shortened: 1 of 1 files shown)


content/docs/use-cases/shared-development-server.md
```diff
@@ -55,13 +55,21 @@ your <abbr>project</abbr>:
 $ dvc cache dir /home/shared/dvc-cache
 ```
 
-And tell DVC to set group permissions on newly created or downloaded cache
-files:
+External cache configuration:
 
 ```dvc
 $ dvc config cache.shared group
+$ dvc config cache.type "reflink,symlink,hardlink,copy"
```
shcheklein, line 62: would be good to include here links - to the links guide and to the relevant section in the dvc config

also, not clear if `reflink` is a very meaningful option here, tbh ... it would work only in a very specific setup - may be we can mention that, but I would not expect that it would become commonly used tbh

jorgeorpinel: Linked to file links guide and moved the link to `config cache` closer to this p. Removed `reflink`

jorgeorpinel: What about `copy`? It's a bit unclear why we leave it in if we want to prevent copying.

efiop (approved):

content/docs/use-cases/shared-development-server.md
shcheklein: yep, sounds good to me

content/docs/use-cases/shared-development-server.md
```diff
@@ -55,14 +55,23 @@ your <abbr>project</abbr>:
...
 
 ```dvc
 $ dvc config cache.shared group
+$ dvc config cache.type "symlink,hardlink,copy"
 ```
 
-> See `dvc cache dir` and `dvc config cache` for more information.
+Above, we first tell DVC to set group permissions on new cache files. Then we
+enable soft and hard [file links](/doc/user-guide/large-dataset-optimization)
+(you can pick only one or the other) to avoid having copies from the external
```
shcheklein, line 67: `you can pick only one or the other` sounds like you should pick in the command itself

btw, hardlinks most likely won't work also :)

so, it's only about symlinks usually

shcheklein (approved):

content/docs/use-cases/shared-development-server.md
jorgeorpinel: Makes sense. Left symlinks only and improved the note that links to the other types.

PR #2478, opened 2021-05-15
feat: focus search box on forward slash keydown

### LINKED ISSUE #722: ui: forward slash / focus the search bar
It is sort of common to focus the search bar when hitting forward slash (`/`), making it easier to navigate a webpage in a keyboard-only way.

Not sure how to implement this but a quick search threw this one: https://github.com/greena13/react-hotkeys

evan-cohen:
Adds a reusable `focusElementWithHotkey` function to listen for the `keydown` event on a particular key and respond by setting focus on the specified element. Implementing this with the "/" key and "#doc-search" element will fix #722.

Note to reviewer: This PR adds the `useEffect` for initializing this event listener to the main Layout component, but I am open to feedback on whether there is an alternate location that would be more appropriate.

### DIFF
--- a/src/components/Documentation/Layout/index.tsx
+++ b/src/components/Documentation/Layout/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 import cn from 'classnames'
 
 import MainLayout, { LayoutComponent, LayoutModifiers } from '../../MainLayout'
@@ -7,6 +7,7 @@ import HamburgerIcon from '../../HamburgerIcon'
 import SearchForm from './SearchForm'
 import SidebarMenu from './SidebarMenu'
 import { matchMedia } from '../../../utils/front/breakpoints'
+import { focusElementWithHotkey } from '../../../utils/front/focusElementWithHotkey'
 
 import styles from './styles.module.css'
 
@@ -17,6 +18,10 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
   } = restProps
 
   const toggleMenu = useCallback(() => setIsMenuOpen(!isMenuOpen), [isMenuOpen])
+
+  useEffect(() => {
+    focusElementWithHotkey('#doc-search', '/')
+  }, [])
 
   return (
     <MainLayout
--- a/src/components/Documentation/Layout/index.tsx
+++ b/src/components/Documentation/Layout/index.tsx
@@ -20,7 +20,8 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
   const toggleMenu = useCallback(() => setIsMenuOpen(!isMenuOpen), [isMenuOpen])
 
   useEffect(() => {
-    focusElementWithHotkey('#doc-search', '/')
+    const closeEventListener = focusElementWithHotkey('#doc-search', '/')
+    return closeEventListener
   }, [])
 
   return (
--- a/src/components/Documentation/Layout/index.tsx
+++ b/src/components/Documentation/Layout/index.tsx
@@ -20,6 +20,9 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
   const toggleMenu = useCallback(() => setIsMenuOpen(!isMenuOpen), [isMenuOpen])
 
   useEffect(() => {
+    if (matchMedia('--xs-scr')) {
+      return
+    }
     const closeEventListener = focusElementWithHotkey('#doc-search', '/')
     return closeEventListener
   }, [])
--- a/src/components/Documentation/Layout/index.tsx
+++ b/src/components/Documentation/Layout/index.tsx
@@ -10,6 +10,7 @@ import { matchMedia } from '../../../utils/front/breakpoints'
 import { focusElementWithHotkey } from '../../../utils/front/focusElementWithHotkey'
 
 import styles from './styles.module.css'
+import { useWindowSize } from 'react-use'
 
 const Layout: LayoutComponent = ({ children, ...restProps }) => {
   const [isMenuOpen, setIsMenuOpen] = useState(false)
@@ -19,13 +20,15 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
 
   const toggleMenu = useCallback(() => setIsMenuOpen(!isMenuOpen), [isMenuOpen])
 
+  const windowSize = useWindowSize()
+
   useEffect(() => {
     if (matchMedia('--xs-scr')) {
       return
     }
     const closeEventListener = focusElementWithHotkey('#doc-search', '/')
     return closeEventListener
-  }, [])
+  }, [windowSize])
 
   return (
     <MainLayout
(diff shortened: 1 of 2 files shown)


jorgeorpinel:
Thanks @evan-cohen ! I tried this in https://dvc-org-add-search-box--yhlul0.herokuapp.com/doc and it works.

src/utils/front/focusElementWithHotkey.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Utility function to focus on an element by pressing a specified hotkey.
+ * @param target CSS selector for the element to be focused
+ * @param hotkey The key that should trigger the target element to be focused
+ */
+export const focusElementWithHotkey = (target: string, hotkey: string) => {
```
julieg18, line 6: I think these functions need return types([eslint-typescript docs](https://github.com/typescript-eslint/typescript-eslint/blob/v2.34.0/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)). Example: `focusElementWithHotkey = (target: string, hotkey: string): void => {`

src/components/Documentation/Layout/index.tsx
```diff
@@ -18,6 +19,10 @@ const Layout: LayoutComponent = ({ children, ...restProps }) => {
 
   const toggleMenu = useCallback(() => setIsMenuOpen(!isMenuOpen), [isMenuOpen])
 
+  useEffect(() => {
+    focusElementWithHotkey('#doc-search', '/')
```
julieg18, line 23: We need to be able to close this event listener when the `Documentation/Layout` component unmounts. Otherwise this function will keep listening for keystrokes even if the user moves to a different page. One way I can think of to accomplish this is to have`focusElementWithHotkey` return a function that closes the event listener, so the `useEffect` hook can return it.
Example:
```
// in Layout
  useEffect(() => {
    const closeEventListener = focusElementWithHotkey('#doc-search', '/')
    return closeEventListener
  }, [])

...

// in focusElementWithHotkey

export const focusElementWithHotkey = (
  target: string,
  hotkey: string
): (() => void) => {
  const func = (e: KeyboardEvent): void => {
    // place code for keyboard event here
  }
  document.addEventListener('keydown', func)
  return (): void => document.removeEventListener('keydown', func)
}
```

julieg18 (changes requested):
Nice job so far! The main issue I found was about the event listener not closing when `Documentation/Layout` unmounts.

src/utils/front/focusElementWithHotkey.ts
```diff
@@ -0,0 +1,15 @@
...
+  document.addEventListener('keydown', $event => {
+    if ($event.key !== hotkey || document.activeElement?.matches(target)) {
+      return
+    }
+    const targetElement = document.querySelector(target) as HTMLElement
+    targetElement?.focus()
+    $event.preventDefault()
+  })
```
rogermparent, line 14: While `$event` won't hurt anything, all or at least most of our existing event listeners use `e` for the event's name.

src/utils/front/focusElementWithHotkey.ts
evan-cohen: All set! `npm run lint-ts` no longer returns any warnings or errors.

src/components/Documentation/Layout/index.tsx
evan-cohen: Thanks for the feedback! I've addressed this issue in 44a3221

src/utils/front/focusElementWithHotkey.ts
evan-cohen: Sure thing. I've resolved this in dc3f8b5. Thanks!

rogermparent (changes requested):
I noticed an issue on the deploy preview where using this feature on mobile breakpoints would have broken behavior: search is focused while off-screen, and typing starts the search with the results partially offscreen:

example

This could be fixed in a few ways, two that immediately come to me are:

- Disable the feature when the screen is within the lowest breakpoint and the sidebar is hidden.
- Expand the sidebar when it's focused off-screen.

I think the latter, while it needs a little extra work, adds enough value to this feature to be worth it. While most phones won't use it, this can still be useful for users that have the docs open in a thin window on a desktop- if we make sure to hide the sidebar again when a search is selected, it would enable a quick navigation within small windows of DVC's docs- I could even see myself using it!

jorgeorpinel:
> this can still be useful for users that have the docs open in a thin window on a desktop

Good find there @rogermparent. My 2c on this is that it's probably an edge case and it would be easier to disable the feature on thin screens for now to get this contrib merged.

We could consider implementing that as a separate feature (create an issue or if @evan-cohen would like to work on that too it can be a separate PR). WDYT?

shcheklein:
@rogermparent @jorgeorpinel please take some action here :)

shcheklein:
@evan-cohen could you please disable it for now on mobile screens so that we can move forward?

@julieg18 any other concerns or we can approve this?

evan-cohen:
> @evan-cohen could you please disable it for now on mobile screens so that we can move forward?

Sure, I have a few potential fixes for this. One is to simply add an early return at the same breakpoint where the mobile menu is displayed:
```jsx
useEffect(() => {
  if (matchMedia('--xs-scr')) {
    return
  }
  const closeEventListener = focusElementWithHotkey('#doc-search', '/')
  return closeEventListener
}, [])
```

However, this exposes another edge case when a user resizes the window because the `useEffect` is currently only running on the initial component mount. I can account for this as well with a hook to track window size ([example](https://usehooks.com/useWindowSize/)) and pass that as a `useEffect` dependency.  Thoughts?

shcheklein:
@evan-cohen approach sounds reasonable to me, but may be there are other simpler ways? let's wait for @rogermparent and @julieg18 to chime in.

rogermparent:
@evan-cohen @shcheklein I think that'd be a fine temporary solution to squash the bug before merge, though it'll likely be removed once we add proper mobile handling for this feature.

shcheklein:
Thanks @rogermparent . Sounds good then. Let's do a temp workaround, create a ticket to improve this later in the followup PR.

julieg18:
> @evan-cohen @shcheklein I think that'd be a fine temporary solution to squash the bug before merge, though it'll likely be removed once we add proper mobile handling for this feature.

Agreed! @evan-cohen, just in case you didn't know, dvc.org has the [`react-use`](https://github.com/streamich/react-use) package installed which has the `useWindowSize` hook already made and ready to use!

shcheklein:
hey, folks, what is the status of this?

rogermparent:
> hey, folks, what is the status of this?

Seems like from [the latest commit](https://github.com/iterative/dvc.org/pull/2478/commits/49ee94b98f8c33abd39c5f43b36832e3be0a9a16) that this has a quick fix for the mobile issue that can suffer a bit if the user resizes the window while the search menu is open. I fired up the deploy preview so we can check it out, but it seems like we can either:

1. Accept this minor caveat and merge, then fix it later
2. Try to implement a more robust solution that listens for the window to go into mobile midth and dismiss the dialog if it happens. Or just styling that can handle mobile widths.

rogermparent:
Now that I put my hands on it, I think we need a better fix before allowing this in. It's probably best to toss in `react-use`'s `useWindowSize` over the current solution and go from there.

rogermparent:
Looks like the main resizing issue is solved! Some focus weirdness can happen if the user resizes while the search box is focused, but that's way more minor than the previous issue. I'd say this can ship after a master merge!

rogermparent (approved):

PR #2481, opened 2021-05-16
cmd ref: rm 'Implies --no-exec' in stage add

### LINKED ISSUE #2480: dvc stage add: redundant "Implies --no-exec" for -c flag
Description of `-c` option in `dvc stage add`:

> -c <path>, --checkpoints <path> - the same as -o but also marks the output as a checkpoint. **Implies --no-exec.** This makes the stage incompatible with dvc repro.

Since `dvc stage add` doesn't have `--no-exec` option I suppose "Implies --no-exec" should be removed.

nik123:
Fixes #2480


jorgeorpinel (approved):
Ah yes good catch! This was mistakenly copied over from https://dvc.org/doc/command-reference/run#options

jorgeorpinel:
Thanks @nik123

Issue #2496, opened 2021-05-22
start: independent trails


iesahin:
UDPATE: Jump to https://github.com/iterative/dvc.org/issues/2496#issuecomment-1255510563.

## Status

* We have 5 repositories: [example-get-started][egs], [dvc-checkpoints-mnist][dcm], [get-started-experiments][gse], [get-started-checkpoints][gsc], and [get-started-pipelines][gsp].

[egs]: https://github.com/iterative/example-get-started
[dcm]: https://github.com/iterative/dvc-checkpoints-mnist
[gse]: https://github.com/iterative/get-started-experiments
[gsc]: https://github.com/iterative/get-started-checkpoints
[gsp]: https://github.com/iterative/get-started-pipelines

* We have a project that generates most of these. 

The docs use [example-get-started][egs] for:

* [Data and Model Versioning](https://dvc.org/doc/start/data-and-model-versioning)
* [Data and Model Access](https://dvc.org/doc/start/data-and-model-access)
* [Data Pipelines](https://dvc.org/doc/start/data-pipelines)
* [Metrics, Parameters, Plots](https://dvc.org/doc/start/metrics-parameters-plots)
* [Experiments](https://dvc.org/doc/start/experiments)

[Checkpoints guide][cpg] uses [dvc-checkpoints-mnist][dcm]

[cpg]: https://dvc.org/doc/user-guide/experiment-management/checkpoints

## Goals

* We want to improve the content with _minimum changes_ to the existing documents. Adding more content to the already available material is desired. 

* We want to have a common/similar project for the tutorials. A _single_ showcase project to contain all DVC features seems a bit artificial. A set of _similar_ projects may be a better tradeoff for maintenance and usability. 

* DVC has different use cases for different people and we want to emphasize these: 
** Data Versioning
** Data Access
** Sharing Models
** Presenting Models with Metrics and Plots
** Experiment Management and Sharing
** Checkpoints (which may be under "experiment management".)

* There should be more than one entry points for the tutorials, e.g., experiment management should be a first-class citizen. 

## Discussion and Research Points

* Current documentation is mostly pipelines-based. Almost all features revolve around `dvc.yaml` and the pipelines. How can we present DVC as an _experiment management system_ without first telling about the pipelines?

[...]

Issue #2474, opened 2021-05-13
start: improvements


flippedcoder:
- [x] Remove the Sharing experiments section from Get Started > Experiments

We should remove the [Sharing experiments section](https://dvc.org/doc/start/experiments#sharing-experiments) since it adds more complexity to the Get Started tutorial than there needs to be since users have to set up a Git remote as well as a DVC remote. We don't go over this at any other point in the tutorial.

If a user clones the get started repo, it works fine, but if they are following along there are some roadblocks they might run into with getting things set up. For example, I ran into a weird credentials issue with my Git remote and ended up working with @dberenbaum to do a crazy workaround.

Overall, it clouds up the purpose of this section which is running experiments.

- [x] Update the `git add` command for Get Started > Metrics, Parameters, and Plots

In the [Comparing iterations section](https://dvc.org/doc/start/metrics-parameters-plots#collecting-metrics), users can get empty Old and Change values when they run `dvc metrics diff` because the `dvc.lock` and `dvc.yaml` are missing from the `git add`.

We need to update the command to be `git add dvc.lock dvc.yaml scores.json prc.json roc.json`.

dberenbaum:
@flippedcoder Do you mind separating these whenever you have a chance? We try to be pretty strict about keeping issues logically separate so that one part of an issue doesn't block the rest of the issue from being resolved.

jorgeorpinel:
Checkboxes could also work, or be a good first step.

jorgeorpinel:
May also be good to see if we can consolidate with #1943.

jorgeorpinel:
We should also consolidate this one with https://github.com/iterative/dvc.org/issues/2496 cc @iesahin

jorgeorpinel:
@iesahin if you could also double check which parts of this are still/most relevant and consolidate into https://github.com/iterative/dvc.org/issues/2496 that would be great.

PR #2497, opened 2021-05-22
start: use `get-started-experiments` for GS:Experiments tutorial

### LINKED ISSUE #2479: start: A new GS:Experiments document with `get-started-experiments`
Katacoda [GS:Experiments scenario](https://katacoda.com/dvc/courses/get-started/experiments) employs [`dvc-get-started`](https://github.com/iterative/dvc-get-started) to introduce experiment features already. The scenario uses [`dvcorg/doc-katacoda:experiments`](https://hub.docker.com/repository/docker/dvcorg/doc-katacoda) container. 

Let's upgrade [GS:Experiments documentation](https://dvc.org/doc/start/experiments) to have a parallel content with the scenario. 

This will also allow to run [`markdown-code-runner`](https://github.com/iterative/markdown-code-runner) on the page to get _automated tests_ for the commands in the documentation content, for the first time. 

Related #2474 
Related iterative/example-repos-dev#38
Related iterative/example-repos-dev#44

@flippedcoder @jorgeorpinel @shcheklein @dberenbaum

iesahin:
This one is the branched version of #2491 

Closes #2479 

It doesn't contain any content changes but updates the commands with the new project in github.com/iterative/get-started-experiments .

The tables are currently `code blocks` but they will be updated with the screenshots instead.


shcheklein:
There are some conflicts, also could we now close #2491 ?

shcheklein:
@jorgeorpinel please note, that we probably should be reviewing this - it's already a branched version.

iesahin:
I think it may look better to use a terminal with the identical background-color when taking the screenshots. Is there a predefined color scheme for code blocks?

Screen Shot 2021-05-28 at 11 15 03

iesahin:
I think the font is SF Mono and the color scheme is similar to Solarized Dark. This is what I get using these:

Screen Shot 2021-05-28 at 11 31 06

iesahin:
And this is after modifying the background to #40354d and foreground to #ccc

Screen Shot 2021-05-28 at 11 36 52

iesahin:
Replaced all screenshots to have a similar background and font.

Screen Shot 2021-05-28 at 13 24 48

iesahin:
I installed SF Mono Bold as well and it looks iTerm uses the bold font when it finds one. 

Bold font and regular font, please let me know if it's better to replace. (It takes some time to do the experiments.)

Screen Shot 2021-05-28 at 13 30 36
Screen Shot 2021-05-28 at 13 31 40

shcheklein:
@iesahin Looks nice! from this screenshots I like the one that is not bold. No strong opinion here :)

content/docs/start/experiments.md
```diff
@@ -18,17 +18,87 @@ once they're no longer needed.
...
+### 🆕 Click here for the instructions to install the project
+
+These commands are run in
+[`get-started-experiments`](https://github.com/iterative/get-started-experiments)
+project. You can run these commands after cloning the repository and install the
+requirements.
+
+Please clone and create a virtual environment:
+
+```console
```
shcheklein, line 38: console -> dvc

start each line with $ (except output) to make highlighter work

content/docs/start/experiments.md
```diff
@@ -18,17 +18,87 @@ once they're no longer needed.
...
-experiment:
+experiment. We'll use it to search for parameters to increase the classification
+performance of our model using
+[Fashion-MNIST](https://github.com/iterative/get-started-experiments) dataset.
+
+In order to run a baseline experiment with the default parameters defined in
+`params.yaml`:
+
+```dvc
+dvc exp run
```
shcheklein, line 62: start with $

content/docs/start/experiments.md
```diff
@@ -18,17 +18,87 @@ once they're no longer needed.
...
+```
+
+![dvc exp run result screenshot](/img/doc/start/exp-ss-90252.png)
```
shcheklein, line 65: here screenshot is probably not needed?

content/docs/start/experiments.md
```diff
@@ -54,92 +125,84 @@ reflected in the experiment run.
...
-└───────────────┴──────────┴─────────┴────────────┴─────────────────┘
+               --include-params model.cnn.conv_units \
+               --include-metrics categorical_accuracy
 ```
 
-Each experiment is given an arbitrary name by default (although we can specify
-one with `dvc exp run -n`.) We can see that `exp-98a96` performed best among
-both of our metrics, with 100 estimators and a minimum of 64 samples to split a
-node.
+![dvc exp show screenshot](/img/doc/start/exp-ss-36794.png)
```
shcheklein, line 188: Let's create a ticket to explore a way to use syntax highlighter to do markup for these tables

content/docs/start/experiments.md
```diff
@@ -192,29 +271,26 @@ Storage, HTTP, HDFS, etc.). The Git remote is often a central Git server
 
 </details>
 
-Experiments that have not been made persistent will not be stored or shared
```
shcheklein, line 195: ^^^^ a few lines about - fix code block, lines should start with $

content/docs/start/experiments.md
iesahin: I thought `dvc` is specifically for DVC commands. Thank you.

content/docs/start/experiments.md
iesahin: I had thought something about putting the text instead but decided to put the screenshot but the reason seems not something worthwhile to remember. 😅  I'm removing it now.

iesahin:
> @iesahin Looks nice! from this screenshots I like the one that is not bold. No strong opinion here :)

I had to locate away from my Mac on the weekend. I'm taking the screenshots with Gnome Terminal but it doesn't look as good. 

exp-ss-63714

This is probably about horizontal spacing config in iTerm. I used 0.95 or 0.90, but Gnome Terminal accepts only >1. Let's keep the current screenshots for the time being and I'd update theem in a later iteration. 
 

I think it's better

content/docs/start/experiments.md
iesahin: #2516

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+using `dvc repro` we need to update `params.yaml` manually, run the pipeline, if
+the results are worth it commit them to DVC and Git. `dvc exp` automates this
+process through its subcommands.
+
+Let's see some metrics produced by this baseline experiment:
 
 ```dvc
-$ dvc exp run --set-param featurize.max_features=3000
+$ dvc exp show --include-metrics categorical_accuracy \
+               --include-params  model.name,model.cnn.conv_units
```
dberenbaum, line 79: Should we have a version of the project that has only these parameters and metrics instead of having to pass all these options to `dvc exp show`? It's a lot of parameters for the first time we introduce this command.

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
 ```
 
+![dvc exp show result screenshot](/img/doc/start/exp-ss-63714.png)
+
+Note that the experiment results are identical with the values checked-out from
+Git. By default each experiment is given a name automatically. We can set the
+name by the `--name/-n` argument.
+
+Let's change the number of units in CNN and make another experiment:
```
dberenbaum, line 88:
```suggestion
Let's change the number of units in the CNN and make another experiment:
```

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+Stage 'prepare' didn't change, skipping
+Stage 'preprocess' didn't change, skipping
+Running stage 'train':
+> python3 src/train.py
+...
+```
+
+`--set-param/-S` argument of `dvc exp run` is used to set parameters in
+`params.yaml`. DVC runs _only_ the stages that depend on these parameter values.
+Parameter dependencies are defined via `dvc stage add` command and stored in
```
dberenbaum, line 102: Maybe reference metrics-parameters-plots here?

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+`dvc.yaml`.
+
 <details>
```
dberenbaum, line 105: This whole details section duplicates what was added above.

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
-train:
-  seed: 20170428
-  n_est: 50
-  min_split: 2
-```
-
-Let's setup experiments with different hyperparameters. We can use the `--queue`
-flag to define all the combinations we want to try without executing anything
-(yet):
+We add experiments to the queue using `--queue` option of `dvc exp run`. Here,
```
dberenbaum, line 153:
```suggestion
We add experiments to the queue using the `--queue` option of `dvc exp run`. Here,
```

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
+we also set the names of experiments to observe the results clearly:
```
dberenbaum, line 154: I think the second sentence is self-evident and could be skipped to save space.

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
+To apply the results of an experiment to your workspace run:
+
+	dvc exp apply <exp>
+
+To promote an experiment to a Git branch run:
+
+	dvc exp branch <exp>
 ```
 
+The command will run all queued experiments in a random order.
```
dberenbaum, line 178: This again seems unnecessary since we already state above that this command will run all experiments. Is the ordering so confusing that it needs to be mentioned here? I would think many people wouldn't notice or care, and people familiar with parallel processing would expect random ordering. It's not much to keep if you think it will avoid confusion, but it's all about trying to keep this doc concise.

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
 
 ## Persisting experiments
 
-Now that we know the best parameters, let's keep that experiment and ignore the
-rest.
+The metrics to optimize are usually domain or problem dependent. In these
+experiments we are trying to optimize for `categorical_accuracy`. Let's keep
+experiment with the highest `categorical_accuracy` score and ignore the rest.
 
 `dvc exp apply` rolls back the <abbr>workspace<abbr> to the specified
```
dberenbaum, line 203: In retrospect, I'd prefer to introduce `dvc exp branch` here instead of `dvc exp apply`, but that's best left for a separate issue. cc @jorgeorpinel

content/docs/start/experiments.md
```diff
@@ -163,8 +241,12 @@ reproducing the result without `dvc exp run`. Let's make it persistent in our
...
-$ git commit -a -m "Preserve best random forest experiment"
+$ git add dvc.lock params.yaml logs.csv metrics.json
+$ git commit -m "Preserve the experiment with 128 Conv Units"
+
+[main 84d1c8f] Preserve the experiment with 128 Conv Units
+ 4 files changed, 21 insertions(+), 21 deletions(-)
+ rewrite logs.csv (91%)
 ```
 
 ## Sharing experiments
```
dberenbaum, line 252: This whole section needs to be revised or removed, but again this is a separate issue (see https://github.com/iterative/dvc.org/issues/2474).

content/docs/start/experiments.md
```diff
@@ -226,55 +305,48 @@ Let's take another look at the experiments table:
...
+> objects associated to them. To clean up the cache, use `dvc gc`.
+
+## Going Further
+
+You can continue to experiment with
+[the project](https://github.com/iterative/get-started-experiments). There are
+many parameters available to modify, e.g., `noise` or `units` in Dense layer of
+the network. Please see the `README.md` file of the project for these. Don't
+forget to [notify us](https://dvc.org/chat) if you happen to find good
+parameters.
```
dberenbaum, line 352: Is this to encourage community engagement?

content/docs/start/experiments.md
```diff
@@ -226,55 +305,48 @@ Let's take another look at the experiments table:
...
+![dvc exp show screenshot](/img/doc/start/exp-ss-32408.png)
+
+> `dvc exp gc` only removes references to the experiments, not the cached
+> objects associated to them. To clean up the cache, use `dvc gc`.
+
+## Going Further
+
+You can continue to experiment with
+[the project](https://github.com/iterative/get-started-experiments). There are
+many parameters available to modify, e.g., `noise` or `units` in Dense layer of
```
dberenbaum, line 349:
```suggestion
many parameters available to modify, e.g., `noise` or `units` in the Dense layer of
```

content/docs/start/experiments.md
dberenbaum: Just curious.

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
 
 ## Running experiments
```
dberenbaum, line 19: This section seems too long. Can you separate setup into its own section?

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
+	dvc exp apply <exp>
+
+To promote an experiment to a Git branch run:
+
+	dvc exp branch <exp>
 ```
 
+The command will run all queued experiments in a random order.
+
 ## Comparing many experiments
```
dberenbaum, line 180: It's awkward and redundant to keep this section as is if the command is now already introduced above.

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
 
 ```dvc
```
dberenbaum, line 129: If we end up introducing `dvc exp show` above, I think we could remove `dvc exp diff`. It's mostly a less powerful version of `dvc exp show`.

content/docs/start/experiments.md
dberenbaum: A couple other suggestions to make the commands and tables easier to read:
* Would `accuracy` or `acc` be sufficient?
* Is`model.name` needed here if we are only using CNN?
* Do we need the nested parameter names (can it just be `conv_units` instead of `model.cnn.conv_units`)?

dberenbaum (commented):
Looks good, and most of my comments are minor grammatical suggestions.

I have two broad questions:
1. Is the new baseline experiment necessary? I think the way you introduce this is better than the current approach, especially because it introduces `dvc exp show` earlier, but it doesn't seem well integrated with the other sections now (creates lots of duplication). I'd suggest either leaving this out for a separate PR or else making bigger changes to remove content from the existing doc.
2. Can we reduce the project complexity? The repo looks great, but if we now have the flexibility of having a project specifically for this doc, I'd prefer we keep the project as simple as possible because it makes the example commands, tables, etc. easier to read, and it's easier for user to look through the contents of the repo.

iesahin:
> 1. Is the new baseline experiment necessary? I think the way you introduce this is better than the current approach, especially because it introduces `dvc exp show` earlier, but it doesn't seem well integrated with the other sections now (creates lots of duplication). I'd suggest either leaving this out for a separate PR or else making bigger changes to remove content from the existing doc.

I feel introducing `dvc exp run` at first with `-S` is too much. IMO we need to tell that `dvc exp run` is a replacement to `dvc repro`, but doing this with `--set-param` is a bit distracting IMHO. I can remove that section but it looks we're delving too early to parameters, before telling `exp run` in general. 

You may be right about repeating `exp run` and `exp show`, over and over, but we are telling experimentation here and experimenting usually involves such repeats with some minor changes.  

> 2. Can we reduce the repo complexity? The repo looks great, but if we now have the flexibility of having a repo specifically for this doc, I'd prefer we keep it as simple as possible because it makes the example commands, tables, etc. easier to read, and it's easier for user to look through the contents of the repo.

At first I wanted to set up many parameters to make a wiggle room for us and for the user. I think it's now safe to remove most of the parameters from `dvc.yaml` and have a clean view. But I also would like to show `--include-params` and `--include-metrics`, as in a realistic project these will most likely be required. 

The code still will read the params from `params.yaml`, but they won't be considered as a dependency. Is this OK?

content/docs/start/experiments.md
```diff
@@ -37,12 +110,13 @@ $ dvc exp run --set-param featurize.max_features=3000
 running experiments. The `--set-param` (or `-S`) flag sets the values for
 <abbr>parameters<abbr> as a shortcut for editing `params.yaml`.
```
dberenbaum, line 111: This whole paragraph repeats what you have already explained above.

content/docs/start/experiments.md
```diff
@@ -54,92 +128,84 @@ reflected in the experiment run.
...
+
+	dvc exp branch <exp>
 ```
 
+The command will run all queued experiments in a random order.
+
 ## Comparing many experiments
 
 To compare all of these experiments, we need more than `diff`. `dvc exp show`
 compares any number of experiments in one table:
```
dberenbaum, line 183: This sentence sounds like it's introducing `dvc exp show`, but it has already been shown above.

dberenbaum (commented):
> I feel introducing dvc exp run at first with -S is too much. IMHO we need to tell that dvc exp run is a replacement to dvc repro, but doing this with --set-param is a bit distracting. I can remove that section but it looks like we're delving too early into parameters, before telling exp run in general.
> 
> You may be right about repeating exp run and exp show, over and over, but we are telling experimentation here, and experimenting usually involves such repeats with some minor changes.

I put a couple more comments on sections where I think the duplication doesn't make sense.

> The code still will read the params from params.yaml, but they won't be considered as a dependency. Is this OK?

You mean you will change `dvc.yaml` but leave the rest of the repo the same?

iesahin:
> Why remove the reference to metrics-parameters-plots? The rationale for including it is that users may otherwise be unfamiliar with how to specify parameters and metrics, which is a prerequisite for a lot of what is below. 

Thank you. I began before that link was there and probably lost it in a rebase. Of course it should be there.

content/docs/start/experiments.md
iesahin: Yes. I would like the users to play with the repo and come up with higher metrics just using the hyperparameters. We can have a _hall of fame_ somewhere maybe.

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
 
-Previously, we learned how to tune [ML pipelines](/doc/start/data-pipelines) and
-[compare the changes](/doc/start/metrics-parameters-plots). Let's further
-increase the number of features in the `featurize` stage to see how it compares.
+In the previous page, we learned how to tune
+[ML pipelines](/doc/start/data-pipelines) and
+[compare the changes](/doc/start/metrics-parameters-plots). In this section
+we'll use a
+[Get Started with Experiments project](https://github.com/iterative/get-started-experiments)
+to illustrate experimentation features in DVC 2.0.
```
jorgeorpinel, line 26: > Cont https://github.com/iterative/dvc.org/pull/2491#discussion_r637487008

```suggestion
[compare the changes](/doc/start/metrics-parameters-plots). Let's now
illustrate the experimentation features in DVC 2.0 with a different project
[example](https://github.com/iterative/get-started-experiments).
```
May not even need to link here, it can be just inside the `<details>` section below. Let's focus on concepts and explaining 🙂

content/docs/start/experiments.md
jorgeorpinel: Are you referring to the stuff in the ` <details>` block @dberenbaum ? (Done per https://github.com/iterative/dvc.org/pull/2491#discussion_r637487171 I think)

Seems OK to have that expandable section there, but let's try to shorten it as much as possible @iesahin ? Here's how...

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+<details>
+
+### ⚙️ Click here for the installation instructions
+
+These commands are run in the
+[`get-started-experiments`](https://github.com/iterative/get-started-experiments)
+project. You can run these commands after cloning the repository and installing
+the requirements.
+
+Please clone and create a virtual environment:
```
jorgeorpinel, line 37:
```suggestion
### ⚙️ Click here to setup the example project.

We'll use our
[`get-started-experiments`](https://github.com/iterative/get-started-experiments)
repo for the usage examples in this page. It can be obtained with Git:
```

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+
+```dvc
+$ git clone https://github.com/iterative/get-started-experiments
+$ cd get-started-experiments
+$ python -m venv .venv
+$ . .venv/bin/activate
+$ python -m pip install -r requirements.txt
```
jorgeorpinel, line 44: Let's just mention/link this as a note after the code block, similar to the **Expand to download example code** block in https://dvc.org/doc/start/data-pipelines#pipeline-stages:

image

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+```
+
+Then you can `dvc pull` to get the dataset and run the commands in this
+document. For detailed information on parameters and the project structure
+please refer to the
+[project repository](https://github.com/iterative/get-started-experiments)
```
jorgeorpinel, line 50:
````suggestion
$ dvc pull
```

> See `dvc pull` for more details.
````

**Review by jorgeorpinel — CHANGES_REQUESTED** [review]
Review based on the intro: I think we need to simplify the explanations, avoid low-level details and too many references to the example repo. It looks a bit like this doc is being written to showcase the example repo but it should be the other way around: docs come first, make up command output if needed, and later update the example repo to match (enough). I don't even think it has to match 100%

> Sorry if some of this overlaps with feedback from @dberenbaum. There are lots of comments already,  which may mean that the changes are too much. Maybe some are out of scope?

Some specific feedback and suggestions 👇

File: content/docs/start/experiments.md (hunk 24 above)
Comment (jorgeorpinel) [reply content/docs/start/experiments.md:44]: BTW https://github.com/iterative/get-started-experiments/issues/2

**Comment by jorgeorpinel** [comment]
Logistics:
- Mark resolved comments if possible
- `git pull` (I committed something)
- Resolve conflicts with `master`.
Thanks

File: content/docs/start/experiments.md (hunk 6 above)
Comment (iesahin) [reply content/docs/start/experiments.md:79]: I've updated the repo to contain only the following params and metrics will be limited to `(val_)acc` and `(val_)loss`.

```yaml
epochs: 1
conv_units: 16
dense_units: 128
dropout: 0.5
```

I'll renew all the commands and screenshots. @dberenbaum @jorgeorpinel

content/docs/start/experiments.md
dberenbaum: I mean that `Running experiments` includes:
1. Setup
2. Running a baseline experiment
3. Showing the experiment results
4. Running an experiment with different param values

It's not that any individual topic is too long, but it might be too many concepts under one section. Can we break up into smaller sections?

shcheklein:
My take: 

- this is WIP (as designated by @iesahin )
- let's not merge and do any significant changes until we have some clarify on the #2496 - it might turn out that it's good to have a section that introduces experiments on top of the existing old school "example-get-started"

iesahin:
>     * it might turn out that it's good to have a section that introduces experiments on top of the existing old school "example-get-started"

I think we will want to show some significant experiment results in an experimentation document. With `example-get-started`, it might be a fruitless effort but we can keep this as is for the time being.

shcheklein:
> I think we will want to show some significant experiment results in an experimentation document. With example-get-started, it might be a fruitless effort but we can keep this as is for the time being.

yes, but the first priority should be to show the possibility of this thing. We might btw rewrite the text a bit and when we do bigrams we can them first with `dvc exp` multiple times - this way we won't need a separate section on top of the example-get-started

content/docs/start/experiments.md
iesahin: Done, thank you. 👍🏼

content/docs/start/experiments.md
jendefig: Good idea @iesahin !!

iesahin: I removed this part in #2574 because the repo is very simple now. Maybe we can have multiple versions in the future.

content/docs/start/experiments.md
iesahin: It's removed in #2574, much simpler now.

content/docs/start/experiments.md
iesahin: This is removed in #2574.

content/docs/start/experiments.md
iesahin: I'm using the links in both places in #2574.

content/docs/start/experiments.md
iesahin: This is done in #2574

content/docs/start/experiments.md
iesahin: I replaced `python -m venv` with `virtualenv`

content/docs/start/experiments.md
iesahin: The sections in #2574  are smaller.

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+
+</details>
 
 `dvc exp run` makes it easy to change <abbr>hyperparameters</abbr> and run a new
-experiment:
+experiment. We'll use it to search for parameters to increase the classification
+performance of our model using the
+[Fashion-MNIST](https://github.com/iterative/get-started-experiments) dataset.
```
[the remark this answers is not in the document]
iesahin: The emphasis to the dataset is reduced in #2574

content/docs/start/experiments.md
```diff
@@ -18,17 +18,90 @@ once they're no longer needed.
...
+
+Let's see some metrics produced by this baseline experiment:
 
 ```dvc
-$ dvc exp run --set-param featurize.max_features=3000
+$ dvc exp show --include-metrics categorical_accuracy \
+               --include-params  model.name,model.cnn.conv_units
 ```
 
+![dvc exp show result screenshot](/img/doc/start/exp-ss-63714.png)
```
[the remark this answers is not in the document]
iesahin: I simplified the project to remove these `--include-X` options, they are told separately in #2574 as optional ways to select columns.

iesahin:
We can close this in favor of #2574 now.

PR #2500, opened 2021-05-24
Fix #1333 right panel scroll bar improvements

### LINKED ISSUE #1333: right panel scroll bar improvements
Follow up after https://github.com/iterative/dvc.org/issues/1198

- [ ] When scroll is enabled and some items are not visible as we scroll the main page we want for the right bar to follow (highlighted item should be always visible).
- [ ] Safari - it's very easy (by scrolling the main page) to get it into the state when it completely ignores scroll inputs. Even if you hover mouse over it.
- [ ] FF: scroll is not visible
- [ ] https://github.com/iterative/dvc.org/pull/1284#issuecomment-631171599 - see screenshot, scroll bar overlaps with text

lateemaspencer:
Fix #1333 right panel scroll bar improvements


jorgeorpinel:
@lateemaspencer thanks a lot! I checked (deployed to https://dvc-org-fix-issue-1333-2xx52ck.herokuapp.com/doc/user-guide/contributing/docs) and this seems to fix everything mentioned in #1333 except that I couldn't try Safari as I'm on Windows (cc @shcheklein)

jorgeorpinel:
❓

PR #2514, opened 2021-05-30
Update Nav design

### LINKED ISSUE #2440: Setup Universal Header/Footers
* [Draft Design Mockup](https://www.figma.com/file/FOc5Lm4yy41tSu3GXpqEVN/Websites?node-id=0%3A1)
* Once design is completed, we need to update our header/footer to match design

julieg18:
* Updated nav design based off Figma design in #2440
* I also edited the mobile menu since the tabbing wasn't working correctly.
* There is some crowding in between screen size changes. I lowered the margins in some places to keep the crowding as small as possible but if you have any suggestions to improve it, please let me know!
* Since there were no focus/active styles on Figma, I just went with I thought matched the site styles originally but let me know if any the styles should be changed!
Fixes #2440


julieg18:
Looks like I'll need to do some refactoring with the nav and the header popup JSX to correct the code climate errors.

src/components/LayoutHeader/Nav/styles.module.css
```diff
@@ -20,111 +20,158 @@
...
-      margin-left: 15px;
+    &:first-child {
+      display: none;
+    }
+  }
+
+  @media (--sm-scr) {
+    margin-left: 4px;
+
+    &:nth-child(5),
```
julieg18, line 52: I used CSS to take off the nav items, but we could also use JS to check for screen overflow and take off the items with JS.

julieg18:
Should we close the popups if a user clicks on one of the popup links?  🤔

src/components/HamburgerMenu/styles.module.css
```diff
@@ -20,40 +20,108 @@
...
+}
+
+.company {
+  @mixin active;
+  @mixin hover;
+
+  font-size: 15px;
+  line-height: 20px;
+  font-weight: 500;
+  letter-spacing: 0.02em;
```
arcticbear, line 100: Let's get rid of all letter-spacing in the header elements (0.2em → 0), the reason for this was to improve readability, but it looks a bit weird.

src/components/LayoutHeader/Nav/Popup/styles.module.css
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    .title {
+      grid-area: title;
+      font-weight: 500;
+      font-size: 16px;
+      line-height: 24px;
+      display: flex;
+      align-items: center;
+      letter-spacing: 0.02em;
```
arcticbear, line 76: The same here, let's make it 0em.

src/components/LayoutHeader/Nav/Popup/styles.module.css
```diff
@@ -0,0 +1,94 @@
...
+      width: 32px;
+      grid-area: icon;
+      display: flex;
+      align-items: center;
+    }
+
+    .title {
+      grid-area: title;
+      font-weight: 500;
+      font-size: 16px;
```
arcticbear, line 72: Fonts for dropdowns feel a bit small, my suggestion here to increase the titles font size from 16 to 18 px.

src/components/LayoutHeader/Nav/Popup/styles.module.css
```diff
@@ -0,0 +1,94 @@
...
+      opacity: 0;
+      width: 16px;
+      height: 16px;
+      margin-left: 8px;
+    }
+
+    .description {
+      grid-area: description;
+      color: rgba(15, 22, 36, 0.5);
+      font-size: 14px;
```
arcticbear, line 89: As well here, let's increase the descriptions font size from 14 to 16 px.

arcticbear:
> Should we close the popups if a user clicks on one of the popup links? 🤔

Good question. For example, in the case when a user opened a popup and with Cmd/Ctrl + clicking on links to open them in the separate tabs it would be useful to keep the popup opened after every click. But in the case of single clicks, we should close the popup, especially when we open a new page in the current tab.

rogermparent:
> But in the case of single clicks, we should close the popup, especially when we open a new page in the current tab.

On top of that, the Community menu is all links to different headers within the same page. I think we'll want to dismiss the menu on click there as well, since I don't see users wanting to go through multiple headers in one opening of the menu.

shcheklein:
Is it only me, or it looks too heave with this font size and so many items? (it take almost of all my screen and fonts size is close to some headers on the page?)

rogermparent:
> Is it only me, or it looks too heavy with this font size/weight and so many items? (it take almost of all my screen and fonts size is close to some headers on the page?)

Agreed, this is particularly noticeable at 1>000px width, but even the breakpoint below seems a tad heavy.

julieg18:
> Is it only me, or it looks too heavy with this font size/weight and so many items? (it take almost of all my screen and fonts size is close to some headers on the page?)

Also agreed! What if we made another drop-down that held some of the other nav links on smaller screen sizes?🤔

arcticbear (commented):
Also, the links `Community` and `Meet the community` are similar I guess, they lead to the sections which very close to each other on the top of the same page. What do you think if we hide the second one and rename the first to `Meet the community`?

julieg18:
Ok, I completed all of @arcticbear's comments and update the nav popups to close on click!

rogermparent:
I'd say the font sizes could be a little smaller as well, and I'm not too big a fan of having different sets of items available at different breakpoints- the current dvc.org displays all items on the header or all items in the burger menu. Now that I look at it, we'll also need to add "Other Tools" to the burger menu.

If we decrease the font size, we can and should keep the tap targets 48px with `padding` or some other method of sizing, even if we must reduce the padding/margin of parent containers such that more of the total header area is tappable.

shcheklein:
@arcticbear could you please find some time to review this? :)

julieg18:
Just noticed that some of the footer svgs disappear when the popups disappear:

image
image

rogermparent (approved):
Looks good enough to me as a first iteration!

PR #2539, opened 2021-06-04
Fix the naming of the label template anchors

ChristianLemke:
Fix the naming of the <DVC_METRIC_X_LABEL> and <DVC_METRIC_Y_LABEL> template anchors.


pared:
Seems like we did not modified docs when we renamed the axis markings.

pared (approved):

PR #2546, opened 2021-06-09
azure: remove note about bucket autocreation

efiop:
Per https://github.com/iterative/dvc/pull/6086#issuecomment-857287644

### DIFF
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -327,9 +327,6 @@ storage. Whether they're effective depends on each storage platform.
   $ dvc remote modify myremote url azure://mycontainer/path
   ```
 
-  Note that if the given container name isn't found in your account, DVC will
-  attempt to create it.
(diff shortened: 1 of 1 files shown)


jorgeorpinel (approved):

jorgeorpinel:
Thanks @efiop !

Do you think we should state somewhere (or in all the appropriate remotes) that DVC **does not** create buckets/containers? Cc @dberenbaum

dberenbaum:
Personally, I wouldn't expect DVC to create buckets, but maybe others would, and it probably can't hurt to be explicit.

jorgeorpinel:
You're right. Let's skip it unless we see that it causes confusion with users 👍

dberenbaum:
What was the context for trying to create buckets in the first place?

efiop:
@dberenbaum Lack of standards when it was first contributed and hesitation to break the existing behavior until now.

PR #2547, opened 2021-06-09
document ssl_verify for http/webdav remotes

skshetry:
Adding `ssl_verify` to http and webdav remotes. See https://github.com/iterative/dvc/pull/6142

### DIFF
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -905,11 +905,13 @@ by HDFS. Read more about by expanding the WebHDFS section in
   > `password` is specified, DVC will not prompt the user to enter a password
   > for this remote.
 
-- `ssl_verify` - allows to disable SSH verification, which is enabled by
-  default.
+- `ssl_verify` - whether or not to verify SSL certificates, or a path to a
+  custom CA certificates bundle to do so (implies `true`).
 
   ```dvc
   $ dvc remote modify myremote ssl_verify false
+  # or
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/remote/modify.md
```diff
@@ -905,11 +905,13 @@ by HDFS. Read more about by expanding the WebHDFS section in
   > `password` is specified, DVC will not prompt the user to enter a password
   > for this remote.
 
-- `ssl_verify` - allows to disable SSH verification, which is enabled by
-  default.
+- `ssl_verify` - whether or not to verify SSL certificates, or a path to a
+  custom CA certificates bundle to do so (implies `true`).
```
jorgeorpinel, line 909: I think the most used term out there is just "CA bundle"

jorgeorpinel:
```suggestion
  custom CA bundle to do so (`true` by default).
```

content/docs/command-reference/remote/modify.md
```diff
@@ -966,6 +968,15 @@ by HDFS. Read more about by expanding the WebHDFS section in
   $ dvc remote modify myremote ask_password true
   ```
 
+- `ssl_verify` - whether or not to verify SSL certificates, or a path to a
+  custom CA certificates bundle to do so (implies `true`).
```
jorgeorpinel, line 972:
```suggestion
  custom CA bundle to do so (`true` by default).
```

jorgeorpinel (approved):
Thanks!

Issue #2545, opened 2021-05-31
`exp run`: checkpoints run every time without dep changes


DavidGOrtega:
UPDATE: Skip to https://github.com/iterative/dvc.org/issues/2545#issuecomment-857212104

---

Hi!

Im running into an issue that probably is my mistake so I apologise in advance 🙏 
To follow up you just need to fork https://github.com/iterative/dvc-checkpoints-mnist

## Description

Every time I run ```dvc exp run``` I have 10 epochs even I have not changed any dependancy.

### Reproduce

```sh
git clone https://github.com/iterative/dvc-checkpoints-mnist
cd dvc-checkpoints-mnist
dvc init -f
dvc remote add -d myremote...
dvc exp run

git add dvc.yaml dvc.lock dvclive.json train.py
git commit -m '10 epochs'

dvc push
git push
```

Now if I repeat ```exp run``` without any change It should not run like repro, right? However in my case every single time it runs another 10 epochs.

<details><summary>LOG</summary>
<p>

```sh
Running stage 'train':                                                                                                                                 
> python3 train.py
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration 'd64afd6'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration '8f532c0'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration '9f5b61b'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration '1ba2514'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration 'd70a50f'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration '450cab5'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
Checkpoint experiment iteration 'a599f5e'.

file:///iterative/dvc-checkpoints-mnist/dvclive.html
Updating lock file 'dvc.lock'                                                                                                                          
[...]

PR #2549, opened 2021-06-10
guide: simplify checkpoints info

jorgeorpinel:
Fixes https://github.com/iterative/dvc.org/issues/2545#issuecomment-851740655


content/docs/command-reference/exp/run.md
```diff
@@ -68,37 +68,29 @@ committing them to the Git repo. Unnecessary ones can be removed with
 
 ## Checkpoints
 
-To track successive steps in a longer <abbr>experiment</abbr>, you can register
-checkpoints with DVC during your code or script runtime (similar to a logger).
+To track successive steps in a longer or deeper <abbr>experiment</abbr>, you can
+register checkpoints from your code (similar to a logger). Each `dvc exp run`
```
pmrowla, line 72: I don't really understand the `similar to a logger` note here? Checkpoints don't really have anything to do with logging

content/docs/command-reference/exp/run.md
```diff
@@ -68,37 +68,29 @@ committing them to the Git repo. Unnecessary ones can be removed with
...
+will resume from the last checkpoint.
 
-To do so, first mark stage `outs` with `checkpoint: true` in `dvc.yaml`. At
-least one checkpoint <abbr>output</abbr> is needed so that the experiment can
-later continue from that output's last cached state.
+First, mark at least stage <abbr>output</abbr> with `checkpoint: true` in
+`dvc.yaml`. This is needed so that the experiment can resume later, based on the
+<abbr>cached</abbr> output(s) (circular dependency).
```
pmrowla, line 77: We don't mention circular dependencies anywhere else in this doc, this seems like it might be confusing to include it here

content/docs/command-reference/exp/run.md
jorgeorpinel: Good Q @pmrowla (and sorry for the delay). I didn't change that in this PR BTW.

Aren't checkpoints a way to log experiment steps though? We even have a logger library that registers checkpoints (DVCLive).

Open for suggestions though. Just remove the parenthesis?

content/docs/command-reference/exp/run.md
jorgeorpinel: We don't mention that concept in general (except once in https://dvc.org/doc/api-reference/make_checkpoint). I just decided to stop avoiding it in this PR since it's an important one for checkpoints, I think. Keep in mind this should (eventually will) be moved to a proper guide where we can expand a bit more on this. So I think it's OK to keep it.

content/docs/command-reference/exp/run.md
pmrowla: I would just remove it

jorgeorpinel: OK!
```suggestion
register checkpoints from your code. Each `dvc exp run`
```

jorgeorpinel: Done!

pmrowla (approved):

Issue #2561, opened 2021-06-11
Google Drive: Improve service account instructions


duijf:
Prompted by: https://discuss.dvc.org/t/dvc-studio-we-could-not-detect-a-dvc-data-remote-in-this-repository/788/11

The current instructions on how to set up Google Drive service accounts do not mention that users should share the Google Drive folder with the service account in question. We should emphasize this in our docs (if this is indeed the recommended way to configure this).

PR #2562, opened 2021-06-11
Document sharing folders with Google Drive service accounts

duijf:
Fixes https://github.com/iterative/dvc.org/issues/2561

This PR changes the setup instructions for Google Drive to include instructions on sharing Drive folders with a service account. This step was missing from our guide up until now.

Notes:

 - I could not get the local build to work correctly (running into some weird problems generating thumbnails -- also happens on `master`), so haven't actually inspected the final page yet.
- The demo screenshot I included contains my `laurens@iterative.ai` email. Is this OK or do we have a policy to use test accounts in cases like this?


content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -195,6 +195,12 @@ authentication is needed.
                  gdrive_service_account_json_file_path path/to/file.json
    ```
 
+3. Share the Google Drive folders that you want to use with the service account.
+   Navigate to your Google Drive folder's sharing options and add the service
+   account as an editor:
```
shcheklein, line 200: can it be a reader (depending on the use case?)

duijf: Included this in the text. Kept it brief, but let me know if we want to elaborate on this more.

jorgeorpinel (approved):
LGTM. Feel free to merge if you're done here, @duijf 🙂

PR #2565, opened 2021-06-12
misc. updates

jorgeorpinel:
1. `path` arg usage (`list/get/import`)
2. A typo
3. Docs contrib guide update
4. `get/import-url` don't support `remote://` URLs

- [ ] (1) requires a matching core `dvc` repo update.

### DIFF
--- a/content/docs/command-reference/get.md
+++ b/content/docs/command-reference/get.md
@@ -14,6 +14,7 @@ usage: dvc get [-h] [-q | -v] [-o <path>] [--rev <commit>] [-j <number>]
 positional arguments:
   url              Location of DVC or Git repository to download from
   path             Path to a file or directory within the repository
+                   relative to the root of the repo
 ```
 
 ## Description
--- a/content/docs/command-reference/get.md
+++ b/content/docs/command-reference/get.md
@@ -14,7 +14,6 @@ usage: dvc get [-h] [-q | -v] [-o <path>] [--rev <commit>] [-j <number>]
 positional arguments:
   url              Location of DVC or Git repository to download from
   path             Path to a file or directory within the repository
-                   relative to the root of the repo
 ```
 
 ## Description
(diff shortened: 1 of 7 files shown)


content/docs/command-reference/get.md
```diff
@@ -14,6 +14,7 @@ usage: dvc get [-h] [-q | -v] [-o <path>] [--rev <commit>] [-j <number>]
 positional arguments:
   url              Location of DVC or Git repository to download from
   path             Path to a file or directory within the repository
+                   relative to the root of the repo
```
shcheklein, line 17: how can it be absolute though? 

for `dvc get .` we should probably try to support absolute path and resolve it

shcheklein: also, seems like we can specify specific details like this in docs?

jorgeorpinel: There's a discussion on whether we should even support absolute `path` (https://github.com/iterative/dvc/issues/2613) so maybe we can update the docs again once that's resolved?

I think it's an edge case so didn't mention here, but if you prefer I can mention somewhere in the description below for now... ⌛

jorgeorpinel: > we can specify specific details like this in docs?

I guess so. A user complained that there was no way to know this so I decided to put it up here but I guess it's OK in the desc. ⌛

jorgeorpinel: OK I removed from up here and added this sentence in the descriptions:

> It can be relative to the root of the repo at `url`, or absolute (from the root of the source file system).

The absolute part may be confusing but that's how it works now AFAIK.

shcheklein: > or absolute (from the root of the source file system).

does it work for local repos only, or only for `.`? Depending on the answer we might want to change it to something like:

"It must be a relative path (absolute path is supported for `.` or local repos)"

jorgeorpinel: In theory it can work in either scenario I think, but you would have to know the location of the temp dir where the repo gets cloned to use a correct absolute path. So in practice only for local paths, yes.

> must be a relative path

Yes but relative to what? I think that's the key part

> supported for . or local repos

Added this info now, PTAL.

jorgeorpinel: > So in practice only for local `url`s, yes

BTW also confusing since `url` there is not a URL but a path...

shcheklein: 👍

shcheklein (approved):

PR #2570, opened 2021-06-18
codespaces: custom container and config

jorgeorpinel:
> Followed https://docs.github.com/en/codespaces/setting-up-your-codespace/configuring-codespaces-for-your-project#using-a-predefined-container-configuration

This customization uses a lighter Docker container and avoids running `yarn && yarn build` by default. It also includes a couple useful VSCode extensions for writing docs.

Contributors should be able to use `yarn && yarn develop` (as indicated in our dev env instructions) right away after the codespace is build/restarted.

> Unfortunately this ~~doesn't work at the moment with this repo since the dev app fails to render when shown via a forwarded port~~ works in the desktop app but not in-browser (as a workaround, it can be published with `ngrok` for now, but the dev auto-refresh doesn't work that way).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.177.0/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment if you want to install an additional version of node using nvm
+# ARG EXTRA_NODE_VERSION=10
+# RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
+
+# [Optional] Uncomment if you want to install more global node modules
+# RUN su node -c "npm install -g <your-package-list-here>"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.177.0/containers/javascript-node
+{
+  "name": "Node.js",
+  "build": {
+    "dockerfile": "Dockerfile",
+    // Update 'VARIANT' to pick a Node version: 12, 14, 16
+    "args": { "VARIANT": "14" }
+  },
+
+  // Set *default* container specific settings.json values on container create.
+  "settings": {
+    "terminal.integrated.shell.linux": "/bin/bash"
+  },
+
+  // Add the IDs of extensions you want installed when the container is created.
+  "extensions": [
+    "dbaeumer.vscode-eslint",
+    "stkb.rewrap",
+    "streetsidesoftware.code-spell-checker"
+  ],
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  // "forwardPorts": [],
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  // "postCreateCommand": "yarn install",
+
+  // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
+  "remoteUser": "node"
+}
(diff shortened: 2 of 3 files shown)


jorgeorpinel:
This customization uses a lighter Docker container and avoids running `yarn && yarn build` by default.

Contributors should be able to use `yarn && yarn develop` instead as indicated in our dev env instructions.

> Unfortunately this doesn't work at the moment with this repo since the dev app fails to render when shown via a forwarded port. But we could create an issue to tackle that if needed. As a workaround, it can be published with `ngrok` for now (dev auto-refresh doesn't work though).

jorgeorpinel:
p.s. for more advanced setup options including custom init scripts, we can refer to https://docs.github.com/en/codespaces/setting-up-your-codespace/configuring-codespaces-for-your-project#creating-a-custom-codespace-configuration

jorgeorpinel:
After some time using Codespaces I can vouch for it, esp. in the desktop app flavor (not so much in-browser). And this change does help speed up the env creation + this container comes with Oh My Zsh which speeds up the Terminal experience (esp its pre-enabled [Git plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/git)). The recommended spelling and text wrapping extensions are also useful.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.177.0/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
```
shcheklein, line 4: Do we support Node 16 for the dvc.org cc @rogermparent ?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
```
shcheklein, line 8: can we install DVC here? 

and the VS Code extension (probably we would need some cli command to install from VSIX)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+
+  // Add the IDs of extensions you want installed when the container is created.
+  "extensions": [
+    "dbaeumer.vscode-eslint",
+    "stkb.rewrap",
+    "streetsidesoftware.code-spell-checker"
+  ],
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
```
shcheklein, line 23: This should help to run the website remotely?

README.md
```diff
@@ -23,6 +23,22 @@ details.
 [![](https://sourcerer.io/fame/shcheklein/iterative/dvc.org/images/6)](https://sourcerer.io/fame/shcheklein/iterative/dvc.org/links/6)
 [![](https://sourcerer.io/fame/shcheklein/iterative/dvc.org/images/7)](https://sourcerer.io/fame/shcheklein/iterative/dvc.org/links/7)
 
+## Developing online
+
+Our repository has GitHub [Codespaces](https://docs.github.com/en/codespaces)
```
shcheklein, line 28: I wonder if can now make Edit on Codespaces button?

.devcontainer/Dockerfile
jorgeorpinel: It works well.

rogermparent: I'm pretty sure node 16 is what broke CI builds and that's why we now specify <=15 in package.json.

jorgeorpinel: I think I tried 14 first and it didn't work.

.devcontainer/Dockerfile
jorgeorpinel: > can we install DVC here?

Is it needed for the VCS extension? Otherwise maybe contributors should have `dvc` locally.

> and the VS Code 

But there doesn't seem to be a defined place to get the VSIX file programmatically? You have to go to https://github.com/iterative/vscode-dvc/actions/, choose a passing build (unclear which one will be more stable), and download the artifact (and unzip).

.devcontainer/devcontainer.json
jorgeorpinel: Yep. No need to define them though, as soon as you `yarn develop` and port 8000 is up, Codespaces automatically forwards that port to your localhost:8000 🧙

README.md
jorgeorpinel: Meaning to open the particular file in your VSC codespace? That would be cool!

Keep in mind the first time you open Codespaces the environment needs to build itself, which takes a while (say 5-10 min). And when you reopen after being inactive it takes a minute or so to restart the box.

.devcontainer/Dockerfile
rogermparent: We'll probably need to do a real publish before we install the VSCode extension in this thing. Maybe we can get away with it if we mark the Extension "pre-alpha"? Installing it in codespaces would effectively be making it public anyway.

jorgeorpinel: I guess I'll try this separately in an actual DVC repo like https://github.com/iterative/example-get-started. It may make more sense there.

README.md
jorgeorpinel: I don't think there's a way to tell it what file to open though. At least nothing mentioned in https://docs.github.com/en/codespaces/developing-in-codespaces/creating-a-codespace.

Created https://github.community/t/codespaces-is-it-possible-to-open-on-a-specific/186839.

jorgeorpinel:
OK well this was kind of a proof of concept. We don't really need it and since it was never approved I'm closing as stale.

PR #2574, opened 2021-06-21
start: new experiments introduction

### LINKED ISSUE #2479: start: A new GS:Experiments document with `get-started-experiments`

iesahin:
This PR is a WIP that updates https://dvc.org/doc/start/experiments 

* Keeps run / show / apply and removes the rest of the material
* Uses `get-started-experiments` in command examples
* Some commands may need an update after iterative/dataset-registry#18
* Closes #2479 
* Replaces #2497 
* Related #2496

### DIFF
--- a/content/docs/start/experiments.md
+++ b/content/docs/start/experiments.md
@@ -2,9 +2,7 @@
 title: 'Get Started: Experiments'
 ---
 
-# Get Started: Experiments
-
-⚠️ This feature is only available in DVC 2.0 ⚠️
+# Get Started with Experiments
 
 <abbr>Experiments</abbr> proliferate quickly in ML projects where there are many
 parameters to tune or other permutations of the code. We can organize such
@@ -13,268 +11,205 @@ track experiments for you so there's no need to commit each one to Git. This way
 your repo doesn't become polluted with all of them. You can discard experiments
 once they're no longer needed.
 
-> 📖 See [Experiment Management](/doc/user-guide/experiment-management) for more
-> information on DVC's approach.
-
-## Running experiments
-
 Previously, we learned how to tune [ML pipelines](/doc/start/data-pipelines) and
 [compare the changes](/doc/start/metrics-parameters-plots). Let's further
 increase the number of features in the `featurize` stage to see how it compares.
+<details>
+## Installing and Configuring the Project
 
-`dvc exp run` makes it easy to change <abbr>hyperparameters</abbr> and run a new
-experiment:
+These commands are run in the
+[`get-started-experiments`](https://github.com/iterative/get-started-experiments)
+project. You can run the commands in this document after cloning the repository
+and installing the requirements.
+
+### Clone the project and create venv
+
+Please clone the project and create a virtual environment. 
+
+> We create a virtual environment to keep the libraries we use isolated from
+> the rest of your system. This prevents version conflicts.
 
 ```dvc
-$ dvc exp run --set-param featurize.max_features=3000
+$ git clone https://github.com/iterative/get-started-experiments -b git-init
+$ cd get-started-experiments
+$ virtualenv .venv
+$ . .venv/bin/activate
+$ python -m pip install -r requirements.txt
 ```
+
+### Create the pipeline
+
+DVC tracks the dependencies across project elements with pipelines. Experiments
+also need a pipeline to run.  
+
+We create a two stage pipeline to expand the dataset from a `tar.gz` file
+to a directory and specify the training command to depend on dataset and a
+parameter, and produce a metrics file.
+
+```dvc
+$ dvc stage add -n expand \
+                -d data/fashion-mnist.tar.gz \
+                -o data/images \
+                tar xvzf data/fashion-mnist.tar.gz
+$ dvc stage add -n train \
+                -p conv_units \
+                -d data/images \
+                -m metrics.json \
+                python3 train.py 
+```
+
+### Get the data set
+
+The repository you cloned doesn't contain the dataset. DVC tracks the datafiles
+separately from the text files.  In order to get `fashion-mnist.tar.gz` from
+the `dataset-registry`, we ask DVC to pull the missing data files. 
+
+```dvc
+$ dvc pull
+```
+
+</details> ## Running experiments ### Running with default parameters
+
+The purpose of `dvc exp` commands is to run the pipeline for ephemeral
+experiments. By _ephemeral_ we mean the experiments can be run without
+committing parameter and dependency changes to Git. Instead the artifacts
+produced for each experiment is tracked by DVC and persisted on demand. 
+
+Running the pipeline with default values requires only the command:
+
+``` dvc
+$ dvc exp run
+TK
+```
+
+It runs the pipeline starting from the basic dependencies and produces
+`metrics.json` file for the default state. 
 
 <details>
 
-### 💡 Expand to see what happens under the hood.
+### If you used `dvc repro` before
 
-`dvc exp run` is similar to `dvc repro` but with some added conveniences for
-running experiments. The `--set-param` (or `-S`) flag sets the values for
-<abbr>parameters<abbr> as a shortcut for editing `params.yaml`.
+Earlier versions of DVC uses `dvc repro` to run the pipeline. If you already
+have a DVC project, you may be used to `dvc repro`.
 
-Check that the `featurize.max_features` value has been updated in `params.yaml`:
+In DVC 2.0 `dvc exp run` supersedes `dvc repro`. Both of these commands run the pipeline.
 
-```git
- featurize:
--  max_features: 1500
-+  max_features: 3000
-```
+We use `dvc repro` to run the pipeline as found in the <abbr>workspace</abbr>.
+All the parameters and dependencies are retrieved from the current workspace.
+It doesn't use any special objects to track the experiments.
+	 
+When you have large number of experiments that you don't want to commit into
+Git, it's better to use `dvc exp run`. It allows to change the parameters
+quickly, can track the history of artifacts and has facilities to compare
+these experiments easily.
 
-Any edits to <abbr>dependencies</abbr> (parameters or source code) will be
-reflected in the experiment run.
-
+`dvc repro` is still available to run the pipeline that don't need these extra
+features. 
+ 
 </details>
 
-`dvc exp diff` compares experiments:
+### Running by setting parameters
+
+Now let's do some more experimentation. 
+
+DVC allows to update the parameters defined in the pipeline without modifying
+the files manually. We use this feature to set the convolutional units in
+`train.py`.
 
 ```dvc
-$ dvc exp diff
-Path         Metric    Value    Change
-scores.json  avg_prec  0.56191  0.009322
-scores.json  roc_auc   0.93345  0.018087
-
-Path         Param                   Value    Change
-params.yaml  featurize.max_features  3000     1500
+$ dvc exp run --set-param conv_units=24 
+TK
 ```
 
-## Queueing experiments
+Note that the pipeline didn't run the earliest stage. Only the stages that
+depend on the updated parameter and subsequent stages are run. 
 
-So far, we have been tuning the `featurize` stage, but there are also parameters
-for the `train` stage (which trains a
-[random forest classifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html)).
-
-These are the `train` parameters from `params.yaml`:
-
-```yaml
-train:
-  seed: 20170428
-  n_est: 50
-  min_split: 2
-```
-
-Let's set up experiments with different hyperparameters. We can use the
-`--queue` flag to define all the combinations we want to try without executing
-anything (yet):
+When you run `dvc exp run` with `--set-param`, it updates the parameter file.
+We can see the effect of it by looking at the diff. 
 
 ```dvc
-$ dvc exp run --queue -S train.min_split=8
-Queued experiment 'd3f6d1e' for future execution.
-$ dvc exp run --queue -S train.min_split=64
-Queued experiment 'f1810e0' for future execution.
-$ dvc exp run --queue -S train.min_split=2 -S train.n_est=100
-Queued experiment '7323ea2' for future execution.
-$ dvc exp run --queue -S train.min_split=8 -S train.n_est=100
-Queued experiment 'c605382' for future execution.
-$ dvc exp run --queue -S train.min_split=64 -S train.n_est=100
-Queued experiment '0cdee86' for future execution.
+$ git diff params.yaml
+TK
+```
+
+### Run multiple experiments in parallel
+
+Instead of running the experiments one-by-one, we can define them first to run
+them in a batch. This is especially handy when you have long running
+experiments. 
+
+We add experiments to the queue using the `--queue` option of `dvc exp run`. We
+also use `-S` (`--set-param`) to set a value for the parameter. 
+
+``` dvc
+$ dvc exp run --queue -S conv_units=32
+$ dvc exp run --queue -S conv_units=64
+$ dvc exp run --queue -S conv_units=128
+$ dvc exp run --queue -S conv_units=256
 ```
 
 Next, run all (`--run-all`) queued experiments in parallel (using `--jobs`):
 
 ```dvc
 $ dvc exp run --run-all --jobs 2
+TK
 ```
 
-## Comparing many experiments
+## Comparing experiments
 
-To compare all of these experiments, we need more than `diff`. `dvc exp show`
-compares any number of experiments in one table:
+The pipeline is run several times with different parameters. To compare all of
+these experiments, we use `dvc exp show`. This command presents the parameters
+and metrics produced in experiments in a nicely formatted table.
 
 ```dvc
-$ dvc exp show --no-timestamp \
-               --include-params train.n_est,train.min_split
-┏━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
-┃ Experiment    ┃ avg_prec ┃ roc_auc ┃ train.n_est┃ train.min_split ┃
-┡━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
-│ workspace     │  0.56191 │ 0.93345 │ 50         │ 2               │
-│ master        │  0.55259 │ 0.91536 │ 50         │ 2               │
-│ ├── exp-bfe64 │  0.57833 │ 0.95555 │ 50         │ 8               │
-│ ├── exp-b8082 │  0.59806 │ 0.95287 │ 50         │ 64              │
-│ ├── exp-c7250 │  0.58876 │ 0.94524 │ 100        │ 2               │
-│ ├── exp-b9cd4 │  0.57953 │ 0.95732 │ 100        │ 8               │
-│ ├── exp-98a96 │  0.60405 │  0.9608 │ 100        │ 64              │
-│ └── exp-ad5b1 │  0.56191 │ 0.93345 │ 50         │ 2               │
-└───────────────┴──────────┴─────────┴────────────┴─────────────────┘
+$ dvc exp show
 ```
+TK
 
-Each experiment is given an arbitrary name by default (although we can specify
-one with `dvc exp run -n`.) We can see that `exp-98a96` performed best among
-both of our metrics, with 100 estimators and a minimum of 64 samples to split a
-node.
+By default it shows all the parameters and the metrics along with the
+timestamp. If you have large number of parameters, metrics or experiments, this
+may lead to a cluttered view. You can limit the table to specific metrics, or
+parameters, or hide the timestamp column with `--include-metrics`,
+`--include-params`, or `--no-timestamp` options of the command, respectively.
 
-> See `dvc exp show --help` for more info on its options.
+```dvc
+$ dvc exp show --no-timestamp --include-params conv_units --include-metrics acc
+TK
+```
 
 ## Persisting experiments
 
-Now that we know the best parameters, let's keep that experiment and ignore the
-rest.
+After selecting a experiments from the table, you can commit the
+hyperparameters and other dependencies that produced this successful experiment
+to your Git history. 
 
-`dvc exp apply` rolls back the <abbr>workspace<abbr> to the specified
-experiment:
-
-```dvc
-$ dvc exp apply exp-98a96
-Changes for experiment 'exp-98a96' have been applied to your workspace.
-```
-
-<details>
-
-### 💡 Expand to see what happens under the hood.
-
-`dvc exp apply` is similar to `dvc checkout`, but works with experiments
-instead. DVC tracks everything in the pipeline for each experiment (parameters,
-metrics, dependencies, and outputs), retrieving things later as needed.
-
-Check that `scores.json` reflects the metrics in the table above:
-
-```json
-{ "avg_prec": 0.6040544652105823, "roc_auc": 0.9608017142900953 }
-```
-
-</details>
-
-Once an experiment has been applied to the workspace, it is no different from
-reproducing the result without `dvc exp run`. Let's make it persistent in our
-regular pipeline by committing it in our Git branch:
+`dvc exp apply` brings back all specific artifacts and parameters from the
+experiment to the <abbr>workspace</abbr>.
 
 ```dvc
-$ git add dvc.lock params.yaml prc.json roc.json scores.json
-$ git commit -a -m "Preserve best random forest experiment"
+$ dvc exp apply 
+TK
 ```
 
-## Sharing experiments
-
-After committing the best experiments to our Git branch, we can
-[store and share](/doc/start/data-and-model-versioning#storing-and-sharing) them
-remotely like any other iteration of the pipeline.
+We can see the changes in the repository and commit them to Git. 
 
 ```dvc
-dvc push
-git push
+$ git diff
+$ git add .
+$ git commit -m "Successful experiment"
 ```
 
-<details>
+## Go Further
 
-### 💡 Important information on storing experiments remotely.
+You can continue to experiment with [the
+project](https://github.com/iterative/get-started-experiments).  Please see the
+`README.md` file of the project for these. Don't forget to [notify
+us](https://dvc.org/chat) if you happen to find good parameters.
 
-The commands in this section require both a `dvc remote default` and a
-[Git remote](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes). A
-DVC remote stores the experiment data, and a Git remote stores the code,
-parameters, and other metadata associated with the experiment. DVC supports
-various types of remote storage (local file system, SSH, Amazon S3, Google Cloud
-Storage, HTTP, HDFS, etc.). The Git remote is often a central Git server
-(GitHub, GitLab, BitBucket, etc.).
+There are many other features of `dvc exp`, like cleaning up the unused
+experiments, sharing them without committing into Git or getting differences
+between two experiments.
 
-</details>
-
-Experiments that have not been made persistent will not be stored or shared
-remotely through `dvc push` or `git push`.
-
-`dvc exp push` enables storing and sharing any experiment remotely.
-
-```dvc
-$ dvc exp push gitremote exp-bfe64
-Pushed experiment 'exp-bfe64' to Git remote 'gitremote'.
-```
-
-`dvc exp list` shows all experiments that have been saved.
-
-```dvc
-$ dvc exp list gitremote --all
-72ed9cd:
-        exp-bfe64
-```
-
-`dvc exp pull` retrieves the experiment from a Git remote.
-
-```dvc
-$ dvc exp pull gitremote exp-bfe64
-Pulled experiment 'exp-bfe64' from Git remote 'gitremote'.
-```
-
-> All these commands take a Git remote as an argument. A `dvc remote default` is
-> also required to share the experiment data.
-
-## Cleaning up
-
-Let's take another look at the experiments table:
-
-```dvc
-$ dvc exp show --no-timestamp \
-               --include-params train.n_est,train.min_split
-┏━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
-┃ Experiment ┃ avg_prec ┃ roc_auc ┃ train.n_est┃ train.min_split ┃
-┡━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
-│ workspace  │  0.60405 │  0.9608 │ 100        │ 64              │
-│ master     │  0.60405 │  0.9608 │ 100        │ 64              │
-└────────────┴──────────┴─────────┴────────────┴─────────────────┘
-```
-
-Where did all the experiments go? By default, `dvc exp show` only shows
-experiments since the last commit, but don't worry. The experiments remain
-<abbr>cached</abbr> and can be shown or applied. For example, use `-n` to show
-experiments from the previous _n_ commits:
-
-```dvc
-$ dvc exp show -n 2 --no-timestamp \
-                    --include-params train.n_est,train.min_split
-┏━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
-┃ Experiment    ┃ avg_prec ┃ roc_auc ┃ train.n_est┃ train.min_split ┃
-┡━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
-│ workspace     │  0.60405 │  0.9608 │ 100        │ 64              │
-│ master        │  0.60405 │  0.9608 │ 100        │ 64              │
-│ 64d74b2       │  0.55259 │ 0.91536 │ 50         │ 2               │
-│ ├── exp-bfe64 │  0.57833 │ 0.95555 │ 50         │ 8               │
-│ ├── exp-b8082 │  0.59806 │ 0.95287 │ 50         │ 64              │
-│ ├── exp-c7250 │  0.58876 │ 0.94524 │ 100        │ 2               │
-│ ├── exp-98a96 │  0.60405 │  0.9608 │ 100        │ 64              │
-│ ├── exp-b9cd4 │  0.57953 │ 0.95732 │ 100        │ 8               │
-│ └── exp-ad5b1 │  0.56191 │ 0.93345 │ 50         │ 2               │
-└───────────────┴──────────┴─────────┴────────────┴─────────────────┘
-```
-
-Eventually, old experiments may clutter the experiments table.
-
-`dvc exp gc` removes all references to old experiments:
-
-```dvc
-$ dvc exp gc --workspace
-$ dvc exp show -n 2 --no-timestamp \
-                    --include-params train.n_est,train.min_split
-┏━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
-┃ Experiment ┃ avg_prec ┃ roc_auc ┃ train.n_est┃ train.min_split ┃
-┡━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
-│ workspace  │  0.60405 │  0.9608 │ 100        │ 64              │
-│ master     │  0.60405 │  0.9608 │ 100        │ 64              │
-│ 64d74b2    │  0.55259 │ 0.91536 │ 50         │ 2               │
-└────────────┴──────────┴─────────┴────────────┴─────────────────┘
-```
-
-> `dvc exp gc` only removes references to the experiments; not the cached
-> objects associated with them. To clean up the <abbr>cache</abbr>, use
-> `dvc gc`.
+Please see the section on [Experiment
+Management](/doc/user-guide/experiment-management) in the User's Guide or `dvc
+exp` and subcommands in the Command Reference. 


iesahin:
I used the following map to write the document:

Get Started with Experiments-210621133436

shcheklein:
Suggestion - let's not modify the existing file. Let's start a fresh doc/section (name it "Experiments Trail"), so that there is no confusion - it will be easier to review.

content/docs/start/experiments.md
```diff
@@ -13,268 +11,205 @@ track experiments for you so there's no need to commit each one to Git. This way
...
-experiment:
-
-```dvc
-$ dvc exp run --set-param featurize.max_features=3000
-```
-
 <details>
+## Installing and Configuring the Project
 
-### 💡 Expand to see what happens under the hood.
```
jorgeorpinel, line 34: FYI in our engine `<details>` always need to be followed by a heading (any level of `#`s) which will be the section's title. Otherwise the app will [fail to build](https://dashboard.heroku.com/apps/dvc-org-iesahin-new-gs--oepxvh/activity/builds/e67c1ea9-6e13-42d0-bac2-ab535feb2835).

jorgeorpinel: Hmmm and BTW I'm not sure which one caused that build failure since they all seem to have a header 🤷

jorgeorpinel:
> Let's start a fresh doc/section (name it "Experiments Trail"), so that there is no confusion - it will be easier to review

➕ 1️⃣  on this. Is that the plan? It can be a useful trick even if you end up removing the existing file and reusing it's name for the new one just before merging.

Another option that may work is if you remove the file in one commit and then add it again with totally different content in another commit: hopefully GH won't mix in the changes and will just show all red on top and all green in the bottom.

iesahin:
We now have two drafts for the new GS:Experiments. 

[First of them](https://dvc-org-iesahin-new-gs--lq3od2.herokuapp.com/doc/start/experiments-trail/experiments-1) uses `dvc get` and `dvc stage add` to create a pipeline, then proceeds to experiments. 

[The second one](https://dvc-org-iesahin-new-gs--lq3od2.herokuapp.com/doc/start/experiments-trail/experiments-2) starts from an already prepared pipeline and uses `dvc pull` to get the data dependency. 

PTAL and let me know which one is the route to proceed. I've also added a paragraph about `dvc exp branch` and a section about the experiment naming. Thank you. 

@shcheklein @dberenbaum @jorgeorpinel

shcheklein:
@iesahin could we add both to the side bar for now? to simplify the review? :)

jorgeorpinel:
> First of them uses dvc get and dvc stage add to create a pipeline
> The second one starts from an already prepared pipeline

From my PoV this is important but secondary, so I'll look at the main content for now 🙂 (I understand that's the same in both alternatives).

content/docs/start/experiments-trail/experiments-2.md
```diff
@@ -0,0 +1,324 @@
...
+When an experiment is not run yet, only the former hash value is shown.
+
+You can refer to the experiment in `dvc exp apply` or `dvc exp branch` after
+running the experiment with the name starting with `exp-`, or the name you have
+supplied with `dvc exp run --name`.
+
+## Preparing a project for DVC experiments
+
+At the beginning of this document, we assumed that there is already a configured
+DVC project to simplify the introduction. DVC experiments are a feature added in
```
jorgeorpinel, line 282: Other than the big `<details>` section, this is the only extra section in alternative 2, right? BTW I'm not sure it makes sense to have a preparation section by the end... But again, I wouldn't focus on this discussion first.

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+parameters to tune or other permutations of the code. We can organize such
+projects and keep only what we ultimately need with `dvc experiments`. DVC can
+track experiments for you so there's no need to commit each one to Git. This way
+your repo doesn't become polluted with all of them. You can discard experiments
+once they're no longer needed.
+
+Previously, we learned how to tune [ML pipelines](/doc/start/data-pipelines) and
+[compare the changes](/doc/start/metrics-parameters-plots). In this section, we
+will explore the basic features of DVC experiment management with
+[`get-started-experiments`][gse] project.
```
jorgeorpinel, line 17: Maybe it's possible to have a single intro paragraph? I'd try to summarize a little (e.g. "merge" sentences 1 and 2). Some specific notes:

- `permutations of the code`: complex terms for an intro.
- `You can discard experiments`: may be implicit in the concept of "experiment"
- `Previously, we learned`: Does it matter? I.e. are these new GS pages still mostly connected in a sequence?

jorgeorpinel, line 17: Maybe it's possible to have a single intro paragraph? I'd try to summarize a little (e.g. "merge" sentences 1 and 2). Some specific notes:

- `permutations of the code`: complex terms for an intro.
- `You can discard experiments`: may be implicit in the concept of "experiment"
- `Previously, we learned`: Does it matter? I.e. are these new GS pages still mostly connected in a sequence?

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+
+[gse]: https://github.com/iterative/get-started-experiments
```
jorgeorpinel, line 19: p.s. you can even do this (in case you didn't know, and up to you):

```suggestion
[`get-started-experiments`] project.

[`get-started-experiments`]:
  https://github.com/iterative/get-started-experiments
```

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+DVC is ready to run the experiments now!
+
+</details>
+
+## Running the experiment with default parameters
+
+The purpose of `dvc exp` subcommands is to run the pipeline for ephemeral
+experiments. By _ephemeral_ we mean the experiments can be run without
+committing parameter and dependency changes to Git. Instead the artifacts
+produced for each experiment are tracked by DVC and persisted on demand.
```
jorgeorpinel, line 118: `ephemeral experiments`: at some point we decided not to use that term. All experiments are the same, but some you can chose to make persistent as Git commits.

> For some reason the term has lingered internally e.g. Dmitry uses it.

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+
+Running the experiment with default hyperparameter values requires only the
+command:
+
+```dvc
+$ dvc exp run
```
jorgeorpinel, line 124: ~~`default hyperparameter values`:~~
- ~~What do we mean by "default" here? Sounds like DVC auto-fills them.~~
- ~~Link to https://dvc.org/doc/command-reference/params ?~~

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+...
+Reproduced experiment(s): exp-7683f
+Experiment results have been applied to your workspace.
+
+To promote an experiment to a Git branch run:
+
+        dvc exp branch <exp>
```
jorgeorpinel, line 131: I'd remove this part of the terminal output for now (there's be a persisting section later I think):
```suggestion
```

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+It runs the pipeline starting from the basic dependencies and produces
+`metrics.json` file for the default state.
```
jorgeorpinel, line 135: It feels like we're assuming the user will read the long `<details>` preparation section. That's kind of optional reading, I think? For those who want to actually try the sample commands. I speculate that most people only want to read the page and get an idea of the feature so you may have to rethink how to explain this part (if at all).

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+
+<details>
+
+### If you used `dvc repro` before
+
+Earlier versions of DVC uses `dvc repro` to run the pipeline. If you already
+have a DVC project, you may already be using `dvc repro`.
+
+In DVC 2.0 `dvc exp run` supersedes `dvc repro`. Both of these commands run the
```
jorgeorpinel, line 144: I don't think `exp run` is supposed to replace `repro`. Esp not in the DVC 3.0 directions AFAIK.
```suggestion
have a DVC project, you may already be using `dvc repro`. Both of these commands run the
```

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+
+We use `dvc repro` to run the pipeline as found in the <abbr>workspace</abbr>.
+All the parameters and dependencies are retrieved from the current workspace. It
+doesn't use any special objects to track the experiments. When you have large
+number of experiments that you don't want to commit all to Git, it's better to
+use `dvc exp run`. It allows to change the parameters quickly, can track the
+history of artifacts and has facilities to compare these experiments easily.
+
+`dvc repro` is still available to run the pipeline when these extra features are
+not needed.
```
jorgeorpinel, line 155: Not clear what we're trying to say about `repro`. Can probably be a shorter ~~note~~ paragraph 🙂

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Same applies in some other explanations throughout the text.

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Similarly in other terminal samples

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+To promote an experiment to a Git branch run:
+
+        dvc exp branch <exp>
+```
+
+When you run `dvc exp run` with `--set-param`, it updates the parameter file. We
+can see the effect of it by looking at the diff.
+
+```dvc
+$ git diff params.yaml
```
jorgeorpinel, line 182: Why not introduce `exp diff` instead? Assuming it works with a single experiment vs. `HEAD` version. Otherwise maybe nothing: not sure seeing this diff is necessary.

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Actually, maybe no need to mention params yet since the next section focuses on that? Instead, this section could be more general and talk about data and code changes, or even show basic ones.

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+Queued experiment '8bb6049' for future execution.
+```
+
+Next, run all (`--run-all`) queued experiments in parallel (using `--jobs`):
+
+```dvc
+$ dvc exp run --run-all --jobs 2
+```
+
+It runs all the experiments, with the specified number of processes in parallel.
```
jorgeorpinel, line 217: Repetitive, I think:
```suggestion
```

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+
+```dvc
+$ dvc exp apply exp-cb13f
+Changes for experiment 'exp-cb13f' have been applied to your current workspace.
+```
+
+We can see the changes in the repository and commit them to Git.
+
+```dvc
+$ git diff
```
jorgeorpinel, line 261: Again here, probably no need for `diff`. Maybe `status`?
````suggestion
We can see the changes in the repository and commit them to Git.

```dvc
$ git status
...
````

Comment (jorgeorpinel) [reply content/docs/start/experiments-trail/experiments-1.md:261]: OR just don't mention seeing the repo changes, just focus on `git commit`.

File: content/docs/start/experiments-trail/experiments-1.md [hunk 14]
```diff
@@ -0,0 +1,334 @@
...
+When you create an experiment, DVC generates a hash value from the contents of
+the experiment. This is shown when you use `--queue` option, e.g.,
+
+```dvc
+$ dvc exp run --queue -S model.conv_units=32
+Queued experiment '6518f17' for future execution.
+```
+
+After _running_ the experiment, DVC uses another auto-generated name to refer to
+the experiment. Typically these start with `exp-`, and can be set via
```
jorgeorpinel, line 290: Maybe this whole section can be a `<details>` in the **Comparing experiments** section. The first `exp show` table (img) can have some queued experiments, and after that this hidden part will explain the differences for those interested in clicking on it. Then there's no need for a 3rd table in here.

> Also, it's fine to exclude queued exps from the 2nd table even if the continuity isn't perfect.

content/docs/start/experiments-trail/experiments-1.md
```diff
@@ -0,0 +1,334 @@
...
+│ ├── b2b8586 [exp-2a1d5] │ 0.25036 │ 16               │
+│ └── *86bd8f9            │ -       │ 512              │
+└─────────────────────────┴─────────┴──────────────────┘
+```
+
+When an experiment is not run yet, only the former hash value is shown.
+
+You can refer to the experiment in `dvc exp apply` or `dvc exp branch` after
+running the experiment with the name starting with `exp-`, or the name you have
+supplied with `dvc exp run --name`.
```
jorgeorpinel, line 324: This can be part of **Persisting experiments**.

jorgeorpinel (commented):
Overall it seems like all that planning and maps are paying off @iesahin! I'm liking the simplicity and structure. I think it still needs some work (see review 👇) but great first version 👍 

> There's also a bunch of minor details but I'll make specific suggestions for those in a later round.

content/docs/start/experiments-trail/experiments-2.md
iesahin: Alternative 2 moves the _preparation_ discussion to the end and mentions only `git clone` and `dvc pull` in the _Installation_ part

content/docs/start/experiments-trail/experiments-1.md
iesahin: Initial part is a relic from the earlier version. I agree with your points.

content/docs/start/experiments-trail/experiments-1.md
iesahin: :) I see. I need some word to describe _experiments not yet persisted and may be thrown out._ I agree that _ephemeral_ looks like we are working in void and experiments are like clouds, but what can it be? I added the explanation because of this. 

_Impermanent_ feels experiments are about to collapse. 

All experiments are not the same, we breed them and select some of them as our next-generation livestock.

content/docs/start/experiments-trail/experiments-1.md
iesahin: I know this but I don't like to use ` within the links and `[get-started-experiments]` doesn't work :)

content/docs/start/experiments-trail/experiments-1.md
iesahin: I merged the paragraphs and simplified in 09f0c0e

iesahin:
Each push to the repository renews the installation, so I put the documents to the sidebar. You should be able to see them in Get Started section. 📢 😄

content/docs/start/experiments-trail/experiments-1.md
iesahin: `exp diff` reports the results in _metrics_, we are telling how parameters are updated here.

content/docs/start/experiments-trail/experiments-1.md
iesahin: default hyperparameter values ➡ default project settings

content/docs/start/experiments-trail/experiments-1.md
iesahin: I see your point and reworded a bit but assuming the reader _did not_ read the installation section may result in too verbose or abstract explanations. I rather err on the side of being concrete.

content/docs/start/experiments-trail/experiments-1.md
iesahin: `dvc exp run` does the same thing as `dvc repro`, namely _running the pipeline._ A user that knows _dvc repro_ will recognize that this new command does the same thing as `dvc repro`. In 3.0, there may be _experiments without pipeline_, but that's another topic and even in that case, `dvc exp run` will continue to _run the pipeline_ in `dvc.yaml`.

content/docs/start/experiments-trail/experiments-1.md
iesahin: I'm trying to say that `∂vc exp run` is `repro` + features 🤭😄

content/docs/start/experiments-trail/experiments-1.md
iesahin: How can the user know what changed _in files_ without `git diff`? All experiments supposedly change the same files, that is `metrics.json` here, the reason we are selecting a particular experiment is that these metrics and associated files are better than others.

content/docs/start/experiments-trail/experiments-1.md
iesahin: Maybe. I'd like to skip this completely. :)

iesahin:
I have added a third alternative to GS:Experiments trails. It moves the shorter preparation to the main text and moves the final _Preparing projects for DVC Experiments_ in Alternative-2 to a details section below it. 

Could you review and tell me which one you _dislike_ most. I think it's easier to discard the _most disliked_ rather than finding the _most liked_ one 😃 

@shcheklein @jorgeorpinel @dberenbaum @casperdcl

casperdcl:
did a 3-way diff locally to not miss anything.

dislikes:

1. hiding "Installing and Configuring the Project" in `<details>`
2. putting "Preparing a project for DVC experiments" at bottom
3. none

So I prefer v3.

iesahin:
> dislikes:
> 
>     1. hiding "Installing and Configuring the Project" in `<details>`
> 
>     2. putting "Preparing a project for DVC experiments" at bottom
> 
>     3. none

First two are what @dberenbaum told in the meeting. I think we'll decide on v3.

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Great, thanks!

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: > I need some word to describe experiments not yet persisted and may be thrown out.

That's my point though, ideally you don't need to make that distinction. The general concept of "experiment" already implies that it's ephemeral or disposable.

So you can just say something like "we're making experiments here blah blah..." And later in the **Persisting** section then you go into explaining what persistent experiments are (Git commits or branches, i.e. arbitrary project versions).

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Yes but what is "default" here? The reader may not have context as to the specific contents of the sample project if they're just reading casually. Even if they're following but don't have DVC experience they won't know what the default project settings or params are or where to find them. Maybe there's another more general way to say this, like "Use `dvc exp run` to run an experiment based on the current DVC metafiles".

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Thanks, it's good to be concrete/ specific in these cases. But

> It runs the specified command

What command? Sounds like you mean `dvc exp run`.

I guess this sentence is a bit repetitive since the one before the code block already stated what we're doing. You could just go for "You should now wee metrics.json in your workspace." if that's the point of this sentence.

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: > dvc exp run does the same thing as dvc repro

It's not 100% the same but again, that may change in the near future so I wouldn't see it this way. I don't think it was really the intention. The intention is that repro is the interface to eval and exec pipelines (dvc.yaml) while `exp run` puts project versions (executes with `repro` internally if needed) into virtual workspaces that DVC tracks automatically.

Maybe there's no need to even mention `repro` at all TBH. @shcheklein can probably confirm here.

jorgeorpinel: I guess my point is that this whole section emphasizes that `exp run` is the new `repro` and that's not the case. If the idea is to have a mini migration guide for existing DVC repos, a `<details>` section may suffice.

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: (Same answer 😬)

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: Ok. Well `exp diff` shows both params and metrics. If you only want the metrics there's `params diff` too 🔀

jorgeorpinel: I mean the params lol

content/docs/start/experiments-trail/experiments-1.md
jorgeorpinel: If the user knows Git they will run `git diff` if they want to. The terminal sample above doesn't even show the output of `git diff` so I'm not sure what's being illustrated. I don't think we should document Git per se.

> All experiments supposedly change the same files, that is metrics.json here

You could do `exp/metrics diff` here if needed. Not sure whether it is (needed).

iesahin:
I'm selecting alt-3 from the candidates if that's OK with all @iterative/docs :)

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
+---
+title: 'Get Started: Experiments'
+---
+
+# Get Started with Experiments
+
+<abbr>Experiments</abbr> proliferate quickly in ML projects where there are many
+parameters to tune or other permutations of the code. We can organize such
+projects and keep only what we ultimately need with `dvc experiments`. DVC can
+track experiments for you so there's no need to commit each one to Git. This way
```
shcheklein, line 10: New users that come here won't understand this part `to commit each one to Git.`. Committing experiments to Git is very DVC-specific.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
+---
+title: 'Get Started: Experiments'
+---
+
+# Get Started with Experiments
+
+<abbr>Experiments</abbr> proliferate quickly in ML projects where there are many
```
shcheklein, line 7: or data?

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+parameters to tune or other permutations of the code. We can organize such
```
shcheklein, line 8: `We can organize ...` - I'm not sure we need this sentence, and `organizing` project usually means using some specific structure (like https://github.com/drivendata/cookiecutter-data-science) l... at least that the way I would read this

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+projects and keep only what we ultimately need with `dvc experiments`. DVC can
+track experiments for you so there's no need to commit each one to Git. This way
+your repo doesn't become polluted with all of them. You can discard experiments
+once they're no longer needed.
+
+Previously, we learned how to tune [ML pipelines](/doc/start/data-pipelines) and
```
shcheklein, line 14: I think one of the main goals of using separate "Trails" (I love the idea and this name!) - is to make them completely independent. I would not distract people by these back-references. We can simplify this.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+[compare the changes](/doc/start/metrics-parameters-plots). In this section, we
+will explore the basic features of DVC experiment management with
```
shcheklein, line 16: can we explain this from the "what users get with this " perspective? E.g. fast local iterations and ability to pick the best? or something like this? otherwise we explain Experiments with experiments management - it's not very informative I'm afraid.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+[gse]: https://github.com/iterative/get-started-experiments
+
+## ⚙️ Installing the Example Project
+
+These commands are run in the [`get-started-experiments`][gse] project. You can
+run the commands in this document after cloning the repository and installing
+the requirements.
+
+### 👥 Clone the project and create virtual environment
```
shcheklein, line 27: emojis: I think we should use the basic style that we already have (less emojis), it's a bit too playful now :)

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+once they're no longer needed.
+
+Previously, we learned how to tune [ML pipelines](/doc/start/data-pipelines) and
+[compare the changes](/doc/start/metrics-parameters-plots). In this section, we
+will explore the basic features of DVC experiment management with
+[`get-started-experiments`][gse] project.
+
+[gse]: https://github.com/iterative/get-started-experiments
+
+## ⚙️ Installing the Example Project
```
shcheklein, line 21: (disclaimer: I haven't participated actively in the discussion so far, so sorry if I repeat certain concerns/arguments)

I think second main objective we should be building these documents with- focus on main things first. In this case it's `dvc exp run` - I should be able to see it in the fist screen with a table. Everything else should come as expandable and/or after sections.

My point here is that get-started are not regular tutorials, they should have a bit different structure. Reproducibility can go in the Katakoda, for example.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+$ dvc exp branch exp-05e87 "cnn-256"
+Git branch 'cnn-256' has been created from experiment 'exp-05e87'.
+To switch to the new branch run:
+
+        git checkout cnn-256
+```
+
+You can then checkout and continue from working this branch as usual.
+
+## 📛 Note on experiment names
```
shcheklein, line 264: It think this it a bit too much - to have the whole section like this as a Note. Is there a very important reason to make it of the same importance as "running experiments" and other sections?

Can we hide it?

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+```dvc
+$ dvc exp run --queue --name cnn-512 -S model.conv_units=512
+Queued experiment '86bd8f9' for future execution.
+```
+
+In `dvc exp show` you can see both of these names:
+
+```dvc
+$ dvc exp show --no-pager --no-timestamp \
```
shcheklein, line 287: we need to fix the highlighter? cc @jorgeorpinel

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+```git
+-model:
+-  conv_units: 16
++model:
++  conv_units: 24
+```
+
+## 🏃‍♂️🏃🏾‍♂️🏃🏻‍♂️ Run multiple experiments in parallel
```
shcheklein, line 178: Can we merge three different "running experiments sections" into one or at least to? Otherwise it becomes more of a user guide I think. No intention here to be very comprehensive, we need to make it clear what is possible, ideally with one single command, one single table

shcheklein:
@iesahin looks great! I left some comments around. Overall I think it should be further simplified and very focused on a few commands - `dvc exp run`, `dcv exp show` and what people can get from them. + some comments on styles (e.g. emoji - even if we want to go that far, let's do this as a separate PR across existing get started docs) + some other comments.

Next steps for this "Trail" that we should think about to my mind, that might affect this doc (or might not, it's fine to merge and iterate). Since I think the most important thing is to make trails completely independent (each trail is a separate get started), each of them should cover all major DVC features - data, pipelines, metrics. 

It means that for this trail we start with experiments, but we should go forward. And introduce metrics (especially dvclive), cover data management from the experiments angle, project structure and reproducibility (pipelines).

It can be at the end a section that looks like the existing get started. 3-4 sections. Ideally short (at least 1/2 of the one you put together in this PR), start with experiments of course.

Again, I don't this we should solve this as part of this PR. But I would at least keep this in mind an I would try to simplify this doc and focus on `dvc exp` more.

content/docs/start/experiments-trail/experiments.md
jorgeorpinel: Why? What's wrong? Sorry, review app isn't up rn.

content/docs/start/experiments-trail/experiments.md
iesahin: This is the Experiments Trail :)

content/docs/start/experiments-trail/experiments.md
iesahin: These two paragraphs are mostly from the older docs. I'll rewrite them. Thank you.

content/docs/start/experiments-trail/experiments.md
iesahin: There are various aspects of this question:

- Being fully reproducible is better than hiding some commands I think. If someone doesn't know about `dvc pull` (or `tar`, or `git clone`...) and we don't mention here, that may be a blocker for them. ("This thing doesn't work!!") A _veteran_ can simply skip these but a novice may drown in unknown waters. I prefer to err on the side of novices in this section of the docs. 

- I agree with you hiding these in `details` section, and other versions had these hidden. In the discussion, I thought there is _partial consensus_ to move the preparation to the main text by simplifying it. We can discuss this here I believe. @dberenbaum @casperdcl @jorgeorpinel

content/docs/start/experiments-trail/experiments.md
iesahin: Too playful, you say :))

In the first version I didn't put any and it looked a bit grim to me. The current emoji set doesn't correspond to these topics. I can put a "gear" here but gear here, gear there, it becomes too gearful. :)

Let me look at this.

content/docs/start/experiments-trail/experiments.md
iesahin: I rather would like to keep the "sales pitch" in these docs, and IMO that's "running experiments in parallel" for `dvc exp run`. But in order to tell "you can run experiments in parallel", we first need to tell "you can run experiments".

I think keeping these in separate sections emphasize _multiple experiments_. If I merge these, `--queue` option will become a minor thing in a deeper paragraph.

content/docs/start/experiments-trail/experiments.md
iesahin: The reason I put this into a separate section is @aguschin's comments about experiments names, I think in iterative/dvc#6050 . For a newcomer having two names, one reported by `--queue`, one given after the run is a bit confusing and I'd rather have a single autogenerated commit name for both of these. Current behavior is a bit confusing for the newcomer. 

I can move this to a relevant UG page and hide here.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+The command tells DVC to create an experiment named `train`, and that for any
+change in `data/images/`, `model.conv_units` or `train.epochs`, we run an
+experiment using `src/train.py` that produces a new `metrics.json` file.
+
+Note that the parameters (added with `-p`) are in the default parameters file
+`params.yaml` and used in the code as normal, by reading the file. DVC only
+tracks the changes and updates them with `--set-param`. For other files that
+contain the parameters, DVC expects the filename to be provided similar to
+`-p myparams.yaml:epochs`.
```
dberenbaum, line 98: Are `--set-param` or `-p myparams.yaml:epochs` important here?

content/docs/start/experiments-trail/experiments.md
dberenbaum: Agreed. This is also an issue below when introducing _ephemeral_ experiments. Need to think of how to introduce experiments without assuming user knows about dvc's gitops workflow.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+DVC is ready to run the experiments now!
+
+</details>
+
+## 👟 Running the experiment with default parameters
+
+The purpose of `dvc exp` subcommands is to run the pipeline for ephemeral
```
dberenbaum, line 106: Maybe we don't even need _ephemeral_ since it's only really useful to contrast to the `dvc repro` workflow. If we do keep _ephemeral_, we could make it more generic, like "By _ephemeral_ we mean the experiments can be easily discarded."

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+Reproduced experiment(s): exp-7683f
+Experiment results have been applied to your workspace.
+...
+```
+
+It runs the specified command that writes the metrics values to `metrics.json`.
+
+<details>
+
+### 📜 If you used `dvc repro` before
```
dberenbaum, line 125: _ephemeral_ might make more sense in this section.

content/docs/start/experiments-trail/experiments.md
dberenbaum: This is in the current dvc sprint to fix, and I just updated the importance. Let's assume this will be fixed before publication, and we can bump the priority further if needed. Hopefully, we can hide or drop this whole section or at least greatly simplify.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+│ ├── d0ee7ce [exp-5dccf] │ 0.23818 │ 32               │
+│ ├── 1533e4d [exp-88874] │ 0.24039 │ 128              │
+│ ├── b1f41d3 [cnn-256]   │ 0.23296 │ 256              │
+│ ├── 07e927f [exp-6c06d] │ 0.23279 │ 24               │
+│ ├── b2b8586 [exp-2a1d5] │ 0.25036 │ 16               │
+│ └── *86bd8f9            │ -       │ 512              │
+└─────────────────────────┴─────────┴──────────────────┘
+```
+
+When an experiment is not run yet, only the former hash value is shown.
```
dberenbaum, line 305: To add to the confusion 🤣 , the hash value changes before and after the experiment is run. I don't think it's worth getting into any of these details in this doc.

content/docs/start/experiments-trail/experiments.md
dberenbaum: I think keeping the "sales pitch" could mean having a compelling header of a hidden section or otherwise calling out the idea, but keeping how to do it in the hidden section. You can still have all these experiments in the `dvc exp show` table even if they aren't in the main text. Having that table higher up seems important since it's unclear how `dvc exp run` is helpful until that point.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+```dvc
+$ git diff params.yaml
+```
+
+```git
+-model:
+-  conv_units: 16
++model:
++  conv_units: 24
+```
```
dberenbaum, line 176: This is a nice explanation, but could it be hidden?

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+We can see the changes in the repository and commit them to Git.
+
+```dvc
+$ git diff
+$ git add .
+$ git commit -m "Successful experiment"
+```
+
+It may also be desirable to commit a particular experiment to a Git branch
```
dberenbaum, line 251: Do we need to introduce both `apply` and `branch` here? It seems sufficient to pick one as a default and save the other for the UG.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+cluttered view. You can limit the table to specific metrics, or parameters, or
+hide the timestamp column with `--include-metrics`, `--include-params`, or
+`--no-timestamp` options of the command, respectively.
+
+```dvc
+$ dvc exp show --no-timestamp \
+  --include-params model.conv_units --include-metrics acc
+```
+
+![](/img/start-dvc-exp-show-no-timestamp-210704.png)
```
dberenbaum, line 227: This is another part that might be a candidate for a hidden section.

dberenbaum (commented):
The concept looks great and so do the explanations! Most of my comments are on the structure as far as what commands to show in what sections in what order.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+
+<details>
+
+## ⏲ Preparing a project for DVC experiments
+
+In this document we assume that there is already a configured DVC project to
+simplify the introduction. DVC experiments are a feature added in DVC 2.0 and
+requires a DVC pipeline defined in the project.
+
+Please refer to [install](/doc/install) if DVC is not installed in your system.
```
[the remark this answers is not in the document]
iesahin: User may have other preferences about using DVC with venv. I usually use a system-wide DVC. Easier to upgrade and saves taking up space for each project. `get-started-experiments` doesn't list DVC in its `requirements.txt`. If it's the recommended / best-practice to install DVC to `venv`, I should add it to `requirements.txt` too.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,319 @@
...
+The repository already contains the necessary configuration to run the
+experiments. If you would like to learn how to configure a project to run DVC
+experiments, please refer to the last section.
+
+<details>
+
+## ⏲ Preparing a project for DVC experiments
+
+In this document we assume that there is already a configured DVC project to
+simplify the introduction. DVC experiments are a feature added in DVC 2.0 and
```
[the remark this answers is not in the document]
iesahin: removed it. thank you.

content/docs/start/experiments-trail/experiments.md
iesahin: Removed it. _Preparing your project for DVC_ shouldn't be that detailed. Thank you.

content/docs/start/experiments-trail/experiments.md
iesahin: Removed ephemeral by joining the explanation sentence. :+1:

content/docs/start/experiments-trail/experiments.md
iesahin: It looks it's a sticky word to be avoided :) I kept this the same.

content/docs/start/experiments-trail/experiments.md
iesahin: I put the sections in `details`. Let's see how it looks.

content/docs/start/experiments-trail/experiments.md
dberenbaum: cc @karajan1001. If you have capacity, please pick up this issue next.

content/docs/start/experiments-trail/experiments.md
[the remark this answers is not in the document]
dberenbaum: Right. I don't think it's necessarily recommended, just an idea if we want to save a step in setup.

jorgeorpinel:
@iesahin please pull master when you can to reflect the nicer `exp show` tables per https://github.com/iterative/dvc.org/pull/2625. Thanks

jorgeorpinel:
This probably isn't a draft anymore? Already well into reviewing

content/docs/start/experiments-trail/experiments.md
iesahin: I've updated the intro in 4f56d8.

content/docs/start/experiments-trail/experiments.md
iesahin: Removed that part in 4f56d8f4

content/docs/start/experiments-trail/experiments.md
iesahin: Removed the references in 4f56d8f4

content/docs/start/experiments-trail/experiments.md
iesahin: Updated the span of `details` in 9c324a060665f01de3c4210d9ab86551ab4c8088

content/docs/start/experiments-trail/experiments.md
iesahin: removed the naming section

content/docs/start/experiments-trail/experiments.md
iesahin: removed `dvc exp apply` and kept `branch`, as it's the option advertised in the command output as well.

content/docs/start/experiments-trail/experiments-1.md
iesahin: It's still not possible to run an experiment without pipeline, so I think the user is expected to have some "default values" or "default pipeline" defined in their project.

content/docs/start/experiments-trail/experiments-1.md
iesahin: I want to show how the file has changed with `dvc exp run -S`, with an external tool. I could use the plain `diff` by writing the params file somewhere before the experiment but this is simpler IMO.

content/docs/start/experiments-trail/experiments-1.md
iesahin: Removed completely.

iesahin:
This is ready for another round of reviews @shcheklein @dberenbaum @jorgeorpinel  Thank you.

dberenbaum:
Is it possible to force deployment to see the latest version?

shcheklein:
I've deployed it.

shcheklein:
Thanks @iesahin . It looks good to me. A few things that I see:

1. We need to remove the emojis at least for the consistency sake - it's not the regular style that we use. Also, to be honest, I haven't seen that many emojis in titles.
2. I see some glitches (cc @julieg18 and @rogermparent):

run and show are not highlighted:

Screen Shot 2021-09-22 at 10 26 53 PM

Table header is borken:

Screen Shot 2021-09-22 at 10 26 46 PM

In the right side panel I see items that do not belong there:

Screen Shot 2021-09-22 at 10 30 41 PM

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+metrics, and commit only the ones that we need to Git.
+
+In this section, we explore the basic features of DVC experiment management with
+[`get-started-experiments`][gse] project.
+
+[gse]: https://github.com/iterative/get-started-experiments
+
+<details>
+
+## ⚙️ Installing the Example Project
```
shcheklein, line 18: should be `###` for the title and inside those sections I think

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+│ ├── 9c32227 [exp-17dd9] │ 0.9167 │ 256              │
+│ ├── 8a9cb15 [exp-29d93] │ 0.9134 │ 128              │
+│ ├── dfc536f [exp-a1bd9] │ 0.9151 │ 24               │
+│ └── 1a1d858 [exp-6dccf] │ 0.9152 │ 16               │
+└─────────────────────────┴────────┴──────────────────┘
+```
+
+## 🔏 Persisting experiments
+
+After selecting an experiment from the table, you can create a Git branch that
```
shcheklein, line 234: we should create a sense that it's only one of the possible ways I think

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+DVC also requires commands to be run and their dependencies to be defined as
+stages. We use `dvc stage add` to add a stage and set its dependencies.
+
+```dvc
+$ dvc stage add -n train \
+                -p model.conv_units \
+                -p train.epochs \
+                -d data/images \
+                -m metrics.json \
+                python3 src/train.py
```
skshetry, line 74: We should encourage users to update `dvc.yaml` instead of `stage add`. The command may seem useful for simple cases, but it may prevent users from learning our DSL which is what is driving all these pipelines interaction.

Related: #2578

```yaml
train:
  cmd: python3 src/train.py
  deps:
  - data/images
  params:
  - model.conv_units
  - train.epochs
  metrics:
  - metrics.json

File: content/docs/start/experiments-trail/experiments.md [hunk 38]
```diff
@@ -0,0 +1,255 @@
...
+`dvc exp show` to compare all of these experiments. This command presents the
+parameters and metrics produced in experiments in a nicely formatted table.
+
+```dvc
+$ dvc exp show
+```
+
+```dvctable
+┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
+┃ white:**Experiment**    ┃ white:**Created**┃    yellow:**loss** ┃    yellow:**acc** ┃ blue:**train.epochs** ┃ blue:**model.conv_units** ┃
```
julieg18, line 193: Looks like certain parts of the table are missing some white space, causing the table to not align correctly. 

```suggestion
┃ white:**Experiment**              ┃ white:**Created**      ┃ yellow:**loss**    ┃ yellow:**acc**    ┃ blue:**train.epochs** ┃ blue:**model.conv_units** ┃
```

julieg18:
> run and show are not highlighted:

`exp run` and `exp show` are not in [our list of dvc commands](https://github.com/iterative/dvc.org/blob/master/config/prismjs/dvc-commands.js#L3). If they are supposed to be there, I can open a pull request for updating our dvc commands (or commit them to this branch).

> In the right side panel I see items that do not belong there:

As far as I can tell, our code scans a page's markdown ([`parseHeadings`](https://github.com/iterative/dvc.org/blob/29e5d55bc0965b9919abcc3f747e3f76a010239c/src/gatsby/models/docs/createPages.js#L17)), finding any h2 headings and adds them to the page's list of `headings`, which appears on the right side panel. There are `h2` headings in the `details` blocks, which makes them appear on the side panel. Taking out the `h2` headings from the `details` blocks would stop the headings from appearing in the side panel. We could also update our `parseHeadings` function to not add headings found in details blocks to the page 's list of `headings`. 🤔

content/docs/start/experiments-trail/experiments.md
dberenbaum: I think the goal is to revise once we implement https://github.com/iterative/dvc/issues/6446 to make this simpler. @skshetry we should also consider as part of `dvc exp init` and possible as a broader topic how to provide templates or other starting points for `dvc.yaml`.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
+---
+title: 'Get Started: Experiments'
+---
+
+# Get Started with Experiments
+
+In machine learning projects, number of <abbr>experiments</abbr> grow rapidly.
```
dberenbaum, line 7:
```suggestion
In machine learning projects, the number of <abbr>experiments</abbr> grows rapidly.
```

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+update the missing data files.
+
+```dvc
+$ dvc pull
+```
+
+The repository already contains the necessary configuration to run the
+experiments.
+
+## 💡 Preparing a project for DVC experiments
```
dberenbaum, line 51: It's not immediately clear to me that these steps are already complete for the cloned repo. I guess this is more of a reference so that users know how you got to this point? I think it's important to clarify, or my first step would be `dvc init`, which will throw an error.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+
+### 📖 If you used `dvc repro` before
+
+Earlier versions of DVC uses `dvc repro` to run the pipeline. If you already
+have a DVC project, you may already be using `dvc repro`.
+
+In DVC 2.0 `dvc exp run` supersedes `dvc repro`. We use `dvc repro` to run the
+pipeline as found in the <abbr>workspace</abbr>. All the parameters and
+dependencies are retrieved from the current workspace. It doesn't use any
+special objects to track the experiments or associate parameters with metrics.
```
dberenbaum, line 124: I don't fully understand the 2nd-4th sentences of this paragraph and which ones refer to `repro` vs `exp run` behavior. I'm not sure they are needed, since the 5th sentence onwards makes clear to me the rationale for using `exp run`.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+│ ├── e2647ef [exp-ee8a4] │ 05:14 PM     │ 0.23146 │ 0.9145 │ 10           │ 64               │
+│ ├── 15c9451 [exp-a9be6] │ 05:14 PM     │ 0.25231 │ 0.9102 │ 10           │ 32               │
+│ ├── 9c32227 [exp-17dd9] │ 04:46 PM     │ 0.23687 │ 0.9167 │ 10           │ 256              │
+│ ├── 8a9cb15 [exp-29d93] │ 04:46 PM     │ 0.24459 │ 0.9134 │ 10           │ 128              │
+│ ├── dfc536f [exp-a1bd9] │ 03:35 PM     │ 0.23508 │ 0.9151 │ 10           │ 24               │
+│ └── 1a1d858 [exp-6dccf] │ 03:21 PM     │ 0.23282 │ 0.9152 │ 10           │ 16               │
+└─────────────────────────┴──────────────┴─────────┴────────┴──────────────┴──────────────────┘
+```
+
+By default it shows all the parameters and the metrics with the timestamp. If
```
dberenbaum, line 206: Could this go inside a `<details>` tag? Feel free to ignore if you think it's better in the main text.

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+
+```dvc
+$ dvc exp branch exp-05e87 "cnn-256"
+Git branch 'cnn-256' has been created from experiment 'exp-05e87'.
+To switch to the new branch run:
+
+        git checkout cnn-256
+```
+
+You can then checkout and continue from working this branch as usual.
```
dberenbaum, line 245: Maybe mention that you could also merge it? Again, not important, and I understand it's sort of implied by "working this branch as usual," but it's probably a common workflow.

dberenbaum, line 245:
```suggestion
You can then checkout and continue working from this branch as usual.
```

dberenbaum (commented):
A couple of spots may need minor clarifications, but I think it's looking really good! It was surprisingly easy to read through the full doc ❤️ .

content/docs/sidebar.json
```diff
@@ -60,7 +60,8 @@
         }
       },
       {
-        "slug": "experiments",
+        "slug": "experiments-trail/experiments",
```
shcheklein, line 63: slug defines URL which looks strange now to my mind (redundant `experiments`)

content/docs/start/experiments-trail/experiments.md
skshetry: hmm, I am not sure if the `dvc exp init` is going to be any easier than the `dvc.yaml` for other scenarios. We should not try to keep them away from dvc.yaml.

shcheklein:
Emre, do we want to include the video from the previous doc?

Also, do we want to delete the previous one and do a redirect?

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
+---
+title: 'Get Started: Experiments'
+---
+
+# Get Started with Experiments
+
+In machine learning projects, number of <abbr>experiments</abbr> grow rapidly.
+DVC can track these experiments, list their most relevant parameters and
```
shcheklein, line 8: it helps to run them also, navigate, compare ...

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+
+Please clone the project and create a virtual environment.
+
+> We create a virtual environment to keep the libraries we use isolated from the
+> rest of your system. This prevents version conflicts.
+
+```dvc
+$ git clone https://github.com/iterative/get-started-experiments -b get-started
+$ cd get-started-experiments
+$ virtualenv .venv
```
shcheklein, line 34: it won't work on Mac (need to check Windows) - you need to provide Python 3

I would simplify this with `We strongly recommend creating a virtual environment first.` - the way we handle this in the current example get started

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+
+### ⚡ Clone the project and create virtual environment
+
+Please clone the project and create a virtual environment.
+
+> We create a virtual environment to keep the libraries we use isolated from the
+> rest of your system. This prevents version conflicts.
+
+```dvc
+$ git clone https://github.com/iterative/get-started-experiments -b get-started
```
shcheklein, line 32: why do we need `-b`?

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+$ cd get-started-experiments
+$ virtualenv .venv
+$ . .venv/bin/activate
+$ python -m pip install -r requirements.txt
+```
+
+### ✅ Get the data set
+
+The repository we cloned doesn't contain the dataset. We use `dvc pull` to
```
shcheklein, line 41: here we need a reference to the data management

content/docs/start/experiments-trail/experiments.md
```diff
@@ -0,0 +1,255 @@
...
+```dvc
+$ dvc pull
+```
+
+The repository already contains the necessary configuration to run the
+experiments.
+
+## 💡 Preparing a project for DVC experiments
+
+If DVC is not installed to the system, please refer to [install](/doc/install).
```
shcheklein, line 53: we already mentioned `dvc pull` before?

content/docs/start/experiments-trail/experiments.md
casperdcl: idk - branch `get-started` doesn't even exist at https://github.com/iterative/example-dvc-experiments/branches

iesahin: It's a tag. https://github.com/iterative/example-dvc-experiments/tree/get-started

It shows the point to start this tutorial. I'm moving the tag to a particular commit instead of updating the name here.

content/docs/start/experiments-trail/experiments.md
iesahin: `virtualenv` requires a separate installation in all systems. The standard one in Py3 is `venv`. We discussed this in https://github.com/iterative/example-repos-dev/issues/26 and decided to continue with `virtualenv`. 

I can also strongly recommend, if that's the point :)

content/docs/start/experiments-trail/experiments.md
iesahin: I believe we'll revise all the docs when we'll have done other trails. All steps will link to other relevant parts.

content/docs/start/experiments-trail/experiments.md
iesahin: Editing `dvc.yaml` requires a node in the narrative towards pipelines, which we may not return to experiments easily. We are trying to come up with the simplest. The pipelines will have its own tutorial once we complete these. @skshetry

content/docs/start/experiments-trail/experiments.md
iesahin: Added. Thanks.

content/docs/start/experiments-trail/experiments.md
iesahin: Thank you :)

content/docs/start/experiments-trail/experiments.md
iesahin: Even `venv` requires a separate `apt install python3-venv` in Debian and derivatives.

content/docs/start/experiments-trail/experiments.md
iesahin: removed the section.

dberenbaum:
> A couple of spots may need minor clarifications, but I think it's looking really good! It was surprisingly easy to read through the full doc ❤️ .

@iesahin It was easy relative to the complexity of the topic and the current https://dvc.org/doc/start/experiments. It was not a comment about you 😄 . Hope I didn't offend!

content/docs/start/experiments-trail/experiments.md
iesahin: Removed the first sentence that may be causing the confusion.

content/docs/start/experiments-trail/experiments.md
iesahin: Thank you. :)

iesahin:
> 
> 
> > A couple of spots may need minor clarifications, but I think it's looking really good! It was surprisingly easy to read through the full doc ❤️ .
> 
> @iesahin It was easy relative to the complexity of the topic and the current https://dvc.org/doc/start/experiments. It was not a comment about you 😄 . Hope I didn't offend!

I took it as a compliment actually. Thank you. Writing simpler is harder. Where could I be offended? 😄

content/docs/sidebar.json
iesahin: experiments trail will also include params/metrics and the checkpoints as separate sections. this one is about the experiments proper. 

I can rename it.

iesahin:
> Emre, do we want to include the video from the previous doc?

I think a newer video might be more appropriate, but we can include it for the time being. 

> Also, do we want to delete the previous one and do a redirect?

All the content in the earlier version is already covered in the guide, and here. I think we can rename this to "experiments" (without "trail") for the time being and rename it to "experiments trail" once the others are done.

content/docs/start/experiments-trail/experiments.md
iesahin: Need a new header to this paragraph to put into a `<details>` section. Let me look at it once more.

content/docs/start/experiments-trail/experiments.md
shcheklein: okay, question - since we are not creating this repo from the very beginning like we do in the example get started one - does it make sense to let them just clone the latest version (and may be briefly explain core pieces of it)?

content/docs/start/experiments-trail/experiments.md
shcheklein: can we include it now? not even the link itself, but some phrase - that `dvc pull` is one of those commands to save, load, manage datasets and ML models ... read here ....

shcheklein:
> All the content in the earlier version is already covered in the guide, and here. I think we can rename this to "experiments" (without "trail") for the time being and rename it to "experiments trail" once the others are done.

I think it's fine to keep it trail. I think I was confused by you not deleting the previous file (hiring it only in the sidebar). Also, we need to do a redirect if we are effectively renaming the section.

Also, it should be something like - experiments-trail/start, experiments-trail/metrics, etc? `experiments-trail/experiments` looks repetitive. If possible it's always good to put the structure early since doing redirects hurts - SEO, and managing them.

shcheklein:
Still the same problems:

- expandable sections should have `### title` for them not being visible in the right hand side panel
- emojis - they are out of style for the docs - we don't use it in other places. Reason - for them not being very scalable and having "calm" default docs I think.
- expandable sections that have a structure inside - a sign that we could simplify it - it's a bit too much for those sections I think.

content/docs/start/experiments-trail/experiments.md
shcheklein: My question was about not providing so much details. We can keep `git clone` and maybe `pip install -r requirements` ... we can put a link nearby to the virtualenv like we do in the current get started?

content/docs/start/experiments-trail/experiments.md
shcheklein: I think since we don't include pipelines, and we could start this doc like:

`git clone project`
`pip install req`
`dvc pull`
`dvc exp run`

do we need to explain anything at all here?

or at least if we want to keep doin this, I would instead create an expandable section:

"Explaining structure of the project" where we explain `dvc.yaml`, put a reference to the pipelines/dvc stage add, etc.

content/docs/sidebar.json
iesahin: For the time being we can replace the current GS:Experiments page with this one. After #2856 and #2857 are done, we can rename these to "trails" with multiple links among them. 

Naming this now as "trail" will look like a trail with a single stepping stone.

content/docs/start/experiments-trail/experiments.md
iesahin: We should start somewhere in the middle, before experiments. Otherwise, the experiments repo won't contain any experiments. We are also using this repository for `dvc exp pull/push` and I believe it's better to keep this tag and start from it. The alternative, that is starting from a repo with experiments in it leads to a more complicated narrative.

iesahin:
> I think it's fine to keep it trail. I think I was confused by you not deleting the previous file (hiring it only in the sidebar). Also, we need to do a redirect if we are effectively renaming the section.

Instead of a redirect, I renamed the section for now. PTAL @shcheklein. I believe we can reorganize and put links to these documents after the other trails. 

I can revert the rename and organize the sidebar and redirects, but this way it seems simpler to me.

content/docs/start/experiments-trail/experiments.md
shcheklein: okay, makes sense ... I wonder if long term we can make this tag `get-started` the default state though fo the repo (and add experiments as "you go" in the docs)

content/docs/sidebar.json
shcheklein: sounds good to me!

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
-$ dvc exp run --run-all --jobs 2
-```
+</details>
 
-## Comparing many experiments
+## Running the experiment with default parameters
 
-To compare all of these experiments, we need more than `diff`. `dvc exp show`
-compares any number of experiments in one table:
+The purpose of `dvc exp` subcommands is to run the experiments without
```
shcheklein, line 59: without committing something to Git (here and in other places) - people who will be reading this don't know why would they need to commit anything into Git, they won't know about metrics and params.

I think there should be some general language initially:

`the purpose of the commands ... is to let you run, capture, compare one or multiple experiments at once as you iterate on your code, data, hyperparameters`

also `The purpose of `dvc exp` subcommands is to run the experiments` is not correct - there are other commands besides `run`

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
+committing parameter and dependency changes to Git. The artifacts like models,
+metrics produced by each experiment are tracked by DVC and persisted on demand.
```
shcheklein, line 61: `are tracked by DVC` - people won't know anything about it at this point

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
-│ **master**        │  **0.55259** │ **0.91536** │ **50**         │ **2**               │
-│ ├── exp-bfe64 │  0.57833 │ 0.95555 │ 50         │ 8               │
-│ ├── exp-b8082 │  0.59806 │ 0.95287 │ 50         │ 64              │
-│ ├── exp-c7250 │  0.58876 │ 0.94524 │ 100        │ 2               │
-│ ├── exp-b9cd4 │  0.57953 │ 0.95732 │ 100        │ 8               │
-│ ├── exp-98a96 │  0.60405 │  0.9608 │ 100        │ 64              │
-│ └── exp-ad5b1 │  0.56191 │ 0.93345 │ 50         │ 2               │
-└───────────────┴──────────┴─────────┴────────────┴─────────────────┘
+```dvc
+$ dvc exp run
```
shcheklein, line 67: syntax highlight is broken for `dvc exp run` cc @julieg18

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
 ```
 
-`dvc exp pull` retrieves the experiment from a Git remote.
+</details>
 
-```dvc
-$ dvc exp pull gitremote exp-bfe64
-Pulled experiment 'exp-bfe64' from Git remote 'gitremote'.
-```
+## Comparing experiments
```
shcheklein, line 146: This should go right after the `dvc exp run` ... w/o any subtitle ideally. There is a natural connection and the most important thing:

`dvc exp run` to run ... now to see your staff - check the `dvc exp show` out.

Everything else is secondary in this document and should go in some general sections like:

`Setting params`

If we want to keep that section at all it should be about `--set-param` and about general way - edit file and run.

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
-both of our metrics, with 100 estimators and a minimum of 64 samples to split a
-node.
-
-> See `dvc exp show --help` for more info on its options.
-
-## Persisting experiments
-
-Now that we know the best parameters, let's keep that experiment and ignore the
-rest.
+It runs the specified command (`python train.py`) in `dvc.yaml`. That command
```
shcheklein, line 74: I would put even all this stuff after the `dvc exp show` - it explains how (deep tech details), we need to show the idea from the user perspective (and use expandable sections to go deep if needed).

content/docs/start/experiments.md
```diff
@@ -2,293 +2,225 @@
...
+[gse]: https://github.com/iterative/get-started-experiments
 
 <details>
 
-### 💡 Expand to see what happens under the hood.
-
-`dvc exp run` is similar to `dvc repro` but with some added conveniences for
-running experiments. The `--set-param` (or `-S`) flag sets the values for
-<abbr>parameters<abbr> as a shortcut for editing `params.yaml`.
+### Installing the Example Project
```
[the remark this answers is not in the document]
iesahin: Done. Thanks.

content/docs/start/experiments.md
iesahin: I think it's better to present `dvc exp show` after a few experiments. To run _a few_, we need to specify the parameters. I'll move the `show` section after _run by setting params_.

content/docs/start/experiments.md
iesahin: Added a brief example after the first run

content/docs/start/experiments.md
iesahin: I revised this.

PR #2577, opened 2021-06-21
updating events

jendefig:
Event update.  Also changed the text in the index.tsx to just direct to subscribe, instead of saying "No Upcoming events."  Lmk if this was not the right way to do this.


content/community.json
```diff
@@ -15,7 +15,7 @@
     },
     "events": {
       "title": "Events",
-      "description": "Learn more about DVC and meet us in person! We're appearing at conferences, meetups, and talks around the world.",
+      "description": "Learn more about DVC and meet our team virtually! We're appearing at conferences, meetups, and talks around the world.",
```
shcheklein, line 18: make it `or in person`? so that we don't have to change it next time?

content/community.json
```diff
@@ -71,20 +71,12 @@
   ],
   "events": [
     {
-      "url": "https://dssconf.pl/",
```
shcheklein, line 74: I think, no need to remove anything. Just add a new event.

content/community.json
jendefig: Good idea!

content/community.json
jendefig: Gotcha!  Put it back.

shcheklein (changes requested):
Make image smaller? I think we don't need it to be full size for this also run it through the minifier.

(omg, we really need to something about us saving images into Git :) )

jendefig:
@shcheklein Reduced from a 500x500 pixel image to a 200 x200 pixel image and ran through tiny.png.  I think it should be good now. 👍

jendefig:
> @shcheklein Reduced from a 500x500 pixel image to a 200 x200 pixel image and ran through tiny.png. I think it should be good now. 👍

Actually in my local it looks a bit blurry this small.  Let me know what you think...

shcheklein:
> Actually in my local it looks a bit blurry this small. Let me know what you think...

yes, it should be actually 700x700 (size on the screen x 2 to make it retina displays compatible). So, in this case we should make it even bigger in terms of dimensions and just run through the minifier.

shcheklein (approved):

shcheklein:
Thanks @jendefig !

PR #2581, opened 2021-06-23
community page updates

jendefig:
Updates to user content and another event.  Waiting for Ingamul to redo header desktop and mobile graphics to say instead: "Could you be our next Developer Advocate?"  If that's agreeable. 🙂


shcheklein:
Looks great!

Do we need to wait for the new image to merge?

Also some suggestions:

```
1800+
registered developers
50K+
```

should we update these numbers ^^

Can we do a different image for meetups (June vs July)?

shcheklein:
> Waiting for Ingamul to redo header desktop and mobile graphics to say instead: "Could you be our next Developer Advocate?" If that's agreeable.

💯

jendefig:
> Looks great!
> 
> Do we need to wait for the new image to merge?
> 
> Also some suggestions:
> 
> ```
> 1800+
> registered developers
> 50K+
> ```
> 
> should we update these numbers ^^
> 
> Can we do a different image for meetups (June vs July)?

I updated numbers.  Not sure how the 50K was arrived at.  This number that I put was messages SENT per Discord.  Not total messages.  Even if I add Discuss for the last month, it's just 97 more.  but that may be a good thing?  We are getting more succinct at answering in support and our docs are excellent enough to reduce need for messages?  We may want to put a different metric in that place?  Issues solved?  (How to quantify?) Contributors?  GH Stars for DVC and CML?  Also it says "developers."  It seems clear that not all are developers, but users may be more accurate?

On changing the pics.  I say no if that's ok.  I will be adding an event: DVC Learn on July 14th.  (Still need to create graphic; DeeVee with just glasses, not sunglasses is the idea).  Another type of Meetup for our kicking the tires/learning audience.  The idea is to help people learn with DVC Learn and then they "graduate" to Office Hours.  Milecia will be doing these.  Just half hours.  So those graphics will look different, but this way people will be able to delineate over time the types of events visually.  Also Milecia is waiting to hear back from another Conference she is likely doing in July.  So that will be a different graphic.  How many fit here?  I haven't seen it all filled in. @rogermparent 

On the header image(s) - Let's wait a tad.  He was asleep during my request.  If it arrives tomorrow morning and we can resolve the metrics issue, we can wrap up all of this mañana.  LMK about what you think on metrics.

shcheklein:
> Not sure how the 50K was arrived at.

Probably from the statsbot: `1,825 Days: 52,652 messages` (and I'm not sure it's everything or from the point we started using the bot - we can confirm in the UI for the statsbot and approximate to the total number if we really want to go precise).

> Contributors? GH Stars for DVC and CML? 

They are not very informative unfortunately in terms of community. Contributors - hard to count since they spread across different repos and overlap, stars - not about community.

> Also it says "developers." It seems clear that not all are developers, but users may be more accurate?

Good point ... may be even members?

> On changing the pics. I say no if that's ok

Absolutely :) And thank for the explanation!

Sounds good on everything else!

shcheklein:
Looks like Dev Advocate image is too low res

jendefig:
> Looks like Dev Advocate image is too low res

Corrected this.  Should be ok now.

jendefig:
> Good point ... may be even members?

I still need to figure out how to change this...

shcheklein:
Should we fix the numbers (13K, etc)?

To find where the text is defined in the repo, try to do `git grep "developers"` or even `git grep "registered developers"`

jendefig:
> Probably from the statsbot: `1,825 Days: 52,652 messages` (and I'm not sure it's everything or from the point we started using the bot - we can confirm in the UI for the statsbot and approximate to the total number if we really want to go precise).

The wording says monthly messages on the site.  Do we want monthly or all time messages?  One or the other needs to change.  I think monthly makes more sense, but the stats bot does not give that number.  I need to research how to get that.  It would be a good thing to track.  But what we are tracking is support.  The value of this number is odd because it will forever be fighting itself.  We would want support calls to go down, but we want users to grow.  I need to think about this more.  Maybe we should have something all together different there.  

And also where do I adjust the wording on that page.  I'm not finding in community.json or the index for community?

jendefig:
Ok. @shcheklein I think this is good to go now.  I took an average of the last 3 months of messages on the Discord server.  That seemed logical.  The user content is updated as well as the events. Pics have been run through tiny.png. 👍

jendefig:
Noticed though on DVC Blog portion of the Community page - this auto populates right @rogermparent ?  Why is the Community Gems post not first in line?  Does it wait a day or something before it updates?

shcheklein (approved):

shcheklein:
Thanks @jendefig ! Great update.

Issue #654, opened 2019-09-29
cases: generalized shared dev server


jorgeorpinel:
> From https://github.com/iterative/dvc.org/pull/637#pullrequestreview-293443095

1. ~~Do we need blog-like articles in a semi-static website?~~ UPDATE: We even have a blog now...
  How about extracting the more user manual-like information from the existing use cases into user guides and write actual articles to replace our use cases? These can be published in our blog, Medium, etc. (handled like a devrel matter like any other article).

UPDATE: Jump to summary: https://github.com/iterative/dvc.org/issues/654#issuecomment-738557458

---

OUTDATED:

2. by "use case" we really mean "case studies" - these are actual (mini) guides with some steps to address specific scenarios. A "use case" is a more general problem solved by the system, they typically map directly to features in a system, so really ~~the current~~ some of the [User Guide](https://dvc.org/doc/user-guide docs are closer to actual use cases in my understanding.

   ![image](https://user-images.githubusercontent.com/1477535/65824055-ac166580-e22f-11e9-9b9a-1ec1830e17c0.png)

Related to #425; Could work on them together 🙂

jorgeorpinel:
and maybe even rename **User Guide** to **User Manual**. Same concept but a little better naming because we can call each document inside a "guide" without causing confusion with the parent section (the manual).

jorgeorpinel:
Some comments from @shcheklein:

From https://github.com/iterative/dvc.org/pull/637#pullrequestreview-293500033 (regarding the list of guides in /user-guide/index.md)):

> I hope that we'll have a proper third level soon to cluster some of these by topics. Not for this PR for sure...

From https://github.com/iterative/dvc.org/pull/637#discussion_r329336744 (regarding reorganizing "External Data" guides):

> The idea is to eventually (and I was waiting for the engine to properly support 3 levels) - is to have one single Manage External Data second level section, that will include External Outputs, External Dependencies, External Cache, Shared Cache (or a better title to explain cases like #565, #455), and other related articles.

From https://github.com/iterative/dvc.org/pull/637#discussion_r329337401:

> Regarding the User Guide - there are sections like "Large Files Optimization" - it's not a use case by itself, it's a technical description how to fine fine tune DVC in certain situations. I think UG should include things like `Pipelines -> How to Edit, How to Create`, etc. And Updating Tracked Files is very similar in my mind to those.

dashohoxha:
[...]

PR #2482, opened 2021-05-16
how-to: setup a shared cache (extracted from use cases)

jorgeorpinel:
REPURPOSED to close #2487

---

Would close (no longer) https://github.com/iterative/dvc.org/issues/654#issuecomment-738557458
And (not yet) #820

- [ ] Consider comments from https://discuss.dvc.org/t/multiple-users-in-one-repo/746 (common file permission questions)
- [x] Work on or prioritize #2487 after this.

PR #2587, opened 2021-06-28
cases: Fast and Secure Data Caching Hub

### LINKED ISSUE #820: use-cases: revise so they're more high level "landing pages"

jorgeorpinel:
Closes https://github.com/iterative/dvc.org/issues/654#issuecomment-738557458
Closes #820 (last item)

- [x] Consider comments from https://discuss.dvc.org/t/multiple-users-in-one-repo/746 (common file permission questions)

TODOs:

- [x] Get concept and scope right
- [x] Iterate on content
- [x] Finalize name (see https://github.com/iterative/dvc.org/pull/2587#issuecomment-892240166)
- [x] https://github.com/iterative/dvc.org/issues/2677


redirects-list.json
```diff
@@ -31,6 +31,7 @@
   "^/doc/tutorials(/.*)?                                                                  /doc/start",
 
   "^/doc/use-cases/data-and-model-files-versioning/?$                                     /doc/use-cases/versioning-data-and-model-files",
+  "^/doc/doc/use-cases/shared-development-server$                                         /doc/user-guide/how-to/share-a-dvc-cache",
```
jorgeorpinel, line 34: This is contingent on #2482 being merged first.

content/docs/use-cases/managing-datasets.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Problem**: managing datasets can be a mess. Are file names important? Where do
+you store them? How do you keep track of data [revisions][versioning]? How can
+they be shared and who can access them? How should multiple projects reuse them?
+Can they be published or plugged into production systems?
+
+> [Data lifecycle](https://www.usgs.gov/products/data-and-tools/data-management/data-lifecycle):
+> Plan > Acquire > Process > Analyze > Preserve > Publish/Share
+
+**Solution**: no ad-hoc file names, handle large files and directories (in Git),
```
jorgeorpinel, line 11: Here's that first draft we talked about @shcheklein.

jorgeorpinel: Thouoghts?

content/docs/use-cases/managing-datasets.md
```diff
@@ -0,0 +1,38 @@
...
+smart <abbr>project</abbr> object storage ("cache") with file-level
+deduplication (local or shared), minimize data transfers, integrate cloud or
+on-premises storage (bring-your-own authorization and sharing config).
+
+Standardized patterns, e.g.:
+
+- Multiple users work on the same shared server.
+- A centralized data storage unit or cluster
+- There's a single environment with access to production data.
+- [Data Registries](/doc/use-cases/data-registries)
```
jorgeorpinel, line 21: Working on improving these scenarios for now ⌛

content/docs/use-cases/managing-datasets.md
shcheklein: We went into data versioning and dataset registries?

This should be about efficient, performant data storage.

Let's coin a term here (like we did with Datasets Registry)? 

E.g.

Data Cache Layer (I would not worry about people mixing Cache and dvc cache - they are not familiar with that DVC implementation detail)
Datasets Cache
Datasets Cache Layer

I would think in this direction ^^

This what is it about. We could use a diagram where we show remote, cache (mention that is it on NAS, Ceph, EFS, etc), clients attached vs regular setup where cache is per client.

jorgeorpinel: > This should be about efficient, performant data storage.

OK how about title **Effective Dataset Storage**?

*Dataset Cache Layer* sounds nice too but seems indirect/ hard to relate to. In fact I'm not sure I get how it is some kind of layer. Between what? (Storage systems and your project I guess?)

jorgeorpinel: (Updated title and intro for now)

content/docs/use-cases/managing-datasets.md
jorgeorpinel: Updated scenarios too.

content/docs/use-cases/managing-datasets.md
shcheklein: Between remote storage and your projects. Yes. It is _cache_ by definition (that's why it's called this way in DVC). But when we make it global we kinda make it an explicit part of your infrastructure and it becomes Objects Cache, Cache Layer, etc. Similar to the regular Datasets Storage that can mean S3 remote (and that's why Datasets Storage is nice, but I'm hesitant a bit since it's not enough to differentiate it from remotes, and it's not enough to emphasize the goal - faster access - that's what any cache is about - more expensive storage, smaller, faster).

jorgeorpinel: OK I think I got this part now.

content/docs/use-cases/data-storage-layers.md
```diff
@@ -0,0 +1,45 @@
...
+**Scenarios**:
+
+- You need to design a system for multiple users collaborating on the same
+  datasets.
+- Communicating with your existing storage (e.g. Amazon S3) takes too long. Can
+  you introduce a local storage cache layer for faster access?
+- You want to move data from your machine to an external drive without losing
+  the ability to use it locally.
+- Your team shares access to a GPU server to train machine learning models, how
+  do you connect different datasets to run [experiments]?
```
jorgeorpinel, line 22: @shcheklein sorry for the delay. PTAL at this new intro, figure, and scenarios.

content/docs/use-cases/data-storage-layers.md
```diff
@@ -0,0 +1,59 @@
...
+
+**Scenarios**:
+
+- You need to design a system for multiple people working on the same datasets.
+- Transfers to/from cloud storage (e.g. Amazon S3) takes too long. Can you
+  easily introduce a local cache layer for faster access?
+- You want to move data from your machine to an external drive without losing
+  the ability to use it locally.
+- Your team shares access to a GPU server for machine learning [experiments].
+  How do you connect different datasets?
```
jorgeorpinel, line 21: I've updated the intro and scenarios (again). PTAL 👀

content/docs/use-cases/data-storage-layers.md
```diff
@@ -0,0 +1,48 @@
...
+configure a [shared cache](/doc/user-guide/how-to/share-a-dvc-cache) among
+multiple <abbr>projects</abbr>, which compounds the deduplication and
+performance benefits. Accessing the data is as fast as the location where you
+set this common storage, e.g. a network-attached storage (NAS) or an SSH server
+in the local network.
+
+Going a step further and making DVC storage global for all your teams' projects,
+it can become an independent layer in your infrastructure. For example, you can
+use Ceph, EFS, cloud object storage, etc. and setup your own access controls as
+supported by that platform.
```
jorgeorpinel, line 46: I've also written a first version of the solution part. I'm still to mention remote storage (⌛) but this should give a good idea of where I'm going with the text.

content/docs/use-cases/data-storage-layers.md
```diff
@@ -0,0 +1,57 @@
...
+[EFS](https://aws.amazon.com/efs/), cloud object storage (S3,
+[GCS](https://cloud.google.com/storage/), Azure, etc.), or any other platform;
+and manage _data security_ policies there.
+
+![]() _Data storage as infrastructure_
+
+To balance speed and cost, you can introduce a secondary storage layer with a
+`dvc remote`. Data associated with older versions of your project that you won't
+need often can be kept or backed up there. Remote storage is also a great way to
+[share datasets and ML models](/doc/use-cases/sharing-data-and-model-files).
```
jorgeorpinel, line 55: @shcheklein here's the full 1st version of the "solution" part (including remote storage).

jorgeorpinel: BTW I'm still trying to add 2 more parts/sections: advanced patterns (only remote storage as primary storage i.e. an external cache in a remote location); and garbage collection. And of course an example (shared dev server, so we can keep that somewhere since users actually implement it). But I'll hold off on those until we've done another review round here.

content/docs/use-cases/data-storage-layer.md
```diff
@@ -0,0 +1,66 @@
+# Data Storage Layer
+
+Datasets used in data science tend to exceed the capacity of typical storage and
+computing environments. What infrastructure do you need to work with large
+files, and can your team utilize it effectively?
+
+For example, you may be looking to speed up data transfers from a massive object
```
jorgeorpinel, line 7: PTAL (mainly at the first half) @shcheklein . Thanks

jorgeorpinel: OK actually most of it is ready for review.

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
...
+efficient storage layer globally — for all your team's projects. This approach
+can help in situations such as:
+
+- You need to speed up data transfers from a massive object store on the cloud,
+  or to move a growing data store out of your machine without slowing things
+  down. How do you plug in something in the middle (like a NAS drive) to
+  existing projects?
+
+- Upgrading your storage platform is expensive. Can you pay only for fast access
+  to the data sets that are needed regularly?
```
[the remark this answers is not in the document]
dberenbaum: Related: should EFS be in the remote column of the image above? Seems more likely that EFS will be the cache layer than the remote layer.

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
...
+
+[file links]:
+  /doc/user-guide/large-dataset-optimization#file-link-types-for-the-dvc-cache
+
+To unify data storage across multiple projects, you can configure a [shared
+cache]. This compounds the deduplication and performance benefits. Having
+separated data storage from other aspects of your projects, you can treat it as
+an independent infrastructure layer, using Ceph,
+[EFS](https://aws.amazon.com/efs/), object storage (e.g. S3,
+[GCS](https://cloud.google.com/storage/), Azure), or any other platform.
```
dberenbaum, line 49: Sure, it's possible to use object storage for your shared cache, but is this really typical? It seems we are mixing the same technologies between the cache and remote concepts, which makes it hard for me to understand the suggested patterns. What do you think about moving object storage (and the example types) down to the next paragraph where remotes are mentioned?

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
...
+projects acquire the same data, creating duplication and increasing cost. Is
+there an effective way to perform this process?
+
+![](/img/storage-layers.png) _Data storage infrastructure_
+
+DVC's built-in data <abbr>caching</abbr> enables you to implement a simple and
+efficient storage layer globally — for all your team's projects. This approach
+can help in situations such as:
+
+- You need to speed up data transfers from a massive object store on the cloud,
```
dberenbaum, line 15:
```suggestion
- You need to minimize data transfers from a massive object store on the cloud,
```
Thoughts on this? Current wording makes me think DVC will somehow increase network speed.

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
...
+be kept there. Remote storage is also a great way to back up and
+[share datasets and ML models](/doc/use-cases/sharing-data-and-model-files).
+
+<!-- Advances scenarios? 1. No storage; 2. Only remote -->
+
+## Example: Shared Development Server
+
+Some teams may prefer using a single shared machine to run their experiments.
+This allows better resource utilization, such as GPU access and centralized data
+storage. Each person can work in their own separate <abbr>workspace</abbr> as
```
dberenbaum, line 65: Is this sentence redundant with the info above now?

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
+# Datasets Cache
```
dberenbaum, line 1: Why "datasets" instead of "data"? To follow the structure of "Versioning Data and Models" and "Sharing Data and Model Files", what about "Caching Data and Models"?

content/docs/use-cases/datasets-cache.md
```diff
@@ -0,0 +1,107 @@
...
+
+Datasets used in data science tend to exceed typical storage and computing
```
dberenbaum, line 3: Should we limit to data science? How is computing capacity relevant (maybe network capacity would make more sense)? From the rest of the paragraph, it seems like the problem is not exceeding capacity, but reducing time, cost, and complexity of storing and accessing data.

dberenbaum (commented):
The content looks great!

One comment: I know this is the hardest part that you have probably already struggled with the most 😬, but I struggled to understand the use case from the title and first sentence.

content/docs/use-cases/datasets-cache.md
shcheklein: I can give my 2cs on the naming (and maybe we can brainstorm how to make to more explicit by changing the title and/or improving the first paragraph a bit). This is indeed the most tricky and the most challenging part of the use cases usually :)

My take on this use case is that by making DVC cache and shared global we are making it an explicit infrastructure element (along with training machines or cluster, along with regular storage that is usually S3 or something, etc). It gives clear benefits, goals, lifecycle, connections to other infrastructure, etc. In this sense, it should be more about organizing infrastructure, then workflow. 

That's why probably names like "Data Cache Layer", "Datasets Cache" (pretty much the same, but domain-specific) resonate a bit better with me than "Caching Data and Models". The latter for me also covers the simple project level cache for example. And may be that's also fine, btw - we can expand and cover the local project cache as one way to do this! 

I hope it makes at least some sense. Happy to explore this together :)

jorgeorpinel: > Why "datasets" instead of "data"?

Yeah probably "Data" is better @dberenbaum. The trickier part is the term "cache" I think, especially readers already immersed in DVC concepts (like us). Try to read this from the PoV of a new user. "Caching" has a general meaning (not necessarily DVC cache), i.e. an intermediate storage layer that's faster but limited (smaller, more expensive). Regardless, I'll try to disambiguate for existing users (same applies for term "remote").

Also, I brainstormed on this with @iesahin and some other ideas we had are: Common Data Cache, Frequent-access Cache, Fast Data Access, Multiple Storage Layers. An interesting principle to include (maybe implicitly): "20% data used 80% of the time"

> it should be more about organizing infrastructure, then workflow

Yes I am also improving on this, to make it clear that we are presenting a pattern in this use case, not a specific DVC project setup (again tricky due to ambiguous terms).

> "Caching Data and Models" also covers the simple project level cache for example... we can expand and cover the local project cache as one way to do this!

Let's pin that for now 😅

dberenbaum: Yup, I know a lot of thought was put into it, which was why I felt like it was necessary to comment. Could just be my bias. Will be interested to hear opinions of others.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: > How is computing capacity relevant

One of the scenarios is that a team shares a GPU server. You can plug an external cache shared in local envs with the training env too. That was my thought process but I can remove it from here and just leave the scenario bullet if it's distracting.

jorgeorpinel: > maybe network capacity would make more sense

Good idea, I'll use that for now.

> seems like the problem is not exceeding capacity, but reducing time, cost, and complexity

I reordered the sentences to make it more logical.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: > Current wording makes me think DVC will somehow increase network speed

Implementing the pattern we're suggesting will increase speed. Not of the network of course but by bringing the data closer. The suggestion works too but I think "speed" is a keyword in this doc.

jorgeorpinel: Changing to "from a massive object which is currently on the cloud" for now

content/docs/use-cases/datasets-cache.md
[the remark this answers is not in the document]
jorgeorpinel: > Is this a separate question from the first bullet?

This one focuses on cost.

> example for this one to make it less abstract

The main scenarios are intentionally generic 😬 

> a lot like plugging something in the middle to speed up data transfers again

Yes, the solution is the same.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: > it's possible to use object storage for your shared cache, but is this really typical?
moving object storage (and the example types) down

Good point. I guess it wouldn't bring it closer. Moved ⬇️ 

> we are mixing the same technologies between the cache and remote concepts

This is the case though. There's no clear line.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: It sort of recaps as an intro to the experiment, yep. Also, mentions "resources" specifically (possibly a keyword).

jorgeorpinel: I mean example 😆

content/docs/use-cases/datasets-cache.md
jorgeorpinel: Went for **Fast Data Storage Layer** for now 😋

content/docs/use-cases/datasets-cache.md
dberenbaum: Okay, computing capacity makes more sense once you explain it, so up to you.

jorgeorpinel: I like the networking idea better. Computing is implied in that scenario bullet about the GPU server. Thanks!

content/docs/use-cases/datasets-cache.md
jorgeorpinel: p.s. @shcheklein I think naming and intro, as well as improving the figures are the 2 major things missing. PTAL

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,143 @@
...
+
+Using DVC to implement a layered data storage system also gives you flexibility
+to redesign the solution in the future without having to change directory
+structures or code.
+
+<details>
+
+## Click to explore a range of advanced implementations
+
+- On one extreme, you can [prepare your data requirements] in a lightweight
```
jorgeorpinel, line 72: This chunk is a) a conclusion and b) a last bit about going from Layer to Layers and advanced possibilities for this pattern. May needs work still (or removing)

jorgeorpinel: p.s. covers the simple cache project scenario mentioned in https://github.com/iterative/dvc.org/pull/2587#discussion_r671570947.

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,143 @@
...
+
+> ⚠️ It's not recommended to share a single workspace among users, as it may
+> produce file permission issues.
+
+Let's say you are cleaning up raw data for later stages:
+
+```dvc
+$ dvc add raw
+$ dvc run -n clean_data -d cleanup.py -d raw -o clean \
+          ./cleanup.py raw clean
```
jorgeorpinel, line 106: I left `run` since it's shorter than `stage add` and then `repro`. Or should I migrate these samples?

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,143 @@
+# Fast Data Storage Layer
+
+Datasets used in data science tend to exceed typical storage and networking
```
casperdcl, line 3: you mean "stretch the limits of" rather than "exceed?"

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,143 @@
...
+$ dvc add raw
+$ dvc run -n clean_data -d cleanup.py -d raw -o clean \
+          ./cleanup.py raw clean
+
+$ git add cleanup.py raw.dvc dvc.yaml dvc.lock .gitignore
+$ git commit -m "Cleanup raw data"
+$ git push
+```
+
+> 📖 Learn more about `dvc add` and `dvc run`.
```
casperdcl, line 113: broken link?

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,143 @@
...
+
+```dvc
+$ git pull
+$ dvc checkout
+A       raw
+A       clean
+# ... work on processing clean data ...
+
+$ dvc run -n process_data
+          -d process.py -d clean -o processed
```
casperdcl, line 129:
```suggestion
$ dvc run -n process_data \
          -d process.py -d clean -o processed \
```

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: I was thinking straight up exceeding but do you think stretch the limits of reads better? I'm fine with either one.

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: Autolinks 🙂 

image

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: Good catch, thanks!

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: p.s. I wonder if these still work on the docs engine version of cml.dev ?

casperdcl: yes but not sure if the tooltips (which I view as *asides* rather than real links) are enough to justify this awkward floating sentence. Would prefer if \`dvc add\` and \`dvc run\` appeared in-text further up.

content/docs/use-cases/datasets-cache.md
iesahin: What about "Joint Cache Layer" or "Joint Storage Layer" for the name?

iesahin: What about "Joint Cache" or "Joint Data Layer"?

content/docs/use-cases/fast-data-storage-layer.md
iesahin: I think we should move all `dvc run` to `dvc stage add`, and not introduce new uses.  It will probably be gone in 3.0

content/docs/use-cases/fast-data-storage-layer.md
casperdcl: well technically if it "exceeds" it wouldn't be "used" since it would be "unusable" right?

content/docs/use-cases/datasets-cache.md
casperdcl: Intermediate/Near Cache?

jorgeorpinel: "Joint" typically refers to 2 things (e.g. 2 parties working together). So still no perfect title... 🤷

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: Is it deprecated? Idk

jorgeorpinel: It's not officially deprecated but I guess most of us agree it should be, so updated.

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: Not necessarily, it just implies data science usually can't happen on typical computing systems.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: "Intermediate" and "Near" can be implied by "Cache", I think. It's a tricky one

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,71 @@
...
+directories.
+
+Now that your team shares a primary storage, it can be managed independently as
+part of your infrastructure; provisioned depending on data access speed and cost
+requirements. You have the flexibility to change storage providers at any time,
+without having to change the directory structures or code of your projects.
+
+> DVC can also manage [remote storage](/doc/command-reference/remote) on
+> platforms like Amazon S3 or Google Drive, in order to
+> [share datasets and ML models](/doc/use-cases/sharing-data-and-model-files) or
```
dberenbaum, line 42:
```suggestion
> [share datasets and ML models](/doc/use-cases/sharing-data-and-model-files)
```

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,71 @@
...
+  storage platform is expensive).
+
+- de-duplicate files automatically when multiple people are working on the same
+  data (for example on a
+  [shared development server](#example-shared-development-server)).
+
+- switch data inputs quickly (without re-downloading) on a shared GPU server for
+  machine learning [experiments](/doc/user-guide/experiment-management).
+
+![](/img/storage-layers.png) _Data storage middleware for multiple projects_
```
dberenbaum, line 27: I think I mentioned this last review, but the "Remote" column here includes some confusing items like SSH and EFS. IMO this column should repeat the "External" column above - SFTP, S3, GCP, Azure (and maybe the columns should be named the same in both diagrams).

Also, should the caption include "fast data storage layer" to make it clear that this is what the title refers to?

dberenbaum (approved):
LGTM! Just a couple minor comments on the second diagram, but I think it's in good shape!

content/docs/use-cases/fast-data-storage-layer.md
iesahin: "tend to exceed" ≠ exceed. This is good, but "stretch the limits" may be better.

content/docs/use-cases/datasets-cache.md
iesahin: I think "fast" is more misleading than "joint." It emphasizes the speed. We don't know if their connection in LAN is _that_ faster than the cloud. 

I think "joint" having 2 parties is OK. We are discussing to share the cache in a LAN and this implies there are at least 2 parties. 

Some more: 

- Local Shared Cache
- Local Joint Cache
- Local Common Cache
- Local Intermediate Cache
- Warm Cache (vs. Cold Cache)
- Neighbor(ing) Cache (vs. Remote Cache)
- Converging Cache (vs. Diverging Cache)
- Nearby Cache
- Adjacent Cache
- Adjoint Cache (vs. Disjoint Remote / Cache)

iesahin: I think we can describe hot / warm / cold cache layers, hot being local to machine, warm being within the same LAN, cold is in the cloud.

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: I think that "stretch the limits" implies it never actually exceeds though, no? But that's an important scenario here actually, e.g. you simply don't have enough space in your local machine.

iesahin: I like when it turns into a semantics discussion :) Both of them are fine with me 👍🏼

jorgeorpinel: Haha yeah minor thing. I like this way better but no strong opinion, of you both vote for "stretch" I'll change it.

content/docs/use-cases/datasets-cache.md
jorgeorpinel: > "fast" is more misleading than "joint." It emphasizes the speed.

Yep, that's the emphasis we want to give! I think it's OK if we are clear in the content that the speed comes from making an intermediate storage layer that is near the local env (in the local network or VPN for example).

> I think "joint" having 2 parties is OK.

Joint usually refers to exactly 2 things working on something specific together. What are the 2 things and what are they doing together? Maybe I just don't understand your idea.

> Warm/Neighboring/Near/Adjacent

Those terms could be 👍 

> hot / warm / cold cache layers

I don't want to call everything "cache". We're keeping it more general with the term "storage". Remote storage is not a cache for example. And what would "hot cache" be? There's just one cache I think (the "warm" one). Local storage is not even needed (links are used).

content/docs/use-cases/fast-data-storage-layer.md
jorgeorpinel: > IMO this column should repeat the "External" column above

Agree. I was just lazy to update that image haha. Will take care in https://github.com/iterative/dvc.org/issues/2677 (noted your comments there).

> should the caption include "fast data storage layer" to make it clear

Hmmm maybe. I hope it can be pretty evident though. Maybe just "fast layer" in the bottom (green text currently)?

jorgeorpinel: UPDATE: I did update the 2nd figure now (still a draft though).

jorgeorpinel: Also, I'm not sure that "External" is the best term in the 2nd image... The "Near" layer can be external too. Maybe should be "Remote" in both figures?

content/docs/use-cases/datasets-cache.md
iesahin: > Joint usually refers to exactly 2 things working on something specific together. What are the 2 things and what are they doing together? Maybe I just don't understand your idea.

I'm looking at the definition and doesn't see "exactly two". Adjective definition says "two or more". 



We have two or more systems sharing a single ("near", "fast", whatever) cache or storage. "Joint" means "shared" in this context.

jorgeorpinel: You're right, it's 2+ as an adjective. But IMO "joint" is still not great because it refers to things working together as one and here we're separating a storage patter into layers. Each one is independent to some degree, although DVC makes them work together but that's not the main point. They don't work together per se.

I'll summarize names in a separate comment for everyone to weight in.

jorgeorpinel:
@shcheklein gave an OK here too (please approve if possible) so on our side just the use case's naming is left to finalize (I'll work on the figures with @jendefig). Summary:

Key ideas: DS-specific storage best practices; Speed up/ minimizing data transfers; Prevent data storage duplication; Fast/cheap dataset storage outside the local env; Bringing mass cloud storage closer; Securing data in a single location.

Concepts: Data caching (in general, not .dvc/cache); Separate data management from DS workflow with dedicated infrastructure (and DVC); Primary (warm) vs. secondary (coldish) storage; Easy sharing storage access

Some possible titles:
- Data (and Models) Storage Cache
- Data(sets) Cache Layer
- Global/Common Data Cache
- Fast Data Storage/Caching Layer  <-- *current* (with "Storage")
- Layered (Data) Storage Architecture
- Fast and Secure Data Caching

Other possible keywords to mix-and-match: Intermediate, Near, Common, Warm, Neighboring, Adjacent

@dberenbaum @iesahin @casperdcl thanks

content/docs/use-cases/datasets-cache.md
jorgeorpinel: https://github.com/iterative/dvc.org/pull/2587#issuecomment-892240166

jorgeorpinel:
Oops I forgot to mention a similar solution found in [MinIO Gateway](https://docs.min.io/docs/minio-gateway-for-s3.html). Maybe we can use some of that terminology (**bolded** below) and/or write another special example using this.

> MinIO S3 **Gateway** adds MinIO features like MinIO Console and disk caching to AWS S3...
>
> ### MinIO Caching
> MinIO **edge caching** allows storing content closer to the applications. **Frequently accessed** objects are stored in a local disk based cache.

_From [this SO question](https://stackoverflow.com/questions/68559059) involving setting "min.io as **hub** between AWS.S3 and DVC"._

casperdcl:
based purely upon % of readers who stand a chance of understanding what we mean, "Intermediate Cache" is the only viable option to me. It is universally understood to mean a) special and b) "in-between". And we don't need to include "data" or "models" since it seems needlessly verbose...

casperdcl:
AFAIK this article is specifically about something which is *not* fast/high-perf *nor* slow, but in-between?

shcheklein:
I would still vote for `Fast and Secure Data Caching` or `Fast and Secure Data Cache` ... (I hope that we included some note about security at the moment)

jorgeorpinel:
Thanks for all the feedback! So I think "**Caching**" is definitely wanted at least. ("Intermediate" could be implied in "cache" here?)

Agree about "data/models" or even "storage" making it verbose but we want use cases to be landing pages, so on the search result title probably need some general terms for context (see all the other use case titles).

> Edge computing, this is essentially what's happening

That typically involves some kind of server (proxy) running things and DVC is actually working locally. From the MinIO example I like "**gateway**" or similar ("hub, bridge") somewhat.

> this article is specifically about something which is not fast/high-perf nor slow, but in-between?

The remote/external layer is optional actually so the global cache doesn't have to be in between, just **nearby**. Maybe I still need to clarify the text if that is not clear?

Performance is key but "high performance" sounds like the software itself is especially fast or that the network throughput is improved ("fast" may have the same problem BTW). Maybe "Optimized" instead?

Optimized Data Caching Layer/Bridge
Data Storage Gateway

Using **Fast and Secure Data Caching** for now. And yes there's a note on security:

image

content/docs/use-cases/fast-data-storage-layer.md
```diff
@@ -0,0 +1,78 @@
...
+  for machine learning [experiments](/doc/user-guide/experiment-management).
+
+![](/img/storage-layers.png) _Data storage middleware for multiple projects_
+
+You can have a single storage for all you projects by setting up a [shared DVC
+cache] in a near location (network, external drive, etc.). This de-duplicates
+files across datasets and prevents repetitive transfers by
+[linking](/doc/user-guide/large-dataset-optimization) your working files and
+directories. Data security policies can be implemented reliably, as data never
+leaves the central storage. DVC can also help you back up and
```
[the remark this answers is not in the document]
jorgeorpinel: They are separate entities. The security note refers to the near layer specifically. But indeed any external layer (remote storage, optional BTW) would be a backup i.e. another copy so the data can exist in more than one place. I guess in that case you need to handle security twice and I should clarify the benefit. Thoughts @shcheklein ?

casperdcl:
I think:

- "Cache" needs to be in the title
  - also some adjective needs to appear before "Cache" to signify that it's a *special* case. The best I can think of is "Intermediate Cache"
  - alternatively, a verb: "Optimi[z]ing Cache"
- "Edge Computing" should be mentioned with a citation in the text
- maybe include "Edge" in the title?

"Optimi[z]ing and Securing Caches with Edge Computing Principles" seems a bit of a mouthful though.

iesahin:
I agree with @casperdcl that "Cache" is better in the title. 

The word "optimize" has MBA connotations for me. Whether this is good or bad is up to the audience. 

When we say "secure", it brings a whole new set of expectations. I'm not sure if we can meet those in the text. 

I'd propose "Fast Intermediate [Data] Cache" or "Fast Intermediate Caching Hub." (It can also be a model cache, so I'm not sure about "Data" being in the title.)

jorgeorpinel:
> "Edge Computing" should be mentioned with a citation in the text
maybe include "Edge" in the title?

I do like the concept of [Edge Storage](https://www.seagate.com/solutions/edge/edge-storage/) but I'm not sure it's 100% appropriate or whether we want to complicate the text with such specialized jargon. Maybe

> MBA connotations

Master of Business Administration?

> When we say "secure", it brings a whole new set of expectations

It's a bit forced but required by @shcheklein due to strategic needs.

> title based on probable search wording

No need to limit it that way too much. Search engines will look at the whole contents. We'd need a SEO specialist to do this properly anyway. OK to coin our own terms for now I think. We do consider what we remember from user support cases when writing use cases though (like the MinIO example).

---
So how about

**Secure Data and Model Caching Hub**
("and Model" skipped in nav entry)

?
- [x] emphasis on security
- [x] "Caching" implies intermediate and fast/cheap
- [x] "Data and Model" provides context for landing page purpose
- [x] "Hub" implies it can serve multiple projects and makes it seem special/intriguing (hopefully)

casperdcl:
> what would they type into Google?

exactly. I'd type "DVC remote slow speed up fast cache optimise"

jendefig:
> Secure Data and Model Caching Hub

I like this! Nailed it! Present participle for the win! 🎉 🙌

jorgeorpinel:
> I'd type "DVC remote slow speed up fast cache optimise" or "DVC secure cache" 

We're not necessarily targeting existing DVC users but rather new users. DS/ML people looking to improve their data storage and transferring processes in general.

jorgeorpinel:
@shcheklein @iesahin @casperdcl please if you can take a look at the 3 newly produced figures in https://dvc-org-cases-mgn-dsets-fbikyq.herokuapp.com/doc/use-cases/secure-data-caching-hub (thanks @jendefig).

iesahin:
Figures look great @jorgeorpinel @jendefig

casperdcl:
Are we all agreed on the following?

- there are 2 categories of target audience. To catch both,
    1. title requires "Secure" or synonym
    2. title requires "Speed/Optimise" or synonym

Also IMHO this is not about SEO (whether it shows up in my search results). It is purely about being misleading (would I bother clicking on it assuming it was top of my search results).

jorgeorpinel:
> ii. title requires "Speed/Optimise" or synonym

I don't agree on this. I think "Caching" implies speed.

jorgeorpinel:
BTW all the images are updated now! Will show during meeting

content/docs/use-cases/secure-data-caching-hub.md
```diff
@@ -0,0 +1,78 @@
+# Secure Data and Model Caching Hub
```
shcheklein, line 1: let's not make it about models I think

this layer is important primarily to optimize datasets usage

shcheklein: minor: `Hub` - do we need this new term? I'm fine to keep it. just doesn't feel like it's needed

shcheklein: by using Secure exclusively we are overemphasizing it too much now ... it become specifically about "security" now

shcheklein: `Caching Hub` can be replaced with `Cache` I think?

iesahin:
After @shcheklein 's review, for the title: "Fast and Secure Data Cache"?

content/docs/use-cases/secure-data-caching-hub.md
jorgeorpinel: > let's not make it about models

OK, removing "and Models".

> Hub - do we need this new term?
Caching Hub can be replaced with Cache

Could be "Layer", or nothing. Layer could refer to the infrastructural aspect of this patter but it's not clear until you actually read the contents. What I like abut Hub is that a) come people may get that it's infrastructure-related **and** it implies multiple parties (people or projects in this case) can employ the same data caching mechanism.

> by using Secure exclusively we are overemphasizing it too much now

Confused. So don't include "Secure"?

jorgeorpinel: > by using Secure exclusively

Ah I see so you mean because it was "Secure and Fast" before. Well since "and Models" is gone now, we can go back to "Fast and Secure" but fast has some misleading connotations (e.g. improving network throughput)... 🤔

shcheklein:
`Caching Hub vs Cache`

No strong opinion on this one - I see pros and cons for both, I think it's fine to go with Hub.

content/docs/use-cases/secure-data-caching-hub.md
```diff
@@ -0,0 +1,78 @@
...
+
+Datasets used in data science tend to exceed typical storage and networking
+capacities. Storage needs expand rapidly as more people acquire the same data,
+creating duplication (increasing cost). Valuable time is wasted waiting for
+downloads in each environment.
+
+![](/img/dataset-copies.png) _Users wait for repeated transfers and produce
```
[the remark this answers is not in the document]
jorgeorpinel: Communicated with the designer. Will report back when we have updates ⏳

jorgeorpinel:
UPDATE: FYI we're still finalizing the images but they're pretty close already. See https://dvc-org-cases-mgn-dsets-fbikyq.herokuapp.com/doc/use-cases/fast-data-caching-hub:

image

image

jendefig:
> UPDATE: FYI we're still finalizing the images but they're pretty close already. See https://dvc-org-cases-mgn-dsets-fbikyq.herokuapp.com/doc/use-cases/fast-data-caching-hub:
> 
> image
> 
> image

There needs to be a party parrot/blob in Github. 😉

jorgeorpinel:
I vote for a "run for your life parrot"

image

jorgeorpinel:
So this was approved and the images have been worked on so I merged it! Please see the final result in https://dvc.org/doc/use-cases/fast-data-caching-hub and let me know if you have additional feedback. Thanks

PR #2590, opened 2021-06-30
Fix right sidebar to scroll with main page

cjhooker:
This addresses the first item. It makes the right sidebar headings list to scroll along with the main page.

I removed `isScrollToCurrentHeadingHappened` logic. Not sure what the intention of this was, but it was causing the sidebar to only scroll once and not every time the current heading in the main page changed.

`scrollIntoView` had some weird behavior so I am setting `scrollTop` instead.


jorgeorpinel:
Thanks! Looks good from quickly trying on https://dvc-org-sidebar-scroll--pwaktw.herokuapp.com/doc/command-reference/remote/modify#available-parameters-per-storage-type. Idk about the implementation details but will ask someone else to please check that 🙂

cjhooker:
Per @shcheklein I need this PR to be in a private clone, so I am closing this PR.

PR #2605, opened 2021-07-03
scroll sidebar follows along with main scroll

kangzeroo:
## Problem
This PR tackles https://github.com/iterative/dvc.org/issues/1333. The problem is explained in [this video.](https://share.vidyard.com/watch/eGwnBPBjWmyAF2Akp7BBmb)
Basically the right panel of anchor tags doesn't scroll with the page. Thus when readers scroll down to the later parts of the page, the right panel fails to provide value in showing them where they are in the anchor tag table of contents.

DVC_Core_Project___Data_Version_Control_·_DVC

## Solution
We force the right sidebar to scroll with the page so that readers always know where they are. 
At first I used a `<HTMLDivElement>.scrollIntoView()` based solution, but that was not smooth in Firefox because it affected the whole screen. I switched to a `<HTMLDivElement>.scrollTop` based solution, which works smoothly in Chrome/Firefox/Safari. 

[Video Demo](https://share.vidyard.com/watch/fNKZuv7hzco8Ki8kgfh8jJ?)

## Notes:
- The scroll function is debounced for kinder browser performance. You can further increase the debounce duration, but this comes at the cost of the scroll bar appearing to "lag" slightly. The current 100ms is fine as the page is mostly static and isn't scare of processing resources
- On Safari, the very top of the page will not highlight the `How to report a problem` link, which is also the previous behavior in production. Do we want to push further and get that minor part working perfectly too? How does that fit into our triaged list of priorities?
- On Chrome/Firefox/Safari, if you scroll up very very fast, the right panel scrollbar won't keep up. However, you have to try quite hard to make this bug appear. For the majority of users they will not see this. Do we want to push further and get that minor part working perfectly too? How does that fit into our triaged list of priorities?
- Total time, 1.5 hours

## Testing Instructions
0. We will be replicating [these test steps](https://share.vidyard.com/watch/fNKZuv7hzco8Ki8kgfh8jJ?)
1. Visit localhost link http://localhost:8000/doc/user-guide/contributing/core
2. Resize the page shorter to force a smaller right panel (this gives us more scroll bar to see in action)
3. Scroll down the page and watch the right panel scroll with you to show the latest highlighted link
4. Scroll up the page to see the same behavior going upwards
5. Repeat this test for Chrome, Firefox & Safari. This only needs to be done in Desktop as the right panel only appear in desktop width screens


kangzeroo:
Possibly could use the [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) part of native browser

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,35 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   const [currentHeadingSlug, setCurrentHeadingSlug] = useState<string | null>(
     null
   )
+
+  const generateHeadingId = (slug: string): string => `links-${slug}`
```
kangzeroo, line 39: This ID was being used in many places, so I created a helper function to ensure consistency of the ID string.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,35 @@ const RightPanel: React.FC<IRightPanelProps> = ({
...
+    .map(({ slug }) => ({
+      id: generateHeadingId(slug),
+      ref: createRef()
+    }))
+    .reduce((acc, curr) => {
+      return {
+        ...acc,
+        [curr.id]: curr.ref
+      }
+    }, {})
```
kangzeroo, line 53: This dynamically creates `React ref` for us to get references to the link DOM elements.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,34 @@ const RightPanel: React.FC<IRightPanelProps> = ({
...
+
+  const scrollSidebar = (): void => {
+    if (currentHeadingSlug && tableOfContentsRefs) {
+      const target =
+        tableOfContentsRefs[generateHeadingId(currentHeadingSlug)].current
+      if (target) {
+        ;(target.parentNode as HTMLDivElement).scrollTop =
+          target.offsetTop - (target.parentNode as HTMLDivElement).offsetTop
```
kangzeroo, line 61: Using a `<HTMLDivElement>.scrollTop` based solution lets you control the scroll isolated to a div, whereas the alternative `<HTMLDivElement>. scrollIntoView ()` affected multiple scroll divs.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -78,7 +107,10 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   }
 
   useEffect(() => {
-    const throttledSetCurrentHeader = throttle(updateCurrentHeader, 100)
+    const throttledSetCurrentHeader = (): void => {
+      throttle(updateCurrentHeader, 100)()
+      debounce(scrollSidebar, 100)()
```
kangzeroo, line 112: The extra `()` in `throttle(updateCurrentHeader, 100)()` looks weird, but that's because lodash throttle returns a function for invocation. I don't use lodash much, so at first this confused me, but thanks to VSCode tooling ❤️ I can easily see what going on under the hood.

image

kangzeroo (commented):
Self documentation for easier code review

kangzeroo:
Sorry folks! I thought this was my forked repo but I just realized its the main repo.
I will replicate this in my fork 😅 You can close this PR.

cc: @shcheklein @mattseddon @rogermparent

kangzeroo:
Sorry folks! I thought this was my forked repo but I just realized its the main repo.
I have replicate this in my [forked repo PR here](https://github.com/iterative/dvc.org/pull/2606) 😅  I am closing this PR.

PR #2606, opened 2021-07-03
fixed right panel scrollbar sync

### LINKED ISSUE #1333: right panel scroll bar improvements

kangzeroo:
Fix #1333

## Problem
This PR tackles https://github.com/iterative/dvc.org/issues/1333. The problem is explained in [this video.](https://share.vidyard.com/watch/eGwnBPBjWmyAF2Akp7BBmb)
Basically the right panel of anchor tags doesn't scroll with the page. Thus when readers scroll down to the later parts of the page, the right panel fails to provide value in showing them where they are in the anchor tag table of contents.

DVC_Core_Project___Data_Version_Control_·_DVC

## Solution
We force the right sidebar to scroll with the page so that readers always know where they are. 
At first I used a `<HTMLDivElement>.scrollIntoView()` based solution, but that was not smooth in Firefox because it affected the whole screen. I switched to a `<HTMLDivElement>.scrollTop` based solution, which works smoothly in Chrome/Firefox/Safari. 

[Video Demo](https://share.vidyard.com/watch/fNKZuv7hzco8Ki8kgfh8jJ?)

✅  `yarn format-check`
✅  `yarn lint-ts`

## Notes:
- The scroll function is debounced for kinder browser performance. You can further increase the debounce duration, but this comes at the cost of the scroll bar appearing to "lag" slightly. The current 100ms is fine as the page is mostly static and isn't scare of processing resources
- On Safari, the very top of the page will not highlight the `How to report a problem` link, which is also the previous behavior in production. Do we want to push further and get that minor part working perfectly too? How does that fit into our triaged list of priorities?
- On Chrome/Firefox/Safari, if you scroll up very very fast, the right panel scrollbar won't keep up. However, you have to try quite hard to make this bug appear. For the majority of users they will not see this. Do we want to push further and get that minor part working perfectly too? How does that fit into our triaged list of priorities?
- Total time, 1.5 hours

## Testing Instructions
0. We will be replicating [these test steps](https://share.vidyard.com/watch/fNKZuv7hzco8Ki8kgfh8jJ?)
1. Visit localhost link http://localhost:8000/doc/user-guide/contributing/core
2. Resize the page shorter to force a smaller right panel (this gives us more scroll bar to see in action)
3. Scroll down the page and watch the right panel scroll with you to show the latest highlighted link
4. Scroll up the page to see the same behavior going upwards
5. Repeat this test for Chrome, Firefox & Safari. This only needs to be done in Desktop as the right panel only appear in desktop width screens

### DIFF
--- a/src/components/Documentation/RightPanel/index.tsx
+++ b/src/components/Documentation/RightPanel/index.tsx
@@ -1,6 +1,7 @@
-import React, { useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect, useRef, createRef } from 'react'
 import cn from 'classnames'
 import throttle from 'lodash/throttle'
+import debounce from 'lodash/debounce'
 
 import { IHeading } from '../'
 import Link from '../../Link'
@@ -34,6 +35,31 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   const [currentHeadingSlug, setCurrentHeadingSlug] = useState<string | null>(
     null
   )
+  const generateHeadingId = (slug: string): string => `links-${slug}`
+  const tableOfContentsRefs: {
+    [key: string]: React.RefObject<HTMLDivElement>
+  } = headings
+    .map(({ slug }) => ({
+      id: generateHeadingId(slug),
+      ref: createRef()
+    }))
+    .reduce((acc, curr) => {
+      return {
+        ...acc,
+        [curr.id]: curr.ref
+      }
+    }, {})
+
+  const scrollSidebar = (): void => {
+    if (currentHeadingSlug && tableOfContentsRefs) {
+      const target =
+        tableOfContentsRefs[generateHeadingId(currentHeadingSlug)].current
+      if (target) {
+        ;(target.parentNode as HTMLDivElement).scrollTop =
+          target.offsetTop - (target.parentNode as HTMLDivElement).offsetTop
+      }
+    }
+  }
   const updateCurrentHeader = (): void => {
     const currentScroll = getScrollPosition()
     const coordinateKeys = Object.keys(headingsOffsets)
@@ -78,7 +104,10 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   }
 
   useEffect(() => {
-    const throttledSetCurrentHeader = throttle(updateCurrentHeader, 100)
+    const throttledSetCurrentHeader = (): void => {
+      throttle(updateCurrentHeader, 100)()
+      debounce(scrollSidebar, 100)()
+    }
 
     document.addEventListener('scroll', throttledSetCurrentHeader)
     window.addEventListener('resize', updateHeadingsPosition)
@@ -107,7 +136,7 @@ const RightPanel: React.FC<IRightPanelProps> = ({
     if (currentHeadingSlug) {
       setIsScrollToCurrentHeadingHappened(true)
       const currentHeadingSlugElem = document.getElementById(
-        `link-${currentHeadingSlug}`
+        generateHeadingId(currentHeadingSlug)
       )
       const contentBlockElem = contentBlockRef.current
       if (currentHeadingSlugElem && contentBlockElem) {
@@ -133,7 +162,11 @@ const RightPanel: React.FC<IRightPanelProps> = ({
           </div>
           <div className={styles.contentBlock} ref={contentBlockRef}>
             {headings.map(({ slug, text }) => (
-              <div id={`link-${slug}`} key={`link-${slug}`}>
+              <div
+                id={generateHeadingId(slug)}
+                key={generateHeadingId(slug)}
+                ref={tableOfContentsRefs[generateHeadingId(slug)]}
+              >
                 <Link
                   className={cn(
                     styles.headingLink,


kangzeroo:
Another possibility is to use the [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) part of native browser. However, the current solutions make use of the existing code thats on the page (leveraging the change of the `document.addEventListener('scroll', throttledSetCurrentHeader)`. I prefer this over adding a new event listener.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,31 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   const [currentHeadingSlug, setCurrentHeadingSlug] = useState<string | null>(
     null
   )
+  const generateHeadingId = (slug: string): string => `links-${slug}`
```
kangzeroo, line 38: This ID was being used in many places, so I created a helper function to ensure consistency of the ID string.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,31 @@ const RightPanel: React.FC<IRightPanelProps> = ({
...
+    .map(({ slug }) => ({
+      id: generateHeadingId(slug),
+      ref: createRef()
+    }))
+    .reduce((acc, curr) => {
+      return {
+        ...acc,
+        [curr.id]: curr.ref
+      }
+    }, {})
```
kangzeroo, line 51: This dynamically creates React ref for us to get references to the link DOM elements.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -34,6 +35,31 @@ const RightPanel: React.FC<IRightPanelProps> = ({
...
+
+  const scrollSidebar = (): void => {
+    if (currentHeadingSlug && tableOfContentsRefs) {
+      const target =
+        tableOfContentsRefs[generateHeadingId(currentHeadingSlug)].current
+      if (target) {
+        ;(target.parentNode as HTMLDivElement).scrollTop =
+          target.offsetTop - (target.parentNode as HTMLDivElement).offsetTop
```
kangzeroo, line 59: Using a `<HTMLDivElement>.scrollTop` based solution lets you control the scroll isolated to a div, whereas the alternative `<HTMLDivElement>. scrollIntoView ()` affected multiple scroll divs.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -78,7 +104,10 @@ const RightPanel: React.FC<IRightPanelProps> = ({
   }
 
   useEffect(() => {
-    const throttledSetCurrentHeader = throttle(updateCurrentHeader, 100)
+    const throttledSetCurrentHeader = (): void => {
+      throttle(updateCurrentHeader, 100)()
+      debounce(scrollSidebar, 100)()
```
kangzeroo, line 109: The extra `()` in `throttle(updateCurrentHeader, 100)()` looks weird, but that's because lodash throttle returns a function for invocation. I don't use lodash much, so at first this confused me, but thanks to VSCode tooling ❤️ I can easily see what going on under the hood.

image

kangzeroo (commented):
Self documentation for easier code review

kangzeroo:
Originally I made a [PR on the wrong repo](https://github.com/iterative/dvc.org/pull/2605) (the original dvc.org repo) instead of my own fork. This PR is the actual to-be-submitted PR.

kangzeroo:
Originally I made a PR on the wrong repo (the original dvc.org repo) instead of my own fork. This PR is the actual to-be-submitted PR.

Tagging @shcheklein @mattseddon @rogermparent for review (the github reviewer tags are disabled so I am manually tagging here). Thanks folks!
image

src/components/Documentation/RightPanel/index.tsx
rogermparent: Personally I'd do what the `map` does in a single `reduce`, but that's just a stylistic nitpick with a likely negligible performance hit.

src/components/Documentation/RightPanel/index.tsx
rogermparent: This doesn't quite work in its current form, as every time this `throttledSetCurrentHeader` is invoked it's using `throttle` and `debounce` to create a new function every time the regular non-throttled/debounced `throttledSetCurrentHeader` runs, so effectively neither child function actually gets limited.

I tested it out by cranking up the delay to something unreasonable and sticking `console.log`s in all the relevant functions, and the child functions are still being run every time the parent is.

src/components/Documentation/RightPanel/index.tsx
```diff
@@ -133,7 +162,11 @@ const RightPanel: React.FC<IRightPanelProps> = ({
           </div>
           <div className={styles.contentBlock} ref={contentBlockRef}>
             {headings.map(({ slug, text }) => (
-              <div id={`link-${slug}`} key={`link-${slug}`}>
+              <div
+                id={generateHeadingId(slug)}
+                key={generateHeadingId(slug)}
+                ref={tableOfContentsRefs[generateHeadingId(slug)]}
+              >
```
rogermparent, line 169: It'd be better to only run `generateHeadingId` only once, even if it means we need to use the uglier curly-brace syntax.

kangzeroo:
This PR is closed as the ticket was suppose to be done in a completely private repo, which is here:
https://github.com/kangzeroo/dvc.org/pull/1

src/components/Documentation/RightPanel/index.tsx
kangzeroo: Nice catch! Updated in https://github.com/kangzeroo/dvc.org/pull/1/commits/faff75314c86bf59687e66883f50be045c97f0de

src/components/Documentation/RightPanel/index.tsx
kangzeroo: I agree, that is cleaner. Updated in https://github.com/kangzeroo/dvc.org/pull/1/commits/a44334d073dbc212ad1d73ad58daba10537de0dc

src/components/Documentation/RightPanel/index.tsx
kangzeroo: I ended up removing this afterwards in my private repo in my deeper fix.
https://github.com/kangzeroo/dvc.org/pull/1

Good catch, I should have read up into lodash docs more. Makes sense that the throttle should be on the outside. The PR in my private repo removed this extra function entirely, so the throttle is back to how it was before.

PR #2612, opened 2021-07-06
ref: clarify GDrive remote param

jorgeorpinel:
Continues #2594


content/docs/command-reference/remote/modify.md
```diff
@@ -528,20 +528,23 @@ more information.
...
   ```dvc
   $ dvc remote modify myremote \
                       gdrive_service_account_user_email 'myemail-addr'
   ```
 
+  Domain-wide delegation of authority is required to include the following OAuth
+  Scopes (authorized in Google Workspace Admin console):
+
+  - `https://www.googleapis.com/auth/drive`
+  - `https://www.googleapis.com/auth/drive.appdata`
```
jorgeorpinel, line 543: Hi @mmikolajczak and thanks for #2594!

May I ask what does it mean to include OAuth Scopes? Included where? For what? Just trying to make sure this document makes sense.

shcheklein:
Not sure why do we change the text here? :)

Let's do a minor update to the formatting if it's needed, beyond that - what are the concerns that we have?

jorgeorpinel:
@shcheklein the concerns are that the existing explanation is very confusing. E.g. the term "impersonating" is only mentioned once in the Google docs. The actual term here is "delegation". The recent change to this text (#2594) uses the right terminology but then it's all mixed together. This is a pretty minor change in any case, I think.

content/docs/command-reference/remote/modify.md
jorgeorpinel: And also @shcheklein as you can see above even after improving the explanation I'm not even sure what this means still. I realize that only users who need this and know what we're talking about will even find this text, but I want to make sure we're explaining this in a sensical way for them.

shcheklein:
Arguably `impersonating` is more human readable than "delegate" :) But anyways, let's just update the term then? I see that we removed some parts, move things around, etc.

content/docs/command-reference/remote/modify.md
shcheklein: Yes. Should we in this case either try it to make sure that we understand this deep enough to do edits, or rely on actual users (esp considering that this is indeed an advanced thing and people who are willing to setup this probably already have some experience and can tolerate small issues), wdyt?

jorgeorpinel:
Yes, I updated the term, removed the unnecessary note about this being optional (most params are), fixed horizontal scrolling in the code block, and moved/reformatted the recent text introduced in https://github.com/iterative/dvc.org/pull/2594/files (which I didn't have a time to review before it was merged).

Are you saying the changes are wrong or unnecessary? I think it's much easier to read this way.

Now:
image

Changes:
image

p.s. TBH I'd even like to remove the last bullet list and just make a compressed comma-separated list.

jorgeorpinel:
p.s. both "impersonation" and "delegation" are cryptic. But the latter is the name of the feature in Google: https://developers.google.com/admin-sdk/directory/v1/guides/delegation

shcheklein:
Both are fine to me :)

shcheklein (approved):

jorgeorpinel:
OK well since I've already changed it 😬 . Thanks!

content/docs/command-reference/remote/modify.md
mmikolajczak: @jorgeorpinel @shcheklein – I guess I'm a little late to the party (different timezones I belive) as I see that the changes were already merged, but I will try to give some response/context (perhaps it will be useful in the future) nonetheless.

The main point of this delegation (at least  from our experiance/use cases): when using service accounts, if you are uploading files to location that is not in Google Workspace team/shared drive of your organization, unless you assume identity ("delegate") of some other actual user, the ownership of files uploaded will be assigned to the service account.  This is a downside, because this accounts have a separate storage quotas that are quite low by default, opposed to e.g. actual workspace users (which depending on the Workspace tier are higher or even unlimited).
For that reason we would rather want to upload files as an actual Workspace user instead of service account when using DVC commands (which is possible via this impersonation/delegation option).

Now, for the instructions themselves, they were rather clear for me (and I wouldn't count myself as that much advanced user). From the DVC perspective, the setup is quite easy, the only thing that needs to be done is adding
**gdrive_service_account_user_email**   to the remote config. Much more needs to be done in GCP/Google Workspace admin console, but it is also described quite well in the linked guide on delegation. The only thing that was unclear (and actually took us some time, as we naively thought that  **https://www.googleapis.com/auth/drive** is like a wild star that authorizes all possible GDrive operations) is what OAuth Scopes have to be provided during 
the following part of Google guide: https://developers.google.com/admin-sdk/directory/v1/guides/delegation#delegate_domain-wide_authority_to_your_service_account – i.e. which operations should service accounts be able to perform when delegating/impersonating user. This is DVC specific, as they should match the scopes that DVC code reponsible for handling GDrive remote requests/uses (https://github.com/iterative/dvc/blob/9d9ff10be1f516d0c6459370c8089caac52e91fb/dvc/fs/gdrive.py#L239).

Overall, I would say that for someone who more or less knows what he/she wants to achive/setup the instructions are currently rather fine. Perhaps it would be more beneficial to elaborate why/for what use case this option/setup can be useful (e.g. mentioned earlier issue of ownership and avoiding storage quotas) – but it rather specific to remote specifics than DVC itself, so I don't know if something like that should be inluded in the docs here.

(Well, this comment actually turned out to be quite long, I hope it makes things at least a bit less cryptic. If something still could be somehow clarified – let me know and I will try to help).

jorgeorpinel: Wow @mmikolajczak that's great feedback, thanks a lot!

The use case is clearer for me now. Probably no need to explain that motivation since users who need this feature will already know that.

> This is DVC specific, as they should match the scopes that DVC code reponsible for handling GDrive remote requests/uses...

This is a key point though. I've made another update to clarify on this. See https://github.com/iterative/dvc.org/pull/2622/files

PR #2613, opened 2021-07-06
updating events images shrunk

jendefig:
Added two new events.  Pics have been run through tiny.png.


shcheklein:
@jendefig images look blurry a bit (even source images) - is it expected (e.g. no better ones to get from their website)?

jendefig:
> @jendefig images look blurry a bit (even source images) - is it expected (e.g. no better ones to get from their website)?

I picked the largest images from their respective websites.  I requested a logo from DSGO yesterday, but haven't received.  In the interest of time, I think they will have to be good enough?

shcheklein:
> I picked the largest images from their respective websites. I requested a logo from DSGO yesterday, but haven't received. In the interest of time, I think they will have to be good enough?

yep, thanks for clarifying this. It's good to merge as is if there are not immediately available options.

PR #2620, opened 2021-07-08
Add image processing to docs

### LINKED ISSUE #2469: docs: better image processing (similar to blog)
**This issue has been created for unresolved PR comments on #2455 This issue is intended to be closed before the public release.**

_Originally posted by @shcheklein in https://github.com/iterative/dvc.org/pull/2455#discussion_r630528533_:

I think this is about retina-like display support. Unlike the blog engine, DVC.org doesn't process images properly. And effective you have to put extra white-background and do 2xWidth to them not look blurry and at the same moment be smaller then the viewport, or even of the same size but not blurry. E.g. https://dvc.org/doc/use-cases/versioning-data-and-model-files here images are wrong (not enough width to properly render them on the retina-like displays).

> In blog we have a special logic and additional Markdown support for this (check the contributors guide).

Let's get the image processing logic on docs to be on par with the blog's one cc @julieg18 @rogermparent

julieg18:
* adds the same image processing that we use for the blog to the documentation images as well
* edits markdown for smaller images, giving them the same appearance as before(Our `resize-image-plugin` shrinks images that have a certain size, shrinking about 10 of our docs images. For now, to keep all images looking like they had before, I updating the markdown images, keeping the same width as the other images.)
* documentation images now have:
  1. loading animation
  2. webp formats
  3. resizing

Fixes #2469

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -49,6 +49,13 @@ const plugins = [
       path: path.join(__dirname, 'static', 'uploads')
     }
   },
+  {
+    resolve: 'gatsby-source-filesystem',
+    options: {
+      name: 'img',
+      path: path.join(__dirname, 'static')
+    }
+  },
   'community-page',
   {
     resolve: 'gatsby-transformer-remark',
--- a/content/docs/cml/cml-with-dvc.md
+++ b/content/docs/cml/cml-with-dvc.md
@@ -5,7 +5,7 @@ downloaded from external sources. [DVC](https://dvc.org) is a common way to
 bring data to your CML runner. DVC also lets you visualize how metrics differ
 between commits to make reports like this:
 
-![](/img/dvc_cml_long_report.png)
+![](/img/dvc_cml_long_report.png '=700')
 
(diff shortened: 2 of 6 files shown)


gatsby-config.js
```diff
@@ -49,6 +49,13 @@ const plugins = [
       path: path.join(__dirname, 'static', 'uploads')
     }
   },
+  {
+    resolve: 'gatsby-source-filesystem',
+    options: {
+      name: 'img',
+      path: path.join(__dirname, 'static')
```
julieg18, line 56: Currently `img` contains more than just the documentation images. Should we consider creating a folder for just documentation images, like `static/uploads/images`?

rogermparent: Yeah, I think that could help- at least while we discuss whether we want to handle different images with different pipelines instead of through queries at consumption-time. I suspect that isn't possible with our current setup that has docs and blog being completely different MD engines.

rogermparent (approved):
Nice work! I'm surprised it was that easy!

PR #2626, opened 2021-07-13
Take `'=700'` off of documentation images markdown

julieg18:
* Takes off `=700` off of several documentation images, allowing our `resize-image-plugin` to resize them to a smaller size
* Example: https://dvc-org-take-off-doc-im-curosi.herokuapp.com/doc/cml/start-github

### DIFF
--- a/content/docs/cml/cml-with-dvc.md
+++ b/content/docs/cml/cml-with-dvc.md
@@ -5,7 +5,7 @@ downloaded from external sources. [DVC](https://dvc.org) is a common way to
 bring data to your CML runner. DVC also lets you visualize how metrics differ
 between commits to make reports like this:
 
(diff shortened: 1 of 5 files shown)


rogermparent (commented):
I see no problem here, @shcheklein @iterative/docs do the smaller images look acceptable, factoring in this PR adding the ability to click to see full size? If it isn't, we'll want to open up a follow-up ticket to fix that, whether it be a tweak in default resize behavior or a change to the source images.

It the result is unacceptable, this PR is small and targeted enough to be closed while we address the issue in such a way that we don't need to manually specify the width of every image.

Also, as mentioned in the meeting we'll want to look into adding some sort of handling for remote image, even if it's just a wrapper and click-for-full-size feature.

shcheklein (approved):

PR #2650, opened 2021-07-20
Hide DownloadButton dropdown with CSS instead of JS

julieg18:
* Originally, our `DownloadButton` dropdown was hidden by taking it out of the HTML with JS. This stops the dropdown links from being read with a link checker and stops the links from being read by a screen reader.

### DIFF
--- a/src/components/DownloadButton/styles.module.css
+++ b/src/components/DownloadButton/styles.module.css
@@ -33,9 +33,18 @@
   right: 0;
   top: calc(100% + 3px);
   display: flex;
+  visibility: hidden;
+  opacity: 0;
   flex-direction: column;
   background-color: #fff;
   box-shadow: 0 2px 3px 0 rgba(0, 0, 0, 0.15);
+  transition: opacity 0.2s ease-in-out, visibility 0.2s ease-in-out;
+
+  &.open {
+    visibility: visible;
+    opacity: 1;
+    z-index: 1;
+  }
 
   &.openTop {
     bottom: calc(100% + 3px);
--- a/src/components/DownloadButton/index.tsx
+++ b/src/components/DownloadButton/index.tsx
@@ -173,14 +173,18 @@ const DownloadButton: React.FC<IDownloadButtonProps> = ({ openTop }) => {
       >
         <img className={styles.triangle} src="/img/triangle.svg" alt="" />
       </TwoRowsButton>
-      {isOpened && (
-        <div className={cn(styles.dropdown, openTop && styles.openTop)}>
-          <DownloadButtonDropdownItems
-            onClick={download}
-            userOS={userOS.current}
-          />
-        </div>
-      )}
+      <div
+        className={cn(
+          styles.dropdown,
+          isOpened && styles.open,
+          openTop && styles.openTop
+        )}
+      >
+        <DownloadButtonDropdownItems
+          onClick={download}
+          userOS={userOS.current}
+        />
+      </div>
     </span>
   )
 }
--- a/src/components/DownloadButton/index.tsx
+++ b/src/components/DownloadButton/index.tsx
@@ -9,6 +9,9 @@ import { logEvent } from '../../utils/front/ga'
 
 import styles from './styles.module.css'
 
+const DVC_PUBLIC_S3_LINK =
+  'https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs'
+const DVC_REPO_S3_LINK = 'https://s3-us-east-2.amazonaws.com/dvc-s3-repo'
 const VERSION = `2.5.4`
 
 enum OS {
@@ -27,22 +30,22 @@ const itemsByOs = {
   },
   [OS.OSX]: {
     title: 'macOS',
-    url: `/osxpkg/dvc-${VERSION}.pkg`,
+    url: `${DVC_PUBLIC_S3_LINK}/osxpkg/dvc-${VERSION}.pkg`,
     download: true
   },
   [OS.WINDOWS]: {
     title: 'Windows',
-    url: `/exe/dvc-${VERSION}.exe`,
+    url: `${DVC_PUBLIC_S3_LINK}/exe/dvc-${VERSION}.exe`,
     download: true
   },
   [OS.LINUX]: {
     title: 'Linux Deb',
-    url: `/deb/pool/stable/d/dv/dvc_${VERSION}_amd64.deb`,
+    url: `${DVC_REPO_S3_LINK}/deb/pool/stable/d/dv/dvc_${VERSION}_amd64.deb`,
     download: true
   },
   [OS.LINUX_RPM]: {
     title: 'Linux RPM',
-    url: `/rpm/dvc-${VERSION}-1.x86_64.rpm`,
+    url: `${DVC_REPO_S3_LINK}/rpm/dvc-${VERSION}-1.x86_64.rpm`,
     download: true
   }
 }


src/components/DownloadButton/styles.module.css
```diff
@@ -33,9 +33,17 @@
   right: 0;
   top: calc(100% + 3px);
   display: flex;
+  visibility: hidden;
+  opacity: 0;
   flex-direction: column;
   background-color: #fff;
   box-shadow: 0 2px 3px 0 rgba(0, 0, 0, 0.15);
+  transition: opacity 0.2s ease-in-out, visibility 0.2s ease-in-out;
```
julieg18, line 41: I went ahead and added animation, making the dropdown have the same behavior as our nav dropdowns.

julieg18:
I'm not sure why but when I go the site for the first time, the links look like this:
image

The links are still working they just don't open a new tab. Then if I navigate to a different page, and go back to the home page. The links are back to what they normally look like:
image

rogermparent (approved):
Great catch!

rogermparent (changes requested):
Seems there's logic in the `DownloadButtonDropdownItems` that depends on re-rendering, due to it depending on the `userOS.current` value to render specific class names, and at build time it's `null` so that's how it gets hydrated.

We'll probably want to address that before merging this, it may take a bit of a large refactor to make this component hydration-friendly.

julieg18:
I tried a couple things, but while I was able to get the links to look correct while keeping the redirect logic, I wasn't able to get `lychee` to read the links without the links being placed directly on the html without any redirection logic. For now, I placed amazon link constants in `DownloadButton` and formed the links there without any redirect logic. What do you think, @rogermparent?

rogermparent:
I'm not sure what you mean by "redirection logic"- are these download links being redirected? If so, they probably shouldn't. 
I do see that the links have the first half factored out, which seems to be the issue here- Lychee isn't going to actually evaluate source code and check links constructed from templates without pointing the checker to built HTML (which is totally viable).

The idea I had was to take the part of that component module that constructs the list of links and, instead, put those links in their entirety in the list items that are fed into the component. I originally thought we'd put the links in a JSON file which is imported, but we may just be able to put the full links in the list and point lychee to the `DownloadButton` module.

rogermparent:
If updating every link for each version update isn't preferable, we can simply point Lychee to dvc.org's homepage, which has the download button, and it should be able to find all the links in the built HTML after this PR goes through.

julieg18:
> I'm not sure what you mean by "redirection logic"- are these download links being redirected? I do see that the links have the first half factored out, which seems to be the issue here- Lychee isn't going to actually evaluate source code and check links constructed from templates without pointing the checker to built HTML (which is totally viable).

Yes, I was referring to the logic updates the download link endings to full links. Sorry for the confusion 😃 

> The idea I had was to take the part of that component module that constructs the list of links and, instead, put those links in their entirety in the list items that are fed into the component.

Isn't that what the pull request is currently doing. 🤔 I updated `DownloadButton` to use the entire links. 

> If updating every link for each version update isn't preferable, we can simply point Lychee to dvc.org's homepage, which has the download button, and it should be able to find all the links in the built HTML after this PR goes through.

Yes, that is our current plan, having `lychee` scan `dvc.org` for the download links.

rogermparent:
> Isn't that what the pull request is currently doing. :thinking: I updated `DownloadButton` to use the entire links.

This PR currently adds links to the buttons with template strings like this:

```ts
const DVC_PUBLIC_S3_LINK =
  'https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs'
const DVC_REPO_S3_LINK = 'https://s3-us-east-2.amazonaws.com/dvc-s3-repo'
const VERSION = `2.5.4`

[...]

  [OS.OSX]: {
    title: 'macOS',
    url: `${DVC_PUBLIC_S3_LINK}/osxpkg/dvc-${VERSION}.pkg`,
    download: true
  },
```

Whereas I believe link checkers (at least when checking sources) require the full link without any templates, like so:

```
  [OS.OSX]: {
    title: 'macOS',
    url: `https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs/osxpkg/dvc-2.5.4.pkg`,
    download: true
  },
```

> Yes, that is our current plan, having `lychee` scan `dvc.org` for the download links.

Got it! I had assumed we were scanning the source, but if we scan the output then how the source constructed is a moot point.

rogermparent:
Pointing lychee at this PR's preview does indeed check the download links! I feel that as long as they're present in HTML (unlike current master) they should work regardless of how the links get there in the source.

```
lychee https://dvc-org-update-dvc-down-vlrhnm.herokuapp.com/ --scheme https -v
? data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHN2ZyB3aWR0aD0iMTI5cHgiIGhlaWdodD0iMTU4cHgiIHZpZXdCb3g9IjAgMCAxMjkgMTU4IiB2ZXJzaW9uPSIxLjEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiPgogICAgPCEtLSBHZW5lcmF0b3I6IFNrZXRjaCA0OSAoNTEwMDIpIC0gaHR0cDovL3d3dy5ib2hlbWlhbmNvZGluZy5jb20vc2tldGNoIC0tPgogICAgPHRpdGxlPkdseXBoZSAxPC90aXRsZT4KICAgIDxkZXNjPkNyZWF0ZWQgd2l0aCBTa2V0Y2guPC9kZXNjPgogICAgPGRlZnM+PC9kZWZzPgogICAgPGcgaWQ9IlBhZ2UtMSIgc3Ryb2tlPSJub25lIiBzdHJva2Utd2lkdGg9IjEiIGZpbGw9Im5vbmUiIGZpbGwtcnVsZT0iZXZlbm9kZCI+CiAgICAgICAgPGcgaWQ9IkxhbmRpbmctcGFnZSIgdHJhbnNmb3JtPSJ0cmFuc2xhdGUoLTQyLjAwMDAwMCwgLTI4MzkuMDAwMDAwKSI+CiAgICAgICAgICAgIDxnIGlkPSJTdWJzY3JpYmUiIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0zLjAwMDAwMCwgMjgyOS4wMDAwMDApIj4KICAgICAgICAgICAgICAgIDxnIGlkPSJzdWJzY3JpYmUiIHRyYW5zZm9ybT0idHJhbnNsYXRlKDM1LjAwMDAwMCwgMC4wMDAwMDApIj4KICAgICAgICAgICAgICAgICAgICA8ZyBpZD0iR2x5cGhlLTEiPgogICAgICAgICAgICAgICAgICAgICAgICA8cmVjdCBpZD0iUmVjdGFuZ2xlLTciIGZpbGw9IiNGRkZGRkYiIHRyYW5zZm9ybT0idHJhbnNsYXRlKDY5LjUwMDAwMCwgODguOTkyNTkzKSByb3RhdGUoNTQuMDAwMDAwKSB0cmFuc2xhdGUoLTY5LjUwMDAwMCwgLTg4Ljk5MjU5MykgIiB4PSI0NCIgeT0iMzQuNjIyMjIyMiIgd2lkdGg9IjUxIiBoZWlnaHQ9IjEwOC43NDA3NDEiIHJ4PSIyNS41Ij48L3JlY3Q+CiAgICAgICAgICAgICAgICAgICAgICAgIDxyZWN0IGlkPSJSZWN0YW5nbGUtNy1Db3B5LTIiIGZpbGw9IiNGRkZGRkYiIG9wYWNpdHk9IjAuNTEwMTAwNDQ2IiB0cmFuc2Zvcm09InRyYW5zbGF0ZSg1OS41MDAwMDAsIDEyNS4yMzk1MDYpIHJvdGF0ZSg1NC4wMDAwMDApIHRyYW5zbGF0ZSgtNTkuNTAwMDAwLCAtMTI1LjIzOTUwNikgIiB4PSIzNCIgeT0iNzAuODY5MTM1OCIgd2lkdGg9IjUxIiBoZWlnaHQ9IjEwOC43NDA3NDEiIHJ4PSIyNS41Ij48L3JlY3Q+CiAgICAgICAgICAgICAgICAgICAgICAgIDxyZWN0IGlkPSJSZWN0YW5nbGUtNy1Db3B5IiBzdHJva2U9IiMxM0FEQzciIHN0cm9rZS13aWR0aD0iMiIgb3BhY2l0eT0iMC43MTM1NjAyNjgiIHN0eWxlPSJtaXgtYmxlbmQtbW9kZTogbXVsdGlwbHk7IiB0cmFuc2Zvcm09InRyYW5zbGF0ZSg4OS41MDAwMDAsIDUyLjc0NTY3OSkgcm90YXRlKDU0LjAwMDAwMCkgdHJhbnNsYXRlKC04OS41MDAwMDAsIC01Mi43NDU2NzkpICIgeD0iNjUiIHk9Ii0wLjYyNDY5MTM1OCIgd2lkdGg9IjQ5IiBoZWlnaHQ9IjEwNi43NDA3NDEiIHJ4PSIyNC41Ij48L3JlY3Q+CiAgICAgICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICA8L2c+CiAgICAgICAgPC9nPgogICAgPC9nPgo8L3N2Zz4=
? support@dvc.org
? data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHN2ZyB3aWR0aD0iMTI5cHgiIGhlaWdodD0iMTU4cHgiIHZpZXdCb3g9IjAgMCAxMjkgMTU4IiB2ZXJzaW9uPSIxLjEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiPgogICAgPCEtLSBHZW5lcmF0b3I6IFNrZXRjaCA0OSAoNTEwMDIpIC0gaHR0cDovL3d3dy5ib2hlbWlhbmNvZGluZy5jb20vc2tldGNoIC0tPgogICAgPHRpdGxlPkdseXBoIDI8L3RpdGxlPgogICAgPGRlc2M+Q3JlYXRlZCB3aXRoIFNrZXRjaC48L2Rlc2M+CiAgICA8ZGVmcz48L2RlZnM+CiAgICA8ZyBpZD0iUGFnZS0xIiBzdHJva2U9Im5vbmUiIHN0cm9rZS13aWR0aD0iMSIgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIj4KICAgICAgICA8ZyBpZD0iTGFuZGluZy1wYWdlIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgtMTA2OC4wMDAwMDAsIC0zMDMyLjAwMDAwMCkiPgogICAgICAgICAgICA8ZyBpZD0iU3Vic2NyaWJlIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgtMy4wMDAwMDAsIDI4MjkuMDAwMDAwKSI+CiAgICAgICAgICAgICAgICA8ZyBpZD0ic3Vic2NyaWJlIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgzNS4wMDAwMDAsIDAuMDAwMDAwKSI+CiAgICAgICAgICAgICAgICAgICAgPGcgaWQ9IkdseXBoLTIiIHRyYW5zZm9ybT0idHJhbnNsYXRlKDEwMjYuMDAwMDAwLCAxOTMuMDE0ODE1KSI+CiAgICAgICAgICAgICAgICAgICAgICAgIDxyZWN0IGlkPSJSZWN0YW5nbGUtNy1Db3B5LTQiIGZpbGw9IiNGRkZGRkYiIG9wYWNpdHk9IjAuMjY3NjMzOTI5IiB0cmFuc2Zvcm09InRyYW5zbGF0ZSg1OS41Mjk1NDQsIDEyNS4xNDYyMTQpIHJvdGF0ZSg1NC4wMDAwMDApIHRyYW5zbGF0ZSgtNTkuNTI5NTQ0LCAtMTI1LjE0NjIxNCkgIiB4PSIzNC4wMjk1NDM2IiB5PSI3MC43NzU4NDM5IiB3aWR0aD0iNTEiIGhlaWdodD0iMTA4Ljc0MDc0MSIgcng9IjI1LjUiPjwvcmVjdD4KICAgICAgICAgICAgICAgICAgICAgICAgPHJlY3QgaWQ9IlJlY3RhbmdsZS03LUNvcHktMyIgZmlsbD0iI0ZGRkZGRiIgdHJhbnNmb3JtPSJ0cmFuc2xhdGUoODkuNTI5NTQ0LCA1Mi42NTIzODcpIHJvdGF0ZSg1NC4wMDAwMDApIHRyYW5zbGF0ZSgtODkuNTI5NTQ0LCAtNTIuNjUyMzg3KSAiIHg9IjY0LjAyOTU0MzYiIHk9Ii0xLjcxNzk4MzIyIiB3aWR0aD0iNTEiIGhlaWdodD0iMTA4Ljc0MDc0MSIgcng9IjI1LjUiPjwvcmVjdD4KICAgICAgICAgICAgICAgICAgICAgICAgPHJlY3QgaWQ9IlJlY3RhbmdsZS03LUNvcHktNSIgc3Ryb2tlPSIjNkI0NDlBIiBzdHJva2Utd2lkdGg9IjIiIHN0eWxlPSJtaXgtYmxlbmQtbW9kZTogbXVsdGlwbHk7IiB0cmFuc2Zvcm09InRyYW5zbGF0ZSg2OS41Mjk1NDQsIDg4Ljg5OTMwMSkgcm90YXRlKDU0LjAwMDAwMCkgdHJhbnNsYXRlKC02OS41Mjk1NDQsIC04OC44OTkzMDEpICIgeD0iNDUuMDI5NTQzNiIgeT0iMzUuNTI4OTMwNCIgd2lkdGg9IjQ5IiBoZWlnaHQ9IjEwNi43NDA3NDEiIHJ4PSIyNC41Ij48L3JlY3Q+CiAgICAgICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICA8L2c+CiAgICAgICAgPC9nPgogICAgPC9nPgo8L3N2Zz4=
? s3://mybucket
? http://www.w3.org/2000/svg
✔ https://github.githubassets.com/assets/gist-embed-b3b573358bfc66d89e1e95dbf8319c09.css [200 OK]
✔ https://dvc.org/ [200 OK]
✔ https://dvc.org/social-share.png [200 OK]
✔ https://cml.dev/ [200 OK]
✔ https://iterative.ai/ [200 OK]
✔ https://studio.iterative.ai/ [200 OK]
✔ https://www.youtube.com/channel/UC37rp97Go-xIX3aNFVHhXfQ [200 OK]
✔ https://www.youtube.com/embed/UbL7VUpv1Bs?rel=0&amp;controls=0&amp;showinfo=0; [200 OK]
✔ https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs/osxpkg/dvc-2.5.4.pkg [200 OK]
✔ https://s3-us-east-2.amazonaws.com/dvc-s3-repo/deb/pool/stable/d/dv/dvc_2.5.4_amd64.deb [200 OK]
✔ https://s3-us-east-2.amazonaws.com/dvc-s3-repo/rpm/dvc-2.5.4-1.x86_64.rpm [200 OK]
✔ https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs/exe/dvc-2.5.4.exe [200 OK]
✔ https://sweedom.us10.list-manage.com/subscribe/post?u=a08bf93caae4063c4e6a351f6&id=24c0ecc49a [200 OK]
✔ https://iterative.ai/about#career [200 OK]
✔ https://github.com/iterative/dvc [200 OK]
⠒ 20/23 [00:00:00] █████████████████████░░░░ ✔ https://github.com/iterative/dvc [200 OK]
thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code', lychee-lib/src/types/status.rs:66:26
✔ https://www.dvc.org/chat [200 OK]
✔ https://twitter.com/DVCorg [200 OK]

📝 Summary
---------------------
🔍 Total...........22
✅ Successful......17
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........5
🚫 Errors...........0
```

rogermparent:
I finally understand what you mean about the redirects, as in `redirects-list` was what is completing the links with the S3 bits at the start. I'm definitely not a fan, especially where the links are "redirected" before the component is rendered such that users can't even copy a nice-looking dvc.org link, but if somehow this prevents the links from being read by Lychee then that means these ahead-of-time redirects aren't being applied during SSG, only after JS kicks in, which would be its own problem.

julieg18:
> Pointing lychee at this PR's preview does indeed check the download links! I feel that as long as they're present in HTML (unlike current master) they should work regardless of how the links get there in the source.

Yes, with my newest commit that creates the link with template string, `lychee` is able to read them! Before `lychee` was seeing the shortened versions. Are we good to merge this or do we need to update the links differently?

>  I'm definitely not a fan, especially where the links are "redirected" before the component is rendered such that users can't even copy a nice-looking dvc.org link, but if somehow this prevents the links from being read by Lychee then that means these ahead-of-time redirects aren't being applied during SSG, only after JS kicks in, which would be its own problem.

Agreed, maybe we need to look at whatever links are being rendered like this and see if they need to be refactored 🤔

rogermparent (approved):
I think where this PR matches the functionality of master that it's mergable, but we'll need some follow-up:

1. Do we want to provide users the `dvc.org/exe/dvc-${version}` links as opposed to the `s3.amazonaws[...]/exe/dvc-${version}` links? That seems like the original intent, as it makes sharing download links easier and allows us to change the backend while keeping existing instances of the link functional. Doing so will require us to allow some links, like these, to opt-out of the pre-redirect behavior. @shcheklein @iterative/docs
 
2. With this change, we can use Lychee to link-check dvc.org's homepage, probably with a `include` filter that makes it exclusively check the download links. #1 informs this, as if we use redirects (real redirects, not pre-redirects) we'll need to make sure Lychee is configured to follow them.

Anyway, nice work @julieg18!

shcheklein:
> Do we want to provide users the dvc.org/exe/dvc-${version} links 

How did we define links before in code? Did we have `dvc.org/exe`-type URLs?

rogermparent:
> How did we define links before in code? Did we have `dvc.org/exe`-type URLs?

We did not, we defined them as the `dvc.org/exe` URLs but from looking at `dvc.org` before this was merged, the redirect logic changed them to the same big `s3.amazonaws.../exe` links you see now, the same as what would be copied by users. 

Now that I think about it, @julieg18's description of the previous behavior used `dvc.org`-type URLS on SSR but the redirect behavior gave us the `amazon`-type links on on client side.

shcheklein:
Okay, I see. 

Yes, it's not good that we show "raw" links. Also "nice" links also don't look right to me - they should be something like: `dvc.org/download/exe...` (we should use a common prefix or a third level domain name - e.g. download.dvc.org)

In this PR we are hardcoding the value second time (redirects.json is the first one). To my mind we should keep only redirects. And probably wrap react link in some way so that links look normal.

PR #2656, opened 2021-07-21
blog: Predefined models exp

flippedcoder:
This is a post about how you can use DVC experiments to help fine-tune a pre-trained model like AlexNet or SqueezeNet.


jorgeorpinel:
Cool! Please let us know when you'd like reviewers involved. Thanks

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,148 @@
...
+[this repo](https://github.com/iterative/pretrained-model-demo).
+
+In the `pretrained-model-tuner.py` file, you'll find the code that defines both
+the AlexNet and SqueezeNet models. We start by initializing these models so we
+can get the number of model features and the input size we need for fine-tuning.
+
+Since the project has everything we need to initialize the models, we can start
+training and comparing the differences between them with the ants/bees dataset.
+Running experiments to get the best tuning for each model can make it difficult
+to see which changes led to a better result. That's why will be using DVC to
```
jendefig, line 57: insert "we"

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,148 @@
...
+We'll add a stage for the training step in this example. To do that, run:
+
+```dvc
+dvc stage add --name train --deps data/hymenoptera_data --deps pretrained-model-tuner.py --metrics-no-cache results.json python pretrained-model-tuner.py
+```
+
+Let's break down the different parts of this command.
+
+- `--name` is the title of the stage
+- `--deps` defines the files and directories that a stage needs in order to run
```
jendefig, line 107: maybe add: "short for dependencies, defines the ..."  This is probably is clear to the software engineering set, but I think will help clarify for those that are more green.

dberenbaum (commented):
Fun article idea!

With the hyperparameter tuning post, it seemed mostly useful to highlight a dvc pattern that wasn't documented elsewhere - how to do hyperparameter searches using dvc experiments. This article keeps to the basic experiments workflow that users can already learn from other docs, so it seems like the value is more about the interest in transfer learning. In that case, interesting results might be more important here, or at least a meaningful approach to how/why to try different experiments. It might make sense to touch base with @daavoo whose background is in computer vision.

Also, could you add attribution to the original source code used to derive this example (and where to get the data)? I see something sort of similar in https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html. One idea for this post would be to update the code to make it more dvc-like and then show how this can make it easier to track different experiments. Some examples:

* Adding `model_name` to the parameters makes it so you can see at a glance whether an experiment is alexnet or squeezenet.
* Logging each epoch using dvclive or using checkpoints enables tracking results per epoch, which currently are getting printed to the screen and then lost.
* Saving both the training and validation metrics shows a more complete picture (it looks like only the training metrics are being saved now) and you can choose to view one or the other or both.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
+---
+title: Using Experiments to Improve Pre-trained Models
+date: 2021-08-24
+description: |
+There are times it will be easier to take a pretrained model and fine-tune it to work with your data. You can do that with DVC experiments.
```
shcheklein, line 5: 💅 style, make it < 80

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+descriptionLong: |
+When you're working with a pretrained model and you want to test different fine-tuning values, using DVC experiments can help you do that faster.
+picture: 2021-08-17/pretrained-models.png
```
shcheklein, line 8: 2021-08-17 - fix the date?

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+This creates the `.dvc/` directory for configuration, the default cache
+location, and other necessary files and directories for DVC to work.
+
+Make sure you set up a
+[virtual environment](https://python.readthedocs.io/en/stable/library/venv.html)
+and install the dependencies with:
+
+```bash
```
shcheklein, line 76: bash -> dvc

shcheklein:
Screen Shot 2021-08-05 at 5 52 59 PM

Something is wrong, does it work for you locally?

https://dvc-org-pre-defined-mod-kedssi.herokuapp.com/blog/pre-defined-models-exp is also broken it looks like ..

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+## Intro
+
+There are plenty of machine learning models available that have been trained to
+solve one problem and the knowledge gained from that can be applied to a new,
+yet related problem. For example, a model like AlexNet has been trained on
+millions of images so you could potentially use this to classify cars, animals,
+or even people. This is called transfer learning and it can save a lot of time
+on developing a model from scratch.
+
+For us to take advantage of transfer learning, we can use fine-tuning to adopt
```
shcheklein, line 28: Should we use this in the title/descriptions? Something "Faster Transfer Learning with Experiments"? (can capture more attention?)

or `Using Experiments for Transfer Learning`

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+  - Experiments
+  - Reproducibility
+  - DVC
+---
+
+## Intro
+
+There are plenty of machine learning models available that have been trained to
+solve one problem and the knowledge gained from that can be applied to a new,
+yet related problem. For example, a model like AlexNet has been trained on
```
shcheklein, line 23: good to have links - to transfer learning, AlexNet

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: Smooth Transfer Learning

or 

Pain-free Transfer Learning :)

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+In this post, we'll go through an example of fine-tuning
+[AlexNet](https://towardsdatascience.com/alexnet-the-architecture-that-challenged-cnns-e406d5297951)
+and
+[SqueezeNet](https://towardsdatascience.com/review-squeezenet-image-classification-e7414825581a)
+to classify bees and ants. We'll use DVC to handle experiments for us and we'll
+compare the results of both models at the end.
+
+## Initialize the pre-trained model
+
+We'll be fine-tuning the AlexNet model and the SqueezeNet model to classify
```
shcheklein, line 46: is there some image we can put - with bees and ants :)

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+```bash
+$ pip install -r requirements.txt
+```
+
+Now that you have DVC initialized, this is a good place to commit your changes.
+
+```git
+$ git status
+$ git add .
+$ git commit -m "added DVC pipeline"
```
shcheklein, line 85: this not a DVC pipeline yet ... it should be "DVC init"

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+the AlexNet and SqueezeNet models. We start by initializing these models so we
+can get the number of model features and the input size we need for fine-tuning.
+
+Since the project has everything we need to initialize the models, we can start
+training and comparing the differences between them with the ants/bees dataset.
+Running experiments to get the best tuning for each model can make it difficult
+to see which changes led to a better result. That's why will be using DVC to
+track changes in the code and the data.
+
+## Add a DVC pipeline
```
shcheklein, line 60: It doesn't look about pipelines, unless I missing something

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: Last two sentences with minor modifications could make it to the `descpritionLong`, which will benefit from it a lot I think.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+
+Now that you have DVC initialized, this is a good place to commit your changes.
+
+```git
+$ git status
+$ git add .
+$ git commit -m "added DVC pipeline"
+```
+
+With DVC in place, we need to add a stage to the pipeline so that we can train
```
shcheklein, line 88: `Put DVC in place` - good name for the section :)

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+
+Stages in DVC let us define individual data processes and can be used to build
+detailed machine learning pipelines. You have the ability to define the
+different steps of model creation like preprocessing, featurization, and
+training.
+
+We'll add a stage for the training step in this example. To do that, run:
+
+```dvc
+$ dvc stage add --name train --deps data/hymenoptera_data --deps pretrained_model_tuner.py --metrics-no-cache --params lr,momentum results.json python pretrained_model_tuner.py
```
shcheklein, line 101: so long lines could be split with `\` - please take a look at the docs, there are examples in the pipelines get started

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+- `--metrics-no-cache` specifies the metrics file produced by the stage, but DVC
+  doesn't track it
+
+After the metrics specification, there's the `python pretrained_model_tuner.py`
+and this defines the command for executing the training script.
+
+This will generate a new stage in the `dvc.yaml` file. If you take a look at it,
+you'll see something like:
+
+```yaml
```
shcheklein, line 118: We can start with this YAML (and then briefly explain the helper command that we were using to generate it). Also it would be good to explain why do we need this file (for DVC to start managing data, understand where metrics to pay attention to, etc - some friendly explanations)

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+to classify bees and ants. We'll use DVC to handle experiments for us and we'll
+compare the results of both models at the end.
+
+## Initialize the pre-trained model
+
+We'll be fine-tuning the AlexNet model and the SqueezeNet model to classify
+images of bees and ants. You can find the project we're working with in
+[this repo](https://github.com/iterative/pretrained-model-demo).
+
+In the `pretrained_model_tuner.py` file, you'll find the code that defines both
```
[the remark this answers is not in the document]
flippedcoder: Good catch! I got rid of that hyphenated one.

flippedcoder:
> * Adding `model_name` to the parameters makes it so you can see at a glance whether an experiment is alexnet or squeezenet.

Good point! I added that param to the stage creation and in the repo.

flippedcoder:
> * Logging each epoch using dvclive or using checkpoints enables tracking results per epoch, which currently are getting printed to the screen and then lost.

I'll do that! I'm trying to break myself out of thinking in just experiments. 😅

flippedcoder:
> * Saving both the training and validation metrics shows a more complete picture (it looks like only the training metrics are being saved now) and you can choose to view one or the other or both.

Done!

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+
+```python
+with open("params.yaml") as f:
+    yaml=YAML(typ='safe')
+    params = yaml.load(f)
+```
+
+With all of this in place, we can finally start running experiments to fine-tune
+the two models.
+
```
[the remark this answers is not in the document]
flippedcoder: If you clone the repo, then nope, there's no need for the steps above. I can cut that and just push for them to clone the repo and I'll start doing that for other blog posts.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,352 @@
...
+### Adding the train stage
+
+Stages in DVC let us define individual data processes and can be used to build
+detailed machine learning pipelines. You have the ability to define the
+different steps of model creation like preprocessing, featurization, and
+training.
+
+We'll add a stage for the training step in this example. To do that, run:
+
+```dvc
```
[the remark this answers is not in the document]
flippedcoder: 1. Done!
2. Working on adding checkpoints.
3. I'll add the model as an output to the stage.

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: I like `Using Experiments for Transfer Learning`!

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: That's a good idea! I'll run it by Jeny. :)

daavoo:
> Fun article idea!

I also like the idea very much. 

I would expect the target audience to be someone at least familiar with "finetuning" but that would like to know about how to do it the "DVC way" and my general feedback is based on that assumption, although some things might be also benefitial to other targets. 

Trying to view it from that target perspective, I would love to read something that feels easily reusable for my custom use case. Could be even a way to engage with people at the end of the post with stuff like: now you go and try it with your data, or try to improve our results by switching params, etc.

In that regard, I think that it could be good to dig a little bit in the data preparation phase, using as base the [Data Versioning Tutorial](https://dvc.org/doc/use-cases/versioning-data-and-model-files/tutorial) could be useful for adding references to a more detailed tutorial, where the experimentation part kind of feels outdated. 

In addition, I have no strong preferences regarding *ML Frameworks* but my general feeling is that using a higher level framework (i.e. keras) would reduce the amount of boilerplate code that could make it harder to differenciate the relevant "finetuning with DVC" parts from regular boilerplate.
 
>  It might make sense to touch base with @daavoo whose background is in computer vision.

 It's kind of tricky to know beforehand which experiments would be meaningful 😅  but I would try to share a few ideas/intuitions that might showcase the benefits of experimentation. 

For pretty much every idea, I think the following things would be helpful to support discussions:

- Add per-epoch and per split (train/val) metrics (could be done via DVCLive as mentioned by Dave)
- Add `training time` or `time per epoch` as a metric. This might be tricky for many reasons but even a naive measure would be useful.

Here are those ideas for experimentation "à la DVC":

### Adding more training data

**Depends on**: Having different dataset versions. Having `training_time` as metric.

**Showcases**: DVC data versioning.

**I would expect that**: extending the training set would improve perfomance. Even if that its not the case we might use the results to introduce another useful but less discussed topic: extending the validation set. 

### Comparing model architectures

**Depends on**: Having `model_name` as param. Having `training_time` as metric.

**Showcases**: `DVC params` 

**I would expect that**: 

Using a smaller version (lower `training_time`) of one family of architectures (i.e. `ResNet18`) would have worse validation accuracy than a bigger version of the same family (i.e. `ResNet50`). If that it's not the case might indicate that our problem/dataset is too simple (i.e. need to extend validation dataset to properly showcase this). 

Using a smaller model from newer families (i.e. `MobilenetV3`) would have better or similar accuracy than a bigger model from older families (i.e. `VGG19`)

**Note**: We are currently limiting the models I assume that due to torchvision not being very friendly about finetuning (i.e. can't easily reuse the same "add a new head" code for different architectures) I would either add a few if/elses for different architectures or switch to a framework where the code for adding a new head is more reusable.

### Comparing Finetuning strategies

We could use similar terminology to the one used in https://lightning-flash.readthedocs.io/en/latest/general/finetuning.html#finetune-strategies

**Depends on**: Having `feature_extract` as param. Having `training_time` as metric.

**Showcases**: `DVC params` 

**I would expect that**:  Training the entire model would result in higher `training_time`, for sure.; I'm not sure about `accuracy`. The theory says that, given a large enough dataset, training the entire model would have better perfomance than training only the head.


### Comparing Input Sizes / Data Augmentation

**Depends on**: Adapting the instantiation of `data_transforms` to be dynamic based on input `params` (I can share a code snippet for this). Having access to both `train` and `validation` metrics.

**Showcases**: `DVC params` 

**I would expect that**:  

Using smaller input size would result in lower `training_time` and  potentially lower `accuracy`.

Assuming we remove any data augmentation from the baseline, adding data augmentation (i.e. `RandomHorizontalFlip`) should improve `validation accuracy`.

---

There are probably other things that could be exposed in a similar way to `model_name`, like the Optimizer to be used (`optimizer_name`) however I think that the above ideas would be more understable for users less familiar with Deep Learning  and have more predictable results. 

Smaller vs bigger model, more data vs less data, "fancy new" vs "old school" architecture, etc. feels simpler than `momentum=0.9` vs `momentum=0.5`

flippedcoder:
> Screen Shot 2021-08-05 at 5 52 59 PM
> 
> Something is wrong, does it work for you locally?
> 
> https://dvc-org-pre-defined-mod-kedssi.herokuapp.com/blog/pre-defined-models-exp is also broken it looks like ..

I'm having issues with the `sharp` package locally. Not sure if that's related to this though.

shcheklein:
@flippedcoder 

that's what I see in the build logs:

```
error Error processing Markdown file /tmp/build_6baa4287/content/blog/2021-08-24-pre-defined-models-exp.md:
      can not read a block mapping entry; a multiline key may not be an implicit key at line 6, column 16:
    descriptionLong: |
                   ^
       not finished source and transform nodes - 1.106s
```

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: I have some links for AlexNet and SqueezeNet a little later and I added a link explaining transfer learning.

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: I have the description of the `dvc stage add` command before this YAML. So you want me to move that description below the YAML? And I added friendlier explanations for each part of the command.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,369 @@
+---
+title: Using Experiments for Transfer Learning
+date: 2021-08-24
+description: |
+  You can work with pretrained models and fine-tune them with DVC experiments.
+descriptionLong: |
+  Running experiments to get the best tuning for a model can make it difficult to see which changes led to a better result. That's why we will be using DVC to track changes in the code and the data.
```
shcheklein, line 7: 💅 please let's wrap it at 80 symbols please

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: I meant something actual, from the dataset?

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: is it resolved? should we update the git commit message?

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: yep, it can be simler to some extent - YAML is more or less friendly and something folks are already familiar vs some new command they haven't seen before. No strong opinion though. @dberenbaum wdyt? :)

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,369 @@
...
+[virtual environment](https://python.readthedocs.io/en/stable/library/venv.html)
+and install the dependencies with:
+
+```dvc
+$ pip install -r requirements.txt
+```
+
+Now that you have DVC initialized, this is a good place to commit your changes.
+
+```git
```
shcheklein, line 86: git -> dvc

content/blog/2021-08-24-pre-defined-models-exp.md
shcheklein: yep, it's fine .. usual convention (for docs) is to put link the first time you mention the concept. Not a big deal clearly.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,369 @@
...
+We'll be fine-tuning the AlexNet model and the SqueezeNet model to classify
+images of bees and ants. You can find the project we're working with in
+[this repo](https://github.com/iterative/pretrained-model-demo).
+
+In the `pretrained_model_tuner.py` file, you'll find the code that defines both
+the AlexNet and SqueezeNet models. We start by initializing these models so we
+can get the number of model features and the input size we need for fine-tuning.
+
+_You can find the original code and tutorial over at
+[this post](https://pytorch.org/tutorials/beginner/finetuning_torchvision_models_tutorial.html)._
```
shcheklein, line 56: rendering is broken

content/blog/2021-08-24-pre-defined-models-exp.md
dberenbaum: Yes, I think the first step could be cloning the repo and then explaining parts of the repo as necessary like `dvc.yaml`. I wouldn't repeat the steps that you've already done in that repo since it could be confusing and it takes longer to get to the transfer learning part. You can also give people the option to skip the explanations and jump right into the transfer learning part. I'm not even sure we really need a detailed explanation of the setup at the start of each blog post, but at least we should give the option of getting right to the main subject matter.

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: I removed this section because we all agreed that we don't need to set up DVC in every blog post.

flippedcoder:
@dberenbaum @shcheklein I added checkpoints to the project, updated the explanation of the yaml to match this, and updated the tables. Let me know if there's anything else you see that should be cleaned up or added!

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,398 @@
...
+
+Let's start by updating the hyperparameter values in `params.yaml`. You should
+have these values in your file.
+
+```yaml
+lr: 0.009
+momentum: 0.017
+```
+
+Now run another experiment with `dvc exp run` and take a look at the metrics
```
dberenbaum, line 216: The downside to adding checkpoints is that it changes the workflow. As you see in the table, `dvc exp run` will continue training additional epochs on top of your previous experiment. To start back at step 0, you will need to do `dvc exp run --reset`.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,398 @@
...
+│ │ ╟ 4e2f7c4             │ 10:30 AM │    1 │ 0.86066 │  67.566 │        75.783 │ 0.9  │ 0.35     │ alexnet    │ 2           │ 8          │ 5          │
+│ ├─╨ 6ad572d             │ 10:29 AM │    0 │    0.75 │  102.48 │         25.17 │ 0.9  │ 0.35     │ alexnet    │ 2           │ 8          │ 5          │
+└─────────────────────────┴──────────┴──────┴─────────┴─────────┴───────────────┴──────┴──────────┴────────────┴─────────────┴────────────┴────────────┘
+```
+
+Now let's update the hyperparameters and run another experiment. There are
+several ways to do this with DVC:
+
+- Change the hyperparameter values directly in `params.yaml`
+- Update the values using the `--set-param` option on `dvc exp run`
```
dberenbaum, line 202: Minor suggestion to introduce `-S` here also so that you can shorten all the commands below.

dberenbaum (changes requested):
Looks good, thanks @flippedcoder! I have a few more comments, mostly about either how to keep winnowing down the setup section or else addressing changes introduced by dvclive and checkpoints. IMO it's ready to go once those are fixed.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,398 @@
...
+and
+[SqueezeNet](https://towardsdatascience.com/review-squeezenet-image-classification-e7414825581a)
+to classify bees and ants. We'll use DVC to handle experiments for us and we'll
+compare the results of both models at the end.
+
+## Initialize the pre-trained model
+
+We'll be fine-tuning the AlexNet model and the SqueezeNet model to classify
+images of bees and ants. You can find the project we're working with in
+[this repo](https://github.com/iterative/pretrained-model-demo).
```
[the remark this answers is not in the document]
flippedcoder: Yeah, I like how that flows better. I got rid of the other link to the original project in favor of this.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,398 @@
...
+set up, we can jump straight into fine-tuning this model to see which
+hyperparameters give us the best accuracy.
+
+We'll run the first experiment with the following command.
+
+```dvc
+$ dvc exp run
+```
+
+This will execute the `pretrained_model_tuner.py` script and run for 5 epochs
```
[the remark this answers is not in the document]
flippedcoder: Ope! I pushed my stuff to the repo after running my experiments. 😅 That _should_ be 5 now.

jorgeorpinel:
@dberenbaum @shcheklein @daavoo thanks for reviewing. It looks like @flippedcoder has addressed everything? Hoping for some approvals soon

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,374 @@
...
+
+```yaml
+lr: 0.009
+momentum: 0.017
+```
+
+Now run another experiment with `dvc exp run` and take a look at the metrics
+with `dvc exp show`. Your table should look something like this now. Since we're
+using checkpoints, note that we continue training additional epochs on top of
+your previous experiment.
```
dberenbaum, line 196: Do you think it's worth mentioning `--reset` here or saying that training from scratch will be shown below? Just a thought, but up to you.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,374 @@
...
+
+```dvctable
+┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━┳━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━┓
+┃ neutral:**Experiment**              ┃ neutral:**Created**  ┃ metric:**step** ┃     metric:**acc** ┃    metric:**loss** ┃ metric:**training_time** ┃ param:**lr**   ┃ param:**momentum** ┃ param:**model_name** ┃ param:**num_classes** ┃ param:**batch_size** ┃ param:**num_epochs** ┃
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━╇━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━┩
+│ **workspace**               │ -        │   **13** │ **0.86066** │ **0.36708** │        **87.417** │ **0.025** │ **0.5**      │ **squeezenet** │ **2**           │ **8**          │ **2**          │
+│ **main**                    │ **10:27 AM** │    **1** │ **0.88525** │ **0.28284** │        **82.143** │ **0.01**  │ **0.09**     │ **squeezenet** │ **2**           │ **8**          │ **2**          │
+│ │ ╓ a771043 [exp-3d76f] │ 10:46 AM     │    1 │ 0.89754 │ 0.28754 │        82.447 │ 0.025 │ 0.5      │ squeezenet │ 2           │ 8          │ 2          │
```
dberenbaum, line 273: Something happened with this table so that the columns aren't aligning.

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,374 @@
...
+this command one more time with different values.
+
+```dvc
+$ dvc exp run --queue -S lr=0.001 -S momentum=0.09 -S num_epochs=2
+```
+
+You can check out the details for the queues you have in place by looking at the
+experiments table with `dvc exp show`. You'll see something like this.
+
+```dvctable
```
dberenbaum, line 303: `exp-7aac` in these last two tables looks off to me. It looks like you have squeezenet checkpoints that are running on top of alexnet checkpoints, which doesn't seem like the workflow you have described.

dberenbaum:
There are a couple of table formatting issues I see. Once those look good, I'm happy to approve 🙂 .

content/blog/2021-08-24-pre-defined-models-exp.md
```diff
@@ -0,0 +1,374 @@
...
+this is one of the ways you can add new stages or update existing ones.
+
+With the `train` stage defined, let's look at where the metrics actually come
+from in the code. If you open `pretrained_model_tuner`, you'll see a line where
+we dump the accuracy and loss for the training epochs into the `results.json`
+file. We're also saving the model on the epoch run and recording metrics for
+each epoch using `dvclive` logging.
+
+```python
+if phase == 'train':
```
dberenbaum, line 110: One more thought after taking a quick look at https://github.com/iterative/pretrained-model-demo/blob/132b6928ab11bbdb500e204c18e6491a03944b6b/pretrained_model_tuner.py#L84. The validation metrics are still being manually logged the old way and not captured by dvc. You could consider logging them with dvclive also like:

```python
            if phase == 'train':
                torch.save(model.state_dict(), "model.pt")

                dvclive.log('train_acc', epoch_acc.item())
                dvclive.log('train_loss', epoch_loss)
                dvclive.log('train_time', epoch_time_elapsed)

            if phase == 'val':
                dvclive.log('val_acc', epoch_acc.item())
                dvclive.log('val_loss', epoch_loss)
               
                dvclive.next_step()
```

Unfortunately, that does make your table even wider since you are adding more metrics. Another option might be to only log the validation metrics, since those are usually more helpful than training metrics, or to introduce the `--include-metrics`/`--include-params` flags to narrow the table.

dberenbaum:
Final note: feel free to ping me again to take a final look. I might miss when the comments get resolved otherwise.

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: I added those metrics to the table and I took out a few of the params we aren't adjusting.

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: Yeah, not sure what I did there but it should be cleared up now!

content/blog/2021-08-24-pre-defined-models-exp.md
dberenbaum: Sounds good! Not sure if you are still making updates, but make sure to put `dvclive.next_step()` inside `if phase == 'val'` or else you will be making two steps for every epoch (once for train, once for val).

flippedcoder:
@dberenbaum I fixed all of the tables/workflow stuff!

dberenbaum:
Thanks, @flippedcoder! Did you see my last comment in https://github.com/iterative/dvc.org/pull/2656#discussion_r692150367?

content/blog/2021-08-24-pre-defined-models-exp.md
flippedcoder: Oh! That makes sense. I wasn't quite sure about that. I'll make those updates here soon!

flippedcoder: @dberenbaum Just pushed the update for this!

dberenbaum:
Thanks again @flippedcoder! There's still some misalignments between the table headers (starting with the `val_loss` column) and other rows in some of the tables. Are you seeing the same?

flippedcoder:
@dberenbaum I've been fiddling with the table for a while and it seems like it's a combination of the formatting and the number of columns. It lines up a little better without the colors and bolding on rows, otherwise I didn't find anything else I could do about it.

dberenbaum:
@flippedcoder So that's the way it looks when you copy and paste from your terminal? Does the table look fine in your terminal?

IMO it's still worth cleaning up by manually adding/deleting spaces as needed so that it looks like the actual table output. I updated one of the tables in https://github.com/iterative/dvc.org/pull/2656/commits/7ba79bb29dc00a7f42fb8090a66ee429b43a46d5. Could you try to do the same for the tables below?

flippedcoder:
> So that's the way it looks when you copy and paste from your terminal? Does the table look fine in your terminal?

@dberenbaum It's weird because in my terminal it looks fine, but when I apply the formatting code it shifts things. I'll go through and manually update them then!

flippedcoder:
@dberenbaum The problem was with the table headers. It should be all cleared up now!

dberenbaum (approved):

shcheklein:
There is a a broken link https://github.com/iterative/pretrained-model-demo (project is private - do we want to make it public before merge?)

Could we also try to minify static/uploads/images/2021-08-24/pretrained-models.png ? tinypng or some other similar service works.

jendefig:
> Could we also try to minify static/uploads/images/2021-08-24/pretrained-models.png ? tinypng or some other similar service works.

@shcheklein I just ran the image through tinypng a second time.  It shaved off another 2%.  I guess it's bigger than normal because it has all the images in it?  Roman had already compressed it.

flippedcoder:
> There is a a broken link https://github.com/iterative/pretrained-model-demo (project is private - do we want to make it public before merge?)

@shcheklein I just changed the repo to public and it looks like everything's good to go! Would you mind merging this?

shcheklein:
@flippedcoder done, merged 🎉 !

Issue #2579, opened 2021-06-22
Run a CI/CD check that makes sure DVC download links are working on dvc update


julieg18:
* have check make sure links for downloading DVC are working
* we could either run the check after every new dvc release or run it daily

jorgeorpinel:
Can you clarify what "on dvc update" means? And if possible mention what the check would do on the issue title. Thanks

julieg18:
So far I have only possible solution for this ticket:
1. Move download links to an easier to access file.
2. Write a bash script that loops through links checking if links are correct, assigning answer to a variable. I found a way to do that with `curl`:
```
curl -I "https://s3-us-east-2.amazonaws.com/dvc-public/dvc-pkgs/osxpkg/dvc-2.1.0.pkg" 2>&1 | awk '/HTTP\// { print($2 == "200" ? "true" : "false") }'`)
```
3. Run script in Github Actions, and responding according depending on what variable says.

This solution feels kind of overcomplicated to me though. 🤔 Is there a simpler solution to accomplish this task?
cc @rogermparent @shcheklein

rogermparent:
> 1. Move download links to an easier to access file.

Agreed, this will be needed to access them from multiple points. A JSON file with all the different download links should do.

> 2. Write a bash script that loops through links checking if links are correct, assigning answer to a variable. I found a way to do that with `curl`:
> [...]
> This solution feels kind of overcomplicated to me though. thinking Is there a simpler solution to accomplish this task?

Agreed that a bash script isn't the way we want to go, and implementing the actual link checking logic is a nightmare which should be avoided, although a lesser nightmare because all our links are AWS so they should behave consistently.

[Lychee](https://github.com/lycheeverse/lychee) is a relatively popular open-source family of projects built around flexibly checking links, and though it's billed for HTML and MD, it also works in JSON files! I think we can get what we want here by:

1. Making a json file that keeps all our DVC download links, wired into the Gatsby site to feed the current components.

2. Adding a GitHub Workflow that points [the Lychee Action](https://github.com/lycheeverse/lychee-action) to that file, checking all the DVC download links. I think this specific check would be best targeted once per day on master, and on master merges- doing it on every PR seems a bit much, and the diff link checker should handle it, though I suppose in its current state it's unwise to put too much trust it in that project right now.

shcheklein:
Q: we have already our own link checker running every day - does it detect it now?

My2cs - custom Github action with a CLI command to try all specific links in one sounds fine to me also.

[...]

PR #2660, opened 2021-07-22
Add Github action for checking dvc download links

julieg18:
* This github action checks if links that include `/download` all work correctly with `lychee`(mentioned in #2579)
* Even though this action doesn't check for the amount of download links, I suppose it works for a first iteration and we could work on checking for amount at a later date.
* You can see an example of the action by looking at previous commits([Example](https://github.com/iterative/dvc.org/pull/2660/checks?sha=3ef3587bc50f237e43472fd69c51d7aebd27d8ce))


.github/workflows/download-link-check.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  deployment_status:
+  schedule:
+    - cron: '0 1-23 * * *'
+
+jobs:
+  run:
+    name: Initialize
+    runs-on: ubuntu-latest
+    if:
```
julieg18, line 12: I hope I wrote this `if` logic correctly 😅

.github/workflows/download-link-check.yml
```diff
@@ -0,0 +1,30 @@
...
+      github.event_name == 'schedule' || (github.event.deployment.ref !=
+      'master' && github.event.deployment_status.state == 'success')
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download Link Checker
+        uses: lycheeverse/lychee-action@v1.0.8
```
julieg18, line 18: From what I can tell, `lychee` places the output file inside a `tmp` folder which gets deleted. Without this output file, I'm not sure how we could get the amount of links without creating our own version of `lychee-action`  that doesn't delete the output file.🤔

rogermparent (changes requested):
Decent proof of concept, it seems like it'll work! We'll need some minor changes before it actually handles all our cases though, doesn't look like PR checks will work at all currently.

.github/workflows/download-link-check.yml
julieg18: Yes, if errors are found in the links, `lychee-action` creates a `./lychee/out.md`. But since the file is only created if errors are found, we can't use that file to check for amount of links, since we'd have to check for amount of links when there are no errors found.

.github/workflows/download-link-check.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          args:
+            --verbose https://dvc.org/ --base-url https://dvc.org/ --include
```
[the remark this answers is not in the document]
julieg18: Oh, you're right! Totally forgot about updating the arguments!

.github/workflows/download-link-check.yml
julieg18: Agreed! I'll create two workflows!

julieg18:
Not sure why but when I attempted to merge this branch with master, I'm told I need to prettify the `DownloadButton` file if I want to commit the merge. But when I prettify the `DownloadButton` file and pushed the merge to remote the `commit` test [failed](https://app.circleci.com/pipelines/github/iterative/dvc.org/9955/workflows/fd547fca-835a-4bfa-bdcd-75019b7b29ee/jobs/10014) telling me I needed to undo the what the pre commit hook told me to do. I took off the merge and pushed the code for now.

My output when I attempted to commit the merge to master:

image

julieg18:
Sorry, didn't mean to request a review. Still working on properly failing the deployment file!

julieg18:
Alright, that took multiple testing attempts but pull request is finally ready for another review!

.github/workflows/download-link-check-deploy.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: lycheeverse/lychee-action@v1.0.8
+        with:
+          args:
+            --verbose "${{ github.event.deployment.payload.web_url }}"
+            --base-url "${{ github.event.deployment.payload.web_url }}"
+            --include /download/
+
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Check for Error Report
```
julieg18, line 25: There's probably a better way to fail this action if the file exists 🤔

jorgeorpinel:
> when I attempted to merge this branch with master, I'm told I need to prettify the DownloadButton file if I want to commit the merge. But when I prettify the DownloadButton file and pushed the merge to remote the commit test failed telling me I needed to undo the what the pre commit hook told me to do

@julieg18 have these problems persisted? Recently Prettier was enabled for all files so may be related to that? We could roll back some of that if needed. See https://github.com/iterative/dvc.org/pull/2624

julieg18:
> @julieg18 have these problems persisted? Recently Prettier was enabled for all files so may be related to that? We could roll back some of that if needed. See #2624

Yep, but I don't think its because of #2624 though. I think it has something to do with circleci possibly installing a different version of prettier than the one we use locally. To fix the issue, we'd might have to update our circleci config or our package.json packages. @rogermparent, what do you think?

rogermparent:
> Yep, but I don't think its because of #2624 though. 

Agreed, #2624 seems to be about restyled, which as far as I can tell isn't part of the CircleCI commit hook

> I think it has something to do with circleci possibly installing a different version of prettier than the one we use locally. To fix the issue, we'd might have to update our circleci config or our package.json packages. @rogermparent, what do you think?

I'm not quite sure, to be honest, at first glance like a conflict between prettier and eslint to me. Do you know which CircleCI step failed? `lint-ts` and `format-check` are run as separate steps so knowing that can help diagnose this issue.

julieg18:
> Do you know which CircleCI step failed? `lint-ts` and `format-check` are run as separate steps so knowing that can help diagnose this issue.

`lint-ts`. I can push another merge to `master` if it helps!

julieg18:
> `lint-ts`. I can push another merge to `master` if it helps! Here is a link to the [failed commit check](https://app.circleci.com/pipelines/github/iterative/dvc.org/9989/workflows/19b732b2-4024-44bd-8bfe-ec090d87f932/jobs/10048). Also, if I run `lint-ts` locally, I don't see any errors about `DownloadButton/index.tsx`.

Update! I reinstalled my node_modules, now `lint-ts` is now working the same on circle-ci and my local version! My local version of prettier is for some reason incorrectly formatting `DownloadButton/index.tsx` 🤔

rogermparent (approved):
Looks good to me!

.github/workflows/download-link-check-schedule.yml
```diff
@@ -0,0 +1,26 @@
+name: DVC Download Link Check
+
+on:
+  schedule:
+    - cron: '0 1-23 * * *'
```
rogermparent, line 5: Wouldn't `1-23` run on only 23 hours, skipping midnight? Practically the same, I suppose, but I think `0-23` is needed for truly every hour.

julieg18: Yes, that's a typo. Thanks for spotting that!

PR #2665, opened 2021-07-23
Update modify.md with params for #6359

meierale:
https://github.com/iterative/dvc/pull/6359 adds some additional parameters to the azure config. This PR is the documentation for those.


jorgeorpinel:
Thanks @meierale . I moved this information to a hidden note for Windows users above all of the custom auth params. That's what this affects right?

@isidentical PTAL when you can (please approve if correct).

isidentical (approved):
Looks good to me

PR #2666, opened 2021-07-23
docs: copy edit DVC 2.0 notes

jorgeorpinel:
Follow up to https://github.com/iterative/dvc.org/pull/2663 (copy edit)

"Since" typically refers to a specific time. Reads a but strange for a version number, I think.

### DIFF
--- a/content/docs/command-reference/exp/index.md
+++ b/content/docs/command-reference/exp/index.md
@@ -1,6 +1,6 @@
 # experiments
 
-⚠️ This feature is available since DVC 2.0 ⚠️
+⚠️ This feature is available on DVC 2.0 and above ⚠️
 
 A set of commands to generate and manage <abbr>experiments</abbr>:
 [run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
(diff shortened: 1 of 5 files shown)


shcheklein:
Since can refer to version (I think it's pretty common). E.g. @since in Java https://stackoverflow.com/questions/578363/javadoc-version-and-since . `on ...` doesn't sound familiar to me at all to be honest.

shcheklein:
No strong opinion at all though :) the phrase becomes longer ... but it's still clear I think, as the previous one

shcheklein (approved):

jorgeorpinel:
> E.g. @SInCE in Java

I think `@since` makes sense because it's code and you need short keywords that don't necessarily read perfectly in natural language.

A shorter way could be "New in version x.y.z" notes (taken from the Python lib ref) e.g. https://docs.python.org/3/library/functions.html

daavoo:
> A shorter way could be "New in version x.y.z" notes (taken from the Python lib ref) e.g. https://docs.python.org/3/library/functions.html

I like this :point_up:

shcheklein:
Yep, sounds good to me too.

shcheklein:

jorgeorpinel:
Done.

PR #2685, opened 2021-08-04
fix file hash examples in internal-files.md #2678

### LINKED ISSUE #2678: Possible typo within internal-files.md
Though all md5 hashes should be 32 characters long, I happened to find out that there is a hash with 34 characters in the docs.

https://github.com/iterative/dvc.org/blob/master/content/docs/user-guide/project-structure/internal-files.md#directories

Looks like `0b40427ee0998e9802335d98f08cd98f` should be changed to sth like `0b40427ee0998e9802335d98f08cd9` (2 characters shorter). May I try to regenerate the hashes for the docs in the `internal-files.md` ?

seungwonpark:
Closes #2678

### DIFF
--- a/content/docs/user-guide/project-structure/internal-files.md
+++ b/content/docs/user-guide/project-structure/internal-files.md
@@ -107,12 +107,12 @@ The resulting cache dir looks like this:
 
 ```dvc
 .dvc/cache/
-├── 19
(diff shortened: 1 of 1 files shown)


jorgeorpinel (approved):
Thanks @seungwonpark !

PR #2688, opened 2021-08-04
Update Link and DownloadButton components

### LINKED ISSUE #2668: Allow `Link` to be able to opt out of "pre redirect"
* Currently the `Link` component automatically [checks](https://github.com/iterative/dvc.org/blob/master/src/components/Link/index.tsx#L107) if `href` prop has a redirect and updates href accordingly. 
* We want to be able to stop the `href` from being updated if needed, like for our download links for example([mentioned](https://github.com/iterative/dvc.org/pull/2650#issuecomment-884597137) in pull request #2650).

julieg18:
* adds `optOutPredirect` prop to `Link`, stopping links from being pre redirected if desired
* ~~have Link render on mount, stopping any href changes due to hydration(fixes "CML" link shows 404 page bug mentioned in #2659)~~
* converts DownloadButton download absolute links to relative links

Fixes #2668

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -12,6 +12,7 @@ export type ILinkProps = {
   target?: undefined | '_blank'
   state?: unknown
   scrollOptions?: object
+  optOutPreRedirect?: undefined | true
 } & React.AnchorHTMLAttributes<HTMLAnchorElement>
 
 const PROTOCOL_REGEXP = /^https?:\/\//
@@ -23,6 +24,7 @@ const ResultLinkComponent: React.FC<ILinkProps> = ({
   children,
   rel,
   target,
+  download = false,
   ...restProps
 }) => {
   // Handle all situations where a basic `a` must be used over Gatsby Link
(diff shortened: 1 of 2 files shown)


src/components/Link/index.tsx
```diff
@@ -114,7 +136,9 @@ const Link: React.FC<ILinkProps> = ({ href, scrollOptions, ...restProps }) => {
     }
   }
 
-  return <ResultLinkComponent href={href} {...restProps} onClick={onClick} />
+  return mounted ? (
+    <ResultLinkComponent href={href} {...restProps} onClick={onClick} />
+  ) : null
```
rogermparent, line 141: I'm not so sure about this, does it prevent links from being seen without JS- e.g. from the `lychee` checks we have planned? Do the issues still persist without this state and conditional?

julieg18: You're right! `lychee` checks would break and would stop links from being seen without JS! I updated the code, replacing `null` with the jsx. Essential just telling the component to re render on mount. There's probably a cleaner way to re render a component though 🤔

rogermparent (approved):

PR #2654, opened 2021-07-21
guide: expand Experiments guide

jorgeorpinel:
Restructures the guide into an intro and a few pages.

Per #2548 (main files changed linked in each item below)

- [x] Copy edit https://github.com/iterative/dvc.org/pull/2618 ([file changes](https://github.com/iterative/dvc.org/pull/2654/files?authenticity_token=k53ZEbxYcI77Rntmd%2F88fdbqSzBJQG9kq41%2B3sW2npFBko0txa9AdJPbx%2Fgf9W6FzJhWIlvCitKufyRYsGKPTg%3D%3D&file-filters%5B%5D=.json#diff-1b150c58a1c943d3149b36bf01defdcac47b31b74eed095adad8298827c3d9c4)) via https://github.com/iterative/dvc.org/pull/2711 and https://github.com/iterative/dvc.org/pull/2719 (nested into this PR)
- [x] ~~Review~~ Copy edit [Checkpoints guide](https://github.com/iterative/dvc.org/pull/2654/files#diff-b9806bbdaf8e3ac58a0a85b160ac6853229c83336abf60e24e2556c3ae054186) for now. ~~Can it be broader yet shorter?  
  Move deep details from https://dvc.org/doc/command-reference/exp/run#checkpoints
  Done in https://github.com/iterative/dvc.org/pull/2753 (not nested)~~
  Extracted to https://github.com/iterative/dvc.org/issues/2769
- [x] Extracted **DVC Experiments** section [from index](https://github.com/iterative/dvc.org/pull/2654/files?authenticity_token=k53ZEbxYcI77Rntmd%2F88fdbqSzBJQG9kq41%2B3sW2npFBko0txa9AdJPbx%2Fgf9W6FzJhWIlvCitKufyRYsGKPTg%3D%3D&file-filters%5B%5D=.json#diff-128fef4265ae36168840329951fb19fd9ce86c7fa965fc244bd024f3ec96346e) into [it's own page](https://github.com/iterative/dvc.org/pull/2654/files?authenticity_token=k53ZEbxYcI77Rntmd%2F88fdbqSzBJQG9kq41%2B3sW2npFBko0txa9AdJPbx%2Fgf9W6FzJhWIlvCitKufyRYsGKPTg%3D%3D&file-filters%5B%5D=.json#diff-17f833349dd6c62cd54dfedc4c73234be01aab006300e4ef9c2879fc1b1d4e72).
- [x] Related links and some other copy edits

Potentially:

[...]

PR #2631, opened 2021-07-14
guide: Cleaning Up Experiments

iesahin:
Adds a section to UG:Experiment Management about cleaning up the experiments. 

We are updating the GS:Experiments, we remove the information on `dvc exp gc`
and `dvc gc` in GS:Experiments. We need a new place for this information in a
rather detailed fashion. 

- Explains `dvc exp gc`
- Explains how to remove experiments from remotes
- Explains how to remove associated DVC-cached elements

Closes #2592

## Waiting for

- [x] iterative/example-repos-dev#68
---
## Follow-Up

The follow-up issue to this PR is #2777

Issue #2548, opened 2021-06-10
guide: Experiment Management (phase 1)


jorgeorpinel:
We have some overview of Experiments in https://dvc.org/doc/user-guide/experiment-management and now a Checkpoints guide under that (which is at a very different level). We should probably

- [x] split the sections of that index into a page each (see https://github.com/iterative/dvc.org/pull/2654); and
- [x] extract technical details from https://dvc.org/doc/command-reference/exp (esp. `run`) to guides
- [ ] extract some? examples from https://dvc.org/doc/command-reference/exp to guides
- [x] review https://dvc.org/doc/user-guide/experiment-management/checkpoints for consistency
- [x] cover more checkpoints scenarios; at least #2568 (see https://github.com/iterative/dvc.org/pull/2753)
- [x]  #2367
- [x] ~~#2675~~ #2690
- [x] #2737
- [x] #2736
- ~~#2844 (my be~~ done by the 2 above)
- [x] #2592 
- \+ General points from https://github.com/iterative/dvc.org/issues/2995 ?

> Next: https://github.com/iterative/dvc.org/issues/2911

jorgeorpinel:
What priority should this have @shcheklein @dberenbaum ?

dberenbaum:
It might make sense to wait until after we finish the new get started sections for experiments and maybe the use case if we also do that one. At that point, we might have a better idea of what to do with these.

What do you think @jorgeorpinel?

PR #2690, opened 2021-08-04
guide: Running Experiments

### LINKED ISSUE #2675: guide: Running Experiments
We are removing some parts of the experiments trail that may be better put into a UG document. 

Related to #2548 

A guide dedicated to running experiments should contain:

- Running experiments
- Running experiments in parallel
- Managing experiments queue
- Naming experiments

iesahin:
This adds a new section to the UG for running experiments. 

Closes #2675

### DIFF
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -1,0 +1,193 @@
+# Running Experiments
+
+## Motivation
+
+Machine Learning and Data Science projects usually involve experimentation.
+Motivation of these experiments can range from finding good hyperparameters to
+checking data and concept drift. DVC 2 introduced a new set of commands to
+manage experiments with minimum boilerplate. It allows to run experiments
+defined by pipelines, track their associated data and model files, set
+parameters for each, push experiment parameters and code to Git remotes without
+committing them, create branches and persist them.
+
+In this part of the DVC User's Guide we will explain how DVC runs the
+experiments, setting parameters, running the experiments in parallel, or in
+temporary directories.
+
+If this is the first time you're introduced to Machine Learning experimentation,
+it may be better to start with the
+[Experiment's Trail](/doc/start/experiments/).
+
+## Parameters
+
+Before running an experiment, you'll probably want to make modifications such as
+data and code updates, or <abbr>hyperparameter</abbr> tuning. For the latter,
+you can use the `--set-param` (`-S`) option of this command to change
+`dvc param` values on-the fly.
+
+### Setting Multiple Parameters for Experiments
+
+### How DVC updates the parameters?
+
+## The Experiments Queue
+
+The `--queue` option lets you create an experiment as usual, except that nothing
+is actually run. Instead, the experiment is put in a wait-list for later
+execution. `dvc exp show` will mark queued experiments with an asterisk `*`.
+
+> Note that queuing an experiment that uses checkpoints implies `--reset`,
+> unless a `--rev` is provided (refer to the previous section).
+
+Use `dvc exp run --run-all` to process the queue. This is done outside your
+<abbr>workspace</abbr> (in temporary dirs in `.dvc/tmp/exps`) to preserve any
+changes between/after queueing runs.
+
+💡 You can also run a single experiment outside the workspace with
+`dvc exp run --temp`, for example to continue working on the project meanwhile
+(e.g. on another terminal).
+
+> ⚠️ Note that only tracked files and directories will be included in
+> `--queue/temp` experiments. To include untracked files, stage them with
+> `git add` first (before `dvc exp run`). Feel free to `git reset` them
+> afterwards. Git-ignored files/dirs are explicitly excluded from runs outside
+> the workspace to avoid committing unwanted files into experiments.
+
+<details>
+
+### ⚙️ How are experiments queued?
+
+A custom [Git stash](https://www.git-scm.com/docs/git-stash) is used to queue
+pre-experiment commits.
+
+</details>
+
+Adding `-j` (`--jobs`), experiment queues can be run in parallel for better
+performance (creates a tmp dir for each job).
+
+⚠️ Parallel runs are experimental and may be unstable at this time. ⚠️ Make sure
+you're using a number of jobs that your environment can handle (no more than the
+CPU cores).
+
+> Note that each job runs the entire pipeline (or `targets`) serially. DVC makes
+> no attempt to distribute stage commands among jobs. The order in which they
+> were queued is also not preserved when running them.
+
+### Adding Experiments to the Queue
+
+### Running Experiments in Parallel
+
+### Cleaning Up the Experiment Queue
+
+## Checkpoints
+
+To track successive steps in a longer or deeper <abbr>experiment</abbr>, you can
+register checkpoints from your code. Each `dvc exp run` will resume from the
+last checkpoint.
+
+First, mark at least stage <abbr>output</abbr> with `checkpoint: true` in
+`dvc.yaml`. This is needed so that the experiment can resume later, based on the
+<abbr>cached</abbr> output(s) (circular dependency).
+
+⚠️ Note that using `checkpoint` in `dvc.yaml` makes it incompatible with
+`dvc repro`.
+
+Then, use the `dvc.api.make_checkpoint()` function (Python code), or write a
+signal file (any programming language) following the same steps as that
+function.
+
+You can now use `dvc exp run` to begin the experiment. All checkpoints
+registered at runtime will be preserved, even if the process gets interrupted
+(e.g. with `[Ctrl] C`, or by an error). Without interruption, a "wrap-up"
+checkpoint will be added (if needed), so that changes to pipeline outputs don't
+remain in the workspace.
+
+Subsequent uses of `dvc exp run` will continue from the latest checkpoint (using
+the latest cached versions of all outputs).
+
+<details>
+
+### ⚙️ How are checkpoints captured?
+
+Instead of a single commit, checkpoint experiments have multiple commits under
+the custom Git reference (in `.git/refs/exps`), similar to a branch.
+
+</details>
+
+List previous checkpoints with `dvc exp show`. To resume from a previous
+checkpoint, you must first `dvc exp apply` it before using `dvc exp run`. For
+`--queue` or `--temp` runs (see next section), use `--rev` instead to specify
+the checkpoint to continue from.
+
+Alternatively, use `--reset` to start over (discards previous checkpoints and
+their outputs). This is useful for re-training ML models, for example.
+
+## The Pipeline
+
+### Running a Single Stage
+
+### Running all Pipelines
+
+### Interactive Reproduction
+
+### Recursive Dependencies
+
+### Improvements over `dvc repro`
+
+## Git and Experiments
+
+Experiments are custom
+[Git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+(found in `.git/refs/exps`) with a single commit based on `HEAD` (not checked
+out by DVC). Note that these commits are not pushed to the Git remote by default
+
+## Note on Experiment Names
+
+Each experiment creates and tracks a project variation based on your
+<abbr>workspace</abbr> changes. Experiments will have an auto-generated name
+like `exp-bfe64` by default, which can be customized using the `--name` (`-n`)
+option.
+
+When you create an experiment, DVC generates a hash value from the contents of
+the experiment. This is shown when you use `--queue` option, e.g.,
+
+```dvc
+$ dvc exp run --queue -S model.conv_units=32
+Queued experiment '6518f17' for future execution.
+```
+
+After _running_ the experiment, DVC uses another auto-generated name to refer to
+the experiment. Typically these start with `exp-`, and can be set via
+`--name / -n` option of `dvc exp run`. So when you add an experiment by setting
+the name, you can see the hash value as _queued experiment_:
+
+```dvc
+$ dvc exp run --queue --name cnn-512 -S model.conv_units=512
+Queued experiment '86bd8f9' for future execution.
+```
+
+In `dvc exp show` you can see both of these names:
+
+```dvc
+$ dvc exp show --no-pager --no-timestamp \
+      --include-metrics loss --include-params model.conv_units
+
+┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
+┃ Experiment              ┃ loss    ┃ model.conv_units ┃
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
+│ workspace               │ 0.23534 │ 64               │
+│ 3973b6b                 │ -       │ 16               │
+│ ├── aeaabb0 [exp-cb13f] │ 0.23534 │ 64               │
+│ ├── d0ee7ce [exp-5dccf] │ 0.23818 │ 32               │
+│ ├── 1533e4d [exp-88874] │ 0.24039 │ 128              │
+│ ├── b1f41d3 [cnn-256]   │ 0.23296 │ 256              │
+│ ├── 07e927f [exp-6c06d] │ 0.23279 │ 24               │
+│ ├── b2b8586 [exp-2a1d5] │ 0.25036 │ 16               │
+│ └── *86bd8f9            │ -       │ 512              │
+└─────────────────────────┴─────────┴──────────────────┘
+```
+
+When an experiment is not run yet, only the former hash value is shown.
+
+You can refer to the experiment in `dvc exp apply` or `dvc exp branch` after
+running the experiment with the name starting with `exp-`, or the name you have
+supplied with `dvc exp run --name`.


iesahin:
This and #2631 are prerequisites for #2574. We are removing some content from start:Experiments and these guides are aimed at to provide that information. @jorgeorpinel

iesahin:
Could you please review the structure and scope of this guide? 

@shcheklein @jorgeorpinel @casperdcl @dberenbaum

dberenbaum:
Thanks @iesahin!

My thoughts from skimming the doc:
1. The pipeline section seems too deep to me. I think the guide could link to other parts of the docs more liberally. It makes sense to introduce pipelines to make this guide self-contained, but the level of detail about targets and different options that can be passed seems unnecessary.
2. There's a section on params, but I don't see much about metrics, nor explanation that experiments can also track changes to data, code, etc.
3. Are we keeping a separate checkpoints guide? Do we need a section for it in this doc?
4. The "Git and Experiments" and "Experiment Names" seem less about running experiments and more about tracking experiments. They might fit better under a separate guide like "Tracking Experiments," which could also dive more into `dvc exp show`, `dvc exp apply`, and `dvc exp branch`, or those commands might need to be fleshed out in this guide. I'm not totally clear on the intended scope of this particular guide.

jorgeorpinel:
> This and #2631 are prerequisites for #2574.

I think this should not be the case. Let's find a way to structure our work so that one already heavy task (in this case Get Started rewrite) doesn't span multiple larger tasks as pre-requisites. Probably it's fine to just remove all this info from GS for now and merge those changes while we work on these guides one by one. Meaning please don't see these as pre-reqs to finish/focus on #2574, I think that should be the priority.

As for the guides, I already started reviewing #2631 since that PR came earlier but it seems like this one is more important right? If possible mark that other one as draft or even close it while we work on this one. WDYT @iesahin ?

#logistics

> I'm checking structure and scope now ⏳

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+parameters for each, push experiment parameters and code to Git remotes without
+committing them, create branches and persist them.
+
+In this part of the DVC User's Guide we will explain how DVC runs the
+experiments, setting parameters, running the experiments in parallel, or in
+temporary directories.
+
+If this is the first time you're introduced to Machine Learning experimentation,
+it may be better to start with the
+[Experiment's Trail](/doc/start/experiments/).
```
jorgeorpinel, line 19: All of this is already in the index (**Experiment Management** main page), probably no need here.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+command is considered the output of experiment. Additionally DVC tracks
+artifacts from intermediate stages, and doesn't reproduce them if their
+dependents have not changed.
+
+DVC pipelines are a detailed topic that we cover in [Get Started][gs-pipelines]
+and the [User's Guide][ug-pipeline-files]. Here we assume that there is already
+a pipeline defined in `dvc.yaml` file in the <abbr>project</abbr>.
+
+[gs-pipelines]: /doc/start/data-pipelines
+[ug-pipeline-files]: /doc/user-guide/project-structure/pipelines-files
```
jorgeorpinel, line 38: I'm not seeing a clear connection between needing to explaining pipelines and running experiments. It could just be a single word linked to https://dvc.org/doc/user-guide/project-structure/pipelines-files or https://dvc.org/doc/command-reference/dag . Clearly we need a separate guide about pipelines or at least a basic concept page for now, but that's out of scope.

p.s. We'll get to it so not much need to worry about it here and when we do we will update all these links too.

jorgeorpinel: I see the following section already has the link I was thinking about:

image

I think the guide can start there. If needed one short paragraph for background could be OK before that IMO.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+
+### Running the pipeline
+
+The pipeline defined in `dvc.yaml` file is run with default settings using:
+
+```dvc
+$ dvc exp run
```
jorgeorpinel, line 45: I wonder if we want to avoid the concept of pipeline in this guide actually. We'll probably have to mention and link it (as discussed above) but let's try to keep the language around experimentation? Something like "The current version/variation of your DVC project can be run with `dvc exp run`..."

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+If all dependencies and the experiment output are not changed and available in
+the workspace, the `dvc exp run` doesn't rerun the commands.
+
+For a pipeline composed of `extract`, `transform`, `train`, `evaluate`, if a
+dependency of `train` stage has changed, the dependent stages (`evaluate`) are
+also run.
+
+### Specifying Targets
+
+By default `dvc exp run` uses `dvc.yaml` file in the current directory. You can
```
jorgeorpinel, line 57: Actually it should use all the dvc.yaml files in the repo by default. See https://dvc.org/doc/command-reference/repro#options (BTW this may be wrong in the help output, if so feel free to open a core repo issue or contribute a change).

jorgeorpinel: p.s. I'm not 100% sure about this.

jorgeorpinel: BTW since we'll have a guide should we remove many of the details from that cmd ref **Options** section I linked to? `targets` is not even an option (flag), but rather a (required) argument.

In fact this idea of removing details from the cmd ref can be generalized but Idk if we want to do as part of this same PR (maybe after some more review iterations) or in a different one. WDYT?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+`my-sub-project/sub-module/...` to run all the pipelines.
+
+### Improvements over `dvc repro`
+
+The classical way of running the pipeline is using `dvc repro`. DVC introduced
+`dvc exp` as an _experiment management_ feature in version 2.0. _Experiment
+management_ involves more than executing the pipelines.
+
+In the following sections, we describe the features that make `dvc exp run` more
+suitable for machine learning experimentation.
```
jorgeorpinel, line 128: Good note but probably not needed or maybe just as a note (md quote block) somewhere else. I think we just need the "New in DVC 2.0" line at the top of this (and all Exp Mgmt) guide(s). Like in https://dvc.org/doc/command-reference/exp

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+last checkpoint.
+
+Checkpoints provide a way to train models iteratively, keeping the metrics
+associated with each epoch.
+
+### Creating Checkpoints
+
+#### Adding Checkpoints to the Pipeline
+
+There are various ways to add checkpoints to a project. In common, these all
```
jorgeorpinel, line 361: We have a separate Checkpoints guide. Let's only link there from here?

The work you've done here could be merged into that guide if it's a significant improvement though, but as a separate PR I think (BTW I'm reviewing that guide too).

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,569 @@
...
+the custom Git reference (in `.git/refs/exps`), similar to a branch.
+
+</details>
+
+## Git and Experiments
+
+Experiments are custom
+[Git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
+(found in `.git/refs/exps`) with a single commit based on `HEAD` (not checked
+out by DVC). Note that these commits are not pushed to the Git remote by default
```
jorgeorpinel, line 517: All of this should probably be in the section index (https://dvc.org/doc/user-guide/experiment-management) or maybe it could work as the document's short intro/background.

jorgeorpinel: Or as PART of the intro, but as a hidden section (a bit too technical).

jorgeorpinel:
Left some specific mid-level comments above. ☝️ My review overlaps a lot with @dberenbaum's comments. In summary I find this too long in it's present form.

Don't get me wrong your holistic approach is great @iesahin but I don't think we need all this info in a single page. Guides can focus on a topic without having to provide all the context. Basically, the sections I didn't comment much on (yet) constitute the main material we need here IMO (plus maybe a bit on metrics and code like @dberenbaum suggested).

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: I think to incorporate this text into the index page.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: IMO this guide should be a bit self-contained. Without pipelines, the user cannot run the experiments. 

Learning/linking pipelines elsewhere is asking the reader to learn `dvc repro`, possibly `dvc run`, `dvc stage` and related commands before coming to `dvc exp`. This summary is aimed towards to "save" the user from those topics and focus only on "pipelines in experiments." Some parts of it may be a bit too much as @dberenbaum suggested, but linking other parts of the documentation causes reader to divert attention.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: We don't want to avoid the concept of pipelines in the UG. Pipelines are an integral part of experimentation and we would like to present the full details to the user.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: As we progress to develop the guide, we can remove the repeating sections in the command reference. Command samples with a single sentence may remain in the cmd ref, but longer discussions around examples should be in the UG I think.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: Users of former versions may ask "what's the difference between this and that?" Also someone just learning about DVC and had met `dvc repro` just a few pages before may wonder why there are two commands doing a similar job.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: That guide is more like a tutorial and I think better be replaced by this section. It also has to tell the experiments because these two topics cannot be clearly separated. When we have a GS page for the checkpoints, we can remove that page I think.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: It may be separated into a section by adding some content from https://dvc.org/blog/experiment-refs I believe. It's a bit alien to this section, you're correct.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: I tested `dvc exp run`, `dvc exp run --all-pipelines`, `dvc exp run dir/dvc.yaml` and the first of these doesn't run the pipeline in another directory. Even `dvc exp run -R .` doesn't run.

iesahin: I think `dvc exp run` has no feature parity with `dvc repro` when it comes to recursive directories or multiple `dvc.yaml`s.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,570 @@
...
+specify other directories or pipeline elements to run. These are specified as
+the last element of the command.
+
+Pipelines can be of the following:
+
+- A particular `dvc.yaml` file: `dvc exp run my-project/dvc.yaml`
+- A stage from the default `dvc.yaml` file: `dvc exp run extract`
+- A stage from a specific `dvc.yaml` file:
+  `dvc exp run my-project/dvc.yaml:extract`
+- A set of stages from the default `dvc.yaml` file:
```
iesahin, line 68: I'm removing  `--recursive` and `--glob` from these docs until iterative/dvc#6456 and iterative/dvc#6457 are resolved.

iesahin: Opened an issue to restore them in #2730

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: This can be a summary of the differences from `dvc repro` but I'm not sure. Keeping this as is.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: What do you mean self-contained though?  The docs are self-contained but we don't need to cover every feature in every guide. There are 2 chapters in this guide before you even get to the topic (running experiments).

> linking pipelines elsewhere is asking the reader to learn dvc repro, possibly dvc run, dvc stage

No, why? Again, the fact that those other concepts don't have a guide yet doesn't mean we have to write them all at the same time. Let's focus on running exps for now. We'll get to writing about pipelines later, meanwhile the cmd ref is where we can link.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > We don't want to avoid the concept of pipelines in the UG

Again, we can have a separate pipeline guide in the UG. I'm talking about this one guide which should be about running experiments. I think we have to limit the scope of our guides content or we'll be writing very long tutorials every time which overlap 50% or more with each other.

I'm not so sure that pipelines are an integral part of experimentation, why? You can experiment on a single stage. They're meant as separate features AFAIK.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: I think I was wrong about the default target being all dvc.yaml files. I know that repro does check all the dvc.yaml files in the repo for consistency though, but only executes pipelines included in the targets.

> As we progress to develop the guide, we can remove the repeating sections in the command reference

Agree, that's what I meant. My question is when. I think if the changes are not major it could be included in this same PR so we're done with it. But a separate one is also OK (add a checkbox?)

> dvc exp run has no feature parity with dvc repro when it comes to recursive directories or multiple dvc.yamls

`exp run` should accept all the same arguments and options as `repro` and behave the same way. See https://dvc.org/doc/command-reference/exp/run#options

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: I don't see how telling readers the history of when each command was developed and why is helpful. This isn't about different versions or improvements either since both commands exist at the same time and behave the same way (one is not better than the other).

I think bringing this up has the potential to confuse more people than those who happened to wonder about this. Also, the explanation given "Experiment management_ involves more than executing the pipelines" doesn't explain what are the improvements.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Yes it's like a tutorial so it needs revising, not discarding it and replacing it by a relatively short section here. Checkpoints definitely deservers a guide don't you think? But that's a separate problem, not necessarily in the scope of this "running experiments" guide.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > adding some content from https://dvc.org/blog/experiment-refs

The reason it was written as a blog was to avoid those implementation details here 😅

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: I think it's OK to document the intended behavior but OK up to you.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: > exp run should accept all the same arguments and options as repro and behave the same way. See https://dvc.org/doc/command-reference/exp/run#options

Yes, _should_ :) 

Before the bug reports (iterative/dvc#6456 and iterative/dvc#6457), I had no strong opinion on this but from the development and future directions, I think it's better to separate `dvc exp run` from `dvc repro` in documentation. Some elements of `repro` may be obsolete in `exp run`, or vice versa. Considering these as two distinct commands seems more prudent to me.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: By _self-contained_, I mean to cover the necessary part of the pipelines within the context of `exp run`. The pipelines section doesn't cover how to create or update them, for example, I think that's not relevant here. But _running_ pipelines is what _dvc exp run_ does, and without a proper summary here, the user is directed to all sorts of details. 

Also, to learn the pipelines itself, one needs to learn `dvc stage`, `dvc repro`, `dvc run`, etc. I would like the reader not to be distracted with other commands within the experiments guide.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: `dvc exp run` is running the pipelines, and I really would like to have experiments without pipelines, but we have this restriction. Which content does this section repeat, apart from `dvc exp run`'s cmd ref?

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: We are not telling the history here, there are two (or including `dvc run`, three) commands that run the pipeline and I believe it's a common topic to discuss the differences.  Why DVC team bothered to add another command that seemingly does the same thing? When I should use one and not the other?

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: Checkpoints are extensions to experiments, they cannot be discussed without first adding a stage, experimenting it, etc. That tutorial mostly deals with the setup of checkpoints. It doesn't even tell different ways to use it. I believe the content here covers more about the checkpoints itself than the tutorial.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: Why do we put extra-technical details in a blog?

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: We can't have a timeframe for the implementation. I'd remove these options from the command reference as well.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: Removing this section from here. Maybe we can tell elsewhere.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: If your edits to the index are done, I'm moving this. @jorgeorpinel

iesahin:
>     1. The pipeline section seems too deep to me. I think the guide could link to other parts of the docs more liberally. It makes sense to introduce pipelines to make this guide self-contained, but the level of detail about targets and different options that can be passed seems unnecessary.

Simplified the section a bit. I believe telling "`dvc exp run` takes the same options as `dvc repro`" makes the command always tied to `dvc repro`. We can change/extend/restrict the meaning of these parameters for `dvc exp run` in the future.

iesahin:
>     4\. The "Git and Experiments" and "Experiment Names" seem less about running experiments and more about tracking experiments. They might fit better under a separate guide like "Tracking Experiments," which could also dive more into `dvc exp show`, `dvc exp apply`, and `dvc exp branch`, or those commands might need to be fleshed out in this guide. I'm not totally clear on the intended scope of this particular guide.

Removed those sections. The first will probably have a dedicated section to tell the technical details, the second may better be in another section. @dberenbaum

iesahin:
>     3\. Are we keeping a separate checkpoints guide? Do we need a section for it in this doc?

The current guide is a nice document but runs like a tutorial. I believe we can modify it for the GS, or have a separate Tutorials section for it. Checkpoints cannot be told without experiments and I see them as natural extensions to experiments, so I put them here. We can have a dedicated section for the checkpoints too.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: May take a while since #2654 is the branch of a series of nested PRs. Maybe you can stash this for now or make a separate branch (I can incorporate it when ready)? Thanks

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > running pipelines is what dvc exp run does

Makes sense but I still think we don't need the concept of pipelines here necessarily. Anyway, I'll try to finalize/merge this and open a separate issue to discuss/improve if needed ⏳

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: p.s. if you could review #2719 that would help BTW!

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > Which content does this section repeat

It will overlap completely with a future data pipelines guide.

Resolving this for now...

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > a separate one is also OK (add a checkbox?)

Added a checkbox to the description of this PR for now. Resolving

> it's better to separate dvc exp run from dvc repro in documentation. Some elements of repro may be obsolete in exp run.

Let's wait until that's an official decision of the DVC product team.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Valid discussion but the guide should focus on running experiments. With the goal of moving on I think I'll remove this or make it into a note somewhere for now... ⏳

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > That tutorial mostly deals with the setup of checkpoints. It doesn't even tell different ways to use it

Again, it's separate problem to improve that other guide. If we have a checkpoints guide, let's put everything-checkpoints there and just mention & link from here 🙂

jorgeorpinel: p.s. working on that ⏳

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > I'd remove these options from the command reference as well.

Probably not necessary (haven't noticed many support cases about it at least) but feel free to!

Resolving

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: OK summarized in af9469ff

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Summarized in b88e34c5 for now but it's strange to see "running exps" as title and then it mainly talks about pipelines and running pipelines. confusing...

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Done in fd1d1c5f.

content/docs/user-guide/experiment-management/index.md
```diff
@@ -1,5 +1,15 @@
...
 
+<!--
+Machine Learning and Data Science projects usually involve experimentation.
+These experiments' goals can range from finding good hyperparameters to testing
+for data and concept drift. DVC 2.0 introduced a new set of commands to manage
+experiments with minimum boilerplate. It allows to run experiments defined by
+pipelines, track their associated data and model files, set parameters for each,
+push experiment parameters and code to Git remotes without committing them,
+create branches and persist them in Git.
+-->
```
jorgeorpinel, line 11: This is temporary ⏳

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,498 @@
...
+
+<details>
+
+#### How does DVC update parameters?
+
+DVC updates the parameters by parsing the parameters file and updating the
+specified value. Because of this it restricts the set of formats accepted to
+YAML 1.2, TOML, JSON and Python.
+
+When you specify a parameter, you can see its effects using `git diff`.
```
jorgeorpinel, line 174: I made this into a `<details>` thingy but actually I think it's better to remove, this has been explained before and not sure how useful the Git command hint is. Just noting here in case someone disagrees... ⏳

jorgeorpinel: Removed in 27a475ce.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,480 @@
...
+DVC allows to run queued experiments in parallel by specifying a number of
+execution processes (`--jobs`):
+
+```dvc
+$ dvc exp run --run-all --jobs 4
+```
+
+> Note that since each experiment runs in an independent temporary directory,
+> common <abbr>stages</abbr> across experiment pipelines are always run
+> separately even if they share param values or other dependencies.
```
jorgeorpinel, line 265: Is this correct? I thought all temp dirs share the same cache. Cc @pmrowla thanks

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
+# Managing Experiments
+
+Once you have defined and/or [run experiments] in your project, you can use
+several features of DVC to see, compare, reproduce, share, or remove them.
+
+[run experiments]: /doc/user-guide/experiment-management/running-experiments
+
+## Experiment names
+
+Experiments created with `dvc exp run` will have an auto-generated name like
```
jorgeorpinel, line 10: Extracted the whole Experiment names section into a new doc per @dberenbaum's suggestion.

This is temporary though, probably won't be included in the PR for merge... ⏳

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Done in d6b4d5cf. Just left the section on **running checkpoints** that links to the full guide.

jorgeorpinel (approved):
OK I think this looks good overall but would you like to take another quick look while we get a confirmation on https://github.com/iterative/dvc.org/pull/2690#pullrequestreview-735437914 and https://github.com/iterative/dvc.org/pull/2690#pullrequestreview-735438083 @dberenbaum ? Thanks

content/docs/user-guide/experiment-management/running-experiments.md
pmrowla: They are executed in their own workspaces, so stages are run separately.

They do share the original workspace's cache directory, so if run cache is available from a previous stage run, that cache entry would be reused to skip execution. But that's an implementation detail, and it's still true either way that the stage run is evaluated separately for each workspace.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,480 @@
...
+
+[persisted]: /doc/user-guide/experiment-management#persistent-experiments
+
+</details>
+
+Run them all one-by-one with the `--run-all` flag. The order of execution is
+independent of their creation order.
+
+```dvc
+$ dvc exp run --run-all
```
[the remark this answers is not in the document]
pmrowla: It's not really a proper queue. But w/the current implementation they are started in order of their creation, but if you are running more than one at a time there is no guarantee as to which will finish first.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+DVC keeps track of the dependency graph and runs only the stages with changed
+dependencies or missing outputs.
+
+> Example: for a pipeline composed of `prepare`, `train`, and `evaluate` stages,
+> if a dependency of `prepare` stage has changed, the downstream stages
+> (`train`, `evaluate`) are also run.
+
+### Running specific stages
```
dberenbaum, line 39: I'm still not clear why the guide should explain running specific stages, `--single-item`, `--all-pipelines`, and `--interactive` options below. There are too many different options about specifying pipelines or stages to explain them all (`--pipeline`, `--recursive`, etc.), and the explanations here don't add much beyond what would be in the command reference. If we want to explain pipeline concepts, we could have a pipelines guide. If we want to explain `exp run` and `repro` options separately, we could make changes in the command reference.  Why include these here?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+> experimentation, you may want to check the basics in
+> [Get Started: Experiments](/doc/start/experiments/) first.
+
+## The pipeline
+
+DVC relies on <abbr>pipelines</abbr> that codify experiment workflows (code,
+<abbr>stages</abbr>, <abbr>parameters</abbr>, <abbr>outputs</abbr>, etc.) in a
+`dvc.yaml` file. These contain the commands to run the experiments.
+
+> 📖 See [Get Started: Data Pipelines](/doc/start/data-pipeline) for an intro to
```
dberenbaum, line 17: Is this something that exists or will exist before this PR is merged?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+> Note that parameters are attached to experiments so you can view them together
+> with `dvc exp show` and `dvc exp diff`.
+
+To set more than one param for the same experiment, use the `-S` option multiple
+times, or supply a comma-delimited list to `-S`:
+
+```dvc
+$ dvc exp run -S learning_rate=0.001 -S units=128
+# is equivalent to
+$ dvc exp run -S learning_rate=0.001,units=128
```
dberenbaum, line 166: I don't believe that comma-separated parameters are supported?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+> ⚠️ Note that DVC doesn't check whether parameters given to `--set-param` are
```
dberenbaum, line 169: This is another case where we could prioritize this if it's important enough to be mentioned in the guide: https://github.com/iterative/dvc/issues/5477

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+## The experiments queue
+
+The `--queue` option of `dvc exp run` tells DVC to append an experiment for
+later execution. Nothing is actually run yet.
+
+<details>
+
+### How are experiments queued?
+
+Queued experiments are created similar to
```
dberenbaum, line 182: Since we aren't otherwise going into implementation details in this guide, why do so here? I worry it goes very deep without much context and is likely to confuse more than help.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+Queued experiment '1cac8ca' for future execution.
+$ dvc exp run --queue -S units=64
+Queued experiment '23660bb' for future execution.
+$ dvc exp run --queue -S units=128
+Queued experiment '3591a5c' for future execution.
+$ dvc exp run --queue -S units=256
+Queued experiment '4109ead' for future execution.
+```
+
+> `dvc exp show` will mark queued experiments with an asterisk `*`.
```
dberenbaum, line 202: This is no longer true, and it's probably best to stay away from trying to describe the output, especially for experiments, since it's not really stable yet. The table now has a column that lists "Queued" for the relevant experiments, which hopefully makes it self-evident.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+$ nohup dvc exp run --temp &
+[1] 30473
+nohup: ignoring input and appending output to 'nohup.out'
+```
+
+> The above example creates a `nohup.log` file in the original workspace with
+> the output of the DVC process.
+
+Note that Git-ignored files/dirs are explicitly excluded from queued/temp runs
+to avoid committing unwanted files into final ([persisted]) experiments.
```
dberenbaum, line 230: This is minor, but I don't think this should be strictly about final/persisted experiments. For example, someone can `dvc exp push` and expose sensitive info.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+</details>
+
+Run them all one-by-one with the `--run-all` flag. The order of execution is
+independent of their creation order.
+
+```dvc
+$ dvc exp run --run-all
+```
+
+> ⚠️ Note that only tracked files and directories will be included in
```
dberenbaum, line 243: This seems partially redundant with the note above about excluding git-ignored files. Could these two notes be combined in one place?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+> `--queue/temp` experiments. To include untracked files, stage them with
+> `git add` first (before `dvc exp run`) and `git reset` them afterwards.
+
+To remove all experiments from the queue and start over, you can use
+`dvc exp remove --queue`. 📖 See [Cleaning-Up Experiments][clean-up] for
```
dberenbaum, line 248: Is this going to exist by the time this PR is merged?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -0,0 +1,300 @@
...
+
+> Note that since each experiment runs in an independent temporary directory,
+> common <abbr>stages</abbr> across experiment pipelines are always run
+> separately even if they share param values or other dependencies.
+
+⚠️ Parallel runs are experimental and may be unstable at this time. ⚠️ Make sure
+you're using a number of jobs that your environment can handle (no more than the
+CPU cores).
+
+## Checkpoint experiments
```
dberenbaum, line 270: What's the plan for this section? It's a nice summary, but I see it as a comment in the checkpoints guide and as a section of this guide.

content/docs/user-guide/experiment-management/index.md
```diff
@@ -1,5 +1,18 @@
 # Experiment Management
 
+<!--
```
dberenbaum, line 3: What's the plan for this text?

content/docs/user-guide/experiment-management/managing-experiments.md
dberenbaum: Need to add to the index if we are keeping this section.

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
...
+`exp-bfe64` by default. It can be customized using the `--name` (`-n`) option:
+
+```dvc
+$ dvc exp run --name cnn-512 --set-param model.conv_units=512
+```
+
+When you create an experiment, DVC generates a Git-like SHA-1 hash from its
```
dberenbaum, line 17: Do people care if it's SHA-1? This seems like it's not worth the hassle to change if the naming method changes.

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
...
+contents. This is shown when you [queue experiments] with `--queue`:
```
dberenbaum, line 18: This is actually different from the hash generated after the experiments is committed, so I'm not sure it makes sense to mention it here.

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
...
+
+[queue experiments]:
+  /doc/user-guide/experiment-management/running-experiments#the-experiments-queue
+
+```dvc
+$ dvc exp run --queue -S model.conv_units=32
+Queued experiment '6518f17' for future execution.
+```
+
+After running queued experiment, DVC uses the regular name mentioned earlier.
```
dberenbaum, line 28: The experiment name should show in `dvc exp show` even for queued experiments now.

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
+# Managing Experiments
```
dberenbaum, line 1: I guess I'm the reason for this new section? I'm okay with leaving this in `Running experiments` depending on what others think. It does seem a bit disjointed as is. I think we need to at least link here from `Running experiments` so it's clear how to see and compare experiments.

dberenbaum (commented):
Thanks, @iesahin and @jorgeorpinel! It looks good, but I wasn't sure how deep to go. It still seems like a draft with the commented out sections and links to non-existent docs. What's the plan for this PR?

content/docs/user-guide/experiment-management/running-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: Sounds good thanks @pmrowla. Unrelated: should we consider renaming the feature in the future? maybe `exp waitlist` cc @dberenbaum 

Resolving here.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Hmm. It comes down to what we're trying to say with "run separately". I think it can be misunderstood as they always run (even if there's a matching run-cache entry) which is not correct it seems. That would only happen in some parallel runs right? I'll rephrase... ⏳ 

Idk if this is an implementation detail, the run-cache is a user-facing feature. We have it documented and mentioned in a few places (granted, not a lot but still).

jorgeorpinel: Updated in 7a8f155f.

jorgeorpinel: but I wonder whether we should just delete this whole note.

content/docs/user-guide/experiment-management/managing-experiments.md
jorgeorpinel: I'll make a separate PR for this ⏳

content/docs/user-guide/experiment-management/index.md
jorgeorpinel: @dberenbaum I extracted this and the non-existing page (Managing Exps) to https://github.com/iterative/dvc.org/pull/2751.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: @dberenbaum the guide is where we want to go into detail about all the possibilities and how every option works, including some implementation details if needed.

> the explanations here don't add much beyond what would be in the command reference

We plan to remove those details from the cmd ref (only leave short descriptions and link here) but it's not included in this PR (should follow immediately).

jorgeorpinel: That said I'm also not 100% satisfied with this guide but I want to merge it and create a new issue with a list of things to fix from it (e.g. regarding mixing in the pipeline concept) so let's keep this comment unresolved for now.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Exists: https://dvc.org/doc/start/data-pipelines

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Good Q, Idk haha @pmrowla ? 😬

jorgeorpinel: Nvmd @pmrowla sorry. I tested this and it does work @dberenbaum. Also in `stage add -p param2,param2,...`. Resolving

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Well since it's not a bug and that issue has p3 I doubt it'll be implemented soon so I think it's OK to leave this note for now.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: I agree the level of detail now is inconsistent but again the guide is where we can go this deep. Let's leave this one unresolved to mention in the improvement task later (about the level of detail) — may be addressed when we review details from the cmd ref to be moved in here.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Removed.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Good catch. I rephrased a little in 85a38c97.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Agree. I combined them in 0d5148de (also hid the trick since it can be dangerous).

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: No, good catch. Removed for now.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: This is basically pulled from https://dvc.org/doc/command-reference/exp/run#checkpoints. We need to determine what to leave on the cmd ref (after this PR). There's also a task about reviewing the whole checkpoints guide in https://github.com/iterative/dvc.org/issues/2548.

jorgeorpinel: p.s. please leave this comment unresolved.

content/docs/user-guide/experiment-management/managing-experiments.md
jorgeorpinel: Moved this to https://github.com/iterative/dvc.org/pull/2752#pullrequestreview-737794776.

content/docs/user-guide/experiment-management/managing-experiments.md
jorgeorpinel: Moved review

content/docs/user-guide/experiment-management/managing-experiments.md
jorgeorpinel: Moved

jorgeorpinel:
I'll merge this later but PTAL @dberenbaum. If I can address any other feedback today I will, otherwise I'll move any unresolved items to a follow-up ticket. Thanks!

content/docs/user-guide/experiment-management/running-experiments.md
dberenbaum: Hm, maybe it's my shell, but I get an error when doing `dvc exp run -S train.n_est=25,train.seed=1` on the example-get-started repo.

dberenbaum: Related: https://github.com/iterative/dvc/issues/5302

jorgeorpinel: This is what I did @dberenbaum 

image

I have DVC 2.6.4 on Ubuntu (WSL)

jorgeorpinel: Ah lol wait that didn't do what I wanted...

image

I guess it's not supported then... Will add a checkbox to the follow-up issue.

jorgeorpinel: Created quick PR instead since it's a bug: https://github.com/iterative/dvc.org/pull/2767

jorgeorpinel:
Promised follow-up: https://github.com/iterative/dvc.org/issues/2768 cc @iesahin

Issue #2641, opened 2021-07-16
restyled: looks like there is a discrepancy between prettier and restyled


shcheklein:
On this file content/docs/user-guide/contributing/blog.md if you try to update it a bit, restyled will complain, while prettier not.

jorgeorpinel:
> if you try to update it a bit, restyled will complain, while prettier not.

I tried in #2774 and everything I'm getting the opposite. Prettier complains (for bad changes) but Restyled always works... ⏳

jorgeorpinel:
Another thing I've noticed is that Restyled sometimes decides not to run e.g. for https://github.com/iterative/dvc.org/pull/2776 (only ran twice 6 days ago https://restyled.io/gh/iterative/repos/dvc.org/pulls/2776).

PR #2692, opened 2021-08-05
Added links to required software for Windows.

### LINKED ISSUE #2684: guide: Improve contributing instructions for Windows
In https://dvc.org/doc/user-guide/contributing/docs, we don't have any OS-specific instructions and for the last 3 hours I'm trying to install `yarn` on Windows by hunting information from SO. 

At least we need to mention the packages required needs Python, Visual Studio Build Tools and Windows SDK.

iesahin:
Fix #2684

### DIFF
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -70,6 +70,15 @@ project dependencies with Yarn:
 $ yarn
 ```
 
+> In Windows, you may need to install [Python], [Visual Studio Build Tools], and
+> [Windows SDK] beforehand for some packages.
+
+[windows sdk]:
+  https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/
+[visual studio build tools]:
+  https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019
+[python]: https://www.python.org/downloads/
+
 You may need to resolve dependencies at this point by running:
 
 ```dvc


iesahin:
ping @jorgeorpinel :)

content/docs/user-guide/contributing/docs.md
```diff
@@ -140,8 +149,8 @@ Some available variables:
 
 ## Doc style guidelines (JavaScript and Markdown)
 
-Some the rules below are be applied automatically by a pre-commit Git hook that
-is installed when `yarn` runs (see [dev env](#development-environment)).
+Following rules are applied automatically by a pre-commit Git hook that is
```
jorgeorpinel, line 152:
```suggestion
Some of the following rules are applied automatically by a pre-commit Git hook that is
```

content/docs/user-guide/contributing/docs.md
```diff
@@ -70,6 +70,15 @@ project dependencies with Yarn:
...
 ```
 
+> In Windows, you may need to install [Python], [Visual Studio Build Tools], and
+> [Windows SDK] beforehand for dependencies.
+
+[windows sdk]:
+  https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/
+[visual studio build tools]:
+  https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019
+[python]: https://www.python.org/downloads/
```
jorgeorpinel, line 80: Let's put this before all the code blocks? Otherwise it kind of interrupts the doc flow a bit.

jorgeorpinel: TBH maybe we should recommend a POSIX env instead and link to https://dvc.org/doc/user-guide/running-dvc-on-windows for details.

content/docs/user-guide/contributing/docs.md
```diff
@@ -140,8 +149,8 @@ Some available variables:
 
 ## Doc style guidelines (JavaScript and Markdown)
 
-Some the rules below are be applied automatically by a pre-commit Git hook that
-is installed when `yarn` runs (see [dev env](#development-environment)).
+Some of the following rules are applied automatically by a pre-commit Git hook that is
+installed when `yarn` runs (see [dev env](#development-environment)).
```
jorgeorpinel, line 153: Restyled didn't catch this for some reason... 😕 

```suggestion
Some of the following rules are applied automatically by a pre-commit Git hook
that is installed when `yarn` runs (see [dev env](#development-environment)).
```

content/docs/user-guide/contributing/docs.md
iesahin: I was able to run the node on Windows. This is not about running DVC itself on Windows, this is about running dvc.org clone on Windows. 

I agree that using WSL / POSIX is better. But in order to contribute to documents on Windows (without all this WSL stuff first), one should be able to run node.js first. Otherwise, for example, running `yarn format-all` doesn't work for contributors.

iesahin: I moved it before the node installation command.

content/docs/user-guide/contributing/docs.md
iesahin: Somehow my VSCode prettier and yarn's prettier behave differently sometimes, in minor ways. There is also some discrepancy between Git's handling of CRLF and Prettier's config. It should be OK in my latest commit.

content/docs/user-guide/contributing/docs.md
```diff
@@ -59,6 +59,15 @@ formatted and linted as well, which is also ensured by the full setup below.
 Make sure you have a recent LTS version of [Node.js](https://nodejs.org/en/)
 (`>=12.0.0`, `<=15.x`), and install [Yarn](https://yarnpkg.com/):
 
+> In Windows, you may need to install [Python], [Visual Studio Build Tools], and
+> [Windows SDK] additionally for dependencies.
```
jorgeorpinel, line 63:
```suggestion
> [Windows SDK] first.
```

jorgeorpinel: Why would the Windows SDK be needed though? Seems overkill, isn't it several GBs?

content/docs/user-guide/contributing/docs.md
```diff
@@ -59,6 +59,15 @@ formatted and linted as well, which is also ensured by the full setup below.
 Make sure you have a recent LTS version of [Node.js](https://nodejs.org/en/)
 (`>=12.0.0`, `<=15.x`), and install [Yarn](https://yarnpkg.com/):
 
+> In Windows, you may need to install [Python], [Visual Studio Build Tools], and
```
jorgeorpinel, line 62: Also why do you need Python to run `npm` and `yarn`?

content/docs/user-guide/contributing/docs.md
iesahin: I remember some package in the requirements asking for it. Some of these are compiled packages and probably need one or two functions from the SDK. Seems overkill, yes.

content/docs/user-guide/contributing/docs.md
iesahin: Not `npm` or `yarn` itself, but one of the (thousands of) packages require it. :)

content/docs/user-guide/contributing/docs.md
jorgeorpinel: Maybe related: https://github.com/iterative/dvc.org/issues/2641

content/docs/user-guide/contributing/docs.md
jorgeorpinel: To run the docs engine locally? I'd be surprised if that's the case. @rogermparent can you confirm? Thanks

And if it's true we should probably mention it in general, not just for Windows.

jorgeorpinel:
Strange link check failed: https://github.com/iterative/dvc.org/pull/2692/checks?check_run_id=3298902860 🤔

content/docs/user-guide/contributing/docs.md
iesahin: I didn't take the note of which package, and I don't want to test by removing everything but yes, Python was required to be installed by a package.

rogermparent: [`node-gyp`](https://github.com/nodejs/node-gyp), which is used to build native addons like `sharp` (our image processor) depends on Python. No way around it, though I've never run into the situation where I don't have Python (probably most likely on Windows I'd guess).

rogermparent:
> Strange link check

That _is_ bizarre, and I've seen it before, let's do a master merge and see if that fixes it, I have a hunch this is the case (and it needs to happen at some point anyway).

content/docs/user-guide/contributing/docs.md
jorgeorpinel: Interesting.

@iesahin mind mentioning Python in general then?

iesahin: I don't think there is a Linux or Mac installation without Python. Should we really mention it?

jorgeorpinel: Depends, what version of Python is needed? If it's 3.x then we should mention it as 2.x can still be the built-in one on MacOS and Linux I think.

jorgeorpinel (approved):
OK if Python 2 is required. If it's 3 then please mention it for all platforms as I think most distros still come with 2 (unless I'm outdated). See https://github.com/iterative/dvc.org/pull/2692#discussion_r690915004 above

rogermparent:
Seems `node-gyp` uses python 3.6+, according to [its README](https://github.com/nodejs/node-gyp)

content/docs/user-guide/contributing/docs.md
jorgeorpinel: Seems like we should mention for all platforms after all @iesahin ? Per https://github.com/iterative/dvc.org/pull/2692#issuecomment-929581152

jorgeorpinel (changes requested):

jorgeorpinel:
Ping @iesahin can we finalize or close this please? Thanks

content/docs/user-guide/contributing/docs.md
iesahin: Added Python 3.6+ as a general requirement like Yarn.

content/docs/user-guide/contributing/docs.md
```diff
@@ -140,8 +150,8 @@ Some available variables:
 
 ## Doc style guidelines (JavaScript and Markdown)
 
-Some the rules below are be applied automatically by a pre-commit Git hook that
-is installed when `yarn` runs (see [dev env](#development-environment)).
+Some of the following rules are applied automatically by a pre-commit Git hook
```
jorgeorpinel, line 153: Thanks for the typo fix.

content/docs/user-guide/contributing/docs.md
```diff
@@ -193,9 +203,9 @@ We try to use a casual and fun tone in our docs. We also avoid authoritative
 language such as "As you can see, clearly this is what happened, of course" etc.
 which while good-intentioned, may scare readers off.
 
-We prefer human-friendly language than exact jargon, as long as it's correct,
-even if using general terminology. Example: avoid Git jargon such as _revision_
-or _reference_, preferring the more basic concept _commit_.
+We prefer human-friendly language rather than exact jargon as long as it's
+correct. Example: avoid Git jargon such as _revision_ or _reference_, preferring
+the more basic concept _commit_.
```
jorgeorpinel, line 208: Out of scope to review the language in this guide but OK...
```suggestion
We prefer general, human-friendly language rather than exact jargon as long as it's
correct. Example: avoid Git jargon such as _revision_ or _reference_, preferring
the more basic terms _commit_ or _version_.
```

content/docs/user-guide/contributing/docs.md
```diff
@@ -204,7 +214,7 @@ we use expandable sections for complex implementation details.
 Start by writing the essence in simple terms, and complete it with
 clarifications, edge cases, or other precisions in a separate iteration.
 
-We use **bold** text for emphasis, and _italics_ for special terms.
+We use **bold** text for emphasis and _italics_ for special terms.
```
jorgeorpinel, line 217: This was OK.

PR #2694, opened 2021-08-05
Remove `gatsby-plugin-catch-links`

### LINKED ISSUE #2659: redirects show a 404 page

julieg18:
* removes `gatsby-plugin-catch-links` as it breaks our redirect links(see #2659) and makes little difference to our page links
* alternatively we can keep plugin if desired and add a `excludePattern` for redirect links

Fixes #2659

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -123,7 +123,6 @@ const plugins = [
   },
   'gatsby-transformer-sharp',
   'gatsby-plugin-sharp',
-  'gatsby-plugin-catch-links',
   {
     resolve: 'gatsby-plugin-manifest',
     options: {
--- a/package.json
+++ b/package.json
@@ -114,7 +114,6 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "^7.19.0",
-    "gatsby-plugin-catch-links": "^2.2.1",
     "gatsby-plugin-feed": "^2.4.1",
     "gatsby-plugin-google-analytics": "^2.2.2",
     "gatsby-plugin-manifest": "2.2.23",
--- a/yarn.lock
+++ b/yarn.lock
@@ -7122,14 +7122,6 @@ gatsby-page-utils@^0.2.7:
     lodash "^4.17.15"
     micromatch "^3.1.10"
 
-gatsby-plugin-catch-links@^2.2.1:
-  version "2.3.2"
-  resolved "https://registry.yarnpkg.com/gatsby-plugin-catch-links/-/gatsby-plugin-catch-links-2.3.2.tgz#c6a24b7c1ea067bc48684db01448d0c44de658c5"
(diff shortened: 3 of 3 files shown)


rogermparent:
I noticed that this affects internal blog links (like links to the docs), breaking CSR (though the links still work as normal links, of course)

screenshot of link not having an event

While this isn't a dealbreaker, it's something we should look into fixing if we go this route. My personal ideal is that we stop using two different markdown rendering engines and use the mdast approach on the blog, but if that's not an option then we can find another way around with something like an `onClick` event that does what `gatsby-plugin-catch-links` does but only applied on blog MD content.

That said, the issue of "internal blog content links don't have CSR" is so small that I think we can let this in to fix the more visible redirect issue and then take care of it later.

julieg18:
> That said, the issue of "internal blog content links don't have CSR" is so small that I think we can let this in to fix the more visible redirect issue and then take care of it later.

We could also update the plugin with `excludePattern`. Looking at the redirect function we have in `utils` it should be simple enough.

rogermparent:
After another look, it also seems like we depend on this plugin for internal links within `basic-concepts` tooltips, and given there may be other hidden things that it covers, it may be better to just add the links we need excluded to the `gatsby-plugin-catch-links` exclusion regex for now and in the future work on making the engine sane enough that we can remove it later.

shcheklein:
@rogermparent 

it's all pretty complicated and sad :)

> the future work on making the engine sane enough that we can remove it later

Roger, any ideas for the future how it all can be improved? is there a way to keep pages prefetch and that nice stuff of Gatsby + make redirects work (external and internal)?

rogermparent:
> Roger, any ideas for the future how it all can be improved? is there a way to keep pages prefetch and that nice stuff of Gatsby + make redirects work (external and internal)?

I think using the `htmlAst`-type rendering that we do in docs everywhere else MD is used (which mostly means blog and tooltips) would go a long way in letting us fix this issue and give granular control over the components rendered in content, since it allows us to swap out normal MD elements for true React components when necessary, like `a` => `Link` in a way that the raw remark+html way we currently do blog elements can't handle.

Aside from that, a simpler solution would be to create something like `gatsby-plugin-catch-links` that can be applied onto certain classes instead of the whole body, as there's distinct areas of the site that use raw HTML where we can modify the wrapper but not the contents. I suppose we could also try to take `gatsby-plugin-catch-links` itself and add this functionality- I took a look into it earlier and it doesn't seem to have it easily out of the box but the meat of the plugin in the `catch-links` module seems importable in such a way that we can override the behavior for our needs by making it redirect-aware.

julieg18:
> it may be better to just add the links we need excluded to the `gatsby-plugin-catch-links` exclusion regex for now...

Agreed! I suppose I can go ahead and close this pull request and open one for adding `/doc/cml` type redirects to `gatsby-plugin-catch-links`, fixing the main bug. We could discuss what to do next either in #2659 or in a new issue if needed😃

PR #2701, opened 2021-08-09
updating events

jendefig:
Updating events on community.json


content/community.json
```diff
@@ -71,60 +71,44 @@
   ],
   "events": [
     {
-      "url": "https://hasgeek.com/fifthelephant/mlops-conference/schedule",
-      "title": "HasGeek",
-      "description": "Milecia McGregor presents at HasGeek Conference from Fifth Elephant on Tuning Hyperparameters with DVC Experiments",
-      "city": "Virtual, Miami, FL",
-      "date": "2021-07-23",
-      "pictureUrl": "/img/community/events/hasgeek.png"
```
shcheklein, line 79: if we remove these events, let's also remove images that belong to them (but I think there is not need to remove events, as far as I remember).

shcheklein (approved):

content/community.json
jendefig: Ok.  But there's a high likelihood that they will repeat in the future, but I can keep them all offline, so it's not bloated.

Issue #2610, opened 2020-02-11
import: allow "chaining" imports somehow (except circular?)


jorgeorpinel:
> More context around https://discordapp.com/channels/485586884165107732/485596304961962003/676405501940072453

As of now, imported data is not cached by default, so you won't be able to import any imported data:

repo1 -> ✔️ dvc import data -> repo2 🙂 -> ✖️ dvc import data -> repo3 🙁 

Somehow allowing this could be useful for the case where you're building a data registry based on other previous smaller DVC repos, for example. Right now you have to `dvc get` and then `dvc add` those artifacts from scratch in the data registry (so they can be imported in further DVC repos).

Ruslan mentioned something about using "links" to implement this (on [Discord](https://discordapp.com/channels/485586884165107732/485596304961962003/676466989035945984)).

UPDATE: Go to https://github.com/iterative/dvc/issues/3305#issuecomment-836503176

dberenbaum:
Moving to dvc.org since it seems like the remaining work is documentation. 

@jorgeorpinel We might be able to better support https://github.com/iterative/dvc/issues/4527 in the future, but until then it IMO it would be useful to be more clear that imports rely on the original source remote and users need access to that, including for chained imports. Thoughts?

@pmrowla Can you take this one if we need to document it? Is there someone else who should be assigned?

pmrowla:
@dberenbaum I'm guessing it will have to be me since I did the current implementation, you can just assign it to me if/when needed

jorgeorpinel:
Thanks guys, we'll take it over as soon as we have all the basic info.

> running dvc update dir from repo D will only check if dir has changed in repo C

This part may deserve it's own issue in iterative/dvc though?

dberenbaum:
My thoughts: The current behavior could be expected or unexpected depending on the circumstances. There's not a lot of user feedback outside of the one user in https://github.com/iterative/dvc/issues/4527, so I'm not sure it's worth keeping the discussion going. Can we document what we have and revisit if users complain or are confused?

pmrowla:
> > running dvc update dir from repo D will only check if dir has changed in repo C
> 
> This part may deserve it's own issue in iterative/dvc though?

[...]

PR #2705, opened 2021-08-10
docs: document chained `dvc import` behavior

pmrowla:
Closes https://github.com/iterative/dvc.org/issues/2610

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -73,8 +73,10 @@ Note that `dvc repro` doesn't check or update import `.dvc` files (see
 `dvc freeze`), use `dvc update` to bring the import up to date from the data
 source.
 
-Also note that chained imports (importing data that was imported into the source
-repo at `url`) are not supported.
+DVC includes limited support for chained imports (importing data that was
+imported into the source repo at `url`).
+
+> See [Chained imports](#example-chained-imports) below for more details.
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


jorgeorpinel:
Thanks @pmrowla !

@iesahin mind taking a look? I'll take over the review at some point since I opened the related issue but it'd be great to get your fresh perspective on the feature and explanation. Thanks

iesahin:
Taking over for copy edits. Thank you Peter.

jorgeorpinel:
Thanks @iesahin I can take it over actually 🙂 but can you leave a review please? I'll address it.

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
 
 The `url` and `rev_lock` subfields under `repo` are used to save the origin and
 [version](https://git-scm.com/docs/revisions) of the dependency, respectively.
+
+## Example: Chained imports
+
+DVC supports importing data importing data that was imported into the source
```
iesahin, line 284: "importing data importing data that was imported" looks fun 😄

iesahin (changes requested):

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+    nfiles: 4
+    path: dir
+```
+
+Each time that we run `dvc import dir` (from C), `dvc update dir` or
+`dvc pull dir`, DVC will first look up the contents of `dir` in repo C. At that
+point in time, DVC will see that `foo` and `bar` are chained imports from repos
+A and B, and then continue resolving the chain as needed (by looking up `foo`
+from A, and `bar` from B).
+
```
iesahin, line 395: I think the example can be simplified by having `A/foo.dvc` -> `B/foo.dvc` -> `C/foo.dvc`. No need for more than one file to show the chained import.

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+This means that all three repos A, B and C must be reachable when DVC commands
+are run from repo D, otherwise the import chain resolution would fail, and the
+`dvc import` command will error out.
+
+⚠️ When fetching files imported from the chain, DVC will fetch them from their
+original source locations. This means that the default DVC remotes for all repos
+in the import chain must be accessible from the final destination repo. So in
```
iesahin, line 402: This is repeating above. Requiring to be "accessible" is standard, I don't think this emphasis is needed.

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+this example, a user running `dvc pull` from repo D must have the appropriate
+credentials to access the default remotes for all three source repos (A, B and
+C).
+
+> Note that when running `dvc update dir` from repo D, DVC will only check
```
iesahin, line 407: I think this needs more attention like ⚠️ .

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+
+## Example: Chained imports
+
+DVC supports importing data importing data that was imported into the source
+repo, as long as all repos in the import chain are accessible from the final
+destination repo. Each source repo's default DVC remote must also be accessible
+from the final destination repo.
+
+Consider an example where we start with two DVC repos, A and B, with the
+following contents:
```
iesahin, line 290: Instead of starting with two repos, it may be better to start with one.

iesahin (dismissed):

content/docs/command-reference/import.md
jorgeorpinel: Probably just a typo. Fixing...

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+    rev_lock: 32ab3ddc8a0b5cbf7ed8cb252f93915a34b130eb
+outs:
+- md5: acbd18db4cc2f85cedef654fccc4a4d8
+  size: 3
+  path: foo
+outs:
+- md5: 630bd47b538d2a513c7d267d07e0bc44.dir
+  size: 3
+  nfiles: 1
+  path: subdir
```
jorgeorpinel, line 345: This doesn't belong in `dir/foo.dvc` of your example I think right @pmrowla ? I'll remove it for now.

content/docs/command-reference/import.md
jorgeorpinel: Simplifying...

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,133 @@ outs:
...
+```
+
+Now, lets say that we want to create a fourth DVC repo, D, and run the following
+command:
+
+```dvc
+# In repo D
+$ dvc import /repo/c dir
+```
```
jorgeorpinel, line 354: How can you import `dir` if there's no `dir.dvc`? That actually works @pmrowla ? That's cool

jorgeorpinel: p.s. I'm renaming all the files and dirs to make it look a bit more realistic.

content/docs/command-reference/import.md
pmrowla:
```
outs:
- md5: acbd18db4cc2f85cedef654fccc4a4d8
  size: 3
  path: foo
```
this part was part of a bad copy/paste and should be removed. The `.dir` hash is supposed to be there.

`dir/` contains `foo` which is imported, and a `subdir/` which is a new directory for this repo, tracked by `dvc add dir/subdir`

The point is to show that later in the chain, we can import `dir/` which contains both dvc tracked data specific to this repo, as well as imports from other DVC repos

content/docs/command-reference/import.md
pmrowla: It works the same way as importing git files & dirs. `dvc import dir/` just imports everything inside `dir/` whether it is git-tracked or dvc-tracked. You could already do that before the import chaining changes were made.

content/docs/command-reference/import.md
jorgeorpinel: This is emphasizing that the remote storage is also used. I'll shorten and merge this with the previous paragraph.

content/docs/command-reference/import.md
jorgeorpinel: I don't think it's such a critical note. It's not even about `import`.

content/docs/command-reference/import.md
jorgeorpinel: I updated the whole example, PTAL @iesahin

content/docs/command-reference/import.md
jorgeorpinel: A bit confused. TBH I didn't have a chance to actually test all this. Mind taking a look at my rewrite and confirm it's correct? Thanks

content/docs/command-reference/import.md
```diff
@@ -276,3 +278,108 @@ outs:
...
+└── training
+    ├── data.csv
+    ├── data.csv.dvc
+    ├── labels
+    │   ├── test.txt
+    │   └── truth.txt
+    └── labels.dvc
+```
+
+> Notice that the `training/labels` directory is also tracked by DVC separately.
```
pmrowla, line 316: @jorgeorpinel This is where the removed `outs` section came from (it's the contents of `labels.dvc`). It may be better to move this below the `training/data.csv.dvc` block so it says something like:

If we also examine `training/labels.dvc` we can see that it is a directory tracked by DVC and not an import. Meaning that `training/` contains a both data unique to repo B and imports from repo A.
```
outs:
- md5: 630bd47b538d2a513c7d267d07e0bc44.dir
  size: 3
  nfiles: 1
  path: labels
```

jorgeorpinel: Got it. Updated to use some of your language in 4bf55012. Thanks!

shcheklein (approved):

jorgeorpinel:
The failed check has nothing to do with this PR (strange). I'll fix it separately.

Thanks again @pmrowla 🎉

PR #2706, opened 2021-08-10
Fix syntax error on `dvc push`

lucabonifacio:
Some examples use the command `dvc remote push` which, if executed, results into an error.

Most probably, the correct command should be: `dvc push`.


shcheklein (approved):

shcheklein:
thanks @lucabonifacio ! 🙏

PR #2721, opened 2021-08-16
Update pipelines-files.md

skshetry:
Introduced in 2.0. Related: https://github.com/iterative/dvc/pull/5194

There was a docs patch for `repro` (https://github.com/iterative/dvc.org/pull/2071) which is missing (most likely got lost during merge in v1 -> v2 docs transition). I'll try adding that as well.


content/docs/command-reference/repro.md
```diff
@@ -25,9 +25,10 @@ graph (a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)) implicitly
 defined by the stages listed in `dvc.yaml`. The commands defined in these stages
 are then executed in the correct order.
 
-For stages with multiple commands (having a list in the `cmd` field), commands
-are run one after the other in the order they are defined. The failure of any
-command will halt the remaining stage execution, and raises an error.
+For stages with multiple commands (having a list or a multiline string in the
```
shcheklein, line 28: is there a standard for "multiline strings" in YAML (starts with | or > or any string with newlines)?

skshetry: It's the one with `|`. The `>` is quite opposite, changes a newline into a space, i.e. multiline to a single line string.

```yaml
stages:
  stage1:
    cmd: |
       wget https://code.dvc.org/get-started/code.zip
       unzip code.zip
       rm -f code.zip 
```

shcheklein: Thanks @skshetry ! That's exactly what I was looking for - that link. We can make the `multiline string` a link? (and probably the list also). But even in that case I would include explicitly that we mean `|` style blocks, right here in the text.

iesahin: I think we can also document this in https://dvc.org/doc/user-guide/project-structure/pipelines-files

jorgeorpinel: > haven't found a good place to accomodate this as an example, we don't have examples on dvc.yaml guide

There are many snippet examples all over. https://dvc.org/doc/user-guide/project-structure/pipelines-files#templating has an anti-multiline one for example. We should change that one if we recommend `|` better:

image

jorgeorpinel: p.s. Agree on linking and adding the info to the pipelines file guide.

skshetry: We do have examples, but not for the possibilities of every keyword, just for the high-level features.

skshetry:
We still don't support multiline string in `dvc.lock` properly. Need to see what we can do, without causing compatibility issues.
Keeping it in draft till then.

content/docs/command-reference/repro.md
jorgeorpinel: > Would have been great if it were collapsible

jorgeorpinel: > Would have been great if it were collapsible

Good idea! It can still be great, feel free to contribute

PR #2731, opened 2021-08-18
Added `Invalid authentication for Git Remote` to troubleshooting

daavoo:
Core P.R: https://github.com/iterative/dvc/pull/6493


content/docs/command-reference/exp/pull.md
```diff
@@ -27,6 +27,11 @@ experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
 A working `git_remote` name (e.g. `origin`) or valid Git repo's URL is required,
 as well as an `experiment` name or hash (see `dvc exp run`) to pull.
 
+> ⚠️ Only [SSH Git URLs] are valid for `git_remote`.
```
jorgeorpinel, line 30: We can include in the paragraph above like this:

```suggestion
A working `git_remote` name (e.g. `origin`) or valid [SSH Git URL] is required,
as well as an `experiment` name or hash (see `dvc exp run`) to pull.
```

jorgeorpinel (changes requested):
One thing to apply in the 3 docs ☝️ 

Also probably we should update the short description of `git_remote` under **Synopsis** (and then we can make an issue in the core repo for the corresponding update).

Thanks @daavoo

jorgeorpinel:
> It doesn't have to be an SSH URL in every case though.

Sounds like this needs another rewrite then? To clarify

daavoo:
> > It doesn't have to be an SSH URL in every case though.
> 
> Sounds like this needs another rewrite then? To clarify

Yep. Also I think that some catch and show an error message pointing to docs might be added in dvc core.

content/docs/command-reference/exp/pull.md
```diff
@@ -24,9 +24,12 @@ experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
 > these are saved under custom Git references. See **How does DVC track
 > experiments?** in `dvc exp run` to learn more about DVC experiment storage.
 
-A working `git_remote` name (e.g. `origin`) or valid Git repo's URL is required,
+A working `git_remote` name (e.g. `origin`) or valid [SSH Git URL] is required,
 as well as an `experiment` name or hash (see `dvc exp run`) to pull.
 
+[ssh git url]:
+  https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_protocols
```
karajan1001, line 31: Isn't an SSH Git URL in the form 'git@github.com:iterative/dvc.git'?
and `https://github.com/iterative/dvc.git` is an HTTPS Git URL

pmrowla: This is a link to the git documentation regarding supported protocols and URLs (it's the markdown syntax for reference links, so the `[SSH Git URL]` in the previous paragraph will be a link w/this URL)

daavoo:
> One important note here: [iterative/dvc#6493 (comment)](https://github.com/iterative/dvc/pull/6493#issuecomment-906269837)
> 
> > It doesn't have to be an SSH URL in every case though.
> > The SSH requirement is specifically for pushing (writing) to a remote git server that requires authentication (meaning pushing to github/gitlab). exp pull and exp list will work with any git URL (including HTTP) as long as the git server is public. If you are hosting a personal git server somewhere that was configured to allow pushing over unauthenticated HTTP, you could use the HTTP git URL format (and would not be required to use SSH).
> > And yes, read operations that require authentication (pull/list for private github/gitlab repos) would require using SSH.

Updated the warning to better clarify that

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,18 @@ using:
...
 ```
+
+## Invalid authentication for Git Remote {#git-auth}
+
+Some DVC commands accept `git_remote` as argument. In order to run those
+commands, you must be properly authenticated having _read_ (`dvc exp list`,
+`dvc exp pull`) or _write_ (`dvc exp push`) permissions on the provided Git
+Remote name or Git URL.
+
+This means that, unless you are hosting a private git server that allows
```
iesahin, line 102: ~~This means that~~

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,18 @@ using:
 ```dvc
 $ dvc checkout --relink
 ```
+
+## Invalid authentication for Git Remote {#git-auth}
+
+Some DVC commands accept `git_remote` as argument. In order to run those
+commands, you must be properly authenticated having _read_ (`dvc exp list`,
```
iesahin, line 98: having -> to have

content/docs/command-reference/exp/push.md
```diff
@@ -24,8 +24,8 @@ experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
 > these are saved under custom Git references. See **How does DVC track
 > experiments?** in `dvc exp run` to learn more about DVC experiment storage.
 
-A working `git_remote` name (e.g. `origin`) or valid Git repo's URL is required,
-as well as an `experiment` name or hash (see `dvc exp run`) to push.
+A working `git_remote` name (e.g. `origin`) or valid Git URL is required, as
+well as an `experiment` name or hash (see `dvc exp run`) to pull.
```
iesahin, line 28: why `pull`? This is ref for `dvc exp push`.

iesahin (commented):
Some minor comments. Thank you @daavoo

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,18 @@ using:
 ```dvc
 $ dvc checkout --relink
 ```
+
+## Invalid authentication for Git Remote {#git-auth}
```
jorgeorpinel, line 95: Is this an actual link output from `dvc` CLI?

jorgeorpinel: Ah yes I see it's in https://github.com/iterative/dvc/pull/6493/files#diff-bf8bef3f4902b5642f7280464c2f9c52574fb2f20b37c0c8fce28075c09e2b24.

content/docs/command-reference/exp/list.md
```diff
@@ -10,7 +10,7 @@ usage: dvc exp list [-h] [-q | -v] [--rev <rev>]
                     [git_remote]
 
 positional arguments:
-  git_remote    Optional Git remote name or repo URL
+  git_remote    Optional Git remote name or valid Git URL
```
jorgeorpinel, line 13:
```suggestion
  git_remote    Optional Git remote name or Git URL (SSH auth only)
```

content/docs/command-reference/exp/list.md
```diff
@@ -20,12 +20,12 @@ or commit they're based on. This is similar to `dvc exp show --no-pager`, but
 limited to experiment names and with very simple formatting. See also
 `dvc exp run`.
 
-If a working `git_remote` name (e.g. `origin`) or valid Git repo's URL is
-provided, lists experiments in that <abbr>repository</abbr> instead (if any,
-based on the `dvc remote default`).
+If a working `git_remote` name (e.g. `origin`) or valid Git URL is provided,
+lists experiments in that <abbr>repository</abbr> instead (if any, based on the
```
jorgeorpinel, line 24:
```suggestion
If a working `git_remote` name (e.g. `origin`) or valid Git URL is provided (only SSH authentication is supported),
this command lists the experiments in that <abbr>repository</abbr> instead (if any, based on the
```

content/docs/command-reference/exp/list.md
```diff
@@ -20,12 +20,12 @@ or commit they're based on. This is similar to `dvc exp show --no-pager`, but
...
+`dvc remote default`).
```
jorgeorpinel, line 25: > if any, based on `dvc remote default`

Unrelated but is that bit correct? I think the experiments may be listed whether or not they've been pushed to the default remote or any DVC remote for that matter. Not sure

content/docs/command-reference/exp/pull.md
```diff
@@ -11,7 +11,7 @@ usage: dvc exp pull [-h] [-q | -v] [-f] [--no-cache]
                     git_remote experiment
 
 positional arguments:
-  git_remote          Git remote name or repo URL
+  git_remote          Git remote name or valid Git URL
```
jorgeorpinel, line 14:
```suggestion
  git_remote    Optional Git remote name or Git URL (SSH auth only)
```

content/docs/command-reference/exp/pull.md
```diff
@@ -24,8 +24,8 @@ experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
 > these are saved under custom Git references. See **How does DVC track
 > experiments?** in `dvc exp run` to learn more about DVC experiment storage.
 
-A working `git_remote` name (e.g. `origin`) or valid Git repo's URL is required,
-as well as an `experiment` name or hash (see `dvc exp run`) to pull.
+A working `git_remote` name (e.g. `origin`) or valid Git URL is required, as
```
jorgeorpinel, line 27:
```suggestion
A working `git_remote` name (e.g. `origin`) or valid Git URL is required (only SSH authentication is supported), as
```

content/docs/command-reference/exp/push.md
```diff
@@ -11,7 +11,7 @@ usage: dvc exp push [-h] [-q | -v] [-f] [--no-cache]
                     git_remote experiment
 
 positional arguments:
-  git_remote          Git remote name or repo URL
+  git_remote          Git remote name or valid Git URL
```
jorgeorpinel, line 14:
```suggestion
  git_remote          Git remote name or Git URL (SSH auth only)
```

content/docs/command-reference/exp/push.md
```diff
@@ -24,8 +24,8 @@ experiments across <abbr>repository</abbr> copies via Git (and DVC) remotes.
 > these are saved under custom Git references. See **How does DVC track
 > experiments?** in `dvc exp run` to learn more about DVC experiment storage.
 
-A working `git_remote` name (e.g. `origin`) or valid Git repo's URL is required,
-as well as an `experiment` name or hash (see `dvc exp run`) to push.
+A working `git_remote` name (e.g. `origin`) or valid Git URL is required, as
```
jorgeorpinel, line 27:
```suggestion
A working `git_remote` name (e.g. `origin`) or valid Git URL is required (only SSH authentication is supported), as
```

jorgeorpinel (commented):
Per https://github.com/iterative/dvc/pull/6493#issuecomment-914431340

content/docs/command-reference/exp/list.md
```diff
@@ -20,12 +20,12 @@ or commit they're based on. This is similar to `dvc exp show --no-pager`, but
...
 
-> Note that this utility doesn't require an existing <abbr>DVC project</abbr> to
-> run from when a `git_remote` URL is given.
+> This utility doesn't require an existing <abbr>DVC project</abbr> to run from
+> when a `git_remote` URL is given.
```
jorgeorpinel, line 28: Seems unrelated. Why change this?

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,18 @@ using:
...
+
+Some DVC commands accept `git_remote` as argument. In order to run those
+commands, you must be properly authenticated to have _read_ (`dvc exp list`,
+`dvc exp pull`) or _write_ (`dvc exp push`) permissions on the provided Git
+Remote name or Git URL.
+
+Unless you are hosting a private git server that allows unauthenticated HTTP
+read/write, you would need to use an [SSH Git URL] for using `dvc exp list` /
+`dvc exp pull` **in private repositories** and for using `dvc exp push` **in any
+case**.
```
jorgeorpinel, line 105: Gonna commit this one:

```suggestion
## HTTP Git authentication is not supported {#git-auth}

[Experiment sharing](/doc/user-guide/experiment-management/sharing-experiments)
commands accept a `git_remote` argument. In order to access the Git remote,
you may need to authenticate for _reading_ (`dvc exp list`,
`dvc exp pull`) or _writing_ (`dvc exp push`) permissions.

This means that unless the Git server allows unauthenticated HTTP
read/write, you should use an [SSH Git URL] when pushing experiments,
and when listing or pulling experiments from private repositories.
```

jorgeorpinel: But I think it's not 100% correct still... ⏳

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,17 @@ using:
 ```dvc
 $ dvc checkout --relink
 ```
+
+## HTTP Git authentication is not supported {#git-auth}
+
+[Experiment sharing](/doc/user-guide/experiment-management/sharing-experiments)
+commands accept a `git_remote` argument. In order to access the Git remote,
```
jorgeorpinel, line 98: This is not just about `git_remote` right? Even if that's not provided, the default remote should also use SSH auth (if auth is needed).

content/docs/user-guide/troubleshooting.md
```diff
@@ -91,3 +91,17 @@ using:
...
+you may need to authenticate for _reading_ (`dvc exp list`,
+`dvc exp pull`) or _writing_ (`dvc exp push`) permissions.
+
+This means that unless the Git server allows unauthenticated HTTP
+read/write, you should use an [SSH Git URL] when pushing experiments,
+and when listing or pulling experiments from private repositories.
```
jorgeorpinel, line 104: I think it should be simpler like this? The "unless" part already implies the cases where HTTP is OK.
```suggestion
This means that unless the Git server allows unauthenticated HTTP
write/read, you should use an [SSH Git URL] when pushing, listing,
or pushing experiments.
```

content/docs/user-guide/troubleshooting.md
daavoo: This error will only happen in commands using `git_remote`. Other errors regarding DVC remote auth would depend on the remote type used and will be handled with a different exception and error message like [`GDriveAuthError`](https://github.com/iterative/dvc/blob/master/dvc/fs/gdrive.py)

content/docs/command-reference/exp/list.md
daavoo: But what about: iterative/dvc#6493 (comment) . This change would only really apply to push, right?

content/docs/command-reference/exp/list.md
dberenbaum: Yes, whether the data was pushed to the dvc remote is irrelevant to `dvc exp list`. Cc @pmrowla for confirmation.

content/docs/command-reference/exp/pull.md
dberenbaum: The Git remote is not optional for `dvc exp pull`.

content/docs/command-reference/exp/list.md
dberenbaum: @daavoo As your user guide comment explains, this applies to reads also (for example, private repos), although authentication may not be necessary for reads as often as for writes. Do you think the suggested change is unclear?

iesahin: I think we can say "authenticated Git URL" without mentioning SSH. The requirement is _authentication_, not a particular protocol. It may be a `file://...` URL on NFS and it still works if you're authenticated, and doesn't if you're not. The reason we can't use HTTP is that we don't get user credentials for authentication and HTTP doesn't allow unauthenticated writes. I believe it's better to mention "authentication"  rather than SSH here.

content/docs/command-reference/exp/list.md
iesahin: Here, also. We can use cloned bare repositories on local disk to `dvc exp push`. It's not about particular connection protocol.

content/docs/command-reference/exp/list.md
daavoo: @dberenbaum  "SSH Git URL" was my first proposed change and then @pmrowla clarified that, for reading (`list`, `pull`), "SSH" is not a requirement in some cases. So I'm confused about getting back to that terminology. 

@iesahin I like "authenticated" very much 👍

content/docs/command-reference/exp/list.md
daavoo: It was not intentional. Removing that

content/docs/command-reference/exp/list.md
dberenbaum: > The requirement is _authentication_, not a particular protocol.

👍  I'm not sure "authenticated" sounds right. I would interpret that as meaning that DVC won't work with URLs where authentication is not required, like reading from a public Github url, or writing to an unprotected NFS URL.

What needs to be clear is that DVC will not work if credentials need to be provided to authenticate (see https://git-scm.com/docs/gitcredentials). Unfortunately, even this may be confusing because we aren't currently using Git's credentials helpers, so even if someone has their credentials stored in Git and doesn't need to enter credentials when doing `git push`, `dvc exp push` may still fail. I opened https://github.com/iterative/dvc/issues/6586 to possibly implement support for credentials, which would negate this whole PR and discussion 😄 .

For now, I think it's fine to do any of:

1. Close this PR altogether since https://github.com/iterative/dvc/pull/6493 should make the problem more self-explanatory.
2. Add a note that the command will fail if credentials need to be provided to authenticate.
3. Merge as is if I'm the only one who finds "authenticated" confusing.

jorgeorpinel: > The requirement is authentication

Authentication is NOT always a requirement from what I understand? So "authenticated Git URL" (current text) seems equally misleading... I agree with the statement in https://github.com/iterative/dvc/pull/6493#issuecomment-914431340:

> we could probably summarize behavior as not supporting HTTP authentication

So maybe for now we can go with `Git URL (HTTP auth is not supported)`

jorgeorpinel: If that's not convincing let's just go with something simple like `Git URL` and let the error happen in the edge cases, and explain the details in the docs. Hard to be super correct in a short description.

content/docs/user-guide/troubleshooting.md
jorgeorpinel: Hmm that sounds a bit strange but OK. In any case the problem itself is broader right? It also applies to configured remotes so why not explain it in it's broader context here? Even if this particular troubleshooting section is only reached (now) via `exp push/list/push <git_remote>`

No strong opinion but I think it would be make sense to make it broader.

content/docs/command-reference/exp/list.md
```diff
@@ -10,7 +10,7 @@ usage: dvc exp list [-h] [-q | -v] [--rev <rev>]
                     [git_remote]
 
 positional arguments:
-  git_remote    Optional Git remote name or repo URL
+  git_remote    Optional Git remote name or authenticated Git URL
```
pmrowla, line 13: Why not just `Git URL`? Using `authenticated` here (and in the other places this change was added) doesn't make sense to me. This also makes it seem confusing since we don't mention authentication at all for `Git remote name`, but we do for `Git URL` (even though the potential auth requirements apply in both situations).

In the docs for `remote add`/`remote modify` we document URLs as

> `url` - remote location

We don't say

> `url` - authenticated remote location

content/docs/command-reference/exp/list.md
daavoo: Changed the following:

- Just using `Git URL` in command reference.
- Added note about git credentials in the troubleshooting section.

And let users reach the troubleshooting page when the exception is raised

content/docs/user-guide/troubleshooting.md
daavoo: I think that the thing is that, for example, the DVC HTTP remote does ask for users/password (hence my comment that errors for each remote should be handled with separated exceptions). The `git_remote` parts of `dvc exp` do not share code with `dvc remote`.

dberenbaum (approved):

content/docs/user-guide/troubleshooting.md
jorgeorpinel: Hmmm. OK it sounds like this is closely related with remote config per https://github.com/iterative/dvc.org/pull/2731#pullrequestreview-751064480 so I still think we could explain the broader context but maybe I'm not getting why that's not a good option. But OK

content/docs/command-reference/exp/list.md
```diff
@@ -20,9 +20,8 @@ or commit they're based on. This is similar to `dvc exp show --no-pager`, but
 limited to experiment names and with very simple formatting. See also
 `dvc exp run`.
 
-If a working `git_remote` name (e.g. `origin`) or valid Git repo's URL is
-provided, lists experiments in that <abbr>repository</abbr> instead (if any,
-based on the `dvc remote default`).
+If a working `git_remote` name (e.g. `origin`) or Git URL is provided, lists
+experiments in that <abbr>repository</abbr> instead (if any).
```
jorgeorpinel, line 24: Git URL or file system path to the repo right? Should probably include all the options down here in the description similar to how the `url` arg is explained in https://dvc.org/doc/command-reference/import.

jorgeorpinel: But I guess that's out of scope for the PR...

jorgeorpinel (approved):

jorgeorpinel:
Please merge along with https://github.com/iterative/dvc/pull/6493.

PR #2740, opened 2021-08-23
Add --show-csv flag to dvc exp show

karajan1001:
per iterative/dvc#6468

### DIFF
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -14,7 +14,7 @@ usage: dvc exp show [-h] [-q | -v] [-a] [-T] [-A] [-n <num>]
                     [--exclude-params <params_list>] [--param-deps]
                     [--sort-by <metric/param>]
                     [--sort-order {asc,desc}] [--no-timestamp] [--sha]
-                    [--show-json] [--precision <n>]
+                    [--show-json] [--show-csv] [--precision <n>]
 ```
 
 ## Description
@@ -113,7 +113,9 @@ metric or param.
   experiment names.
 
 - `--show-json` - prints the command's output in easily parsable JSON format,
-  instead of a human-readable table.
+  instead of a human-readable table .
+
+- `--show-csv` - Print output in csv format instead of a human-readable table.
 
 - `--precision <n>` -
   [round](https://docs.python.org/3/library/functions.html#round) decimal values
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -115,7 +115,8 @@ metric or param.
 - `--show-json` - prints the command's output in easily parsable JSON format,
   instead of a human-readable table.
 
-- `--show-csv` - prints the command's output in CSV format instead of a human-readable table.
+- `--show-csv` - prints the command's output in CSV format instead of a
+  human-readable table.
 
 - `--precision <n>` -
   [round](https://docs.python.org/3/library/functions.html#round) decimal values
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -113,9 +113,9 @@ metric or param.
   experiment names.
 
 - `--show-json` - prints the command's output in easily parsable JSON format,
-  instead of a human-readable table .
+  instead of a human-readable table.
 
-- `--show-csv` - Print output in csv format instead of a human-readable table.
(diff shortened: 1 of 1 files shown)


karajan1001:
related to PR https://github.com/iterative/dvc/pull/6468

content/docs/command-reference/exp/show.md
```diff
@@ -111,7 +111,9 @@ metric or param.
   experiment names.
 
 - `--show-json` - prints the command's output in easily parsable JSON format,
-  instead of a human-readable table.
+  instead of a human-readable table .
+
+- `--show-csv` - Print output in csv format instead of a human-readable table.
```
shcheklein, line 116: Print -> prints

csv -> CSV

`table .` -> table.`

shcheklein: also, for consistency why don't we do `the command's output`

content/docs/command-reference/exp/show.md
```diff
@@ -111,7 +111,9 @@ metric or param.
   experiment names.
 
 - `--show-json` - prints the command's output in easily parsable JSON format,
-  instead of a human-readable table.
+  instead of a human-readable table .
```
jorgeorpinel, line 114:
```suggestion
  instead of a human-readable table.
```

content/docs/command-reference/exp/show.md
jorgeorpinel:
```suggestion
- `--show-csv` - prints the command's output in CSV format instead of a human-readable table.
```

jorgeorpinel:
> related to iterative/dvc#5446
related to PR iterative/dvc#6468

@karajan1001 thanks for the context! It's OK to link to the PR only 🙂

Anyway, please let us know when this is ready for review.

karajan1001:
> > related to [iterative/dvc#5446](https://github.com/iterative/dvc/issues/5446)
> > related to PR [iterative/dvc#6468](https://github.com/iterative/dvc/pull/6468)
> 
> @karajan1001 thanks for the context! It's OK to link to the PR only 🙂
> 
> Anyway, please let us know when this is ready for review.

Still in progress, so maybe need a little more day to finish it.

jorgeorpinel:
No rush. Keep an eye on the checks though please. Thanks

shcheklein:
@karajan1001 is it still in draft?

karajan1001:
As the core PR been merged, we can work on this one now. 
Need an example? but the `--show-json` flag don't have an example either.

shcheklein:
Let's fix the linters and merge. If you have time it would be nice to add some example (no need to make them reproducible, etc), but I don't this is a requirement to be honest. 

What would be great is to actually describe the possible use case for this and show an example for that - DS produces this in CSV and process then somewhere. It might be even beneficial to expand in the option description (cc @dberenbaum )

dberenbaum:
Agree that we can merge once linters are fixed!

I'm not sure about needing a usage description or example. I don't see much harm, but it seems self-evident to me: viewing the table in Excel, reading into Pandas for sorting/filtering, etc. As far as an example, we could make up one like read into pandas and filter to only experiments above a certain metric threshold, but it seems obvious and outside the scope of dvc. I think any data scientist would get it as soon as they see "csv." @karajan1001 might have good insight here also.

jorgeorpinel:
> Need an example? but the --show-json flag don't have an example either.

Adding a quick example that includes both would be nice if you can come up with it very quickly. Don't worry about it otherwise (cc @iesahin maybe you want to include that in https://github.com/iterative/dvc.org/pull/2752)

_@karajan1001 also please try to use branches directly on this repo instead of a fork, so it's easier for us to contribute e.g. if I want to fix the formatting so the checks pass I have to add your fork, fetch that one branch, etc. Thanks_

jorgeorpinel:
p.s. for this one if you just merge `master` now, Restyled should work again and that will fix the formatting automatically (in a separate PR that we can merge).

dberenbaum:
Thinking about the examples some more, I would prefer example outputs for both `--show-json` and `--show-csv` over a usage example, since what would be less obvious to users is the format of those JSON and CSV outputs, particularly since they are pretty different from one another.

jorgeorpinel:
@karajan1001 please just merge `master`. We'll figure out the examples separately. Thanks

jorgeorpinel (approved):
I'll make an issue about the examples discussed. Thanks!

PR #2748, opened 2021-08-23
dvclive - Catalyst integration

pacifikus:
PR for the second task in iterative/dvclive#112


daavoo (changes requested):
Thanks @pacifikus !

Two minor things:

- [ ]  Add a new children to the sidebar https://github.com/iterative/dvc.org/blob/master/content/docs/sidebar.json#L599
- [ ] Add a new item to the list in https://github.com/iterative/dvc.org/blob/master/content/docs/dvclive/user-guide/ml-frameworks/index.md

daavoo (approved):

PR #2750, opened 2021-08-24
Close open redirect

### LINKED ISSUE #2749: security: open redirect vulnerability
This is an open redirect vulnerability issue, that leads to victim redirecting to unwanted phishing or malicious web pages.

### Vulnerable url:

https://dvc.org//example.com/
https://dvc.org//google.com/

### Steps to reproduction :

1 : Add any url after slash endpoint
2 : While users visit the above url they will get redirected.
https://dvc.org//evil.com/

rogermparent:
Fixes #2749 

This PR tweaks our "trim trailing slashes" redirect entry to not apply to root-relative links that start with two slashes, which should close the open redirect bug since single slashes should only redirect to our site.

### DIFF
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -59,5 +59,5 @@
 
   "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/views/view-settings",
 
-  "^/([^/].+)/$                                                                           /$1"
+  "^/(.+)/$                                                                               /$1"
 ]
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -59,5 +59,5 @@
 
   "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/views/view-settings",
 
-  "^/(.+)/$                                                                               /$1"
+  "^/([^/].+)/$                                                                           /$1"
 ]
--- a/src/utils/shared/redirects.test.js
+++ b/src/utils/shared/redirects.test.js
@@ -241,9 +241,8 @@ describe('getRedirects', () => {
   })
 
   describe('Does not accidentally redirect to an external site', () => {
-    itRedirects('https://dvc.org//google.com/', '/google.com')
     itRedirects('//google.com/', '/google.com')
-    itRedirects('///evil.com/', '/evil.com')
+    itRedirects('https://dvc.org//google.com/', '/google.com')
     itRedirects('/////////evil.com/', '/evil.com')
   })
 })
--- a/src/utils/shared/redirects.test.js
+++ b/src/utils/shared/redirects.test.js
@@ -239,4 +239,8 @@ describe('getRedirects', () => {
       '/doc/use-cases/versioning-data-and-model-files'
     )
   })
+
+  describe('Does not have an open redirect bug on trailing slash removal', () => {
+    itRedirects('//google.com/', '/google.com')
+  })
 })
--- a/src/utils/shared/redirects.test.js
+++ b/src/utils/shared/redirects.test.js
@@ -240,7 +240,8 @@ describe('getRedirects', () => {
     )
   })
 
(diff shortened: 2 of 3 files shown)


rogermparent:
Looks like this fixes it! https://dvc-org-close-open-redi-pbfzmj.herokuapp.com//google.com/ doesn't redirect to Google.
I checked to make sure other review apps do have the bug, and it does as seen with https://dvc-org-dvc-exp-ssh-ihkxxryi9y.herokuapp.com//google.com/, so testing against this one should mean it works. 

Also, existing trailing slash trimming isn't broken as seen with https://dvc-org-close-open-redi-pbfzmj.herokuapp.com/doc/.

redirects-list.json
```diff
@@ -59,5 +59,5 @@
 
   "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/views/view-settings",
 
-  "^/(.+)/$                                                                               /$1"
+  "^/([^/].+)/$                                                                           /$1"
```
shcheklein, line 62: I would suggest we make it a bit more robust. Catch it in the places where we for those redirects. E.g. in the 

`/src/server/middleware/redirects/index.js`

do something like:

```
redirectLocation = redirectLocation.replace(/([^:]\/)\/+/g, "$1")
```

wdyt?

we need to prevent this from happening at the core level, so if someone makes a mistake in the redirects list we won't go back to the regression.

Ideally we should also add a test for this (we already have tests for redirects)

shcheklein:
Ideally, we need to make sure that it works in dev env as well- sometimes they differ.

redirects-list.json
rogermparent: I think that's a fine idea, especially the test, since if the regression happens it won't be obvious otherwise.

shcheklein: Yep, I would check if that place alone is enough though. I think we consume redirects link in a few places (?).

rogermparent:
The latest iteration works on `getRedirect` (or rather `matchRedirectList`, which `getRedirect` uses) to replace multiple starting slashes slashes (of any number, since it turns out any amount of slashes over one will trigger the bug) with a single slash instead. For this bug, it will redirect `dvc.org//google.com/` to `dvc.org/google.com` which then safely 404s instead of redirecting to an external site.

`getRedirect` powers all redirects across the site, so it works in tests and the dev environment.

While it feels a little heavy-handed, I can't think of any valid use case that this change would clobber.

shcheklein (approved):

src/utils/shared/redirects.test.js
```diff
@@ -239,4 +239,11 @@ describe('getRedirects', () => {
       '/doc/use-cases/versioning-data-and-model-files'
     )
   })
+
+  describe('Does not accidentally redirect to an external site', () => {
+    itRedirects('https://dvc.org//google.com/', '/google.com')
+    itRedirects('//google.com/', '/google.com')
+    itRedirects('///evil.com/', '/evil.com')
+    itRedirects('/////////evil.com/', '/evil.com')
```
jorgeorpinel, line 247: 2 of these tests are probably enough?
Also google = evil lol

jorgeorpinel (approved):

src/utils/shared/redirects.test.js
rogermparent: I removed the three slashes, but the remaining ones each cover enough to warrant inclusion IMO:
1. `//google.com/` - base case
2. `/////////evil.com/` - multiple slashes (to an extreme) which do trigger the bug when it happens
3. `http://dvc.org//google.com/` - including protocol

rogermparent:
Just did another manual check on the site to make sure, and we're good to go!

Issue #2737, opened 2021-08-20
guide: Comparing Experiments


iesahin:
It may be worthwhile to have a section in the Experiment Management about `dvc exp show`, `dvc exp diff` and their options. 

Related #2548 

- [x] Is https://github.com/iterative/dvc.org/issues/2844 still needed after this?

jorgeorpinel:
I think it could be included in https://github.com/iterative/dvc.org/pull/2752 for now, as comparing them is probably a basic need to manage them. May not even need an explicit section but rather implied throughout that guide.

iesahin:
"Managing" feels too broad for me. It can include everything we write in the "Experiment Management" chapter. 

I understand what you mean by "managing", after we have some set of experiments by running them we compare them, persist them, push them, discard them, etc.

jorgeorpinel:
> "Managing" feels too broad

Yes it's pretty broad you're right. But it should be clear what we mean given the context: listed under "management" and at the same level as "running", "sharing", "removing".

> compare them, persist them, push them, discard them

If there's a better title than "managing" let's change it, but we agree on the contents of the doc I think? Pushing is part of "sharing" though.

jorgeorpinel:
p.s. to clarify, we'll include this in the doc being created in https://github.com/iterative/dvc.org/pull/2752 right? No need to be in the same PR (although it wouldn't hurt).

jorgeorpinel:
I think "Comparing" isn't the best title here. We'll basically include `exp show` and `exp diff` right? So maybe "Listing and Comparing"? Thanks

iesahin:
"Listing and comparing" is rather verbose I think. Users will use these commands to compare the experiments. What other cases do you think of using these commands?

"Comparing experiments" might not be the best title but "Listing and comparing experiments" doesn't also seem that best title too.

shcheklein:
I think comparing should be fine (at the end you list them primarily for that reason?) ... can be also Navigating experiments (in this case `diff` won't fit precisely. Naming is hard :)

jorgeorpinel:
Navigating and Comparing?

Let's get the content first and the name should be easier to decide then 🙂

Issue #2736, opened 2021-08-20
guide: Persisting Experiments


iesahin:
A document in Experiment Management section to tell different ways of persisting experiments may be nice. 

Related #2548

jorgeorpinel:
Could also be part of https://github.com/iterative/dvc.org/issues/2548 possibly. At least initially

jorgeorpinel:
I mean https://github.com/iterative/dvc.org/pull/2752

PR #2752, opened 2021-08-25
guide: Managing Experiments

jorgeorpinel:
Extracted from https://github.com/iterative/dvc.org/pull/2690#pullrequestreview-735446663

- [x] Un-nest from #2690 when that PR is merged
- [ ] Combine and summarize the motivation text in the section index
- [ ] Finish Managing Exps guide (mainly about `exp show/compare`)
- [ ] Remove deep details from the command reference that can be moved here instead

Maybe also include (if natural)
- [ ] https://github.com/iterative/dvc.org/issues/2736
- [ ] https://github.com/iterative/dvc.org/issues/2737


content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
...
+
+Experiments created with `dvc exp run` will have an auto-generated name like
+`exp-bfe64` by default. It can be customized using the `--name` (`-n`) option:
+
+```dvc
+$ dvc exp run --name cnn-512 --set-param model.conv_units=512
+```
+
+When you create an experiment, DVC generates a Git-like SHA-1 hash from its
+contents. This is shown when you [queue experiments] with `--queue`:
```
jorgeorpinel, line 18: > From https://github.com/iterative/dvc.org/pull/2690#discussion_r694263173

> This is actually different from the hash generated after the experiments is committed, so I'm not sure it makes sense to mention it here.

content/docs/user-guide/experiment-management/managing-experiments.md
```diff
@@ -0,0 +1,68 @@
...
+
+[queue experiments]:
+  /doc/user-guide/experiment-management/running-experiments#the-experiments-queue
+
+```dvc
+$ dvc exp run --queue -S model.conv_units=32
+Queued experiment '6518f17' for future execution.
+```
+
+After running queued experiment, DVC uses the regular name mentioned earlier.
```
jorgeorpinel, line 28: > From https://github.com/iterative/dvc.org/pull/2690#discussion_r694264184

> The experiment name should show in dvc exp show even for queued experiments now.

content/docs/user-guide/experiment-management/index.md
```diff
@@ -1,5 +1,18 @@
 # Experiment Management
```
jorgeorpinel, line 1: > From https://github.com/iterative/dvc.org/pull/2690#discussion_r694266392

> I guess I'm the reason for this new section? I'm okay with leaving this in Running experiments depending on what others think. It does seem a bit disjointed as is. I think we need to at least link here from Running experiments so it's clear how to see and compare experiments.

iesahin:
For 

> Finish Managing Exps guide (mainly about exp show/compare)

I have #2737. 

I think I can also create issues for other items as well.

jorgeorpinel:
> I have #2737.

Yeah @iesahin I created this one based on https://github.com/iterative/dvc.org/pull/2690#pullrequestreview-735446663 before noticing your Exp Compare guide.

I think it could be included in this guide for now, as comparing them is probably a basic need to manage them. May not even need an explicit section but rather implied throughout this guide.

jorgeorpinel:
@iesahin should we close this and turn it into an issue?

jorgeorpinel:
Thanks @iesahin !

PR #2755, opened 2021-08-25
docs: couple workspace tooltips

jorgeorpinel:
Follow-up to https://github.com/iterative/dvc.org/pull/2453


content/docs/user-guide/basic-concepts/workspace.md
```diff
@@ -4,7 +4,7 @@ match: [workspace]
 tooltip: >-
   Directory containing all your DVC project files, e.g. raw data, source code,
   ML models. One project version at a time is visible in the workspace.  
-  See full [concept page](/doc/user-guide/basic-concepts/workspace).
+  See [full concept](/doc/user-guide/basic-concepts/workspace).
```
jorgeorpinel, line 7: Unrelated to this PR but I noticed 2 trailing spaces like in line 6 above don't turn into a new line as they should per MD rules I think. Would that be easy to enable @rogermparent ? Thanks

image

rogermparent: I'm actually not at all sure about the implementation off the top of my head, but it should certainly be possible. I thought this would be something on by default- can you try `>` over `>-`? It's possible the multi-line chomping is removing the trailing spaces like it does newlines.

shcheklein (approved):

PR #2722, opened 2021-08-16
Remove all experiments

karajan1001:
per https://github.com/iterative/dvc/pull/6394

PR #2715, opened 2021-08-13
remove specific experiments in queue

karajan1001:
per https://github.com/iterative/dvc/pull/6393

PR #2758, opened 2021-08-25
Remove all exps (2)

karajan1001:
Remaining issues from https://github.com/iterative/dvc.org/pull/2722

### DIFF
--- a/content/docs/command-reference/exp/remove.md
+++ b/content/docs/command-reference/exp/remove.md
@@ -6,7 +6,7 @@ Delete specific `dvc experiments` from the <abbr>project</abbr>.
 
 ```usage
 usage: dvc exp remove [-h] [-q | -v] [--queue | -A]
-                      [experiment [experiment ...]]
+                      [<name> [<name> ...]]
 
 positional arguments:
    experiment    Experiments to remove.


jorgeorpinel:
Thanks! What about https://github.com/iterative/dvc.org/pull/2715#pullrequestreview-737875800, could we include that here too? 🙂

jorgeorpinel:
OK @karajan1001 thanks. So you'll be adding an **Example: remove all experiments** example in this PR?

karajan1001:
> OK @karajan1001 thanks. So you'll be adding an **Example: remove all experiments** example in this PR?

Sorry not quite understand what do you mean.  In #2722 we already have
```
$ dvc exp remove -A
$ dvc exp list
```
Is this an Example: remove all experiments

jorgeorpinel:
@karajan1001 as explained in https://github.com/iterative/dvc.org/pull/2715#pullrequestreview-737875800 were just missing an example of `exp remove --queue`. I think it should be combined with the `exp remove -A` into a separate example instead of just adding to the generic **Examples** section. But feel free to use another PR, I'm merging this one so we move on 🙂

jorgeorpinel (approved):

jorgeorpinel:
p.s. we will have an example in https://github.com/iterative/dvc.org/pull/2631 now that I think about it so adding one here too is less important. Up to you @karajan1001 (what do you think @iesahin ?)

PR #2759, opened 2021-08-25
blog: cats and dogs data post

flippedcoder:
Tentative publish date: 10/05/21 ~09/14/2021~

This is a blog post building off of the pretrained model post. We show off how DVC can handle large dataset changes using remotes and how you can run new experiments on an existing model.


flippedcoder:
@dberenbaum @shcheklein Would you mind reviewing this?

content/blog/2021-09-14-cats-and-dogs.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Switching the Dataset for Your Pre-trained Models
+date: 2021-09-14
+description: |
+  You can easily switch datasets for pre-trained models with DVC.
+descriptionLong: |
+  Sometimes your problem definition changes and you need to update your dataset.
```
shcheklein, line 7: can we do a bit longer, informative descriptions? Ideally it would serve as an intro even or a summary of an intro.  We can apply styles here, emojis, etc.

content/blog/2021-09-14-cats-and-dogs.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Switching the Dataset for Your Pre-trained Models
+date: 2021-09-14
+description: |
+  You can easily switch datasets for pre-trained models with DVC.
```
shcheklein, line 5: It also almost repeats the title. Should we come up with some description that describes the problem?

content/blog/2021-09-14-cats-and-dogs.md
```diff
@@ -0,0 +1,145 @@
...
+picture: 2021-09-14/cats-and-dogs.png
+pictureComment: Switching the Dataset for Your Pre-trained Models
+author: milecia_mcgregor
+commentsUrl: https://discuss.dvc.org/t/tuning-hyperparameters-with-reproducible-experiments/821
+tags:
+  - MLOps
+  - DVC
+  - Git
+  - Experiments
+  - Pre-trained models
```
shcheklein, line 17: models -> Models ... did we use it before in the transfer learning?

content/blog/2021-09-14-cats-and-dogs.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Switching the Dataset for Your Pre-trained Models
```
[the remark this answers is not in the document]
shcheklein: btw, is it different (on a conceptual level) from the transfer learning blog post we put together before?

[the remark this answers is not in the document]
daavoo: > @dberenbaum @daavoo would you, guys, understand from this title the idea?

I'm not 100% sure if I get it. Are we pulling the weights of the "bees and ants" model and fine-tuning on a new "cats and dogs" dataset?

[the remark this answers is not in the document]
daavoo: Or is it more about we are going to take the same code used for the "bees and ants" model and reuse it for training with a new "dogs and cats" dataset?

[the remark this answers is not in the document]
dberenbaum: There has been a lot of focus on "data-centric" ML over "model-centric" ML lately, and this seems like it's addressing that (correct me if I'm wrong @flippedcoder). If so, I like the idea again and think it makes sense as a follow-up to the last one 🔥 !

However, I think it's unlikely that business problems change so dramatically that someone discards the original dataset and adopts a completely unrelated one. A more realistic scenario might be to analyze performance separately on bees and ants, and then add more bees if the model performs worse on those or vice versa (maybe you could start with a subset of the data if it's too hard to find additional bee/ant images 😄 ). A middle ground might be adding dogs and cats to the bees and ants to get a more generic animal classifier rather than discarding the original data completely.

Even better would be to combine these ideas and iterate on the data multiple times. For example, you could start with a subset of bees and ants, then add some cats, then some dogs, and finally some more of whatever animal is performing worst in the model.

I'm also not sure it's worth doing the hyperparameter tuning again in this post since it was already well covered last time and doesn't seem like it's the focus here.

I'll take a closer look once we narrow down the high-level flow of the post!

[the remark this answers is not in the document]
shcheklein: > However, I think it's unlikely that business problems change so dramatically that someone discards the original dataset and adopts a completely unrelated one.

The way I understand this post is more of a transfer learning application. In this case it would make sense to take a pre-trained model and start fine-tuning it on the new dataset, right?

But then, what is the difference with the previous one?

content/blog/2021-09-14-cats-and-dogs.md
flippedcoder: The descriptions are this short because usually we try to keep it > 80 characters. Do you want me to make the `descriptionLong` longer or just say something different and keep the same character count?

content/blog/2021-09-14-cats-and-dogs.md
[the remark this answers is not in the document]
flippedcoder: It's more about using DVC to handle data versioning, but I definitely see how that got lost in the way I wrote this. I really like @dberenbaum's idea of iterating on the data multiple time because it gets the point across better.

I'll work on this update and get it back to y'all in a day or two! Thanks for the really good feedback!

content/blog/2021-09-14-cats-and-dogs.md
flippedcoder: We actually didn't. I'll add that tag as a small thing in this PR.

content/blog/2021-09-14-cats-and-dogs.md
shcheklein: < 80 usually means - wrap lines at 80 symbols. So it's per-line style convention, not for the `description` or `descriptionLong`. Sorry for the confusion here.

content/blog/2021-09-14-cats-and-dogs.md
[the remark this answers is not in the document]
iesahin: If finding/extending the dataset seems to be difficult, you can update the post using `dataset-registry/fashion-mnist`. It has 10 classes and you can tell like "we have a socks and trousers classifier, now we add another class called shirts and see its performance."

content/blog/2021-09-14-cats-and-dogs.md
flippedcoder: Oh... Ok, that makes sense! No worries. Thanks for clearing that up.

flippedcoder:
@shcheklein @dberenbaum @daavoo @iesahin I've updated the text for the blog post if you want to start reviewing that. I'm working on getting the new tables and figuring out how to get the data registry up for the project, but the text won't change much.

iesahin:
image

I think you can also mention `dvc exp show` and use `--include-metrics`, `--include-params` for a smaller table. The table seems a bit too wide.

iesahin:
image

Also, here, I think you can use one or two params/metrics and get a smaller table. There are too many numbers to compare :)

iesahin:
The content is much improved from the previous one. Great post. Thank you @flippedcoder :)

dberenbaum:
It looks like the use case is more clear now. It still looks like transfer learning with cats and dogs instead of bees and ants, so is that okay?

Another approach is to do something like in https://thedatafrog.com/en/articles/dogs-vs-cats/, where the user found some mislabeled/bad data and removed it from the dataset. Not sure if this is the same data, or if something similar could be done here.

flippedcoder:
> It looks like the use case is more clear now. It still looks like transfer learning with cats and dogs instead of bees and ants, so is that okay?

It's transfer learning with bees, ants, cats and dogs so nothing gets replaced. This is making a more general classifier if that makes sense.

Sorry I have been able to finish this one up. I had a super unexpected conference pop up last Friday and there are 4 more over the next week. My plan is to get the updates finished this week.

dberenbaum:
> It's transfer learning with bees, ants, cats and dogs so nothing gets replaced. This is making a more general classifier if that makes sense.

Makes sense! It's a little weird since the original SqueezeNet/AlexNet models are already generic classifiers, but if we frame it as starting from a bees/ants classifier then I doubt it will be too confusing.

I do have a few suggestions 😅:
1. Are you starting from the generic AlexNet/SqueezeNet models or the fine-tuned bees/ants model? This isn't clear to me. I don't think you need to switch between architectures since that was well covered in the last blog post.
2. I also don't see the bees/ants data. I think you will want to keep that data in there if the idea is to make a more generic classifier rather than making one for dogs/cats specifically.
3. It would probably make sense to add all of the dogs/cats validation data at the start so you can see how the model performs on the more generic validation data and how it improves as you add in cats and dogs to the training data.

jendefig:
Working on graphic.  Can we adjust title?  Lots of 'm's' and two 'more's.  
Suggestion:  "Adding Data to Create a More Generic Model" or "Adding Data to Build a More Generic Model" ? @flippedcoder

flippedcoder:
@dberenbaum @iesahin Finally got the changes up for this one! I think I addressed all of the suggestions but let me know if you all see anything else.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,242 @@
...
+installed.
+
+This will make a new directory called `./cats-dogs/data/` that was downloaded
+from the DVC remote and it has images for cats and dogs. The data is in a
+similar format to the ants and bees directories so there aren't any other code
+changes we need to make.
+
+We do need to slowly add in the new data though. For starters, we'll move the
+`val` data for `cats` and `dogs` from the `/cats-dogs/data/` directory to the
+corresponding directory in `data/hymenoptera_data`.
```
dberenbaum, line 73: This is picky, but I think hymenoptera is the classification for bees and ants or something like that. It would probably be more natural to add cats and dogs to the base `data` directory, but that requires minor code and pipeline changes. Not sure if it's problematic to make changes from the previous blog post.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,242 @@
...
+Once the training epochs are finished, run the following command.
+
+```dvc
+dvc exp show --no-timestamp --include-metrics step,acc,val_acc,loss,val_loss --include-params lr,momentum
+```
+
+The `--no-timestamp` hides the timestamps from table. The `--includes-metrics`
+option lets us choose which metrics we want to show in the table. The
+`--includes-params` option does the same for hyperparameters. This gives us a
+table that's eaiser to read quickly.
```
dberenbaum, line 96:
```suggestion
table that's easier to read quickly.
```

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,242 @@
...
+│ **workspace**               │    **3** │ **0.86885** │    **0.46** │ **0.31573** │   **3.7067** │ **0.001** │ **0.09**     │
+│ **data-change**             │    **-** │       **-** │       **-** │       **-** │        **-** │ **0.001** │ **0.09**     │
+│ │ ╓ 3b3a2a2 [exp-23593] │    3 │ 0.86885 │    0.46 │ 0.31573 │   3.7067 │ 0.001 │ 0.09     │
+│ │ ╟ 93d015d             │    2 │ 0.83197 │ 0.41333 │ 0.36851 │   3.4259 │ 0.001 │ 0.09     │
+│ │ ╟ d474c42             │    1 │ 0.79918 │ 0.43333 │ 0.46612 │    3.286 │ 0.001 │ 0.09     │
+│ ├─╨ 1582b4b             │    0 │ 0.52869 │    0.39 │ 0.94102 │   2.5967 │ 0.001 │ 0.09     │
+└─────────────────────────┴──────┴─────────┴─────────┴─────────┴──────────┴───────┴──────────┘
+```
+
+You'll notice that the validation accuracy doesn't seem right compared to the
```
dberenbaum, line 111: 👍 This is exactly what we should expect because the training metrics are based on bees/ants data and the validation metrics are based on bees/ants/cats/dogs. If we broke down validation metrics by class, they would probably be way better for bees/ants than cats/dogs.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,242 @@
...
+│ │ ╟ 23dc98f             │    1 │ 0.87671 │    0.86 │ 0.35964 │  0.61713 │ 0.001 │ 0.09     │
+│ ├─╨ 99a3c34             │    0 │ 0.71429 │    0.82 │ 0.67674 │  0.62798 │ 0.001 │ 0.09     │
+│ │ ╓ 3b3a2a2 [exp-23593] │    3 │ 0.86885 │    0.46 │ 0.31573 │   3.7067 │ 0.001 │ 0.09     │
+│ │ ╟ 93d015d             │    2 │ 0.83197 │ 0.41333 │ 0.36851 │   3.4259 │ 0.001 │ 0.09     │
+│ │ ╟ d474c42             │    1 │ 0.79918 │ 0.43333 │ 0.46612 │    3.286 │ 0.001 │ 0.09     │
+│ ├─╨ 1582b4b             │    0 │ 0.52869 │    0.39 │ 0.94102 │   2.5967 │ 0.001 │ 0.09     │
+└─────────────────────────┴──────┴─────────┴─────────┴─────────┴──────────┴───────┴──────────┘
+```
+
+Suprisingly, it looks like the model did better with just the `cats` data. This
```
dberenbaum, line 219: I think the results make sense, and the last experiment with cats and dogs looks better because validation accuracy is what matters here. The training metrics for the first experiment are only for bees/ants, the second are only for bees/ants/cats, and only the third is actually evaluating performance on bees/ants/cats/dogs, so these aren't really comparable. That's why it's important that you added all the validation data at the start, so you have an apples to apples comparison for validation metrics. The training metrics only show how the model overfit to the subset of data on which it was trained.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,242 @@
...
+is a sign that we probably need to change some of the hyperparameter values and
+do some other experiments. We can look at a comparison between the experiments
+with the `cats` data and both the `cats` and `dogs` data.
+
+```dvc
+dvc plots diff exp-54e8a exp-82f70
```
dberenbaum, line 225: You could consider doing all 3 experiments here: `dvc plots diff exp-23593 exp-54e8a exp-82f70`.

dberenbaum (commented):
Looks great! I love the comparison between experiments as you add more data. I actually think you can change some of the conclusions because these results look pretty much like what should be expected!

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: I think if there's an explanation for moving the bees and ants data out of the `hymenoptera_data` folder into the `data` folder, that should be fine. Let me know what you think about the update.

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: 😱 I didn't know you can use this command with more than two experiments! Now I'm not sure why I thought it was only two... But thanks for letting me know this is a thing!

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: Thanks for this explanation! That makes this make more sense to me. I included it in the post because it'll definitely help other people too!

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: Thanks again for the really good explanation! It helps to have a better idea of what data scientists are looking at and expecting!

content/blog/2021-10-05-extending-models-with-more-data.md
dberenbaum: I probably shouldn't have said anything since it's a minor detail that has nothing to do with the technical content 😄 . 

The explanation and changes in the update look good, but it also feels like an unnecessary distraction. Maybe it's better to go back to how it was and just have a quick note, like "(cats and dogs don't really belong in the [hymenoptera](https://en.wikipedia.org/wiki/Hymenoptera) directory, but it's the easiest way to add the data)." 

Totally up to you here.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+│ │ ╟ 93d015d             │    2 │ 0.83197 │ 0.41333 │ 0.36851 │   3.4259 │ 0.001 │ 0.09     │
+│ │ ╟ d474c42             │    1 │ 0.79918 │ 0.43333 │ 0.46612 │    3.286 │ 0.001 │ 0.09     │
+│ ├─╨ 1582b4b             │    0 │ 0.52869 │    0.39 │ 0.94102 │   2.5967 │ 0.001 │ 0.09     │
+└─────────────────────────┴──────┴─────────┴─────────┴─────────┴──────────┴───────┴──────────┘
+```
+
+You'll notice that the validation accuracy is really low. That's because the
+training metrics are based on bees and ants while the validation metrics are
+based on bees, ants, cats, and dogs. If we looked at the validation metrics by
+class, they'd likely be better for bees and ants than cats and dogs
```
dberenbaum, line 124:
```suggestion
class, they'd likely be better for bees and ants than cats and dogs.
```

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+
+That means we should probably add more data to the training dataset.
+
+## Adding the cats data to the training dataset
+
+Let's add the `train` data for `cats` to the corresponding directory in
+`data/hymenoptera_data` and go through another experiment run with a different
+learning rate. With this new data, can run another experiment. One important
```
dberenbaum, line 132:
```suggestion
learning rate. With this new data, we can run another experiment. One important
```

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+The `exp-23593` and `exp-54e8a` values are the ids for the experiments you want
+to compare. You'll see a new file gets generated in the `dvc_plots` directory in
+your project. That's where you'll find the `index.html` file you should open in
+your browser. You'll see something similar to this.
+
+![plots comparing the accuracy, validation accuracy, loss, and validation loss for all epochs of each experiment](2021-10-05/with-cats-data.png)
+
+There's a huge difference in the accuracy of our model after we've add this
```
dberenbaum, line 187:
```suggestion
There's a huge difference in the accuracy of our model after we've added this
```

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+We can look at a comparison between the experiments with the `cats` data and
+both the `cats` and `dogs` data.
+
+```dvc
+dvc plots diff exp-23593 exp-54e8a exp-82f70
+```
+
+![plot of differences between model with just cats data and model with both cats and dogs data](2021-10-05/with-cats-and-dogs-data.png)
+
+There's definitely something happening with the validation metrics, so keep
```
dberenbaum, line 251: This sentence seems a little off now that you've updated the text above.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+These results make sense for the experiments we've run. We're paying attention
+to the validation accuracy here because this gives us a fair comparison of
+what's happening as we add more data.
+
+The first experiment's training metrics are for bees and ants. The second
+experiment's training metrics are for bees, ants, and cats. And the third
+experiment's training metrics are for all four classes. So we can't really
+compare these metrics. That's why it looks like this last experiment has some
```
dberenbaum, line 239: I would drop this last sentence. I think of overfitting as being more like the first experiment, where training metrics are high but validation metrics are low because the model is overfit to the specific data included in training and doesn't fit other data well.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,263 @@
...
+model overfitting happening.
+
+We can look at a comparison between the experiments with the `cats` data and
+both the `cats` and `dogs` data.
+
```
dberenbaum, line 244: It might be easier to track if you named each experiment, but it also introduces more complexity to the commands you are running, so up to you.

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: I went back to the old version and added that little note.

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: That makes sense. I updated it!

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: I think I'll skip that for this post and make sure to do it in one of the next ones.

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: Gotcha. Thanks for all of these explanations on this post! They've been super helpful for my general understanding of things. 😄

dberenbaum (approved):
Great work, @flippedcoder!

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,256 @@
+---
+title: Adding Data to Build a More Generic Model
+date: 2021-10-05
+description: |
+  You can easily make changes to your dataset using DVC to handle data versioning. This will let you extend your models to handle
```
shcheklein, line 5: 💅 style

content/blog/2021-08-24-transfer-learning-experiments.md
```diff
@@ -14,6 +14,7 @@ tags:
   - Experiments
   - Reproducibility
   - DVC
+  - Pre-trained Models
```
shcheklein, line 17: why changing this here?

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,256 @@
...
+
+Then we'll add the cats and dogs data to the training data and watch how the
+model improves as we run experiments.
+
+## Updating the dataset with DVC
+
+To add the new cats and dogs dataset to the project, we'll use this DVC command.
+
+```dvc
+dvc get https://github.com/iterative/dataset-registry blog/cats-dogs
```
shcheklein, line 58: `$ dvc get`

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,256 @@
...
+a virtual environment enabled. Then run this command:
+
+```dvc
+dvc exp run
+```
+
+Once the training epochs are finished, run the following command.
+
+```dvc
+dvc exp show --no-timestamp --include-metrics step,acc,val_acc,loss,val_loss --include-params lr,momentum
```
shcheklein, line 93: use `\` to split this long command

shcheklein: we should create a way to save table filters and sorting I think in DVC, wdyt @dberenbaum ?

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+Now you'll have a table that shows both experiments and you can see how much
+better the new one did with the `cats` data added.
+
+```dvctable
+┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┓
```
shcheklein, line 149: btw, this is a good example where we want to see in the table changes to data I think? cc @dberenbaum

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,256 @@
...
+┃ neutral:**Experiment**              ┃ metric:**step** ┃     metric:**acc** ┃ metric:**val_acc** ┃    metric:**loss** ┃ metric:**val_loss** ┃ param:**lr**    ┃ param:**momentum** ┃
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━╇━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━┩
+│ **workspace**               │    **3** │ **0.91389** │    **0.87** │ **0.20506** │  **0.66306** │ **0.001** │ **0.09**     │
+│ **data-change**             │    **-** │       **-** │       **-** │       **-** │        **-** │ **0.001** │ **0.09**     │
+│ │ ╓ 9405575 [exp-54e8a] │    3 │ 0.91389 │    0.87 │ 0.20506 │  0.66306 │ 0.001 │ 0.09     │
```
shcheklein, line 154: should we name these experiments somehow to make it easier for readers? e.g. `exp-old-data`, `exp-new-data`, etc

content/blog/2021-10-05-extending-models-with-more-data.md
dberenbaum: I extracted this point to https://github.com/iterative/dvc/issues/6708. We plan to tackle these kind of improvements soon, but there are a lot of ideas that we need to sort through and prioritize.

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: I updated this to have fewer characters per line.

content/blog/2021-08-24-transfer-learning-experiments.md
flippedcoder: I think I was just updating the tag to match the topic better. I can remove it if you think it's unnecessary.

dberenbaum:
@jendefig LGTM!

content/blog/2021-08-24-transfer-learning-experiments.md
shcheklein, line 17: no worries, I was curious about the new post though - this tag can potentially also apply there (but we are reusing the model in the new post also?)

content/blog/2021-10-05-extending-models-with-more-data.md
flippedcoder: I think @dberenbaum mentioned this earlier, but since this post has been delayed so long I didn't want to delay it further.

content/blog/2021-08-24-transfer-learning-experiments.md
flippedcoder: I think since the focus of this post is on the data changes and not the pretrained model that it's ok to leave it off of the new post.

content/blog/2021-10-05-extending-models-with-more-data.md
```diff
@@ -0,0 +1,262 @@
+---
```
shcheklein, line 1: should we rename the file, btw? ^^

shcheklein:
Looks like we have extra image files + some images are broken in the post, also it's complaining about some links to the images.

Otherwise looks good to me!

shcheklein:
@rogermparent @julieg18 hey, could please tal why does link checker fail but blog post is rendered properly?

rogermparent:
The links that are failing are image links that get heavily reprocessed into hashed images (e.g. `2021-10-05/with-cats-data.png` becomes `/static/dc01547ad9f771f11e39ba81d45658b3/f714d/with-cats-data.png` in output), which are essentially impossible to detect with source alone like the current link checker does. To handle these, we'll have to check the output of the site instead, which is something to consider when we eventually overhaul the checker. 

For this PR, these particular errors are fine to ignore.

shcheklein (approved):

flippedcoder:
@shcheklein @dberenbaum Could one of you please merge this post today? 🙂

PR #2766, opened 2021-08-30
Dvclive linker

daavoo:
Close iterative/dvclive#114


shcheklein (approved):

rogermparent (approved):
Looks fine to me, nice work!

rogermparent:
Ideally we'd consolidate all our auto-linkers into one tool that can handle all the cases we need, but the current implementation does what we need and this text-only processing is likely an extremely minor part of our whole build time even if there's a little bit of redundant work.

julieg18 (commented):
I don't see any issues! Great work!

PR #2774, opened 2021-09-01
config: sync local/CI Prettier and Restyled

### LINKED ISSUE #2641: restyled: looks like there is a discrepancy between prettier and restyled

jorgeorpinel:
for https://github.com/iterative/dvc.org/issues/2641

Should close #2641


jorgeorpinel:
☝️ Local format checks passed for the above commit:

image

CI checks and Restyled also passed:

image

jorgeorpinel:
☝️ I skipped local tests with the above commit (`git commit -n`) since I knew the formatting is wrong.

The CI tests [failed](https://app.circleci.com/jobs/github/iterative/dvc.org/10342?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) but Restyled (which took a while BTW) is [passing](https://restyled.io/gh/iterative/repos/dvc.org/jobs/1096496) 🤔

jorgeorpinel:
☝️ For some reason reverting https://github.com/iterative/dvc.org/pull/2624/commits/fb8b7727f522e9c8b1a027c40f4bb155bdcde267#diff-eb41540235b5687201bb19ed968c284453eea68563f5908a6e15ed8cbfa959de makes Restyled work again. Maybe it can't process that many files? Cc @casperdcl

jorgeorpinel:
Please review but don't merge (since checks are failing) @shcheklein. We need to revert the changes to content/docs/user-guide/contributing/blog.md once this is approved. Thanks

shcheklein:
Thanks @jorgeorpinel . What was the reason for the discrepancy?

jorgeorpinel:
I don't think it's a discrepancy. My hypothesis is that Restyled isn't able to process that many files (see https://github.com/iterative/dvc.org/pull/2624#pullrequestreview-705725815). TBH I'm not sure what the problem is here but for now this seems to help.

jorgeorpinel:
If there's a discrepancy then there are 2 problems. I didn't investigate that aspect much.

shcheklein:
Is there a link to logs / Restyled PR that we think it did wrong?

jorgeorpinel:
> If there's a discrepancy

I sorted the Prettier settings in both config files in 36ae1cff to compare them easily and they seem to be the same. There could be a Prettier version discrepancy I guess but I doubt it or that it would cause a problem.

> Is there a link to logs / Restyled PR that we think it did wrong?

There's https://restyled.io/gh/iterative/repos/dvc.org/jobs, let me see if I can find something in there... ⏳

jorgeorpinel:
> let me see if I can find something in there

The logs are not super informative. They only output when it formats files, but doesn't show the output of Prettier itself.

image
*Examlpe: https://restyled.io/gh/iterative/repos/dvc.org/jobs/1096523*

shcheklein:
so, it does change `blog.md` .. but that's expected, right?

jorgeorpinel:
> it does change blog.md

@shcheklein yeah sorry for the confusion. That log is from the current config (where it works as intended). For the previous config (when Restyled always passed) it doesn't say anything useful:

image
*Example: https://restyled.io/gh/iterative/repos/dvc.org/jobs/1096496*

casperdcl:
note there may also be issues with eslint (https://github.com/iterative/setup-dvc/issues/22)

For me, doing this:

- current `master` (7139bde)
- node v14.17.4
- `yarn && npm install` as recommended in https://dvc.org/doc/user-guide/contributing/docs#development-environment

then 

- `yarn lint-ts`: finds 5 errors
- `yarn lint-css`: succeeds
- `yarn format-all`: 10 files changed, 57 insertions(+), 60 deletions(-)
- `npm update`: updates 10 package versions in `package.json`

Opened a PR at #2779

shcheklein:
so, still not sure about the fix - why would we change the include options?

what about that PR that @casperdcl has prepared? if we update versions - does it help?

jorgeorpinel:
> should we highlight the header (colorize params, metrics, etc)?
what about that PR that @casperdcl has prepared?

I don't think https://github.com/iterative/dvc.org/pull/2779 is directly related.

> still not sure about the fix - why would we change the include options?

I'm not sure why we changed them in https://github.com/iterative/dvc.org/pull/2624#pullrequestreview-705725815 in the first place, which is what unexpectedly broke this, it seems. It was working, we "fixed" it. This may not be a propper fix but rolls back that regression at least.

My only theory remains that Restyled can't handle that many files. The Prettier env runs out of memory going over utility files or blogs, etc. and never gets to the actual content files we care about.

Should it only format changed files? Like our `format-staged` node script (used in Git hooks locally). Won't be the same as the format check but should have the same effect, I think. That script uses `pretty-quick` though, not sure if it's available in Restyled.

shcheklein:
>I'm not sure why we changed them in #2624 (review) in the first place, which is what unexpectedly broke this, it seems. It was working, we "fixed" it. This may not be a propper fix but rolls back that regression at least.

to make consistent with the `package.json` settings:

```
    "format-check": "prettier --check '**/*.{js,jsx,md,tsx,ts,json}'",
    "format-all": "prettier --write '**/*.{js,jsx,md,tsx,ts,json}'",
```

it means being compatible with CI/CD since it runs `format-check` I think

shcheklein:
Let's try to get run if `{ }` in the config, but keep all the patterns:

```
    include: 
      - "**/*.md"
      - "**/*.jsx"
      - "**/*.tsx"
      - "**/*.js"
      - "**/*.ts"
      - "**/*.json"
  ```

**Comment by shcheklein** [comment]
I would also create a ticket/ask in the resyled repo- it's open source, they can help.

**Comment by jorgeorpinel** [comment]
> to make it consistent with the package.json settings

Right, the check does `format-check` indeed. I think originally Restyled intended to replace the local Git hook instead though, for people who don't install `yarn` locally (e.g. edit via Github) and that only runs `format-staged`. But anyway, it should be consistent with the GH action check indeed, I suppose.

> Let's try to get run if { } in the config, but keep all the patterns

Done in 26fc2e3f (after `master` merge). Seems to be working for now 🤔

**Comment by jorgeorpinel** [comment]
> Seems to be working

I retested the previous config file and Restyled stopped failing again (unwanted behavior). So I guess it was just a config file formatting issue after all. Good catch @shcheklein ! I force pushed 26fc2e3f again now and it looks good:

image

I'm going to merge https://github.com/iterative/dvc.org/pull/2775 to fix the formatting here. The check should pass after that. Please merge/approve 🙂

**Review by shcheklein — DISMISSED** [review]

**Comment by jorgeorpinel** [comment]
Wait I still need to roll back the blog changes. I'll merge it...

File: content/docs/user-guide/contributing/blog.md [hunk 1]
```diff
@@ -1,7 +1,7 @@
 # Writing Blog Posts
 
-To create a new post, you need to create a PR to the DVC website
-[project](https://github.com/iterative/dvc.org).
+To create a new post, you need to create a PR to the DVC website super mega
+ultra great [project](https://github.com/iterative/dvc.org).
```
jorgeorpinel, line 4: OK hold on so https://github.com/iterative/dvc.org/pull/2775/files supposedly fixed this ☝️ (which I need to undo) but it's still not fixed according to [the check](https://app.circleci.com/pipelines/github/iterative/dvc.org/10439/workflows/08bb5f08-74d4-4487-9518-7ac3ce117250/jobs/10498). The line looks technically good to me... ⏳

content/docs/user-guide/contributing/blog.md
```diff
@@ -41,7 +41,6 @@ tags:
   - Version Control
   - AI
 ---
-
 ```
```
jorgeorpinel, line 44: Shoot so now Restyled removed this line but `prettier` adds it back, so the check is failing... Looks like there's still a config discrepancy somewhere 😕

jorgeorpinel: Hmmm I reintroduced the line in a9b208e6 and Restyled left it alone this time 🤷

jorgeorpinel:
OK everything is actually passing now. I think we can reapprove and merge this now @shcheklein. Sorry for the confusion. I'll keep an eye on it... 👀

shcheklein (approved):

.restyled.yaml
```diff
@@ -3,13 +3,19 @@ restylers:
...
+        '--no-semi',
+        '--single-quote=true',
         '--tab-width=2',
-        '--use-tabs=false',
-        '--prose-wrap=always'
+        '--trailing-comma=none',
+        '--use-tabs=false'
       ]
-    include: ['**/*.{js,json,md,yaml,yml}']
+    include:
```
shcheklein, line 15: I would put a comment here, btw

jorgeorpinel: Done in f61f8be8.

PR #2776, opened 2021-09-01
Update config.md for #6419 (i.e. #6111)

meierale:
Updated the config.md to document the new state.dir and index.dir options.

Per https://github.com/iterative/dvc/pull/6419

- [x] Also finish #2826


content/docs/command-reference/config.md
```diff
@@ -238,6 +238,25 @@ to learn more about the state file (database) that is used for optimization.
   When an entry in the database is used (e.g. during the `dvc status`), DVC
   updates the timestamp on that entry. This way, when the database needs a
   cleanup, DVC can sort entries chronologically, and remove the oldest ones.
+  
+- `state.dir` - specify a specific location for storing the state. This is
+  necessary when using `dvc` on a NFS or some other mounted volume.
```
jorgeorpinel, line 243:
```suggestion
- `state.dir` - specify a custom location for the state file. This is
  necessary when using DVC on NFS or other mounted volumes.
```

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,25 @@ to learn more about the state file (database) that is used for optimization.
...
+  (By default, store SQLite-based remote indexes and state's
+  `links` and `md5s` caches in the repository itself to avoid any
+  possible state corruption in 'shared cache dir' scenario.) 
+  Use this override with care.
+  
+### index
+
+See
+[Internal directories and files](/doc/user-guide/project-structure/internal-files), 
+specifically `.dvc/tmp/index` to learn more about the remote index files.
```
jorgeorpinel, line 253:
```suggestion
See `.dvc/tmp/index` in 
[Internal directories and files](/doc/user-guide/project-structure/internal-files), 
to learn more about remote index files.
```

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,25 @@ to learn more about the state file (database) that is used for optimization.
...
+
+- `index.dir` - overwrite the default location of the directory for remote index 
+  files that are used for optimizing `dvc push`, `dvc pull`, `dvc fetch` and 
+  `dvc status -c` operations.
+  This may be necessary when using `dvc` on a NFS or some other mounted volume.
```
jorgeorpinel, line 258:
```suggestion
- `index.dir` - specify a custom location for the directory where remote index 
  files will be stored. This may be necessary when using DVC on NFS or other mounted volumes.
```

jorgeorpinel (commented):
A few initial copy edits

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,23 @@ to learn more about the state file (database) that is used for optimization.
   When an entry in the database is used (e.g. during the `dvc status`), DVC
   updates the timestamp on that entry. This way, when the database needs a
   cleanup, DVC can sort entries chronologically, and remove the oldest ones.
+  
+- `state.dir` - specify a custom location for the state file. This is
+  necessary when using DVC on NFS or other mounted volumes.
+  (By default, store SQLite-based remote indexes and state's
+  `links` and `md5s` caches in the repository itself to avoid any
+  possible state corruption in 'shared cache dir' scenario.)
```
jorgeorpinel, line 246: @meierale can you clarify what this long parenthesis means? Does it intend to document the default behavior when this is not set?

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,23 @@ to learn more about the state file (database) that is used for optimization.
   When an entry in the database is used (e.g. during the `dvc status`), DVC
   updates the timestamp on that entry. This way, when the database needs a
   cleanup, DVC can sort entries chronologically, and remove the oldest ones.
+  
+- `state.dir` - specify a custom location for the state file. This is
+  necessary when using DVC on NFS or other mounted volumes.
```
jorgeorpinel, line 243: > This is necessary when using DVC on NFS or other mounted volumes

@meierale is that always the case, or should it be "this may be necessary"?

content/docs/command-reference/config.md
jorgeorpinel, line 246: @meierale can you clarify what this long parenthesis means? Does it intend to document the default behavior when this is not set?

jorgeorpinel: If this is about default behavior, I thin kit would be best to document that in https://dvc.org/doc/user-guide/project-structure/internal-files instead.

jorgeorpinel: BTW https://dvc.org/doc/user-guide/project-structure/internal-files doesn't mention "state file" so probably a terminology update is needed in this doc or in that other one? I.e. what internal files does `state.dir` affect exactly? (I guess `.dvc/tmp/md5s` and `.dvc/tmp/links`) Thanks

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,23 @@ to learn more about the state file (database) that is used for optimization.
...
+  (By default, store SQLite-based remote indexes and state's
+  `links` and `md5s` caches in the repository itself to avoid any
+  possible state corruption in 'shared cache dir' scenario.) 
+  Use this override with care.
```
jorgeorpinel, line 247: What's the risk of overriding this? Maybe the warning should be more specific/scary 👻

content/docs/command-reference/config.md
meierale: @jorgeorpinel That's a good question. To be honest I don't really know whether it is always necessary. It certainly is when working with azure ml. Maybe it's better to write "may be necessary", I agree.

content/docs/command-reference/config.md
meierale: > If this is about default behavior, I thin kit would be best to document that in https://dvc.org/doc/user-guide/project-structure/internal-files instead.

Yes that's correct in my opinion: the default behavior of dvc (and its internal sqlite use) is to store the caches of `links` and `md5s` internally inside the .dvc folder of the project. I think it does makes sense indeed to document that on the ìnternal-files` page.
This default behavior however may cause problems if the project (its local folder location) resides on a NFS or some other mounted volume that is not directly attached to the users machine. (e.g. `dvc init` or other commands will fail due to some weird SQLite error.)
Thus the PR and this documentation ;-)

content/docs/command-reference/config.md
meierale: What should be avoided is that all of a sudden multiple users that each have the project checked out locally are using the same directory for the state.dir and index.dir. 
Let's assume we have a linux machine with multiple users and that each user has the project checked out locally at `/home/username/projectname`. No problems so far, because every folder is in fact different.
But if on that machine the users would decide to override the default behavior and put the `links` and/or `md5s` to `/tmp`, then we would end up with the same folder for state and index dir being used by multiple users. This would likely result in state curruptions.

Do you know what I mean? (but I figured such a scenario is overkill to describe here)... What do you think?

content/docs/command-reference/config.md
meierale: > BTW https://dvc.org/doc/user-guide/project-structure/internal-files doesn't mention "state file" so probably a terminology update is needed in this doc or in that other one? I.e. what internal files does `state.dir` affect exactly? (I guess `.dvc/tmp/md5s` and `.dvc/tmp/links`) Thanks

correct 👍

content/docs/command-reference/config.md
jorgeorpinel:
```suggestion
- `state.dir` - specify a custom location for the state file. This may be
  necessary when using DVC on NFS or other mounted volumes.
```

content/docs/command-reference/config.md
jorgeorpinel: > I think it does makes sense indeed to document that on the internal-files page.

Please edit that file when you can and I'll help finalize the text in both places with some Github suggestions 🙂 although if you become unavailable np, just please lmk and we'll take it over when possible.

> > internal-files doesn't mention "state file" so probably a terminology update is needed
>
> correct 👍

OK so for now can remove this whole defaults parenthesis probably, since the section already links to the Internal Files doc where that's explained. Thanks

```suggestion
```

content/docs/command-reference/config.md
jorgeorpinel: Thanks for the explanation! I think we can summarize it like this:

```suggestion
  ⚠️ Make sure no other project shares the same `state.dir` location.
```

And same for `.dvc/tmp/index`, I guess?

jorgeorpinel:
So https://github.com/iterative/dvc/pull/6419 is still undecided though? We may want to wait until that's near final to finish this PR. Please remind me when to look again @meierale ! Thanks

meierale:
> So [iterative/dvc#6419](https://github.com/iterative/dvc/pull/6419) is still undecided though? We may want to wait until that's near final to finish this PR. Please remind me when to look again @meierale ! Thanks

Hi @jorgeorpinel - I suggest to prepare the docs (this PR and https://github.com/iterative/dvc.org/pull/2826) so that they are ready when needed. But only merge when #6419 is being merged as well. 
@efiop: Now that 2.7.x are out, are you still going to merge #6419 and introduce your changes on top? Or what's the current plan/approach?

efiop:
> Now that 2.7.x are out, are you still going to merge #6419 and introduce your changes on top? Or what's the current plan/approach?

@meierale Correct, waiting for the right moment with other stuff getting in and needing a release.

jorgeorpinel:
> I suggest to prepare the docs (this PR and #2826)

Absolutely, thanks. I just wonder whether https://github.com/iterative/dvc/pull/6419 can change in a way that requires reviewing the docs further so please lmk once that is approved to make a final review. I'll probably commit some suggestions for now though.

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,20 @@ to learn more about the state file (database) that is used for optimization.
...
+  
+### index
+
+See `.dvc/tmp/index` in 
+[Internal directories and files](/doc/user-guide/project-structure/internal-files), 
+to learn more about remote index files.
+
+- `index.dir` - specify a custom location for the directory where remote index 
+  files will be stored. This may be necessary when using DVC on NFS or other mounted volumes.
+  Use this override with care.
```
jorgeorpinel, line 254:
```suggestion
  ⚠️ Make sure no other project shares the same `state.dir` location.
```

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,20 @@ to learn more about the state file (database) that is used for optimization.
   When an entry in the database is used (e.g. during the `dvc status`), DVC
   updates the timestamp on that entry. This way, when the database needs a
   cleanup, DVC can sort entries chronologically, and remove the oldest ones.
+  
+- `state.dir` - specify a custom location for the state file. This may be
+  necessary when using DVC on NFS or other mounted volumes.
+  ⚠️ Make sure no other project shares the same `state.dir` location.
```
jorgeorpinel, line 244: Sorry, still confused. Which state file does this refer to? https://dvc.org/doc/user-guide/project-structure/internal-files doesn't use that term and the note introduced in #2826 doesn't clarify this either. Does this affect .dvc/tmp/md5s or .dvc/tmp/links ? Or both

meierale: I tried to document it more clearly: state.dir affects the `links` and `md5s` directories, index.dir the `index` directory.

efiop:
Thanks for the PR, @meierale ! 🙏 

@jorgeorpinel Let's hold on this PR for now, please, we need to reconsider some stuff on dvc side and we'll then come back to this ourselves.

jorgeorpinel:
OK @efiop closing for now then... But https://github.com/iterative/dvc/pull/6419 was merged so shouldn't we have this documented? And yes thanks again @meierale !

efiop:
Ok, keeping and releasing this for now. So reopening.

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,21 @@ to learn more about the state file (database) that is used for optimization.
   When an entry in the database is used (e.g. during the `dvc status`), DVC
   updates the timestamp on that entry. This way, when the database needs a
   cleanup, DVC can sort entries chronologically, and remove the oldest ones.
+  
+- `state.dir` - specify a custom location for the state file directories 
+  (the `links` and `md5s` directory which by default reside in `./dvc/tmp`). 
+  This may be necessary when using DVC on NFS or other mounted volumes.
+  ⚠️ Make sure no other project shares the same `state.dir` location.
```
efiop, line 245: This and the one for `index` are not necessary, as we generate a unique hash based on the path to the repo that is using this, so there is no real danger in multiple projects using the same dirs.
```suggestion
```

content/docs/command-reference/config.md
```diff
@@ -238,6 +238,21 @@ to learn more about the state file (database) that is used for optimization.
...
+  
+### index
+
+See `.dvc/tmp/index` in 
+[Internal directories and files](/doc/user-guide/project-structure/internal-files), 
+to learn more about remote index files.
+
+- `index.dir` - specify a custom location for the directory where remote index 
+  files will be stored. This may be necessary when using DVC on NFS or other mounted volumes.
+  ⚠️ Make sure no other project shares the same `index.dir` location.
```
efiop, line 255:
```suggestion
```

efiop (approved):

jorgeorpinel:
Merged this via https://github.com/iterative/dvc.org/pull/2891 but this is still showing open for some reason... 🤔 

The changes are already in `master`: https://github.com/iterative/dvc.org/blob/master/content/docs/command-reference/config.md#index

jorgeorpinel:
Resolved conflicts, now this PR has 0 changes 🤷 Merging!

jorgeorpinel:
>  Also finish Add note regarding default behavior of sqlite dirs #2826

Need to do that though...

jorgeorpinel:
p.s. thanks for taking over @efiop

PR #2802, opened 2021-09-07
ref: deprecate `digest` option from HTTP remote

isidentical:
We have removed the digest option in https://github.com/iterative/dvc/pull/6525


efiop (changes requested):
Let's not remove it, but just clarify that it has been removed and put a link to a ticket in dvc repo, so people could give us feedback.

jorgeorpinel:
Agree to leave it for now with a note on which version it was removed but do we need to link to some GH ticket? I'd prefer to avoid that. Users can reach us via regular channels with feedback. Or is there a special reason we want to encourage feedback on this? Thanks

efiop:
@jorgeorpinel Just to have 1 place to point people to.

content/docs/command-reference/remote/modify.md
```diff
@@ -909,10 +909,6 @@ by HDFS. Read more about by expanding the WebHDFS section in
   - `basic` -
     [basic authentication scheme](https://tools.ietf.org/html/rfc7617). `user`
     and `password` (or `ask_password`) parameters should also be configured.
-  - `digest` -
-    [digest Access Authentication Scheme](https://tools.ietf.org/html/rfc7616).
-    `user` and `password` (or `ask_password`) parameters should also be
-    configured.
   - `custom` - an additional HTTP header field will be set for all HTTP requests
```
jorgeorpinel, line 912: Restoring per https://github.com/iterative/dvc.org/pull/2802#issuecomment-916507027
```suggestion
  - `digest` -
    [digest Access Authentication Scheme](https://tools.ietf.org/html/rfc7616).
    `user` and `password` (or `ask_password`) parameters should also be
    configured.
  - `custom` - an additional HTTP header field will be set for all HTTP requests
```

jorgeorpinel (dismissed):

efiop (approved):

shcheklein:
Was it deprecated or removed though?

efiop:
The difference is vague. I've suggested to include a link above, but that might be excessive. The current approach will work fine for now.

shcheklein:
Agreed on the link. It's more of a question - keep `deprecated` term or use `remove`. Even from the conversation you both have here:

`but just clarify that it has been _removed_`
`which version it was removed`
`We have removed the digest option`

it seems that it was removed? So, why don't explicitly say `removed in ...`

jorgeorpinel:
I changed to "removed".

I don't like the idea of linking to GH issues in general (seems fragile; we already have a few which are probably no longer relevant) but feel free to add the link to the appropriate ticket (I'm not sure which one @efiop referred to) and merge @isidentical.

PR #2816, opened 2021-09-08
Fix chrome auto-dark mode bugs in docs

julieg18:
* mentioned in #2794


src/components/Documentation/Layout/SearchForm/styles.module.css
```diff
@@ -2,13 +2,6 @@
   box-sizing: border-box;
   height: 60px;
   padding-top: 10px;
-  background: rgb(238, 244, 248);
-  background: linear-gradient(
```
julieg18, line 6: Currently [google dark mode doesn't work with linear-gradients](https://bugs.chromium.org/p/chromium/issues/detail?id=1223383&q=component%3ABlink%3EPaint%3EDarkMode&can=2), which is why the docs search input and our blog items look wrong in dark mode. I took off the linear gradient in the search input since (as far as I could tell) it was barely noticeable. I couldn't find a easy way to fix the blog items, but we could wait for Google Chrome to fix the bug 🤔

rogermparent: Seems barely noticeable when working, if it's causing an issue then removing it is totally fine.

rogermparent (approved):

PR #2830, opened 2021-09-16
Fix broken link to Azure authentication examples

josecelano:
In the section https://dvc.org/doc/command-reference/remote/add#supported-storage-types, there is a broken link on "Click for Microsoft Azure Blob Storage". The link should point to: https://dvc.org/doc/command-reference/remote/modify#example-some-azure-authentication-methods.


shcheklein (approved):

shcheklein:
thanks @josecelano !

Issue #2730, opened 2021-08-18
guide: Add `--glob` and `--recursive` options to Running Experiments


iesahin:
It looks there are bugs in `dvc exp run` related to these options. It's better to remove these options from the ongoing guide in #2690. 

## Waiting For

- [x] iterative/dvc#6456
- [x] iterative/dvc#6457
- [ ] ref: Revise `exp run` not to contain any options automatically inherited from `dvc repro`

iesahin:
It looks #6457 isn't going forward, so I'll just add `--recursive` to the guide.

jorgeorpinel:
So will https://github.com/iterative/dvc.org/pull/2846 close this?

iesahin:
I couldn't decide if it closes. It may be better to remove `dvc repro options` statement in the usage text `dvc exp run` command reference and list the supported `dvc repro` options only.

jorgeorpinel:
Sorry @iesahin can you clarify your answer? I did not understand how that's related to this issue about the guide.

Also did you add

> Revise exp run not to contain any options automatically inherited from dvc repro

? Why would that be and how is it related here again? I'm lost, can you re-scope this issue and/or create other ones so they're clear? Thanks

jorgeorpinel:
OK I checked the PR and I think we don't need to cover `--recursive` in the Running Exps guide, just like we don't cover all the `repro` options in it (I think, right?). This is more related to data pipelines.

> list the supported dvc repro options only

Let's definitely create an issue about this. What options are we listing in `exp run` that are actually not supported?

Thanks

PR #2846, opened 2021-09-22
guide: Add --recursive` options to Running Experiments

iesahin:
Related #2730

`--glob` seems to take some time, so this one only adds the `--recursive`

### DIFF
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
 > Note that the order in which pipelines are executed is not guaranteed; Only
 > the internal order of stage execution is.
 
-### Interactive reproduction
+### Running pipelines recursively
+
+DVC supports pipelines defined in more than one `dvc.yaml` file. These can
+reside in nested folders and you may want to run all these in a single command.
+
+```
+$ tree
+.
+├── dir
+│   ├── deeper-dir
+│   │   └── dvc.yaml
+│   ├── dvc.lock
+│   └── dvc.yaml
+├── dvc.lock
+└── dvc.yaml
+```
+
+In this example, if you want to run the pipelines defined in `dir/dvc.yaml` and
+`dir/deeper-dir/dvc.yaml` in a single command, you can issue:
+
+```dvc
+$ dvc exp run --recursive dir/
+```
+
+It will run all the pipelines under `dir/`. You may also use `.` as a directory,
+so `dvc exp run -R .` to run all the pipelines recursively, starting from the
+current directory.
+
+### Running stages interactively
 
 When you want to have more granular control over which stages are run, you can
 use the `--interactive` option. This flag allows you to confirm each stage


content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
 > Note that the order in which pipelines are executed is not guaranteed; Only
 > the internal order of stage execution is.
 
-### Interactive reproduction
+### Running pipelines recursively
+
+DVC supports pipelines defined in more than one `dvc.yaml` file. These can
```
jorgeorpinel, line 84: Or a single dvc.yaml file can have multiple pipelines.

jorgeorpinel: Also each pipeline can be spread among several dvc.yaml files.

Let's find a way to describe all these possibilities in general without explaining each one here.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
...
+reside in nested folders and you may want to run all these in a single command.
```
jorgeorpinel, line 85:
```suggestion
reside in subfolders inside the workspace, and you may want to run all of them at once.
```

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
...
+
+```
+$ tree
```
jorgeorpinel, line 88:
````suggestion
```dvc
````

May or may not need `dvc` here, does it make a difference visually?

File: content/docs/user-guide/experiment-management/running-experiments.md [hunk 4]
```diff
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
...
+│   ├── deeper-dir
+│   │   └── dvc.yaml
+│   ├── dvc.lock
+│   └── dvc.yaml
+├── dvc.lock
+└── dvc.yaml
+```
+
+In this example, if you want to run the pipelines defined in `dir/dvc.yaml` and
+`dir/deeper-dir/dvc.yaml` in a single command, you can issue:
```
jorgeorpinel, line 100:
```suggestion
To run the pipelines defined in `dir/dvc.yaml` as well as the ones in
`dir/deeper-dir/dvc.yaml` with a single command, you can use:
```

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -79,7 +79,35 @@ $ dvc exp run --all-pipelines
...
+
+```dvc
+$ dvc exp run --recursive dir/
+```
+
+It will run all the pipelines under `dir/`. You may also use `.` as a directory,
+so `dvc exp run -R .` to run all the pipelines recursively, starting from the
+current directory.
```
jorgeorpinel, line 108:
```suggestion
It will run all the pipelines under `dir/`.

You may also use `.` as target (e.g.
`dvc exp run -R .`) to run all pipelines in the entire project
(recursively starting from the current directory).
```

jorgeorpinel: (Committing this last one copy edit)

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: Running recursively is about "more than one `dvc.yaml` file." Multiple pipelines is a different topic, and we may need to update `targets` discussion to include this possibility. I think introducing multiple pipelines here would be distracting to the base topic.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: I didn't try but I don't think so. We may better put `bash` if we remove `dvc` though, it may have some syntax coloring.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: If you're trying to avoid the topic of multiple pipelines you shouldn't use plural "pipelines" anywhere in this section. Not also in the text above but also bellow:

> To run the pipelines defined in `dir/dvc.yaml` as well as the ones in
`dir/deeper-dir/dvc.yaml` with a single command

and 

> It will run all the pipelines under `dir/`.

I agree we should focus on one thing or the other here so please make those singular pipeline.

jorgeorpinel: p.s. do we need an issue or check box about covering multiple pipelines? TBH probably not in the context of experiments.

jorgeorpinel: > we should focus on recursion only if possible

p.p.s. actually I'm not too sure about this. Probably the main use of `--recursive` is to run multiple pipelines... It makes sense to put a single pipeline in each dvc.yaml file. Splitting pipelines in different files seems much more advanced and we don't even have a single example of that throughout our docs.

jorgeorpinel: OK I thought about this. TBH I don't think we need this section here. The option just comes from `repro` and it will be documented in the pipelines guide as well as the `repro` and `exp run` cmd refs. This is not really about running experiments.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: `dvc` extends from `bash` or some other terminal highlighting I believe, which is why we don't use any other one. Please always render contributions to check how they look. Thanks

jorgeorpinel:
I'd just close this (and https://github.com/iterative/dvc.org/issues/2730) per https://github.com/iterative/dvc.org/pull/2846#discussion_r717249749.

iesahin:
> I'd just close this (and maybe #2730) per [#2846 (comment)](https://github.com/iterative/dvc.org/pull/2846#discussion_r717249749).

`dvc exp run` does not support all options of `dvc repro` and considering cmd ref as a substitute for the UG is something against our decision to write a UG in the first place. UG contains (or may contain) some duplicate content from the REF. That seems normal to be.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: I still this is not about running experiments at all. It's about how to structure data pipelines. I even plan to remove the concept of pipelines completely from this guide when possible, so there's no point in adding this. We merged the guide including that section for now for the sake of making progress.

jorgeorpinel: BTW we can reuse this content when we get to the pipelines guide so it won't be lost even if we close the PR, which is my recommended action.

shcheklein: hmm, @jorgeorpinel why do you think that is not about experiments? If `dvc.yaml` files are independent then you have to specify explicitly or in some other way an entry point to run, right? `--recursive` can be considered as a way to specify it. 

I'm not sure about the details level for (e.g. do we need the whole subsection or not for this, etc), also not sure about the name. I would create a general section like "Specifying pipelines to run" and would name a few options there how to put a target, limit the scope, etc, etc when you run an experiment. To simplify we can even make a reference to other sections related to pipelines or whatnot.

WDYT, @jorgeorpinel @iesahin ?

content/docs/user-guide/experiment-management/running-experiments.md
shcheklein: `dvc` is needed - it expands bash, also It handles things like `$` (it becomes non-active text - no way to copy it), etc.

jorgeorpinel:
> dvc exp run doesn't support all dvc repro options BTW. We can't assume the reader will be familiar with repro.
> and considering cmd ref as a substitute for the UG is something against our decision to write a UG in the first place.

I'm not suggesting to only have `repro/exp run --recursive` info in the ref. We'll put it in a Data Pipelines guide at some point. It's not about experiment management.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > If dvc.yaml files are independent then you have to specify explicitly or in some other way

OK, this is true. So do we want to cover this scenario in the guide as it's own section? But then maybe we need a separate "Organizing Experiments" page after all (instead of https://dvc.org/doc/user-guide/experiment-management#organization-patterns) and we mention it there as a tool. Or just mention `exp run --recursive` in that existing section for now (it can be a single sentence probably and link to the cmd ref for now).

I think the part about multiple or recursive pipelines is a topic for the Data Pipelines guide though.

> general section like "Specifying pipelines to run"

I think we should stop including the pipeline concept here since we plan to remove it from the experiments guide not to mix in concepts. Conceptually you don't need a pipeline (just technically a single stage that codifies some code) to run experiments.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Yeo but in the suggestion above I removed the `$ tree` line 😬 which is why I'm not sure. I'll merge it and see how it renders... ⏳

content/docs/user-guide/experiment-management/running-experiments.md
shcheklein: I think we should indeed try to avoid this, but I'm not sure why they could not be connected at all. Let's say we indeed have a repo with pipelines, some complicated set of pipelines (e.g. monorepo with a lot of models). When I run an experiment I understand that I have a lot of them and I need to specify something to run, right?

> Conceptually you don't need a pipeline (just technically a single stage that codifies some code) to run experiments.

You don't need != you don't use them. You don't need them for a "happy path", but if you use pipelines in the project you essentially run them when you run an experiment.

jorgeorpinel: Avoiding the concept doesn't mean they can not be connected. They can and if it's too hard to avoid that word then it's OK, we can include it and even link to Pipeline docs. I just think we should always be able to say "run the experiment" in this guide (focus on experimentation) and not "run pipelines". After all it's `exp run`, not `repro`.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: OK so it doesn't seem to matter if there's no `$ tree` but I say we keep it since it doesn't hurt and the md code looks more consistent.

image
https://dvc-org-iesahin-issue27-pwlhbh.herokuapp.com/doc/user-guide/experiment-management/running-experiments

> Unless you have a reason to keep `$ tree` @iesahin — keeping in mind it's Linux-specific (not a Windows tool and would technically be `tree -A` in Mac (close though).

content/docs/user-guide/experiment-management/running-experiments.md
shcheklein: Fair enough. But I would not go into extremes here. I don't see we should be avoiding it completely. `exp` doesn't replace pipelines at all. They are two different things. And we do this section, the way I understand it is that it describes an advanced case - when you have a project with multiple pipelines and you do experiments with it.

jorgeorpinel: > BTW, I created https://github.com/iterative/dvc.org/issues/2883 (long awaited) so we can add check boxes there for things we want to cover related to pipelines.

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -79,7 +79,37 @@ $ dvc exp run --all-pipelines
 > Note that the order in which pipelines are executed is not guaranteed; Only
 > the internal order of stage execution is.
 
-### Interactive reproduction
+### Running pipelines recursively
+
+DVC supports pipelines defined in more than one `dvc.yaml` file. These can
+reside in subfolders inside the workspace, and you may want to run all of them
+at once. Example project:
```
jorgeorpinel, line 86: > Continues https://github.com/iterative/dvc.org/pull/2846#pullrequestreview-763545319

To summarize on this, `exp run --recursive` has to do with how your project structure is organized so in case people have several dvc.yaml files they may need it. I don't think we need to mention pipelines to explain this, even when they user may be using them indeed — they would know what they are and how/where they codified them already.

That said if you want to mention both the cases with multiple dvc.yaml files and the one with multiple pipelines as part of the explanation here, that's OK too. We just shouldn't focus on the concept of pipelines I believe.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Moving this discussion to a new review as this one grew too much! See https://github.com/iterative/dvc.org/pull/2846#pullrequestreview-769287727 @iesahin @shcheklein thanks

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Another option or an additional thing we may want to do is to create a separate "Organizing Experiments" page (instead of https://dvc.org/doc/user-guide/experiment-management#organization-patterns) and mention `exp run --recursive` as a tool in there. This could be a separate issue or check box somewhere. Up to you @iesahin

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: I didn't get the problem with `tree`. We're using it all over the documentation, even supposedly hands-on parts. We can skip it or show only a directory hierarchy but I don't think that will matter much.

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: This section is about "running experiments" and I believe it involves experiments that use pipelines in directory hierarchies. We can mention `--recursive` in other places as a way to organize the experiments, but _here_ we are discussing _how to run experiments_ if their parts are found in recursive directories.

iesahin: Pipelines are an inherent part of experiments, even if we have "experiments without pipelines" at some point, this guide will have to cover "experiments with pipelines." Specifying pipelines, organizing them or adding/deleting stages to them are irrelevant topics to this section but "running pipelines" should be covered here IMO. If we'd remove all "pipeline related stuff" from this section, about 90% of the material should be deleted. We don't have any means to run the experiments other than running the pipelines. 

This individual section tells how to _run_ the experiments _if_ their parts are organized in directory hierarchies, not _organizing_ experiments into directory hierarchies.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: `tree` is GNU-specific. Ideally we should remove it from other places too IMO. The figure with the directory structure (`tree` output) is self explanatory I think. But yes this is minor so up to you.

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: > here we are discussing how to run experiments if their parts are found in recursive directories

Right, it makes sense to mention `exp --recursive` here. I just don't think we need a full section that describes the project's organization. It can be a paragraph (maybe even a single sentence) in the existing **Running all pipelines** section.

> Pipelines are an inherent part of experiments
at some point, this guide will have to cover "experiments with pipelines."

Disagree. You don't need a pipeline for experiments, you just need a dvc.yaml file. It's more of a technicality and in the future this implementation will be made transparent with `exp init` anyway. But we've discussed this a lot already...

> If we'd remove all "pipeline related stuff" from this section, about 90% of the material

Most would stay, just explained from a different perspective, as mentioned in https://github.com/iterative/dvc.org/pull/2846#discussion_r718142001.

iesahin: > Disagree. You don't need a pipeline for experiments, you just need a dvc.yaml file.

In the current implementation, we need a pipeline at least with a single stage to run an experiment. Even `dvc exp init` will create a _default pipeline_ (`dvc.yaml` with some defaults) and run it.

PR #2847, opened 2021-09-22
Updating Events page

jendefig:
Adding all the new events to the Community page through 11/18/21


shcheklein:
I wonder if we should change the banner now - make it about events also? or move events to the top?

jendefig:
> I wonder if we should change the banner now - make it about events also? or move events to the top?

Hmm!  Good point.  I think moving Learn and Events up is a good idea.  Then we need to decide on how the testimonial.to get incorporated which would be good with the Community and  contribute.   Need to think about this more....

PR #2851, opened 2021-09-24
Fix bugs in Gatsby 3 upgrade

rogermparent:
Attempting to get #2848 to build


julieg18:
Weird, this is what site looks like on my end: 

image

Do we need to try rebuilding it?

rogermparent:
That was weird, I got it too and a rebuild did indeed fix the fonts for me. The public folder was also malformed and triggered a full rebuild- not a good portent :grimacing:

PR #2853, opened 2021-09-25
ref: alias `ls` for `list`

ap-kulkarni:
Per https://github.com/iterative/dvc/pull/6683 (added alias to `dvc list` as `dvc ls` to fix https://github.com/iterative/dvc/issues/6651)

### DIFF
--- a/content/docs/command-reference/list.md
+++ b/content/docs/command-reference/list.md
@@ -3,6 +3,8 @@
 List project contents, including files, models, and directories tracked by DVC
 and by Git.
 
+> Aliased to `dvc ls`.
+
 > Useful to find data to `dvc get`, `dvc import`, or for `dvc.api` functions.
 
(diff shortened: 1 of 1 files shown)


jorgeorpinel (approved):
Thanks @ap-kulkarni .

@skshetry do you think we also want to list this in the nav like `dvc doctor`?

Cc @shcheklein, may be time for an engine feature that deals with this cases?

skshetry:
I'd prefer to make `ls` the default in the docs, and make `list` an alias, as it's more intuitive and easier to type.

jorgeorpinel:
I like the idea but I think docs need to reflect the product as much as possible so if `ls` is the alias we should keep `list` as the main entry for now. I don't think the question on whether to add both to the nav was answered but I'm merging this for now, lmk though!

PR #2877, opened 2021-10-01
Document short aliases `--csv/--json/--md` flags

skshetry:
We replaced `--show-csv`/`--show-json`/`--show-md` with short aliases `--csv/`--json`/`--md` respectively in https://github.com/iterative/dvc/pull/6711. Though even if the `--show-*` are still supported, I am removing those mentions from the docs, as it may confuse users. I did not touch blogs when making the changes.

The change is not released yet, so we may want to wait for merging this.

### DIFF
--- a/content/docs/command-reference/diff.md
+++ b/content/docs/command-reference/diff.md
@@ -9,7 +9,7 @@ workspace.
 ```usage
 usage: dvc diff [-h] [-q | -v]
                 [--targets [<paths> [<paths> ...]]]
-                [--show-json] [--show-hash] [--show-md]
+                [--json] [--show-hash] [--md]
                 [a_rev] [b_rev]
 
 positional arguments:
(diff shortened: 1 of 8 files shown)


daavoo (commented):
@iterative/docs this feels like the kind of change that should be associated with a release number somehow. 

I can imagine people trying to run the command from previous versions and failing.

casperdcl (approved):
I suppose it's fine to leave `--show-hash` as-is.

Also older versions of DVC will fail with e.g. `--json` as @daavoo pointed out but I don't think it's a problem:

- would require old version of DVC but reading new version of docs (affects few people)
- it's easy to run `--help` to find out supported flags

jorgeorpinel:
> this feels like the kind of change that should be associated with a release number somehow

That's up to the core team cc @efiop what's the planned version for this, perhaps a minor release increment?

jorgeorpinel (approved):
@skshetry could you remind us about this when the feature is released? Thanks

PR #2884, opened 2021-10-01
doc: PyTorch Lightning integration

### LINKED ISSUE #2882: docs: new page about dvclife integration with pytorch_lightning
I've made a [PR](https://github.com/iterative/dvclive/pull/168) with the feature. It's approved, so I write a documetation page about it.

sirily:
Fix #2882 

Documentation about the new [feature](https://github.com/iterative/dvclive/pull/168)

### DIFF
--- a/content/docs/dvclive/ml-frameworks/pytorch-lightning.md
+++ b/content/docs/dvclive/ml-frameworks/pytorch-lightning.md
@@ -62,8 +62,7 @@ trainer = Trainer(
 trainer.fit(model)
 ```
 
-> 📖 PyTorch Lightning by default creates directory with the logger's name to
-> store checkpoints, and it will create DvcLiveLogger in your workspace. If you
-> want to avoid it use checkpoint callback, change checkpoint path or disable
-> checkpointing at all as described in the
+> 📖 By default, PyTorch Lightning creates a directory to store checkpoints
+> using the logger's name (`DvcLiveLogger`). You can change the checkpoint path
+> or disable checkpointing at all as described in the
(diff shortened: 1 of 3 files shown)


content/docs/dvclive/ml-frameworks/pytorch-lightning.md
```diff
@@ -0,0 +1,69 @@
...
+    path='my_logs_path',
+    summary=False
+)
+trainer = Trainer(
+    logger=dvclive_logger,
+)
+trainer.fit(model)
+```
+
+> 📖 PyTorch Lightning by default creates directory with the logger's name to
```
daavoo, line 65: This reminded me that we should include logic for saving the model inside `DvcLiveLogger` to be consistent with the other integrations (which support the `model_file` arg).

daavoo (approved):

content/docs/dvclive/ml-frameworks/pytorch-lightning.md
daavoo: https://github.com/iterative/dvclive/issues/170

sirily: In Lightning saving is controlled by another callback and is independent of logger. I suppose, we can't include this logic and keep compatibility with native pytorch lightning checkpoint callbacks

sirily: `LightningModule` implements method `log`, which user call to store everything he wants, and then `LightningModule` calls logger. Checkpoints are saved at the end of batch/epoch via hooks like `on_train_batch_end`, etc. Logger doesn't has access to these hooks.
To save models we should write callback and encourage users to use it with our logger. But users can use native callbacks and control saving everything himself, and track it with DVC.

PR #2895, opened 2021-10-05
doc: correcting cropped image

jendefig:
The original image had the logos at the top cropped.  This image corrects that and addresses some remaining alignment issues.


casperdcl (commented):
- should be higher res? most of `/static/img/*.png` are about twice the resolution
- have to update the other image as well to match (right now the central "run train/test script" box is a different size)

PR #2908, opened 2021-10-08
guide: revisit Exp Sharing

jorgeorpinel:
Re-do of https://github.com/iterative/dvc.org/pull/2711 and https://github.com/iterative/dvc.org/pull/2719
Per https://github.com/iterative/dvc.org/pull/2654#issuecomment-936772206

- Links to sharing from other docs and other copy edits around the topic
- Introduce simple diagram for explanation about the necessary remotes
- Improve explanations (more specific and direct)
- Remove repeated or unrelated content (covered in other guides) e.g. about `dvc exp list`
- Extract short example about looping -> turned to a How to page

PENDING:

- [x] Question on long experiment by the end
  Discussion starting at https://github.com/iterative/dvc.org/pull/2908#pullrequestreview-811970391. -- REMOVED
- [ ] Double check Experiments motivation is clear in more general guide pages?
  See https://github.com/iterative/dvc.org/pull/2908#pullrequestreview-835727834


jorgeorpinel:
@shcheklein this was already approved (as a nested PR) in #2711 and #2719.

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,59 +1,59 @@
...
-```dvc
 $ dvc remote list
-storage https://remote.dvc.org/example-dvc-experiments
+storage s3://mybucket/my-dvc-store
 ```
 
-## Uploading experiments to remotes
+</details>
+
+> ⚠️ Note that only [SSH Git URLs] are compatible with DVC Experiment sharing.
```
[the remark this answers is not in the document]
iesahin: But the note says "only SSH Git URLs", not "URLs in the form of Git SSH URLs." The former phrasing also means "no file URLs or paths, only SSH Git URLs."

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,59 +1,59 @@
...
-This pushes the necessary DVC-tracked files from the cache to the default DVC
-remote (similar to `dvc push`). You can prevent this behavior by using the
-`--no-cache` option to the command above.
-
-If there's no default DVC remote, it will ask you to define one with
-`dvc remote default`. If you don't want a default remote, or if you want to use
-a different remote, you can specify one with the `--remote` (`-r`) option.
+The [default DVC remote](/doc/command-reference/remote/default) is used unless
+one is specified with the `--remote` (`-r`) option. To prevent pushing
+DVC-tracked files to remote storage altogether, use the `--no-cache` option.
```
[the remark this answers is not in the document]
iesahin: In general, _flag_ is used for _boolean_ options without any arguments. (https://unix.stackexchange.com/questions/285575/whats-the-difference-between-a-flag-an-option-and-an-argument) But there is no consensus about these terms.

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: It says only SSH URLs **are compatible**. That does not imply that only URLs are accepted by `exp push`.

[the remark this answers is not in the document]
jorgeorpinel: And in fact if you go to https://dvc.org/doc/command-reference/exp/push it only mentions Git remote names or URLs for `git_url` so if that info is incomplete (also paths are accepted) it's a more general problem not really specific to this guide only and especially not to this PR I think. It wasn't mentioned in the original PR either ([current published doc](https://dvc.org/doc/user-guide/experiment-management/sharing-experiments)).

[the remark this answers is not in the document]
jorgeorpinel: Extracted to https://github.com/iterative/dvc.org/issues/2992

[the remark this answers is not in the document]
jorgeorpinel: @iesahin `git_remote` does not accept paths. Additionally, the paragraph above reads "If your team uses Git remotes on a Git server or hosting" so here the Git remote will contain a URL, not a path.

Do you have a suggestion on how to improve the whole intro or some other section to leave room for Git remotes that are paths? TBH it seems like an edge case and more related documenting Git than DVC; It should be clear to people who have that setup. And you do use that in the example already.

And again, the current doc also doesn't explain path-based Git remotes. On the other hand we definitely want to put this warning about SSH URLs so this is an improvement.

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: > In general, flag is used for boolean options without any arguments

I'm happy to adopt this going forward (we should communicate to the team and/or add in some internal terminology doc) but if we want to retrofit existing mentions, we'd need to create a ticket. Not sure it's worth it 🙂

content/docs/user-guide/experiment-management/index.md
```diff
@@ -12,16 +12,22 @@ optimization, etc. DVC helps you codify and manage all of your
...
    from scratch like typical project changes.
 
-   At this point you may also want to consider the different
-   [ways to organize](#organization-patterns) experiments in your project (as
-   Git branches, as folders, etc.).
+   > At this point you may also want to consider the different
+   > [ways to organize](#organization-patterns) experiments in your project as
+   > well.
+
+4. Easily [share experiments] across your team.
```
shcheklein, line 28: minor: I think it's usually good not to use `easy` and similar terms

content/docs/user-guide/experiment-management/index.md
```diff
@@ -35,6 +41,8 @@ meaningful measures for the experimental results.
 
 ## Experiments
 
+> Note: these features require a Git-enabled DVC <abbr>repository</abbr>.
```
shcheklein, line 44: do we actually need this note? does it apply to the whole experiment management?

Git-enabled DVC repo - still too complicated to my mind.

I would move the note one paragraph below (do not start with some notes that are not even clear if they are relevant for me) and would make it other way around:

`dvc exp` commands won't work if you don't use Git to version control your code

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,59 +1,59 @@
 # Sharing Experiments
 
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+If your team uses [Git remotes] on a Git server or hosting (e.g. GitHub, GitLab,
```
shcheklein, line 3: what are the chances that teams do not use Git remotes? considering that we already mentioned that `dvc exp` won't work w/o Git?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,59 +1,59 @@
...
+
+[ssh git urls]:
+  https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_protocols
 
-You can upload an experiment and its files to both remotes using `dvc exp push`
-(requires the Git remote name and experiment name as arguments).
+## Uploading experiments
+
+You can upload an experiment with all of its files and data using `dvc exp push`
+(requires a Git remote name and experiment name as arguments).
```
shcheklein, line 46: > (requires a Git remote name and experiment name as arguments).

why do we need to clarify it here this way?

shcheklein:
Branch has failed to deploy also, could we merge master fix it @rogermparent ?

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
iesahin: `git_remote` is a Git remote name, not a URL or path. You can define a `git_remote` in the form: 

```
git remote add myremote ~/myrepobackups/myrepo.git 
```

and `myremote` will work with `dvc exp push`. 

When you say _only_, it reads like the above form doesn't work.

rogermparent:
I think a master merge could fix this deploy, since it doesn't have a Gatsby 2 cache populated.

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -63,13 +57,11 @@ performance also depend on the connection bandwidth and remote configurations.
 
 [run-cache]: /doc/user-guide/project-structure/internal-files#run-cache
 
-## Listing experiments remotely
+## Listing experiments on remotes
 
-In order to list experiments in a DVC project, you can use the `dvc exp list`
-command. With no command line options, it lists the experiments in the current
-project.
```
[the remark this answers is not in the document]
jorgeorpinel: Done in 32ccd25b.

content/docs/user-guide/experiment-management/comparing-experiments.md
```diff
@@ -42,10 +42,10 @@ refs/tags/baseline-experiment:
         cnn-64
 ```
 
-This command lists the experiments originated from `HEAD`. You can add any other
-options to the remote command, including `--all`. (see previous section).
+This command lists remote experiments originated from `HEAD`. You can add any
+other options to the remote command, including `--all` (see previous section).
```
jorgeorpinel, line 46: > From https://github.com/iterative/dvc.org/pull/2862#pullrequestreview-798142029

How is `HEAD` is determined for a remote to list experiments from (e.g. `dvc exp list origin`)? Is it the local repo's HEAD? Is it the default branch HEAD always? We should probably clarify. Cc @iesahin do you know?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -63,13 +57,11 @@ performance also depend on the connection bandwidth and remote configurations.
 
 [run-cache]: /doc/user-guide/project-structure/internal-files#run-cache
 
-## Listing experiments remotely
-
-In order to list experiments in a DVC project, you can use the `dvc exp list`
-command. With no command line options, it lists the experiments in the current
-project.
+## Listing experiments on remotes
```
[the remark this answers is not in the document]
jorgeorpinel: > > listing saved experiments
>
> Saved where? It's not clear we refer to the remote ones

I guess they don't have to be "remote" necessarily, could just be in an external repo in the local machine. Seems like an edge case we don't need to customize language for though?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -130,31 +111,28 @@ can set the number with `--jobs` (`-j`).
 If an experiment being pulled already exists in the local project, DVC won't
 overwrite it unless you supply `--force`.
 
-### Example: Pushing or pulling multiple experiments
+## Example: Sharing multiple experiments
```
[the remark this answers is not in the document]
jorgeorpinel: What do you guys think about 👨‍🏫 for guide example headers @shcheklein @iesahin ? I could add a task to apply throughout Exp Mgmt guides (at least) in https://github.com/iterative/dvc.org/issues/2911.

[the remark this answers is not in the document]
shcheklein: No strong opinion on this.

Re the structure - I would prefer to have example embedded/close to the text that is relevant for them. Unless we are writing some mini-tutorials we'd like to link to.

content/docs/user-guide/experiment-management/index.md
jorgeorpinel: Changed to "seamlessly" for now. We kind of need a qualifier here I think.

content/docs/user-guide/experiment-management/index.md
jorgeorpinel: Moved down and rephrased for now. I don't know if we need the note. Most realistic uses of DVC need Git I suppose so maybe we should only mention that Git is NOT required in the few that dont? I.e. basic data management with `init --no-scm`.

> use Git to version control your code

This may also be misleading though? It's not just code but config files (importantly dvc metafiles).

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: We mentioned that in a different page though, I think it's worth reiterating a little here for context. I rephrased it a little for clarity, PTAL.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: It's just explaining the command usage. It takes those 2 arguments. I rephrased this to remove the parenthesis.

content/docs/user-guide/experiment-management/index.md
```diff
@@ -88,7 +96,7 @@ These are the main alternatives:
 - **Git tags and branches** - use the repo's "time dimension" to distribute your
   experiments. This makes the most sense for experiments that build on each
   other. Helpful if the Git [revisions](https://git-scm.com/docs/revisions) can
-  be easily visualized, for example with tools
+  be visualized easily, for example with tools
```
shcheklein, line 99: minor, style: I would remove "easily"

content/docs/user-guide/experiment-management/index.md
```diff
@@ -52,6 +58,8 @@ roll back to the baseline. The basic workflow goes like this:
 - Make the selected experiment persistent by committing its results to Git. This
   cleans the slate so you can repeat the process.
 
+> Note that these features won't work if you don't use Git to version your code.
```
shcheklein, line 61: move it before the list? (feels a bit too late at this stage)

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,83 @@
 # Sharing Experiments
 
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+If your team uses Git server or hosting (e.g. GitHub, GitLab, etc.) to
+collaborate on projects, you can also use them to save and share DVC
+Experiments.
```
shcheklein, line 5: Experiments -> experiments?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,83 @@
...
 
-[DVC remotes](/doc/command-reference/remote) on the other hand are
-storage-specific locations (e.g. Amazon S3 or Google Drive) which we can
-configure with `dvc remote`. DVC uses them to store and fetch large files that
-don't normally fit inside Git repos.
+> You will need both [Git remotes] and DVC
```
shcheklein, line 7: I still don't see a strong reason for these notes to be honest ... we mention GH already before ... what are the chances people won't have Git remote?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,83 @@
...
 
-[git remotes guide]:
-  https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes
+Sharing experiments is similar to [sharing regular project data] and artifacts
+by synchronizing from remotes. DVC-tracked data, models, etc. are in your
+project's <abbr>cache</abbr> and thus will be transferred to/from
+[remote storage](/doc/command-reference/remote) (e.g. Amazon S3 or Google
+Drive). Small files like experimental code and
+[DVC metafiles](/doc/user-guide/project-structure) files are uploaded or
+downloaded with Git automatically as needed.
```
shcheklein, line 22: probably not automatically, `dvc exp pull` is needed, right?

again, ideally, we want to be explicit that it's not the regular way you would use Git - it's based on custom refs, it doesn't affect / pollute history, etc

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
shcheklein: yep, agreed with Emre on this from a brief look. Local path is also URI. Btw, what are other options besides SSH? 

HTTP is not supported (no writable), what else?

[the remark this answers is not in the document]
shcheklein: may be it's easier to say that HTTP URLs are not supported?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -130,31 +92,28 @@ can set the number with `--jobs` (`-j`).
 If an experiment being pulled already exists in the local project, DVC won't
 overwrite it unless you supply `--force`.
 
-### Example: Pushing or pulling multiple experiments
+## 👨‍🏫 Example: Sharing multiple experiments
```
shcheklein, line 95: let's do emojis as a separate PR?

do we have links to these examples in the text? 

should we after all embed examples into the text (if you read the book do you expect examples at the end?)

content/docs/user-guide/experiment-management/index.md
jorgeorpinel: We're not selling DVC here, it's a conditional: if Git revisions can be visualized easily. Easily is key. They can always be visualized (e.g. with `git log` if you know how to use it well), but if it's hard then this doesn't apply.

jorgeorpinel: Changed completely to

> Git-based experiment structures are especially helpful along with Git history exploration tools like in GitHub.

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: > have example embedded/close to the text that is relevant for them

Good idea and that's how this page is structured, but will keep it in mind when considering which examples to move from the command reference.

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: > Local path is also URI

@iesahin @shcheklein This is a larger issue then, since we published it also in https://dvc.org/doc/user-guide/troubleshooting#git-auth and don't mention the edge case of path-based Git remote definitions. Should we address it in this PR?

> say that HTTP URLs are not supported

HTTP URLs are supported, even for pushing, if the remote accepts unauthenticated writes. It's more about the authentication. Only SSH auth is supported. Local paths never require auth AFAIK.

content/docs/user-guide/experiment-management/index.md
jorgeorpinel: OK sure, moved.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: It's "DVC Experiments" (branded). Unsure whether this will stay in the long run but it's how we're differentiating regular exps from our implementation for now (we've discussed this I believe).

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: Yeah, I removed this but just incorporated the link to Git remotes into the details section below.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: > probably not automatically... be explicit that it's not the regular way you would use Git

Changed "automatically as needed" for "by DVC" in this details section. The other details are explained further down.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: Hmmm. OK removing emojis for now but we need an issue then to address this in all guides. I wonder if it's really that important

There are no manual links to these examples anywhere but the ToC nav shows them.

And yes, there are short examples or what I call "code samples" embedded in the text as needed.

These examples by the end are interesting applications that don't map exactly to specific paragraphs from the previous sections. In fact the last example is the longest section in the whole doc.

content/docs/user-guide/experiment-management/sharing-experiments.md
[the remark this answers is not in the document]
jorgeorpinel: OK what about "Note that DVC can only authenticate with Git remotes via [SSH URLs]." ? Using that for now. WDYT @iesahin @shcheklein ?

content/docs/user-guide/troubleshooting.md
```diff
@@ -92,16 +92,17 @@ using:
...
-may need to authenticate for _write_ (`dvc exp push`) or _read_ (`dvc exp list`,
+commands accept a `git_remote` argument. You may need to authenticate to use the
+Git remote, for _write_ (`dvc exp push`) or _read_ (`dvc exp list`,
 `dvc exp pull`) permissions.
 
 DVC does not currently support authentication with [Git credentials]. This means
 that unless the Git server allows unauthenticated HTTP write/read, you should
-use an [SSH Git URL] when listing, pulling or pushing experiments.
+use an [SSH Git URL] for Git remotes used for listing, pulling or pushing
+experiments.
```
jorgeorpinel, line 105: Hello @daavoo can you confirm this change us OK? Thanks

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: Rephrased to "DVC can rely on existing Git servers or hosting (e.g. GitHub, GitLab, etc.) to
save and share DVC Experiments." PTAL

content/docs/user-guide/experiment-management/comparing-experiments.md
jorgeorpinel: Ping @iesahin

jorgeorpinel: Or, I guess @pspeter can also confirm. Thanks

content/docs/user-guide/troubleshooting.md
daavoo: I think that it goes back to the original discussion 😅 . The title feels a little misleading for me, because this issue only applies to `experiments` and the new phrase seems to imply that it applies to DVC in general (maybe `DVC experiments` instead of `DVC`?).

The body changes lgtm

jorgeorpinel: Agree but the title is already a bit long and users mainly get here from an error printed when using `dvc exp` so shouldn't be a problem really 🙂

content/docs/user-guide/experiment-management/comparing-experiments.md
iesahin: I thought I've replied this, sorry. 

It's the `HEAD` of the remote repository. So when `dvc exp list origin` it lists the `HEAD` of the `origin` even if we are on another branch.

content/docs/user-guide/experiment-management/comparing-experiments.md
```diff
@@ -41,8 +41,9 @@ refs/tags/baseline-experiment:
         cnn-64
 ```
 
-This command lists remote experiments originated from `HEAD`. You can add any
-other options to the remote command, including `--all` (see previous section).
+This command lists remote experiments based on `HEAD`. You can use `--all` to
+list all experiments, or add any other supported option to the remote
+`dvc exp list` command.
```
jorgeorpinel, line 46: Sorry, slightly unrelated copy edit...

content/docs/user-guide/experiment-management/comparing-experiments.md
jorgeorpinel: I see... OK, updated.

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,78 @@
...
-```dvc
 $ dvc remote list
-storage https://remote.dvc.org/example-dvc-experiments
+storage s3://mybucket/my-dvc-store
 ```
 
-## Uploading experiments to remotes
+</details>
+
+> ⚠️ Note that DVC can only authenticate with Git remotes using [SSH URLs].
```
shcheklein, line 35: remove square brackets?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,78 @@
...
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+DVC can rely on existing Git servers or hosting (e.g. GitHub, GitLab, etc.) to
+save and share DVC Experiments.
 
-[DVC remotes](/doc/command-reference/remote) on the other hand are
-storage-specific locations (e.g. Amazon S3 or Google Drive) which we can
-configure with `dvc remote`. DVC uses them to store and fetch large files that
-don't normally fit inside Git repos.
+<details>
```
shcheklein, line 6: hmm .. I don't think we should be using expandable sections here. That's User Guide - it's purpose is to be more comprehensive, right?

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -130,31 +87,33 @@ can set the number with `--jobs` (`-j`).
...
-> pushed/pulled.
+> 📖 See also [Listing experiments].
 
-## Example: Creating a directory for an experiment
+[listing experiments]:
+  /doc/user-guide/experiment-management/comparing-experiments#list-experiments-in-the-project
 
-A good way to isolate experiments is to create a separate home directory for
-each one.
+## Example: Dedicated experiment directories
```
shcheklein, line 106: I don't quite understand where does this example come from, what is the motivation ... it also feels very unrelated to the section itself

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -130,31 +87,33 @@ can set the number with `--jobs` (`-j`).
 If an experiment being pulled already exists in the local project, DVC won't
 overwrite it unless you supply `--force`.
 
-### Example: Pushing or pulling multiple experiments
+## Example: Sharing multiple experiments
 
-You can create a loop to upload or download all experiments like this:
+You can create a loop to push or pull all experiments. For example in a Linux
```
shcheklein, line 92: This is not example - this is How to. Not clear again what is the motivation for this "example"

shcheklein (changes requested):
I would say section should be focusing more on explaining (in simple terms, with diagrams) how it works + all basic related coomands - push/pull/list.

Now it lack that explanation (current one is hidden and too complicated still). 

Examples in the section are not actual examples, but some mini "How tos" w/o clear motivation.

cc @iesahin

iesahin:
Without examples, I think UG would be a rehash of the command reference.

shcheklein:
> Without examples, I think UG would be a rehash of the command reference.

Yes. Current "examples" are not examples though. They are separate advanced mini-guides / how-tos. Some of them very specific and it's not clear even why do we prioritize them not some other how-tos.

My take on this - actual example should be embedded right into the text (no separate sections). If we want to include some advanced `How tos` - let's rename them, prioritize and include. It should be clear though _why_ those. There should be some logic in the text that connects them to the main text.

content/docs/user-guide/experiment-management/sharing-experiments.md
iesahin: Suppose you have thousands of experiments, would you push/pull/share them one by one?

content/docs/user-guide/experiment-management/sharing-experiments.md
iesahin: For example, the user has experimented for some time and found a good result. This is to isolate that experiment in a single directory and share with folks that don't use Git or DVC.

iesahin:
The text is already too long and it's true that the context is not always clear. However splitting the examples are not worth it IMO.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: Hmmm we still have some expanding sections for implementation details. But yeah I expanded this and it looks good.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: I don't have a strong opinion on this example. I think it's clever but also agree it's not clear whether someone would ever do this. Keeping it for now as it's at the bottom but I'm happy to remove it if you guys decide ( we already have other examples both here and in https://dvc.org/doc/command-reference/exp/push#examples).

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: Agree it's more of a how to... Moved to a new how-to page for now. PTAL

I also agree it's a clear need @iesahin - maybe we should open a feature request in the core repo though? Maybe `exp push/pull --head` or/and `<branch>` or/and `--all`

jorgeorpinel:
> Without examples, I think UG would be a rehash of the command reference.

User guide examples can be embedded in the story though, not necessarily in separate **Example: XYZ** sections. Those look more like cmd ref-y to me (maybe it's bias).

jorgeorpinel:
> should be focusing more on explaining (in simple terms, with diagrams) how it works

I reworked the first section and included a very basic diagram. Do you think this is closer to what the guide should explain @shcheklein ? Thanks

shcheklein:
> I reworked the first section and included a very basic diagram. Do you think this is closer to what the guide should explain @shcheklein ? Thanks

It's better. I still don't understand the logic behind the "example", lack of simple actual **examples** (if we want to include them?), explaining `--remote` and `--j` doesn't make much sense to me here, at least not in command reference style.

To my mind if the page is done right there should be one section around diagram that explain the logic behind this explicitly (that we don's use commits). We should write about `dvc exp list`.

jorgeorpinel:
OK @shcheklein but can we follow up on the examples in a separate PR? Or do you also think the main section is not good enough yet? I.e. do you mean a much more formal diagram closer to something like https://dvc.org/doc/use-cases/fast-data-caching-hub#example-shared-development-server ? Thanks

jorgeorpinel:
p.s. note that I removed unrelated changes (extracted into https://github.com/iterative/dvc.org/pull/3080).

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,90 @@
 # Sharing Experiments
 
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+Saving and sharing experiments is similar to [sharing regular project versions],
+done by synchronizing with DVC and Git remotes. DVC takes care of pushing and
+pulling to/from Git in the case of experiments, however.
```
shcheklein, line 5: Git -> Git remote

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,121 +1,90 @@
 # Sharing Experiments
 
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+Saving and sharing experiments is similar to [sharing regular project versions],
```
shcheklein, line 3: This is better! This is still suboptimal intro though. The way I would read it - we are pushing commits (sorry, if we are going in circles). How can we explain this to people?

```
Aw we learned before, DVC experiments are utilizing Git internals to capture their metadata. As we mentioned, it's similar to Git commits but w/o actually polluting your Git repository history and everything is automated. This is great for one person iterating on the project, but what if we want to share this? If they are not Git commits, how do we make experiments history available?
```

can something like this help?

shcheklein:
@jorgeorpinel looks better, how about this

`explaining --remote and --j doesn't make much sense to me here, at least not in command reference style.

To my mind if the page is done right there should be one section around diagram that explain the logic behind this explicitly (that we don's use commits). We should write about dvc exp list.` - not sure those were addressed.

examples- up to you, I just don't see any meaning in having them here in this way to be honest. I'm not opposed to creating a separate subsection "Recipes" or something.

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: I see what you mean ⌛

jorgeorpinel: OK @shcheklein . PTAL at the latest intro here (introduced in 675aacec).

> I didn't use the exact same ideas you suggested as to avoid reiterating on basic explanations of `dvc exp` or going into implementation details (which exist in the overview page)

jorgeorpinel:
> explaining --remote and --j doesn't make much sense to me here

It's a good question. I think that @iesahin intended to cover all the possible details in the guide so that we don't have them in the **Description** of the cmd refs. We may have even agreed to do that before.

But maybe it's OK to leave out less relevant option explanations (will be covered under the **Options** section of each cmd ref. This improves the docs focus and maintainability -- it's unclear how to define relevant though.

I'm removing them from here and the cmd ref descs. for now. WDYT?

> if the page is done right there should be one section around diagram...

The intro currently does that. I'm not trying to completely rewrite the guide. The existing structure also makes sense I think.

> We should write about `dvc exp list`

It's already mentioned here, but mainly covered in [Comparing Exps](https://dvc.org/doc/user-guide/experiment-management/comparing-experiments). We have [an issue](https://github.com/iterative/dvc.org/issues/2995) to rewrite that guide though, which structure may be too linear/ disconnected.

> examples- up to you... separate subsection "Recipes" or something.

TBH I also just left it because it's already there but didn't find it super relevant either (clever though). I'll remove it for now. And we do have a recipes section, the How To! But there should be a clear need for each step-by-step solution there.

shcheklein:
> I think that @iesahin intended to cover all the possible details in the guide so that we don't have them in the Description of the cmd refs. 

To my mind details of the options and their behavior for a specific command should be explained in the cmd ref for that command. UG should be for long sections (remotes) or cross command explanations (experiments overview). Or lengthy explanation how things work. 

And that doesn't matter, since `-j` is still not an essential option, I don't see anything on top of cmd ref options description.

content/docs/user-guide/experiment-management/sharing-experiments.md
```diff
@@ -1,199 +1,96 @@
 # Sharing Experiments
 
-There are two types of remotes that can store experiments. Git remotes are
-distributed copies of the Git repository, for example on GitHub or GitLab.
+Regular <abbr>DVC repository</abbr> versions are typically synchronized among
+team members per regular Git workflow. [DVC Experiments] are directly woven into
+this commit history. But they also intend to avoid cluttering shared repos, so
```
shcheklein, line 5: Sorry, I'm not sure I understand this

`DVC Experiments] are directly woven into this commit history.`

Let's try to write is simple and very explicit.

regular Git workflow, namely creating Git commits, pushing them to Github or from Github. 

It means that experiments become part of the Git history, they are visible in Git, etc. As we mentioned this is not idea since doesn't scale well, creates mess in Git history etc. DVC experiments still rely on Git and Github, but they are not directly attached ... so won't visible ... etc ..

shcheklein:
> . The existing structure also makes sense I think.

Yes, but it's not ideal. Those small subsections could be removed completely - just one paragraph after the diagram is enough . Hey, you have push/pull/list to share experiments this way. Here are the links to read the details (links to the cmd ref).

Thus the whole purpose of this page is the diagram and text around it - that is the most valuable part. We can write more and more details, caveats that are general for that workflow, write advanced stuff how to see git index, etc.

Also, diagram might include `exp pull`, `exp push`?

content/docs/user-guide/experiment-management/sharing-experiments.md
jorgeorpinel: > Let's try to write is simple and very explicit.

Simplified in 0fb52cb3.

jorgeorpinel: I think the rest of your suggestions are great but do we want to elaborate so much on general motivation for DVC Experiments in this guide @shcheklein ? Maybe we should separately review that this is clearly included in the Overview page and/or the Running Exps page.

For now the only mention of motivation is "to avoid cluttering everyone's copies of the repo, by default experiments will only exist in the local environment where they were created."

jorgeorpinel:
> > The existing structure also makes sense I think.
>
> Yes, but it's not ideal. Those small subsections could be removed completely

I agree with you @shcheklein. I argued for that from the original PR but we decided to merge something. And this PR is about revisiting that guide so it makes sense to bring it up again. Buuuttt that's actually a task in https://github.com/iterative/dvc.org/issues/2911. I actually intended to only improve the guide with it's current structure here. If you don't think this is valuable enough then I can plan to repurpose the PR, but I think merging the work so far would be a good first step.

> diagram might include exp pull, exp push?

Added pull in d6de4f57.

iesahin:
I tend to diverge from you about the reading habits of users, so I'm including more headers the readers could skim. Unifying those sections might produce a more beautiful document but the users will spend a longer time finding the information they need.

shcheklein:
I think we diverge here primarily on the content and the focus in the document, not reading habits. For me removing subsections make sense since I don't see much content in them that should stay - we don't it in the user guide in this form (e.g. `-j` - why only this option? what is the benefit to repeat it this way, etc).

shcheklein (approved):

jorgeorpinel:
TBH in my answer I was thinking more of https://dvc.org/doc/user-guide/experiment-management/comparing-experiments, not the one (sharing exps), sorry for the confusion.

As for reading habits, FWIW personally I usually try to Ctrl+F key words first. Reading the ToC or headers is a probably next strategy but I expect that kind of structure more in reference-type docs. For a guide I'd assume that to really get it you need to read the whole thing or at least a good chunk at the beginning before skipping ahead.

jorgeorpinel:
p.s. I have to admit I'm still not sure when we want guides to be more technical vs. refs.

PR #2921, opened 2021-10-13
oct heartbeat draft

jendefig:
Publish date: Oct 15th

This is ready for review from a content perspective.  Still adding some things (Amsterdam) and polishing.

PR #2931, opened 2021-10-15
re-adding October Heartbeat

jendefig:
Re-adding October Heartbeat here.
Apologies for the confusion.

@casperdcl - I have addressed all your comments from https://github.com/iterative/dvc.org/pull/2921 with the exception of the image of the text.  I'm not sure I agree.  But I can't get yarn develop to work on my computer (it keeps giving me some avatar error 🤷‍♀️ ) so I have to look at it here in heroku.  I also changed the manhaes pic to another from the article, but unsure if it's better.  Finally, I added the code (I think) from the CML with DVC doc.  (this was what originally messed everything up because I didn't put it inside the ````python` markdown. 😬 🤦‍♀️ Thanks for the feedback, I definitely would never have caught all the Oxford comma faults. 

I still need to run images through tiny.png.  And look at again myself where I can read it in a normal way, instead of the code way, which just doesn't work for my ability to edit.

### DIFF
--- a/content/blog/2021-10-15-october-21-heartbeat.md
+++ b/content/blog/2021-10-15-october-21-heartbeat.md
@@ -7,6 +7,7 @@ description: >
   from the team, new learning opportunities, and more!
 descriptionLong: |
   This month you will find:
+    
     🗺 MLOps workflows,
 
     🤔 Lots of ways to learn,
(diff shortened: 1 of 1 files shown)


jendefig:
I think this is done now.

I concur with @casperdcl on Apsilon quotes in quote block instead of image.

I added the picture block things for the meetups.

I need to know if you are good with me putting the code block in.  To me the docs section looked slimmer this month so if gave it some useful visual heft.

jendefig:
Ok.  This last one should do it.

PR #2932, opened 2021-10-18
Mention libvips/sharp gotcha in setup instructions

duijf:
In order to document some stuff about Studio, I wanted to set up `dvc.org` for local development. I ran into this gotcha, and wanted to mention it in the setup instructions to save others some time.

Gatsby has a dependency on the native `libvips` library via the `sharp` NPM package. Sharp by default uses the system `libvips` if it's available, but that can lead to compatibility problems if there is a version mismatch (or build options are different).

On my system I have `libvips` installed for something else. This led to the following error when running `yarn develop`:

```
ERROR #11321  PLUGIN

"gatsby-plugin-manifest" threw an error while running the onPostBootstrap lifecycle:

Input file contains unsupported image format

[ long stacktrace ]
```

The fix is to make `sharp` ignore the system `libvips`. This needs to be done at dependency install time, as documented in this Gatsby issue: https://github.com/gatsbyjs/gatsby/issues/20698

After removing `node_modules` and running `yarn` with this env var, the problem went away.

### DIFF
--- a/.github/workflows/link-check-deploy.yml
+++ b/.github/workflows/link-check-deploy.yml
@@ -15,8 +15,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - run: git merge-base origin/master HEAD
-      - run: git merge-base origin/master...HEAD
       - id: build_check
         uses: LouisBrunner/checks-action@v1.0.0
         with:


rogermparent:
Great catch! I've run into this before as well, it's related to us using old Gatsby packages that have been left behind by more recent versions of libvips. I'd suggest a minor change to the text that makes it read better to me:

> ⚠️ Do you get an error from `gatsby-plugin-manifest` about unsupported image
> formats? You may be running into
> [this issue](https://github.com/gatsbyjs/gatsby/issues/20698). You can fix it
> by reinstalling `node_modules` with a flag that forces `sharp` to ignore your
> globally installed `libvips`:

Someone on the docs team might have a different/better suggestion. Other than that, thanks for the contribution!

iesahin:
I think it might be better to have a _Troubleshooting_ section header and put this in a bullet list. WDYT @jorgeorpinel ?

jorgeorpinel:
> I've run into this before as well, it's related to us using old Gatsby packages that have been left behind by more recent versions of libvips

So maybe we should make an issue to update Gatsby paks instead @rogermparent ? It's great to be aware of this but documenting it means we don't have a solution.

PR #2933, opened 2021-10-18
Document Studio error: `Failed to push experiment`

duijf:
Fixes https://github.com/iterative/viewer/issues/2454

This PR documents currently known causes of failing to push experiments to the remote repository. It also points users to our contact information in case they are running into something else.

In addition, fix some pre-commit lints for this page in the first commit. The lints were attempting to reformat a section. I also added a couple of word break opportunity tags so the content flows more nicely in case of certain line wrapping scenario's.


duijf:
Derp, thanks @tapadipti for catching those grammar problems. Fixed now :smile:

PR #2954, opened 2021-10-21
Fix and update link-check

rogermparent:
> Context: https://github.com/iterative/dvc.org/pull/2932

This PR fixes the odd failures that are happening on PRs by using `fetch-depth: 0` on the checkout action, allowing full access to all branches' git history and allowing so we can do a proper github-like diff that doesn't include `origin/master` changes that have happened since the PR in question.

Separately, this PR upgrades `link-check` to a new version that uses a small change that uses a single command of `git diff origin/master...HEAD` instead of a two-step `merge-base` and `diff`. Functionally the same thing (probably didn't need a minor bump, just a patch, but oh well we're still in `0.x.x`), just more simple and elegant- the only reason we didn't do it before is I wasn't aware of triple-dot syntax. I tested this new version a couple times to make sure it doesn't just break everything, and it works fine.

One big caveat is that merging master (or any Gatsby 3 commit) into our current Gatsby 2 PRs that currently have this issue will cause caching havoc that needs manual intervention to be remedied. I can do so for PRs that _really_ need it, but I'd suggest using local link-check instead since everyone probably has full commit history locally and it should work. At least we'll have it solved for the future!

### DIFF
--- a/.github/workflows/link-check-deploy.yml
+++ b/.github/workflows/link-check-deploy.yml
@@ -23,7 +23,7 @@ jobs:
           status: queued
       - name: Run Link Check
         id: check
-        uses: iterative/link-check.action@v0.8
+        uses: iterative/link-check.action@v0.9
         with:
           diff: true
           configFile: config/link-check/config.yml


jorgeorpinel:
Thanks @rogermparent ! Looks good i.e. I approve but

> merging master (or any Gatsby 3 commit) into our current Gatsby 2 PRs that currently have this issue will cause caching havoc that needs manual intervention to be remedied

That sounds scary haha. No matter when we merge this there will be several PRs open that may require merging `master` to solve conflicts. Are you suggesting we just ignore the CI link-checker for those?

PR #2966, opened 2021-10-25
ref: `exp show --only-changed`

daavoo:
docs updates for https://github.com/iterative/dvc/pull/6867

Not sure if this will actually get merged before #2862 but seems relevant to eventually include the option in the guide.

### DIFF
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -175,7 +175,8 @@ $ dvc exp show --include-params=featurize
 └───────────────────────┴──────────────┴─────────┴────────────────────────┴──────────────────┘
 ```
 
-You can also filter out the metrics and parameters that didn't change:
+You can also filter out the metrics and parameters that are the same
+across the shown experiments:
 
 ```dvc
 $ dvc exp show --only-changed
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -175,8 +175,8 @@ $ dvc exp show --include-params=featurize
 └───────────────────────┴──────────────┴─────────┴────────────────────────┴──────────────────┘
 ```
 
-You can also filter out the metrics and parameters that are the same
-across the shown experiments:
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/exp/show.md
```diff
@@ -73,6 +74,13 @@ metric or param.
 
 - `--param-deps` - include only parameters that are stage dependencies.
 
+- `--only-changed` - Only show metrics/params with values varying across the
+  selected experiments. When used along with
+  `--include-params`/`--include-metrics`, `--only-changed` prevails over the
+  specific values passed to `--include`. For example, given
+  `--only-changed --include-params=foo`, if `foo` doesn't vary across the
+  selected experiments, it won't be shown in the final table.
```
jorgeorpinel, line 82:
```suggestion
- `--only-changed` - show only parameters and metrics with values that vary
  across experiments. Note that this option takes precedence over
  `--include-params` and `--include-metrics`, for example given
  `--include-params=foo --only-changed`, param `foo` would still be hidden
  if its value is the same in all experiments.
```

content/docs/command-reference/exp/show.md
```diff
@@ -171,6 +179,22 @@ $ dvc exp show --include-params=featurize
 └───────────────────────┴──────────────┴─────────┴────────────────────────┴──────────────────┘
 ```
 
+You can also filter out the metrics and parameters that are the same across the
+shown experiments:
```
jorgeorpinel, line 183:
```suggestion
You can also filter out any metrics and parameters that do not change across the
shown experiments:
```

jorgeorpinel (approved):
Couple copy edits ☝️

jorgeorpinel:
> seems relevant to eventually include the option in the guide

Added to https://github.com/iterative/dvc.org/issues/2995

jorgeorpinel:
Oops I merged before it's even merged in the core repo... It's minor so I think we can let it be (doubt users will find it in docs before it's released)

PR #1360, opened 2020-05-26
metrics: 1.0 update

dmpetrov:
DVC 1.0 metrics

PR #2988, opened 2021-10-29
Updated exp diff tables

daavoo:
P.R. for https://github.com/iterative/dvc/pull/6888

### DIFF
--- a/content/docs/command-reference/exp/diff.md
+++ b/content/docs/command-reference/exp/diff.md
@@ -7,7 +7,7 @@ Show changes in [metrics](/doc/command-reference/metrics) and
 
 ```usage
 usage: dvc exp diff [-h] [-q | -v] [--all] [--param-deps]
-                    [--json] [--md] [--old]
+                    [--json] [--md]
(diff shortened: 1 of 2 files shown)


dberenbaum (changes requested):
In the synopsis section of the cmd ref, `--old` is still shown as an option.

Also, where are the related core/docs PRs for `metrics diff --old`?

daavoo:
> Also, where are the related core/docs PRs for `metrics diff --old`?

There is not such an option for `dvc metrics diff` (And, afaik, it has never existed). It looks like the docs are not in sync with the command

dberenbaum (approved):

jorgeorpinel:
> There is not such an option

It must have existed when @iesahin wrote this. Anyway, thanks for the update!

iesahin:
FYI `--old` joined to `dvc metrics diff` options in https://github.com/iterative/dvc.org/pull/1360/files

@daavoo @jorgeorpinel

jorgeorpinel:
Now, that's --old.

Issue #2989, opened 2021-10-29
ref: consistent cmd alias strategy


jorgeorpinel:
**Current CLI aliases**
> Bugs marked with check boxes 🐞 

- [version](https://dvc.org/doc/command-reference/version)-> [doctor](https://dvc.org/doc/command-reference/doctor): the former **doesn't** mention the latter; The latter has it's own short page linking to the former (not great).
  > - [x] There are no links to either one from any other refs or guides. There should be some at least.

- [list](https://dvc.org/doc/command-reference/list) -> ls: the former mentions the `ls` alias; ~~The latter [doesn't exist](https://dvc.org/doc/command-reference/ls).~~
  > - [ ] Other docs link to `list` only ATM. Update?
  > - [x] `ls` not existing seems like a bug ATM (quite inconsisntent)

- [experiments](https://dvc.org/doc/command-reference/experments) -> [exp](https://dvc.org/doc/command-reference/exp): ~~The former doesn't exist;~~ Only the latter (the alias) is on the cmd ref nav.
  > - [x] `experments` not existing is a bug -- as it used to be the path, it should be redirecting to `exp`.
  > - [x] Also, the now `exp` page says it's "aliased to `exp`" which no longer makes sense.

**Solution**

Let's pick which name is considered most important, whether original or alias (I vote for `version`, `ls`, `exp`) and only have that on the nav (and terminal block highlighter?). WDYT @dberenbaum ?

Questions:

- Do we even mention the alternative name in docs?
- [ ] We should also review what the `-h` help output is doing for these, probably (Core DVC question)

**Also...**

We sometimes call the special `remote://` remote URL format a "remote alias". Should probably use this term consistently or remove it altogether.

UPDATE: Some decisions in https://github.com/iterative/dvc.org/issues/2989#issuecomment-966546066 below

jorgeorpinel:
Related product question @dberenbaum @efiop : should `exp list` and other listing commands have `ls` aliases too? Thanks

dberenbaum:
TBH I'm not sure what the motivation was for each of these aliases (except for `exp` which is obviously just shorter). @efiop any reason why we developed these aliases?

I would probably lea

> Let's pick which name is considered most important, whether original or alias (I vote for `version`, `ls`, `exp`) and only have that on the nav (and terminal block highlighter?). WDYT @dberenbaum ?

Sounds good.

>     * Do we even mention the alternative name in docs?

I don't think we need to.

[...]

PR #2990, opened 2021-10-29
ref: fix aliasing bugs

jorgeorpinel:
per https://github.com/iterative/dvc.org/issues/2989

  > - [x] There are no links to either one (`version/doctor`) from any other docs.
  > - [ ] `ls` not existing seems like a bug ATM (quite inconsisntent)
  > - [x] `experments` not existing is a bug -- as it used to be the path, it should be redirecting to `exp`.
  > - [x] Also, the now `exp` page says it's "aliased to `exp`" which no longer makes sense.

Also
- [x] Missing `ls` and `experiments` from the terminal blocks highlighter

Finally
- [ ] Update https://github.com/iterative/dvc.org/issues/2989 description after merge

### DIFF
--- a/content/linked-terms.js
+++ b/content/linked-terms.js
@@ -22,5 +22,9 @@ module.exports = [
   {
     matches: 'dvc experiments',
     url: '/doc/command-reference/exp'
+  },
+  {
+    matches: 'dvc ls',
+    url: '/doc/command-reference/list'
   }
 ]
--- a/config/prismjs/dvc-commands.js
+++ b/config/prismjs/dvc-commands.js
@@ -34,6 +34,7 @@ module.exports = [
   'metrics',
   'params diff',
   'params',
+  'ls',
   'list',
   'install',
   'init',
--- a/config/prismjs/dvc-commands.js
+++ b/config/prismjs/dvc-commands.js
@@ -56,6 +56,7 @@ module.exports = [
   'exp branch',
   'exp apply',
   'exp',
+  'experiments',
   'doctor',
   'diff',
   'destroy',
--- a/content/docs/command-reference/exp/index.md
+++ b/content/docs/command-reference/exp/index.md
@@ -1,6 +1,6 @@
 # exp
 
-_New in DVC 2.0_
+_New in DVC 2.0 (see `dvc version`)_
 
 A set of commands to generate and manage <abbr>experiments</abbr>:
 [run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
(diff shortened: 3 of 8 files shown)


jorgeorpinel:
> `ls` not existing seems like a bug
(https://dvc.org/doc/command-reference/ls)

@rogermparent sorry: where do auto-links come from again? `ls` isn't in sidebar.json and [won't auto-link](https://dvc-org-ref-aliases-xeonj8d4vz.herokuapp.com/doc/command-reference/list) but `experiments` also isn't, yet [it does](https://dvc-org-ref-aliases-xeonj8d4vz.herokuapp.com/doc/command-reference/exp)..?

rogermparent:
@jorgeorpinel There's an explicit entry in `simpleLinker` for `dvc experiments`, which is how it gets around the `commandLinker` requirement of being in the sidebar.

dberenbaum (approved):

content/linked-terms.js
```diff
@@ -22,5 +22,9 @@ module.exports = [
   {
     matches: 'dvc experiments',
     url: '/doc/command-reference/exp'
+  },
+  {
+    matches: 'dvc ls',
+    url: '/doc/command-reference/list'
```
jorgeorpinel, line 28: @rogermparent I added `dvc ls` here but it's still not linking. See https://dvc-org-ref-aliases-xeonj8d4vz.herokuapp.com/doc/command-reference/list

rogermparent: It works for me locally, so it seems like this is a cache issue.

jorgeorpinel: Thanks. Is it possible to destroy the Heroku review app and recreate it from scratch?

rogermparent: If you set the `DEPLOY_OPTIONS` env var to `build,upload,clean` it should clear out the cache, though it's possible the same issue could persist in master. If you bump `version` in `plugins/gatsby-remark-dvc-linker`, it should do a cache clear naturally in here as well as master. 
I'm looking into a way to invalidate cache when these settings change, but it'll take some changes on Gatsby's end as well as ours even if we pass options through the proper API.

jorgeorpinel: No need to complicate things. But so you think merging this will actually address the problem in question? If so please do @rogermparent !

rogermparent: Not quite, I think a proper addressing of the problem will take some extra work. Merging this into master _may_ not exhibit this problem, but it's also possible it could until the next automatic cache clear (most likely a package upgrade).

I think until then, bumping a patch version on `plugin/gatsby-remark-dvc-linker/package.json` is a quick and simple way to solve the problem here and make sure it doesn't happen on master.

jorgeorpinel: > bumping a patch version

Done and merged. Let's see what happens... ⌛

jorgeorpinel: https://dvc.org/doc/command-reference/list looks good 👍

PR #3005, opened 2021-11-04
Removed non existing `dvc metrics diff --old` option

daavoo:
https://github.com/iterative/dvc.org/pull/2988#issuecomment-961253726


dberenbaum (approved):

jorgeorpinel:
Thanks!

PR #2909, opened 2021-10-09
guide: DVC Experiments Overview

jorgeorpinel:
> Originally extracted from https://github.com/iterative/dvc.org/pull/2654#issuecomment-936772206

- The [guide's index](https://dvc-org-exp-dvc-exps-pa-qhfdse.herokuapp.com/doc/user-guide/experiment-management) goes over Exp Mgmt in general (not only `exp` commands).
- The [new Exps Overview page](https://github.com/iterative/dvc.org/pull/2909/files#diff-59c6f50a088ff9d6ffbb2a0cb284901ea7c6c732fda9b937dbb34df43a8b883c) explains the basic  `dvc exp` workflow (happy path), contains implementation details (extracted [from Running Exps](https://github.com/iterative/dvc.org/pull/2909/files#diff-213c105f41cfb7e3abe29cba4fd80d9cf459bb94703b0a49708b23cf6b6cf5c5)), and **introduces `exp init`**.
- Additionally, [Persisting Exps](https://github.com/iterative/dvc.org/pull/2909/files#diff-64edd45172545bed656fae31b6639b4f9cf6fe51853eeca8f9208eb161527bbd) is now in a single place (stand-alone guide).

OTHER possible goals (may follow-up after this PR if needed):
- [x] Explain about exp ~~IDs and~~ names? From #2768
- [ ] What to say about **Organization patterns**?
  See https://github.com/iterative/dvc.org/pull/2909#pullrequestreview-775621414
  and https://github.com/iterative/dvc.org/pull/2909#pullrequestreview-812329316
- [ ] Improve **Basic workflow** section motivation and clearly differentiate it from the "new" DVC Experiments workflow.
  See https://github.com/iterative/dvc.org/pull/2909#pullrequestreview-819618983
- [ ] Create a diagram for different `exp` workflows? (at least the basic one)
  See https://github.com/iterative/dvc.org/pull/2909#pullrequestreview-794880808 below.
- [ ] Longer `exp init` section? Even rename or split from Overview? See [possible structure](https://github.com/iterative/dvc.org/pull/3015#pullrequestreview-804151331).

Issue #2768, opened 2021-08-31
guide: Running Experiments improvements


jorgeorpinel:
> Follow up to https://github.com/iterative/dvc.org/pull/2690

- [ ] Extract unnecessary details from https://dvc.org/doc/command-reference/exp (mainly from `run`) since a lot of it is explained here now.
- [ ] The pipeline section seems too deep. I think the guide could link to other parts of the docs more liberally -- for now https://dvc.org/doc/start/data-pipelines or https://dvc.org/doc/command-reference/dag#directed-acyclic-graph?
- [ ] Add about metrics.
- [ ] Explain that experiments can also track changes to data, code, etc.
- [ ] Delete note on "each experiment runs in an independent temporary directory" ?
- [ ] Review conceptual level of each section e.g. H2s for main subtopics, H3s for methods, hidden blocks for implementation `<details>`
- [ ] **Checkpoint experiments** is basically pulled from https://dvc.org/doc/command-reference/exp/run#checkpoints. We need to determine what to leave on the cmd ref. (rel 32548)

iesahin:
I believe the command reference should contain examples with one or two sentence explanations and the rest should be moved to UG. Command examples can continue to live in the ref. WDYT @jorgeorpinel

jorgeorpinel:
Yes, we need some cmd ref examples focused on how to use the CLI (not so much on explaining concepts). There can be examples in both places.

PR #3006, opened 2021-11-04
[NESTED] guide: Exp implementation details, naming into Overview

jorgeorpinel:
> Per https://github.com/iterative/dvc.org/pull/2909#discussion_r743028065

### DIFF
--- a/content/docs/user-guide/experiment-management/experiments-overview.md
+++ b/content/docs/user-guide/experiment-management/experiments-overview.md
@@ -16,7 +16,7 @@ are not pushed to Git remotes by default (see `dvc exp push`).
 </details>
 
 Experiments will have an auto-generated ID like `exp-bfe64` by default. A custom
-name can be given instead, using the `--name` (`-n`) option of `dvc exp run`.
+name can be given instead (using the `--name`/`-n` option of `dvc exp run`).
 
 > ID or name can be used to reference experiments with `dvc exp` subcommands.
 
--- a/content/docs/user-guide/experiment-management/experiments-overview.md
+++ b/content/docs/user-guide/experiment-management/experiments-overview.md
@@ -1,10 +1,29 @@
 # DVC Experiments Overview
 
-DVC Experiments are captures automatically by DVC when you [run them].
+DVC Experiments are captures automatically by DVC when you [run them]. Each
+experiment creates and tracks a project variation based on the changes in your
+<abbr>workspace</abbr>. Experiments preserve the latest commit in the current
+branch (Git `HEAD`) as their parent or _baseline_.
+
+<details>
+
+### ⚙️ How does DVC track experiments?
+
+Experiments are custom [Git references] (found in `.git/refs/exps`) with a
+single commit based on `HEAD` (not checked out by DVC). Note that these commits
+are not pushed to Git remotes by default (see `dvc exp push`).
+
+</details>
+
+Experiments will have an auto-generated ID like `exp-bfe64` by default. A custom
+name can be given instead, using the `--name` (`-n`) option of `dvc exp run`.
+
+> ID or name can be used to reference experiments with `dvc exp` subcommands.
 
 [run them]: /doc/user-guide/experiment-management/running-experiments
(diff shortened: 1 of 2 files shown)


content/docs/user-guide/experiment-management/experiments-overview.md
```diff
@@ -1,10 +1,29 @@
...
+Experiments are custom [Git references] (found in `.git/refs/exps`) with a
+single commit based on `HEAD` (not checked out by DVC). Note that these commits
+are not pushed to Git remotes by default (see `dvc exp push`).
+
+</details>
+
+Experiments will have an auto-generated ID like `exp-bfe64` by default. A custom
+name can be given instead (using the `--name`/`-n` option of `dvc exp run`).
+
+> ID or name can be used to reference experiments with `dvc exp` subcommands.
```
jorgeorpinel, line 21: @iesahin I'm not 100% sure this is right or consistent with the language you've used in other guides. Are `exp-abcde` called "IDs" or "auto-generated names"? Are the Git hashes called IDs instead? The hashes don't exist for queued experiments so it may be confusing to call those IDs.

We should make sure this terminology makes sense and is consistent 🙂 - let's figure it out here and I'll apply everywhere else.

content/docs/user-guide/experiment-management/experiments-overview.md
```diff
@@ -1,10 +1,29 @@
...
+### ⚙️ How does DVC track experiments?
+
+Experiments are custom [Git references] (found in `.git/refs/exps`) with a
+single commit based on `HEAD` (not checked out by DVC). Note that these commits
+are not pushed to Git remotes by default (see `dvc exp push`).
+
+</details>
+
+Experiments will have an auto-generated ID like `exp-bfe64` by default. A custom
+name can be given instead (using the `--name`/`-n` option of `dvc exp run`).
```
jorgeorpinel, line 19: Also (@iesahin), [Running Experiments](https://dvc.org/doc/user-guide/experiment-management/running-experiments) doesn't cover `--name` so I had to mention the command directly from here. It'd be ideal if there's a section there to link from here instead...

jorgeorpinel: Added a check box about this to https://github.com/iterative/dvc.org/issues/2768.

shcheklein (approved):

shcheklein:
@jorgeorpinel let's merge this, to then review the main one. It's hard otherwise. Probably not the best case to use nesting (it's better when changes do not overlap much / independent by their nature - not when they both change the same files for the same reason).

jorgeorpinel:
> let's merge this, to then review the main one

Are you sure? It's going to make reviewing https://github.com/iterative/dvc.org/pull/2909 harder. We could leave this unmerged until the main one is approved instead.

Agree that so far nesting hasn't been helpful. I'll stop trying 😓

content/docs/user-guide/experiment-management/experiments-overview.md
iesahin: I was using ID and name interchangeably to tell `exp-abc123` like names, which can be specified with `--name` option to `dvc  exp run`. Unfortunately, there are places that we need to tell `refs/exps/abc123eeefff333/` IDs as commit IDs. 

You're right that we should avoid ID to denote the former, and call the latter _commit ID_ instead. Could this be the naming convention?

jorgeorpinel: "commit ID" may also be confusing though. We already have a term for that I think? I.e. SHA or commit hash.

Maybe we only need "SHA/hash" and "name"? Avoid "ID" completely. WDYT?

iesahin: Could we put "experiment name" to the glossary? When we mention it for the first time, we can link to it. 

I'll try to avoid ID completely. 

Commit hash = SHA-256 of Git commits
Experiment name = `exp-123ab` like names in experiments.

iesahin: Can we have a wiki page about these notes, and review time to time? @jorgeorpinel

Issue #2974, opened 2021-10-27
ref: Document `dvc exp init`


iesahin:
It looks the help text is there: 

```
$ dvc exp init --help
usage: dvc experiments init [-h] [-q | -v] [--run] [--interactive] [-f] [--explicit] [--name NAME] [--code CODE] [--data DATA] [--models MODELS]
                            [--params PARAMS] [--metrics METRICS] [--plots PLOTS] [--live LIVE] [--type {default,live}]
                            ...

Initialize experiments.
Documentation: <https://man.dvc.org/exp/init>

positional arguments:
  command               Command to execute.

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Be quiet.
  -v, --verbose         Be verbose.
  --run                 Run the experiment after initializing it
  --interactive, -i     Prompt for values that are not provided
  -f, --force           Overwrite existing stage
  --explicit            Only use the path values explicitly provided
  --name NAME, -n NAME  Name of the stage to create
  --code CODE           Path to the source file or directory which your experiments depend (default: src)
  --data DATA           Path to the data file or directory which your experiments depend (default: data)
  --models MODELS       Path to the model file or directory for your experiments (default: models)
  --params PARAMS       Path to the parameters file for your experiments (default: params.yaml)
  --metrics METRICS     Path to the metrics file for your experiments (default: metrics.json)
  --plots PLOTS         Path to the plots file or directory for your experiments (default: plots)
  --live LIVE           Path to log dvclive outputs for your experiments (default: dvclive)
  --type {default,live}
                        Select type of stage to create (default: default)

```

We don't have a `dvc exp init` ref page as of now.

jorgeorpinel:
Yeah I think we need a larger epic to document the upgraded Experiments release actually. Let's plan it tomorrow? Cc @dberenbaum

iesahin:
What's the approximate release schedule for `dvc exp init`? @dberenbaum @skshetry

dberenbaum:
November 18th

jorgeorpinel:
So we did not get started on this yet for several reasons, but we should soon! @dberenbaum @skshetry at this point are you available to work on a cmd ref draft? Thanks

PR #3015, opened 2021-11-11
[WIP] cmd-ref: document exp init

skshetry:
Fixes https://github.com/iterative/dvc.org/issues/2974.


dberenbaum:
@jorgeorpinel What sections do you think are needed for the cmd-ref? We will definitely want a user guide for this, but since we are trying to keep the cmd-ref more narrow, maybe we can eliminate some of these sections.

content/docs/command-reference/exp/index.md
```diff
@@ -4,6 +4,7 @@ _New in DVC 2.0_
 
 A set of commands to generate and manage <abbr>experiments</abbr>:
 [run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
+[init](/doc/command-reference/exp/init),
```
jorgeorpinel, line 7: Should we feature init first?
```suggestion
[init](/doc/command-reference/exp/init),
[run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
```

jorgeorpinel, line 7: Should we feature init first?
```suggestion
[init](/doc/command-reference/exp/init),
[run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
```

jorgeorpinel, line 7: Should we feature init first?
```suggestion
[init](/doc/command-reference/exp/init),
[run](/doc/command-reference/exp/run), [show](/doc/command-reference/exp/show),
```

jorgeorpinel (commented):
> GH seems to have a bug where I can't leave single comments, only review comments... Sorry for the noise.

content/docs/command-reference/exp/index.md
```diff
@@ -37,6 +38,7 @@ positional arguments:
     push                Push a local experiment to a Git remote.
     pull                Pull an experiment from a Git remote.
     remove              Remove local experiments.
+    init                Initialize experiments.
```
jorgeorpinel, line 41: Sounds like a requirement (like `dvc init`. How can we better describe what this does for the user? (I need to read the rest of the changes before I can suggest something.)

jorgeorpinel (commented):
> More noise... Ignore these

content/docs/command-reference/exp/index.md
jorgeorpinel: Key phrases:
> helps you quickly get started with experiments
reduces boilerplate
initializing pipeline stages

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
+# exp init
+
+Initializes experiments.
```
jorgeorpinel, line 3: p.s applies here too.

jorgeorpinel (commented):
On the command description

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+customized through config.
+
+It also offers guided `--interactive` mode for creating a stage to be `exp run`
+later. `dvc exp init` supports creating different types of stages, eg: live if
+you are using dvclive to track metrics and plots, etc.
+
+### The `command` argument
+
+The `command` argument is optional, if you are using `--interactive` mode. The
+`command` sent to `dvc exp init` can be anything your terminal would accept and
```
jorgeorpinel, line 34: Should we put this under **Options** like for `targets` in https://dvc.org/doc/command-reference/repro#options (and other places I think) ? That was an initiative of @skshetry actually 🙂

jorgeorpinel (commented):
> Noise

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+  the invocation.
+
+- `--metrics` - override the path to metrics file to track, which your
+  experiment produces. Default is `metrics.json` file.
+
+- `--plots` - override the path to plots file or directory, which your
+  experiment produces. The default is `plots`.
+
+- `--live` - override the directory `path` for
+  [DVCLive](/doc/dvclive/user-guide/dvclive-with-dvc), which your experiment
```
jorgeorpinel, line 89: Fixed link checker error
```suggestion
  [DVCLive](/doc/dvclive/dvclive-with-dvc), which your experiment
```

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Guided/Interactive mode
+
+## Examples
+
+#### Default stage
+
+#### Dvclive stage
+
+diff --git a/content/docs/sidebar.json b/content/docs/sidebar.json
```
jorgeorpinel, line 137: Typo?
```suggestion
```

jorgeorpinel (commented):
Couple small fixes (committing...)

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Non-interactive mode
+
+## Guided/Interactive mode
+
+## Examples
+
+#### Default stage
+
+#### Dvclive stage
```
jorgeorpinel, line 135: > @jorgeorpinel What sections do you think are needed for the cmd-ref? We will definitely want a user guide for this, but since we are trying to keep the cmd-ref more narrow, maybe we can eliminate some of these sections.

@dberenbaum yes we can keep it much leaner now and leave much of the details for a guide. But it's good to have a proposed structure in advance so thanks @skshetry 

For now each proposed section can be a short paragraph in the description without any subtitle structure. Let's just mention the features without explaining exactly how they work as the guide will do that. For now just awareness + the help output will have to suffice I think.

jorgeorpinel (commented):
> Noise

content/docs/command-reference/exp/init.md
jorgeorpinel: OK so for the short description, how about something like

> Helper command to setup and run experiments from scratch.

Or something more specific like

> Codify a project variation using [DVC metafiles] and run it as a controlled [experiment]

content/docs/command-reference/exp/index.md
jorgeorpinel: For the help output maybe an even shorter form of the specific desc.

> Codify a project variation and run it as an experiment

p.s. and re "sounds like a requirement (like `dvc init`)" - can we still reconsider the name? 😅

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+  different kinds of stages: `default` and `live`. If unspecified, `default`
+  stage is created.
+
+  `default` stage creates a stage with `metrics` and `plots` tracked by DVC
+  itself, and does not track live-created artifacts (unless explicitly
+  specified).
+
+  With `--type=live`, `dvc exp init` will create a stage where `metrics` and
+  `plots` are tracked by `dvclive` itself (unless `--metrics` and `--plots` are
+  explicitly provided).
```
dberenbaum, line 103: It might be helpful to discuss the intended use cases. The default stage is for classical ML experiments, while live is for deep learning. We should probably also mention the checkpoints in live.

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+## Description
+
+`dvc exp init` helps you quickly get started with experiments, with the command
+to run and tracks data, params, source code, models, metrics and plots by
+default.
+
+It reduces boilerplate for initializing [pipeline](/doc/command-reference/dag)
+stages in a `dvc.yaml` file by assuming sane defaults about the location of your
+data, parameters, source code, models, metrics and plots, which can be
+customized through config.
```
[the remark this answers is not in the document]
jorgeorpinel: How do you customize "through config"? `dvc config`? Config files? Which one(s). Thanks

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+## Description
+
+`dvc exp init` helps you quickly get started with experiments, with the command
+to run and tracks data, params, source code, models, metrics and plots by
+default.
+
+It reduces boilerplate for initializing [pipeline](/doc/command-reference/dag)
+stages in a `dvc.yaml` file by assuming sane defaults about the location of your
```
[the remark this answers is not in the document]
jorgeorpinel: And `dvc exp run`. And maybe `params` and `metrics` for that matter.

content/docs/command-reference/exp/init.md
jorgeorpinel: Not sure though. That should be explained in DVCLive docs and possibly also in a future `exp init` guide? I'd just link to DVC here too.

dberenbaum: It shouldn't really be about dvclive at all. It's about deep learning scenarios for which checkpoints and dvclive were built.

dberenbaum: We can still put it in the guide. Just want to clarify this doesn't really belong in dvclive docs.

content/docs/command-reference/exp/index.md
skshetry: Makes sense.

content/docs/command-reference/exp/init.md
skshetry: I don't want to add `helper` as it feels redudant. I don't want to emphasize `and run experiments` here even though it does have `--run` flag. And, `from scratch` may not always be the case.

So, `Command to setup experiments` does not seem that different to `Initializes experiments`.

content/docs/command-reference/exp/init.md
skshetry: Let's keep it as it is for now, as we also need to do something for `stage add`/`run`?

content/docs/command-reference/exp/init.md
skshetry: yeah, looks like I copied something from `github.dev` editor.

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
skshetry: > How do you customize "through config"? dvc config? Config files? Which one(s).

Both? But I am not sure we need to specify that, as `dvc config` operates on config files.

content/docs/command-reference/exp/init.md
jorgeorpinel: Sounds good. I was confused about the running part so we can avoid that. And the word "helper" is not necessary. But the point is to have a more descriptive description. Do you have a suggestion @skshetry ?

> Command to setup experiments does not seem that different to Initializes experiments

It's actually quite different. More informative. And I think we can improve it even more.

content/docs/command-reference/exp/init.md
jorgeorpinel: Why keep it as a section here when we moved it in `repro` ? We had lots of good reasons (which you proposed), what's different?

> do something for stage add/run

Out of scope but we can create an issue.

content/docs/command-reference/exp/init.md
jorgeorpinel: No worries.

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
jorgeorpinel: OK. My point is that need to be a bit more specific and/or link to what that config looks like (how it's setup). Also, separate Q:

> to reuse across projects

Does that mean using user/system config? I think all these details should be somewhere, maybe a separate subsection for now and link there from here? Thanks

content/docs/command-reference/exp/init.md
skshetry: Because it's better to do it at the same time for all commands. I am going with what we have right now, let's handle it later?

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
skshetry: yeah, I have a section below with the title `Setting up custom config/Sharing configs`. Just need to hyperlink it, and fill that section.

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
skshetry: `exp run` should be linked in the next line. `params` and `metrics` should be easy. I'm having trouble adding `stage add` here.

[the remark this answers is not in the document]
skshetry: hmm, maybe we can bring `stage add` by mentioning why `exp init` exists instead?

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+You can leave some configurations to use default values. This can be useful in a
+system or global config so that you can avoid repeating these paths if all of
+your projects share a similar structure.
+
+## Non-interactive mode
+
+## Guided/Interactive mode
+
```
skshetry, line 153: I don't really want to show all the prompts. And I think most of the information are already mentioned elsewhere, so I am thinking of leaving these fields.

content/docs/command-reference/exp/init.md
```diff
@@ -0,0 +1,158 @@
...
+## Examples
+
+#### Default stage
+
+#### Dvclive stage
```
skshetry, line 158: Also thinking of leaving these fields out for now.

dberenbaum:
What's the status on this @skshetry @jorgeorpinel? I can take over if needed. This should get merged today.

dberenbaum (approved):
Merging this since we need some reference available.

content/docs/command-reference/exp/init.md
jorgeorpinel: p.s. we don't always make changes for all commands at once, we can take them as they come. BTW I actually like a section better than having it under Options, I'm just confused since we discussed this a lot for `repro` and you (and Ivan) were very sure about putting it under Options 😆

jorgeorpinel: > Because it's better to do it at the same time for all commands. I am going with what we have right now, let's handle it later?

https://github.com/iterative/dvc.org/pull/3071#pullrequestreview-829161122

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
jorgeorpinel: Hi @skshetry is there some more info on `exp init` configuration files in the wiki or somewhere? Did anything change in the `config` command? Thanks

content/docs/command-reference/exp/init.md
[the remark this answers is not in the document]
jorgeorpinel: > would be good to link to stage add

FYI in #3071 I changed that link for `dvc.yaml` because we want to encourage people to write that manually. And added a link to `stage add` from there.

content/docs/command-reference/exp/init.md
jorgeorpinel: Good call. If we ever make a detailed guide we can include all prompts but it's probably unnecessary either way.

content/docs/command-reference/exp/init.md
jorgeorpinel: We'll want some examples though. Added task in #3071

PR #3019, opened 2021-11-13
ref: copy edit Kerberos info in `remote add` (WebHDFS)

jorgeorpinel:
rel. https://github.com/iterative/dvc.org/pull/3009


content/docs/command-reference/remote/add.md
```diff
@@ -304,18 +304,15 @@ $ dvc remote add -d myremote hdfs://user@example.com/path
 
 **HDFS and WebHDFS:**
 
-Both remotes, HDFS and WebHDFS, allow using a Hadoop cluster as a remote
-repository. However, HDFS requires `libhdfs`, an interface to the Java Hadoop
-client, that must be installed separately. Meanwhile, WebHDFS has no need for
-this requirement as it communicates with the Hadoop cluster via a REST API.
+Both HDFS and WebHDFS allow using a Hadoop cluster as remote storage. However,
+HDFS requires `libhdfs`, a Python library to use the Java Hadoop client, and it
+must be installed separately. WebHDFS has no need for this, as it uses a REST
```
jorgeorpinel, line 309: @shcheklein should we move the info about HDFS needing some Python library to the HDFS section? Then we can remove the confusing **HDFS and WebHDFS** bold header. WDYT? I'm not 100% sure how this works so maybe that's not right.

jorgeorpinel: Cc @efiop or maybe you can confirm? Thanks

efiop: libhdfs is not a python library it is just a separate library to install. We use pyarrow python library that requires it.

jorgeorpinel: OK. Found some more details in https://github.com/iterative/dvc/issues/5858#issuecomment-824216700. I separated the notes now (4ed771f6).

shcheklein (approved):

Issue #2447, opened 2021-05-05
Upgrade to Gatsby 3


julieg18:
Currently we are at Gatsby 2, which holds back our image processing.

By upgrading to Gatsby v3, we can:
- [ ] Upgrade `sharp` and corresponding image packages(`gatsby-plugin-sharp`, `gatsby-transformer-sharp`, `gatsby-image`, etc.)
- [ ] Upgrade to Node version 16(Currently at 15)

Issues that will need to be resolved when we upgrade:
- [ ] Upgrading to v3 will likely break our custom caching mechanism, we'll need to make sure it's working and correct any issues that may arise.

shcheklein:
@julieg18 this is a great summary, thanks! I wonder if we have some ticket related to caching already (cc @rogermparent )?

julieg18:
I couldn't find one, but there are quite a few issues, I could have easily missed one 🤔

rogermparent:
#1063 is an old ticket about image caching that can provide some context, particularly the latest comment from me on there can be considered a call for this Issue (though I think it was still Gatsby 2 then)

julieg18:
> There's been a lot of iterations on those plugins since then, so it may be worth trying an update in a PR to see if the cache behavior improvements also fixed our issue. If we are able to use the latest versions of those plugins, we'll get access to a slew of new features like that greatly improve DX like lazy queries/images and true parallel processing

I attempted to update dvc.org to Gatsby 3 but even with the latest updates to Gatsby and its packages, our caching mechanism still breaks when deploying. Though I'm not sure if its for a different reason than last time we attempted this.

Error Message in Failed Build Script:
<details>

```
success Building HTML renderer - 20.597s
error Page data from page-data.json for the failed page "/doc": {
  "componentChunkName": "component---src-templates-doc-tsx",
  "path": "/doc",
  "result": {
    "data": {
      "page": {
        "title": null,
        "description": null,
        "htmlAst": {
          "type": "root",
          "children": [
            {
              "type": "element",
              "tagName": "h1",
              "properties": {
                "id": "dvc-documentation",
                "style": "position:relative;"
              },
              "children": [
                {
                  "type": "text",
                  "value": "DVC Documentation"
                },
                {
                  "type": "element",
                  "tagName": "a",
                  "properties": {
                    "href": "#dvc-documentation",
                    "ariaLabel": "dvc documentation permalink",
                    "className": [
                      "anchor",
                      "after"
                    ]
                  },
                  "children": [
[...]

PR #3021, opened 2021-11-15
Upgrade to Gatsby 4 and fix images paths that broke builds

rogermparent:
I decided to use Gatsby 3 in #2447 since Gatsby 4 was still in beta, most plugins hadn't updated to match it, and it seemed to be a bit less stable in Heroku builds.

Since then, Gatsby 4 has left beta and most plugins have updated to work with it. Also, Gatsby Cloud's image processing has integrations specific to 4, so this PR will be used to test on Gatsby Cloud.

Also, it seems one of the upgrades since then has made it to that builds fail when an image path isn't found- this is a good thing, since we had a few images that were completely broken and just unnoticed, and some "phantom" images that worked when they shouldn't have.

This PR also updates the `typescript` and `eslint` packages, which was required because `react-use` uses named tuples, a feature introduced in newer TypeScript versions, and eslint needed to be upgraded alongside TypeScript. 
I also upgraded `stylelint`, which then discovered a whole ton of CSS problems that we had before, including obvious stuff like duplicated keys. The results of `stylelint --fix` ended up being the bulk of the PR's diff.


julieg18 (commented):
Looks great! 

I noticed that the cards in the Doc's home page is styled differently:

image

I have no idea why though 🤔

rogermparent:
@julieg18 Great catch on the card collapsing! I'm not sure why that would have changed, I know the major upgrade on `github-markdown-css` caused some issues, but I don't see how it would cause this completely custom flex setup to start breaking down.

I just fixed it by using slightly more proper `margin`-based spacing, though I had to apply it vertically to `cardWrapper`s and horizontally to `card`s, it wouldn't work otherwise.

rogermparent:
Not sure why Restyled is trying to change this, running `format-all` locally doesn't apply any changes, and the changes restyled is asking for cause `lint-ts` to break.

julieg18:
> Not sure why Restyled is trying to change this, running `format-all` locally doesn't apply any changes, and the changes restyled is asking for cause `lint-ts` to break.

I think I had the same issue when I tried to edit that particular file that Restyled is trying to restyle. I think it had something to do with the kind of environment that Restyled is using to run code.

rogermparent:
I took a check around all the top-level pages and usages of the `cards` custom component, I haven't caught anything obvious. Willing to give this time for another once-over to be extra sure, but I think we're ready to send this in (and probably deal with some link check errors :sob:)

julieg18 (approved):
Looks good!

PR #3035, opened 2021-11-18
ref: Update `dvc run` references in `dvc commit`

### LINKED ISSUE #3031: ref: Update `dvc run` references in `dvc commit`
This is to replace `dvc run` examples with `dvc stage add` in `dvc commit` reference.

iesahin:
Fixes #3031

### DIFF
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -130,9 +130,9 @@ $ dvc pull -aT
 
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
-the <abbr>cache</abbr> with undesired intermediate results, you can
-use the `--no-commit` option of `dvc repro`. Once your progress is good
-enough, `dvc commit` can be used to store data files in the cache.
+the <abbr>cache</abbr> with undesired intermediate results, you can use the
+`--no-commit` option of `dvc repro`. Once your progress is good enough,
+`dvc commit` can be used to store data files in the cache.
 
 In the `featurize` stage, `src/featurization.py` is executed. A useful change to
 make is adjusting the parameters for that script. The parameters are defined in
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -131,9 +131,10 @@ $ dvc pull -aT
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
 the <abbr>cache</abbr> with undesired intermediate results, we can run a single
-stage with `dvc exp run --no-commit`, or `dvc repro --no-commit`. This prevents
-data from being pushed to cache. When development of the stage is finished,
-`dvc commit` can be used to store data files in the cache.
+stage with `dvc repro --no-commit <target>` or reproduce an entire pipeline with
+`dvc repro --no-commit`. This prevents data from being pushed to the cache. When
+development of stage is finished, `dvc commit` can be used to store data files
+in the cache.
 
 In the `featurize` stage, `src/featurization.py` is executed. A useful change to
 make is adjusting the parameters for that script. The parameters are defined in
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -128,12 +128,12 @@ $ dvc pull -aT
 
 ## Example: Rapid iterations
 
-Sometimes we want to iterate through multiple changes to configuration, code,
-or data, trying different ways to improve the output of a stage. To avoid
-filling the <abbr>cache</abbr> with undesired intermediate results, we can run
-a single stage with `dvc exp run --no-commit`, or `dvc repro --no-commit`. This
-prevents data from being pushed to cache. When development of the stage is
-finished, `dvc commit` can be used to store data files in the cache.
+Sometimes we want to iterate through multiple changes to configuration, code, or
+data, trying different ways to improve the output of a stage. To avoid filling
+the <abbr>cache</abbr> with undesired intermediate results, we can run a single
+stage with `dvc exp run --no-commit`, or `dvc repro --no-commit`. This prevents
+data from being pushed to cache. When development of the stage is finished,
+`dvc commit` can be used to store data files in the cache.
 
 In the `featurize` stage, `src/featurization.py` is executed. A useful change to
 make is adjusting the parameters for that script. The parameters are defined in
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -34,7 +34,8 @@ these step(s) during the process of tracking each file or directory:
 - Store the file contents in the cache.
 
 > Skipping these steps is typically done to avoid caching unfinished data, for
-> example when [experimenting](/doc/command-reference/exp).
+> example when exploring different data or
+> [stages](/doc/command-reference/stage)
 
 More specifically, scenarios for `dvc commit` include:
 
@@ -131,10 +132,10 @@ $ dvc pull -aT
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
 the <abbr>cache</abbr> with undesired intermediate results, we can run a single
-stage with `dvc repro --no-commit <target>` or reproduce an entire pipeline with
-`dvc repro --no-commit`. This prevents data from being pushed to the cache. When
-development of stage is finished, `dvc commit` can be used to store data files
-in the cache.
+stage with `dvc repro --no-commit <target>`, or reproduce an entire pipeline
+with `dvc repro --no-commit`. This prevents data from being pushed to the cache.
+When development of stage is finished, `dvc commit` can be used to store data
+files in the cache.
 
 In the `featurize` stage, `src/featurization.py` is executed. A useful change to
 make is adjusting the parameters for that script. The parameters are defined in
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -128,13 +128,13 @@ $ dvc pull -aT
 
 ## Example: Rapid iterations
 
-Sometimes we want to iterate through multiple changes to configuration, code, or
-data, trying different ways to improve the output of a stage. To avoid filling
-the <abbr>cache</abbr> with undesired intermediate results, we can run a single
-stage with `dvc repro --no-commit <target>`, or reproduce an entire pipeline
-with `dvc repro --no-commit`. This prevents data from being pushed to the cache.
-When development of stage is finished, `dvc commit` can be used to store data
-files in the cache.
+Sometimes we want to iterate through multiple changes to configuration, code,
+or data, trying different ways to improve the output of a stage. To avoid
+filling the <abbr>cache</abbr> with undesired intermediate results, we can run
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/commit.md
```diff
@@ -34,8 +34,7 @@ these step(s) during the process of tracking each file or directory:
 - Store the file contents in the cache.
 
 > Skipping these steps is typically done to avoid caching unfinished data, for
-> example when exploring different data or
-> [stages](/doc/command-reference/run)).
+> example when [experimenting](/doc/command-reference/exp).
```
jorgeorpinel, line 37:
```suggestion
> example when [experimenting](/doc/user-guide/experiment-management).
```

jorgeorpinel: But I'm not sure we want to plug experiments here. I don't remember `commit` used much if at all in any exp ref or guide. Probably roll back

content/docs/command-reference/commit.md
```diff
@@ -132,7 +131,7 @@ $ dvc pull -aT
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
 the <abbr>cache</abbr> with undesired intermediate results, we can run a single
-stage with `dvc run --no-commit`, or reproduce an entire pipeline using
+stage with `dvc exp run --no-commit`, or reproduce an entire pipeline using
```
jorgeorpinel, line 134: Idk. This produces an experiment which also fills the cache. Should be `repro` I think.

jorgeorpinel (changes requested):
☝️

content/docs/command-reference/commit.md
iesahin: They'll need when  using `--no-commit` flag to `dvc exp run`. For the singular use, `exp run` is no different from `dvc repro`, it leaves artifacts in the workspace and those may be committed or not.

content/docs/command-reference/commit.md
iesahin: Which cache? Run cache or the file cache? `dvc exp run` command works similar to repro, commits artifacts to `.dvc/cache`

content/docs/command-reference/commit.md
jorgeorpinel: It's different. It creates an experiment (Git custom ref. etc.) If this is not about experimentation let's use non-exp concepts.

content/docs/command-reference/commit.md
jorgeorpinel: Regardless of the cache question, `exp run` produces an experiment. We have not decided to use experiment runs as the default pipeline workflow that I know of so any non-experiment specific docs should not use exp commands. Thus `repro` here.

content/docs/command-reference/commit.md
shcheklein: Agreed, it's also about `dvc add` which might be used outside experiments.

content/docs/command-reference/commit.md
shcheklein: Agreed, this is not a valid change semantically.

Probably `dvc stage add` and then `dvc repro --no-commit <stage>` is the way to replace it.

Or ideally `dvc stage add` should have `--execute`.

shcheklein (changes requested):
Both changes do not look semantically correct at the moment. Let's try to fix / improve this.

content/docs/command-reference/commit.md
iesahin: I replace `repro` with `exp` here, because the sentence mentions "unfinished data." Usually we have "unfinished data" while experimenting. I can add "or [running pipelines] (repro)" here, or remove this sentence completely. 

`dvc add` is mentioned in the paragraph one above.

jorgeorpinel: > Usually we have "unfinished data" while experimenting.

Sure, "experimenting" in a very general sense, as in "trying things out". That's still not the same as a data science experiment (param tuning, etc.)

> I can add "or [running pipelines] (repro)" here, or remove this sentence completely.I don't think it's a problem to mention "unfinished data" here without associating it with `dvc exp`.

>

jorgeorpinel: OOPS FIXING PREVIOUS COMMENT:

> I can add "or [running pipelines] (repro)" here, or remove this sentence completely.

I don't think it's a problem to mention "unfinished data" here without associating it with `dvc exp`. And not sure I see how running pipelines relates exactly to `--no-commit` -- you can still run it right? But again that's a jump away from the topic I think. We can just roll this link back IMO.

content/docs/command-reference/commit.md
jorgeorpinel: > Probably dvc stage add and then dvc repro --no-commit <stage> is the way to replace it.

Agree, emphasizing the `<stage>` given as `target`.

> I removed the "entire pipeline" phrase

Why? It's a valid situation we want to mention. If we're going to change that, rather generalize to `repro --no-commit` only and don't mention either specific-stage(s) nor entire-pipeline cases, as that distinction is besides the point, really.

jorgeorpinel (commented):
If we're going to introduce `exp run` here we should start by adding it to line 29 where the other commands that have `--no-commit` are mentioned. But I'm still not sure we want to mix the concepts here. Do we really expect users to `exp run --no-commit` ? Cc @dberenbaum WDYT

For now can you please address existing feedback avoiding `exp run` @iesahin ? Introducing that may require more relevant changes to the text/story (probably best to create a separate ticket). Thanks

dberenbaum:
> If we're going to introduce `exp run` here we should start by adding it to line 29 where the other commands that have `--no-commit` are mentioned. But I'm still not sure we want to mix the concepts here. Do we really expect users to `exp run --no-commit` ? Cc @dberenbaum WDYT
> 
> For now can you please address existing feedback avoiding `exp run` @iesahin ? Introducing that may require more relevant changes to the text/story (probably best to create a separate ticket). Thanks

I can see your point @iesahin. It feels like this section was written for experimentation before we had `exp run`, so it makes sense to either update it to use experiments or change the examples.

However, let's not introduce `exp run` as part of replacing `dvc run` references and keep this as a separate issue. Doing `dvc exp run --no-commit` seems pretty dangerous since it makes the experiment not reproducible. In `dvc run/repro --no-commit`, the user controls when the commit happens, so they can still make sure to `dvc commit` before doing git commit, but there's no (easy) way to update the experiment ref with the committed data.

jorgeorpinel:
Per https://github.com/iterative/dvc/issues/7110#issuecomment-988234008 let's double check that `--no-commit` even does anything in `dvc exp` before considering introducing that here (as a separate ticket).

pmrowla:
from: https://github.com/iterative/dvc/issues/7110#issuecomment-988406787

`--no-commit` has no effect in `exp run`, by definition, in order for DVC to be saving an experiment so it can be accessed later, the data has to be cached

This is already mentioned in the command reference in the docs (https://dvc.org/doc/command-reference/exp/run#options)

> In addition to the following, `dvc exp` run accepts all the options in `dvc repro`, with the exception that `--no-commit` has no effect here.

---

wrt DVC pipelines, the docs for `commit` should only mention `dvc run --no-commit` and/or `dvc repro --no-commit`.

Like Ivan already mentioned (https://github.com/iterative/dvc.org/pull/3035#discussion_r757722530), `dvc run --no-commit ...` does actually still have a valid use case (at least until `stage add` has some execution flag), where a user is still designing/debugging a stage and wants to test what the current version of their command would output (but does not want to commit the resulting data into DVC cache).

content/docs/command-reference/commit.md
iesahin: Restored the phrase now. PTAL

content/docs/command-reference/commit.md
iesahin: Removed it. PTAL. @jorgeorpinel

content/docs/command-reference/commit.md
```diff
@@ -35,7 +35,7 @@ these step(s) during the process of tracking each file or directory:
 
 > Skipping these steps is typically done to avoid caching unfinished data, for
 > example when exploring different data or
-> [stages](/doc/command-reference/run)).
+> [stages](/doc/command-reference/stage)
```
jorgeorpinel, line 38: There's also `<abbr>stage</abbr>` - probably better?

content/docs/command-reference/commit.md
```diff
@@ -132,9 +132,9 @@ $ dvc pull -aT
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
 the <abbr>cache</abbr> with undesired intermediate results, we can run a single
-stage with `dvc run --no-commit`, or reproduce an entire pipeline using
-`dvc repro --no-commit`. This prevents data from being pushed to cache. When
-development of the stage is finished, `dvc commit` can be used to store data
+stage with `dvc repro --no-commit <target>`, or reproduce an entire pipeline
+with `dvc repro --no-commit`. This prevents data from being pushed to the cache.
```
jorgeorpinel, line 136: I'd combine these into a single one (probably no need to differentiate between single or multiple stages) but OK, not the direct purpose of this PR.

jorgeorpinel (approved):

content/docs/command-reference/commit.md
jorgeorpinel: Thanks.

content/docs/command-reference/commit.md
iesahin: Done. Thx.

content/docs/command-reference/commit.md
iesahin: I've merged them, could you PTAL?

content/docs/command-reference/commit.md
```diff
@@ -132,10 +131,9 @@ $ dvc pull -aT
 Sometimes we want to iterate through multiple changes to configuration, code, or
 data, trying different ways to improve the output of a stage. To avoid filling
 the <abbr>cache</abbr> with undesired intermediate results, we can run a single
```
jorgeorpinel, line 133:
```suggestion
the <abbr>cache</abbr> with undesired intermediate results, you can
```

content/docs/command-reference/commit.md
```diff
@@ -132,10 +131,9 @@ $ dvc pull -aT
...
-stage with `dvc run --no-commit`, or reproduce an entire pipeline using
-`dvc repro --no-commit`. This prevents data from being pushed to cache. When
-development of the stage is finished, `dvc commit` can be used to store data
-files in the cache.
+stage, or reproduce an entire pipeline by supplying `--no-commit` option to
+`dvc repro`. This prevents data from being pushed to the cache. When development
+of stage is finished, `dvc commit` can be used to store data files in the cache.
```
jorgeorpinel, line 136:
```suggestion
use the `--no-commit` option of `dvc repro`. Once your progress is good
enough, `dvc commit` can be used to store data files in the cache.
```

content/docs/command-reference/commit.md
jorgeorpinel: Thanks, I'll leave a suggestion that further generalizes this. ⌛

jorgeorpinel (commented):
How about this @iesahin ? If you agree please commit, Restyle, and merge.

shcheklein (approved):

PR #2979, opened 2021-10-27
blog: Exp versioning

dberenbaum:
First draft of experiment versioning blog post. Mostly worked on the intro. The tutorial section is outdated, mainly from when we were still focusing more on "lightweight experiments" rather than "experiment versioning."

PR #3051, opened 2021-11-30
Update GS experiments with `dvc exp init`

iesahin:
This one is migrated from Notion. I have added a hidden section on hyperparameters there. 

I have modified here: 

- It starts with `dvc exp init` now. Without mentioning any pipelines stuff or `dvc.yaml`

- I'm planning to remove "If you've run `repro` before" hidden section, the section where we tell how to prepare pipelines for arbitrary projects, and keep the hyperparameter stuff to a minimum.

### DIFF
--- a/content/docs/start/experiments.md
+++ b/content/docs/start/experiments.md
@@ -138,6 +138,55 @@ Experiment results have been applied to your workspace.
 
 <details>
 
+### ℹ️ More information about (Hyper)parameters and metrics
+
+It's pretty common for data science pipelines to include configuration files
+that define adjustable parameters to train a model, do pre-processing, etc. DVC
+provides a mechanism for stages to depend on the values of specific sections of
+such a file (YAML, JSON, TOML, and Python formats are supported).
+
+The `train` stage is created with the following `dvc stage add` command. Notice
+the arguments with `-p` option (short for `--params`)
+
+```dvc
+dvc stage add -n train \
+                -d data/images/ \
+                -d src/train.py \
+                -p model.conv_units \
+                -p train.epochs \
+                -o models/model.h5 \
+                --live metrics \
+                python3 src/train.py
+```
+
+The `params` section defines the parameter dependencies of the `train` stage. By
+default, DVC reads those values (`model.conv_units` and `train.epochs`) from
+`params.yaml` file. You can also set the parameter file name by supplying it
+before the parameter name, like `-p myparams.json:model.units`.
+
+Here is the contents of `params.yaml` file:
+
+```yaml
+train:
+  epochs: 10
+model:
+  conv_units: 16
+```
+
+When you use `--set-param`option for `dvc exp run`, DVC updates these values
+with the values you set in the command line before running the experiment.
+
+Metrics are what you use to evaluate your models. DVC allows any scalar values
+to be used as metrics. It's able to track the metrics we defined in the code
+with the Keras integration introduced recently. Before that, we were using
+`--metrics` and `--metrics-no-cache` options of `dvc stage add` to define
+metrics to DVC, and write the metrics in the code manually. Please see
+`dvc metrics` for this kind of explicitly defined metrics.
+
+</details>
+
+<details>
+
 ### ⚙️ Run multiple experiments in parallel
 
 Instead of running the experiments one-by-one, we can define them to run in a
--- a/content/docs/start/experiments.md
+++ b/content/docs/start/experiments.md
@@ -19,13 +19,38 @@ the [`example-dvc-experiments`][ede] project.
 
 [ede]: https://github.com/iterative/example-dvc-experiments
 
+## Initializing a project into DVC experiments
+
+If you already have a DVC project, that's great. You can start to use `dvc exp`
+commands right away to create experiments in your project. (See the [user's
+guide] for detailed information.) In this section, we'll focus how to structure
+an ML project into a DVC experiments project with `dvc exp init`.
+
+A typical machine learning project has some data, a set of scripts that trains a
+model, a bunch of hyperparameters that modify these models. DVC makes certain
+assumptions about the names of these elements to initialize an experimentation
+project with:
+
+```dvc
+$ dvc exp init python src/train.py
+```
+
+Here, `python src/train.py` describes how our project runs the training script.
+It could be any other command. For DVC, this defines how you run an experiment.
+
+You can also set code (default: `src`), data (`data/`), models (`models/`),
+hyperparameters (`params.yaml`), metrics (`metrics.json`), and plots (`plots/`)
+dependencies if they differ from the defaults.
+
+We run these experiments in [`example-dvc-experiments`][ede] project, and it's
+structured around these defaults so we don't specify any custom paths.
+
 <details>
 
 ### ⚙️ Installing the example project
 
-These commands are run in the [`example-dvc-experiments`][ede] project. You can
-run the commands in this document after cloning the repository, installing the
-requirements, and pulling the data.
+You can run the commands in this document after cloning the repository,
+installing the requirements, and pulling the data.
 
 #### Clone the project and create virtual environment
 
@@ -42,6 +67,8 @@ $ . .venv/bin/activate
 $ python -m pip install -r requirements.txt
 ```
 
+</details>
+
 #### Get the data set
 
 The repository we cloned doesn't contain the dataset. Instead of storing the
@@ -51,11 +78,6 @@ this case, we use `dvc pull` to update the missing data files.
 ```dvc
 $ dvc pull
 ```
-
-The repository already contains the necessary configuration to run the
-experiments.
-
-</details>
 
 Running the experiment with the default project settings requires only the
 command:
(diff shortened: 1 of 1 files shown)


content/docs/start/experiments.md
```diff
@@ -138,6 +160,55 @@ Experiment results have been applied to your workspace.
 
 <details>
 
+### ℹ️ More information about (Hyper)parameters and metrics
```
dberenbaum, line 163: Hopefully, we can condense this section and incorporate it with initializing the project, or directly before or after.

content/docs/start/experiments.md
```diff
@@ -19,13 +19,38 @@ the [`example-dvc-experiments`][ede] project.
...
+guide] for detailed information.) In this section, we'll focus how to structure
+an ML project into a DVC experiments project with `dvc exp init`.
+
+A typical machine learning project has some data, a set of scripts that trains a
+model, a bunch of hyperparameters that modify these models. DVC makes certain
+assumptions about the names of these elements to initialize an experimentation
+project with:
+
+```dvc
+$ dvc exp init python src/train.py
```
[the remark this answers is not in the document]
dberenbaum: Example that was being used for https://github.com/iterative/dvc.org/pull/2979 (it's since been cut down):

```
$ dvc exp init -i
This command will guide you to set up a default stage in dvc.yaml.
See https://dvc.org/doc/user-guide/project-structure/pipelines-files.

DVC assumes the following workspace structure:
├── data
├── metrics.json
├── models
├── params.yaml
├── plots
└── src

Command to execute: python src/train.py
Path to a code file/directory [src, n to omit]: src/train.py
Path to a data file/directory [data, n to omit]: data/images/
Path to a model file/directory [models, n to omit]:
Path to a parameters file [params.yaml, n to omit]:
Path to a metrics file [metrics.json, n to omit]:
Path to a plots file/directory [plots, n to omit]: logs.csv
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
default:
  cmd: python src/train.py
  deps:
  - data/images/
  - src/train.py
  params:
  - model
  - train
  outs:
  - models
  metrics:
  - metrics.json:
      cache: false
  plots:
  - logs.csv:
      cache: false
Do you want to add the above contents to dvc.yaml? [y/n]: y

Created default stage in dvc.yaml. To run, use "dvc exp run".
See https://dvc.org/doc/user-guide/experiment-management/running-experiments.
```

content/docs/start/experiments.md
dberenbaum: The key points to me are that code should read parameters from a file and write metrics to a file. Not sure we need to go deeper than that.

dberenbaum (commented):
Let me know if you think this direction makes sense @iesahin!

iesahin:
I'm copying @jorgeorpinel 's comments in Notion: 

> TBH I'm not sure I love hidden sections for metrics/params. They're usually reserved for off-topic implementation details. Maybe some of the metrics info can be collapsed but a basic description that is relevant to the trail should be in a regular paragraph, I think.

About the sentence starting with "This experiment is then associated with the values in parameters file":

> Parameters mention start all the way up here. If we're going to insert params info after the exp show table, is there any way to avoid the concept before it? Let's try to all of the params info until after the table so it's consolidated and more logical.

iesahin:
It looks we still need to use `dvc init` before `dvc exp init`, right? 

```
dvc exp init python3 src/train.py
ERROR: you are not inside of a DVC repository (checked up to mount point '/home/iex/work')
```

dberenbaum:
> It looks we still need to use `dvc init` before `dvc exp init`, right?
> 
> ```
> dvc exp init python3 src/train.py
> ERROR: you are not inside of a DVC repository (checked up to mount point '/home/iex/work')
> ```
> 
> Is there any work to check whether we're inside a DVC repository and `dvc init` if we're not? @dberenbaum

This has been deprioritized for now, so `dvc init` is still needed.

content/docs/start/experiments.md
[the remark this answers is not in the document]
iesahin: I think `--interactive` deserves a mention, though the screen estate it occupies may be too large.

content/docs/start/experiments.md
```diff
@@ -16,16 +16,49 @@ https://youtu.be/FHQq_zZz5ms
 
 In this section, we explore the basic features of DVC experiment management with
 the [`example-dvc-experiments`][ede] project.
-
 [ede]: https://github.com/iterative/example-dvc-experiments
 
+<details>
```
julieg18, line 21: This change should get the app to build correctly 😃 

```suggestion
<details>
```

iesahin: I looked for the reason of failure for a few ~~hours~~  minutes and decided it's a local hiccup :)) Thank you @julieg18

content/docs/start/experiments.md
```diff
@@ -104,24 +134,6 @@ experiment in a separate row, along with the Git commit IDs they are attached
...
-When you have a large number of experiments that you don't want to commit all to
-Git, it's better to use `dvc exp run`. It allows to change the parameters
-quickly, can track the history of artifacts and has facilities to compare these
-experiments easily.
-
-</details>
-
 Now let's do some more experimentation.
 
 DVC allows to update the parameters defined in the pipeline without modifying
```
jorgeorpinel, line 139: I think it makes sense to remove the `repro` section for sure. But then we go and mention "the pipeline" out of the blue right after. What is a pipeline? My first thought is to also remove that reference but maybe linking to https://dvc.org/doc/start/data-pipelines instead could be an option (since we're in the GS and there's no specific pipelines guide anyway).

iesahin: I replaced two occurrences of "pipeline" to "project."

iesahin:
This one is ready for review @shcheklein @dberenbaum @jorgeorpinel @julieg18

content/docs/start/experiments.md
iesahin: I simplified the section. Divided params and metrics. PTAL.

dberenbaum:
Thanks @iesahin! We might have to pause this effort pending decisions on https://github.com/iterative/dvc/issues/7130 and related issues. It's not worth iterating on this if we have breaking changes to `exp init` soon after. Sorry for the delay.

iesahin:
If there is not a definite time frame for the changes in `dvc exp init` or it'll take, say, more than one month, could we review and merge this to the master? It includes more updates than only `dvc exp init` itself. 

I can revisit the page once more when the changes are implemented. WDYT @dberenbaum @shcheklein ?

jorgeorpinel:
If we're not sure we want `exp init` in the Get Started I don't think it's a good idea to merge it.

> includes more updates than only dvc exp init itself

You could remove the `exp init` changes from this PR then (and rename it?), so we can review/approve/merge those other updates.

jorgeorpinel:
p.s. needs merging master in.

iesahin:
> If we're not sure we want `exp init` in the Get Started I don't think it's a good idea to merge it.

The question is not whether we want it in the GS, but whether `dvc exp init` will stay in its current form. I believe no major changes will be released in Q1, from the shape of discussions in iterative/dvc#7130, it looks like implementing them will take more time than the current version. Instead of keeping the GS docs in its current form, we can review/merge with the current `dvc exp init`, and wait the changes. 

> You could remove the `exp init` changes from this PR then (and rename it?), so we can review/approve/merge those other updates.

That's also possible but that will mean we'll continue to live with the current document structure indefinitely.

dberenbaum:
I'm fine with moving forward as is for now. Looking at the doc, any `exp init` changes are likely to have minimal impact on the doc, and we can make minor changes as needed.

dberenbaum:
> You could remove the `exp init` changes from this PR then (and rename it?), so we can review/approve/merge those other updates.

After doing a quick review, I agree that this is the way to go regardless of `exp init` plans. It looks like there are multiple topics addressed by this PR:
* Replacing the setup with `exp init`.
* Adding hidden sections to introduce the concepts of parameters and metrics.
* Replacing `--include/--exclude` flags with `--drop/--keep` flags.
* Wording changes.

Some of these are going to generate a lot more discussion than others, so it would be helpful to have more than one PR so that simple changes like wording and updating flags can get merged quickly.

iesahin:
Converted to draft. Will work on these in separate branches and this will be closed after other PRs are submitted.

iesahin:
#3112 modified to include `--drop` (but not `--keep`) in `experiments.md` instead of `--include-params` etc. I'll base the modifications on the current master and won't deal with it again.

content/docs/start/experiments.md
```diff
@@ -138,6 +111,34 @@ Experiment results have been applied to your workspace.
 
 <details>
 
+### ℹ️ More information about (Hyper)parameters
```
iesahin, line 114: This section is in #3163

jorgeorpinel: #3163 is merged. Should `master` be merged into this branch so this change goes away, and resolve the comment?

iesahin:
> Interesting method. Similar to a nested PR or PR series but less messy. However, it can also be confusing to review in 2 places.
> 
> What's the exact use of this draft if it's split into other PRs? Should anyone be looking at the changes? What parts)? Thanks

This is for my own tracking. For longer documents, I can rebase time to time to see the remaining parts. For this one, I don't expect any review. It's like a guide to move the document.

iesahin:
This one is completed, now GS:Experiments starts with `dvc exp init.` Thank you for your reviews @jorgeorpinel @dberenbaum @shcheklein

PR #3060, opened 2021-12-02
guide: update Node version (docs contrib)

jorgeorpinel:
`yarn` started failing for me today. It wanted node 14+ I think (I closed the terminal, oops).

### DIFF
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -57,7 +57,7 @@ the website engine itself. Source code and content files need to be properly
 formatted and linted as well, which is also ensured by the full setup below.
 
 Make sure you have [Python](https://www.python.org/downloads/) 3.7+, a recent
-LTS version of [Node.js](https://nodejs.org/en/) (`>=14.0.0`, `<=17.x`), and
+LTS version of [Node.js](https://nodejs.org/en/) (`>=14.0.0`, `<=16.x`), and
 install [Yarn](https://yarnpkg.com/):
 
 > In Windows, you may need to install [Visual Studio Build Tools], and the
--- a/content/docs/user-guide/contributing/docs.md
+++ b/content/docs/user-guide/contributing/docs.md
@@ -57,7 +57,7 @@ the website engine itself. Source code and content files need to be properly
 formatted and linted as well, which is also ensured by the full setup below.
 
 Make sure you have [Python](https://www.python.org/downloads/) 3.7+, a recent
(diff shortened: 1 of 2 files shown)


rogermparent:
Our `package.json` is set for `"node": "<=16.x"`, 16 being the current LTS and 14 being the last one.

jorgeorpinel:
I had 12 (the default from `apt-get` on Ubuntu) which worked up til yesterday for me.

content/docs/user-guide/contributing/docs.md
```diff
@@ -57,7 +57,7 @@ the website engine itself. Source code and content files need to be properly
 formatted and linted as well, which is also ensured by the full setup below.
 
 Make sure you have [Python](https://www.python.org/downloads/) 3.7+, a recent
-LTS version of [Node.js](https://nodejs.org/en/) (`>=12.0.0`, `<=15.x`), and
+LTS version of [Node.js](https://nodejs.org/en/) (`>=14.0.0`, `<=17.x`), and
```
jorgeorpinel, line 60: So is this the correct change then @rogermparent ?
```suggestion
LTS version of [Node.js](https://nodejs.org/en/) (`>=14.0.0`, `<=16.x`), and
```

jorgeorpinel: Yeah I guess this should be it. Committing, please approve/merge.

rogermparent (approved):

rogermparent:
We should also update the `engines.node` field in `package.json`. According to [yarn docs](https://classic.yarnpkg.com/en/docs/package-json#toc-engines), our value should be `">=14.x <=16.x"`.

skshetry:
> We should also update the `engines.node` field in `package.json`. According to [yarn docs](https://classic.yarnpkg.com/en/docs/package-json#toc-engines), our value should be `">=14.x <=16.x"`.

I just tried yesterday with node 17 and node 16. Both failed while installing/compiling `sharp`.

jorgeorpinel:
> We should also update the engines.node field in package.json

Done @rogermparent . I removed node_modules and ran `yarn` again successfully. PTAL

> tried yesterday with node 17 and node 16. Both failed

@skshetry does that relate to `engines.node`? Can you provide more context? I'm on WSL2/Ubuntu with Node 16 and this currently works fine.

rogermparent:
> I just tried yesterday with node 17 and node 16. Both failed while installing/compiling `sharp`.

Works for me on node v16.11.1, 17 should be rejected by yarn with the message "Found incompatible module." (though it also works for me if forced by changing the `engines` versions).

As far as this PR's changes, they look fine to me.

rogermparent (approved):

PR #3062, opened 2021-12-02
ref: review cache/etc. "corruption" concept...

jorgeorpinel:
Mainly about cache <-> dvc.lock links

Currently we are a bit alarmist about the possibility of corruption. DVC's built-in protection mechanism make this difficult.

### DIFF
--- a/content/docs/command-reference/config.md
+++ b/content/docs/command-reference/config.md
@@ -158,25 +158,21 @@ This section contains the following options, which affect the project's
   > directory into paths relative to the config file location.
 
 - `cache.type` - link type that DVC should use to link data files from cache to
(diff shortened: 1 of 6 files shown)


jorgeorpinel:
Ping @efiop any change you can take a look here? Thanks

shcheklein:
@efiop @jorgeorpinel what do we do here folks? how can we move these things faster?

jorgeorpinel:
@shcheklein actually this comes from a suggestion you left somewhere (sorry, I can't find the ref. now) so feel free to review. I thought @efiop would be best equipped to confirm this info. but have had no response.

shcheklein (approved):

PR #3065, opened 2021-12-07
ref: state exp names should be unique

jorgeorpinel:
rel. https://github.com/iterative/dvc/pull/6848


jorgeorpinel:
Cc @iesahin do you think we should mention in the guide or other refs as well? Or GS

iesahin:
Can we have an experiment naming guide somewhere? :)

karajan1001 (approved):

shcheklein (approved):

jorgeorpinel:
> experiment naming guide somewhere

What info would you think people need guidance for @iesahin ?

See also https://github.com/iterative/dvc.org/pull/2909#discussion_r760506777

iesahin:
I think experiment names are a minor point and the content we have already covers it, but it began to feel we better have a section in the overview for them as bullet points. Something like:

* Experiments get their names from the content of the pipeline, unless specified by the `--name` option. 
* Queued experiments have a names before they run, and they get another name after running. 
* Experiment names should be unique. 
* Experiment names cannot contain `/` 

etc.

jorgeorpinel:
@iesahin now there are some exp naming details hidden in https://dvc.org/doc/user-guide/experiment-management/experiments-overview, including that exp names should be unique and where they come from

image

It doesn't cover:

> - Experiments get their names from the content of the pipeline

Not sure what you mean by that though. The default ones are actually derived from the custom ref commit hash.

> Queued experiments have a names before they run

We don't have this info anywhere rn. Do you want to add it to the hidden section in https://dvc.org/doc/user-guide/experiment-management/running-experiments#the-experiments-queue for now?

> Experiment names cannot contain /

Yeah the naming rules should be the same as https://git-scm.com/docs/git-check-ref-format. Feel free to include that in the existing passage(s) as well!

---

The larger question is whether this info deserves to be unhidden and/or centralized. And if so where

shcheklein:
I think it can be a separate section - `Naming experiments` with some details around this and where those names could be used.  I would even make it a top level in the UG. Even it's simple (better for it to be simple).

Issue #2911, opened 2021-10-10
guide: Experiment Management (phase 2)


jorgeorpinel:
> Extracted from https://github.com/iterative/dvc.org/issues/2548

- [ ] & #2768
- [ ] & #2313 (could be included above)
- [ ] #2737
- [ ] #2736
- [ ] https://github.com/iterative/dvc.org/issues/2844 (my be done by the 2 above)
- [ ] #2777 
- [ ] #2769

PR #3071, opened 2021-12-07
ref: `exp init` improvements

jorgeorpinel:
> Follow-up to #3015

- [x] Move before `exp run` in nav
- [x] Copy edits/ Improve explanations
- [x] Address ~~https://github.com/iterative/dvc.org/pull/3015#pullrequestreview-804140978 if still needed~~ https://github.com/iterative/dvc.org/pull/3071#pullrequestreview-825977103.
- ~~Full [Configuration](https://github.com/iterative/dvc.org/pull/3015#discussion_r748627442) section~~ Linked to https://dvc.org/doc/command-reference/config#exp + https://github.com/iterative/dvc.org/pull/3098

---

- [ ] Add **Examples**?
- [ ] Keep `command` section or move to **Options**?
  See https://github.com/iterative/dvc.org/pull/3071#pullrequestreview-829161122
- [ ] Update DVC core with changes to `exp (init)` help output decided here.
- Extract details to new Exps guide? Review this [possible structure](https://github.com/iterative/dvc.org/pull/3015#pullrequestreview-804151331) -- May be part of https://github.com/iterative/dvc.org/issues/2911


jorgeorpinel:
My intended changes are not nearly done (see checkboxes) but if there's approval we can merge the improvements so far, so feel free to review @dberenbaum @shcheklein . Thanks

content/docs/command-reference/exp/init.md
```diff
@@ -1,6 +1,7 @@
 # exp init
 
-Codify project using [DVC metafiles](/doc/user-guide/project-structure) to run
```
shcheklein, line 3: sorry folks for going into review on top of review, it's way too complicated to my mind :)

(this change itself is not substantial enough to be honest)

Codify ... DVC metafiles .... experiments ... an existing project - I'm afraid that people won't understand this. And this is part of the happy path.

What does this command essentially do?

`The fastest way to initialize needed stuff in a project for DVC to be able to run experiments`

why don't we write like this? the current summary should be part of the description (but description starts nicely now, so I would not touch it)

content/docs/command-reference/exp/init.md
```diff
@@ -17,37 +18,52 @@ usage: dvc exp init [-h] [-q | -v] [--run] [--interactive] [-f]
...
+```
+├── data/
+├── metrics.json
+├── models/
+├── params.yaml  # required
+├── plots/
+└── src/
+```
+
+> Note that `params.yaml` is the only required file (see `dvc params`).
```
shcheklein, line 36: side note ... for the sake of my education: why is it required?

content/docs/command-reference/exp/init.md
```diff
@@ -17,37 +18,52 @@ usage: dvc exp init [-h] [-q | -v] [--run] [--interactive] [-f]
...
+
+To use this feature, provide a `command` argument or use the `--interactive`
```
shcheklein, line 38: `this feature` refers to?

content/docs/command-reference/exp/init.md
```diff
@@ -56,10 +72,13 @@ $ dvc exp init './another_script.sh $MYENVVAR'
 
 ## Options
 
-- `-i`, `--interactive` - prompts user for the command to execute and different
-  paths for tracking outputs and dependencies, unless they are provided through
-  arguments explicitly. Interactive mode allows users to set those locations
-  from default values or omit them.
+- `-i`, `--interactive` - prompts user for the `command` to execute and for the
```
shcheklein, line 75: to execute -> experiment `command` or something. otherwise it's not clear what command

content/docs/command-reference/exp/init.md
dberenbaum: To parse the parameters and put them into `dvc.yaml`.

content/docs/command-reference/exp/init.md
dberenbaum: Note that any changes should also apply to docs/command-reference/exp/index.md.

jorgeorpinel: > going into review on top of review

No problem, that's kind of the intention of this PR 🙂

content/docs/command-reference/exp/init.md
jorgeorpinel: Related to https://github.com/iterative/dvc/issues/6446#issuecomment-988284225

content/docs/command-reference/exp/init.md
```diff
@@ -17,37 +18,52 @@ usage: dvc exp init [-h] [-q | -v] [--run] [--interactive] [-f]
...
-This command is intended to be a quick way to start running experiments. To
-create more complex stages and pipeliens, use `dvc stage add`.
+> This command is intended as a quick way to start running experiments. To
+> codify complex data pipelines, see the `dvc.yaml` specification.
+
+[checkpoints]: /doc/user-guide/experiment-management/checkpoints
 
 ### The `command` argument
 
-The `command` argument is optional, if you are using `--interactive` mode. The
```
jorgeorpinel, line 37: BTW a question I have is whether to keep this section or move `command` under **Options** like we did for `targets` in https://dvc.org/doc/command-reference/repro#options. I personally like the section more but I remember we discussed with (cc @shcheklein) this and using Options was picked so for consistency I'd move this to Options as well.

> From https://github.com/iterative/dvc.org/pull/3015#pullrequestreview-804146896

content/docs/command-reference/exp/init.md
dberenbaum: It's also related to https://github.com/iterative/dvc/issues/6605. The `params` section of `dvc.yaml` currently expects keys from a file (see below), not a filename itself, so all the parameters need to be defined before creating the stage. If we want to stop requiring this file at the time of `dvc exp init`, we need to change `params` to accept filenames.

```
stages:
  train:
    cmd: python train.py
    deps:
    - data
    - src
    params:
    - params.py:
      - batch_size
      - data_path
      - epochs
      - latent_dim
      - num_samples
```

jorgeorpinel: What about generating a dvc.yaml without `params` if there's no params.yaml file? This is a core discussion though!

dberenbaum: Sorry, I misunderstood you, and now I think it's actually more of a docs issue 🤔 . It is possible to use `dvc exp init` without `params`. You can either use `--explicit` and not provide `--params`, or you can use `dvc exp init -i` and type `n` at the params prompt.

The note here is to indicate that if you do use either the default or some other `params` path, then that path must exist and be parseable to extract the parameters for the stage. So the current note probably needs to be reworded.

content/docs/command-reference/exp/init.md
jorgeorpinel: > What does this command essentially do?

How about `Quickly setup any project to use [DVC Experiments].`?

content/docs/command-reference/exp/init.md
jorgeorpinel: UPDATE: Looks like the discussion about empty params.yaml files is in a few core repo tickets now (e.g. https://github.com/iterative/dvc/issues/7138) so I won't keep discussing the current functionality here (still a bit confused though).

> It is possible to use dvc exp init without params. You can either use --explicit and not provide --params, or you can use dvc exp init -i and type n at the params prompt...
So the current note probably needs to be reworded.

Clarified in 2273431a.

content/docs/command-reference/exp/init.md
jorgeorpinel: "this feature" referred to `exp init`. I didn't want to say "this command" because `exp init` accepts a `command` argument so it's confusing (same happens in `run` and `stage add`). Anyway, I'm just going to be explicit every time instead. Solved

content/docs/command-reference/exp/init.md
jorgeorpinel: Done in 3d916568.

content/docs/command-reference/exp/init.md
```diff
@@ -19,43 +18,57 @@ usage: dvc exp init [-h] [-q | -v] [--run] [--interactive] [-f]
...
+├── data/
+├── metrics.json
+├── models/
+├── params.yaml  # required by default
+├── plots/
+└── src/
+```
+
+> Note that by default `dvc exp init` expects at least a `params.yaml` file
+> present. It can be avoided via the `--params`, `-i`, or `--explicit` options.
```
dberenbaum, line 40: I like the explanation under `--params`, and I'm not sure this explains enough about how to solve the issue. Maybe something like:

Note that by default `dvc exp init` expects at least a `params.yaml` file present. This is because DVC reads that file to find parameters to include in the stage. Modify the parameters file path via `--params`, omit all default paths via `--explicit`, or use `-i` to configure via prompts.

content/docs/command-reference/exp/init.md
```diff
@@ -19,43 +18,57 @@ usage: dvc exp init [-h] [-q | -v] [--run] [--interactive] [-f]
 
 ## Description
```
dberenbaum, line 19: Somewhere in the description, it might be useful to explain that `dvc exp init` by default expects that input data, parameters, and source code paths exist before running an experiment, and that the command is expected to generate models, metrics, and plots.

dberenbaum (changes requested):
Looks good! A couple small notes to help clarify for users what the expectations are for the different paths.

content/docs/command-reference/exp/init.md
shcheklein: > Quickly setup any project to use [DVC Experiments]

sounds good to me!

shcheklein:
looks good to me (I didn't check the details though, since Dave've done this already)

Do we want to add an example? (may be even animation?)

content/docs/command-reference/exp/init.md
jorgeorpinel: Updated in 600e83ca.

content/docs/command-reference/exp/init.md
jorgeorpinel: That's more or a problem for `exp run` though @dberenbaum right (already linked form `--run` option). But should we say that `exp run` (and `repro`) for that matter expect that the stage definition and code are valid and working (resp.)? Ideally it should be evident IMO.

jorgeorpinel:
> Do we want to add an example?

Yep @shcheklein it's a check box I didn't get to yet ⌛ (could be a separate PR) 

> may be even animation?

Good idea in general to consider gifs but in this case I think a static console sample with the entire `-i` output would be better so you can read them calmly. Animations would be good for samples that focus on progress, mutation, etc. Let's keep it in mind.

jorgeorpinel:
UPDATE: I added a simple example just to give a sense of the interactive-mode UI. Not sure which features we want to highlight with examples otherwise.

content/docs/command-reference/exp/init.md
dberenbaum: > That's more or a problem for `exp run` though

Well that's the point of `exp init`, right? Better to have users understand what's needed up front than to have them run `exp init` only to fail on `exp run`.

Doesn't need to be part of this PR. It could also be handled by some of the suggested changes to the core command rather than the docs.

content/docs/command-reference/exp/init.md
```diff
@@ -64,71 +80,123 @@ $ dvc exp init './another_script.sh $MYENVVAR'
...
 - `-q`, `--quiet` - do not write anything to standard output. Exit with 0 if no
   problems arise, otherwise 1.
 
 - `-v`, `--verbose` - displays detailed tracing information.
+
+## Example: interactive mode
+
+Let's prepare a raw data ingestion script to start running experiments on it.
+
+```dvc
```
dberenbaum, line 155: Why not a model training example? Data ingestion is a nice application to show how this can be useful even for non-training stages, but model training is the primary use case. It's unlikely for someone to run experiments on a data ingestion script.

content/docs/command-reference/exp/init.md
```diff
@@ -64,71 +80,123 @@ $ dvc exp init './another_script.sh $MYENVVAR'
...
+Path to a plots file/directory [plots, n to omit]: n
+...
+```
+
+Notice that the code and data locations were specified above, to avoid using the
+defaults which are too broad. `params.yaml` exists in the workspace and is used
+by our code in question, so the default is used. Models, metrics, and plots are
+omitted as there are none for an ingestion process.
+
+The resulting `dvc.yaml` file is simply:
```
dberenbaum, line 185: Might be worth mentioning pipelines briefly to explain what's in `dvc.yaml`.

dberenbaum (commented):
Looks good. Adding an example is great, but why not focus on a model training example?

jorgeorpinel:
> why not focus on a model training example?

OK @dberenbaum . Done in 56f4a467.

content/docs/command-reference/exp/init.md
jorgeorpinel: > Better to have users understand what's needed up front than to have them run exp init only to fail on exp run.

I'm not sure. If users need to read that on the cmd ref., is `exp init` serving as a self-explanatory command? I think those notes should be in the command's output fist (and added here as a result of that change).

content/docs/command-reference/exp/init.md
jorgeorpinel: The data ingestion example produced a very small dvc.yaml file and then I had that note about using `stage add` to produce the same. Idk, I thought it was interesting. Anyway, I agree the training scenario is more applicable so 
I replaced all that now.

content/docs/command-reference/exp/init.md
jorgeorpinel: I'm actively avoiding mixing in that concept here but it is mentioned strategically once in the Description. I did elaborate on this sentence a bit though.

dberenbaum (approved):
Nice!

jorgeorpinel:
Thanks!

jorgeorpinel:
Looks like the [failing link check](https://github.com/iterative/dvc.org/pull/3071/checks?check_run_id=4611926372) is a false positive. I'll see if this is already reported to the link checker repo. Merging 🔀

content/docs/command-reference/exp/index.md
```diff
@@ -26,19 +26,19 @@ usage: dvc exp [-h] [-q | -v]
...
                         repository.
-    run                 Reproduce complete or partial experiment pipelines.
-    gc                  Garbage collect unneeded experiments.
-    branch              Promote an experiment to a Git branch.
     list                List local and remote experiments.
+    apply               Apply the changes from an experiment to your
+                        workspace.
+    branch              Promote an experiment to a Git branch.
+    remove              Remove local experiments.
+    gc                  Garbage collect unneeded experiments.
```
jorgeorpinel, line 39: Question @dberenbaum @skshetry : did we update the core repo with these texts? Thanks

jorgeorpinel: Wait... There are no changes to the text here ⌛

content/docs/command-reference/exp/init.md
```diff
@@ -1,7 +1,6 @@
 # exp init
 
-Codify project using [DVC metafiles](/doc/user-guide/project-structure) to run
-[experiments](/doc/user-guide/experiment-management).
+Quickly setup any project to use [DVC Experiments].
```
jorgeorpinel, line 3: I meant this (per last check box in this PR desc.) @dberenbaum @skshetry . Currently `dvc exp init -h` prints `Initialize DVC in the current directory. Expects directory to be a Git repository...` Do we want to update that text?

dberenbaum: I think that's the `dvc init -h` text 😄 . `dvc exp init -h` says `Initialize DVC experiments`. However, I think we agreed this new language is better, so I'm happy to put in a PR to update in the core repo.

jorgeorpinel: > I think that's the dvc init -h text 😄

yep 🤦 sry. OK, thanks for the PR

PR #3077, opened 2021-12-09
updated syntax for dvc import example

oslobowl:
From https://discord.com/channels/485586884165107732/485596304961962003/918463103560196116

> Question: I'm reading the documentation and how to use dvc import properly ... under [Importing and updating fixed revisions](https://dvc.org/doc/command-reference/import#example-importing-and-updating-fixed-revisions) that when you have used `dvc import --rev` with a git hash and we want to force-update it to a different commit, we might use `dvc update --rev`, and the example given is: `dvc update --rev cats-dogs-v2`.
Should it not be `dvc update --rev cats-dogs-v2 <stage>.dvc`, though?

> seems like you found a bug in our docs. You are correct that there should be <stage>.dvc in the end of this line.

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -177,20 +177,12 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
 under `repo`:
 
 ```yaml
-md5: 7ff366c716a376ec009054f1c141dc17
-frozen: true
 deps:
-- path: use-cases/cats-dogs
-  repo:
-    url: git@github.com:iterative/dataset-registry.git
-    rev: cats-dogs-v1
-    ^^^^^^^^^^^^^^^^^
-    rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
-outs:
-- md5: b6923e1e4ad16ea1a7e2b328842d56a2.dir
-  size: 41149064
-  nfiles: 1800
-  path: cats-dogs
+  - path: use-cases/cats-dogs
+    repo:
+      url: git@github.com:iterative/dataset-registry.git
+      rev: cats-dogs-v1
+      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
 ```
 
 If `rev` is a Git branch or tag (where the underlying commit changes), the data
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -193,7 +193,7 @@ will not have an effect on the import `.dvc` file. You may force-update it to a
 different commit with `dvc update --rev`:
 
 ```dvc
-$ dvc update --rev cats-dogs-v2 <stage>.dvc
+$ dvc update --rev cats-dogs-v2 cats-dogs.dvc
 ```
 
 > In the above example, the value for `rev` in the new `.dvc` file will be
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -177,12 +177,20 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
 under `repo`:
 
 ```yaml
+md5: 7ff366c716a376ec009054f1c141dc17
+frozen: true
 deps:
-  - path: data/data.xml
-    repo:
-      url: git@github.com:iterative/dataset-registry.git
-      rev: cats-dogs-v1
-      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
+- path: use-cases/cats-dogs
+  repo:
+    url: git@github.com:iterative/dataset-registry.git
+    rev: cats-dogs-v1
+    ^^^^^^^^^^^^^^^^^
+    rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
+outs:
+- md5: b6923e1e4ad16ea1a7e2b328842d56a2.dir
+  size: 41149064
+  nfiles: 1800
+  path: cats-dogs
 ```
 
 If `rev` is a Git branch or tag (where the underlying commit changes), the data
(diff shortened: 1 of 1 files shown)


pared:
The discussion:
https://discord.com/channels/485586884165107732/485596304961962003/918463103560196116

Hmm, now that I inspect the whole paragraph closely there seems to be more issues with it.

1. We import `cats-dogs` but the sample `.yaml` content is for `data.xml`:
Import:

```
dvc import --rev cats-dogs-v1 \
             git@github.com:iterative/dataset-registry.git \
             use-cases/cats-dogs
```

`yaml` content: 

```
deps:
  - path: data/data.xml
    repo:
      url: git@github.com:iterative/dataset-registry.git
      rev: cats-dogs-v1
      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
```

So, What I think we should do is:

1. update the `yaml` content to, so that it reflects actual result from the import statement.
2. update the `dvc update` call to target `cats-dogs.dvc`: `dvc update --rev cats-dogs-v2 cats-dogs.dvc`

@jorgeorpinel can you confirm that discrepancy between `cats-dogs` and `data.xml` is not intentional?

iesahin (commented):
Thank you for the contribution @oslobowl 🙏🏼

iesahin:
I've modified the `.yaml` file content, but didn't update the `dvc import` call. PTAL and pull. @pared @oslobowl @jorgeorpinel

iesahin (approved):

content/docs/command-reference/import.md
```diff
@@ -177,12 +177,20 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
...
-  - path: data/data.xml
-    repo:
-      url: git@github.com:iterative/dataset-registry.git
-      rev: cats-dogs-v1
-      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
+- path: use-cases/cats-dogs
+  repo:
+    url: git@github.com:iterative/dataset-registry.git
+    rev: cats-dogs-v1
+    ^^^^^^^^^^^^^^^^^
```
shcheklein, line 187: it's not a valid YAML syntax anymore, it might confuse people I think, Emre ... if we want to highlight the diff - we can do diff as we do in some other places, wdyt?

content/docs/command-reference/import.md
```diff
@@ -177,12 +177,20 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
 under `repo`:
 
 ```yaml
+md5: 7ff366c716a376ec009054f1c141dc17
+frozen: true
```
jorgeorpinel, line 181: Thanks but we didn't include this on purpose. Keeping snippets short
```suggestion
```

content/docs/command-reference/import.md
```diff
@@ -177,12 +177,20 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
...
+  repo:
+    url: git@github.com:iterative/dataset-registry.git
+    rev: cats-dogs-v1
+    ^^^^^^^^^^^^^^^^^
+    rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
+outs:
+- md5: b6923e1e4ad16ea1a7e2b328842d56a2.dir
+  size: 41149064
+  nfiles: 1800
+  path: cats-dogs
```
jorgeorpinel, line 193: Same. No need to include the whole file, only the relevant section
```suggestion
```

content/docs/command-reference/import.md
```diff
@@ -177,12 +177,20 @@ When using this option, the import `.dvc` file will also have a `rev` subfield
...
-    repo:
-      url: git@github.com:iterative/dataset-registry.git
-      rev: cats-dogs-v1
-      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
+- path: use-cases/cats-dogs
+  repo:
+    url: git@github.com:iterative/dataset-registry.git
+    rev: cats-dogs-v1
+    ^^^^^^^^^^^^^^^^^
+    rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
```
jorgeorpinel, line 188: Fixes format (re https://github.com/iterative/dvc.org/pull/3077#pullrequestreview-831921076)
```suggestion
  - path: use-cases/cats-dogs
    repo:
      url: git@github.com:iterative/dataset-registry.git
      rev: cats-dogs-v1
      rev_lock: 0547f5883fb18e523e35578e2f0d19648c8f2d5c
```

jorgeorpinel (commented):
I'll commit these:

content/docs/command-reference/import.md
```diff
@@ -193,7 +201,7 @@ will not have an effect on the import `.dvc` file. You may force-update it to a
 different commit with `dvc update --rev`:
 
 ```dvc
-$ dvc update --rev cats-dogs-v2
+$ dvc update --rev cats-dogs-v2 <stage>.dvc
```
jorgeorpinel, line 196: We should be specific here I think? It's an example
```suggestion
$ dvc update --rev cats-dogs-v2 cats-dogs.dvc
```
Committing...

content/docs/command-reference/import.md
iesahin: It might be good, but using diff also doesn't lead to proper YAML. 

It looks it's already merged, but I can replace this with a `git` +/- example. DYT it will look better?

shcheklein: may be :) I'm missing some context I think ... what is the reason to include it in the first place, btw? Looks like this current fix is enough, no?

iesahin: The surrounding text mentions how `--rev` changes the `.dvc` file contents, but the content itself is long to see. I've added `^^^` under `rev:` because of this. I don't think the user is expecting a valid YAML at this point, it's obvious that we're making them to notify the difference.

jorgeorpinel: I see what you were going for @iesahin but currently the YAML block is shown before using the `update` cmd so it wouldn't make sense to have a git-style diff here unless you rearrange that somehow. Feel free to open a PR suggestion. Thanks

PR #3079, opened 2021-12-10
fix(seo): prevent expensive calculations on every render

jdnichollsc:


src/components/SEO/helper.ts
```diff
@@ -0,0 +1,82 @@
...
+      content: description
+    },
+    {
+      name: 'twitter:description',
+      content: description
+    }
+  )
+}
+
+export const buildMetaImage = (
```
julieg18, line 38: I don't think it's the best practice to have these functions directly mutate `prebuiltMeta`. Maybe just have functions like `buildMetaImage` just return the meta data itself and keep the `prebuildMeta` inside of `buildMetadata`?

Example:
```
// etc...

export const getMetaImage = (
  imageUrl: string,
  imageAlt?: string,
  imageData?: IGatsbyImageData
): Array<MetaProps> => {
  return [
    { property: 'og:image', content: imageUrl },
    { name: 'og:image:alt', content: imageAlt },
    { property: 'og:image:width', content: String(imageData?.width) },
    { property: 'og:image:height', content: String(imageData?.height) },
    { name: 'twitter:image', content: imageUrl },
    { name: 'twitter:image:alt', content: imageAlt }
  ]
}

export const buildMetadata = (
  siteUrl: string,
  title?: string,
  defaultMetaTitle?: boolean,
  description?: string,
  keywords?: string,
  image?: IGatsbyImageData,
  imageAlt?: string
) => {
  const prebuildMeta: MetaProps[] = []

  // etc...

  if (image) {
    const imageUrl = siteUrl + getSrc(image)
    const imageData = getImage(image)
    prebuildMeta.push(...getMetaImage(imageUrl, imageAlt, imageData))
  }
  return prebuildMeta
}

```

julieg18 (commented):
Nice work! Not only does this cause less rerendering, it also makes the code inside `SEO` more readable. Minus one minor thing, I don't see any issues!

rogermparent (commented):
Great idea to memoize all the SEO logic! It may be possible to do a bit more granular memoization with a bit of refactoring on how we use `react-helmet`, but this simple change alone definitely stops the whole SEO generation process from happening on every render.

That said, after running the profiler on this PR compared to master, the difference in render times for the SEO component in particular are negligible.

Before memoization:
before memoization screenshot

After memoization:
after memoization screenshot

Were you able to get different profiling results? I could be missing something.

src/components/SEO/helper.ts
jdnichollsc: I totally agree with you! using functional programming approach looks much better for this scenario to avoid mutating the array or returning a new one <3

jdnichollsc:
@rogermparent great question, maybe for now these changes are negligible, but for the future, having a lot of pages or components using this logic, you will want to avoid the same calculations in every render, right now the project is so simple in that aspect :)

rogermparent:
> @rogermparent great question, maybe for now these changes are negligible, but for the future, having a lot of pages or components using this logic, you will want to avoid the same calculations in every render, right now the project is so simple in that aspect :)

Agreed! And even though we don't see many benefits now, we also don't see any drawbacks, and a memoized approach almost certainly scales better. I agree with @julieg18's comment, but after that's taken care of I see no issues with merging this in.

Thanks for the contribution and the good discussion!

jdnichollsc:
Thanks for your comments folks, I just updated the PR with the changes requested by @julieg18! <3

julieg18 (approved):
Looks great!

PR #3091, opened 2021-12-13
ref: `exp show` Use dvctable

daavoo:


jorgeorpinel (changes requested):
Thanks @daavoo but it also needs coloring codes otherwise it just looks like 2 split blocks:

image

You can copy the codes from existing table examples 🙂

daavoo:
> Thanks @daavoo but it also needs coloring codes otherwise it just looks like 2 split blocks:
> 
> image
> 
> You can copy the codes from existing table examples 🙂

I have just discovered that when checking the deployment 😅

jorgeorpinel (approved):
Thanks!

PR #3093, opened 2021-12-13
Add nested md tabs

julieg18:
* Since this is a copy of the pull request on cml.dev, I didn't add test tabs (it worked fine locally). But if we want to be sure, I'll add them.


jorgeorpinel (approved):
No worries. When we start using tabs, if there's any problem we'll report it. Thanks for porting this!

PR #3101, opened 2021-12-16
ref: Add `plots.auto_open` in `dvc.config`

daavoo:
P.R. for https://github.com/iterative/dvc/pull/7159

PR #3105, opened 2021-12-16
Parallel coordinates plot

daavoo:
For https://github.com/iterative/dvc/pull/6933

- [ ] Include https://github.com/iterative/dvc.org/pull/3101


content/docs/command-reference/exp/show.md
```diff
@@ -127,6 +129,14 @@ metric or param.
 
 - `--md` - prints the command's output in Markdown table format.
 
+- `--html` - generates an interactive parallel coordinates plot from the table.
```
jorgeorpinel, line 132: Can we link somewhere that explains what a PCP is perhaps? Or even better, briefly explain it here... In fact ideally we should do so as a new passage in the **Description** (that mentions this option).

jorgeorpinel: p.s. unless we split this feature to a separate command per https://github.com/iterative/dvc/pull/6933#issuecomment-973159973

content/docs/command-reference/exp/show.md
```diff
@@ -127,6 +129,14 @@ metric or param.
...
+
+- `-o <folder>, --out <folder>` - destination folder of the `html` plot. By
+  default, `dvc_plots`.
```
jorgeorpinel, line 135: > Question about default value in https://github.com/iterative/dvc/pull/6933#issuecomment-996068883

jorgeorpinel: If kept here, it should state it has no effect without `--html`.

content/docs/user-guide/experiment-management/comparing-experiments.md
```diff
@@ -348,6 +348,20 @@ $ dvc exp show --csv | csvstat
 
 [csvkit]: https://csvkit.readthedocs.io/en/latest/
 
+## Show the experiments table as a Parallel Coordinates plot
+
+`dvc exp show` can also generate an interactive Parallel Coordinates plot, with
```
jorgeorpinel, line 353: Hmmm. How is a table anymore? 🙂

content/docs/command-reference/exp/show.md
jorgeorpinel: If we keep this feature here (included in `exp show`) we need to redefine the entire `exp show` description. It's no longer just "prints a customizable table of experiments, their metrics and parameters." PCP is a plot (not a table) and only includes metrics right? Or can it also include params

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +245,32 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
 └───────────────────────┴──────────────┴─────────┴────────────────────────┴──────────────────┘
 ```
 
-Note that in the final example, Git commits remain in chronological order. The
+Note that in the last example, Git commits remain in chronological order. The
```
jorgeorpinel, line 248:
```suggestion
Note that in the previous example, Git commits remain in chronological order. The
```

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +245,32 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
...
 sorting only applies to experiment groups (sharing a parent commit).
 
+To generate an interactive parallel coordinates plot from the table:
+
+```dvc
+$ dvc exp show --html
```
jorgeorpinel, line 254: Let's give this example a separate header?

````suggestion
## Example: Parallel coordinates plot

To generate an interactive parallel coordinates plot based on the experiments
and their parameters:

```dvc
$ dvc exp show --html
````

Comment (jorgeorpinel) [reply content/docs/command-reference/exp/show.md:254]: But again, definition/description of what a PCP is is still missing somewhere.

File: content/docs/command-reference/exp/show.md (hunk 1 above)
Comment (daavoo) [reply content/docs/command-reference/exp/show.md:132]: The PCP contains the same data that it's in the columns of the table. To reduce noise, we always apply a filter to drop columns that have the same value for all rows. But if a `param` has different values per row, it will be shown in the plot.

File: content/docs/user-guide/experiment-management/comparing-experiments.md (hunk 3 above)
Comment (daavoo) [reply content/docs/user-guide/experiment-management/comparing-experiments.md:353]: Indeed, it's not a table anymore, but it does contain the same data as the table. 

So, it's a different representation of the same data. 

`--json` output is neither a table and it's named `Get experiments table in JSON`

File: content/docs/command-reference/exp/show.md (hunk 5 above)
Comment (daavoo) [reply content/docs/command-reference/exp/show.md:254]: Do you think that it should be a link or we should write our own explanation of PCP ?

File: content/docs/user-guide/experiment-management/comparing-experiments.md (hunk 3 above)
Comment (iesahin) [reply content/docs/user-guide/experiment-management/comparing-experiments.md:353]: "Experiments table" refers to the output of `dvc exp show` in default format. When you supply `--json`, you get the content of this table in JSON format.

Here we can call "Show experiment results as a PCP" not to confuse.

File: content/docs/user-guide/experiment-management/comparing-experiments.md [hunk 6]
```diff
@@ -348,6 +348,20 @@ $ dvc exp show --csv | csvstat
...
+`--html`.
+
+The `--html` option can be combined with the other flags of `dvc exp show`. For
+example, the flag `--sort-by` would set the color of the lines.
+
```
iesahin, line 358: I think instead of "sets the color", we should be saying, "the intensity of the color is determined by the value of the column you select with `--sort-by`." The user is not selecting a color, they select a column to highlight the results.

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +245,34 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
...
+
+To generate an interactive PCP based on the experiments and their parameters:
+
+```dvc
+$ dvc exp show --html
+```
+
+![](/img/ref_parallel_coordinates.png) _Parallel Coordinates Plot_
+
+Use `--sort-by` to customize the colors of the lines:
```
iesahin, line 261: Instead of customize, we should be saying something like "the intensity of the background color is determined by the value of the column you select with `--sort-by`."

content/docs/command-reference/exp/show.md
```diff
@@ -109,7 +109,9 @@ metric or param.
 - `--sort-by <name>` - sort experiments by the specified metric or param
   (`name`). Only one visible column (either metric or param) can be used for
   sorting. This only affects the ordering of experiments derived from the same
-  parent commit. Parent commits are always sorted chronologically.
+  parent commit. Parent commits are always sorted chronologically. When combined
+  with the `--html` flag, `name` will be used to set the color of the lines in
```
iesahin, line 113: This "sets the color" expression is a bit confusing. Could you clarify how the color is determined?

iesahin (changes requested):

content/docs/user-guide/experiment-management/comparing-experiments.md
daavoo: I'm ok with any title. Just wanted to point out that the content of the PCP plot is the same as the `exp show` table, just with a different representation. 

On the contrary, the `--json` output contains different content to the `exp show` table but it's treated similar to other output types (`--csv`).

The `--json` output is dumped way before we create and apply flags to the table and it's content is very different:

https://github.com/iterative/dvc/blob/main/dvc/command/experiments/show.py#L543-L573

shcheklein: Is it a discussion on the title `Show the experiments table as a Parallel Coordinates plot` or something else? Title sounds fine to me personally (even though I still think that naming option `--html` can lead to confusion now and some potential renamings in the future).

daavoo: Originated with the title but I guess that it is going a little in circles and it's really about https://github.com/iterative/dvc/issues/7165

iesahin: Technically it may be the other way around, `dvc exp show` may be the beautified, filtered and tabulated version of `--json` output but from the user POV, `--json` contains a similar data. If you're fine with "experiments table", that also feels fine with me.

content/docs/command-reference/exp/show.md
iesahin: Could we have a L4 title that says `PCP options` and lists these two together?

dberenbaum:
FYI, we discussed in refinement today that it's probably better to make this a separate subcommand, like `dvc exp plots`, `dvc exp report`, or similar. It gives flexibility to add more to the HTML and change the options, and it promotes it more than another option in `dvc exp show`.

Although TBH I don't feel too strongly either way. If I'm already looking at the table and decide it would be easier to view it as a PCP, does it make sense to have to use a different command?

content/docs/command-reference/exp/show.md
daavoo: Not sure about adding a title, it might look like the other options don't have effect on the plot (but some do)

content/docs/command-reference/exp/show.md
```diff
@@ -53,6 +53,20 @@ sorted inside each group, chronologically by default. The `--sort-by` and
 `--sort-order` options can change this ordering, based on any single, visible
 metric or param.
 
+### Parallel Coordinates Plot
+
+When the `--pcp` option is passed, an interactive Parallel Coordinates Plot will
+be generated using the data from the table.
+
+This plot is usefull to exploring the relationships between the metrics _and_
```
dberenbaum, line 61:
```suggestion
This plot is useful to explore the relationships between the metrics _and_
```

content/docs/user-guide/experiment-management/comparing-experiments.md
```diff
@@ -348,6 +348,22 @@ $ dvc exp show --csv | csvstat
 
 [csvkit]: https://csvkit.readthedocs.io/en/latest/
 
+## Show the experiments table as a Parallel Coordinates plot
```
dberenbaum, line 351: Nice that you already added this!

dberenbaum (commented):
Unfortunately, it's pretty hard to read the example images because they are so small. Not sure if there's anything we can do about it 🤔 .

content/docs/command-reference/exp/show.md
```diff
@@ -53,6 +53,20 @@ sorted inside each group, chronologically by default. The `--sort-by` and
...
 
+### Parallel Coordinates Plot
+
+When the `--pcp` option is passed, an interactive Parallel Coordinates Plot will
+be generated using the data from the table.
+
+This plot is useful to explore the relationships between the metrics _and_
+params used in experiments.
+
+![](/img/user_guide_parallel_coordinates.gif) _Parallel Coordinates Plot_
```
daavoo, line 64: > Unfortunately, it's pretty hard to read the example images because they are so small. Not sure if there's anything we can do about it 🤔 .

@dberenbaum Do you refer to this GIF? 

For the other images, I am able to click and open a full page view (where I am able to read all letters)

dberenbaum: Yeah, I can open them, but they look small when embedded within the page.

daavoo: Hacked the plotly layout a bit to increase height and font size (also updated plots to use same examples as the above tables, from example-get-started).

Line width, unfortunately, is a need sponsor feature https://github.com/plotly/plotly.js/issues/2573 😅

dberenbaum: Font size was the real concern. Looks better now, thanks!

dberenbaum (approved):

content/docs/command-reference/exp/show.md
jorgeorpinel: Right, right. It has metrics and params (some filtered) I was confused. But still, the description of `exp show` needs to be a bit more general if we keep this feature here. Currently it's "prints a customizable table of experiments..." Maybe "displays your experiments in a customizable table or parallel coordinates plot"?

jorgeorpinel: > p.s. sorry for the delay.

content/docs/command-reference/exp/show.md
```diff
@@ -138,6 +155,14 @@ metric or param.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+- `--pcp` - generates an interactive Parallel Coordinates Plot from the table.
```
jorgeorpinel, line 158:
```suggestion
- `--pcp` - generates an interactive [parallel coordinates plot](#parallel-coordinates-plot) from the table.
```

content/docs/command-reference/exp/show.md
```diff
@@ -138,6 +155,14 @@ metric or param.
...
+
+- `-o <folder>, --out <folder>` - destination folder of the `--pcp` plot. By
+  default, `dvc_plots`.
```
jorgeorpinel, line 161: Per https://github.com/iterative/dvc.org/pull/3105#pullrequestreview-834458070:
```suggestion
- `-o <folder>, --out <folder>` - when used with `--pcp`, specifies a destination
  `folder` of the plot. By default its `dvc_plots`.
```

content/docs/command-reference/exp/show.md
jorgeorpinel: > should state it has no effect without (`--pcp`)

I'll update this. Resolving.

content/docs/user-guide/experiment-management/comparing-experiments.md
jorgeorpinel: > --json output is neither a table and it's named Get experiments table in JSON

Good point I guess we could separately update that title too... See https://github.com/iterative/dvc.org/pull/3137

I see the example title in question has been updated, thanks. Resolving!

content/docs/command-reference/exp/show.md
```diff
@@ -138,6 +155,14 @@ metric or param.
...
+
+- `--open` - opens the generated `--pcp` plot in the browser automatically. Only
+  used if `--pcp` was also passed.
```
jorgeorpinel, line 164:
```suggestion
- `--open` - when used with `--pcp`, opens the generated plot in a browser
  automatically.
```

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +260,38 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
 └───────────────────────┴──────────────┴─────────┴────────────────────────┴──────────────────┘
 ```
 
-Note that in the final example, Git commits remain in chronological order. The
+Note that in this example, Git commits remain in chronological order. The
 sorting only applies to experiment groups (sharing a parent commit).
 
+## Example: Parallel coordinates plot (PCP)
+
+To generate an interactive PCP based on the experiments and their parameters:
```
jorgeorpinel, line 268:
```suggestion
To generate an interactive [PCP](#parallel-coordinates-plot) based on the experiments
and their parameters:
```

jorgeorpinel (commented):
Copy edits and linking (committing)

content/docs/command-reference/exp/show.md
```diff
@@ -53,6 +53,20 @@ sorted inside each group, chronologically by default. The `--sort-by` and
...
+be generated using the data from the table.
+
+This plot is useful to explore the relationships between the metrics _and_
+params used in experiments.
+
+![](/img/pcp_interaction.gif) _Parallel Coordinates Plot_
+
+The interactivity of the plot allows you to extract different insights from the
+data. For example, the order of the columns can make some patterns more easily
+visible.
```
jorgeorpinel, line 68:
```suggestion
```

content/docs/command-reference/exp/show.md
```diff
@@ -53,6 +53,20 @@ sorted inside each group, chronologically by default. The `--sort-by` and
 `--sort-order` options can change this ordering, based on any single, visible
 metric or param.
 
+### Parallel Coordinates Plot
+
+When the `--pcp` option is passed, an interactive Parallel Coordinates Plot will
+be generated using the data from the table.
+
+This plot is useful to explore the relationships between the metrics _and_
+params used in experiments.
```
jorgeorpinel, line 62:
```suggestion
When the `--pcp` option is passed, an interactive parallel coordinates plot will
be generated using the same data from the table. This plot is useful to explore the
relationships between the metrics and params used in experiments. You can reorder
the columns to make some patterns more easily visible.
```

jorgeorpinel: + maybe link to https://en.wikipedia.org/wiki/Parallel_coordinates ?

content/docs/command-reference/exp/show.md
```diff
@@ -53,6 +53,20 @@ sorted inside each group, chronologically by default. The `--sort-by` and
 `--sort-order` options can change this ordering, based on any single, visible
 metric or param.
 
+### Parallel Coordinates Plot
+
```
jorgeorpinel, line 57:
```suggestion
```

jorgeorpinel (commented):
Idea to simplify the PCP section of the description.

> Ideally this should be in the guide actually but it's OK like this for now.

content/docs/command-reference/exp/show.md
jorgeorpinel: > > hard to read the example images because they are so small
>
> Do you refer to this GIF?

I do think the GIF could use a smaller example. Maybe just 2 params and a single metric ([example](https://docs.wandb.ai/ref/app/features/panels/parallel-coordinates)). I'd also reduce the number of experiments (7 max) if possible so this sample is short.

The ones in the longer example seem fine to me since you can click on them to examine them (Still preferably a few less experiments would be ideal IMO, but up to you).

content/docs/command-reference/exp/show.md
```diff
@@ -111,6 +125,9 @@ metric or param.
   sorting. This only affects the ordering of experiments derived from the same
   parent commit. Parent commits are always sorted chronologically.
 
+  When combined with the `--pcp` flag, the values of the column `name` will be
+  used to determine the color of the lines in the Parallel Coordinates Plot.
```
jorgeorpinel, line 129: Misc. copy edit
```suggestion
  > Note that the sort order will determine the color of the lines in `--pcp` mode.
```
Not sure this note is needed though, it'll be evident upon usage.

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +260,39 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
...
+and their parameters:
+
+```dvc
+$ dvc exp show --all-branches --pcp
+```
+
+![](/img/ref_pcp_default.png) _Parallel Coordinates Plot_
+
+The color of the lines is determined by the value of the column you select with
+`--sort-by`:
```
jorgeorpinel, line 278: Copy edit (committing):
```suggestion
Using `--sort-by` will reorder the plot experiments as expected, and determine
the color of the lines that represent them:
```

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +260,39 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
...
+
+```dvc
+$ dvc exp show --all-branches --pcp --sort-by roc_auc
+```
+
+> When `--sort-by` is not supplied, the color of the lines will be determined by
+> the `Experiment` column.
```
jorgeorpinel, line 285:
```suggestion
```

Probably unnecessary note 🙂

content/docs/command-reference/exp/show.md
```diff
@@ -235,8 +260,39 @@ $ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-ord
...
+
+![](/img/ref_pcp_sortby.png) _Colorized by roc_auc_
+
+Combine with other flags for further filtering:
+
+```dvc
+$ dvc exp show --all-branches --pcp --sort-by roc_auc --exclude-metrics avg_prec
```
jorgeorpinel, line 292: To avoid horizontal scroll when rendered (committing):
```suggestion
$ dvc exp show --all-branches --pcp --sort-by roc_auc
               --exclude-metrics avg_prec
```

jorgeorpinel (commented):
Small updates to the PCP example in the ref:

content/docs/command-reference/exp/show.md
jorgeorpinel: NOTE: I just realized we'll need to remove links to this section though. I think that's best. Replace them for links to [/doc/user-guide/experiment-management/comparing-experiments#parallel-coordinates-plot](https://dvc-org-dvc-parallel-co-flymou.herokuapp.com/doc/user-guide/experiment-management/comparing-experiments#parallel-coordinates-plot) please.

content/docs/user-guide/experiment-management/comparing-experiments.md
```diff
@@ -189,6 +189,22 @@ $ dvc exp show --sort-by auc --sort-order desc
...
 
+## Parallel Coordinates Plot
+
+`dvc exp show` can also generate an interactive
+[Parallel Coordinates Plot](/doc/command-reference/exp/show#parallel-coordinates-plot),
+with `--pcp`.
+
+The `--pcp` option can be combined with the other flags of `dvc exp show`. For
+example, the color of the lines will be determined by the values in the column
+passed to `--sort-by`.
```
jorgeorpinel, line 200: Suggestion on the guide section:

```suggestion
You can also generate an interactive parallel coordinates plot with
`dvc exp show --pcp`. This flag can be combined with other options of the
command. For example, use  `--sort-by` to sort the experiments and determine
the color of the lines that represent them.
```

- [ ] But link to https://en.wikipedia.org/wiki/Parallel_coordinates instead? Or some other (more authoritative) source

jorgeorpinel:
This is in pretty good shape but I left some specific changes that I think would be ideal. Not trying to block merge but if we have some time please address those. There's also (from the decription):

> - [ ] Include https://github.com/iterative/dvc.org/pull/3101

Thanks!

iesahin (approved):
LGTM. Thank you @daavoo

content/docs/command-reference/exp/show.md
daavoo: I am not sure about reducing the number of experiments. 

I think that the main usage is when you have many experiments and the plot helps to perceive some patterns more easily than the table.

daavoo: Agree on reducing to 2 params 1 metric

content/docs/command-reference/exp/show.md
```diff
@@ -111,6 +125,9 @@ metric or param.
   sorting. This only affects the ordering of experiments derived from the same
   parent commit. Parent commits are always sorted chronologically.
 
+  When combined with the `--pcp` flag, the values of the column `name` will be
+  used to determine the color of the lines in the Parallel Coordinates Plot.
+
```
daavoo, line 130:
```suggestion
```

content/docs/command-reference/exp/show.md
daavoo: I will wait for https://github.com/iterative/dvc/pull/7141 in order to update the plot (to perhaps remove the Experiment column). Leaving the current GIF for now, it that's ok

daavoo:
> This is in pretty good shape @daavoo, but I left some specific suggestion that I think would be ideal (6 unresolved comments as of now, starting at [#3105 (review)](https://github.com/iterative/dvc.org/pull/3105#pullrequestreview-841094139)). Not trying to block merge but if we have some time please address those. There's also (from the description):
> 

I think I addressed all comments. About the item in description, I think we will have to wait until it gets merged on core and then update

shcheklein:
@jorgeorpinel any updates? can we merge this?

content/docs/command-reference/exp/show.md
```diff
@@ -1,6 +1,7 @@
 # exp show
 
-Print a customizable table of experiments, their metrics and parameters.
+Displays your experiments in a customizable table or
+[parallel coordinates plot](#example-parallel-coordinates-plot-pcp).
```
jorgeorpinel, line 4: Here...

content/docs/command-reference/exp/show.md
```diff
@@ -54,6 +55,14 @@ sorted inside each group, chronologically by default. The `--sort-by` and
 `--sort-order` options can change this ordering, based on any single, visible
 metric or param.
 
+When the `--pcp` option is passed, an interactive
+[parallel coordinates plot](#example-parallel-coordinates-plot-pcp) will be
```
jorgeorpinel, line 59: Also here...
> But maybe also add a paragraph linking to the example like `(see an [example](...))`

content/docs/command-reference/exp/show.md
```diff
@@ -115,6 +124,16 @@ metric or param.
 
 - `-v`, `--verbose` - displays detailed tracing information.
 
+- `--pcp` - generates an interactive
+  [parallel coordinates plot](#example-parallel-coordinates-plot-pcp) from the
```
jorgeorpinel, line 128: And here. Or remove this one -- not a big deal here.

jorgeorpinel (commented):
I think we should link to /doc/user-guide/experiment-management/comparing-experiments#parallel-coordinates-plot instead in:

jorgeorpinel (commented):
> @jorgeorpinel any updates? can we merge this?

Sure, it's not my intention to bock this. It's def. mergeable (and already approved).

Thanks again for addressing all the feedback @daavoo

PR #3107, opened 2021-12-16
Fix site crashing with certain anchor links

### LINKED ISSUE #2395: URLs that contain anchors with reserved chars break the website
broken page and:

`Error: SyntaxError: Failed to execute 'querySelector' on 'Document': '#option!134' is not a valid selector.`

on 

https://dvc.org/doc/command-reference/gc#option!134

Priority depends on if we have titles (anchors with `!` and other funny symbols).

julieg18:
* stop code from running `querySelector` on invalid id strings

Fixes #2395

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -4,6 +4,7 @@ import { useLocation } from '@reach/router'
 import GatsbyLink from 'gatsby-link'
 import { getRedirect } from '../../utils/shared/redirects'
 import { scrollIntoLayout, getScrollNode } from '../../utils/front/scroll'
+import safeQuerySelector from '../../utils/front/safeQuerySelector'
 
 export type ILinkProps = {
   children: React.ReactNode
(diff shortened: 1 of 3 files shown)


src/components/Link/index.tsx
```diff
@@ -77,7 +77,7 @@ const ResultLinkComponent: React.FC<ILinkProps> = ({
 }
 
 const scrollToHash = (hash: string, scrollOptions = {}): void => {
-  if (hash) {
+  if (hash && /^[A-Za-z][-A-Za-z0-9_:.]*$/.test(hash)) {
```
julieg18, line 80: Whenever an anchor link didn't match a valid HTML id, `querySelector` through an error and crashed the site. I added a regex stopping `querySelector` from running unless the anchor link is valid.

src/components/Link/index.tsx
```diff
@@ -77,7 +77,7 @@ const ResultLinkComponent: React.FC<ILinkProps> = ({
 }
 
 const scrollToHash = (hash: string, scrollOptions = {}): void => {
-  if (hash) {
+  if (hash && /^#[A-Za-z][-A-Za-z0-9_:.]*$/.test(hash)) {
```
julieg18, line 80: Whenever an anchor link didn't match a valid HTML id, querySelector through an error and crashed the site. I added a regex test stopping querySelector from running unless the anchor link is a valid id.

rogermparent (commented):
I may be better to create a new function called something like `safeQuerySelector` and reuse that across both of these areas instead of copying the regex in two places.

julieg18:
> I may be better to create a new function called something like `safeQuerySelector` and reuse that across both of these areas instead of copying the regex in two places. We could even wrap the `querySelector` call in a `try...catch` instead of regexing.

Good idea! I created a `safeQuerySelector` which uses a `try...catch` instead of using a regex.

rogermparent (approved):
Looks good!

PR #3112, opened 2021-12-17
exp show: Add --keep and --drop

daavoo:
P.R. for https://github.com/iterative/dvc/pull/7141

- [ ] After this PR: https://github.com/iterative/dvc.org/issues/3118


daavoo (commented):
I have only included changes in the ref because I have a few questions about the user guide:

- What's the scope of the `Examples`section in `ref` v.s. the user guide (`Comparing Experiments`)?

- Should we use a more "complex" table in the user guide?

It currently states `The table output may become cluttered if you have a large number of parameters and metrics.` but the tables have only 2 params and 2 metrics so all the filtering seems a little pointless.

- Should we ensure that the tables and commands used in the user guide (or better, everywhere) are somehow reproducible?

@jorgeorpinel @iesahin

iesahin:
> I have only included changes in the ref because I have a few questions about the user guide:
> 
>     * What's the scope of the `Examples`section in `ref` v.s. the user guide (`Comparing Experiments`)?

I believe "one liners" can be included in the REF, whereas longer examples are more suitable to the UG. This is my personal view though, and I tend to favor UG for examples than the REF. 

> 
>     * Should we use a more "complex" table in the user guide?

No need to make them artificially complex I believe, they already look complex. 

> 
> 
> It currently states `The table output may become cluttered if you have a large number of parameters and metrics.` but the tables have only 2 params and 2 metrics so all the filtering seems a little pointless.
> 
>     * Should we ensure that the tables and commands used in the user guide (or better, everywhere) are somehow reproducible?
> 

We try to make only the Get Started fully reproducible. Other parts are a bit disparate that trying to make them fully reproducible looks futile. 

> 
> @jorgeorpinel @iesahin

content/docs/command-reference/exp/show.md
```diff
@@ -48,10 +46,10 @@ which you can typically exit by typing `Q`. Use `--no-pager` to print the table
 to standard output.
 
 By default, the printed experiments table will include columns for all metrics
-and params from the entire project. The `--param-deps`, `--include-metrics`, and
+and params from the entire project. The `--param-deps`, `--drop`, `--keep` and
```
dberenbaum, line 49: Unrelated to this PR, but why call out `--param-deps` first here? It seems like a less common option. Could `--only-changed` be called out first instead since it's often a useful option?

content/docs/command-reference/exp/show.md
```diff
@@ -48,10 +46,10 @@ which you can typically exit by typing `Q`. Use `--no-pager` to print the table
...
 other [options](#options) can determine which ones should be displayed.
 
-Experiments in the table are first grouped (by parent commit). They are then
+experiments in the table are first grouped (by parent commit). They are then
```
dberenbaum, line 52: Why this change?

content/docs/command-reference/exp/show.md
```diff
@@ -78,36 +76,12 @@ metric or param.
...
-
-- `--exclude-metrics <list>` - hide the specified `dvc metrics` from the table
-  (all param will be shown except for these). Accepts a comma-separated `list`
-  of metric names. Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`,
-  and `**`. If a `path:` prefix is included in the name, the wildcard pattern
-  will **not** be matched against that prefix but only against the rest of the
-  name.
+  across experiments.
+
+- `--drop <list>` - remove the columns matching the specified regex patterns.
```
dberenbaum, line 81: Does this accept a comma-separated list or just a single regex? Could you explain this a bit more and how to use in the basic case where regex isn't needed? Probably not all users will be familiar with regex.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,129 @@ metric or param.
...
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
+│ workspace               │  0.60405 │  0.9608 │ 3000                   │ 100         │ 64              │
+│ master                  │  0.60405 │  0.9608 │ 3000                   │ 100         │ 64              │
+│ ├── d384680 [exp-bc055] │  0.51799 │ 0.92333 │ 500                    │ 100         │ 64              │
+│ ├── 6b338f8 [exp-3315b] │  0.58589 │   0.945 │ 2000                   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │  0.56447 │ 0.94713 │ 1500                   │ 100         │ 64              │
+└─────────────────────────┴──────────┴─────────┴────────────────────────┴─────────────┴─────────────────┘
 ```
 
 Sort experiments by the `auc` metric, in ascending order:
```
dberenbaum, line 199: Hm, is it better to specify `ascending order` here if the command doesn't use `--sort-order`? It shows default behavior but doesn't show use of `--sort-order` like it did before.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,129 @@ metric or param.
...
+│ ├── 6b338f8 [exp-3315b] │ 08:03 PM     │  0.58589 │   0.945 │ 2000                   │ 2                │ 20170428   │ 100         │ 64              │
+│ ├── d384680 [exp-bc055] │ 08:03 PM     │  0.51799 │ 0.92333 │ 500                    │ 2                │ 20170428   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │ 08:03 PM     │  0.56447 │ 0.94713 │ 1500                   │ 2                │ 20170428   │ 100         │ 64              │
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┴──────────────────┴────────────┴─────────────┴─────────────────┘
+```
+
+Combine `--only-changed` with `--drop` and/or `--keep`:
+
+```dvc
+$ dvc exp show --only-changed --drop 'neutral:**Created**' --keep 'train.(?!seed)'
```
dberenbaum, line 184:
```suggestion
$ dvc exp show --only-changed --drop Created --keep 'train.(?!seed)'
```

Was it intentional to include this syntax formatting? It's showing up as text in the deployment:

Screen Shot 2021-12-20 at 10 38 33 AM

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,129 @@ metric or param.
...
+│ ├── 6b338f8 [exp-3315b] │  0.58589 │   0.945 │ 2000                   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │  0.56447 │ 0.94713 │ 1500                   │ 100         │ 64              │
+└─────────────────────────┴──────────┴─────────┴────────────────────────┴─────────────┴─────────────────┘
 ```
 
 Sort experiments by the `auc` metric, in ascending order:
 
 ```dvc
-$ dvc exp show --include-params=featurize --sort-by=auc --sort-order=asc
+$ dvc exp show --only-changed --sort-by=metric:**roc_auc**
```
dberenbaum, line 202:
```suggestion
$ dvc exp show --only-changed --sort-by=roc_auc
```
Same as above.

dberenbaum: It probably is good to bold `--sort-by=roc_auc` to highlight the relevant part of the command.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,129 @@ metric or param.
...
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┘
 ```
 
-You can also filter out any metrics and parameters that do not change across the
-shown experiments:
+You can also use `--drop` to filter columns matching specific regex patterns:
 
 ```dvc
-$ dvc exp show --only-changed
+$ dvc exp show --drop 'prepare'
```
dberenbaum, line 166: Very minor, but maybe exclude quotes when doing simple exact column matches so they don't appear to be required.

dberenbaum (changes requested):
Minor changes requested. Looking good!

content/docs/command-reference/exp/show.md
daavoo: It was not intentional

content/docs/command-reference/exp/show.md
daavoo: Should I replace with `descending` order ? 

I think showing an option that it's not needed might be counterproductive.

content/docs/command-reference/exp/show.md
daavoo: I don't think it's possible to do inside a code block.

content/docs/command-reference/exp/show.md
daavoo: Does it make sense to keep supporting a list (comma separated `--drop 'x,y'` or multiple `--drop x --drop y`) now that it supports regex?

I understand that some users might not be familiar with regex but, if we show an example of how `--drop 'x|y'` would be equivalent as passing multiple values, I'm no longer sure if the lack of regex knowledge would be a problem at all.

content/docs/command-reference/exp/show.md
daavoo: messed up with search and replace

content/docs/command-reference/exp/show.md
daavoo: cc @iterative/websites ?

daavoo:
> I believe "one liners" can be included in the REF, whereas longer examples are more suitable to the UG. This is my personal view though, and I tend to favor UG for examples than the REF.

I will try to follow that line and update the user guide as part of this P.R. for revision.

> We try to make only the Get Started fully reproducible. Other parts are a bit disparate that trying to make them fully reproducible looks futile.

Didn't mean fully reproducible. In cmd ref we have a note:

https://github.com/iterative/dvc.org/blob/master/content/docs/command-reference/exp/show.md?plain=1#L146-L149

But in user guide I wasn't sure how to replicate the output tables.

content/docs/command-reference/exp/show.md
dberenbaum: Agreed. I'm fine with whatever you want to do here.

content/docs/command-reference/exp/show.md
rogermparent: I don't think it's possible, but we could look into adding this as a feature if it's something we want.

dberenbaum: Not a big deal. I guess it's also possible to do with [html](https://meta.stackexchange.com/questions/183610/how-to-combine-bold-and-code-sample-in-markdown), but let's not worry about it now.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,129 @@ metric or param.
...
+│ ├── 6b338f8 [exp-3315b] │ 08:03 PM     │  0.58589 │   0.945 │ 0.2           │ 20170428     │ 2000                   │ 2                │ 20170428   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │ 08:03 PM     │  0.56447 │ 0.94713 │ 0.2           │ 20170428     │ 1500                   │ 2                │ 20170428   │ 100         │ 64              │
+└─────────────────────────┴──────────────┴──────────┴─────────┴───────────────┴──────────────┴────────────────────────┴──────────────────┴────────────┴─────────────┴─────────────────┘
 ```
 
 > You can exit this screen with `Q`, typically.
 
-Let's limit the param columns to only include the `featurize` group:
+As a quick way of reducing noise, `--only-changed` will drop any column with
+values that do not change across the shown experiments:
```
iesahin, line 145: ~~shown~~

content/docs/command-reference/exp/show.md
iesahin: Could there be more than one `--drop` or `--keep` in a command? Like `dvc exp show --drop this --drop that`?

iesahin (approved):
LGTM. Thank you @daavoo

content/docs/command-reference/exp/show.md
daavoo: Updated core to accept single value only and added example for `x|y`

daavoo: > Could there be more than one `--drop` or `--keep` in a command? Like `dvc exp show --drop this --drop that`?

No. I added an example on how `this|that` could achieve the same

jorgeorpinel:
> What's the scope of the Examplessection in ref v.s. the user guide (Comparing Experiments)?

Great question. It's something we only started considering recently when writing https://dvc.org/doc/user-guide/experiment-management. I tend to think contrary to @iesahin. Examples in the ref. should serve as quick ways for users to grasp how the tool does its job and cover some main use cases. In the UG it's unclear we want separate example sections, as we can embed the examples throughout it. I believe @shcheklein also favors this view. That said we may currently have too many examples in the cmd ref that should just be part of the guide explanations (replaced with more simple stuff in the ref).

> Should we use a more "complex" table in the user guide?
the tables have only 2 params and 2 metrics so all the filtering seems a little pointless.

Good point. If the small table is limiting your explanations then I think yes, it should be updated to something longer but not larger than totally necessary.

> Should we ensure that the tables and commands used in the user guide (or better, everywhere) are somehow reproducible?

Agree with @iesahin on this: we only worry about "reproducibility" in the GS at this point. We don't have the capacity to maintain all other examples to that level and hopefully it's not necessary.

content/docs/command-reference/exp/show.md
```diff
@@ -78,36 +76,12 @@ metric or param.
...
-  of metric names. Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`,
-  and `**`. If a `path:` prefix is included in the name, the wildcard pattern
-  will **not** be matched against that prefix but only against the rest of the
-  name.
+  across experiments.
+
+- `--drop <regex>` - remove the columns matching the specified [regex][regex].
+
+- `--keep <regex>` - prevent the columns matching the specified [regex][regex]
+  to be removed by any of the other options.
```
jorgeorpinel, line 84: I hope we can introduce them without the concept of regex at first, and then mention separately any regex is accepted.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,147 @@ metric or param.
 > can find the actual source code.
 
 Let's say we have run 3 experiments in our project. The basic usage shows the
-workspace (Git working tree) and experiments derived from `HEAD`
-(`10-bigrams-experiment` branch in this case), and all of their metrics and
-params (scroll right to see all):
+workspace (Git working tree) and experiments derived from `HEAD` (`master`
+branch in this case), and all of their metrics and params (scroll right to see
```
jorgeorpinel, line 123: Great job updating the example @daavoo, thanks. It's basically a guide so I see why you're not sure what to do in the UG. I think you shouldn't worry about the guide except minimal updates to reflect this change. We have not finished moving some of this info from the ref to the guide anyway so we'll have to revisit this.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,147 @@ metric or param.
...
+│ master                  │ May 29, 2021 │  0.60405 │  0.9608 │ 3000                   │
+│ ├── d7fdde2 [exp-1b262] │ 08:03 PM     │  0.56447 │ 0.94713 │ 1500                   │
+│ ├── 6b338f8 [exp-3315b] │ 08:03 PM     │  0.58589 │   0.945 │ 2000                   │
+│ └── d384680 [exp-bc055] │ 08:03 PM     │  0.51799 │ 0.92333 │ 500                    │
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┘
 ```
 
-You can also filter out any metrics and parameters that do not change across the
-shown experiments:
+You can also use `--drop` to filter columns matching a specific [regex][regex]:
```
jorgeorpinel, line 163: Again the concept of regex is too technical. I wouldn't mention it down here:
```suggestion
You can also use `--drop` to filter specific columns:
```
> Unless you actually put a (clear) regex in the example

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,147 @@ metric or param.
...
+│ └── d7fdde2 [exp-1b262] │ 08:03 PM     │  0.56447 │ 0.94713 │ 1500                   │ 2                │ 20170428   │ 100         │ 64              │
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┴──────────────────┴────────────┴─────────────┴─────────────────┘
 ```
 
-Sort experiments by the `auc` metric, in ascending order:
+You can use `|` in order to remove a list of columns:
 
 ```dvc
-$ dvc exp show --include-params=featurize --sort-by=auc --sort-order=asc
+$ dvc exp show --drop 'avg_prec|train.min_split'
```
jorgeorpinel, line 184: Hmmmm. Do we recommend this over several `--drop`s? Not sure. Feels like we're documenting regular expressions here. We could skip this one (but maybe add another `--drop` to the previous one).

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,147 @@ metric or param.
...
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━┩
+│ workspace               │ -            │  0.9608 │ 0.2           │ 20170428     │ 3000                   │ 2                │ 20170428   │ 100         │
+│ master                  │ May 29, 2021 │  0.9608 │ 0.2           │ 20170428     │ 3000                   │ 2                │ 20170428   │ 100         │
+│ ├── d384680 [exp-bc055] │ Dec 17, 2021 │ 0.92333 │ 0.2           │ 20170428     │ 500                    │ 2                │ 20170428   │ 100         │
+│ ├── d7fdde2 [exp-1b262] │ Dec 17, 2021 │ 0.94713 │ 0.2           │ 20170428     │ 1500                   │ 2                │ 20170428   │ 100         │
+│ └── 6b338f8 [exp-3315b] │ Dec 17, 2021 │   0.945 │ 0.2           │ 20170428     │ 2000                   │ 2                │ 20170428   │ 100         │
+└─────────────────────────┴──────────────┴─────────┴───────────────┴──────────────┴────────────────────────┴──────────────────┴────────────┴─────────────┘
+```
+
+Combine `--only-changed` with `--drop` and/or `--keep`:
```
jorgeorpinel, line 199:
```suggestion
If combined `--only-changed` has the least priority, `--drop` comes next,
and `--keep` has the last word:
```

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,147 @@ metric or param.
...
+│ ├── d384680 [exp-bc055] │  0.51799 │ 0.92333 │ 500                    │ 100         │ 64              │
+│ ├── 6b338f8 [exp-3315b] │  0.58589 │   0.945 │ 2000                   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │  0.56447 │ 0.94713 │ 1500                   │ 100         │ 64              │
+└─────────────────────────┴──────────┴─────────┴────────────────────────┴─────────────┴─────────────────┘
+```
+
+Sort experiments by the `auc` metric, in descending order:
+
+```dvc
+$ dvc exp show --only-changed --sort-by=roc_auc --sort-order desc
```
jorgeorpinel, line 220: Seems unrelated? And there's already a sorting example after this.

jorgeorpinel (commented):
My comments and a couple suggestions. Thanks again @daavoo

content/docs/command-reference/exp/show.md
daavoo: I've removed support for multiple `--drop` in core.

I've considered that joining using `|`  was an easy to teach replacement for either multiple `--drop` or comma-separated lists.

Do you consider that we should still support multiple values?

content/docs/command-reference/exp/show.md
daavoo: In the original examples there were also 2 examples for `--sort-by`. 

Only change I did here is to replace `--sort-order asc` with `--sort-order desc` because `--sort-order asc` has the same behavior as not passing the option at all, and could give the impression that you need to pass the value in order to actually sort.

content/docs/command-reference/exp/show.md
daavoo: Do you think we should still mention priorities on the `Options` section of this would be enough ?

content/docs/command-reference/exp/show.md
daavoo: I'm not sure what term to use here instead of regex.

content/docs/command-reference/exp/show.md
iesahin: > tend to think contrary to @iesahin. Examples in the ref. should serve as quick ways for users to grasp how the tool does its job and cover some main use cases. In the UG it's unclear we want separate example sections, as we can embed the examples throughout it. I believe @shcheklein also favors this view.

I believe REF examples should be like [tldr](https://tldr.sh/) pages, and without much explanation. People (like me) usually refer to REF to find the exact option quickly.

For UG, I tend to have more explanations, describing "why" and "how" rather verbosely. These may be separated from the UG, though I believe having UG around workflows (rather than commands) might make more sense. In this case, we'll have larger examples.

content/docs/command-reference/exp/show.md
dberenbaum: "Pattern"? "Name"? Or simply "remove the matching columns."

content/docs/command-reference/exp/show.md
dberenbaum: How about something like, "You can use regex to match columns. For example, to remove multiple columns:"

content/docs/command-reference/exp/show.md
```diff
@@ -78,36 +76,12 @@ metric or param.
 - `--param-deps` - include only parameters that are stage dependencies.
 
 - `--only-changed` - show only parameters and metrics with values that vary
-  across experiments. Note that this option takes precedence over
-  `--include-params` and `--include-metrics`, for example given
```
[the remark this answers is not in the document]
dberenbaum: Agreed. I see it in the examples, but probably good to be explicit here as well.

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,148 @@ metric or param.
...
+┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
+│ workspace               │  0.60405 │  0.9608 │ 3000                   │ 100         │ 64              │
+│ master                  │  0.60405 │  0.9608 │ 3000                   │ 100         │ 64              │
+│ ├── d384680 [exp-bc055] │  0.51799 │ 0.92333 │ 500                    │ 100         │ 64              │
+│ ├── 6b338f8 [exp-3315b] │  0.58589 │   0.945 │ 2000                   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │  0.56447 │ 0.94713 │ 1500                   │ 100         │ 64              │
+└─────────────────────────┴──────────┴─────────┴────────────────────────┴─────────────┴─────────────────┘
+```
+
+Sort experiments by the `auc` metric, in descending order:
```
dberenbaum, line 218:
```suggestion
Sort experiments by the `roc_auc` metric, in descending order:
```

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,148 @@ metric or param.
...
+│ ├── 6b338f8 [exp-3315b] │ 08:03 PM     │  0.58589 │   0.945 │ 2000                   │
+│ └── d384680 [exp-bc055] │ 08:03 PM     │  0.51799 │ 0.92333 │ 500                    │
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┘
 ```
 
 To see all experiments throughout the Git history:
 
 ```dvc
-$ dvc exp show --all-commits --include-params=featurize --sort-by=auc --sort-order=asc
+$ dvc exp show --all-commits --only-changed --sort-by=auc
```
dberenbaum, line 239:
```suggestion
$ dvc exp show --all-commits --only-changed --sort-by=roc_auc
```

dberenbaum (changes requested):
Minor changes requested 🙏

content/docs/command-reference/exp/show.md
```diff
@@ -147,109 +119,148 @@ metric or param.
...
+│ workspace               │ -            │  0.60405 │  0.9608 │ 3000                   │ 2                │ 20170428   │ 100         │ 64              │
+│ master                  │ May 29, 2021 │  0.60405 │  0.9608 │ 3000                   │ 2                │ 20170428   │ 100         │ 64              │
+│ ├── 6b338f8 [exp-3315b] │ 08:03 PM     │  0.58589 │   0.945 │ 2000                   │ 2                │ 20170428   │ 100         │ 64              │
+│ ├── d384680 [exp-bc055] │ 08:03 PM     │  0.51799 │ 0.92333 │ 500                    │ 2                │ 20170428   │ 100         │ 64              │
+│ └── d7fdde2 [exp-1b262] │ 08:03 PM     │  0.56447 │ 0.94713 │ 1500                   │ 2                │ 20170428   │ 100         │ 64              │
+└─────────────────────────┴──────────────┴──────────┴─────────┴────────────────────────┴──────────────────┴────────────┴─────────────┴─────────────────┘
 ```
 
-Sort experiments by the `auc` metric, in ascending order:
+You can use `|` in order to remove a list of columns:
```
daavoo, line 181:
```suggestion
You can use [regex][regex] to match columns. For example, to remove multiple columns:
```

content/docs/command-reference/exp/show.md
daavoo:
```suggestion
- `--drop <regex>` - remove the matching columns.

- `--keep <regex>` - prevent the matching columns to be removed by any of the other options.
```

iesahin:
A bit off-topic but I wonder if there is a way to turn off all params or metrics? Like `--include-metrics ""` before?

daavoo:
> A bit off-topic but I wonder if there is a way to turn off all params or metrics? Like --include-metrics "" before?

No, the new filtering operates on column names and it has no concept of the type of column.

content/docs/command-reference/exp/show.md
```diff
@@ -258,3 +272,5 @@ sorting only applies to experiment groups (sharing a parent commit).
 
 📖 See [Metrics, Parameters, and Plots](/doc/start/metrics-parameters-plots) for
 an introduction to parameters, metrics, plots.
+
+[regex]: https://en.wikipedia.org/wiki/Regular_expression
```
iesahin, line 276: I think reading this page makes people more confused than searching for "regexp." Could you replace it with a simpler page, like https://towardsdatascience.com/an-introduction-to-regular-expressions-5dd762afc5e4 ?

dberenbaum (approved):

content/docs/command-reference/exp/show.md
daavoo: Replaced with https://regexone.com/

PR #3113, opened 2021-12-18
Move docs node generation to a local plugin

rogermparent:
This PR is a first step in porting our docs engine into a reusable Gatsby theme. It moves all the files from our docs Gatsby Model to a new local plugin at `plugins/gatsby-theme-iterative-docs`.

From here, we can continue to move more parts of our docs engine (sidebar, components) to this plugin directory. Once we have everything moved and ensure it still works, we can turn the local plugin folder into its own node module and start work on reintegrating it with the other sites.


gatsby-config.js
```diff
@@ -36,6 +36,13 @@ const plugins = [
   'gatsby-plugin-react-helmet',
   'gatsby-plugin-sitemap',
   'gatsby-plugin-twitter',
+  {
+    resolve: 'gatsby-theme-iterative-docs',
+    options: {
+      docComponent: require.resolve('./src/templates/doc.tsx'),
```
julieg18, line 42: Where is the `docComponent` being used in the theme? I'm not seeing it being used anywhere 🤔

julieg18 (approved):
Looks great @rogermparent! Just one small question :)

gatsby-config.js
rogermparent: Whoops! I renamed this to `defaultTemplate` in `createPages` but didn't end up updating it here.

yathomasi (commented):
Everything looks good @rogermparent .

PR #3115, opened 2021-12-20
Update landing page

### LINKED ISSUE #3108: Revert Landing Page back to normal
@iterative/websites We need to revert the landing page to how it was before.  Question though.  Would it be possible to keep the modal if it can auto feed what's in the blog?  If it's something we have to keep changing, then nevermind, just revert.  But if it can be auto populated with what's the latest blog post, it may be worth keeping.

julieg18:
* take off youtube video from hero, using old animation
* Update the "What's New" modal, populating it with the latest blog post

Fixes #3108


rogermparent (approved):
Looks good, nice work!

yathomasi (commented):
Looks good @julieg18 .

jendefig (approved):
Love it!  This is awesome @julieg18 Very excited about this! 🙏

PR #3117, opened 2021-12-21
Replace Google Analytics with Plausible

### LINKED ISSUE #1113: analytics: improve Google Analytics integration
Currently we only track page views through `gatsby-plugin-google-analytics`. We can track much more using their `<OutboundLink>` component and their `trackCustomEvent` function.

> the first candidate would be to log next/previous page views, social share buttons, comments button click, tags click, etc

@shcheklein I'm not sure what you mean with next/previous page clicks, do you mean the clicks on the pagination buttons?

Just an aside to keep in mind: I found an issue in the plugin. If we want to receive a callback when the event is sent, that [only occurs if google analytics script loads correctly](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-analytics/src/index.js#L101) (like with tracking protection mode, a privacy browser extension or an ad blocker). This is why I chose to call `window.ga` directly in my PR regarding tracking subscriptions. If we need this callback for our own flow control, we will need to issue a PR upstream and get it accepted.

### LINKED ISSUE #1179: analytics: should we strip queries?
dvc.org/blog/a-public-reddit-dataset

should be probably the same as

dvc.org/blog/a-public-reddit-dataset?fbcid=aassdfsdfsdf

(Need to check if it's a good practice or not)

### LINKED ISSUE #2596: Replace Google Analytics
* There are a couple ways we could do this. Two possibilities are using plausible analytics or have our backend handle analytics.

julieg18:
* I'll be sure check over all the events, making sure everything is configured correctly after we merge this. 

Fixes #2596, fixes #1113, fixes #1179


src/components/Page/DefaultSEO/index.tsx
```diff
@@ -116,7 +116,13 @@ const DefaultSEO: React.FC<IDefaultSEOProps> = ({ pathname }) => {
           href: fullUrl
         }
       ]}
-    />
+    >
+      <script
+        defer
+        data-domain="dvc.org"
+        src="https://plausible.io/js/plausible.outbound-links.js"
```
julieg18, line 123: I configured plausible to track outbound links for dvc.org, fixing #1113.

src/components/Community/Contribute/index.tsx
```diff
@@ -5,15 +5,17 @@ import LayoutWidthContainer from '../../LayoutWidthContainer'
...
 
 import { useCommunityData } from '../../../utils/front/community'
 import * as sharedStyles from '../styles.module.css'
 
-const logPR = (): void => logEvent('community', 'contribute-pr')
-const logBlogpost = (): void => logEvent('community', 'contribute-blogpost')
-const logTalk = (): void => logEvent('community', 'contribute-talk')
-const logAmbassador = (): void => logEvent('community', 'contribute-ambassador')
+const log = (type: string): void =>
+  logEvent('Community', { Section: 'contribute', Contribute: type })
```
julieg18, line 14: Since GA custom events work a little differently than Plausible, I had to tweak the events a bit. I also updated the naming on some of the events, making it more clear what the event meant.

rogermparent (approved):
I can't help but wonder if there could be a less manually intensive way to handle link tracking, but that's aside from the point of this PR- looks like a fine port!

PR #3127, opened 2021-12-22
Fix visibile outline in /community section headers

### LINKED ISSUE #1290: community page: remove border around section names
It should not affect accessibility (when you click Tab). It should only disable them after I click. 

Alternatively make them enabled only on Mobile (tablet?) and make them look nice.

Screen Shot 2020-05-13 at 4 15 23 PM

julieg18:
* replaces clickable div with button inside `Community/Section`

Fixes #1290

### DIFF
--- a/src/components/Community/Section/index.tsx
+++ b/src/components/Community/Section/index.tsx
@@ -4,7 +4,6 @@ import { useLocation } from '@reach/router'
 import { Collapse } from 'react-collapse'
 import { useWindowSize } from 'react-use'
 
-import { isTriggeredFromKB } from '../../../utils/front/keyboard'
 import { screens } from '../../../../config/postcss/media'
 
 import * as styles from './styles.module.css'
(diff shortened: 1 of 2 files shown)


shcheklein (approved):

rogermparent (commented):
This change itself is fine to me, but I just noticed that the community sections no longer collapse. Is that intentional? If so we should probably make these static, and if not we'll need to find a way to fix it.

julieg18:
> This change itself is fine to me, but I just noticed that the community sections no longer collapse. Is that intentional? If so we should probably make these static, and if not we'll need to find a way to fix it.

Are you referring to how the sections collapse on smaller screens? If so, they are collapsing on my end: 

https://user-images.githubusercontent.com/43496356/147242525-d74033cc-31cc-4c17-8083-069af64c6f90.mov

I also tested firefox and safari. I also tested the hyper links which work as well. What browser/OS are they not collapsing on @rogermparent?

rogermparent:
I must have missed that we only have the collapsing feature on smaller screens now! Sorry about that.

rogermparent (approved):

PR #3132, opened 2021-12-27
ref: Add `ui.theme` to `config`.

daavoo:
P.R. for https://github.com/iterative/dvc/pull/7202


content/docs/command-reference/config.md
```diff
@@ -281,6 +281,13 @@ be overidden explicitly through CLI arguments or through responses in prompts
   [custom HTML template](/doc/command-reference/plots#html-templates) for
   `dvc plots`. Accepts a path relative to the `.dvc/` folder.
 
+### ui
+
+- `ui.theme` - Optionally indicates the color theme being used in the terminal.
+  Accepts values `dark`, `light` and `None` (default). If not `None`, this
+  option is used to alternate row's background color in the table generated by
```
iesahin, line 288: row's -> rows'

iesahin: I think the second sentence can be rephrased to something like: 

If set to `light` or `dark`, this option is used to determine the alternating row colors in `dvc exp show` table output.

iesahin (changes requested):
Minor

PR #3133, opened 2021-12-27
Move sidebar logic to theme

rogermparent:
#3143 was originally part of this, but moved to make reviewing each individual feature easier.

This PR moves all our sidebar logic to our new reusable Gatsby theme.

Since the sidebars and logic have to be imported in both Gatsby Webpack and plain node environments while still being reusable, the sidebar leans on Gatsby Themes but in such a way that it can work in regular node. To help make this possible, sidebar normalization and helpers have been split and the helpers must be called on an imported instance of the sidebar- this helps particularly with plain Node.


gatsby-config.js
```diff
@@ -36,7 +36,12 @@ const plugins = [
   'gatsby-plugin-react-helmet',
   'gatsby-plugin-sitemap',
   'gatsby-plugin-twitter',
-  'gatsby-theme-iterative-docs',
+  {
+    resolve: 'gatsby-theme-iterative-docs',
+    options: {
+      remark: false
+    }
```
rogermparent, line 43: dvc.org's remark plugins are configured differently from the more basic ones from cml.dev (i.e. it has the local plugins and uses an external const for max image width), so we suppress the one from the theme and use our more specific one.

plugins/gatsby-theme-iterative-docs/gatsby-config.js
```diff
@@ -0,0 +1,57 @@
...
+            resolve: 'gatsby-remark-images',
+            options: {
+              withWebp: true
+            }
+          },
+          'gatsby-remark-responsive-iframe'
+        ]
+      }
+    })
+  }
```
rogermparent, line 53: The remark config for dvc.org is special, but the one from cml.dev used here is a more general configuration that is the minimum we'd apply to any docs site.
Ideally on most sites we won't set `remark:false` and let this take care of the remark config.

plugins/gatsby-theme-iterative-docs/normalize-sidebar.js
```diff
@@ -122,12 +82,10 @@ function normalizeItem({ rawItem, parentPath, resultRef, prevRef }) {
...
 
-function normalizeSidebar({
+function normalizeSidebar(
   data,
-  parentPath,
-  parentResultRef,
-  startingPrevRef
-}) {
+  { parentPath = '', parentResultRef, startingPrevRef } = {}
+) {
```
rogermparent, line 88: Small tweak to the signature to make this function more natural to use IMO, the rest of the args are just used to recursive calls so that function's practically unary.

plugins/gatsby-theme-iterative-docs/src/sidebar.js
```diff
@@ -0,0 +1 @@
+module.exports = []
```
rogermparent, line 1: Just here to be shadowed

src/components/Documentation/Layout/SidebarMenu/index.tsx
```diff
@@ -31,26 +33,15 @@ const ICONS: { [key: string]: React.FC<{ className?: string }> } = {
...
-  children,
-  label,
-  path,
+  item: { children, label, path, style, icon, type },
   activePaths,
-  onClick,
-  style,
-  icon,
-  type
+  onClick
```
rogermparent, line 44: Slight signature tweak which makes more sense to me now that we have a type def for sidebar items.

src/gatsby-theme-iterative-docs/sidebar.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const normalizeSidebar = require('../../plugins/gatsby-theme-iterative-docs/normalize-sidebar')
+const sidebar = require('../../content/docs/sidebar.json')
+module.exports = normalizeSidebar(sidebar)
```
rogermparent, line 4: This shadowed file definition must be present and default export a normalized sidebar. It also must be ES5 format to be used in `gatsby-node` contexts like remark plugins and server redirects- while those can't do shadows, they can still require this file directly.

yathomasi (commented):
Looks good 👍

julieg18 (approved):
Looks good!

shcheklein:
@rogermparent is it still a draft, should we merge this?

rogermparent:
> Should we merge this?

Sorry about the lack of elaboration, #3170 takes the place of this and #3143 because it was easier to restart from master due to how these earlier PRs changed imports in many components.

Issue #2980, opened 2021-10-28
Make Youtube Embeds GDPR compliant


julieg18:
Currently, we use Youtube embeds for videos on the site. But Youtube is using cookies to track user's info. If we don't want to have to ask a user's permission to play Youtube videos, we'll need to find an alternative to youtube. Youtube does offer `youtube-nocookie.com`, but that just places tracking info in local storage and adds cookies once the user presses play. 

cc @shcheklein @rogermparent

julieg18:
We could definitely use a different video platform, but we would be forced to copy over any youtube videos we want to post on our website over to that platform. And we would less traffic on our youtube channel 🤔 Maybe we should consider turning our youtube embeds into links to youtube? We could use the youtube API and CSS to make it look like a video (allowing us to not have to update our website layouts), but when you click on it opens a new tab to youtube. @iterative/websites, what do you think?

rogermparent:
> We could definitely use a different video platform, but we would be forced to copy over any youtube videos we want to post on our website over to that platform. And we would less traffic on our youtube channel thinking

Agreed, would be a rough time to port all videos we want to embed as well as trying to enforce no youtube embeds throughout the site.

> Maybe we should consider turning our youtube embeds into links to youtube? We could use the youtube API and CSS to make it look like a video (allowing us to not have to update our website layouts), but when you click on it opens a new tab to youtube.

I don't think this is quite what we'd want, I get the impression most people like to have the embeds within the site instead of bringing the user to another site which makes it more likely they'll bounce after.

Maybe we could re-tune our youtube-embeds such that the iframe src is not loaded until the user clicks a button? I found some inspiration [from this article](https://cookieplugins.com/en/embedding-gdpr-compliant-youtube-videos/) and threw together a little demo:

https://user-images.githubusercontent.com/9111807/147594671-99f8f993-07d3-4141-9c7e-fa18af8bd895.mp4

We'll have to provide our own background image if we want one, but this method of loading ensures that the child iframe can't even load if we don't want it to, ensuring that no user info can be stored and we stay GDPR compliant while keeping videos embedded in the website.

julieg18:
> Maybe we could re-tune our youtube-embeds such that the iframe src is not loaded until the user clicks a button?

[...]

PR #3136, opened 2021-12-28
Change youtube embed component to only load src on user request

rogermparent:
First mentioned in [this comment](https://github.com/iterative/dvc.org/issues/2980#issuecomment-1002224872)

https://user-images.githubusercontent.com/9111807/147594671-99f8f993-07d3-4141-9c7e-fa18af8bd895.mp4

This PR modifies our YouTube embed component such that the `src` is only set after the user clicks a button to watch the video. With this, the embedded site has no way of breaking GDPR compliance through cookies, localStorage, or any other API because the site isn't even loaded until the user wants it to be. I assume we'll probably need to add a message like "By playing this video you accept YouTube's cookies" to be truly GDPR compliant.

We'll have to add our own background images if we want them, but that shouldn't be too difficult given that we can just grab youtube's image and display it on the element- just not as easy to work with should we change the video in the future.

### DIFF
--- a/src/components/Home/UseCases/Video/index.tsx
+++ b/src/components/Home/UseCases/Video/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react'
+import React, { useState, useCallback, useRef } from 'react'
 
 import TwoRowsButton from '../../../TwoRowsButton'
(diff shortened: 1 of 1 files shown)


julieg18:
@rogermparent, are we good to close this since we now just need to look into adding a hover animation?

PR #3140, opened 2021-12-29
Interactive synopsis Fix #2024

### LINKED ISSUE #2024: cmd: make Synopsis/args/options more interactive
> Extracted from https://github.com/iterative/dvc.org/pull/1983#issuecomment-744081574

Needs:

- Special place/way to specify command arguments in pages of https://dvc.org/doc/command-reference
- link to them from the synopsis? (as well as to `--` options)
- Have anchors for each argument and option

yathomasi:
Feature mentioned on #2024 .
> Needs:
>
> - Special place/way to specify command arguments in pages of https://dvc.org/doc/command-reference
> - link to them from the synopsis? (as well as to -- options)
> - Have anchors for each argument and option

We may not actually need the different/special way to specify command arguments because current enclosing with enclosing brackets should be fine. But, we need to define some set of rules/guidelines so that we are consistent throughout the documentation and also it's would be great to define the parser.

- [x] ~~define some guidelines for arguments and options~~(no need as it's defined by argsparse on similar pattern)
- [x] parse synopsis code block and add anchor `#` tags to the arguments
- [x] make sure change in Code component doesn't affect existing syntax highlighting
- [x] parse options and add respective `id` for anchor tags

@iterative/websites @shcheklein @iterative/docs 

---

After merge:
- [ ] Change isolated option mentions throughout important docs so they auto-link e.g. from "...use the `--out` option of `dvc import`," to just "use `dvc import --out`."


src/components/Code/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return pipe(text, wrapUsage, wrapDvc, linkifyArgs)
+}
+
+const formatText = (text: string) => {
+  const wrapped = wrapWithTags(text)
+  return htmlToReact(wrapped)
+}
+
+const Code: React.FC<{ className: string }> = ({ className, children }) => {
+  if (className === 'language-usage') {
```
julieg18, line 40: Great idea reusing the classes placed by `gatsby-remark-prismjs` to find the code blocks that need linking!

julieg18 (commented):
Great work so far!

julieg18:
> We may not actually need the different/special way to specify command arguments because current enclosing with enclosing brackets should be fine. But, we need to define some set of rules/guidelines so that we are consistent throughout the documentation and also it's would be great to define the parser.

One idea that comes to mind is having our `/command-reference/RANDOM_COMMAND#options` section setup like our like [our glossary page](https://dvc.org/doc/user-guide/glossary), with each item in the list having a link attached to it. The links in the Synopsis section would take you the desired list item in the options section.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es5",
     "module": "commonjs",
     "resolveJsonModule": true,
-    "lib": ["dom", "es2015", "es2017"],
+    "lib": ["dom", "es2015", "es2017", "es2021"],
```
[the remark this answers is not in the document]
yathomasi: Its to use `String.replaceAll()`

src/components/Documentation/Markdown/Main/styles.module.css
```diff
@@ -70,6 +70,15 @@
       }
     }
 
+    a.token.args {
+      color: var(--color-light-blue);
+    }
+
+    a.token.args:hover {
+      outline: 1px solid green;
+      text-decoration: none;
```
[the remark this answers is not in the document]
yathomasi: Yes. It was just supposed to be for dev preview. It will be changed.

src/utils/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Parser as HtmlToReactParser } from 'html-to-react'
+
+const htmlToReactParser = new HtmlToReactParser()
+
+export function htmlToReact(html: any) {
+  return htmlToReactParser.parse(html)
```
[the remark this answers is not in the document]
yathomasi: Yes, we can use that. I assumed we needed to convert that to React component but turns out we don't.

src/components/Code/index.tsx
yathomasi: Yes, we needed to make sure the existing syntax highlighting worked.

yathomasi:
> One idea that comes to mind is having our `/command-reference/RANDOM_COMMAND#options` section setup like [our glossary page](https://dvc.org/doc/user-guide/glossary), with each item in the list having a link attached to it. The links in the Synopsis section would take you the desired list item in the options section.

Yeah, that's the idea. But, I am having a little bit issue with adding the HTML `id` attribute to the markdown AST.

julieg18:
> Yeah, that's the idea. But, I am having a little bit issue with adding the HTML id attribute to the markdown AST.

Feel free to message me in Slack if you want to discuss possible solutions!

yathomasi:
@iterative/docs 
Some assumptions here are:
 1. arguments are always enclosed in square brackets eg: `[-h]`
 2. arguments that need to be linked start with `-` or `--`. eg: `[-h] [--version]`
 3. Option definition on the list starts with the respective inline code block which also starts with `-` or `--` 
 Eg:
- `-h` 

 These are the assumptions taken and these need to be converted into guidelines/rules.
  For my quick overview look, I could not see edge cases for now. Please, mention them if there could be some.

casperdcl:
> These are the assumptions taken and these need to be converted into guidelines/rules.

It's just the output from `dvc ... --help` as generated by [argparse](https://docs.python.org/3/library/argparse.html) so I don't think any manual rules are needed.

gatsby-config.js
```diff
@@ -4,7 +4,7 @@ require('dotenv').config()
 const path = require('path')
 
 require('./config/prismjs/dvc')
-require('./config/prismjs/usage')
+// require('./config/prismjs/usage')
```
casperdcl, line 7: does this need deleting?

casperdcl (commented):
looking good

rogermparent:
screenshot

As mentioned in the meeting, we'll want to at least hide these icons until the user hovers over the list item, but the ideal would probably be adding links to the inline code blocks themselves. After this is handled, I think this PR will be good enough to merge as a first iteration!

There's also the issue of there possibly being a synopsis item which isn't documented in the options list, which adds a dead link to the synopsis. It'd take an order of magnitude more work to prevent this so I think it's fine for a first iteration, but it's worth mentioning for a possible rewrite in the future.

src/components/Documentation/Markdown/Main/styles.module.css
```diff
@@ -70,6 +70,10 @@
       }
     }
 
+    a.token.args {
+      color: var(--color-light-blue);
```
julieg18, line 74: A `a`'s `outline-color` is `--color-gray-hover` by default which isn't viewable on dark backgrounds. Adding `outline-color: var(--color-light-blue-hover);` here should allow keyboard users to see when these links are being focused on.

julieg18 (commented):
Looks great! Agreeing with @rogermparent's comments, this should be good to merge after we have the link icons appear on hover and we can open an issue later about making sure we don't make any [dead links](https://dvc-org-interactive-syn-zo8hwd.herokuapp.com/doc/command-reference/stage#-v) in the synopsis.

jorgeorpinel:
Hi. Thanks for addressing #2024. I'm not sure I understand what these changes do exactly. Could there be a summary?

> current enclosing with enclosing brackets should be fine

What enclosing brackets are we talking about here?

>  define some guidelines for arguments and options

What kind of guidelines? I think ideally just using backquote "code" blocks in markdown is the easiest for writers and if the engine can recognize when command options or other arguments are being used and link/or not as needed.

Related: https://github.com/iterative/dvc.org/issues/1614

jorgeorpinel:
> Could there be a summary?

Oh, I see. You are linking from synopsis blocks to arguments. Indeed this was a feature to consider (marked with question mark in #2024) but to be honest now that I see it in action I'm not sure it's that useful. No strong opinion, we could keep it. Please link the entire enclosed text though (including any argument):

image

jorgeorpinel:
The really useful feature would be to link from command mentions from other places right into the options section, for example in https://dvc.org/doc/command-reference/add there's a `dvc init --no-scm` which links to https://dvc.org/doc/command-reference/init (add `#--no-scm` in that case).

Also, plain mentions of options inside a cmd ref. could be auto-linked to that anchor e.g. in the same ref's description there's this mention of `--no-commit` which could be linked to `#--no-commit`:

image

> An exception to this would be option names that don't exist in the current reference (probably because they refer to some other command).

gatsby-config.js
yathomasi: Yes, because we are using custom component for Code block and the logic on `prismjs/usage` are also migrated to that custom Code component. 
We needed parsing the code block and it was relative easy to work with raw data than complex structure provided after allowing through prismjs.

yathomasi:
> What kind of guidelines? I think ideally just using backquote "code" blocks in markdown is the easiest for writers and if the engine can recognize when command options or other arguments are being used and link/or not as needed.

Okay, we don't need such guide lines. The engine will parse the arguments from the square brackets. And, for each option, the line would start with the `inline codeblock` which we are using now. So, basically it's exactly how it is there as of now.

yathomasi:
>Please link the entire enclosed text though (including any argument):

Sure, that's already been done.

gatsby-config.js
shcheklein: we could completely remove this prism plugin then? let's do this then

plugins/gatsby-remark-args-linker/index.js
```diff
@@ -0,0 +1,74 @@
...
+const argsRegex = new RegExp(/\-{1,2}[a-zA-Z-]*/, 'ig')
+
+function patch(context, key, value) {
+  if (!context[key]) {
+    context[key] = value
+  }
+
+  return context[key]
+}
+const svgIcon = `<svg aria-hidden="true" focusable="false" height="16" version="1.1" viewBox="0 0 16 16" width="16"><path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path></svg>`
```
shcheklein, line 12: is it a link icon? we should have this icon somewhere already - do we want to reuse it?

plugins/gatsby-remark-args-linker/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+module.exports = (
+  { markdownAST },
+  { icon = svgIcon, className = 'anchor', isIconAfterHeader = false }
+) => {
+  visit(
+    markdownAST,
+    node =>
+      node.type === 'listItem' &&
+      node.children.some(
```
shcheklein, line 22: will it match lists like "- some text `--flag`"? also does it mean that we math lists outside the `Options` section?

plugins/gatsby-remark-args-linker/index.js
```diff
@@ -0,0 +1,74 @@
...
+      if (String(value).startsWith('-')) {
+        const id = value.match(argsRegex)[0]
+        const data = patch(listItemNode, `data`, {})
+
+        patch(data, `id`, id)
+        patch(data, `htmlAttributes`, {})
+        patch(data, `hProperties`, {})
+        patch(data.htmlAttributes, `id`, id)
+        patch(data.hProperties, `id`, id)
+        if (icon) {
```
shcheklein, line 45: [Q] why do we need this `if`?

plugins/gatsby-remark-args-linker/index.js
```diff
@@ -0,0 +1,74 @@
+const visit = require('unist-util-visit')
+
+const argsRegex = new RegExp(/\-{1,2}[a-zA-Z-]*/, 'ig')
+
+function patch(context, key, value) {
+  if (!context[key]) {
```
shcheklein, line 6: better to do `in` here? (less questions regarding booleans, etc - I know it's probably less relevant here

src/components/Code/index.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
```
shcheklein, line 1: to some extent we are reimplementing prism here a bit (parsing custom DSL - dvc commands, usage, etc). I wonder if we could use prism hooks instead - https://prismjs.com/extending.html#writing-plugins ?

yathomasi:
> The really useful feature would be to link from command mentions from any other pages right into the corresponding option anchor. For example, in https://dvc.org/doc/command-reference/add there's a `dvc init --no-scm` which links to https://dvc.org/doc/command-reference/init . This would add `#--no-scm` to that URL.
> 
I have added it on last push. If it doesn't cover every cases then it would be better to create a separate issue.

> Also, plain mentions of options inside a cmd ref. could be auto-linked to that anchor e.g. in the same page there's this mention of `--no-commit`, which would be linked to `#--no-commit`. It's currently not linked:
> 
> image
> 
> > An exception to this would be option names that don't exist in the current reference (probably because they refer to some other command).

Regarding this, I think we need further discussion because this can have more exceptions. We can proceed after we are sure there are no such exception. Again, we can work on this on separate issue. It can be same with above one or also a separate one.

plugins/gatsby-remark-args-linker/index.js
yathomasi: Yes, it's a link icon. It's on the plugin we used for header. Not sure, we can use them.

plugins/gatsby-remark-args-linker/index.js
yathomasi: it matches list with 1st line start with inline Code block i.e `` and that starts with `-`.  Yes, If it exactly matches this then it's possible outside options too.  We can work on this more if it collide but I found that unique for my quick overlook over the docs.

plugins/gatsby-remark-args-linker/index.js
yathomasi: icon is taken as parameter for plugin so if we don't want to set it we can pass false or empty value on config.

plugins/gatsby-remark-args-linker/index.js
yathomasi: Sorry, which `in` ?

src/components/Code/index.tsx
yathomasi: I think we already had them.
I did so because I found it relatively easy to parse the raw data and add dvc commands, usage, etc along with links to args then working with prismjs encoded data.

plugins/gatsby-remark-args-linker/index.js
julieg18: We do have a list that matches the pattern in our DVCLive docs: https://dvc-org-interactive-syn-zo8hwd.herokuapp.com/doc/dvclive/dvclive-with-dvc#--live-no-cache. Not sure if it does any harm to have links there... Though if we wanted, we could have `gatsby-remark-args-linker` only look through pages in our `/command-reference/` directory 🤔

shcheklein: thanks, Julie and Thomas. q: how hard it is to make it strict and well-defined?

plugins/gatsby-remark-args-linker/index.js
shcheklein: let's simplify this, wdyt? we don't need to pass an icon to the plugin as far as I understand. We always use the same icon

plugins/gatsby-remark-args-linker/index.js
shcheklein: like `key in context` - would that work better and be a bit better defined?

src/components/Code/index.tsx
shcheklein: > I think we already had them.

yep, there are different ways to hook into prism though as far as I understand. Not 100% sure though. 

This way we are reimplementing prism itself a bit.

jorgeorpinel:
> > link from command mentions from any other pages...
>
> I have added it on last push.

Hmmm I don't see it working in https://dvc-org-interactive-syn-zo8hwd.herokuapp.com/doc/command-reference/add . The `dvc init --no-scm` link still goes to https://dvc-org-interactive-syn-zo8hwd.herokuapp.com/doc/command-reference/init (no anchor)

> If it doesn't cover every cases then it would be better to create a separate issue.

We already have https://github.com/iterative/dvc.org/issues/2024 though. I was just clarifying that issue's original intention.

Of course it can be another PR though. Seems like what's covered in this one so far is already a significant change to review code-wise.

> we need further discussion because this can have more exceptions

Agree. And again, relates to https://github.com/iterative/dvc.org/issues/1614 and to making the regex checks more strict in general (mentioned in some review above). Thanks

yathomasi:
@iterative/websites
For some options like 
- `-n` , `--name`
1. I tried adding `id` for second argument as well. At first I tried adding `id` to respective inline code block, which to my surprise didn't worked. So, instead I added first argument `id` to list i.e `li` and second argument (if any) to child paragraph i.e `p`. You can check [this last commit](https://github.com/iterative/dvc.org/pull/3140/commits/27960ac897fedf3fc247752cd75c027a5b425a26)

2. I have done and pushed the changes asked by @jorgeorpinel
>For example, in https://dvc.org/doc/command-reference/add there's a dvc init --no-scm which links to https://dvc.org/doc/command-reference/init . This would add #--no-scm to that URL.

I can see it's working locally but it's not showing on deployment. 

@julieg18  @rogermparent  can you please review these two points? Please mention if there are any optimization we want.

julieg18 (commented):
> I tried adding id for second argument as well. At first I tried adding id to respective inline code block, which to my surprise didn't worked. So, instead I added first argument id to list i.e li and second argument (if any) to child paragraph i.e p. You can check this last commit

Weird. I wonder if it has something to do with the inline code elements being on a 3rd level instead of a 2nd level like the paragraph or 1st level like the list item. 🤔 If @rogermparent doesn't have any ideas for solutions, maybe we could just merge how it is for now and look into it later.  

> I have done and pushed the changes asked by @jorgeorpinel...I can see it's working locally but it's not showing on deployment.

I deleted the app and rebuild it in Heroku. Looks to be working now!

jorgeorpinel:
> For example, in https://dvc.org/doc/command-reference/add there's a dvc init --no-scm ...

I can confirm it's working on the latest deployment 🙂 Thanks!

casperdcl (commented):
v. minor: removing the `-`/`--` prefixes from the URL might be nice (`#no-commit` not `#--no-commit`)?

rogermparent:
> If @rogermparent doesn't have any ideas for solutions, maybe we could just merge how it is for now and look into it later.

Nothing off the top of my head, I think looking into it later would be fine.

rogermparent:
Just found out the reason `code` blocks aren't taking an `id`: it's passed as a prop, but the custom `code` component isn't set up to handle that `id` prop.
Changing the code component to something like this makes it able to handle ids the same way other components do:
```tsx
const Code: React.FC<{ className: string; id?: string }> = ({
  className,
  id,
  children
}) => {
  if (className === 'language-usage') {
    const codeText = Array.isArray(children)
      ? children.join('')
      : String(children) || ''
    const ReactText = formatText(codeText)
    return (
      <code
        id={id}
        className={cn(className)}
        dangerouslySetInnerHTML={{ __html: ReactText }}
      ></code>
    )
  }
  return (
    <code id={id} className={cn(className)}>
      {children}
    </code>
  )
}
```

yathomasi:
> Just found out the reason `code` blocks aren't taking an `id`: it's passed as a prop, but the custom `code` component isn't set up to handle that `id` prop. 

I was also assuming some other Code component/plugin was interfering but didn't know it was the current Custom Code component I was working on 😃. I thought inline code blocks were handled on it's own code component.

Anyway thanks for finding @rogermparent . I will fix that.

yathomasi:
> v. minor: removing the `-`/`--` prefixes from the URL might be nice (`#no-commit` not `#--no-commit`)?

@casperdcl 
I think it's better with dashes because we need to set ids which needs to be unique on a page.
If we removed the dashes it become common words which might already be used on pages. So, there will be possibility of colliding. For eg:  
`--help` => `help` 
`--global` => `global`
and so on.

plugins/gatsby-remark-args-linker/index.js
yathomasi: Since we are using lodash, the `.has` utility function would be a better implementation.

plugins/gatsby-remark-args-linker/index.js
yathomasi: The possible way I could think of was to make it strict or well defined is by checking heading Options and looking into a consecutive list and to the list item. 
But, I found it difficult to do so with Markdown AST. We could easily achieve that if we had something easy/well-defined APIs/functions as in DOM manipulation. Otherwise, we need to come up with some hacky way to do so.

casperdcl:
> If we removed the dashes it become common words which might already be used on pages. So, there will be possibility of colliding. For eg: `--help` => `help` `--global` => `global` and so on.

Sure. Also

```diff
-https://dvc.org/doc/command-reference/exp/run#queue
+https://dvc.org/doc/command-reference/exp/run#--queue
```

I guess the longer version looks more like what you type in the CLI so perhaps better.

plugins/gatsby-remark-args-linker/index.js
rogermparent: I looked into this a little earlier, simple checks like the dir being in `command-reference` can be done with parts of the Gatsby remark plugin API that we don't currently use: the plugins are actually passed the Markdown node of the page in question as well as the capability to reach into other nodes like its File parent.

```ts
module.exports = api => {
  const { markdownAST, getNode, markdownNode } = api
  const parentNode = getNode(markdownNode.parent)
  const isDoc = parentNode.relativeDirectory.startsWith(
    'docs/command-reference'
  )
```

More granular checks like ensuring the header text is "Options" would require dipping into the AST, which is certainly doable but the implementation can border on hacky- might be worth looking into in the future, but limiting this to `command-reference` pages should be enough limitation for a first iteration.

Doing either of these would require us to make this linker plugin live outside of the main `flow` statement that only passes in the markdown AST, but I think that's a convention worth breaking.

shcheklein: sounds good, but it's also about replacing `node.children.some` with something like `node.children.first`

'docs/command-reference' - we need to make it a param? since it might be different across different websites later.

src/components/Code/index.tsx
rogermparent: Found that Prism has an [autolinker plugin](https://github.com/PrismJS/prism/blob/master/plugins/autolinker/prism-autolinker.js) that shows how Prism hooks can be used to add links. I think it could be worth the extra work of converting this React-based children replacement since what's going on here really is quite similar to Prism.

yathomasi:
@rogermparent 
Finally ended up using a prism hook instead of a custom component.
@julieg18 can you verify if there are any edge cases?

config/prismjs/usage.js
```diff
@@ -9,5 +9,13 @@ Prism.languages.usage = {
   },
   usage: {
     pattern: /(^|\n)\s*(usage|positional arguments|optional arguments)/
+  },
+  args: {
+    pattern: /(?<=\[)(?:[^\]\[]+|\[(?:[^\]\[]+|\[[^\]\[]*\])*\])*(?=\])/,
+    inside: {
+      arg: {
+        pattern: /\-{1,2}[a-zA-Z- <>\]\[.]*/
```
julieg18, line 17: Some of our option arguments are more complex than this. ([Example in `/exps/show`](https://dvc-org-interactive-syn-s47nhu.herokuapp.com/doc/command-reference/exp/show)) We might want to consider expanding this regex to include `{}`, `/`, `,`, and `_`.

image

plugins/gatsby-remark-dvc-linker/commandLinker.js
```diff
@@ -5,31 +5,29 @@ const { getItemByPath } = require('../../src/utils/shared/sidebar')
...
+const ARGS_REGEXP = new RegExp(/\-{1,2}[a-zA-Z-]*/, 'ig')
 const COMMAND_ROOT = '/doc/command-reference/'
 
 module.exports = astNode => {
   const node = astNode[0]
   const parent = astNode[2]
-
   if (parent.type !== 'link' && DVC_REGEXP.test(node.value)) {
     const parts = node.value.split(/\s+/)
+    const baseUrl = `${COMMAND_ROOT}${parts[1]}`
```
julieg18, line 16: Not 100% what changed in this plugin to cause this but this phrase in [`/config`](https://dvc-org-interactive-syn-s47nhu.herokuapp.com/doc/command-reference/config#--system) is now a broken link ([in dvc.org](https://dvc.org/doc/command-reference/config#command-options-flags), this phrase isn't a link at all):

image

julieg18 (commented):
Looks good! I did notice some minor things, but I think we could still merge this and fix them on a later iteration.

plugins/gatsby-remark-args-linker/index.js
julieg18: > sounds good, but it's also about replacing node.children.some with something like node.children.first

This would check the first child in the list, right? While this would almost all use cases [`/repro`](https://dvc-org-interactive-syn-s47nhu.herokuapp.com/doc/command-reference/repro) options starts with `targets`:

image

Is the only one as far as I know, though :)

plugins/gatsby-remark-args-linker/index.js
rogermparent: This is how `gatsby-remark-autolink-headers` does icons as well, though we could make one that can be shared everywhere it's used so we can ensure consistency.

config/prismjs/usage.js
yathomasi: Thanks for pointing this out. I have simply generalized it to include all.

plugins/gatsby-remark-dvc-linker/commandLinker.js
yathomasi: We have updated the command linker as per the requirement. I have now fixed the logic and the above command should be linkable as `/config#--system`.

plugins/gatsby-remark-args-linker/index.js
yathomasi: I added one to the static folder and passed it as param on both `gatsby-remark-autolink-headers` and `gatsby-remark-args-linkers` for consistency.

julieg18 (approved):
Looks awesome! Reading through all of the comments, there is still a couple things to do but we could create a issue listing the rest of the tasks to do later. If everyone else agrees, I think we could merge this!

rogermparent (approved):
Definitely looks good enough to me to be merged as a first iteration!

shcheklein (changes requested):
Looks good to me! Good stuff with the prism hook - less code == better sleep :)

Two small things. First, could we please move icon a bit down to align it better with the bullets?

Screen Shot 2022-01-10 at 4 18 39 PM

And please still review and resolve the existing comments (it's totally fine if you think that it's not the right time to address, or you want to keep certain things the way it is, but it's a good practice to react and say that we are moving forward).

plugins/gatsby-remark-args-linker/index.js
yathomasi: This one was also kind of discussed in another conversation as well. We use `gatsby-autolink-headers` as well. So, for consistency, we defined the link-icon separately and passed that icon as parameters on both.

src/components/Code/index.tsx
yathomasi: Mentioned on comment, did that extra work, and ended up using prism hooks.

plugins/gatsby-remark-args-linker/index.js
yathomasi: Oops sorry, I missed this conversation.

yathomasi: I have added the params for paths which can be a string(path) or an array of paths. For now, I have set it to `docs/command-reference`.

yathomasi: > sounds good, but it's also about replacing `node.children.some` with something like `node.children.first`

I have made some changes and pushed.

yathomasi:
We had headings icons aligned similarly, so I had left it as it was. 
I have just made the change and this looks nice with bullet points.
Screen Shot 2022-01-11 at 15 00 11
If we want more aesthetic changes, we can continue on another iteration.

Similarly, looks like I had missed another conversation. I have gone through and updated them with the necessary changes. 

Here is the listing we would need discussion or move to another iteration:
  - plain mentions of options inside a cmd ref. could be auto-linked to that anchor e.g. mention of `--no-commit`, which would be linked to #--no-commit 
  - making `args-linker` more strict to the Options section
  - making aesthetic/UI changes for the anchor tags on synopsis or anchor icon on list options.
  
I think these are all, please mention if I have missed any.

shcheklein (approved):

shcheklein:
Thanks @yathomasi !

PR #3143, opened 2021-12-29
Add default remark plugin definition to theme

rogermparent:
Extracted from #3133

This PR adds a default remark transformer definition to our Gatsby theme. It's actually the one from cml.dev and isn't used on dvc.org since dvc.org has some special plugins that cml.dev doesn't, but it seems like the CML version is the one we'd use on most sites anyway.

Ideally, this remark definition can be used on most of our sites that will use this theme.


plugins/gatsby-theme-iterative-docs/index.js
```diff
@@ -0,0 +1 @@
+// noop
```
rogermparent, line 1: Just something I noticed in other gatsby plugins/themes that seems to be standard practice if you're not using index.js.

julieg18 (approved):
Looks good!

shcheklein:
@julieg18 @rogermparent why this is not merged still? :)

rogermparent:
> why this is not merged still?

Same as [#3133](https://github.com/iterative/dvc.org/pull/3133#issuecomment-1019696228), #3170 takes its place.

PR #3144, opened 2021-12-29
Move ToggleProvider to theme

rogermparent:
This PR simply moves the `ToggleProvider` component to the theme. This component is identical in both dvc.org and cml, so no configuration is needed.


julieg18 (approved):
Looks good!

rogermparent:
Closing because we'll be moving the whole `Documentation` component to this theme

PR #3147, opened 2022-01-03
dvclive: Updates for implicit no step

daavoo:
P.R. for https://github.com/iterative/dvclive/pull/203


dberenbaum (commented):
Looks good! Can you also update https://dvc.org/doc/dvclive/api-reference/live/log to mention the metrics summary and/or `no_step`?

jorgeorpinel:
A bit confused by the concept of DVCLive without steps (or why that addresses https://github.com/iterative/dvclive/issues/182). Plus I see there's a discussion on the semantics in https://github.com/iterative/dvclive/pull/203#issuecomment-1005066518 so I'll wait until when/if you request my review to take a closer look. Thanks

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -54,6 +55,11 @@ files.
   _metrics summary_ (usable by `dvc metrics`). The _summary_ will be located at
   `{path}.json`. _Default_: `True`.
 
+- `no_step` - If `True`, DVCLive will **not** generate
+  [_metrics logs_](/doc/dvclive/get-started#metrics-logs),
+  [_html report_](/doc/dvclive/dvclive-with-dvc#html-report) nor
+  [_checkpoints_](/doc/dvclive/dvclive-with-dvc#checkpoints)
```
jorgeorpinel, line 61: I do have one generic note on existing DVCLive docs writing style: It's preferable to avoid manual styling of text even when it is a feature in Markdown. Especially of link elements which already have CSS styling, but other italics and bolds should be pretty rare if any at all. When they're common (e.g. on every "not") it kind of defeats their purpose of applying special emphasis ("will/does/can not" is already emphatic). I also know italics can be used on certain key terms but maintaining this kind of rule manually is nearly impossible -- if needed it should be automated.

jorgeorpinel: p.s. I'm not saying we should un-do existing styling since it's already in though. But maybe consider stopping applying it and create an issue to remove it later if you agree.

daavoo:
> A bit confused by the concept of DVCLive without steps (or why that addresses [iterative/dvclive#182](https://github.com/iterative/dvclive/issues/182)).

One of the outputs of DVCLive are linear plots, containing metric values associated to each iteration (step) of the training process.

When exploring DVCLive for ML scenarios where there are no iterations we realize that these outputs don't have any use so in iterative/dvclive#203 we introduce an implicit `no step` mode. 

This makes DVCLive to don't generate linear plots unless `next_step` or `set_step` is called in the code, addressing iterative/dvclive#182.

content/docs/dvclive/api-reference/live/log.md
```diff
@@ -19,10 +16,16 @@ live.log("loss", 0.9)
 
 ## Description
 
-The first call to `live.log(name, val)` will create a new file in
-`{path}/{name}.tsv` including the header and first row.
+If `summary` is True, `Live.log()` DVCLive will update the
```
dberenbaum, line 19:
```suggestion
If `summary` is True, `live.log(name, val)` will update the
```

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -51,7 +47,7 @@ files.
   `path` directory and start from that point. _Default_: `False`.
 
 - `summary` - If `True`, upon each `Live.log()` call, DVCLive will generate a
-  _metrics summary_ (usable by `dvc metrics`). The _summary_ will be located at
+  summary (usable by `dvc metrics`). The summary will be located at
   `{path}.json`. _Default_: `True`.
```
dberenbaum, line 51: What do you think about a warnings to not change `resume` or `summary` in no-step scenarios?

dberenbaum (changes requested):
Lots of comments since this is a big change.

This also probably requires an update to the dvclive readme. We probably need some process to keep that updated.

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -23,25 +21,23 @@ live = Live()
...
+Initializes a DVCLive logger.
 
 ⚠️ If `path` already exists, `Live()` will remove all existing DVCLive related
 files.
 
 ## Attributes
 
-- `dir` - Location of the
-  [_metrics logs_](/doc/dvclive/get-started#metrics-logs) directory.
+- `dir` - Location of the [linear plots](/doc/dvclive/get-started#linear-plots)
```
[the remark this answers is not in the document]
daavoo: Indeed, I have a draf P.R. that introduces "data" (both images and plots from https://github.com/iterative/dvclive/pull/189) revisiting all this sections

content/docs/dvclive/get-started.md
```diff
@@ -16,25 +17,37 @@ live = Live()
...
+
+### No step
+
+In ML scenarios where there are no iterations (steps) as part of the training
+process, there is no need to update the step number.
+
+When you don't update the step number in your code, DVCLive will **not**
+generate:
+
+- [linear plots](/doc/dvclive/get-started#linear-plots)
```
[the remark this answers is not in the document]
daavoo: > - What was wrong with `metrics logs`?

I just find `metrics logs` or the other alternatives are a little abstract. I think that "data series", "scalar plots", or line/linear plots better indicate what it's being saved.

content/docs/dvclive/get-started.md
```diff
@@ -1,12 +1,13 @@
 # Get Started
 
-DVCLive is a simple Python library whose interface consists of three main steps.
+DVCLive is a simple Python library for logging data in a format usable by DVC.
```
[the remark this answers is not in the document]
daavoo: Updated

content/docs/dvclive/api-reference/live/index.md
daavoo: Added warning. TBH I would remove `summary` as I don't think it'd be much trouble for users if we make the summary mandatory

content/docs/dvclive/api-reference/live/index.md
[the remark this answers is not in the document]
daavoo: https://github.com/iterative/dvc.org/pull/3207

content/docs/dvclive/get-started.md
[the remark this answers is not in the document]
dberenbaum: Without `dvc plots`, they are `.tsv` files, and that's what's actually being saved. IMO plots is a bit unnatural and DVC-specific. It also seems to me like the key feature of these are that they capture the metrics over time at each step, showing training progression. What would describe that? What about something like "step `.tsv` files"? I don't really like it but not sure what's better.

Do you feel the same about "metrics summary"? That seems similarly abstract to me.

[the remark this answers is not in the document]
daavoo: > It also seems to me like the key feature of these are that they capture the metrics over time at each step, showing training progression. What would describe that?

Metrics history? 

> Do you feel the same about "metrics summary"? That seems similarly abstract to me.

I don't know why but I don't feel the same for summary. Maybe because I can't think of alternatives 😅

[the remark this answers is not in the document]
dberenbaum: I think "metrics history" is pretty good. What do you think?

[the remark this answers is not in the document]
daavoo: Updated with metrics history. Overall I like how it reads

content/docs/dvclive/api-reference/live/get_step.md
```diff
@@ -1,7 +1,5 @@
 # Live.get_step()
 
-Returns the current `step` value.
```
jorgeorpinel, line 3: Just curious: why did you decide to remove these short descriptions on top? It's kind of nice to read what something does right from the start IMO but maybe I'm biased bc/ that's what we do in all `dvc` refs.

jorgeorpinel: Also the longer Definition above isn't super specific, it talks about this fn. very generally along with `set_step`. Should be more direct I think (esp. if we're removing these shorter ones).

content/docs/dvclive/get-started.md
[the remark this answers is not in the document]
jorgeorpinel: I don't see how "metrics logs" is abstract. It's pretty concrete and specific to what those TSV files are. It's just a bit confusing in the current Get Started since there's a "Log Metrics" section and later a "Metrics Logs" subsection. Probably there's no need for those subsections (under **Outputs**): just use paragraphs, then the terms don't become titles. "Metrics logs" does also read funny because of the final s in both words though so maybe be more descriptive like "TSV log files of your metrics" or something like that.

"Metrics history" to me sounds like it could relate to the Git history i.e. how metrics have evolved between commits.

content/docs/dvclive/dvclive-with-dvc.md
```diff
@@ -85,11 +85,12 @@ dvc.lock  training_metrics       train.py
 dvc.yaml  training_metrics.json
 ```
 
-The `.tsv` files generated under `training_metrics` can be visualized with
-`dvc plots`.
+The [metrics history](/doc/dvclive/get-started#history)) generated under
```
dberenbaum, line 88:
```suggestion
The [metrics history](/doc/dvclive/get-started#history) generated under
```

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -50,10 +46,16 @@ Its parameters are focused on configuring the behavior of subsequent calls to
 - `resume` - If `True`, DVCLive will try to read the previous `step` from the
   `path` directory and start from that point. _Default_: `False`.
 
+  ⚠️ Don't change the default value unless you are going to update the `step`
+  value during training.
```
dberenbaum, line 50: Maybe it can be explained more. For example: "If you are not using `steps`, don't use `True` since DVCLive will preserve the previous run's files to try to read the step value."

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -50,10 +46,16 @@ Its parameters are focused on configuring the behavior of subsequent calls to
...
+
 - `summary` - If `True`, upon each `Live.log()` call, DVCLive will generate a
-  _metrics summary_ (usable by `dvc metrics`). The _summary_ will be located at
+  summary (usable by `dvc metrics`). The summary will be located at
   `{path}.json`. _Default_: `True`.
 
+  ⚠️ Don't change the default value unless you are going to update the `step`
+  value during training.
```
dberenbaum, line 57: Also could explain more here: `If you are not using `steps`, don't set to `False` since the _summary_ will be the only record of your metrics.`

content/docs/dvclive/api-reference/live/log.md
```diff
@@ -19,10 +16,16 @@ live.log("loss", 0.9)
...
 ## Description
 
-The first call to `live.log(name, val)` will create a new file in
-`{path}/{name}.tsv` including the header and first row.
+If `summary` is True, `live.log(name, val)` will update the
+[summary](/doc/dvclive/get-started#summary) with the latest value logged.
+
+💡 The summary `{path}.json` is usable by `dvc metrics`.
+
+### Step updates
```
dberenbaum, line 24: Overall, the `step` info seems a bit scattered. Maybe we can move this section into get started, possibly make it more high level, and then reference to it throughout the docs when discussing steps (like `summary` and `history`).

content/docs/dvclive/api-reference/live/get_step.md
daavoo: Makes sense, thanks! Updated

content/docs/dvclive/api-reference/live/log.md
daavoo: I kind of have gone in the opposite direction in #3207 because the implications of updating `step` vary across the type of data being logged.

I removed `summary` / `history` info from `Outputs` section in favor of:


> The contents of the dvclive folder and dvclive.json would vary depending on the type of data you have logged and whether you have updated the step value or not. 
> See Live.log(), Live.log_image() and Live.log_plot() for more details.

content/docs/dvclive/api-reference/live/index.md
daavoo: Added but with some rephrasing please TAL

content/docs/dvclive/api-reference/live/index.md
daavoo: Added but with some rephrasing please TAL

content/docs/dvclive/get-started.md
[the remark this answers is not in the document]
daavoo: For me, `history` makes more sense than `logs` when combined with the other output, named `summary`. It feels like `summary` and `history` are "connected".

`history` is also a term [used in Keras ](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/History) with similar meaning as here, so data scientist might be already familiar with it.

> Probably there's no need for those subsections (under **Outputs**): just use paragraphs, then the terms don't become titles.

In subsequent P.R. I will remove the outputs subsections https://dvc-org-dvclive-data-hpgunbngf.herokuapp.com/doc/dvclive/get-started#outputs

content/docs/dvclive/api-reference/live/log.md
dberenbaum: Sounds good, we can revisit in that PR.

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -50,10 +49,16 @@ Its parameters are focused on configuring the behavior of subsequent calls to
 - `resume` - If `True`, DVCLive will try to read the previous `step` from the
   `path` directory and start from that point. _Default_: `False`.
 
+  ⚠️ If you are not using steps, don't set to `True` since DVCLive will preserve
+  existing files from previous run's and assume that `step` has been enabled.
```
dberenbaum, line 53:
```suggestion
  previous run's files and assume that `step` has been enabled.
```
Minor style edit.

content/docs/dvclive/api-reference/live/next_step.md
```diff
@@ -1,7 +1,5 @@
 # Live.next_step()
 
-Signals that the current step has ended and increases `step` value by 1 (one).
```
dberenbaum, line 3: Per the previous comment by @jorgeorpinel, why get rid of this? I find it a useful summary.

content/docs/dvclive/api-reference/live/set_step.md
```diff
@@ -1,7 +1,5 @@
 # dvclive.set_step()
 
-Signals that the current step has ended and sets `step` to the given value.
```
dberenbaum, line 3: Same as above. Why get rid of this?

dberenbaum:
Looks good! Just checking on previous comment from @jorgeorpinel about keeping the summaries at the top of the api methods.

content/docs/dvclive/api-reference/live/set_step.md
daavoo: That's odd I thought I've added back all short descriptions

dberenbaum (approved):

PR #3149, opened 2022-01-04
Fix broken tab order on docs sidebar menu

julieg18:
* Turn the menu's dropdown bullet points into buttons, allowing keyboard users to tab over them (if you try to tab through a docs page inside dvc.org, you'll get stuck on a dropdown)

Part of #1149

### DIFF
--- a/src/components/Documentation/Layout/SidebarMenu/index.tsx
+++ b/src/components/Documentation/Layout/SidebarMenu/index.tsx
@@ -75,11 +75,6 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
     onClick(isLeafItem)
   }
 
-  const bulletIconClick = (event: SyntheticEvent<HTMLSpanElement>): void => {
-    event.preventDefault()
-    setIsExpanded(!isExpanded)
-  }
-
   // Fetch a special icon if one is defined
   const IconComponent = icon && ICONS[icon]
   const iconElement = IconComponent ? (
@@ -131,13 +126,7 @@ const SidebarMenuItem: React.FC<ISidebarMenuItemProps> = ({
         {iconElement ? (
           iconElement
         ) : (
-          <span
-            className={bulletIconClassName}
-            onClick={bulletIconClick}
-            onKeyDown={bulletIconClick}
-            role="button"
-            tabIndex={0}
-          ></span>
+          <span className={bulletIconClassName}></span>
         )}
         {label}
       </Link>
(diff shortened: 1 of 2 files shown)


rogermparent (approved):
Awesome! Huge accessibility gains from this small change.

shcheklein:
It breaks the existing UX logic though? I can now toggle the menu section (by clicking on a bullet) w/o activating it.

julieg18:
> It breaks the existing UX logic though? I can now toggle the menu section (by clicking on a bullet) w/o activating it.

@shcheklein, could you explain further what's broken? As far as I could tell, everything works the same as dvc.org 🤔

shcheklein:
Open https://dvc.org/doc
Click the Use Cases arrow bullet point in the side bar menu
Website should stay on the "DVC Documentation" page

with the current changes - Use Cases page is activated.

I hope that explains it. Please let me know if it's still not clear. I can probably record the vide.

julieg18:
> I hope that explains it. Please let me know if it's still not clear. I can probably record the vide.

Oh, I didn't realize we had that feature! Cool! I'll update this code so that we can keep that and allow users to tab over it!

julieg18:
Ok, the dropdown bullet icons are now buttons! You can click on them (allowing you to open the dropdown without clicking on the link directly) and keyboard users can tab over or click on them as well.

shcheklein:
Thanks Julie, we now have a minor problem with the border.

Does it help with accessibility to have two separate elements like this, btw? Should we skip them on tab action?

(overall, I think we should fix borders in general - defaults look too ugly)

Screen Shot 2022-01-04 at 11 14 35 AM

julieg18:
> Thanks Julie, we now have a minor problem with the border.

Should be easy to fix this! 

> Does it help with accessibility to have two separate elements like this, btw? Should we skip them on tab action?

I think it helps. Web standards say that clickable elements should be clickable with keyboard as well. 

> (overall, I think we should fix borders in general - defaults look too ugly)

I'll open an issue! Though our options for changing outline are limited to the color, thickness, and type.

rogermparent:
Still some slight overflow issues with focused bullets, particularly when also mousing over the bullet:

normal screenshot
mouseover screenshot

The normal state could be fixed with just a few pixels more in the direction it was in before. The hover state is niche enough to possibly be ignored, but if we do want to fix it I could see using a color darkening effect instead of a size transform solving the issue.

julieg18:
> Still some slight cropping issues on the left outline of focused bullets for me on Firefox

Nice catch! I took off the offset entirely since it was already at 1px

> The hover state is niche enough to possibly be ignored, but if we do want to fix it I could see using a color darkening effect instead of a size transform solving the issue.

I vote we keep the scale increase even if it messes with the outline on hover since the area can be hard to click on as is.

yathomasi (approved):
Great  Work @julieg18 
It's working fine. Regarding outline on hover, I quickly tried adding z-index but looks like it doesn't work. Still, I think it's okay as it is.

PR #3152, opened 2022-01-05
Use `youtube-nocookie` domain in Video component

julieg18:
Part of #2980

### DIFF
--- a/src/components/Home/UseCases/Video/index.tsx
+++ b/src/components/Home/UseCases/Video/index.tsx
@@ -37,7 +37,7 @@ const Video: React.FC<{ id: string }> = ({ id }) => {
           title="Video"
           width="560"
           height="315"
-          src={`https://www.youtube.com/embed/${id}?rel=0&amp;controls=0&amp;showinfo=0;${
+          src={`https://www.youtube-nocookie.com/embed/${id}?rel=0&amp;controls=0&amp;showinfo=0;${
             isWatching ? `&autoplay=1` : ''
           }`}
           frameBorder="0"


shcheklein:
Interesting, in the incognito mode I local storage is clean and I see two cookies `NID` and `1P_JAR` 🤔

rogermparent (approved):
Looks good! This should handle all our embeds since `gatsby-remark-embedder` uses `nocookie` automatically.

shcheklein:
I wonder also if the set different things in different countries/regions.

rogermparent:
Just grabbed a random GDPR checker, assuming it checks from EU, and found something interesting:

https://www.cookiemetrix.com/display-report/dvc-org-use-youtube-noc-nchy4d.herokuapp.com/d31ca1f42e90bf8768df44de574e4ee3

There's one stored cookie: `CONSENT` from `youtube-nocookie.com` with the value `PENDING+342`. I'd imagine this specific cookie could be exempt since it isn't tracking the user, just a setting, and it does serve a technical purpose in remembering whether the user has consented to cookies.

shcheklein:
Also things to keep in mind (regarding different regions and countries).

For example, when I open youtube.com itself in the empty newly installed Firefox - it doesn't ask anything at all regarding cookies (and I'm pretty sure they have to be 100% compliant with CA laws). I wonder what happens if we open it from Europe.

Just saying that we should be ideally doing checks from some European country with an empty browser (user is not logged in into Youtube and didn't accepted anything).

shcheklein:
Tested clean Safari and Firefox from CA - no cookies, only some non-essential local store things like:

Screen Shot 2022-01-05 at 9 39 51 AM

Can't test completely clean Chrome. And can't test from Europe now.

But so far it looks reasonable for me.

PR #3157, opened 2022-01-07
Add `WhatsNewModal` Image Check

julieg18:
* Have the modal check if image exists before adding `GatsbyImage`. Otherwise a build will fail if a main blog image isn't provided

### DIFF
--- a/src/components/Home/WhatsNewModal/index.tsx
+++ b/src/components/Home/WhatsNewModal/index.tsx
@@ -91,12 +91,14 @@ const WhatsNewModal: React.FC = () => {
               <h2>{latestPost.title}</h2>
             </Link>
             <Link href={latestPost.slug}>
-              <GatsbyImage
-                className={styles.image}
-                alt=""
-                objectFit="contain"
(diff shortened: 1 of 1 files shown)


yathomasi (commented):
Looks good 👍

yathomasi (approved):

src/components/Home/WhatsNewModal/index.tsx
```diff
@@ -91,12 +91,14 @@ const WhatsNewModal: React.FC = () => {
...
-              />
+              {latestPost.picture && (
+                <GatsbyImage
+                  className={styles.image}
+                  alt=""
+                  objectFit="contain"
+                  image={latestPost.picture.gatsbyImageData}
+                />
+              )}
             </Link>
```
rogermparent, line 102: Should we put the conditional on the `Link` component so we don't render an empty link?

julieg18: Nice catch! I'll update the code!

rogermparent (approved):
Looks good!

Issue #1238, opened 2020-05-05
blog: make meta image URLs stable


shcheklein:
Right now when we change some libraries, setting, redeploy, etc, image URLs change and break our forum topics, for example:

https://discuss.dvc.org/t/dvc-heartbeat-feburary-20/318

We need to make some "proxy" URLs that will be stable and depend only on a post slug.

julieg18:
This should be doable! Technically, we already have our blog images saved on a stable url. ([December 2021 Community Gems Image](https://dvc.org/uploads/images/2021-12-21/dec-community-gems.png)) But we would want our meta images to be smaller than that... Two solutions I could think of so far:
* Running some kind of post build script that creates blog images with a fixed width (currently we have our meta blog images at 850) and saves them to a fixed place on a site. (`/blog/DATE_OF_POST/IMAGE_NAME` for example)
* Write some redirect code that redirects from a stable url to the image url (`/blog/NAME_OF_BLOG_POST/main-img` redirects to the unstable image link)

julieg18:
Update! I looked into redirecting the meta image urls, but couldn't think of way to accomplish this with the way our app is set up. 🤔 But this solution doesn't involve recreating any images, so if anyone has any ideas for this, please let me know!

As for my other solution idea, I looked into editing the blog images and giving them fixed url paths. The only way to give an image a fixed path in Gatsby (as far as I know) is to place images in the `static` folder before gatsby starts. There are probably a couple ways to edit and save images in the `static` folder before a gatsby build, but I tried one way in my draft pr, #3161. 

#3161 has a prebuild script that runs in production. When it runs it uses `nodejs` packages and a package that works with images called `Jimp` to take all images in the `static/uploads/images`, create a resized version with a fixed width (800px), and saves them in a `/static/blog` folder. Then gatsby copies over all the images in the `static` folder to `public` folder and our blog posts have access to 800px images with a fixed path. While this works, we are forced to edit and save a whole new set of blog images on our website. 

What do you think, @iterative/websites? Any other ideas for easier solutions?

iesahin:
How Gatsby determines the generated image URLs? If it's not random and can be customized, it may be easier to set a URL/pathname formatting option.

julieg18:
> How Gatsby determines the generated image URLs? If it's not random and can be customized, it may be easier to set a URL/pathname formatting option.

[...]

PR #3161, opened 2022-01-07
Test blog meta image path update

julieg18:
Testing solution for #1238

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,9 +1,66 @@
 require('dotenv').config()
 
+const { statSync } = require('fs')
+const { readdir } = require('fs/promises')
+const path = require('path')
+const Jimp = require('jimp')
+
 const { setPageContext } = require('./src/gatsby/common')
 
 const models = require('./src/gatsby/models.js')
 const callOnModels = require('./src/gatsby/utils/models')
+const { isProduction } = require('./src/server/utils')
+
+exports.onPreBootstrap = async () => {
+  if (!isProduction) {
+    return
+  }
+
+  const getDirectories = async path => {
+    try {
+      const files = await readdir(path)
+      const directoryPaths = files.filter(file =>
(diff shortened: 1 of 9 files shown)


gatsby-node.js
```diff
@@ -40,3 +44,110 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+
+//         return mkDirIfItDoesNotExist(`./public/blog/images/${date}`)
+//       })
+//     )
+
+//     await Promise.all(
+//       nodes.map(async node => {
+//         const imagePath = node.picture.gatsbyImageData.images.fallback.src
+//         const date = node.date.slice(0, 10)
+//         return copyFile(
```
julieg18, line 99: I commented out my original solution which involved copying over the images created by `gatsby-plugin-image` to a new place in the `public` folder. At the time `copyFile` runs, gatsby hasn't created all of the images yet so you get the `can't find directory or file` error on some of the images. Maybe we could look into having Gatsby creating the images earlier, guaranteeing they are already built when we run the `copyFile` code?

gatsby-node.js
```diff
@@ -40,3 +44,110 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+            date
+          }
+        }
+      }
+    `)
+
+    await Promise.all(
+      nodes.map(async node => {
+        const imagePath = node.picture.fields.sourcePath
+        return Jimp.read(path.join(__dirname, 'static', 'uploads', imagePath))
```
julieg18, line 139: Since I couldn't use `copyFile`, I used `Jimp` to take the main images and copy them over to the public folder. But if anyone has any ideas on getting the commented solution to work, it would be probably be better to just use `gatsby-plugin-image` and not using `Jimp` at all.

julieg18:
Closing in favor of #3206

PR #3162, opened 2022-01-10
redirects: add man.dvc.org/completion

dtrifiro:
Fixes 404 on https://man.dvc.org/completion

### DIFF
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -56,6 +56,7 @@
   "^/doc/user-guide/dvcignore$                                                            /doc/user-guide/project-structure/dvcignore-files",
   "^/doc/understanding-dvc(/.*)?$                                                         /doc/user-guide/what-is-dvc",
   "^/doc/commands-reference(/.*)?$                                                        /doc/command-reference$1",
+  "^/doc/command-reference/completion$                                                    /doc/install/completion",
   "^/doc/command-reference/plot$                                                          /doc/command-reference/plots",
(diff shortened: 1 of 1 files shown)


iesahin (approved):
Thank you Daniele.

shcheklein:
@dtrifiro thanks! out of curiosity - can we pls provide more context to the PR? where do we have this link in DVC?

dtrifiro:
Hey, this comes from `dvc completion --help`:
image

shcheklein:
Thanks, makes sense. Should we then have a CR page - https://dvc.org/doc/command-reference/completion? We can link from it the guide in Install ... also we need to add syntax highlighter for `dvc completion` I guess. cc @jorgeorpinel @iesahin

jorgeorpinel:
> Should we then have a CR page - https://dvc.org/doc/command-reference/completion?

Due to this PR that URL has been permanently redirected to https://dvc.org/doc/install/completion :/ we should undo this ASAP if we want to have the cmd ref. And I agree we should! Not sure how that slept through (probably a while ago). @dberenbaum WDYT?

jorgeorpinel:
> For now I would make the redirect temporary (302 code) to avoid permanent 301 cache.

Made https://github.com/iterative/dvc.org/pull/3172

Cc @iesahin please keep in mind the nature or permanent redirects before merging redirect changes 🙂 We should carefully consider which ones go 302 or even 303 I think.

jorgeorpinel:
And/or... Maybe the engine should default to 302 and we make 301s explicit?

PR #3167, opened 2022-01-10
Added january events

flippedcoder:
Adding the upcoming events we have for January.


shcheklein (changes requested):
Thanks for the update @flippedcoder ! Can we make the python web conf image smaller / run it through tinypng? It's 0.5MB which looks too much.

jendefig (approved):

shcheklein:
I assume this is outdated @jendefig @flippedcoder ? Do we need this? If not - let's close, if yes - please let's discuss how is responsible for moving things forward when there is no response for a review for a while. wdyt?

flippedcoder:
@shcheklein I lost track of this PR so sorry about that. We'd still like this merged in since there are some upcoming events and I've updated the image.

shcheklein (approved):

shcheklein:
Thanks @flippedcoder , merged.

PR #3170, opened 2022-01-10
Move Documentation components to theme

rogermparent:
This PR is a bottom-up moving of the `Documentation` component, and any components it depends on (and there's a lot) to the `gatsby-theme-iterative-docs` plugin.


julieg18 (commented):
It's looking good so far! Are you still getting that weird error you mentioned or have you found a fix for it? If you're still getting it, where/when are you receiving it?

rogermparent:
Ah sorry, I only mentioned the fix in a standuply message: it was just Node 17, and 17 caused issues on master. Using Node 16 fixed it!

yathomasi (commented):
Looking good. 
Are we planning to pack this into an npm package? Do we require updating the current package.json on `gatsby-theme-iterative-docs` with all the packages used ?

rogermparent:
> Looking good. Are we planning to pack this into an npm package? Do we require updating the current package.json on `gatsby-theme-iterative-docs` with all the packages used ?

Yes to both!

plugins/gatsby-theme-iterative-docs/gatsby-config.js
```diff
@@ -0,0 +1,57 @@
...
+          },
+          'gatsby-remark-responsive-iframe'
+        ]
+      }
+    })
+  }
+  return {
+    plugins
+  }
+}
```
rogermparent, line 57: Reimplemented #3143

plugins/gatsby-theme-iterative-docs/gatsby-node.js
```diff
@@ -33,6 +34,25 @@ exports.createSchemaCustomization = async api => {
...
+        ),
+        [path.resolve(__dirname, 'redirects')]: path.resolve(
+          'src',
+          'gatsby-theme-iterative-docs',
+          'redirects'
+        )
+      }
+    }
+  })
+}
```
rogermparent, line 54: The meat and potatoes to getting sidebar resolution working: this is similar to theme shadowing within Gatsby and uses the same directory structure on the consumer end, but the actual node files in the same location allow `gatsby-node` and other pure node modules to work with the same imports, and so allowing the modules that import these to be used in both environments, enabling the `utils/shared` modules to work even in an external module.

plugins/gatsby-theme-iterative-docs/src/components/SEO/index.tsx
```diff
@@ -3,9 +3,14 @@ import Helmet from 'react-helmet'
...
 import getSiteMeta from '../../queries/siteMeta'
-import { IPaginatorPageInfo } from '../Paginator'
 import { buildMetadata, MetaProps } from './helper'
 
+export interface IPaginatorPageInfo {
+  currentPage: number
+  nextPage?: string
+  previousPage?: string
+}
+
```
rogermparent, line 13: We need this interface but not the rest of `Paginator`

src/components/Page/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 import MainLayout from '../MainLayout'
 import DefaultSEO from './DefaultSEO'
-import DocumentationLayout from '../Documentation/Layout'
+import DocumentationLayout from '../DocumentationLayout'
```
rogermparent, line 5: We took the rest of the `Documentation` component but left `Documentation/Layout` behind and promoted it to its own standalone component.
Taking this into the theme would mean a lot more components the layout depends on would have to be pulled in, but this layout actually isn't very closely coupled with the rest of the `Documentation` component so breaking it apart works nicely.

tsconfig.json
```diff
@@ -18,7 +18,12 @@
     "noFallthroughCasesInSwitch": true,
     "allowSyntheticDefaultImports": true,
     "allowJs": true,
-    "alwaysStrict": true
+    "alwaysStrict": true,
+    "paths": {
+      "gatsby-theme-iterative-docs/*": [
+        "./plugins/gatsby-theme-iterative-docs/*"
+      ]
+    }
```
rogermparent, line 26: This enables `tsc` linting to work with the faked package imports, and will have to be removed once the theme becomes a real package.

plugins/gatsby-theme-iterative-docs/gatsby-config.js
yathomasi: Is there any website where we don't use remark plugins?

rogermparent: iterative.ai doesn't, and likely won't, but it also wouldn't be using this theme. Also, cml.dev doesn't use the custom local remark plugins that dvc.org does, though it'll probably want to in the future.

src/components/DocumentationLayout/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+      modifiers={[LayoutModifiers.Wide, LayoutModifiers.Collapsed]}
+    >
+      <ThemeDocumentationLayout currentPath={pathname}>
+        {children}
+      </ThemeDocumentationLayout>
+    </MainLayout>
+  )
+}
+
+export default Layout
```
rogermparent, line 23: In the latest iteration, I noticed the inner part of `DocumentationLayout` is pretty necessary to make the whole thing work, but thankfully only `MainLayout` pulls in a bunch of website dependencies so it can be easily broken off here.

julieg18 (approved):
Looks good and everything seems to be working correctly!

rogermparent:
> Looks good and everything seems to be working correctly!

Here's hoping it stays that way on merge!

yathomasi (approved):
Looks good on my end too.

rogermparent:
There was one build where images were broken in docs, but it seemed to clear up after a cache clear. I'll try another cached build and see if it happens again, otherwise I'll merge.

PR #3175, opened 2022-01-12
ref: clarify mechanics around `-d/R` options

jorgeorpinel:
> Context: https://discord.com/channels/485586884165107732/563406153334128681/930551738711310426

Very similar change in most files other than a few phrases updated in a couple Description sections.


content/docs/command-reference/add.md
```diff
@@ -128,8 +128,9 @@ not.
 
 - `-R`, `--recursive` - determines the files to add by searching each target
   directory and its subdirectories for data files. If there are no directories
-  among the `targets`, this option is ignored. For each file found, a new `.dvc`
-  file is created using the process described in this command's description.
+  among the `targets`, this option has no effect. For each file found, a new
```
jorgeorpinel, line 131: is ignored -> has no effect

jorgeorpinel, line 131: is ignored -> has no effect

content/docs/command-reference/checkout.md
```diff
@@ -18,9 +18,9 @@ positional arguments:
 ## Description
 
 This command is usually needed after `git checkout`, `git clone`, or any other
-operation that changes the current <abbr>DVC files</abbr>. It restores the
-corresponding versions of the DVC-tracked data files and directories from the
-<abbr>cache</abbr> to the workspace.
+operation that changes the current `dvc.lock` or `.dvc` files in the
+<abbr>project</abbr>. It restores the corresponding versions of all DVC-tracked
```
jorgeorpinel, line 22: More specific

content/docs/command-reference/checkout.md
```diff
@@ -75,15 +75,16 @@ the [pipeline](/doc/command-reference/dag) must be reproduced (using
 - `--summary` - display a short summary of the changes done by this command in
   the workspace, instead of a full list of changes.
 
+- `-d`, `--with-deps` - only meaningful when specifying `targets`. This
+  determines files to update by resolving all dependencies of the target stages
```
jorgeorpinel, line 79: List this option first (as in most other refs) since it's more broad: applies to any target (`-R` only applies to dir targets).

jorgeorpinel: Also, the wording is standardized across refs.

content/docs/command-reference/commit.md
```diff
@@ -19,8 +19,7 @@ positional arguments:
 Stores the current contents of files and directories tracked by DVC in the
 <abbr>cache</abbr>, and updates `dvc.lock` or `.dvc` files if/as needed. This
 forces DVC to accept any changed contents of tracked data currently in the
-<abbr>workspace</abbr>. We explore the scenarios in which this can be useful
-next.
+<abbr>workspace</abbr>.
```
jorgeorpinel, line 22: Several small copy edits to this ref Desc.

content/docs/command-reference/fetch.md
```diff
@@ -12,7 +12,7 @@ usage: dvc fetch [-h] [-q | -v] [-j <number>] [-r <name>] [-a] [-T]
 
 positional arguments:
   targets       Limit command scope to these tracked files/directories,
-                .dvc files, or stage names.
+                stages, or .dvc files.
```
jorgeorpinel, line 15: Mention stages before .dvc files (as in other definitions).

- [ ] This technically needs a core repo update...

content/docs/command-reference/fetch.md
```diff
@@ -72,19 +72,18 @@ The `dvc remote` used is determined in order, based on
...
+  searches backward from the targets in the corresponding pipelines. This will
+  not fetch files referenced in later stages than the `targets`.
 
 - `-R`, `--recursive` - determines the files to fetch by searching each target
   directory and its subdirectories for `dvc.yaml` and `.dvc` files to inspect.
-  If there are no directories among the `targets`, this option is ignored.
+  If there are no directories among the `targets`, this option has no effect.
+
+- `--run-cache` - downloads all available history of
+  [stage runs](/doc/user-guide/project-structure/internal-files#run-cache) from
```
jorgeorpinel, line 85: Moved this option under -d and -R. Seems less relevant

jorgeorpinel (commented):
Summary of changes 👇

content/docs/command-reference/commit.md
```diff
@@ -67,17 +65,20 @@ Note that it's best to try avoiding these scenarios, where the
...
 
-- `-d`, `--with-deps` - determines files to commit by tracking dependencies to
-  the target stages or `.dvc` files. If no `targets` are provided, this option
-  is ignored. By traversing all stage dependencies, DVC searches backward from
-  the target stages in the corresponding pipelines. This means DVC will not
-  commit files referenced in later stages than the `targets`.
+- `-d`, `--with-deps` - only meaningful when specifying `targets`. This
+  determines files to commit by resolving all dependencies of the target stages
+  or `.dvc` files: DVC searches backward from the targets in the corresponding
+  pipelines. This will not commit files referenced in later stages than the
```
jorgeorpinel, line 75: Question: does this imply a kind of reclusiveness as well? In that stage dependencies are resolved recursively (not just once level before target stages). And if so should this be specified? I'd prefer to avoid using the term "recursive" here in such note though, not to create confusion with `-R`.

iesahin:
Looks good. Is this done?

jorgeorpinel:
Yes, it's ready for review @iesahin .

content/docs/command-reference/commit.md
iesahin: What reclusiveness? :)

jorgeorpinel: Oops meant recursiveness. Stage dependencies are resolved recursively.

content/docs/command-reference/metrics/diff.md
```diff
@@ -61,8 +61,8 @@ all the current metrics (without comparisons).
   ```
 
 - `-R`, `--recursive` - determines the metrics files to use by searching each
-  target directory and its subdirectories for DVC-tracked files to inspect. If
-  there are no directories among the `targets`, this option is ignored.
+  target directory and its subdirectories for `dvc.yaml` files to inspect. If
```
dberenbaum, line 64:
```suggestion
  target directory and its subdirectories for metrics files to inspect. If
```

`metrics` has different behavior because targets can be anything that can be parsed as metrics, even if not specified in `dvc.yaml`. For example, a JSON file with `{"acc": 0.9}` will be included as long as its somewhere in the target directory, even if no `dvc.yaml` exists.

content/docs/command-reference/fetch.md
dberenbaum: What other definitions? Looking at the core help text, I just checked `push`, `pull`, and `checkout`, and all had `files/directories, .dvc files, or stage names`.

content/docs/command-reference/metrics/diff.md
jorgeorpinel: Oh true, good catch, updated! Is that also true for `metrics show`?

jorgeorpinel: >  Is that also true for metrics show -R?

I tested it on 2.9.3 and it seems so, so updated as well.

`metrics diff -R` isn't working as expected for me though... I'll open an issue ⌛

jorgeorpinel: > `metrics diff -R` isn't working as expected for me though...

Never mind, I forgot to provide an `a_rev` 😬

content/docs/command-reference/fetch.md
dberenbaum: @jorgeorpinel This is the only question I have left. Is this change actually needed? Where did it come from?

jorgeorpinel: This particular change in the Synopsis is not needed @dberenbaum . It's a marginal improvement IMO (stage names are more important than .dvc files).

dberenbaum: Would you mind if we take this change out? I don't think it's worth it because:
1. It requires a core change
2. All the similar commands seem to use the current order
3. It makes more sense to me in the current order because the first two are files.

iesahin (approved):
LGTM. Minor point and I think @dberenbaum 's [comments about the order](https://github.com/iterative/dvc.org/pull/3175/files#r787022744) seems correct.

content/docs/command-reference/fetch.md
jorgeorpinel: Taken out for now!

content/docs/command-reference/commit.md
```diff
@@ -34,13 +33,13 @@ these step(s) during the process of tracking each file or directory:
 - Store the file contents in the cache.
 
 > Skipping these steps is typically done to avoid caching unfinished data, for
-> example when exploring different data or <abbr>stages</abbr>.
+> example when exploring different data or code.
```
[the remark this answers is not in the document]
jorgeorpinel: Sure, going with "datasets" then. I mainly wanted to remove the stage concept from this note.

content/docs/command-reference/fetch.md
jorgeorpinel: > What other definitions? ... `push`, `pull`, and `checkout` all have "tracked files/directories, .dvc files, or stage names".

Yeah you're right, not sure why I thought we'd changed those before. Also `status`

> makes more sense to me in the current order because the first two are files

The first one is "tracked files/dirs" so not exactly files and ".dvc files" in that sense is more of a configuration item. That is to say that to my mind they're in different buckets anyway. So stage names still seems more important to me (marginally), but I'll leave it up to you to consider whether to make an issue about it in the core repo.

Thanks again @dberenbaum

Issue #2939, opened 2021-10-18
docs: apply more language tabs


jorgeorpinel:
> Follow up to https://github.com/iterative/dvc.org/issues/759

Per tabs system introduced in https://github.com/iterative/dvc.org/pull/2914

### Usage

```md
<!-- /some/page.md -->
<toggle>
<tab title="Foo">

some text about Foo

</tab>
<tab title="Bar">

some text about Bar

</tab>
</toggle>
```

- ~~Put this usage (and other special features) in https://dvc.org/doc/user-guide/contributing/docs ? Coincidentally, we could use tabs to list the special feats~~ UPDATE: https://github.com/iterative/dvc.org/issues/3506

---

<a name="ideas-places"></a>
#### Main use

- [ ] Upgrade Linux-specific terminal/shell or code samples to cover Windows too.
- [ ] Installation page sections https://dvc.org/doc/install
- [ ] #3311 
- [ ] Throughout for params in different formats (guides, refs)
- [ ] Throughout for metrics/plots in different formats (guides, refs)
- [ ] Throughout for some `--json/csv/md` options (guides, refs.)

#### Other ideas (exploratory)

- [ ] Throughout (some) examples of https://dvc.org/doc/user-guide/project-structure/pipelines-files to show writing dvc.yaml vs. using `stage add` -- +& related guides/refs. ([ex](https://dvc.org/doc/user-guide/how-to/add-deps-or-outs-to-a-stage#example))
- [ ] To show different ways for [Registering checkpoints](https://dvc.org/doc/user-guide/experiment-management/checkpoints#registering-checkpoints-in-your-code) (currently only DVCLive is shown, not signal files/`make_checkpoint()`).
- [ ] Try in Examples of https://dvc.org/doc/user-guide/external-dependencies and https://dvc.org/doc/user-guide/managing-external-data (and https://dvc.org/doc/user-guide/contributing/core#testing-remotes)
- [ ] Try in https://dvc.org/doc/user-guide/what-is-dvc#core-features and/or https://dvc.org/doc/user-guide/related-technologies (pages could  be merged)
- [x] ~Try for the feature highlights in https://dvc.org/doc/use-cases/ci-cd-for-machine-learning -- make the page shorter/cleaner~
- [ ] Try in subsections of https://dvc.org/doc/use-cases/data-registries#using-registries
- Maybe for Experiment Tracking tool comparison (instead of table)
- Note: may be useful for #2866
[...]

PR #3178, opened 2022-01-12
guide: move Exp org patterns to Persisting Exps, improve

jorgeorpinel:
~~rel. https://github.com/iterative/dvc.org/issues/2939~~

- [x] Merge paragraphs above and below.
- [x] Add examples? (e.g. `tree`-like folder views, Git tree pic, etc.)


jorgeorpinel:
Got a styling issue @julieg18 . Maybe someone in the web team can update the margin of the tabs section so there is some white space before the next paragraph? Currently it's gray -- looks odd:

image
> See https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management#organization-patterns

Thanks

julieg18:
> Got a styling issue @julieg18 . Maybe someone in the web team can update the margin of the tabs section so there is some white space before the next paragraph? Currently it's gray -- looks odd:

Thanks for spotting that! I'll open a pr for that adds some bottom margin to the tabs!

jorgeorpinel:
Thanks @julieg18 !

julieg18:
Pull request for adding a margin bottom is merged!

jorgeorpinel:
>  Add examples? (e.g. tree-like folder views, etc.)

WDYT @shcheklein ? Should we add some visual examples to [each tab](https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management#organization-patterns)?

iesahin:
It's visually nicer but I'm not sure this is a good way to use tabs. Tabs are used, in general, not for overlapping content but for clearly distinct alternatives: Programming languages, operating systems, etc. The reader sees the tabs more than the content inside them, so I'd use the tabs for content rather conservatively.

iesahin:
Also, it becomes _too much work_ to read all. We expect the reader to read and understand all these options, not select and apply one them.

jorgeorpinel:
> Tabs are used, in general, not for overlapping content but for clearly distinct alternatives

Not sure what overlapping means here. I think the org patterns are clear alternatives. I don't think we have to limit tabs use to languages or platforms although that's their original intention indeed.

shcheklein:
I don't have a strong opinion on this. I still think that both pages - index and overview should be significantly reconsidered (e.g. run cache on the index page is a way bigger problem), even organization patterns looks like out of the place for the index page - it's too specific information, or at least it's written this way. Point here is that most likely users won't understand this section anyway here, so no strong opinion on this :)

I would move this into the Persisting experiments. Index and Overview - should be about bigger picture, general workflow diagrams, benefits, help users better navigate in the section stucture (map subsections to the workflow), etc.

jorgeorpinel:
> Index and Overview - should be about bigger picture, general workflow diagrams, benefits, help users better navigate in the section stucture (map subsections to the workflow), etc.

Agree. There's tasks for this in #2911 (updated with your notes).

> I would move this into the Persisting experiments.

Good idea ⌛ 
So assuming we put it there, what do we think about using tabs? And having "freed up some screen space" that way, about adding examples for each pattern?

jorgeorpinel:
> Usually tabs serve as a way to customize "same" content

@shcheklein I think we could use them some times to present alternatives but I'm not sure when it's appropriate (vs. different headings or just a list for example).

I'm experimenting a little here, so PTAL at https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management/persisting-experiments#organization-patterns to see a concept of what I mean. 

> Unfortunately the page jumps a lot since the tabs are at the end but there's an issue about this: #3179

content/docs/user-guide/experiment-management/persisting-experiments.md
```diff
@@ -94,3 +94,91 @@ files, etc.) can be stored in Git.
...
+  </card>
+
+</cards>
+
+</tab>
+<tab title="Labels (ad hoc)">
+
+In general, you can record experiments in a separate system and structure them
+using custom labeling. This is typical in dedicated experiment tracking tools. A
+possible problem with this approach is that it's easy to lose the connection
```
jorgeorpinel, line 175: p.s. We could plug in LDB here at some point.

jendefig:
> I'm experimenting a little here, so PTAL at https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management/persisting-experiments#organization-patterns to see a concept of what I mean.

I do like this.  In this case, it's alternatives on doing the same thing "setting up experiment structure" and I think having the succinct titles of the tabs, frames the "I have options" experience for the user better than rolling through all the options down the page, which can start to feel overwhelming.  

But yes, the jumpy thing is a bit uncomfortable.  Need to figure that one out.

jorgeorpinel:
To recap guys. Sounds like we have 2 votes against (@iesahin, @shcheklein) and 2 for (@jendefig, myself)? 

> the jumpy thing is a bit uncomfortable. Need to figure that one out.

Being discussed in #3179 but not seen as high priority as of now.

iesahin:
What will we do to resolve the tie? :) Duel?

shcheklein:
I still think this is not the right thing to do. If we try to go to the bottom of why usually people don't use tabs for content (vs tabs for customizing the same thing) it can lead among other things stuff like broken search (browser and the site one), if we decide to export docs into PDF / book - there will be a problem also - we would need to flatten it anyway, right side bar navigation becomes a bit broken (it doesn't reflect the structure), we need to double check that it's easy to share the link to a section, etc, etc.

jorgeorpinel:
> the jumpy thing is a bit uncomfortable. Need to figure that one out.

That's been fixed. See https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management/persisting-experiments#organization-patterns

Also, here's some validation to have info about alternative experiment organization patterns (via @dberenbaum): https://www.reddit.com/r/MachineLearning/comments/smnw3o/comment/hw05eb2/?utm_source=share&utm_medium=web2x&context=3

jorgeorpinel:
> it can lead among other things stuff like broken search

@shcheklein good question. I can't test that on the review app

> if we decide to export docs into PDF / book

We already have that problem with hidden sections anyway but yes, also true.

> right side bar navigation becomes a bit broken

I don't see that though, the internal story or structure of each section doesn't need to be reflected in the ToC IMO.

> we need to double check that it's easy to share the link

Yep, we can do this. Example: https://dvc-org-guide-exp-tabs-sz1jswq.herokuapp.com/doc/user-guide/experiment-management/persisting-experiments?tab=Directories#organization-patterns (notice `?tab=Directories` URL query)

shcheklein:
> We already have that problem with hidden sections anyway but yes, also true.

hidden sections are easier to my mind (and to some extent they are hidden since they non-essential, so less of a problem by default) - it can be flattened as-is - just render it expanded

> search broken 

it means that you could not Ctrl-F and search on the page I think

jorgeorpinel:
> you could not Ctrl-F and search

Ah true. Also a problem in hidden sections.

So yeah unless we are OK with hiding parts of the content then tab toggles wouldn't be appropriate. And if we do want to hide it then it's unclear why pick tabs vs regular hidden sections 🤔 

Anyway, I'll flatten this one for now... ⌛

jorgeorpinel:
> unless we are OK with hiding parts of the content then tab toggles wouldn't be appropriate

p.s. This implies we are OK with hiding secondary systems or languages such as Gitlab and Bitbucket in https://cml.dev/doc/usage but could be Windows, CSV, etc. (refer to list in https://github.com/iterative/dvc.org/issues/2939)

shcheklein:
Good stuff, Jorge, but before we review this, could you please help me with the context :):

What was the intention behind this PR? (I see the link to making tabs, but title says different). I'm curios was the initial problem discussed somewhere, any specific motivation? 

PR itself:

- diagram is blurry and doesn't look "native" to docs
- cards are clickable and lead to Get Started, seem a bit too much to use them (use table instead?)
- arguably organizing experiments deserves it's own page or be part of the overview / intro (in some simple form)
- can we reuse some lang from the reddit post (probably even in intro) to start answering their concerns from the very beginning ...

jorgeorpinel:
> diagram is blurry and doesn't look "native"

Fixed blur I think (still adapting to Macbook). The graph is from https://github.com/iterative/example-dvc-checkpoints/network ... do you mean that the labels don't seem to connect or that the style seems off? Not sure how to address the latter

> cards are clickable and lead to Get Started, seem a bit too much to use them (use table instead?)

Oops, removed links.

Actually something in the site CSS won't let me make a full-width table. I tried [this code](https://jsfiddle.net/1wgmv63d/) and it shows it like this:

image

And it has borders like the cards do so either way we need custom styles to remove those if that's what you meant. Lmk

⌛

jorgeorpinel:
> arguably organizing experiments deserves it's own page or be part of the overview / intro 

Idk. We've discussed this a few times and seem to go in circles. The info. seems relevant but maybe `exp` features don't make it easy to get to alternative organizations yet, so it could be a product issue. I think that keeping it here for now is appropriate.

> can we reuse some lang from the reddit post

OK I tried incorporating some of the concepts in a304823b, basically adding some more motivation to the section intro.

jorgeorpinel:
> This image needs to be branded. Shall I make a thread for Roman

Is it still an issue? I think it looks less odd now... Maybe I should just remove the **Sep** header and maybe add a caption linking to the source?

image

> having a dedicated page for this is a good idea

Yeah but we already have 7 pages in the Experiment Management guide so unless it's absolutely clear that this is a major exp-related topic I'd try to avoid that.

shcheklein:
> Is it still an issue?

it's still blurry ... also if we have somewhere commits tree image I would try to use the same style


On separate page or making it part of the Overview/Index.

I guess for me it's hard to connect intro like `decide how to organize them once completed. ` with directories. May be we need to add some examples? How can we bring a completed experiment into a subdirectory?

To me all these patterns are more general than persisting experiments after they done. It's more about how do you structure your project to accommodate experimentation in the first place. wdyt?

jorgeorpinel:
> it's still blurry ... also if we have somewhere commits tree image

I zoomed in as much as possible. I don't think we have any Git tree image other than custom made ([2nd fig. here](https://dvc.org/doc/use-cases/versioning-data-and-model-files)).

> To me all these patterns are more general than persisting experiments

I agree. This is meant as an incremental improvement but the deeper iteration should be on the core product itself I believe, and probably the core team is thinking about it cc @dberenbaum ? But for now this is an improvement IMO.

shcheklein:
> This is meant as an incremental improvement

I guess that's what I struggle a bit to see we had a section that was a bit more compact, arguably in a better place and now we have the same + some images (that ideally we would want to unify somehow). I'm not sure about this being an improvement to be honest. And text is confusing now since a bunch of those organizational patterns are not about `dvc exp`.

content/docs/user-guide/experiment-management/persisting-experiments.md
```diff
@@ -94,3 +94,86 @@ files, etc.) can be stored in Git.
...
+
+  </card>
+  <card>
+    v0.2.0
+
+    ```
+    └── experiments
+        ├── cnn_128
+        └── cnn_512
+    ```
```
iesahin, line 164: I think instead of naming these `cnn_128`, `cnn_512`, etc. it may be more understandable to use the same words and common steps in an ML pipeline, like `data augmentation`, `dataset1`, `dataset2` etc. `cnn` is always in the training phase, and may be different versions of `model` directory.

PR #3182, opened 2022-01-12
ref: extract technical details from `exp run` to guides

jorgeorpinel:
For tasks in #2548

> - extract technical details from ...reference/exp (esp. run) to guides

_Most of the material *should* already be in guides._

and per #2768

> - The pipeline section seems too deep...
> - The text under (Hyper)parameters has nothing to do with running experiments

---

- See [summary](https://github.com/iterative/dvc.org/pull/3182#pullrequestreview-851239689) below. UPDATE: Jump to https://github.com/iterative/dvc.org/pull/3182#issuecomment-1012558741 for latest logic behind these changes.
- Plus some misc. copy edits in related/near texts as needed
- See also ref. vs guide discussion points in https://github.com/iterative/dvc.org/pull/3182#issuecomment-1012558741.
- Also happens to close #2745

PR #3189, opened 2022-01-18
guide: simplify Checkpoints guide

jorgeorpinel:
And misc. related changes

Extracted from https://github.com/iterative/dvc.org/pull/3182

Begins addressing #3189


content/docs/user-guide/experiment-management/checkpoints.md
```diff
@@ -2,24 +2,35 @@
...
-example to track the progress in deep learning techniques such as evolving
-neural networks.
-
-_Checkpoint experiments_ track a series of variations (the checkpoints) and
-their execution can be stopped and resumed as needed. You interact with them
-using the `--rev` and `--reset` options of `dvc exp run` (see also the
-`checkpoint` field in `dvc.yaml` `outs`). They can help you
+To track successive steps in a longer machine learning experiment, you can
+register checkpoints from your code at runtime, for example to track the
+progress with deep learning techniques. They can help you
```
iesahin, line 7: I think a : after "you" looks better. Or something like

They can help you to: 

- implement
- track
- see
- automate

content/docs/user-guide/experiment-management/checkpoints.md
```diff
@@ -2,24 +2,35 @@
...
   checkpoints.
 - track all code and data changes corresponding to the checkpoints.
 - see when metrics start diverging and revert to the optimal checkpoint.
 - automate the process of tracking every training epoch.
 
-> Experiments and checkpoints are [implemented](/blog/experiment-refs) with
-> hidden Git experiment commits branches.
+Checkpoint [execution] can be stopped and resumed as needed. You interact with
+them using the `--rev` and `--reset` options of `dvc exp run` (see also the
+`checkpoint` field in `dvc.yaml` `outs`).
```
iesahin, line 17: It might be better to use `dvc exp run --rev` to link directly to the option. Might need rephrasing though.

iesahin (commented):
Minor points. Thank you Jorge.

content/docs/user-guide/experiment-management/checkpoints.md
jorgeorpinel: Thanks for the suggestion. No `:` is needed though. It's a stylistic choice.

content/docs/user-guide/experiment-management/checkpoints.md
jorgeorpinel: I tend to agree but since there are 2 options here it would be too much to link both that way or weird to only link one. Could link them manually I guess... I'm not sure it's needed.

Please approve if there's no major feedback 🙂 Thanks @iesahin

iesahin (approved):
Thanks Jorge, good to go :)

jorgeorpinel:
Why thank you!

PR #3187, opened 2022-01-17
`exp show`: add `--rev` option

karajan1001:
Related to https://github.com/iterative/dvc/pull/7204

> & https://github.com/iterative/dvc/pull/7245

PR #3191, opened 2022-01-18
`exp list`: add `--num` option, rename `--all` to `--all-commits`

karajan1001:
Related to https://github.com/iterative/dvc/pull/7245

### DIFF
--- a/content/docs/command-reference/exp/list.md
+++ b/content/docs/command-reference/exp/list.md
@@ -5,8 +5,8 @@ List experiments in a <abbr>DVC repository</abbr> (remote or local).
 ## Synopsis
 
 ```usage
-usage: dvc exp list [-h] [-q | -v] [--rev <rev>]
-                    [--all] [--names-only]
+usage: dvc exp list [-h] [-q | -v] [-A] [--rev <commit>]
+                    [-n <num>] [--names-only]
                     [git_remote]
 
 positional arguments:
@@ -34,10 +34,15 @@ options below).
 
 ## Options
 
-- `--rev <commit>` - list experiments derived from the specified Git commit
-  (instead of `HEAD`).
+- `--rev <commit>` - list experiments derived from the specified `<commit>` as
+  baseline (HEAD by default).
 
-- `--all` - list all experiments in the repository (overrides `--rev`).
+- `-n <num>`, `--num <num>` - show experiments from the previous `num` commits
+  (before the `--rev` baseline). For example `-n -2` includes experiments from
+  `HEAD` and its 2 first ancestors.
+
+- `-A, --all-commits` - list all experiments in the repository (overrides
+  `--rev` and `--num`).
 
 - `--names-only` - print only the names of the experiments without their parent
   Git commit.
--- a/content/docs/command-reference/exp/list.md
+++ b/content/docs/command-reference/exp/list.md
@@ -38,8 +38,8 @@ options below).
   baseline (HEAD by default).
 
 - `-n <num>`, `--num <num>` - show experiments from the last `num` commits
-  (first parents) starting from the `--rev` baseline. Give a negative value to include
-  all first-parent commits (similar to `git log -n`).
+  (first parents) starting from the `--rev` baseline. Give a negative value to
+  include all first-parent commits (similar to `git log -n`).
 
 - `-A, --all-commits` - list all experiments in the repository (overrides
   `--rev` and `--num`).
--- a/content/docs/command-reference/exp/list.md
+++ b/content/docs/command-reference/exp/list.md
@@ -37,9 +37,9 @@ options below).
 - `--rev <commit>` - list experiments derived from the specified `<commit>` as
   baseline (HEAD by default).
 
-- `-n <num>`, `--num <num>` - show experiments from the previous `num` commits
-  (before the `--rev` baseline). For example `-n -2` includes experiments from
-  `HEAD` and its 2 first ancestors.
+- `-n <num>`, `--num <num>` - show experiments from the `--rev` baseline and
+  from `num` commits before it (first parents). Give a negative value to include
+  all first-parent commits (similar to `git log -n`).
 
 - `-A, --all-commits` - list all experiments in the repository (overrides
   `--rev` and `--num`).
--- a/content/docs/command-reference/exp/list.md
+++ b/content/docs/command-reference/exp/list.md
@@ -37,8 +37,8 @@ options below).
 - `--rev <commit>` - list experiments derived from the specified `<commit>` as
   baseline (HEAD by default).
 
-- `-n <num>`, `--num <num>` - show experiments from the `--rev` baseline and
-  from `num` commits before it (first parents). Give a negative value to include
+- `-n <num>`, `--num <num>` - show experiments from the last `num` commits
+  (first parents) starting from the `--rev` baseline. Give a negative value to include
   all first-parent commits (similar to `git log -n`).
 
 - `-A, --all-commits` - list all experiments in the repository (overrides


content/docs/command-reference/exp/list.md
```diff
@@ -5,8 +5,8 @@ List experiments in a <abbr>DVC repository</abbr> (remote or local).
 ## Synopsis
 
 ```usage
-usage: dvc exp list [-h] [-q | -v] [--rev <rev>]
-                    [--all] [--names-only]
+usage: dvc exp list [-h] [-q | -v] [-A] [--rev <commit>]
+                    [-n <num>] [--names-only]
```
shcheklein, line 9: not related: `--names-only` description is missing it seems like

iesahin: Added #3195 for this. I can merge one of these and rebase the other.

iesahin (approved):
LGTM Gao, thank you.

content/docs/command-reference/exp/list.md
```diff
@@ -37,7 +37,10 @@ options below).
 - `--rev <commit>` - list experiments derived from the specified Git commit
   (instead of `HEAD`).
 
-- `--all` - list all experiments in the repository (overrides `--rev`).
+- `-n <num>, --num <num>` - list the last `num` commits from <rev>.
```
shcheklein, line 40: <rev> -> `<rev>`

commits -> experiments? 

Something like this probably (otherwise, it's not clear what are we listing after all):

list experiments from the `<rev>` commit and `num` commits before it

do we include <rev> itself btw?

jorgeorpinel:
> Added #3195 for this. I can merge one of these and rebase the other.

The other PR is merged. Please merge/rebase `master` here.

content/docs/command-reference/exp/list.md
karajan1001: Something like this probably (otherwise, it's not clear what are we listing after all):
> list experiments from the <rev> commit and num commits before it

I think so.

> do we include itself btw?

Yes, we do, it can only be an integer above 1 or equals -1.  

- 1: no history includes.
- n: last n histories.
- -1: all of the histories.

shcheklein: okay, so, that's the best place to expand and clarify this I think. thanks, @karajan1001.

iesahin: IMHO having numeric parameters without clear meaning is not a good UI. I think the option could be split into 3, like

```
--last-n-commits <n>
--no-history
--all-history
```

shcheklein: The most important case here is to provide `n > 0`, right? so, in this case my 2cs - it's better to keep simple `-n`, and keep `-1`, `0` and other things as an advanced stuff (that can be done with a normal set of commands)

content/docs/command-reference/exp/list.md
```diff
@@ -37,7 +37,11 @@ options below).
 - `--rev <commit>` - list experiments derived from the specified Git commit
   (instead of `HEAD`).
 
-- `--all` - list all experiments in the repository (overrides `--rev`).
+- `-n <num>, --num <num>` - list experiments from the `<rev>` commit and `num`
+  commits before it. (`-1` for the whole history of the `<rev>`)
```
jorgeorpinel, line 41: Same as https://github.com/iterative/dvc.org/pull/3187#discussion_r793240744

content/docs/command-reference/exp/list.md
```diff
@@ -37,7 +37,11 @@ options below).
...
+
+- `-A, --all-commits` - list all experiments in the repository (overrides
+  `--rev`).
```
jorgeorpinel, line 44: Same as https://github.com/iterative/dvc.org/pull/3192#pullrequestreview-864364698

jorgeorpinel (changes requested):
Let's be consistent with the other similar PRs please 🙂

content/docs/command-reference/exp/list.md
jorgeorpinel: Agree accepting only positive integers make sense. `-1` is cryptic -- should be another flag instead?

> Lets maybe discuss on the product side in https://github.com/iterative/dvc/pull/7245#discussion_r788133855 though?

iesahin: Yep, if the other 2 seems advanced enough, `-n` should only accept a `> 0` integers and have a definite meaning. It's like `git log -n`.

dberenbaum: No history is the default, so a special option is needed only to represent all history.

>  list experiments from the <rev> commit and num commits before it

Given this language, I would expect 0 to be the default and showing no history, while 1 would show `rev` plus the previous commit. This makes more sense to me, but it's a breaking change. Otherwise, we can tweak this language.

---

I think it's somewhat common for commands/functions to use `-1` or similar to mean "all" for numeric arguments rather than make them separate options (maybe this is specific to certain domains?), and IMO it would be cumbersome to have a separate flag for all history (that would be 2 separate flags to specifically interact with `--rev`). I was only arguing to make it a bit more accepting of unexpected values, like saying "The default is 1, and any number <1 will show all history" (replace 1 with 0 if that becomes the default).

https://clig.dev/#arguments-and-flags seems to accept that "special" values are acceptable:

> If a flag can accept an optional value, allow a special word like “none.” For example, ssh -F takes an optional filename of an alternative ssh_config file, and ssh -F none runs SSH with no config file.

jorgeorpinel: Good points. Since --num already exists in exp show, I'd extract this discussion to a DVC issue (or is the -1 part new?). Related: https://github.com/iterative/dvc/issues/7314

> Otherwise, we can tweak this language

Yeah for now the language here and related PRs should reflect the actual behavior but I'm still confused actually: These 2 rules seem contradictory:

> - 1: no history includes.
> - n: last n histories.

What if n=1, is it "no history" or "last 1 history"?

dberenbaum: `n=1` shows experiments from `<commit>` only, so `<commit>` is currently included in the count. That's why the suggested wording of "list experiments from the commit and num commits before it" is good but inaccurate currently.

dberenbaum: > Since --num already exists in exp show, I'd extract this discussion to a DVC issue (or is the -1 part new?). Related: [iterative/dvc#7314](https://github.com/iterative/dvc/issues/7314)

-1 is new I think, and we are expanding `--num` to apply to many other `exp` commands besides `show`. We can move to a new issue if you prefer, although I don't think we can merge this until we decide.

iesahin: From the [discussion](https://github.com/iterative/dvc/pull/7245#discussion_r793538068) and demo in iterative/dvc#7245, the word is _commits_, not _experiments._

shcheklein: @iesahin I think in this case there are commits _and_ experiments. Just by the nature of this options. Or we can use `experiments` and `revisions` (e.g. list all experiments from N past revisions), but I don't see how can we make it (list commits from the last N commits). Or may be I'm missing your point?

iesahin: The number specified by `-n` is the same number given to `git log -n`. Each commit may have more than one associated experiment and `-n` does not specify the number of experiments. 

```
commit-ABCD
    - exp 1
    - exp 2
    - exp 3
commit-DEF1
    - exp 4
    - exp 5
    - exp 6
    - exp 7
commit-6375
   - exp 8
   - exp 9
```

FWIU, when you specify `-n 2`, all experiments from the last two commits (`exp 1` to `exp 7` here) are listed.

jorgeorpinel: > n=1 shows experiments from <commit> only, so <commit> is currently included in the count...
We can move to a new issue if you prefer

@dberenbaum Then there's no way to list/show experiments from the given rev `<commit>` and from it's parent only (`n=2` would give more than that). Seems like a core issue indeed, so I commented in https://github.com/iterative/dvc/pull/7245#pullrequestreview-870006159 for now

> The number specified by -n is the same number given to git log -n

@iesahin except in the case of n=1. See link above 🙂

dberenbaum: Following up here because the example above from @iesahin is helpful. Here is what will show with various `-n` values:

```
$ dvc exp list
commit-ABCD
    - exp 1
    - exp 2
    - exp 3

$ dvc exp list -n 1
commit-ABCD
    - exp 1
    - exp 2
    - exp 3

$ dvc exp list -n 2
commit-ABCD
    - exp 1
    - exp 2
    - exp 3
commit-DEF1
    - exp 4
    - exp 5
    - exp 6
    - exp 7

$ dvc exp list --rev commit-DEF1
commit-DEF1
    - exp 4
    - exp 5
    - exp 6
    - exp 7

$ dvc exp list --rev commit-DEF1 -n 1
commit-DEF1
    - exp 4
    - exp 5
    - exp 6
    - exp 7

$ dvc exp list --rev commit-DEF1 -n 2
commit-DEF1
    - exp 4
    - exp 5
    - exp 6
    - exp 7
commit-6375
   - exp 8
   - exp 9
```

content/docs/command-reference/exp/list.md
jorgeorpinel: Looks like this text will have to change again... See https://github.com/iterative/dvc.org/pull/3187#pullrequestreview-883803230.

content/docs/command-reference/exp/list.md
```diff
@@ -34,10 +34,15 @@ options below).
...
 
-- `--rev <commit>` - list experiments derived from the specified Git commit
-  (instead of `HEAD`).
+- `--rev <commit>` - list experiments derived from the specified `<commit>` as
+  baseline (HEAD by default).
 
-- `--all` - list all experiments in the repository (overrides `--rev`).
+- `-n <num>`, `--num <num>` - show experiments from the previous `num` commits
+  (before the `--rev` baseline). For example `-n -2` includes experiments from
+  `HEAD` and its 2 first ancestors.
```
jorgeorpinel, line 42: See https://github.com/iterative/dvc.org/pull/3187#pullrequestreview-902553693

content/docs/command-reference/exp/list.md
```diff
@@ -34,10 +34,15 @@ options below).
...
 
-- `--rev <commit>` - list experiments derived from the specified Git commit
-  (instead of `HEAD`).
+- `--rev <commit>` - list experiments derived from the specified `<commit>` as
+  baseline (HEAD by default).
 
-- `--all` - list all experiments in the repository (overrides `--rev`).
+- `-n <num>`, `--num <num>` - show experiments from the `--rev` baseline and
+  from `num` commits before it (first parents). Give a negative value to include
+  all first-parent commits (similar to `git log -n`).
```
jorgeorpinel, line 42: Same as https://github.com/iterative/dvc.org/pull/3187#pullrequestreview-903319359
```suggestion
- `-n <num>`, `--num <num>` - show experiments from the last `num` commits
  (first parents) starting from the `--rev` baseline. Give a negative value to include
  all first-parent commits (similar to `git log -n`).
```

jorgeorpinel (approved):

dberenbaum (dismissed):

jorgeorpinel:
@iesahin no need to do cross merges. Let's just merge the Restyled PR https://github.com/iterative/dvc.org/pull/3344 when this is ready to be published.

iesahin (changes requested):
Instead of this, #3344 will be merged as @jorgeorpinel suggested.

jorgeorpinel (approved):
Oh, this is properly formatted now, actually.

Ok so are we ready to merge?

jorgeorpinel:
I take this

image

as a yes!

PR #3195, opened 2022-01-18
ref: add `--names-only` option description to `dvc exp list`

iesahin:
added `--names-only` description


jorgeorpinel (approved):
Thanks, good catch! Do we know what core PR this option was merged in? Looks like they missed that ref. update.

PR #3197, opened 2022-01-19
start: add a section about metrics to experiments

iesahin:
This one adds a brief hidden section about metrics to https://dvc.org/start/experiments/

Related #3051

### DIFF
--- a/content/docs/start/experiments.md
+++ b/content/docs/start/experiments.md
@@ -226,12 +226,23 @@ $ dvc exp show --drop 'Created|train|loss'
 
 ### ℹ️ More information about metrics
 
-Metrics are what you use to evaluate your models. DVC allows any scalar values
-to be used as metrics. It's able to track the metrics we defined in the code
-with the Keras integration introduced recently. You can also specify a custom
-metrics file using `--metrics` option of `dvc exp init`, and write the metrics
-in the code manually. Please see `dvc metrics` for this kind of explicitly
-defined metrics.
+Metrics are what you use to evaluate your models. DVC associates metrics to
+experiments to compare them. DVC allows any scalar values to be used as metrics.
+You can specify metrics JSON or YAML 1.2 formatted file using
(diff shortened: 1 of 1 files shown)


content/docs/start/experiments.md
```diff
@@ -222,6 +222,19 @@ $ dvc exp show --drop 'Created|train|loss'
...
 ```
 
+<details>
+
+### ℹ️ More information about metrics
+
+Metrics are what you use to evaluate your models. DVC allows any scalar values
+to be used as metrics. It's able to track the metrics we defined in the code
+with the Keras integration introduced recently. You can also specify a custom
+metrics file using `--metrics` option of `dvc exp init`, and write the metrics
```
shcheklein, line 232: also edit `dvc.yaml`. 

I think if we keep this section we should say something like - To capture project metrics in DVC you use JSON, TSV, whatever file or multiple files, specify them as metrics files in dvc.yaml , and your app writes with arbitrary structure into them. There are a bunch of commands (like `dvc exp show`) or services (mentioned Studio?) that interpret those files ...



Essense of the metrics, less abstract repetitive things

jorgeorpinel: Also the Keras integration is completely out of context here. Either link to the DVCLive page (I assume that's what it refers to?) or avoid! 🙂

iesahin:
This one needs a review too: @shcheklein @jorgeorpinel

shcheklein (approved):
It looks like a good improvement. I think it makes sense to move both - params and metrics details up (close to the first time they are mentioned, may be combining them into one section)

iesahin:
I created an issue, #3296 for this. Can we merge this @jorgeorpinel ?

jorgeorpinel:
Sorry for the delay. The section's content looks good, thanks. I would suggest moving it up earlier near where params are mentioned first (that would also avoid 2 details sections together), and in that case the last paragraph is probably unnecessary. I'll make a draft PR ⌛

PR #3198, opened 2022-01-19
Search blogs 

### LINKED ISSUE #2712: Add Search functionality to Blog
As the content grows, it would be good to have an internal search function for the blog like we do on the docs.

yathomasi:
This should bring the search functionality to the blogs section. We still need to work on UI/UX.

For testing this functionality locally, we need an application on Algolia for credentials keys. I have added `.env.example` for required variables. Since the Algolia index _updates only on build time_ the application needs to be built once before running the development server.

Also, the types are passed for now.( ~~WIP: fix them soon~~)

For the review app, I will be adding those environment variables, so it should be working.

Fixes #2712


.env.example
```diff
@@ -0,0 +1,4 @@
+GATSBY_ALGOLIA_APP_ID=
```
shcheklein, line 1: Q: we already have a few Algolia search boxes implemented - how did we pass these variables before?

gatsby-config.js
```diff
@@ -145,6 +145,14 @@ const plugins = [
       excludePattern: /\/doc\/cml/
     }
   },
+  {
+    resolve: `gatsby-plugin-algolia`,
```
shcheklein, line 149: Q: what does it exactly do? send data to Algolia during the build-time? or just helps to render JS?

src/gatsby/hooks/useClickOutside.tsx
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react'
```
shcheklein, line 1: Q: where do we use this hook, should it be placed locally? (vs globally into gatsby/hooks)

src/utils/algolia-queries.js
```diff
@@ -0,0 +1,51 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const escapeStringRegexp = require('escape-string-regexp')
```
shcheklein, line 2: Q: how does it/does it overlap with existing Algolia stuff?

.env.example
yathomasi: Yes, we have them implemented and the required variables are directly set to the required places. But, in this specific case, we have the ADMIN_KEY variable which should not be exposed with code. So, it's passed into this variable.

gatsby-config.js
yathomasi: Yes, exactly this helps to send data to Algolia for indexing at build-time. No, this doesn't help to render. We have another utility package called `react-instantsearch-dom` by the Algolia team for rendering.

src/gatsby/hooks/useClickOutside.tsx
yathomasi: This hook is to detect when we click outside a certain element block. Most use cases are on Modal, dropdowns, etc. where we want to close them when we click outside of that. This hook is reusable and often used on those use cases.

src/utils/algolia-queries.js
yathomasi: I don't this overlaps with the existing one. 
Because, as I mentioned the product we are using is DocSearch in which we don't do much in terms of setting index as well as rendering the results. 
For indexing, they run crawlers on their server every hour or day to extract data and index accordingly.
For rendering the results, they have provided `docsearch.js` package which is complete and we only need to specify some variables and everything will be good to go.

.env.example
shcheklein: Got it, I see that workflow is very different.

gatsby-config.js
shcheklein: thanks, Thomas, it's clear now.

Q: how much does it slow down the build? how will it work with the preview apps if they share the same index or if there is a way to do subindex? (for local testing, for preview apps, etc?)

src/gatsby/hooks/useClickOutside.tsx
shcheklein: do we have already cases like this? we have quite a lot of dropdowns

src/utils/algolia-queries.js
shcheklein: good, thanks ... I understand that those are two different approaches

gatsby-config.js
yathomasi: It's around 2-3 seconds. I just checked on preview apps build.
Screen Shot 2022-01-24 at 10 09 44
For indexing, we can set different environment variables for each environment. 
Screen Shot 2022-01-24 at 10 13 47
I have set `stag_blogs` as indexing for preview apps. 
We can only set one environment variable for preview apps, so the same index will be used for all preview apps. I don't that will be a problem.
Similarly, we can set a variable for production as well. For eg: `prod_blogs` should separate it from others.

src/gatsby/hooks/useClickOutside.tsx
yathomasi: Yeah actually. I just checked the `What's new` section on our homepage. That part is also somewhat similar to implementation but is not reusable. We can actually use this reusable hook there as well and also on upcoming similar functionality.

gatsby-config.js
yathomasi: I have now updated the keys on preview apps with the Algolia application `DVC Blog`(One that you have invited me to). So, newer builds should start using those.

src/gatsby/hooks/useClickOutside.tsx
shcheklein: how about the download button? top menu drop down?

Let's create a ticket please to unify this, so that we don't duplicate and have 3 different implementations please.

gatsby-config.js
shcheklein: sounds good, Thomas.

Still heve mixed feelings about sending every blog text to Algolia every time we rebuild 🤔 

@rogermparent @julieg18 what's your take?

rogermparent: I think running a full update every time is fine for a first iteration, particularly if it's only a few seconds with the amount we have now, but we can also consider another iteration on top of this that takes advantage of the [partial updates feature](https://www.gatsbyjs.com/plugins/gatsby-plugin-algolia/#partial-updates) of `gatsby-plugin-algolia`. We'll have to add our own `modified` field providing either a boolean or datetime, but I'm sure we can figure that out.

src/utils/algolia-queries.js
```diff
@@ -0,0 +1,51 @@
...
+          description
+        }
+        childBlogPost {
+          slug
+        }
+        excerpt(pruneLength: 1000)
+      }
+    }
+  }
+}`
```
rogermparent, line 29: I think it'd be better to query on our custom `BlogPost` nodes, as the process of making those notes had already done the work of sorting out which `MarkdownRemark` nodes are blog posts as opposed to docs instead of re-doing that work here with a regex filter as they're known to be [particularly slow and less scalable than other methods of filtering](https://www.gatsbyjs.com/docs/query-filters/#performance).

src/gatsby/hooks/useClickOutside.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    for (const event of events) {
+      document.addEventListener(event, onClick)
+    }
+
+    return () => {
+      for (const event of events) document.removeEventListener(event, onClick)
+    }
+  })
```
julieg18, line 26: Do we need this `useEffect` to run on every render? Could we achieve the same thing if we just run it on once on mount 🤔

.env.example
julieg18, line 1: If I try to run `yarn build`, I get an error. Is this because I don't have any of the env variables? If so, we probably want to make sure users are able to create a finished build even if users don't have the required api keys. 

image

gatsby-config.js
julieg18: Agreed with @rogermparent! We can use this for a first iteration and open a issue for improving this at a later date.

julieg18 (commented):
Great job so far!

gatsby-config.js
yathomasi: >We'll have to add our own modified field providing either a boolean or datetime, but I'm sure we can figure that out.

Thanks, for mentioning that out. Yeah, that would be better and more scalable. And, for `modified`, I found we can't just use the file modified datetime as that changes when the file is on different places(local, server) instead we could use the git method to get the modified datetime. [Stackoverflow](https://stackoverflow.com/a/66118168).
I would check more on that.

src/utils/algolia-queries.js
yathomasi: Yeah, we could use that approach. The issue was we couldn't find all the fields but it looks like we could use the inline query over the parent for those missing fields. So, I will consider this approach.

src/gatsby/hooks/useClickOutside.tsx
yathomasi: Great and Thanks for pointing that out. Yeah, we don't need this to run on every render. We can have `ref` and `handler` as dependencies.

.env.example
yathomasi: Sure, there is a skip indexing option as well. We can set that to true if no variables are provided.

src/components/Blog/Search/SearchResult/styles.module.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.searchResultShow {
+  display: block;
+}
+
+.searchResultContent {
+  padding-right: 4px;
+  max-height: 650px;
+  overflow: scroll;
```
rogermparent, line 24: How about `overflow-y: auto`? `-y` so we don't get a horizontal scrollbar and `auto` so it hides when there's only a few results.

rogermparent:
Seems like the Heroku deploy doesn't have the git info needed for the modification date, I ran into something similar with GH Actions and needed to change to a non-shallow checkout- maybe we just need to `git checkout` in the Heroku deploy script?

yathomasi:
@yalozhkin can you please confirm UI with the latest deployment. 
For the back button on blog pages, can you give me a little idea of the icon, text, and placement?

I have reverted the changes getting modified info using git as Heroku doesn't seem to work that way. I feel introducing git to the Heroku deployment flow would not be the optimal solution.
So, I have thought to introduce a field `modified_at` or `updated_at` and add a pre-commit hook which will update the date if there are any changes on those posts.

@iterative/websites If the above approach seems good then I will proceed with that else as @rogermparent  mentioned, we can merge this for 1st iteration and discuss and work on another issue/PR for improving the search especially for partial updating the indexes.

julieg18:
> So, I have thought to introduce a field `modified_at` or `updated_at` and add a pre-commit hook which will update the date if there are any changes on those posts.

This field would be in the blog's frontmatter? Would this approach be completely automated or would blog post writers need to manually add this field? If we do something like this, we'd probably want it to be completely automated 🤔 

> @iterative/websites If the above approach seems good then I will proceed with that else as @rogermparent mentioned, we can merge this for 1st iteration and discuss and work on another issue/PR for improving the search especially for partial updating the indexes.

Personally, I think we could merge as a 1st iteration and then work on partial updating later.

yathomasi:
>This field would be in the blog's frontmatter? Would this approach be completely automated or would blog post writers need to manually add this field? If we do something like this, we'd probably want it to be completely automated 🤔
Yes, the field will be on the blog's front matter. I just tried with the pre-commit script and no the blog author doesn't need to add the field. It will be automatically added on commit. The cons would be a little bit of delay on commit. 

I can push those changes here or should I do it on another branch?

julieg18:
@iterative/websites, is there anything blocking this pr from being merged as a first iteration?

rogermparent:
> is there anything blocking this pr from being merged as a first iteration?

I don't see anything that disqualifies it, and it seems the deploy error is just a random Heroku error, but I do think we should run another deploy attempt to ensure everything builds before merging to master.

rogermparent:
The redeploy works! Once conflicts are resolved, we should be good to go.

yathomasi:
> @iterative/websites, is there anything blocking this pr from being merged as a first iteration?

Yes, I think we are waiting for the review on UI changes @yalozhkin requested. 


> That said, do we have the right environment variables set up? I think we had a discussion on whether to have an index for review apps or not and were learning toward no- 

Yes, I had them setup on review env vars as well as production vars on heroku.

>does the build break if we don't have the algolia env vars on?

No, the build would not break if we don't have them set. It will just skip them.

yathomasi:
It looks like we have got positive feedback from @yalozhkin on slack.

Screen Shot 2022-02-03 at 23 41 39

We already have another PR opened for #3239. If we want to merge this we can continue with the discussion there.

yathomasi:
Looks like I had missed the response on the Figma board. 
Is there any reference for UI/UX, I can take for this implementation?

Screen Shot 2022-02-04 at 00 12 10

shcheklein:
Where would the back link lead? Back the main blog page - then it's not really needed, it's not critical at least.

As a next step (separate PR), let's please make the search bar hidden under the icon in the top menu or some other place (see medium for example)

I think now it grabs a bit too much attention at the main page + if we put an icon we can keep it at all the pages.

yathomasi:
>Where would the backlink lead? Back the main blog page - then it's not really needed, it's not critical at least.

Yeah, it's "Back to blog". Also, we can rely on browser back button, too.

>As a next step (separate PR), let's please make the search bar hidden under the icon in the top menu or some other place (see medium for example)
>
>I think now it grabs a bit too much attention at the main page + if we put an icon we can keep it at all the pages.

That sounds good. We can start a separate PR on improving the search placements after this get merged. I like how [Gatsby](https://www.gatsbyjs.com/) has setup the search.

julieg18 (approved):
Great work!

rogermparent (commented):
Looks fine to me, but weren't we going to remove the `dev_blogs` index and make only production index?

yathomasi:
> Looks fine to me, but weren't we going to remove the `dev_blogs` index and make only production index?

If so we would want to either make the search component visible only on the production server or make the base URL to the current website base URL like it's on Search Docs( the results always open on live production URL not locally.)

rogermparent:
> If so we would want to either make the search component visible only on the production server or make the base URL to the current website base URL like it's on Search Docs( the results always open on live production URL not locally.)

The latter sounds like a perfectly fine solution, even if we're unable to change the base url on deploy previews' search such that it's slightly broken on those previews like the docs search, the search is primarily meant for users on production so I'd say it's worth taking that downside.

yathomasi:
I think #3264 is also ready to get merged to this branch.
Once that gets merged, I think we can merge this to master as well.

yathomasi:
changes that need to be done:

- [ ]  change trigger from current branch to `master` branch
- [ ] remove current environment variable for index name which will default to `dvc_blogs` for the index name.

PR #3199, opened 2022-01-20
ref: missing `adds --jobs`

daavoo:


content/docs/command-reference/add.md
```diff
@@ -169,6 +169,12 @@ not.
   [remote storage](/doc/command-reference/remote) to transfer external target to
   (can only be used with `--to-remote`).
 
+- `-j <number>`, `--jobs <number>` - Only used if `--to-remote` is also passed.
+  Parallelism level for DVC to upload data to remote storage. The default value
+  is 4 \* cpu_count(). For SSH remotes, the default is 4. Note that the default
+  value can be set using the jobs config option with dvc remote modify. Using
+  more jobs may speed up the operation.
```
jorgeorpinel, line 176: For closer consistency with other refs that have this option:
```suggestion
- `-j <number>`, `--jobs <number>` -
  parallelism level for DVC to transfer data when using `--to-remote`. The default value
  is `4 \* cpu_count()`. For SSH remotes, the default is `4`. Using
  more jobs may speed up the operation.
```

jorgeorpinel: Committed. I removed "Note that the default value can be set using the jobs config option with dvc remote modify" too. It's a good note though, but it kind of overcomplicates the already-complicated description. I wonder if there's a better place to comment on this... ⌛

jorgeorpinel:
I wonder if `--jobs` also has this same effect in `import-url --to-remote`. [Currently](https://dvc.org/doc/command-reference/import-url#-j) it only refers to downloading the target data. @dberenbaum or @efiop, do you know? Thanks

jorgeorpinel (dismissed):

content/docs/command-reference/add.md
```diff
@@ -169,6 +169,10 @@ not.
   [remote storage](/doc/command-reference/remote) to transfer external target to
   (can only be used with `--to-remote`).
 
+- `-j <number>`, `--jobs <number>` - parallelism level for DVC to transfer data
+  when using `--to-remote`. The default value is `4 \* cpu_count()`. For SSH
+  remotes, the default is `4`. Using more jobs may speed up the operation.
```
jorgeorpinel, line 174: I wonder if `--jobs` also has this same effect in `import-url --to-remote`. [Currently](https://dvc.org/doc/command-reference/import-url#-j) it only refers to downloading the target data. @dberenbaum or @efiop, do you know? Thanks

jorgeorpinel:
@daavoo thanks again for this. Mind also completing the list of commands that have `--jobs` in https://dvc.org/doc/command-reference/remote/modify#available-parameters-for-all-remotes ? Thanks

content/docs/command-reference/add.md
jorgeorpinel: > removed "Note that the default value can be set using the jobs config option...

Maybe we should change the text (in all refs) to something like "The default is 4*cpu (can be set with `dvc remote modify jobs`) :...

> Linked to https://dvc.org/doc/command-reference/remote/modify#available-parameters-for-all-remotes

daavoo:
> @daavoo thanks again for this. Mind also completing the list of commands that have `--jobs` in https://dvc.org/doc/command-reference/remote/modify#available-parameters-for-all-remotes ?

Done.

content/docs/command-reference/add.md
daavoo: `--jobs` affects the `uploading` when using `dvc import-url --to-remote` so the description should be updated. This also affects `dvc update`

jorgeorpinel: > --jobs affects the uploading when using dvc import-url --to-remote

AFAIK `--to-remote` "streams" chunks of the data via the local env into the remote. Do you mean that `--jobs` affects this whole transferring strategy (both download and upload) but has no effect for downloading regular imports (no `--to-remote`) @daavoo ? Thanks

> If you're not sure please lmk and we'll check with someone else in the core team.

jorgeorpinel: Moved this Q to https://github.com/iterative/dvc.org/issues/3233

content/docs/command-reference/remote/modify.md
```diff
@@ -79,8 +79,10 @@ The following config options are available for all remote types:
 
 - `jobs` - change the default number of processes for
   [remote storage](/doc/command-reference/remote) synchronization operations
-  (see the `--jobs` option of `dvc push`, `dvc pull`, `dvc fetch`, `dvc status`,
-  and `dvc gc`). Accepts positive integers. The default is typically `4`.
+  (see the `--jobs` option of `dvc push`, `dvc pull`, `dvc fetch`, `dvc get`,
+  `dvc add`, `dvc import`, `dvc update`, `dvc import-url`, `dvc get-url`,
+  `dvc status` and `dvc gc`). Accepts positive integers. The default is
```
jorgeorpinel, line 84: Now that I see it the list seems unnecessarily long...
```suggestion
  (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc import`,
  `dvc update`, `dvc add --to-remote`, `dvc gc -c`, etc.). Accepts positive
  integers. The default is
```

jorgeorpinel (approved):

PR #3205, opened 2022-01-21
troubleshooting: add entry for pickle protocol errors

pmrowla:
For https://github.com/iterative/dvc/pull/7300

### DIFF
--- a/content/docs/user-guide/troubleshooting.md
+++ b/content/docs/user-guide/troubleshooting.md
@@ -112,15 +112,16 @@ experiments.
 
 You may encounter this error when using DVC on different Python versions with
 the same <abbr>DVC project</abbr> directory, for example having created the
-project on Python 3.8. in one environment and later attempting to update it from a
-Python 3.7 env. This is due to temporary [internal directories] that can be
+project on Python 3.8. in one environment and later attempting to update it from
+a Python 3.7 env. This is due to temporary [internal directories] that can be
 incompatible with older Python versions once created.
 
-In these rare situations, it is safe to remove the corresponding tmp
-directory and retry the DVC command. Specifically, one of:
+In these rare situations, it is safe to remove the corresponding tmp directory
+and retry the DVC command. Specifically, one of:
 
 - `.dvc/tmp/index`
 - `.dvc/tmp/md5s`
 - `.dvc/tmp/links`
 
-[internal files]: https://dvc.org/doc/user-guide/project-structure/internal-files
+[internal files]:
+  https://dvc.org/doc/user-guide/project-structure/internal-files
--- a/content/docs/user-guide/troubleshooting.md
+++ b/content/docs/user-guide/troubleshooting.md
@@ -123,5 +123,5 @@ and retry the DVC command. Specifically, one of:
 - `.dvc/tmp/md5s`
 - `.dvc/tmp/links`
 
-[internal files]:
+[internal directories]:
   https://dvc.org/doc/user-guide/project-structure/internal-files
--- a/content/docs/user-guide/troubleshooting.md
+++ b/content/docs/user-guide/troubleshooting.md
@@ -107,3 +107,20 @@ experiments.
 [git credentials]: https://git-scm.com/docs/gitcredentials
 [ssh git url]:
   https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_protocols
+
+## Failed to open pickled cache {#pickle}
+
+You may encounter this error when using DVC with multiple Python versions within
+a single local repository, such as when you use multiple virtual environments to
+test behavior across a range of Python versions. In certain cases it is possible
+that DVC may generate internal (temporary) files that are incompatibile with
+older Python versions (i.e. DVC run in Python 3.8 may generate files which is
+incompatible with DVC in Python 3.7).
+
+In the event that this occurs, it is safe to remove the relevant temporary
+file(s), and retry the DVC command. Specifically, the following directories (and
+their file contents) can be safely removed in this situation:
+
+- `.dvc/tmp/md5s`
+- `.dvc/tmp/links`
+- `.dvc/tmp/index`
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/troubleshooting.md
```diff
@@ -107,3 +107,20 @@ experiments.
 [git credentials]: https://git-scm.com/docs/gitcredentials
 [ssh git url]:
   https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_protocols
+
+## Failed to open pickled cache {#pickle}
+
+You may encounter this error when using DVC with multiple Python versions within
```
jorgeorpinel, line 113: Can we generalize/simplify this a bit to avoid mentioning pickling? Prob. an unnecessary implementation detail

Please see https://github.com/iterative/dvc/pull/7300#pullrequestreview-864458242

jorgeorpinel: And prob. link to https://dvc.org/doc/user-guide/project-structure/internal-files somewhere here 🙂

pmrowla: I'm not sure it's irrelevant here, because the error specifically happens due to using pickle in mixed python versions.

dberenbaum: Maybe explain the pickling issue so it's clear why it's mentioned? For example: "Pickle protocol version 5 (https://docs.python.org/3/library/pickle.html#data-stream-format) is used in Python 3.8+ but is not supported in earlier Python versions."

Also, maybe it's better to reverse the paragraph order so that the fix is explained up front and implementation details are explained later?

jorgeorpinel: > the error specifically happens due to using pickle in mixed python versions

But that's used internally in DVC right? That's what I mean by implementation detail.

Or is this about user code or data that uses pickling from different Python versions?

jorgeorpinel: UPDATE: OK I read the explanations in https://github.com/iterative/dvc/pull/7222#discussion_r789406572 and https://github.com/iterative/dvc/pull/7300#pullrequestreview-864458242 and I see it's an implementation detail indeed so I still incline to simplify this in some way that pickling isn't mentioned.

I'd be happy to make a specific suggestion or take this over but I still don't fully understand the problem... See my questions in https://github.com/iterative/dvc/pull/7300#discussion_r797215402.

pmrowla: I still think including that the error is due to pickling is relevant, otherwise it's unclear why the error occurs in the first place. Since DVC officially supports both Python 3.8+ and 3.7, it is non-obvious why DVC would work in a given repo in 3.8+ but not 3.7.

content/docs/user-guide/troubleshooting.md
```diff
@@ -107,3 +107,20 @@ experiments.
 [git credentials]: https://git-scm.com/docs/gitcredentials
 [ssh git url]:
   https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_protocols
+
+## Failed to open pickled cache {#pickle}
```
jorgeorpinel, line 111: To match error message in https://github.com/iterative/dvc/pull/7300
```suggestion
## Could not open pickled 'index/md5/links' cache {#pickle}
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -107,3 +107,20 @@ experiments.
...
+
+You may encounter this error when using DVC with multiple Python versions within
+a single local repository, such as when you use multiple virtual environments to
+test behavior across a range of Python versions. In certain cases it is possible
+that DVC may generate internal (temporary) files that are incompatibile with
+older Python versions (i.e. DVC run in Python 3.8 may generate files which is
+incompatible with DVC in Python 3.7).
```
jorgeorpinel, line 118: Per https://github.com/iterative/dvc/pull/7300#discussion_r797284091
```suggestion
You may encounter this error when using DVC on different Python versions with
the same <abbr>DVC project</abbr> directory, for example having created the
project on Python 3.8. in one environment and later attempting to update it from a
Python 3.7 env. This is due to temporary [internal directories] that can be
incompatible with older Python versions once created.
```

content/docs/user-guide/troubleshooting.md
```diff
@@ -107,3 +107,20 @@ experiments.
...
+
+In the event that this occurs, it is safe to remove the relevant temporary
+file(s), and retry the DVC command. Specifically, the following directories (and
+their file contents) can be safely removed in this situation:
+
+- `.dvc/tmp/md5s`
+- `.dvc/tmp/links`
+- `.dvc/tmp/index`
```
jorgeorpinel, line 126:
```suggestion
In these rare situations, it is safe to remove the corresponding tmp
directory and retry the DVC command. Specifically, one of:

- `.dvc/tmp/index`
- `.dvc/tmp/md5s`
- `.dvc/tmp/links`

[internal files]: https://dvc.org/doc/user-guide/project-structure/internal-files
```

jorgeorpinel (commented):
Taking this over

content/docs/user-guide/troubleshooting.md
jorgeorpinel: > I still think including that the error is due to pickling is relevant

Actually I just realized we never included that remark in the text, only in the section title. I would prefer to completely avoid it but since that's how it was merged in the core error message, I'll leave it here too.

jorgeorpinel (approved):
Updated. WDYT @pmrowla @dberenbaum ? Feel free to merge if you approve.

pmrowla:
I think one of the internal links is broken but otherwise it LGTM @jorgeorpinel

content/docs/user-guide/troubleshooting.md
```diff
@@ -107,3 +107,21 @@ experiments.
...
+incompatible with older Python versions once created.
+
+In these rare situations, it is safe to remove the corresponding tmp directory
+and retry the DVC command. Specifically, one of:
+
+- `.dvc/tmp/index`
+- `.dvc/tmp/md5s`
+- `.dvc/tmp/links`
+
+[internal files]:
```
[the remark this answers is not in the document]
jorgeorpinel: Yep, my bad! Fixing...

PR #3206, opened 2022-01-21
Stabilize Blog Post Meta Image Path

yathomasi:
I have continued with @julieg18 work on pr #3161. 

I did a runtime comparison for the script and found this implementation is comparatively faster. So, I ended up pushing the changes.

> Actually, I did some comparisons around the solutions, 
>- First one using gatsby-plugin-image, it took around - `19151 ms` for graphql query itself
>- Current Jimp implementation took around, `26121 ms` to complete
>- Sharp implementation took around, `4039 ms` to complete

For #1238


gatsby-node.js
```diff
@@ -40,3 +44,65 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+        await mkDirIfItDoesNotExist(dirPath)
+        return sharp(path.join(__dirname, 'static', 'uploads', imagePath))
+          .resize({ width: 850 })
+          .toFile(path.join(__dirname, 'public', 'blog', imagePath))
+          .catch(err => {
+            console.error(err)
+          })
+      })
+    )
+    // const end = Date.now()
```
shcheklein, line 103: please, let's not commit commented or dead code

gatsby-node.js
```diff
@@ -40,3 +44,65 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+
+    await Promise.all(
+      nodes.map(async node => {
+        const imagePath = node.picture.fields.sourcePath
+        const dirPath = path.dirname(
+          path.join(__dirname, 'public', 'blog', imagePath)
+        )
+        await mkDirIfItDoesNotExist(dirPath)
+        return sharp(path.join(__dirname, 'static', 'uploads', imagePath))
+          .resize({ width: 850 })
```
shcheklein, line 96: can we take this width from somewhere? it's not good to have it defined in multiple places

shcheklein:
@julieg18 @yathomasi 👍

Looks like: 

https://dvc-org-meta-image-path-yj1gwc.herokuapp.com/blog/images/2021-09-07/hell.png - browser cache doesn't work (it works on the current version dvc.org/blog). Probably it's related to the way it's deployed (preview app) - we need to test that everything works fine in prod. It's fine to merge and test.

Could

src/components/SEO/helper.ts
```diff
@@ -70,7 +70,7 @@ export const buildMetadata = (
     })
   }
   if (image) {
-    const imageUrl = siteUrl + getSrc(image)
+    const imageUrl = siteUrl + (getSrc(image) || image)
```
shcheklein, line 73: Q: why do we need this `||`?

julieg18: Our stable urls are only created in production. So I configured the code to send `SEO` a `gatsby-plugin-image` object if the site is running in development and send `SEO` the stable url string if the site is running in production. So `imageUrl` will either be the default src of a `gatsby-plugin-image` or just the image itself.

But if we don't care if the meta tags in development are showing a image that doesn't exist, we can just take this off and put `const imageUrl = siteUrl + image`.

gatsby-node.js
julieg18: Comments deleted!

gatsby-node.js
julieg18: We have constants for the blog width in `src/constants`.

gatsby-node.js
```diff
@@ -40,3 +45,62 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+
+    await Promise.all(
+      nodes.map(async node => {
+        const imagePath = node.picture.fields.sourcePath
+        const dirPath = path.dirname(
+          path.join(__dirname, 'public', 'blog', imagePath)
+        )
+        await mkDirIfItDoesNotExist(dirPath)
+        return sharp(path.join(__dirname, 'static', 'uploads', imagePath))
+          .resize({ width: BLOG.mageMaxWidthHero })
```
julieg18, line 96: Just noticed that `sharp` isn't actually resizing the images. I'll look into into what's wrong 🤔

src/components/Blog/Post/index.tsx
```diff
@@ -83,8 +83,10 @@ const Post: React.FC<IBlogPostData> = ({
               />
             </div>
           </div>
-
-          <HeroPic picture={picture} pictureComment={pictureComment} />
+          <HeroPic
+            picture={picture?.gatsbyImageData}
+            pictureComment={pictureComment}
+          />
```
rogermparent, line 89: My first inclination seeing this is that it'd be better if `HeroPic` takes `picture` directly and processes it inside the component, but that's at best a nitpick and there may be a good reason to do otherwise.

src/templates/blog-post.tsx
```diff
@@ -55,7 +58,14 @@ const BlogPostPage: React.FC<IBlogPostPageProps> = ({ data }) => {
...
         description={description}
-        image={picture}
+        image={
+          picture &&
+          (isProduction
+            ? `/blog/${picture.fields.sourcePath}`
+            : picture.gatsbyImageData)
+        }
+        imageHeight={picture?.gatsbyImageData.height}
+        imageWidth={picture?.gatsbyImageData.width}
```
rogermparent, line 68: same sentiment here with passing down `picture` instead of disassembling it here, with a similar conclusion of it being a nitpick.

rogermparent (commented):
I'm not loving how we do the resizing with `sharp` manually instead of taking advantage of gatsby queries, but I also didn't personally experience the issue that made us choose to do so. 

I do think the natural placement for these static blog files would be in the directory of the page itself with a static name of `image` or `cover`, but most meta tags require absolute URLs anyway so it wouldn't cut down complexity by that massive an amount.

If this happens to do what we want, I'm fine accepting it as a first iteration.

shcheklein:
@rogermparent 

> I'm not loving how we do the resizing with sharp manually instead of taking advantage of gatsby queries, but I also didn't personally experience the issue that made us choose to do so.

could you please summarize - what is the major difference? less code and it looks better or is it about some caching issues, etc?

what does Gatsby use internally? (I thought that it's also sharp)

src/components/Blog/Post/index.tsx
julieg18: To make the typescript less complex, I set `IBlogPostHeroPic.picture` (aka `HeroPic`'s defined prop value for `picture`) to only take a `IGatsbyImageData` object instead of entirety of `picture` which now includes `gatsbyImageData` and the `fields.sourcePath`.

src/templates/blog-post.tsx
julieg18: Similiar to your previous comment, I set them like so to make the typescript less complex. But if you think it would be better, I'm happy to set the `HeroPic` and `SEO` to take the entire `picture` object!

yathomasi:
@rogermparent 
>I'm not loving how we do the resizing with sharp manually instead of taking advantage of gatsby queries, but I also didn't personally experience the issue that made us choose to do so.

One of the main reasons to use the `sharp` plugin manually over gatsby queries was the execution speed. I found some significant differences. I have also mentioned the comparisons in the description of this PR.

rogermparent:
> could you please summarize - what is the major difference? less code and it looks better or is it about some caching issues, etc?
> what does Gatsby use internally? (I thought that it's also sharp)

My concern is both, but the latter is the most real concern: While the Gatsby image plugins like `gatsby-plugin-sharp` do use sharp internally, they wrap around it to allow use of Gatsby's internal caching and parallelism features (workers API) that don't come easy from using `sharp` directly.

I do want to stress that I'm happy to go with this if it does the job, at least as a first iteration, but the feeling I get is that when using a framework, stepping outside of it should only be done when absolutely necessary and I'm not sure that's the case here.

shcheklein:
@rogermparent I agree with this. Question is - why do we see it being slower that much, how does it affect the total build time (not only processing these specific images)?

rogermparent:
> One of the main reasons to use the sharp plugin manually over gatsby queries was the execution speed.

> why do we see it being slower that much, how does it affect the total build time (not only processing these specific images)?

I didn't see the exact implementation, but it depends on specifically how the plugin was used: my first guess is that if the implementation we tried used an extra `fixed` or `fluid` methods that generate multiple versions of the image instead of the lower-level `resize` method, that would be an easy explanation since that would be generating 4x the images compared to the direct-usage implementations. Otherwise, I'd have to look into it more to see what's going on.

gatsby-node.js
```diff
@@ -40,3 +45,62 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+    }
+  }
+}
+
+const mkDirIfItDoesNotExist = async folderPath => {
+  const doesDirExist = await checkIfDirExists(folderPath)
+  if (!doesDirExist) {
+    await mkdir(folderPath, { recursive: true })
+  }
+}
```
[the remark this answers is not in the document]
julieg18: Good idea! I've never heard of `mkdirp` before!

julieg18:
> > One of the main reasons to use the sharp plugin manually over gatsby queries was the execution speed.

> > why do we see it being slower that much

> I didn't see the exact implementation, but it depends on specifically how the plugin was used: my first guess is that if the implementation we tried used an extra fixed or fluid methods that generate multiple versions of the image instead of the lower-level resize method, that would be an easy explanation since that would be generating 4x the images compared to the direct-usage implementations. Otherwise, I'd have to look into it more to see what's going on.

The first working implementation was using `Jimp` to resize images instead of sharp. [I actually couldn't get the code to work](https://github.com/iterative/dvc.org/pull/3161#discussion_r789255588)  with using gatsby queries to resize the images.

gatsby-node.js
julieg18: Fixed! It's literally just a typo, `BLOG.mageMaxWidthHero` instead of `BLOG.imageMaxWidthHero` 😆

gatsby-node.js
```diff
@@ -40,3 +45,62 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+              fields {
+                sourcePath
+              }
+            }
+            date
+          }
+        }
+      }
+    `)
+
```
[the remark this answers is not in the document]
julieg18: Moved to the blog's `onCreateMarkdownContentNode`!

julieg18:
Updated the code based on everyone's comments! And now that everything is working correctly and all typescript errors are cleared, this pr is ready for another round of reviews!

rogermparent (approved):
Seems at least good enough as a first iteration, great improvements so far! Hopefully the browser caching issues @shcheklein experienced earlier will work out fine.

yathomasi (commented):
@julieg18 
Looks good to me.
Also, the images on preview deployment also look fine.

PR #3210, opened 2022-01-24
ref: clarifications around `add` and `import-url` for external targets

jorgeorpinel:
Related to the last task in #2121


content/docs/command-reference/add.md
```diff
@@ -374,33 +376,33 @@ outs:
...
+
 ## Example: Transfer to remote storage
 
 When you have a large dataset in an external location, you may want to track it
-as if it was in your project, but without downloading it locally (for now). The
-`--to-remote` option lets you do so, while storing a copy
-[remotely](/doc/command-reference/remote) so it can be
-[pulled](/doc/command-reference/plots) later.
+in your <abbr>project</abbr> without downloading it (yet), for example if
+there's not enough space in your current environment.
```
jorgeorpinel, line 388: This example is almost the same as the one in the `import-url` ref. Since the latter has the added benefit of keeping the connection to the data source (for later `update`s) and to remove the docs maintenance surface, should we remove the example from this ref? I can still link to the example in `import-url` from the option description here.

content/docs/command-reference/add.md
```diff
@@ -341,19 +344,18 @@ Only the hash values of the `dir/` directory (with `.dir` file extension) and
 ## Example: Transfer to the cache
 
 When you have a large dataset in an external location, you may want to add it to
-the <abbr>project</abbr> without having to copy it into the
-<abbr>workspace</abbr>. Maybe your local disk doesn't have enough space, but you
-have set up an
-[external cache](/doc/user-guide/managing-external-data#setting-up-an-external-cache)
```
dberenbaum, line 347: Any particular reason to reword this and move the basic use case to the bottom of the explanation? For me, the old way was easier to understand since it first explains that you lack space locally but have an external cache. Adding directly to cache doesn't have obvious benefit if the cache is local.

content/docs/command-reference/add.md
dberenbaum: Makes sense to me!

content/docs/command-reference/add.md
```diff
@@ -374,33 +376,33 @@ outs:
...
-`--to-remote` option lets you do so, while storing a copy
-[remotely](/doc/command-reference/remote) so it can be
-[pulled](/doc/command-reference/plots) later.
+in your <abbr>project</abbr> without downloading it (yet), for example if
+there's not enough space in your current environment.
+
+The `--to-remote` option lets you bootstrap the data locally by creating a
+`.dvc` file, while storing a copy [remotely](/doc/command-reference/remote).
+This way it can be [pulled](/doc/command-reference/plots) later, for example on
+a system tht can handle it.
```
dberenbaum, line 393:
```suggestion
a system that can handle it.
```

content/docs/command-reference/add.md
```diff
@@ -374,33 +376,33 @@ outs:
...
 
 When you have a large dataset in an external location, you may want to track it
-as if it was in your project, but without downloading it locally (for now). The
-`--to-remote` option lets you do so, while storing a copy
-[remotely](/doc/command-reference/remote) so it can be
-[pulled](/doc/command-reference/plots) later.
+in your <abbr>project</abbr> without downloading it (yet), for example if
+there's not enough space in your current environment.
+
+The `--to-remote` option lets you bootstrap the data locally by creating a
```
dberenbaum, line 390: It's not that clear to me what "bootstrap" means. What do you think about simplifying to:

> The `--to-remote` option creates a `.dvc` file locally, ...

content/docs/command-reference/add.md
```diff
@@ -374,33 +376,33 @@ outs:
...
+`.dvc` file, while storing a copy [remotely](/doc/command-reference/remote).
```
dberenbaum, line 391: When I read "storing a copy," it makes me think that the data might still be stored locally, and then copied to the remote. What about "storing the data"?

content/docs/command-reference/add.md
```diff
@@ -374,33 +376,33 @@ outs:
...
-```dvc
-$ ls
-data.xml.dvc
-```
-
-Whenever anyone wants to actually download the added data (for example from a
-system that can handle it), they can use `dvc pull` as usual:
+Even when nothing is downloaded locally, the operation still creates a `.dvc`
+file in the <abbr>workspace</abbr>. So whenever anyone wants to actually
+download the data, they can use `dvc pull` as usual:
```
dberenbaum, line 405: Nice summarization here 🔥

content/docs/command-reference/add.md
jorgeorpinel: @dberenbaum I gave this some thought and realized that adding "without having to copy it into the workspace" isn't the base case. The base case is simply to add data that is currently external. `dvc add` doesn't allow this unless you use `--external` (tracks it in place) or (in this case) specify an `--out path` inside the project.

jorgeorpinel: I clarified this further now though. The part about not having enough space locally (but an external cache) was missing to complete the motivation for using `--out` vs. just `cp` + `add`. PTAL!

jorgeorpinel: p.s. hmmm I thought this was going to be much clearer but I see I'm going in a circle here, a little bit.

The logic behind this case is not simple... I rewrote it completely again 😅 , PTAL -- I think it's better explained now.

content/docs/command-reference/add.md
jorgeorpinel: Removed "bootstrap" term. Went for "... lets you create a `.dvc` file for the target data
without downloading it locally."

content/docs/command-reference/add.md
jorgeorpinel: Removed "storing a copy". Went for "... transfers it directly to remote storage..."

content/docs/command-reference/add.md
jorgeorpinel: > > This example is almost the same as the one in the import-url ... should we remove the example from this ref?
>
> Makes sense to me!

UPDATE: After the latest round of updates I'm no longer 100% sure of this since for `add` you need to combine `--out` with `--to-remote`. So even when the situation is very similar, the actual recipe has that key difference. People who never use imports may prefer to stick to `add` and thus need to figure out that detail.

Keeping for now!

dberenbaum: Good point. Just to clarify, you can do `dvc add --to-remote` without `--out`:

```
$ dvc add --to-remote ../test.sh
100% Adding...|█████████████████████████████████████████████████|1/1 [00:00, 49.91file/s]
````

File: content/docs/command-reference/add.md [hunk 7]
```diff
@@ -374,33 +376,34 @@ outs:
...
+storage space.
+
+Using the `--to-remote` flag lets you create a `.dvc` file for the target data
+without downloading it locally. DVC transfers it directly to [remote storage]
+instead. This way it can be [pulled](/doc/command-reference/plots) later, for
+example on a system that can handle it.
 
-Let's add the `data.xml` to our remote storage from the given remote location:
+Let's add a `data.xml` file via HTTP in this way. Note that the `--out` (`-o`)
+option is also needed since the target data is outside the project:
```
dberenbaum, line 397: As noted above, `--out` should be optional.

content/docs/command-reference/add.md
dberenbaum: Much clearer now!

content/docs/command-reference/add.md
jorgeorpinel: > you can do dvc add --to-remote without --out

Hmm wait does that move the output to the workspace? ⌛

jorgeorpinel: UPDATE: Yeah it does... So `--to-remote` implies `--out .` ... 🤔

content/docs/command-reference/add.md
jorgeorpinel: Fixed, PTAL.

content/docs/command-reference/add.md
jorgeorpinel: > even when the situation is very similar, the actual recipe has that key difference (there's no -o in import-url)

I guess this wasn't that relevant after all since neither example uses `-o` now. The `add` example just notes that `-o .` is implied but that's not critical (already noted in the `--to-remote` option desc.)... We could remove it after all ⌛

content/docs/command-reference/add.md
```diff
@@ -379,41 +377,24 @@ outs:
...
 
 ```dvc
-$ ls
-data.xml.dvc
+$ dvc add https://data.dvc.org/get-started/data.xml --to-remote
 ```
 
-Whenever anyone wants to actually download the added data (for example from a
-system that can handle it), they can use `dvc pull` as usual:
+See this [`import-url` example][iutr] for more details.
```
jorgeorpinel, line 394: I left the example in `add` but reduced it to its bare minimum and linked to the one in `import-url` for details.

content/docs/command-reference/import-url.md
```diff
@@ -360,36 +360,35 @@ Running stage 'prepare' with command:
...
-
 A       data.xml
-1 file added and 1 file fetched
+1 file added
 ```
 
-Note that you can also use `dvc update --to-remote` to bring the import up to
-date in remote storage, without downloading anything.
+Use `dvc update --to-remote` to bring the import up to date in remote storage,
+without downloading anything.
```
jorgeorpinel, line 392: Other misc. copy edits 😬

content/docs/command-reference/add.md
```diff
@@ -155,16 +156,15 @@ not.
   > above).
 
 - `-o <path>`, `--out <path>` - specify a `path` to the desired location in the
-  workspace to place the `targets` (instead of using the current working
-  directory). Directories specified in the path will be created by this command.
+  workspace to place the `targets` (moving them from their current location).
```
dberenbaum, line 159:
```suggestion
  workspace to place the `targets` (copying them from their current location).
```

I think "moving" sounds like they will be removed from their current location.

content/docs/command-reference/import.md
```diff
@@ -82,8 +82,7 @@ To actually [version the data](/doc/start/data-and-model-versioning), `git add`
 
 - `-o <path>`, `--out <path>` - specify a `path` to the desired location in the
   workspace to place the downloaded file or directory (instead of using the
-  current working directory). Directories specified in the path must already
-  exist, otherwise this command will fail.
+  current working directory).
```
jorgeorpinel, line 85: Removed unnecessary note for `import -o` (should be intuitive).

> Tis behavior isn't consistent across commands though (I messaged about it internally).

content/docs/command-reference/add.md
```diff
@@ -155,16 +156,15 @@ not.
...
+  [example](#example-transfer-to-an-external-cache)).
 
-  > Note that this can be combined with `--to-remote` to avoid storing the data
-  > locally, in which case the give `path` is only used in `dvc.yaml`.
-
-- `--to-remote` - add an external target, but don't move it into the workspace,
-  nor cache it. [Transfer it](#example-transfer-to-remote-storage) it directly
-  to remote storage (the default one, unless `-r` is specified) instead. Use
-  `dvc pull` to get the data locally.
+- `--to-remote` - add a target that's outside the project, but neither move it
```
dberenbaum, line 163:
```suggestion
- `--to-remote` - add a target that's outside the project, but neither copy it
```

content/docs/command-reference/update.md
```diff
@@ -10,7 +10,7 @@ usage: dvc update [-h] [-q | -v] [--rev <commit>] [-R] [--to-remote]
                   [-r <name>] [-j <number>] targets [targets ...]
 
 positional arguments:
-  targets      Import stage .dvc files to update. Using -R, directories
+  targets      Import .dvc files to update. Using -R, directories
```
jorgeorpinel, line 13: Found this outdated terminology in the `update` ref. only (will check core repo too ⌛). It's "import `.dvc file` everywhere else in docs.

content/docs/command-reference/update.md
```diff
@@ -46,14 +46,14 @@ $ dvc update --rev master
...
-  No changes are done in the <abbr>workspace</abbr>. Use `dvc pull` to get the
-  data locally. This option can't be used with DVC or Git repository imports.
+  directory and its subdirectories for import `.dvc` files to inspect. If there
+  are no directories among the targets, this option has no effect.
+
+- `--to-remote` - update a `.dvc` file created with `dvc import-url` without
+  downloading the latest data.
+  [Transfer it](/doc/command-reference/import-url#example-transfer-to-remote-storage)
+  directly to remote storage instead (the default one unless one is specified
+  with `-r`). Use `dvc pull` to get the data locally.
```
jorgeorpinel, line 56: Made all `--to-remote` descriptions consistent and as simple as possible, including in `update` now.

jorgeorpinel (commented):
OK @dberenbaum there's sizeable changes after your last comment so here's a summary of them 👇 Thanks

content/docs/command-reference/update.md
jorgeorpinel: p.s. @dberenbaum I see you started reviewing just before I made this summary 😅 

FYI the only newer commit is the one related to this comment, specifically https://github.com/iterative/dvc.org/pull/3210/commits/bf1cde2b5d0d5da37e7059f1fbe978bf8e51d28a.

content/docs/command-reference/add.md
```diff
@@ -379,41 +377,24 @@ outs:
...
-          --to-remote
-```
+## Example: Transfer to remote storage
 
-The only difference that dataset is transferred straight to remote, so DVC won't
-control the remote location you gave but rather continue managing your remote
-storage where the data is now on. The operation will still be resulted with an
-`.dvc` file:
+Use `--to-remote` to create a `.dvc` file for the operation without downloading
+data, transferring it directly to [remote storage] instead:
```
dberenbaum, line 388: Do you think it's worth mentioning the use case here? You may not have enough storage locally, but you want to pull it later on a machine with more space.

dberenbaum:
Thanks @jorgeorpinel! Just left a few small comments.

content/docs/command-reference/add.md
jorgeorpinel: Updated the motivation in both examples and reinstated everything else in the case of `add`. Linking to `import-url` for details stopped making sense (again).

dberenbaum (approved):

PR #3216, opened 2022-01-26
`exp remove`: Update help for `--all` flag

daavoo:
It is not true that `--all` includes `--queue`. Options need to be combined in order to effectively remove all ran and queued experiments

### DIFF
--- a/content/docs/command-reference/exp/remove.md
+++ b/content/docs/command-reference/exp/remove.md
@@ -26,7 +26,7 @@ With `--queue`, the list of experiments awaiting execution is cleared instead.
 - `--queue` - remove all experiments that haven't been run yet (defined via
   `dvc exp run --queue`).
 
-- `-A`, `--all` - remove all experiments (includes `--queue`).
+- `-A`, `--all` - remove all experiments.
 
 - `-g`, `--git-remote` - Name or URL of the Git remote to remove the experiment
   from
--- a/content/docs/command-reference/exp/remove.md
+++ b/content/docs/command-reference/exp/remove.md
@@ -26,7 +26,8 @@ With `--queue`, the list of experiments awaiting execution is cleared instead.
 - `--queue` - remove all experiments that haven't been run yet (defined via
   `dvc exp run --queue`).
 
-- `-A`, `--all` - remove all experiments.
+- `-A`, `--all` - remove all experiments that have been run. Use `--queue` to
+  remove queued ones.
 
 - `-g`, `--git-remote` - Name or URL of the Git remote to remove the experiment
   from


content/docs/command-reference/exp/remove.md
```diff
@@ -26,7 +26,7 @@ With `--queue`, the list of experiments awaiting execution is cleared instead.
 - `--queue` - remove all experiments that haven't been run yet (defined via
   `dvc exp run --queue`).
 
-- `-A`, `--all` - remove all experiments (includes `--queue`).
```
shcheklein, line 29: should we then write it explicitly:

(except `--queue` experiments) ?

content/docs/command-reference/exp/remove.md
```diff
@@ -26,7 +26,7 @@ With `--queue`, the list of experiments awaiting execution is cleared instead.
...
+- `-A`, `--all` - remove all experiments.
```
jorgeorpinel, line 29:
```suggestion
- `-A`, `--all` - remove all experiments that have been run. Use `--queue` to
  remove queued ones.
```

content/docs/command-reference/exp/remove.md
```diff
@@ -26,7 +26,8 @@ With `--queue`, the list of experiments awaiting execution is cleared instead.
 - `--queue` - remove all experiments that haven't been run yet (defined via
   `dvc exp run --queue`).
 
-- `-A`, `--all` - remove all experiments (includes `--queue`).
+- `-A`, `--all` - remove all experiments that have been run. Use `--queue` to
+  remove queued ones.
```
jorgeorpinel, line 30: Does this apply as well for `exp push/list/pull` ? Cc @dberenbaum and @karajan1001 (per #3191-3)

jorgeorpinel: And what about `exp show -a/A`? Should we mention that in that case queued experiments ARE included in the output?

karajan1001: > Does this apply as well for exp push/list/pull ? Cc @dberenbaum and @karajan1001 (per #3191-3)

Currently `push/list/pull` does not work on queued ones. 

> And what about exp show -a/A? Should we mention that in that case queued experiments ARE included in the output? May be a product question on whether the behavior is consistent, not sure.

This is of course a problem. Cc @pmrowla .

dberenbaum: Yeah, they are currently excluded from everything except for `exp run --queue`, and `exp remove --queue`, and `exp show`. For now, I think that behavior is fine if correctly documented.

Eventually, queue operations should be in a separate command. See https://github.com/iterative/dvc/pull/7275#issuecomment-1021947145 and https://github.com/iterative/dvc/issues/5615.

daavoo:
@jorgeorpinel Should we merge this and extract (if needed) https://github.com/iterative/dvc.org/pull/3216#discussion_r793348338 ?

content/docs/command-reference/exp/remove.md
daavoo: > And what about exp show -a/A? Should we mention that in that case queued experiments ARE included in the output?

Just wanted to clarify that queued experiments will be shown even without `-a/A` flag.

iesahin (approved):
Thank you @daavoo

PR #3229, opened 2022-01-27
Fix Broken Anchor Links in `/command-reference/exp/show`

julieg18:
* adds `loading: auto` to gatsby-remark-images options, which turns off lazy loading on doc images and blog images

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -115,7 +115,8 @@ const plugins = [
           options: {
             maxWidth: BLOG.imageMaxWidth,
             withWebp: true,
-            quality: 90
+            quality: 90,
+            loading: 'auto'
           }
         },
         'gatsby-remark-responsive-iframe',


gatsby-config.js
```diff
@@ -115,7 +115,8 @@ const plugins = [
           options: {
             maxWidth: BLOG.imageMaxWidth,
             withWebp: true,
-            quality: 90
+            quality: 90,
+            loading: 'auto'
```
julieg18, line 119: The code that scrolls to the desired element runs only when all the images on the page loads in. `gatsby-remark-plugin` sets our png/jpg doc images to load lazily by default. Since the images in `/exp/show` are at the bottom of the page, they are never loaded in until we start scrolling down, which is breaking the anchor links. 

To fix this, I added the `loading: 'auto'` option to turn off lazy loading in the images.

iesahin (approved):
Thank you @julieg18

PR #3239, opened 2022-01-31
Search blogs - Partial indexing

yathomasi:
added pre-commit script to run node script to add/update `updated_at` field if blog post(markdown) is added or updated.

The script used the `gray-matter` package to parse the markdown file. Gatsby also uses the same package. It seems like it parses the date field and converts it into a `Date` string. I wouldn't consider that an issue, as it would only affect visually and we may implement tweaks as mentioned on the issue [Disable date parsing](https://github.com/jonschlinkert/gray-matter/issues/62).

If this looks good then we can integrate the partial indexing using the `updated_at` field. 
@iterative/websites

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "gatsby-plugin-parent-resolvers": "^1.0.1",
     "github-markdown-css": "^5.0.0",
     "graphql": "^16.0.1",
+    "gray-matter": "^4.0.3",
     "iso-url": "^1.2.1",
     "isomorphic-fetch": "^3.0.0",
     "lodash": "^4.17.21",
@@ -169,7 +170,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn format-staged && yarn lint-staged"
+      "pre-commit": "bash scripts/pre-commit.sh"
     }
   },
   "lint-staged": {
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+git diff --cached --name-status | egrep -i "(A|M).*content\/blog\/.*\.(md|mdx)" | while read a b; do node ./scripts/update-blog-post.js $b; git add $b; done
+yarn format-staged && yarn lint-staged
--- a/scripts/update-blog-post.js
+++ b/scripts/update-blog-post.js
@@ -1,0 +1,15 @@
+const fs = require('fs').promises
+const matter = require('gray-matter')
+
+const filePath = process.argv[2]
+if (filePath) {
+  file = matter.read(filePath)
+  const { data: currentData } = file
+  const updatedData = {
+    ...currentData,
+    updated_at: new Date().toISOString()
+  }
+  file.data = updatedData
+  const updatedFileContent = matter.stringify(file)
+  fs.writeFile(filePath, updatedFileContent)
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -9039,7 +9039,7 @@ graphql@^16.0.1:
 
 gray-matter@^4.0.3:
   version "4.0.3"
-  resolved "https://registry.npmjs.org/gray-matter/-/gray-matter-4.0.3.tgz"
+  resolved "https://registry.yarnpkg.com/gray-matter/-/gray-matter-4.0.3.tgz#e893c064825de73ea1f5f7d88c7a9f7274288798"
   integrity sha512-5v6yZd4JK3eMI3FqqCouswVqwugaA9r4dNZB1wwcmrD02QkV5H0y7XBQW8QwQqEaZY1pM9aqORSORhJRdNK44Q==
   dependencies:
     js-yaml "^3.13.1"


shcheklein (changes requested):
First of all, nice work and an interesting approach @yathomasi ! As we discussed during the planning meeting, though, this implementation might be fragile since it depends on the presence of the hook + additional changes to the files. There is no way to guarantee / see that hook was got activated, there are chances that we might miss the updates and no easy way to detect that.

As we discussed, let's try to do a bit of research on how to use CI/CD for this.

yathomasi:
It looks like we can perform indexing using the official `algoliasearch` package. There might be some caveats we might fall into so we could use the approach followed by [gatsby-plugin-algolia](https://github.com/algolia/gatsby-plugin-algolia/blob/master/gatsby-node.js) as reference.

So, we can write a proper script that can perform the indexing of the data and it can be invoked from a CI/CD pipeline. But, we need to think of the number of indices we would want to use. It could only be a single production_index and is only updated on a production build(our docs search works that way). But, having indices for each build would be more challenging, would consume more resources, and also search would not be that much important for review apps, staging and dev builds.

I am still trying to figure out how I should approach it.

rogermparent:
> we need to think of the number of indices we would want to use. It could only be a single production_index and is only updated on a production build(our docs search works that way). But, having indices for each build would be more challenging, would consume more resources, and also search would not be that much important for review apps, staging and dev builds.

I agree that having a single index is the way to go, as maintaining an index for every review app or even a single one for dev builds would likely be a very unfavorable cost/benefit that overall isn't worth it.

yathomasi:
@iterative/websites Actually I am getting the blocker here. I am not able to figure out how I can get the blog's attributes and content data with some script that runs on some CI/CD pipeline.
Gatsby's data fetching with Graphql only works either during Build time or Client Runtime. A package like `gray-matter` is only suitable for front matter. Is there any approach I should look for?

shcheklein:
@yathomasi sure, let's try to brainstorm together and unblock you.

> Gatsby's data fetching with Graphql only works either during Build time or Client Runtime.

could we create a subcommand (`yarn index`) that would specifically run some isolated part of the Gatsby pipeline?

If we run the build this way - could we then rely on Gatsby cache to detect changes? 


blog file -> Gatsby task (internally pings Algolia) -> Gatsby caches tasks -> next time when it runs if blog file hash is the same Gatsby won't be running build for this Node again

This is by best guess of what is possible, curios about @rogermparent 's take on this.

rogermparent:
> could we create a subcommand (`yarn index`) that would specifically run some isolated part of the Gatsby pipeline?
> If we run the build this way - could we then rely on Gatsby cache to detect changes?
> 
> blog file -> Gatsby task (internally pings Algolia) -> Gatsby caches tasks -> next time when it runs if blog file hash is the same Gatsby won't be running build for this Node again

While theoretically possible, it would be incredibly hacky and almost certainly not work very well without a massive amount of work- conditionally droppping page creation as a whole may omit most image processing, which we all know is the bulk of the time for builds, but if we try to prune the build even further to only run on files in the diff we'd need to go even hackier and engineer some way to selectively build only the pages that change.

Beyond all the challenges of actually implementing this, there's the simple fact that we don't have access to cache in GitHub CI, which is the whole point of running this limited build instead of just piggybacking off the real build. We'd have to implement a whole secondary caching mechanism, and I have heavy doubts it'd work well and be worth the extra work.

From what I can see, there's two viable ways to do partial indexing:

1. Run an independent CI action on master commits that analyzes the diff and updates Algolia by adding/removing search nodes with data derived from non-Gatsby node processing of the markdown files. To @yathomasi's blocker on this, my first assumption would be that using `remark` itself would be the best way to process the markdown content beyond frontmatter, since it's what we use on the real pipeline.

2. Start pulling Git history (even just a minimal depth last commit since we're only interested in what changed since last build) on master builds and go back to the original attempt at using Git history to enable partial indexing.

My preferred method would be the second, but there's also edge cases and considerations with that (e.g. what if we merge another PR before the first's update has gone through?). I'm also unsure how the gatsby plugin can handle removed pages without replacing the index each build, but we rarely if ever remove blog posts so that could be acceptable.

shcheklein:
> there's the simple fact that we don't have access to cache in GitHub CI

just 2cs to this - I think we do, you can cache arbitrary things on CI and restore them. Or I'm missing your point Roger?

When I was describing the pipeline I was thinking about completely separate Gatsby-like Graphql pipeline, not the one we have, so it can be customized to disable images, etc.

> Start pulling Git history

is it on CI or Heroku?

yathomasi:
> 2\. Start pulling Git history (even just a minimal depth last commit since we're only interested in what changed since last build) on master builds and go back to the original attempt at using Git history to enable partial indexing.
> 
> My preferred method would be the second, but there's also edge cases and considerations with that (e.g. what if we merge another PR before the first's update has gone through?)

I would also prefer this method but I think we should also consider pulling git history, querying blog files to get the last git committed date would also add to the build time. 

> I'm also unsure how the gatsby plugin can handle removed pages without replacing the index each build, but we rarely if ever remove blog posts so that could be acceptable.

I think the plugin creates the temporary indices and once finished it replaces the existing indices.

yathomasi:
Closing in favor of #3264

PR #3240, opened 2022-01-31
Stop list item markers from appearing in Google Chrome's auto dark mode

### LINKED ISSUE #2794: Fix auto-dark mode

julieg18:
* hides the nav's list item markers
* fixes the last bug mentioned in #2794

Fixes #2794

### DIFF
--- a/src/components/LayoutHeader/Nav/styles.module.css
+++ b/src/components/LayoutHeader/Nav/styles.module.css
@@ -4,6 +4,10 @@
   flex-direction: row;
   align-items: center;
 
+  ul {
+    list-style-type: none;
+  }
+
   @media (--xs-scr) {
     display: none;
   }


src/components/LayoutHeader/Nav/styles.module.css
```diff
@@ -4,6 +4,10 @@
   flex-direction: row;
   align-items: center;
 
+  ul {
+    list-style-type: none;
```
julieg18, line 8: dvc.org nav in Chrome's Auto Dark Mode:

image

PR nav: 
image

PR #3242, opened 2022-02-01
Small theme bugfixes

rogermparent:
~This PR takes another step in the direction of making the theme package independent by adding all the packages it uses to its `package.json` file.~

On reflection, I think merging the addition of these packages into the local `dvc.org` version isn't necessary, and can only complicate things. I've copied the original changes into the start of the independent theme repo and left some small fixes that I think we still want to pick up.

Remaining steps:
1. Let the theme actually make a `yarn.lock`, give it its own repo, and publish it.
2. Use theme in `dvc.org` and remove the local version of the code in `plugins`
3. Remove the webpack/tsconfig snippets used for this local half-step
4. Remove any node packages that are used in the theme but nowhere else in the site
5. Integrate the theme into cml.dev


plugins/gatsby-theme-iterative-docs/package.json
```diff
@@ -4,8 +4,31 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "lint-ts": "tsc --noEmit --skipLibCheck && eslint --ext .json,.js,.ts,.tsx ."
   },
```
rogermparent, line 8: `lint-ts` helps ensure none of the packages were missed and nothing else ended up broken.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Layout/SearchForm/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import Promise from 'promise-polyfill'
-import { loadResource } from 'gatsby-theme-iterative-docs/src/utils/front/resources'
+import { loadResource } from '../../../../utils/front/resources'
```
rogermparent, line 3: Fixed a mistake left over from #3170

plugins/gatsby-theme-iterative-docs/src/typings.ts
```diff
@@ -0,0 +1,51 @@
...
+  type EaseModule = {
+    [key: string]: EaseFunction
+  }
+  const ease: EaseModule
+  export = ease
+}
+
+declare module 'iso-url' {
+  export const URL: typeof window.URL
+}
```
rogermparent, line 51: This file is copied from the base site because it's required for the same sorts of imports.

plugins/gatsby-theme-iterative-docs/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitThis": true,
+    "noFallthroughCasesInSwitch": true,
+    "allowSyntheticDefaultImports": true,
+    "allowJs": true,
+    "alwaysStrict": true
+  },
+  "include": ["./src/**/*", "./*.js"]
+}
```
rogermparent, line 24: Required for `tsc`-based linting.

plugins/gatsby-theme-iterative-docs/package.json
```diff
@@ -4,8 +4,31 @@
...
   "author": "",
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
```
julieg18, line 11: I looked through the `gatsby-theme-iterative-docs` and spotted some dependences that aren't here:

* [github-slugger](https://github.com/iterative/dvc.org/blob/master/plugins/gatsby-theme-iterative-docs/createPages.js#L1)
* [lodash](https://github.com/iterative/dvc.org/blob/master/plugins/gatsby-theme-iterative-docs/src/components/Documentation/TutorialsLinks/index.tsx#L3)
* [title-case](https://github.com/iterative/dvc.org/blob/master/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js#L23)
* [gatsby-link](https://github.com/iterative/dvc.org/blob/master/plugins/gatsby-theme-iterative-docs/src/components/Link/index.tsx#L4)
* [react-collapse](https://github.com/iterative/dvc.org/blob/master/plugins/gatsby-theme-iterative-docs/src/components/Documentation/Layout/SidebarMenu/Item/index.tsx#L2)

Do they need to be added to the dependencies?

julieg18 (commented):
Question about some missing dependencies, but besides that, we should be good to merge!

plugins/gatsby-theme-iterative-docs/package.json
rogermparent: You're right! One thing to note is that `gatsby-link` should actually be removed as it's recommended to use the export from the main `gatsby` package, as per [the README for `gatsby-link`](https://www.npmjs.com/package/gatsby-link).

I'll add these as well as make the `gatsby-link` change.

Also, having both `react-collapsible` and `react-collapse` confused me a bit.

yathomasi (commented):
Looks good. I think we should publish this to a package now and check with cml.dev to find more.
If it works perfectly great else if we find some issues then we can fix them accordingly.

rogermparent:
Looking back at this, I don't think we actually need to merge the `package.json` into `dvc.org`, but rather just use it in the separated theme package (which I've already started on).
In the meantime, I'll strip out all but the fixes so we can still get those in.

julieg18 (approved):
Looks good!

PR #3251, opened 2022-02-02
lock: update info. around locking and "parallel" dvc processes

### LINKED ISSUE #860: cmd ref: document releasing of locks in `dvc run`
Context: https://github.com/iterative/dvc/pull/2584

jorgeorpinel:
Closes #860

### DIFF
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -287,7 +287,7 @@ sorting only applies to experiment groups (sharing a parent commit).
 
 ## Example: Parallel coordinates plot (PCP)
 
-To generate an interactive parallel coordiantes plot based on the experiments
+To generate an interactive parallel coordinates plot based on the experiments
 and their parameters:
 
 ```dvc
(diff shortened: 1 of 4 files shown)


jorgeorpinel:
@efiop @shcheklein need some help to determine a more complete description for `.dvc/tmp/lock` in https://dvc.org/doc/user-guide/project-structure/internal-files and perhaps an updated text in https://dvc.org/doc/user-guide/troubleshooting#lock-issue. Thanks

efiop:
@jorgeorpinel Sorry for the delay. Not sure I understand what kind of description you have in mind. It is just an internal file that locks the whole repo from the user perspective.

iesahin:
What commands can be run with this lock in place? No commands? Some commands? @efiop

jorgeorpinel:
> Not sure I understand what kind of description you have in mind

@efiop I'm trying to address #860 which is very old. I wanted to close it but we're unsure on whether some existing docs around locking behavior are current (see https://github.com/iterative/dvc.org/issues/860#issuecomment-1023712795).

> What commands can be run with this lock in place?

This is also unclear to me. I think it's only used during `(exp) run/repro` ?

iesahin:
Given the circumstances, it will probably take a very long time to get an answer from @efiop and we can better do whatever we can on our own. I think this is good as it is. @jorgeorpinel

jorgeorpinel:
I'm waiting for an answer on https://github.com/iterative/dvc.org/issues/860#issuecomment-1063651617 for now.

PR #3255, opened 2022-02-03
ref: Remove outdated SSH default jobs.

daavoo:
The default to `4` is no longer True.

Per https://github.com/iterative/dvc/pull/7339

### DIFF
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -81,7 +81,7 @@ The following config options are available for all remote types:
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
   import`, `dvc update`, `dvc add --to-remote`, `dvc gc
-  -c`, etc.). Accepts positive integers. The default is `4 \* cpu_count()`.
+  -c`, etc.). Accepts positive integers. The default is `4 * cpu_count()`.
 
   ```dvc
   $ dvc remote modify myremote jobs 8
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -81,9 +81,7 @@ The following config options are available for all remote types:
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
   import`, `dvc update`, `dvc add --to-remote`, `dvc gc -c`, etc.). Accepts
-  positive integers.
-
-  The default is `4 * cpu_count()`.
+  positive integers. The default is `4 * cpu_count()`.
 
   ```dvc
   $ dvc remote modify myremote jobs 8
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -80,8 +80,10 @@ The following config options are available for all remote types:
 - `jobs` - change the default number of processes for
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
-  import`, `dvc update`, `dvc add --to-remote`, `dvc gc
-  -c`, etc.). Accepts positive integers. The default is `4 * cpu_count()`.
+  import`, `dvc update`, `dvc add --to-remote`, `dvc gc -c`, etc.). Accepts
+  positive integers.
+
+  The default is `4 * cpu_count()`.
 
   ```dvc
   $ dvc remote modify myremote jobs 8
--- a/content/docs/command-reference/remote/modify.md
+++ b/content/docs/command-reference/remote/modify.md
@@ -81,7 +81,7 @@ The following config options are available for all remote types:
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
   import`, `dvc update`, `dvc add --to-remote`, `dvc gc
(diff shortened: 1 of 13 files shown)


content/docs/command-reference/remote/modify.md
```diff
@@ -81,7 +81,7 @@ The following config options are available for all remote types:
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
   import`, `dvc update`, `dvc add --to-remote`, `dvc gc
-  -c`, etc.). Accepts positive integers. The default is typically `4`.
+  -c`, etc.). Accepts positive integers. The default is `4 \* cpu_count()`.
```
iesahin, line 84: `\` in `4 \*` has a special meaning? @daavoo

iesahin (approved):
This PR removes the specific --jobs information for SSH remotes. Thank you @daavoo

content/docs/command-reference/remote/modify.md
iesahin: Can merge after the fix :)

shcheklein:
> The default to 4 is no longer True.

is it a problem that should be fixed? (initially if was done for a reason, curious if we should consider this as a regression)

iesahin (changes requested):
It looks this requires the bug in https://github.com/iterative/dvc/issues/7319 to be resolved. I thought `--jobs` is now identical to the value in other remotes, but it looks it's not? (What is it now? 1?)

dberenbaum:
I believe this is an attempt to accurately reflect current behavior, but I also think @dtrifiro is looking into a fix, so these docs changes will likely be short-lived. Does that sound right @daavoo @dtrifiro?

iesahin:
If we remove those sentences, it means `--jobs` for SSH is no different than other remotes, `4 * cpu_count`. I would like to know the ground truth for this. 

BTW, why do we set number of connections by the CPU count? It's usually not dependent to the number of cores, but to the network speed or server configuration. Setting this to `4` _for all remotes_ looks more sensible to me.

daavoo:
> If we remove those sentences, it means `--jobs` for SSH is no different than other remotes, `4 * cpu_count`. I would like to know the ground truth for this.

We used to have a lower default value hardcoded for SSH remotes and that is no longer the case.  The current ground truth is that all remotes have the same default number of jobs, including SSH ones.

Apart from that, there is another bug, reported in https://github.com/iterative/dvc/issues/7319 , that makes `--jobs` not being correctly passed downstream, making the `--job` argument useless as a workaround for avoiding issues due to having too many default jobs in SSH (or other) remotes.

> BTW, why do we set number of connections by the CPU count? It's usually not dependent to the number of cores, but to the network speed or server configuration. Setting this to `4` _for all remotes_ looks more sensible to me.

I have no context on the reasons behind the chosen default. Based on user reports, all I can tell is that for some remotes the default number seems to be too high and is the reason behind some issues.

>  @dtrifiro is looking into a fix, so these docs changes will likely be short-lived. Does that sound right @daavoo @dtrifiro?

Today we have clarified that @dtrifiro has already looked into the `--jobs` not being passed and will eventually look into having reasonable defaults (correct me if I'm wrong, @dtrifiro). 

It might be True that these changes will be short-lived in that case, closing for now.

dtrifiro:
> Today we have clarified that @dtrifiro has already looked into the `--jobs` not being passed and will eventually look into having reasonable defaults (correct me if I'm wrong, @dtrifiro).

That's correct, working on this as part of #7319

iesahin:
> Today we have clarified that @dtrifiro has already looked into the --jobs not being passed and will eventually look into having reasonable defaults (correct me if I'm wrong, @dtrifiro).

I think this makes the current default value for SSH something other than `4 * cpu_count` but let's not turn this into nitpicking. :) Thank you very much @daavoo @dtrifiro

daavoo:
Reopened per https://github.com/iterative/dvc/pull/7339#pullrequestreview-889523146

content/docs/command-reference/remote/modify.md
```diff
@@ -80,8 +80,10 @@ The following config options are available for all remote types:
 - `jobs` - change the default number of processes for
   [remote storage](/doc/command-reference/remote) synchronization operations
   (see the `--jobs` option of dvc push`, `dvc pull`, `dvc get`, `dvc
-  import`, `dvc update`, `dvc add --to-remote`, `dvc gc
-  -c`, etc.). Accepts positive integers. The default is typically `4`.
+  import`, `dvc update`, `dvc add --to-remote`, `dvc gc -c`, etc.). Accepts
+  positive integers.
+
+  The default is `4 * cpu_count()`.
```
jorgeorpinel, line 86:
```suggestion
  positive integers. The default is `4 * cpu_count()`.
```

jorgeorpinel (approved):
So this no longer an unexpected temporary behavior? Rather temporarily permanent? If so OK (but no rush to merge as https://github.com/iterative/dvc/pull/7339 probably won't be released soon).

iesahin (approved):
The docs removes the specific settings for SSH remotes, now they are to be set by `--jobs`. 

I doubt we'll keep this, as making 32 SSH connections on an 8 core machine will probably lead to more issues/PRs/changes, but now this one correctly reflects the DVC behavior. Thank you @daavoo @dtrifiro

jorgeorpinel:
> removes the specific settings for SSH remotes, now they are to be set by --jobs

So there is no default for SSH remotes? The changes in https://github.com/iterative/dvc/pull/7339/files would imply that it's 4 * CPUs like all other remotes.

iesahin:
> So there is no `--jobs` default for SSH remotes?

The, the core team removed specific setting of `4` and now it's `4 * cpu` as any other remote.

Issue #3179, opened 2022-01-12
tabs: way to have same height in all elements


jorgeorpinel:
Currently the height of the Toggle element varies depending on the active tab. This makes sense as the default behavior, not to make any assumptions.

Could it have a prop to enable a stable height across all tabs though? It'd have to be the tallest one I suppose.

- [ ] Port to other engines if implemented 🙂

jorgeorpinel:
p.s. should we create an **iterative/web** team for tagging on these? @rogermparent @julieg18

julieg18:
> Could it have a prop to enable a stable height across all tabs though? It'd have to be the tallest one I suppose.

Makes sense, it would make tab switching be less jerky. We should be able to implement this! But what if we looked into making the content of a tab behave like a details component? Have the content of the tab expand smoothly instead? Then we could just apply that to all tabs without needing any props 🤔

jorgeorpinel:
> what if we looked into making the content of a tab behave like a details component? Have the content of the tab expand smoothly

Such an animation would definitely improve the default behavior. I may still want this feature on top but it would be more of a nice-to-have at that point 🙂  Up to you.

> We do! **iterative/websites**

Oh lol k

jorgeorpinel:
> I may still want this feature on top but it would be more of a nice-to-have at that point 🙂

Well... I just tried using tabs [at the end of a doc](I may still want this feature on top but it would be more of a nice-to-have at that point 🙂) and the side menus move a lot with the height. I don't see how an animation would help there -- it'd still be quite distracting. Equal/max height may be the best way to go after all @julieg18 ?

julieg18:
> Well... I just tried using tabs [at the end of a doc](I may still want this feature on top but it would be more of a nice-to-have at that point 🙂) and the side menus move a lot with the tab changes (if you scroll down enough to see the page footer). I don't see how an animation would help there -- it'd still be quite distracting. Equal/max height may be the best way to go after all @julieg18 ?

Sounds good! We'll put our focus on adding a prop for equal tab height.

yathomasi:
I have tried setting the height of the tab to maximum content height, which introduces the possibility of having a large gap. 
Here is the demo video:
https://user-images.githubusercontent.com/20840228/150820948-b9dde230-d08c-409d-a63b-5ad871eaf7c5.mov

So, I think we need to introduce some maximum default height which of course will introduce a scrollbar. Will that be okay?
Here is the demo video:

[...]

PR #3257, opened 2022-02-04
added height prop to toggle tag

yathomasi:
This will be a simple addition to the current toggle tab on markdown.
We can set a height through a `height` prop on the `toggle` component `<toggle height="300px">` which will set height for all the tabs.

Ported from [pull cml#193](https://github.com/iterative/cml.dev/pull/193)

This is added to the `gatsby-plugin-iterative-docs` plugin.

Solves #3179


julieg18 (approved):
Looks good!

PR #3263, opened 2022-02-08
Add course link and "..." dropdown to nav bar

### LINKED ISSUE #2955: Educational Course Connection
With the new educational course coming out at the end of the quarter, we need to determine how people will access it from our websites.  

Each website will access the same school: Iterative Tools School.  

I would like for the school to be accessible  from the navigation and in DVC.org's case, from the Community Page.  

[Here is information on how their domains work](https://support.learnworlds.com/support/solutions/articles/12000080095-general-overview-custom-domains#What-is-a-custom-domain?). Let me know how you think it's best to proceed!

@julieg18 @rogermparent @yalozhkin

julieg18:
* refactors nav popup code
* adds `...` dropdown to nav bar (holds items that don't fit on the nav bar)
* add "Course" nav item

Fixes #2955

### DIFF
--- a/src/components/LayoutHeader/Nav/LinkItems/index.tsx
+++ b/src/components/LayoutHeader/Nav/LinkItems/index.tsx
@@ -1,33 +1,43 @@
-import React, { useRef, useState } from 'react'
+import React from 'react'
 import cn from 'classnames'
 
 import Link from 'gatsby-theme-iterative-docs/src/components/Link'
-import { OtherToolsPopup, CommunityPopup } from '../Popup'
+import {
+  OtherToolsPopup,
+  CommunityPopup,
+  OtherPopup,
+  IPopupProps
+} from '../Popup'
 
 import { ReactComponent as ArrowUpSVG } from '../../../../../static/img/arrow-up-icon.svg'
 import { ReactComponent as ArrowDownSVG } from '../../../../../static/img/arrow-down-icon.svg'
 
 import { logEvent } from 'gatsby-theme-iterative-docs/src/utils/front/plausible'
 import { getFirstPage } from 'gatsby-theme-iterative-docs/src/utils/shared/sidebar'
+import usePopup, { IUsePopupReturn } from '../../../../gatsby/hooks/usePopup'
 
 const docsPage = getFirstPage()
 
 import * as styles from './styles.module.css'
 
-type PopupName = 'CommunityPopup' | 'OtherToolsPopup'
+type PopupName = 'communityPopup' | 'otherToolsPopup' | 'otherPopup'
 
 interface INavLinkData {
   href: string
   eventType: string
   text: string
+  className?: string
 }
 
 interface INavLinkPopupData {
   text: string
-  popup: PopupName
+  popupName: PopupName
+  Popup: React.FC<IPopupProps>
+  className?: string
+  href?: string
 }
 
-const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
+export const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
   {
     href: '/features',
     eventType: 'features',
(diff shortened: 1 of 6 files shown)


src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -45,158 +55,91 @@ const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
...
-    if (isOtherToolsPopupOpen) {
-      closeAllPopups()
-    } else {
-      openOtherToolsPopup()
-    }
+): item is INavLinkPopupData =>
+  (item as INavLinkPopupData).popupName !== undefined
+
+const LinkItems: React.FC = () => {
+  const communityPopup = usePopup()
```
julieg18, line 91: I went ahead and created a `usePopup` hook that handles popup functionality. The popup code in the nav was already overly complex and adding a third popup would have made things even more complicated 😅

src/components/LayoutHeader/Nav/LinkItems/styles.module.css
```diff
@@ -66,6 +78,18 @@
     }
   }
 
+  &.other {
```
julieg18, line 81: Here's what the popup looks like: 

image


Is it good enough for a first iteration? cc @yalozhkin

src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -1,33 +1,43 @@
...
   text: string
-  popup: PopupName
+  popupName: PopupName
+  Popup: React.FC<IPopupProps>
+  className?: string
+  href?: string
 }
 
-const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
+export const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
```
julieg18, line 40: I wasn't sure where "Learn" should go in the nav. For now, I placed it after "Support" :)

rogermparent (approved):
Looks fine by me, especially as a first iteration before we try unifying all the navs.

Are we okay with "Community" losing its dropdown when in this menu? I don't see a problem with it, but others may think differently. That said, if we want to change it we can do so in a follow up, since this is still a step up from before where it wasn't present at all!

shcheklein:
Thanks, @julieg18 ! Can we move the Learn link to the Docs., Blog? (unless you discussed this other folks and this decision was made based of that). It's not a strong opinion, but it makes sense to me to keep them nearby.

julieg18:
> Thanks, @julieg18 ! Can we move the Learn link to the Docs., Blog? (unless you discussed this other folks and this decision was made based of that). It's not a strong opinion, but it makes sense to me to keep them nearby.

If I'm understanding your question correctly, you want to move it to after Blog? I moved it there but let me know if I misunderstood :)

> We already have Learn under the community dropdown, so we should consider changing either of these.

You're right! Plus, as we mentioned before, Learn is kind of an unclear name anyway. I renamed it to "Course" for now! 

> I think we also need to find a proper place on the hamburger menu as well.

Thanks for catching that! I'll update it.

> Q: should we consider adding a new window SVG icon similar to what we have with items on other tools menus?

I'm not sure I understand your question 🤔 Where exactly do you want to add a new window SVG icon?

yathomasi:
> I'm not sure I understand your question 🤔 Where exactly do you want to add a new window SVG icon?

screenshot-dvc-org-add-course-conn-nsrmv2 herokuapp com-2022 02 09-14_03_38

I meant something like this like we are using on `Other tools`. But, I am not sure we would want that.

julieg18:
> I meant something like this like we are using on Other tools. But, I am not sure we would want that.

Oh, ok! Thanks for clarifying further! Personally, I don't think it's needed. But @iterative/websites, what do you think?

shcheklein:
Yeah, good point @yathomasi , but I also don't think we need it in this specific case. I would consider adding the link only if it's a completely external website, in this case it's kinda internal link (ecosystem of the websites), so it's fine to avoid it here. And probably in general, menu is to tight on space that it makes sense to sacrifice some consistency here.

PR #3264, opened 2022-02-09
Search blogs indexing on Github actions CI

yathomasi:
This will create a trigger on Github Action. 
- It will be triggered when changes are made to the `content/blogs/**` folder on the `master` branch.
- Also, there is an attempt on adding caching on yarn, output folders i.e `.cache`, `public` and use Gatsby's incremental build flag during build time. This has significantly decreased build time.
- Added back partial indexing using git

NOTE: 
- currently `search-blogs-ci-indexing`(this branch) is used to trigger the push for testing purposes. It will be changed back to `master` if the testing goes alright.
- Also, it depends on Github Actions secrets for Algolia keys. It looks like I don't have access to add secrets. Once, I have access or someone adds those values, I will continue with further testings.

cc: @iterative/websites


.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,46 @@
+name: Indexing blogs
+on:
+  push:
+    branches: ['search-blogs-ci-indexing']
```
yathomasi, line 4: Currently, I am using this on `search-blogs-ci-indexing` for this draft.

.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,46 @@
...
+    paths:
+      - 'content/blog/**'
```
yathomasi, line 6: It will only get triggered changes on `content/blog/**`. I can there is `content/author` too. Should we consider that as well ?

.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          key: gatsby-public-folder
+          path: public
+      - name: Yarn cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
yathomasi, line 42: Caching yarn to GitHub actions for faster builds.

.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install Node.js and use yarn
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+      - name: create env file
+        run: |
+          touch .env
+          echo GATSBY_ALGOLIA_APP_ID=${{ secrets.GATSBY_ALGOLIA_APP_ID }} >> .env
+          echo ALGOLIA_ADMIN_KEY=${{ secrets.ALGOLIA_ADMIN_KEY }} >> .env
+          echo GATSBY_ALGOLIA_INDEX_NAME=${{ secrets.GATSBY_ALGOLIA_INDEX_NAME }} >> .env
```
yathomasi, line 22: Getting secrets for Algolia from Github Action Secret.

package.json
```diff
@@ -9,6 +9,7 @@
     "clean": "gatsby clean",
     "author": "LIMIT_BLOG_PAGES=1 SKIP_DOCS=true gatsby develop",
     "build": "gatsby build",
+    "build:incremental": "GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES=true gatsby build",
```
yathomasi, line 12: Used Gatsby's incremental build flag

gatsby-config.js
```diff
@@ -150,10 +150,15 @@ const plugins = [
...
-      skipIndexing: !(
-        process.env.GATSBY_ALGOLIA_APP_ID && process.env.ALGOLIA_ADMIN_KEY
-      ),
-      queries: require('./src/utils/algolia-queries.js')
+      skipIndexing:
+        process.env.CI &&
+        process.env.GATSBY_ALGOLIA_APP_ID &&
+        process.env.ALGOLIA_ADMIN_KEY
+          ? false
+          : true,
```
yathomasi, line 158: Now, it will run only on CI build.

package.json
shcheklein: Q: what does this key mean? why don't we use it in the regular builds?

src/gatsby/models/blog/onCreateMarkdownContentNode.js
```diff
@@ -16,7 +19,13 @@ async function createMarkdownBlogNode(api, { parentNode, createChildNode }) {
     pictureComment
   } = frontmatter
   const { name, relativePath } = parentNode
-
+  let gitDateTime
+  if (process.env.CI) {
+    const { stdout } = await exec(
+      `git log -1 --pretty=format:%aI ${fileAbsolutePath}`
+    )
+    gitDateTime = stdout
```
shcheklein, line 27: could please describe how exactly do we use this? do we still send all the records to algolia (including this date) or we compare it with something and do not send certain records?

package.json
shcheklein, line 12: Should we avoid building docs? images? (to make it faster)

shcheklein (commented):
Good start @yathomasi !

package.json
yathomasi: It looks like Gatsby has [this feature on by default](https://www.gatsbyjs.com/docs/reference/release-notes/v3.0/#incremental-builds-in-oss). So, caching `public` and `.cache` should be enough.

src/gatsby/models/blog/onCreateMarkdownContentNode.js
yathomasi: We are currently using [`gatsby-plugin-algolia`](https://github.com/algolia/gatsby-plugin-algolia/blob/master/gatsby-node.js) for indexing. 

If we have partial indexing enabled, then what it does is
- first gets the current indexed data from the Algolia with only the fields we specified (for eg: `modified`, `slug`)
- compares it with the freshly generated data and filters the new ones, ones whose fields have changed and removed ones.
- then it generates the data from the filtered and sends only those for indexing.

And for this particular `modified` field we are using the last commit timestamp for each post, which I assumed will remain the same until a new commit is made.

package.json
yathomasi: With cached folders `.cache` and `public`, it looks like it's kind of gets skipped i.e only get built if there are any changes. We can consider doing that as well.

.github/workflows/blogs-indexing.yaml
yathomasi: @shcheklein Can I get access to setting these secrets so that I can check/verify the data are getting indexed properly to Algolia?

.github/workflows/blogs-indexing.yaml
rogermparent: There's [a built-in caching feature for `setup-node`](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies) that takes care of proper implementation for this, though granted we haven't used it up until now.

package.json
rogermparent: This was a Gatsby 3 flag for a feature that became default in Gatsby 4, so it can be removed here.

rogermparent: It'd probably be best to not build images anyway so we don't store a bunch of never-used images in cache nor have to reprocess them all should it ever get cleared.

rogermparent (commented):
Looks cool so far! Hopefully testing goes well.

.github/workflows/blogs-indexing.yaml
yathomasi: Yeah, I had tried and it wasn't working. So, I changed to this way. But, I found later that the value wasn't set properly. So, yeah setting `env` should also work.

.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  build-and-indexing-blogs:
+    name: Build and indexing blogs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Node.js and use yarn
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
```
julieg18, line 16: Should we update this to `16.x` since that's the latest node engine that dvc.org can use?

julieg18 (commented):
Great job so far! Fingers crossed for when we test it 🤞

.github/workflows/blogs-indexing.yaml
yathomasi: Sure, we can use that.

yathomasi: Ended up using `node@14`. It looks like we have some issue with `node@16`. 
It seems like the issue is with `gatsby-plugin-manifest`. Need to look into detail. 

Screen Shot 2022-02-11 at 19 52 28

.github/workflows/blogs-indexing.yaml
```diff
@@ -0,0 +1,36 @@
...
+          cache: 'yarn'
+      - name: Gatsby Cache Folder
+        uses: actions/cache@v2
+        with:
+          key: gatsby-cache-folder
+          path: .cache
```
julieg18, line 27: Maybe I'm misreading the readme, but can't [actions/cache](https://github.com/actions/cache#whats-new) cache multiple directories? If it can, should we cache the `.cache` folder and `.public` folder in the same step?

julieg18 (approved):
Great job! I'm not seeing any issues, just had one minor question!

.github/workflows/blogs-indexing.yaml
yathomasi: Yeah, It looks like that could be done.

src/config/index.ts
```diff
@@ -1,9 +1,10 @@
 const config = {
   algolia: {
-    appId: String(process.env.GATSBY_ALGOLIA_APP_ID),
-    searchKey: String(process.env.GATSBY_ALGOLIA_SEARCH_KEY),
-    adminKey: process.env.ALGOLIA_ADMIN_KEY,
-    indexName: process.env.GATSBY_ALGOLIA_INDEX_NAME || 'dev_blogs'
+    appId: process.env.GATSBY_ALGOLIA_APP_ID || 'B87HVF62EF',
+    searchKey:
+      process.env.GATSBY_ALGOLIA_SEARCH_KEY ||
+      'cd45432547aa825db0ab63f4d90e13e3',
```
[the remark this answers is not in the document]
yathomasi: This refers to the application created by Ivan and also got the `open-source` which is basically the one we are going to use. I have also invited you to the app on iterative mail.
For now, as a testing purpose, we have used`ci_dvc_blogs` but later we will have it as `dvc_blogs` for production.
Also, these keys are safe to use on our frontend.

src/gatsby/models/blog/onCreateMarkdownContentNode.js
```diff
@@ -32,7 +41,8 @@ async function createMarkdownBlogNode(api, { parentNode, createChildNode }) {
     commentsUrl,
     picture,
     pictureComment: markdownToHtml(pictureComment),
-    sourcePath: relativePath
+    sourcePath: relativePath,
+    gitDateTime
```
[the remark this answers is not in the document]
yathomasi: Noted.

PR #3215, opened 2022-01-26
start: use tabs for a Windows command

mahnoor-shahid:
Received and error while removing the data file. 

Remove-Item : Parameter cannot be processed because the parameter name 'f' is ambiguous. Possible
matches include: -Filter -Force. 

I used Force command to make it work.

PR #3267, opened 2022-02-11
Only run merge-scheduler workflow if it's not a fork

julieg18:
* As shown by #3215, the merge-schedule workflow only runs for prs that aren't from forks and will fail if it runs on a forked pr.

### DIFF
--- a/.github/workflows/merge-schedule.yaml
+++ b/.github/workflows/merge-schedule.yaml
@@ -11,7 +11,7 @@ on:
 jobs:
   merge_schedule:
     runs-on: ubuntu-latest
-    if: ${{ github.repository == 'iterative/dvc.org' }}
+    if: ${{ github.repository !== 'iterative/dvc.org' }}
     steps:
       - uses: gr2m/merge-schedule-action@v1
         with:
--- a/.github/workflows/merge-schedule.yaml
+++ b/.github/workflows/merge-schedule.yaml
@@ -11,6 +11,7 @@ on:
 jobs:
   merge_schedule:
     runs-on: ubuntu-latest
+    if: github.repository == 'iterative/dvc.org'
     steps:
       - uses: gr2m/merge-schedule-action@v1
         with:
@@ -18,6 +19,6 @@ jobs:
           # rebase. Default is merge.
           merge_method: squash
           #  Time zone to use. Default is UTC.
-          time_zone: "America/Los_Angeles"
+          time_zone: 'America/Los_Angeles'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/merge-schedule.yaml
+++ b/.github/workflows/merge-schedule.yaml
@@ -11,7 +11,8 @@ on:
 jobs:
   merge_schedule:
     runs-on: ubuntu-latest
-    if: ${{ github.repository != 'iterative/dvc.org' }}
+    if:
+      { { github.event.pull_request.head.repo.full_name == github.repository } }
     steps:
       - uses: gr2m/merge-schedule-action@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-schedule.yaml
```diff
@@ -11,13 +11,15 @@ on:
 jobs:
   merge_schedule:
     runs-on: ubuntu-latest
+    if:
+      ${{ github.event.pull_request.head.repo.full_name == github.repository }}
```
julieg18, line 15: Checks if the repo name of the pull request equals the github repo. I tested it with a fork (#3270) and [it appears to work correctly](https://github.com/iterative/dvc.org/actions/runs/1826967213)!

rogermparent (approved):
Looks good! Thanks for the quick fix!

shcheklein:
Is it expected? Is it bc we were experimenting with the GH workflow? Ot does it create a new check / status every hour? 

Also, how does the merge work - who is the author of the merge? 

Also, do we know why fork didn't work?

Screen Shot 2022-02-11 at 10 19 08 PM

julieg18:
> Is it expected? Is it bc we were experimenting with the GH workflow? Ot does it create a new check / status every hour?

Yes, the multiple checks are expected. The check runs every time a pr is opened, edited, or receives new commits. 

> Also, how does the merge work - who is the author of the merge?

The action uses the github rest api to merge the pr. I believe! the action itself is the author It should look something like this:

image

> Also, do we know why fork didn't work?

Yes. The github action doesn't run on forks for security reasons.

shcheklein:
@julieg18 thanks!

> @shcheklein Yes, the multiple checks are expected. The check runs every time a pr is opened, edited, or receives new commits.

🤔 hmm, why don't we see the same for the link checker, unit tests, etc? (it's the first time I see the checks being duplicated this way). 

anyways, we can't keep it this way I think - it'll be exploding on some long-running PRs

> The action uses the github rest api to merge the pr. So I believe the action itself is the author! It should look something like this:

does it mean that anyone can come and create a PR and it'll be automatically merged (if they put a time to merge it in the description)?

> Yes. The github action doesn't run on forks for security reasons.

what error message do we see in this case? can it be changed - per repo or per specific action? and the most important - do we want this to be changed (somewhat related to the previous question I think)

julieg18:
> 🤔 hmm, why don't we see the same for the link checker, unit tests, etc? (it's the first time I see the checks being duplicated this way).

The same check can be duplicated when it is configured to run for different events. For example, our `link-check-deploy` runs for two different events and appears twice on a commit: 

Screen Shot 2022-02-14 at 4 23 37 PM

But most of our checks only need to run for one event and thus, aren't getting duplicated like `merge-scheduler`

> anyways, we can't keep it this way I think - it'll be exploding on some long-running PRs

Agreed! Maybe we can look for github action alternatives or update the configuration for this one. 

> does it mean that anyone can come and create a PR and it'll be automatically merged (if they put a time to merge it in the description)?

I believe that it can't be merged unless it has atleast one approving review. But I'll create a test right now to be sure.

> what error message do we see in this case? can it be changed - per repo or per specific action? and the most important - do we want this to be changed (somewhat related to the previous question I think)

Here's the error message:

image

I don't think it can be changed. What do you mean exactly by "do we want this to be changed"? Changed to what exactly?

shcheklein:
Sorry for the confusion. `can it be changed` == `can we change the behavior`, the way it behaves on forks.

julieg18:
> Sorry for the confusion. `can it be changed` == `can we change the behavior`, the way it behaves on forks.

Oh, ok! Thanks for the clarification! No, I don't think so. The merge-schedule action doesn't offer a way to change this behavior.

PR #3271, opened 2022-02-11
guide: note that `exp run` commits data

### LINKED ISSUE #2418: `exp run`: internal data committing performance implications
Per https://github.com/iterative/dvc/pull/5859 (released in [2.1.0](https://github.com/iterative/dvc/releases/tag/2.1.0))

UPDATE: Jump to https://github.com/iterative/dvc.org/issues/2418#issuecomment-854183969

> Data dependencies are now dvc commited internally when stashing an experiment so that any modifications to that data dep are preserved in both workspace and tempdir runs (previously the changes were dropped entirely by the exp run dvc checkout step).

Implications to consider documenting:
- [x] Mention the `git stash` "trick" somewhere
- [ ] Large changes to data dependencies in the workspace may slow down experiment queueing
- ~~Even if the run fails after this step, the new cached entries will remain, and metafiles will point to them (seen as Git changes)~~ Also true of `repro` which is already linked from `exp run` ref.

jorgeorpinel:
and other misc edits to Running Exps, and to the `gc` ref.

To close #2418

### DIFF
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -24,12 +24,11 @@ Provides a way to execute and track <abbr>experiments</abbr> in your
 directories, etc.
 
 `dvc exp run` has the same general behavior as `dvc repro` when it comes to
-`targets` and stage execution (restores the dependency graph, etc.). One
-important difference is that changed dependencies are committed to the <abbr>DVC
-cache</abbr> when the experiment begins execution, to guarantee its
-reproducibility later on.
+`targets` and stage execution (restores the dependency graph, etc.).
 
-> See the command [options](#options) for more on the differences.
+> This includes committing any changed data <abbr>dependencies</abbr> to the
+> <abbr>DVC cache</abbr> when preparing the experiment, which can take some
+> time. See the [Options](#options) section for the differences.
 
 Use the `--set-param` (`-S`) option as a shortcut to change
 <abbr>parameter</abbr> values [on-the-fly] before running the experiment.
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -23,9 +23,13 @@ Provides a way to execute and track <abbr>experiments</abbr> in your
 <abbr>project</abbr> without polluting it with unnecessary commits, branches,
 directories, etc.
 
-> `dvc exp run` has the same behavior as `dvc repro` when it comes to `targets`
-> and stage execution (restores the dependency graph, etc.). See the command
-> [options](#options) for more on the differences.
+`dvc exp run` has the same general behavior as `dvc repro` when it comes to
+`targets` and stage execution (restores the dependency graph, etc.). One
+important difference is that changed dependencies are committed to the <abbr>DVC
+cache</abbr> when the experiment begins execution, to guarantee its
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/exp/run.md
```diff
@@ -23,9 +23,13 @@ Provides a way to execute and track <abbr>experiments</abbr> in your
...
 directories, etc.
 
-> `dvc exp run` has the same behavior as `dvc repro` when it comes to `targets`
-> and stage execution (restores the dependency graph, etc.). See the command
-> [options](#options) for more on the differences.
+`dvc exp run` has the same general behavior as `dvc repro` when it comes to
+`targets` and stage execution (restores the dependency graph, etc.). One
+important difference is that changed dependencies are committed to the <abbr>DVC
+cache</abbr> when the experiment begins execution, to guarantee the
+reproducibility of any recorded experiment.
```
jorgeorpinel, line 30: One question for this PR's reviewer(s): #2418 was about performance implications. Should this be stressed further here or in the guide?

> For now I may just link to `gc` and check whether that ref. mentions `exp gc`(as vice versa).

content/docs/command-reference/gc.md
```diff
@@ -74,14 +73,14 @@ use `--remote` as well.
...
-  > \* Not including [DVC experiments](
+  > \* Not including [DVC experiments]
 
 [dvc experiments]: /doc/user-guide/experiment-management#experiments
 
 - `--all-experiments` keep cached objects referenced in all [DVC experiments],
   as well as in the workspace (implying `-w`). This preserves the project's
   [experimental](/doc/user-guide/experiment-management) data (including
-  checkpoints).
+  checkpoints). See also `dvc exp gc`.
```
jorgeorpinel, line 83: This is the only change to this file that is directly related to the task at hand. Sorry, just needed to copy edit the rest when I read it...

content/docs/command-reference/exp/run.md
pmrowla: I think just noting that it is committed is fine

dberenbaum: See my comment in https://github.com/iterative/dvc.org/issues/2418#issuecomment-1036405900. `dvc repro` also commits changed dependencies to the DVC cache, so I don't see a difference here.

I'm having flashbacks to our original discussion about this 😅 . I think this initially came up because `dvc exp run` didn't used to commit changed dependencies, but we changed it to behave more like `dvc repro`.

jorgeorpinel: > this initially came up because dvc exp run didn't used to commit changed dependencies, but we changed it to behave more like dvc repro 

Right I just went all the way back to https://github.com/iterative/dvc/issues/5593, https://github.com/iterative/dvc/pull/5859, and #2418 and it seems that's what the change was. So this paragraph about being different to `repro` is incorrect... ⌛

jorgeorpinel: OK I updated the changes...

> I think just noting that it is committed is fine

Yeah that's what I did now. I didn't specifically mention it in the context of queueing, just left generic language that applies to running or queueing ("when preparing the experiment").

dberenbaum (approved):

PR #3283, opened 2022-02-15
Add clarification to repro.md

alex000kim:
Adding clarification about `dvc repro` behaviour for a stage to be considered "always changed"


content/docs/command-reference/repro.md
```diff
@@ -51,8 +51,8 @@ There are a few ways to restrict what will be regenerated by this command: by
 specifying specific reproduction [`targets`](#options), or by using certain
 command [options](#options), such as `--single-item` or `--all-pipelines`.
 
-> Note that stages without dependencies are considered _always changed_, so
-> `dvc repro` always executes them.
+> Note that stages without dependencies or outputs are considered _always
+> changed_, so `dvc repro` always executes them.
```
daavoo, line 55: Maybe link to https://dvc.org/doc/command-reference/stage/add#--always-changed?

daavoo (approved):

content/docs/command-reference/repro.md
alex000kim: link added

content/docs/command-reference/repro.md
```diff
@@ -51,7 +51,8 @@ There are a few ways to restrict what will be regenerated by this command: by
 specifying specific reproduction [`targets`](#options), or by using certain
 command [options](#options), such as `--single-item` or `--all-pipelines`.
 
-> Note that stages without dependencies are considered _always changed_, so
+> Note that stages without dependencies or outputs are considered
```
daavoo, line 54:
```suggestion
> Note that stages without dependencies are considered
```

daavoo:
My bad for lack of clarification, the outputs is only an issue on the first run (i.e. before dvc had tracked the output)

content/docs/command-reference/repro.md
```diff
@@ -51,7 +51,8 @@ There are a few ways to restrict what will be regenerated by this command: by
 specifying specific reproduction [`targets`](#options), or by using certain
 command [options](#options), such as `--single-item` or `--all-pipelines`.
 
-> Note that stages without dependencies are considered _always changed_, so
+> Note that stages without dependencies are considered
+> [always changed](/doc/command-reference/stage/add#--always-changed), so
 > `dvc repro` always executes them.
```
jorgeorpinel, line 56: Hmmm Idk. `stage add --always-changed` adds an `always-changed` field to the stage in dvc.yaml . That's not needed for stages without dependencies to be considered always changed. This link may end up causing confusion. What was the motivation for this change? Maybe I can help clarify the text in another way.

jorgeorpinel:
From what I see originally this intended to add "or outputs" but that's not actually the case so only a link was added, which I think adds confusion. So I'll revert this I think.

content/docs/command-reference/repro.md
jorgeorpinel: Maybe linking to https://dvc.org/doc/command-reference/status#local-workspace-status could help more in clarifying... 🤔

PR #3286, opened 2022-02-15
Repro docs clarification

daavoo:
Ground truth here: 

https://github.com/iterative/dvc/blob/d1f7cf45405119d8663a146b9493cf4b69e544f3/dvc/stage/__init__.py#L236


alex000kim (approved):
LGTM

content/docs/command-reference/repro.md
```diff
@@ -51,7 +51,7 @@ There are a few ways to restrict what will be regenerated by this command: by
 specifying specific reproduction [`targets`](#options), or by using certain
 command [options](#options), such as `--single-item` or `--all-pipelines`.
 
-> Note that stages without dependencies are considered
+> Note that stages without dependencies or outputs are considered
```
jorgeorpinel, line 54: Does this mean without dependencies (x)or without outputs? Or does it mean both without deps nor outs?

daavoo: It has to be both, but I wasn't really confident about my English and whether `nor` was the correct way to say. 

The code is:

```
not any((self.deps, self.outs))
```

jorgeorpinel: "neither x nor y" would be a proper way to say it in writing. In this case since we already with "without" maybe  we just need to change change or -> nor but I'm not sure it's clear/emphatic enough. Alternatives:

"stages without dependencies and without outputs"

"stages without either dependencies or outputs"

"stages that have neither dependencies nor outputs"

"stages that don't have dependencies nor outputs"

jorgeorpinel: For now I went with "stages without dependencies nor outputs" in the `run` ref. and with "has neither dependencies nor outputs" in the `status` ref. (see https://github.com/iterative/dvc.org/pull/3313/files).

PR #3289, opened 2022-02-16
start: about .dvc file tracking

jorgeorpinel:
This is per user feedback. Full context: https://discord.com/channels/485586884165107732/563406153334128681/943192209958772737

> In my opinion, it is not at all clear from the documentation how DVC tracks data. I was under the impression that all files tracked by DVC us a .dvc file. Do I have to manually use dvc add to track outputs of stages if I want to restore those outputs on a different machine without having to run dvc repro?

UPDATE: More user feedback in this thread - https://discord.com/channels/485586884165107732/943360450698043463


content/docs/start/data-and-model-versioning.md
```diff
@@ -19,8 +19,6 @@ or watch our video to learn about versioning data with DVC!
 
 https://youtu.be/kLKBcPonMYw
 
-To start tracking a file or directory, use `dvc add`:
-
 <details>
```
jorgeorpinel, line 22: I also moved this to after the details block so it's not interrupted by it.

iesahin (approved):
Clarification and link to pipeline files seems good. Thank you Jorge.

iesahin:
Is this still draft? @jorgeorpinel

jorgeorpinel:
Looks good

image

content/docs/start/data-and-model-versioning.md
```diff
@@ -86,6 +85,12 @@ outs:
 
 </details>
 
+> Note that there are [other mechanisms] to track data files/dirs besides `.dvc`
```
shcheklein, line 88: my 2cs - it expands the start page (we should be optimizing for its size) and I'm not sure it solves the problem. It's not informative enough. 

The expectation is that at this stage people are not familiar with pipelines yet. This reference would confuse at best (how important it is to know? should I jump and start reading the second doc?)

I would clarify how did that happen that user was familiar with pipelines but didn't know about them tracking files automatically? May be we should change the pipelines docs instead.

If we want to keep this note, I would make more friendly, less obscure:

Note, when you start using pipelines `dvc add` is needed only for pipeline inputs, usually data. Everything else is handled ...

jorgeorpinel: While this only added one line so I think the size impact isn't that important, now that I read it a week later I do agree that it doesn't necessarily clarify anything. I'll revert and check whether the pipelines page needs the clarification instead... ⌛

PR #3303, opened 2022-02-22
install: drop `snap install --beta/edge` (Linux)

jorgeorpinel:
For https://github.com/iterative/dvc.org/issues/3302


iesahin (approved):
Removes `--edge` and `--beta` from `snap install` options. Thanks Jorge.

PR #3307, opened 2022-02-22
fix links plus a couple other things

jendefig:
A couple more bad links and corrected some things that should have been there the first time.  Sorry.  🤦‍♀️ My attempt to speed up my process on the front end, failed on the backend.

### DIFF
--- a/content/blog/2022-2-17-february-22-heartbeat.md
+++ b/content/blog/2022-2-17-february-22-heartbeat.md
@@ -146,11 +146,12 @@ _Discovering temporal scientometric knowledge in COVID-19 scholarly production
 
 - If you're a [**Guild.Ai**](https://guild.ai/) user, you'll be happy to know
   that Guild now supports DVC! Find out more in
-  [this article](https://my.guild.ai/t/using-guild-ai-with-dvc/803) and the
+  [this article](https://my.guild.ai/t/using-guild-ai-with-dvc/803) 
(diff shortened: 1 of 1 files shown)


jendefig:
@rogermparent So I guess the link checker has been turned off?  I missed a ton of bad links this month. Sorry I didn't realize it was being skipped, I should have noticed that.  Is this on purpose?

rogermparent (approved):

rogermparent:
I'm not sure what exactly happened on the original PR since it's possible it could've run late or something, but looking at it I can see [a failed link checker report](https://github.com/iterative/dvc.org/runs/5233845167) attached to it

iesahin (approved):
Thank you @jendefig

jendefig:
Thanks @rogermparent Missed that the first time. Those are the ones, except Medium. But that wouldn't have picked up the [Raymond Gasper]() which just sent back to the same blog post.  Is this how it should be?

rogermparent:
It not picking up the empty link isn't a deliberate decision, but implicit on the link checker being designed to look for whole links and check for ones with abnormal responses. A lot of edge cases like empty links weren't taken into consideration, but it's something we should probably think about in the planned link check overhaul.

jendefig:
Thanks @rogermparent!  It was totally my bad to miss putting them in. When I write, to go back and forth collecting links just slows me down, so I tried just typing [name]() and kept going to fill in links later, but missed those.  A simple cmd + f - () would find them for me though.  Just need to add to my process.

PR #3310, opened 2022-02-22
`ref`: Duplicated paging info to `exp show` 

### LINKED ISSUE #3217: ref: Move pagination (less/more) section to elsewhere from `dvc dag`
No description provided.

iesahin:
It was referencing `dvc dag` for paging info. Now it's self contained and added an example to use `cat` as a pager. 

Closes #3217

### DIFF
--- a/content/docs/command-reference/exp/show.md
+++ b/content/docs/command-reference/exp/show.md
@@ -44,10 +44,9 @@ $ dvc exp show
  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ```
 
-Your terminal will enter a
-[paginated screen](/doc/command-reference/dag#paginating-the-output) by default,
-which you can typically exit by typing `Q`. Use `--no-pager` to print the table
-to standard output.
+Your terminal will enter a [paginated screen](#paginating-the-output) by
+default, which you can typically exit by typing `q`. Use `--no-pager` to print
+the table to standard output.
 
 By default, the printed experiments table will include columns for all metrics,
 parameters and dependencies from the entire project. The `--only-changed`,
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/exp/show.md
```diff
@@ -65,6 +64,42 @@ will be generated using the same data from the table.
...
+
+This command's output is automatically piped to
+[less](<https://en.wikipedia.org/wiki/Less_(Unix)>) if available in the terminal
+(the exact command used is `less --chop-long-lines --clear-screen`). If `less`
+is not available (e.g. on Windows), the output is simply printed out.
+
+> It's also possible to
+> [enable `less` on Windows](/doc/user-guide/running-dvc-on-windows#enabling-paging-with-less).
+
+> Note that this also applies to `dvc dag`.
```
jorgeorpinel, line 77: I think we don't need this link or the link from dag to here since we're duplicating the content.

content/docs/command-reference/exp/show.md
```diff
@@ -65,6 +64,42 @@ will be generated using the same data from the table.
...
+```dvc
+$ export DVC_PAGER=more
+```
+
+If you'd like to print to stdout always without paging, you can set this
+variable to `cat`:
+
+```dvc
+$ export DVC_PAGER=cat
+```
```
jorgeorpinel, line 101: Let's not further expand on Linux methods? There are others but this doc should be about DVC IMO.

Also, this isn't in https://dvc.org/doc/command-reference/dag#providing-a-custom-pager ATM.

iesahin: Whole environment variable thing is about Linux. The method provides a way to skip the `--no-pager` option everytime. So I think it's a useful tip.

jorgeorpinel: Hey sorry for the delay...

We can improve the env var blocks later to include Windows Cmd (Windows also supports the same env vars I think).

This is a cmd ref, we want it to be more focused/shorter/self contained I think. Again, this note is a Linux tip.

iesahin: removed

content/docs/command-reference/exp/show.md
iesahin: removed

jorgeorpinel (approved):

content/docs/command-reference/exp/show.md
jorgeorpinel: > We can improve the env var blocks later to include Windows Cmd

I'm making it clearer that these are Linux examples instead for now. See https://github.com/iterative/dvc.org/pull/3356

Issue #1762, opened 2020-09-06
ref: clarify about obfuscated remote file structure (same as cache)


antonkulaga:
From my experience of moving the team to DVC there were two major understanding problems with understanding of key DVC concepts:
* when you do a gdrive and other dvc remotes (with exception of dvc import) they have folders/files in a dvc-specific format, you cannot just put regular input files to gdrive and say they are remote (unless you are using dvc import)
* ```dvc run``` is more about creating stages than actually running, if one just wants to run she should use ```dvc repro``` instead

Despite reading dvc tutorials and docs many teammates and students were still confused by this. Maybe there are ways to emphasize it somehow so people who read not attentively will not confuse?

efiop:
Transferred to the docs repo.

jorgeorpinel:
Hei @antonkulaga thanks for the feedback

> dvc remotes... have folders/files in a dvc-specific format, you cannot just put regular input files to gdrive and say they are remote (unless you are using dvc import)

This is something we can definitely address in docs, I think. Probably around sync commands (push/pull/etc) and remote refs. and maybe mention that remotes use the same https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory (and link there from the aforementioned pages).

We'll keep this issue open to be addressed when possible, but feel free to [contribute](https://dvc.org/doc/user-guide/contributing/docs), that would be amazing! (Just a draft PR is enough, we can polish the writing, add links, etc.)

> dvc run is more about creating stages than actually running

Yep, it used to be for running before 1.0 but it's transitioned into a helper to define stages. We're considering renaming it to something like `dvc stage add` I think (cc @efiop that's more of a discussion for the core repo I think) and leaving the running for `repro` exclusively.

For now in docs we already emphasize that it's a helper command so not actionable feedback for docs at the moment, I think.

sarthakforwet:
@jorgeorpinel In `push.md`, we have the sentence "dvc push uploads data from the cache to remote storage."
Doesn't this sentence provide the argument that remote storage would have data in a dvc-specific format?

Also, can you please mention some references for me to learn about this issue in a better way? Thanks!

jorgeorpinel:
Hey @sarthakforwet all the info about the cache dir structure is in the link I shared above (https://dvc.org/doc/user-guide/dvc-files-and-directories#structure-of-the-cache-directory).

> dvc push uploads data from the cache to remote storage

[...]

PR #3315, opened 2022-02-22
Added document on remote cache structure #1762

michael-mayo:
Hi @shcheklein @jorgeorpinel,

I've made a short very introductory document on how the remote cache is structured, based on my understanding after a couple of days. It's currently in the user-guide folder but not sure if that's the right place. Happy for any feedback. Would also like to make second document concerning how versioning works as I have not covered that here.

Mike


content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+
+```dvc
+$ mkdir project
+$ cd project
+$ git init
+$ dvc init
+```
+
+So far, I have not connected `git` to any external repository. 
+You can do that easily enough later on, but we won't worry about that here.
```
jorgeorpinel, line 33: Not sure what this referred to. Connect git to an external repo?

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+data1 data2 data3 data4
+data1.dvc data2.dvc data3.dvc data4.dvc
+```
+
+The `dvc add` command should also have put all data files inside `.gitignore`, which is a good idea because if these four identical data files were in reality very big,
+it could take a long time to upload them to github and we want to avoid that.
+
+Let's check the `.gitignore` file then:
+
+```dvc
```
jorgeorpinel, line 71: Starts to look a bit like a `dvc add` explanation here.

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+Thanksfully, the data files are all there!
+
+Now we can do an initial commit of everything else using `git`,
+but first, let's set up the connection between `dvc` and our remote S3 cache first. 
+
+When the remote cache is added, the file `.dvc/cache` will be updated with the remote cache details, so we will need to add these to git as well:
+
+```dvc
```
jorgeorpinel, line 86: Here it looks like a remote storage explanation. Maybe there's a bit of confusion between local cache and remote storage (granted, they're very similar so it can be confusing).

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+$ dvc remote add s3cache s3://path/to/your/cache/
+$ git add .dvc/config
+```
+
+And finally let's finish with a initial commit:
+
+```dvc
+$ git commit -m "initial"
```
jorgeorpinel, line 94: I'd try to avoid or compress all these satellite operations in order to focus on the next section (actually about the cache structure).

content/docs/user-guide/structure-of-remote-cache-by-example.md
jorgeorpinel: OK I see this is actually about remote storage so the confusion is just in the terminology. We try to avoid "remote cache" even when you could call it that, to prevent this confusion between local and remote data storage. Just a note

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+
+If you managed to get this far, we can now start talking about how the remote cache is structured in relation to our project files.
+
+## Pushing the DVC-tracked files to the remote cache
+
+Now come's the interesting part!
+
+> To fully appreciate how remote caches work, it's useful
+> to understand the concept of _indirection_. Indirection refers
+> to the naming of things using a reference.
```
jorgeorpinel, line 105: Interesting 👍🏼 

A [link here](https://en.wikipedia.org/wiki/Indirection) would be useful too.

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+
+Notice the output of this command: `1 file pushed`.
+If you were following through the previous section,
+you should know that _four_ data files were created.
+You can double check this if you like:
+
+```dvc
+% ls
+data1		data2		data3		data4		
+data1.dvc	data2.dvc	data3.dvc	data4.dvc
```
jorgeorpinel, line 126: Nice example. Seems more about storage optimization (caching mechanism) than on structure of the cache/remote per se -- just a naming question for the title I guess

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+
+![Data in project and remote cache](../../../static/img/project_remote_1.png)
+
+## Adding new data and how the remote cache changes
+
+Let's add two more data files to our repository, but this time make them  different to the first four data files. We should also make sure DVC tracks the new files and add the corresponding `.dvc` files with git: 
+
+```dvc
+$ echo data3636363 > data5
+$ echo blahblahblah > data6
```
jorgeorpinel, line 183: Minor: I think it looks better to use more meaningful samples even if the data contents are irrelevant for the story.

content/docs/user-guide/structure-of-remote-cache-by-example.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+Graphically speaking, the setup will now look like this:
+
+![Data in project and remote cache after adding more data](../../../static/img/project_remote_2.png)
+
+
+## Modifying existing data 
+
+Next, let's change one of the data files that used to be a duplicate and push the changes to the remote cache:
```
jorgeorpinel, line 220: Minor: I would've personally combined this with the previous section since you can add and modify in the same step. But that's me (I tend to summarize as much as possible).

jorgeorpinel:
@michael-mayo I just left feedback since you mentioned it was welcome. Not expecting any follow-ups since this is closed.

In general looks like a longer-form blog format so if that was the intention 👍🏼 

For docs, even guides (our longer format) we do try to summarize a bit more. For example in this case we have a short section about the cache structure in https://dvc.org/doc/user-guide/project-structure/internal-files#structure-of-the-cache-directory and then from pages about remote storage we just link there 🙂  That said some of the ways you explain this are nice and could definitely be valuable additions.

michael-mayo:
Thanks Jorge, I'm current working some more on these so I will generate another PR soon. Might move them to the blogs directory instead. The feedback was good, thanks.

content/docs/user-guide/structure-of-remote-cache-by-example.md
michael-mayo: Resolved. This was removed in a new draft.

content/docs/user-guide/structure-of-remote-cache-by-example.md
michael-mayo: Good point. I've changed all occurrences of "remote cache" to "remote storage".

content/docs/user-guide/structure-of-remote-cache-by-example.md
michael-mayo: Added the link and a better explanation.

content/docs/user-guide/structure-of-remote-cache-by-example.md
michael-mayo: Yes that's a good point. Have changed to the title to "Remote storage optimization example" for now.

PR #3317, opened 2022-02-23
Improve/Reusable components for dropdown

yathomasi:
I ended up updating the @julieg18's implementation.
Made some components more reusable and used them over existing dropdowns.

cc: @iterative/websites


src/gatsby/hooks/useOnEscape.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
```
julieg18, line 1: Great idea putting the escape logic into a hook! It makes the popup hook a lot simpler and we are able to reuse this logic, if needed, elsewhere!

src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -102,7 +102,13 @@ const LinkItems: React.FC = () => {
       {navLinkItemsData.map((item, i) => {
         const popup = isPopup(item) ? popups[item.popupName] : undefined
         return (
-          <li key={i} className={styles.linkItem} ref={popup?.containerEl}>
+          <li
+            key={i}
+            className={styles.linkItem}
+            ref={popup?.containerEl}
+            onMouseEnter={popup?.open}
```
julieg18, line 109: Have either Ivan or Yaroslav agreed on changing the dropdowns to activate on hover? Not a blocker since this is a minor change and we can always change it back later :)

src/gatsby/hooks/useOnEscape.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const useOnEscape = (handler: (event: Event) => void, cleanUp = true) => {
+  useEffect(() => {
+    const listener = (event: KeyboardEvent) => {
+      // check if key is an Escape
```
julieg18, line 6: Is this comment needed? I feel like the if statement speaks for itself 🤔

julieg18 (commented):
Looks great!

src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -75,7 +75,7 @@ export const navLinkItemsData: Array<INavLinkData | INavLinkPopupData> = [
     Popup: OtherToolsPopup
   },
   {
-    text: '...',
+    text: '···',
```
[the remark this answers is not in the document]
yathomasi: Screen Shot 2022-02-23 at 19 25 14

Actually, the dropdown on ellipsis was lower than usual which was due to margin-bottom, so in order to remove that and center the ellipsis, I ended up using `·`. The better one would be to use `SVG`.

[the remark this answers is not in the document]
julieg18: Makes sense! I don't see any issue with using SVG but if we do use it, we will need to add an `aria-label` to the button, letting screen readers know what the button does.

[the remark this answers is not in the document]
yathomasi: It's set on SVG itself. It should be working. If not we can add to the button itself.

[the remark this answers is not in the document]
julieg18: Yes, adding to the button is preferable to the SVG :)

rogermparent (approved):
Looking good to me as well!

julieg18 (approved):

src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -102,18 +106,25 @@ const LinkItems: React.FC = () => {
       {navLinkItemsData.map((item, i) => {
         const popup = isPopup(item) ? popups[item.popupName] : undefined
         return (
-          <li key={i} className={styles.linkItem} ref={popup?.containerEl}>
+          <li
+            key={i}
+            className={styles.linkItem}
+            ref={popup?.containerEl}
+            onMouseEnter={isMobile ? undefined : popup?.open}
```
julieg18, line 113: Instead of adding a package to detect mobile users, I wonder if we should update our popup logic to open on `:focus-within` instead of toggling `button.onClick` 🤔. Then mobile users wouldn't have to double tap... but this code works just fine and we can always update it later :)

julieg18 (approved):
LGTM

src/components/LayoutHeader/Nav/LinkItems/index.tsx
```diff
@@ -109,8 +110,8 @@ const LinkItems: React.FC = () => {
             key={i}
             className={styles.linkItem}
             ref={popup?.containerEl}
-            onMouseEnter={popup?.open}
-            onMouseLeave={popup?.close}
+            onMouseEnter={isMobile ? undefined : popup?.open}
+            onMouseLeave={isMobile ? undefined : popup?.close}
```
rogermparent, line 114: While I'm not incredibly enthused by having to do a JS mobile check just for this, but I see why it's needed with how `mouseEnter` and `click` trigger at the same time on touch.

I think maybe a small debounce on all popover changes would slightly better handle edge-cases such as mobile devices (or other devices detected as mobile) with mice or desktop devices with touch, but that's a minor enough case that I think we can let this through.

rogermparent (approved):

yathomasi:
@julieg18 @rogermparent 
Tried a bunch of ways and finally found out the` pointer up` event fits perfectly in our case. The order of `mouse enter` and `click` was such that it was closing the dropdown. Now, it seems to be working in order. If this is okay, then we are ready to merge.

rogermparent:
> If this is okay, then we are ready to merge.

Just tried it out, and everything looks good to me! Hopefully we don't run into browser support issues, but the [caniuse entry for pointer events](https://caniuse.com/pointer) looks pretty good.

julieg18 (approved):
Great idea!

Issue #2960, opened 2021-10-24
guide: concerns with "Persisting" experiments


jorgeorpinel:
> Follow-up to https://github.com/iterative/dvc.org/pull/2845

### Term/Concept of "persisting"
_Per https://github.com/iterative/dvc.org/pull/2845#issuecomment-937151219_

First, "persisting" usually means to continue something despite it being difficult. We want it to mean to make something (that was ephemeral) persistent. So not sure it's a good title already.

One option is to be explicit, avoid term "persist", and say "commit experiments to Git". This is a probably good first step regardless.

Then we should consider all the reasons why people want to do commit experiments and generalize them into a term. Currently we only mention sharing experiments, but then this whole content should be inside https://dvc.org/doc/user-guide/experiment-management/sharing-experiments instead (which I doubt is right but maybe?)

- [ ] Related: improve motivation in the intro of https://dvc.org/doc/user-guide/experiment-management/persisting-experiments. Probably reuse https://dvc.org/doc/user-guide/experiment-management#persistent-experiments (and remove that from the index or briefly mention somewhere and link).
  _From https://github.com/iterative/dvc.org/pull/2845#issuecomment-939426781_

### Other

- [ ] Don't call the version where experiments are run (in examples) `experiments`. It can be confusing/misleading
  _From https://github.com/iterative/dvc.org/pull/2845#discussion_r726818325_
- General copy edits if needed (e.g. https://github.com/iterative/dvc.org/pull/2845#pullrequestreview-773103226)

iesahin:
IMHO "commit" is too generic and overused.

Users don't only commit to share, they bring experiments to the standard Git workflow. 

@dberenbaum 's "retrieving" is meaningful in this context, we retrieve them from DVC to Git. 

How about "consolidating experiments"? 😄

jorgeorpinel:
I wouldn't call "commit" generic. It's relates specifically to version control. And it's a pretty clear way to express what we're trying to say: We sell DVC Experiments as something that doesn't pollute your Git history (no commits) so it makes sense to say how to actually commit some if needed.

> they bring experiments to the standard Git workflow

Currently we only mention sharing experiments, we should probably emphasize git flow in general more then.

> How about "consolidating experiments"? 😄

Vague. My vote sticks with the very explicit and correct "commit" unless there's a much better alternative.

jorgeorpinel:
This discussion has become overcomplicated and too Git-focused. We're off topic.

[...]

PR #3319, opened 2022-02-24
Minor fixes

omesser:
Minor stuff I hit while glossing over some docs...
Some other thing that bothered me a bit (made me stop and re-read a few times) is the compound noun "pipeline_**s**_ file(s)" instead of "pipeline file(s)" (pipelines plural), but I ... guess it's valid...? 🤔

### DIFF
--- a/content/docs/user-guide/experiment-management/checkpoints.md
+++ b/content/docs/user-guide/experiment-management/checkpoints.md
@@ -32,8 +32,8 @@ similar to a branch.
 
 </details>
 
-Like with regular experiments, checkpoints can be persisted by
-[committing them to Git](#committing-checkpoints-to-git).
+Like with regular experiments, checkpoints can be
+[committing to Git](#committing-checkpoints-to-git).
 
 This guide covers how to implement checkpoints in an ML project using DVC. We're
 going to train a model to identify handwritten digits based on the MNIST
--- a/content/docs/user-guide/experiment-management/checkpoints.md
+++ b/content/docs/user-guide/experiment-management/checkpoints.md
@@ -32,7 +32,7 @@ similar to a branch.
 
 </details>
 
-Like with regular experiments, checkpoints can become persistent by
+Like with regular experiments, checkpoints can be persisted by
 [committing them to Git](#committing-checkpoints-to-git).
 
 This guide covers how to implement checkpoints in an ML project using DVC. We're
--- a/content/docs/user-guide/experiment-management/checkpoints.md
+++ b/content/docs/user-guide/experiment-management/checkpoints.md
@@ -33,7 +33,7 @@ similar to a branch.
 </details>
 
 Like with regular experiments, checkpoints can be
-[committing to Git](#committing-checkpoints-to-git).
+[committed to Git](#committing-checkpoints-to-git).
 
 This guide covers how to implement checkpoints in an ML project using DVC. We're
 going to train a model to identify handwritten digits based on the MNIST
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -8,7 +8,7 @@ details.
 > experimentation, you may want to check the basics in
 > [Get Started: Experiments](/doc/start/experiments/) first.
 
-## Pipeline files
+## Pipelines files
 
 DVC relies on `dvc.yaml` files that contain the commands to run the
 experiment(s). These files codify _pipelines_ that specify one or more
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -61,10 +61,10 @@ and reproduces them.
 
 > 📖 See `dvc params` for more details.
 
-You could manually edit a params file and run an experiment using those as inputs. Since
-this is a common sequence, the built-in option `dvc exp run --set-param` (`-S`)
-is provided as a shortcut. It takes an existing param name and value, and
-updates the file on-the-fly before execution.
+You could manually edit a params file and run an experiment using those as
+inputs. Since this is a common sequence, the built-in option
+`dvc exp run --set-param` (`-S`) is provided as a shortcut. It takes an existing
+param name and value, and updates the file on-the-fly before execution.
 
 ```dvc
 $ cat params.yaml
@@ -193,9 +193,9 @@ outputs don't remain in the workspace.
 
 Subsequent uses of `dvc exp run` will resume from the latest checkpoint (using
 the latest cached versions of all outputs). To resume from a previous checkpoint
-(list them with `dvc exp show`), you must first `dvc exp apply` it before running
-it with `dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to specify the
-checkpoint to resume from.
+(list them with `dvc exp show`), you must first `dvc exp apply` it before
+running it with `dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to
+specify the checkpoint to resume from.
 
 Alternatively, use `--reset` to start over (discards previous checkpoints and
 their outputs). This is useful for re-training ML models, for example.
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -193,9 +193,9 @@ outputs don't remain in the workspace.
 
 Subsequent uses of `dvc exp run` will resume from the latest checkpoint (using
 the latest cached versions of all outputs). To resume from a previous checkpoint
-(list them with `dvc exp show`), you must first `dvc exp apply` it before
-running it with `dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to
-specify the checkpoint to resume from.
+(list them with `dvc exp show`), you must first `dvc exp apply` it before using
+`dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to specify the
+checkpoint to resume from.
 
 Alternatively, use `--reset` to start over (discards previous checkpoints and
 their outputs). This is useful for re-training ML models, for example.
--- a/content/docs/user-guide/experiment-management/running-experiments.md
+++ b/content/docs/user-guide/experiment-management/running-experiments.md
@@ -8,7 +8,7 @@ details.
 > experimentation, you may want to check the basics in
 > [Get Started: Experiments](/doc/start/experiments/) first.
 
-## Pipelines files
+## Pipeline files
 
 DVC relies on `dvc.yaml` files that contain the commands to run the
 experiment(s). These files codify _pipelines_ that specify one or more
@@ -20,7 +20,7 @@ experiment(s). These files codify _pipelines_ that specify one or more
 
 ### Running the pipeline(s)
 
-You can run the experimental pipeline using `dvc exp run`. It uses `./dvc.yaml`
+You can run the experiment pipeline using `dvc exp run`. It uses `./dvc.yaml`
 (in the current directory) by default.
 
 ```dvc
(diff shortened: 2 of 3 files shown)


content/docs/user-guide/experiment-management/checkpoints.md
```diff
@@ -32,7 +32,7 @@ similar to a branch.
 
 </details>
 
-Like with regular experiments, checkpoints can become persistent by
+Like with regular experiments, checkpoints can be persisted by
 [committing them to Git](#committing-checkpoints-to-git).
```
jorgeorpinel, line 36: Yeah we discussed this term and at some point decided not to use it as a verb since it's not really a the actual meaning of "to persist" in general language.

We do still have this problem in a couple places: in the title of https://dvc.org/doc/user-guide/experiment-management/persisting-experiments and in a hidden section under https://dvc.org/doc/user-guide/experiment-management/running-experiments#the-experiments-queue.

Full discussion: https://github.com/iterative/dvc.org/issues/2960

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -20,7 +20,7 @@ experiment(s). These files codify _pipelines_ that specify one or more
 
 ### Running the pipeline(s)
 
-You can run the experimental pipeline using `dvc exp run`. It uses `./dvc.yaml`
+You can run the experiment pipeline using `dvc exp run`. It uses `./dvc.yaml`
```
jorgeorpinel, line 23:
```suggestion
You can run the pipeline of the experiment with `dvc exp run`. It uses `./dvc.yaml`
```

Relates to a point in https://github.com/iterative/dvc.org/issues/2768:

> removing the pipeline and stages concept would be ideal

jorgeorpinel: For now, I'm not sure what's wrong with "experimental pipeline" though. Isn't that what it is?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -192,11 +191,11 @@ gets interrupted (e.g. with `Ctrl+C`, or by an error). Without interruption, a
 "wrap-up" checkpoint will be added (if needed), so that changes to pipeline
 outputs don't remain in the workspace.
 
-Subsequent uses of `dvc exp run` will continue from the latest checkpoint (using
+Subsequent uses of `dvc exp run` will resume from the latest checkpoint (using
```
jorgeorpinel, line 194: 👍🏼 

Should we also change  continue -> resume  in [here](https://dvc.org/doc/command-reference/exp/run#-r)?

content/docs/user-guide/experiment-management/running-experiments.md
```diff
@@ -192,11 +191,11 @@ gets interrupted (e.g. with `Ctrl+C`, or by an error). Without interruption, a
...
 the latest cached versions of all outputs). To resume from a previous checkpoint
-(list them with `dvc exp show`), you must first `dvc exp apply` it before the
-`dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to specify the
-checkpoint to continue from.
+(list them with `dvc exp show`), you must first `dvc exp apply` it before
+running it with `dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to
```
jorgeorpinel, line 197: Unclear what "it" refers to.

```suggestion
(list them with `dvc exp show`), you must first `dvc exp apply` it before
using `dvc exp run`. For `--queue` or `--temp` runs, use `--rev` to
```

jorgeorpinel:
p.s. sorry for the delay @omesser . Assigned you for visibility in https://github.com/orgs/iterative/projects/284.

content/docs/user-guide/experiment-management/checkpoints.md
omesser: @jorgeorpinel  - Thanks for bringing this discussion to my attention.
I was not aware of the rabbit hole 🐇  here 😅 . My edit here was strictly related to the usage as a verb (english) - my 2 cents about that (coincidentally 2 points 😄  ):

1. re: [first point raised in the issue description](https://github.com/iterative/dvc.org/issues/2960) - Though _persist_ also means "continuing to do something in a determined way" the usage here of "continuing to exist (maybe past an expected/usual time)" is definitely _also_ a valid definition in the general language (See: [[1]](https://dictionary.cambridge.org/dictionary/english/persist) [[2]](https://www.merriam-webster.com/dictionary/persist) [[3]](https://www.oxfordlearnersdictionaries.com/definition/american_english/persist) for example).
2. Probably more important - This is how it is conventionally used in our domain. "To persist X" means "to commit X to persistency" (of some sorts). So, the way I see it (and the reason I bothered "fixing" it) "checkpoints can become persistent by..." - would sound wrong / clunky to software engineers & people in associated fields (e.g. data engineering) - we have a verb for that. I would argue that familiarity and conciseness win here (even if it wasn't for my point no. 1)

But, on a more constructive note - As per the [last comment](https://github.com/iterative/dvc.org/issues/2960#issuecomment-999627440) am I correct to assume it's agreed if I make this change (only here): `Like with regular experiments, checkpoints can become persistent by committing them to git` -> `Like with regular experiments, checkpoints can be committed to git` [losing info here, not loving it, but better than now]

I was actually going to suggest following your comment to change the other places you mentioned - namely the section title from `Persisting experiments` to `Committing experiments` - but upon further consideration I actually find that problematic 😢 . IMHO "committing" is more of an implementation detail here. If I were to read/skim these docs and compare to other experiment tracking solutions I would look for (search / grep) the terms "persist" and "track" (=what) and not "commit" (=how) - all the more important when it comes to titles!

content/docs/user-guide/experiment-management/running-experiments.md
omesser: Hmm, the way I read it -  "experimental pipeline" suggests not only possession ("of experiment") but that the pipeline itself is somehow of experimental nature (i.e. not final?).
Here, we just want to convey possession to the experiment. There's nothing "experimental" about the pipeline *itself*.

content/docs/user-guide/experiment-management/running-experiments.md
omesser: @jorgeorpinel  - I change the command reference as well. But please note I noticed the command references fell out of sync (at least partly) with the actual `dvc` command help messages, see [exp run -r](https://github.com/iterative/dvc/blob/main/dvc/commands/experiments/run.py#L69)
I guess this calls for a realignment pass(es) or some mechanism in place to avoid disparity

content/docs/user-guide/experiment-management/running-experiments.md
omesser: Good catch, thanks!

omesser:
Hey @jorgeorpinel,
Pushed some changes and answered some comments - please do another pass when you get a chance

content/docs/user-guide/experiment-management/running-experiments.md
iesahin: I think "experiment pipeline" is better than "pipeline of the experiment"

iesahin (approved):
Other than a small point regarding `checkpoints.md`, looks good to go. Thank you @omesser

content/docs/user-guide/experiment-management/checkpoints.md
jorgeorpinel: > "continuing to exist (maybe past an expected/usual time)" is definitely also a valid definition

Yes but then the phrase should be "checkpoints can persist" which would be unclear. "Become persistent" is a little clunky indeed, but clear: going from temporary to continuous.

Anyway I see you changed it now along the lines of the decisions in #2960 so 👍🏼  

If you want to argue for another phrasing please comment in the issue, we can always revisit all these mentions.

content/docs/user-guide/experiment-management/checkpoints.md
```diff
@@ -32,8 +32,8 @@ similar to a branch.
 
 </details>
 
-Like with regular experiments, checkpoints can become persistent by
-[committing them to Git](#committing-checkpoints-to-git).
+Like with regular experiments, checkpoints can be
+[committing to Git](#committing-checkpoints-to-git).
```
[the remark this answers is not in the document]
jorgeorpinel:
```suggestion
[committed to Git](#committing-checkpoints-to-git).
```

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: "experimental" is not possessive though 🙂  "experiment's" would be possessive. "Experimental" is a qualitative adjective: it indicates a quality of the noun ("pipeline" here). We do not want to express possession above, we're just talking about the pipeline (which is itself an experiment).

> that the pipeline itself is somehow of experimental nature (i.e. not final?)

Yes, that's the intended meaning of the phrase. The hypothetical pipeline here is experimental in fact.

The problem IMO is the unnecessary complication of having the pipeline concept here, which again is something we want to address per #2768.

jorgeorpinel: > something we want to address separately, per https://github.com/iterative/dvc.org/issues/2768

Now that if you want to address that here, why not?

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: Thanks!

> I noticed the command references fell out of sync (at least partly) with the actual `dvc` command help messages, see [exp run -r](https://github.com/iterative/dvc/blob/main/dvc/commands/experiments/run.py#L69)

Thanks for mentioning it.

> I guess this calls for a realignment pass(es) or some mechanism in place to avoid them falling out of sync

The core team has a manual check in their PRs where they decide when docs changes are required but being manual there's room for error of course, so yeah that happens! Cc @dberenbaum

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: p.s. I also have no strong opinion agains "pipeline of the experiment" TBH, esp. since we want to remove the pipeline concept later so this will be reviewed again then.

jorgeorpinel (approved):

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: p.s. Actually, there's no un-alignment here. We don't intend to copy-paste option descriptions from help output. We often rewrite them completely in docs.

jorgeorpinel: p.p.s. if anything I'd say that the CLI output `Continue the specified checkpoint experiment` isn't as clear as the docs `continue an experiment from a specific checkpoint`. You don't need `--rev` to **resume** a checkpoint experiment, you need it to specify from which checkpoint to resume from.

jorgeorpinel: Created https://github.com/iterative/dvc/issues/7451 to follow up on that.

content/docs/user-guide/experiment-management/running-experiments.md
omesser: > "Yes, that's the intended meaning of the phrase. The hypothetical pipeline here is experimental in fact."

Have to admit I'm a bit confused about that sentence 😄 
@jorgeorpinel  - I don't want to go against your judgment & expertise here 
IMO - `experiment pipeline` is more appropriate here than `experimental pipeline`, but definitely can live with both.
I'll go with whichever you decide

content/docs/user-guide/experiment-management/running-experiments.md
omesser: > "We often rewrite them completely in docs."
Intriguing! 👀 
Understood

content/docs/user-guide/experiment-management/running-experiments.md
jorgeorpinel: K let's keep your change 👍🏼

jorgeorpinel:
Thanks @omesser !

content/docs/user-guide/experiment-management/running-experiments.md
omesser: 🙏

PR #3324, opened 2022-03-01
start: reorg params section in Exps page

jorgeorpinel:
per https://github.com/iterative/dvc.org/pull/3197#issuecomment-1055724585


jorgeorpinel:
WDYT @iesahin ?

iesahin (approved):
I think this is a good improvement. Thank you Jorge.

PR #3328, opened 2022-03-02
start: rever note on pipeline tracking from Versioning page

jorgeorpinel:
- [x] Revert note on pipeline tracking from Versioning page
- [ ] Does the Pipelines page need clarifying instead?

Per https://github.com/iterative/dvc.org/pull/3289#pullrequestreview-890336021

### DIFF
--- a/content/docs/start/data-and-model-versioning.md
+++ b/content/docs/start/data-and-model-versioning.md
@@ -85,11 +85,7 @@ outs:
 
 </details>
 
-> Note that there are [other mechanisms] to track data files/dirs besides `.dvc`
-> files.
-
 [format]: /doc/user-guide/project-structure/dvc-files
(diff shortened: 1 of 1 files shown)


content/docs/start/data-pipelines.md
```diff
@@ -75,6 +75,10 @@ A `dvc.yaml` file is generated. It includes information about the command we ran
 (`python src/prepare.py data/data.xml`), its <abbr>dependencies</abbr>, and
 <abbr>outputs</abbr>.
 
+DVC uses this metafiles to track the data used and produced by the stage, so
```
shcheklein, line 78: these

shcheklein (approved):

content/docs/start/data-pipelines.md
jorgeorpinel:
```suggestion
DVC uses these metafiles to track the data used and produced by the stage, so
```

Issue #2544, opened 2021-06-08
cases: list of ideas


jorgeorpinel:
### 1. Data Management

- [x] Data Versioning - https://dvc.org/doc/use-cases/versioning-data-and-model-files
- [ ] "Organizing Team Datasets" per https://github.com/iterative/dvc.org/issues/3186
- [x] Data Registries - https://dvc.org/doc/use-cases/data-registries

### 2. Data Pipeline development
> From https://github.com/iterative/dvc.org/issues/2544#issuecomment-857170399 below

> model development may include data validation and preprocessing followed by model training and evaluation...
compose this as a dag where I can easily and efficiently run only the necessary stages
iteratively update data, add features, tune models, etc (overlaps with 2.)

### 3. Experiment Management
> From https://github.com/iterative/dvc.org/issues/2270#issuecomment-793269476

Preliminary ideas:

- ~~Hyperspace exploration [Tuning/Optimization] ? May be too low level~~ There's a blog about this now.
- [x] Experiment ~~Bookkeeping~~ **Tracking** (with Git): Rapid iterations. UPDATE: https://github.com/iterative/dvc.org/pull/2782
- Visualizing Data Science (experiments + params, metrics/plots) + Viewer
- [ ] Experiment execution and orchestration (`exp`+`machine`+CML?)
  > From https://github.com/iterative/dvc.org/pull/2782#pullrequestreview-781487421

> Here we should sell W&B, MlFlow, etc - rapid iterations, live metrics + other metrics + navigation

### 4. Production environments/ MLOps 
> From https://github.com/iterative/dvc.org/issues/2490#issuecomment-853561046

4.1 **DVC in Production**
Training remotely
Deploying models (CLI or API)
Keep pipelines, artifacts in sync between environments
Batch scoring a.k.a. "DVC for ETL" - see https://github.com/iterative/dvc.org/issues/2512#issuecomment-854999981
\+ Distributed/parallel computing

> Good example of user perspective: https://discord.com/channels/485586884165107732/485596304961962003/872860674529845299

4.2 **ML Model Registry**
Model lifecycle (training, shadow, active, inactive)
Automated/Continuous training (remotely)
Discovery and reusability
Deploying models
Batch scoring example
\+ Real-time inference

4.3 **Production Integrations**
Databases (e.g. SQL dump versioning/preprocessing)
Spark (e.g. remote training)
AirFlow (e.g. batch scoring)
Kafka (e.g. real-time predictions)

[...]

PR #3333, opened 2022-03-03
cases: Model Registries

jorgeorpinel:
Per #2544

> **4.2 ML Model Registry**
Model lifecycle (training, shadow, active, inactive)
Automated/Continuous training (remotely)
Discovery and reusability
Deploying models
Batch scoring example
\+ Real-time inference

Review app page: https://dvcorg-casesmodelregistry.gtsb.io/doc/use-cases/model-registry


jorgeorpinel:
@shcheklein sorry for the long delay on this. Please take a look at the 2nd draft intro, benefits list, and paragraphs around DVC high-level features.

> The figure is still very generic but looking into improving that now... ⌛

content/docs/sidebar.json
```diff
@@ -92,7 +92,11 @@
         "slug": "fast-data-caching-hub"
       },
       "experiment-tracking",
-      "data-registries"
+      "data-registries",
+      {
+        "label": "ML Model Registries",
```
shcheklein, line 97: "Model Registry" sounds better to me (ML is clear from the context), Registries is not needed -usually you would have a single registry per team

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
+# Machine Learning Model Registries
+
+An **ML model registry** (or _model catalog_) works as a central repository
```
shcheklein, line 3: this is not only about being in between development and apps (another way to say - deployment, or production, etc) .... this is about discoverability - teams need this to see what models are available, who was building them, how, etc ...

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
...
+[experiments]: /doc/user-guide/experiment-management
+[metrics and plots]: /doc/start/metrics-parameters-plots
+
+Many of these benefits are built-into DVC: Your [data pipelines] and resulting
+artifacts (including ML models and performance metadata) are **codified** into
+one or mode <abbr>DVC projects</abbr>. This enables managing them with standard
+Git [versioning] workflows. Large files can be pushed to [remote storage], an
+efficient and scalable access point for your models.
+
+To achieve the other goals, the final outputs of these model development
```
shcheklein, line 36: I bet here people will be lost. They won't the motivation for this (why would we want to use a central repo)

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
...
+projects can be **packaged** and published in a global DVC repository. Existing
+model files stored remotely are reused in this collection without copying them,
+so building the registry is fast. Links to the original projects are preserved,
+so you can always retrieve full lineage and provenance info.
+
+Distributed DVC repos and managed storage are already great for sharing data and
+models. And with a model registry, getting them to production is also
+straightforward: Use the DVC CLI to [import] them for deployment, integrate them
```
shcheklein, line 44: it can be also DVC Python API, and probably `dvc get` rather than import

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
...
+  reproducible.
+- Share and control access to early-stage [experiments] and production-ready
+  models alike.
+- Save performance [metrics and plots] to quickly evaluate and compare models.
+- A platform to export, integrate, deploy, or roll-back published models
+
+[experiments]: /doc/user-guide/experiment-management
+[metrics and plots]: /doc/start/metrics-parameters-plots
+
+Many of these benefits are built-into DVC: Your [data pipelines] and resulting
```
shcheklein, line 30: can we do pipelines as second thing in the list or even omit (make it less DVC-specific), "how model is built", "its metrics", etc

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
...
+artifacts (including ML models and performance metadata) are **codified** into
+one or mode <abbr>DVC projects</abbr>. This enables managing them with standard
```
shcheklein, line 32: GitOps?

shcheklein: for SEO

content/docs/use-cases/ml-model-registries.md
```diff
@@ -0,0 +1,54 @@
...
+further improve your model catalogs with more features and better usability.
+
+![](/img/ml_model_registry_placement.jpg) _Model registry placement contrasts
+with general [data registries] for reusable data sources or training sets._
+
+<!-- TODO: Look up package registry diagrams for inspiration -->
+
+[data registries]: /doc/use-cases/data-registries
+
+DVC model registries provide your team with key collaborative capabilities:
```
shcheklein, line 17: can we emphasize basic things - it's Git-based, etc ....

shcheklein: in the list or in the paragraph above

shcheklein (changes requested):
Thanks @jorgeorpinel . It actually looks very good!

I left some comments here and there.

Also, there should be some space to put gto, mlem?

content/docs/sidebar.json
jorgeorpinel: Updated. Wait, should "Data Registries" also be singular?

jorgeorpinel:
> there should be some space to put gto, mlem?

Yep, I have lots of info for a 2nd half going into other tools but struggling a bit to summarize it. Will upload something soon ⌛

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: Updated intro. PTAL when you can

shcheklein: Looks good!

Second part `DVC supports it ` ... can we mention that DVC kinda enables it on top of Git ecosystem ? so, it's the same but a bit more explicit. Or it can be something like ... DVC runs on top of Git to .... wdyt?

jorgeorpinel: I'm not sure we want to mention Git in the intro paragraph or how to do that without getting technical. Let me try... ⌛

shcheklein: yep, maybe (Git ecosystem) would be enough? not sure ... I see your concern, but when we talk about DVC in other use cases we mention Git, so probably it's fine

jorgeorpinel: What about "DVC enables these features on the Git ecosystem, reducing friction between development and operations." ?

jorgeorpinel: or keep "..., leveraging existing development workflows."

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: > here people will be lost. They won't the motivation for this (why would we want to use a central repo

Is this still a concern with the new intro? We now say that "a central model catalog is used to discover, share, deploy, etc."

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: > it can be also DVC Python API

That's already in the text: "integrate them directly into code with the Python API" - changed to "DVC's Python API" to make it more explicit.

> probably dvc get rather than import

Good point, updated.

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: > do pipelines as second thing ... (make it less DVC-specific 

Agreed. Updated to "modeling process and performance metadata" (term "metrics and plots" is already used  earlier).

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: > GitOps for SEO

Inserted.

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: I added that DVC projects are Git-based. I think most other DVC basics are already covered but I'll double check... ⌛

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,52 @@
+# Machine Learning Model Registries
+
+A **model registry** is a central catalog for data artifacts produced in
+multiple data science projects. Models can be discovered, audited, shared,
+tested, and deployed from here. DVC enables these features on the Git ecosystem,
+reducing friction between development and operations. Other tools from
+[Iterative](https://iterative.ai/) provide more advanced features and better
+usability.
```
jorgeorpinel, line 8: > tools from Iterative

💡  It would be nice to have a landing page for the list of Iterative.ai products. Right now it's a menu you have to click on in the home page:

image

jorgeorpinel: Idea moved to https://github.com/iterative/iterative.ai/issues/431

content/docs/use-cases/ml-model-registries.md
shcheklein: a bit too fancy :) if we can do the same a bit more explicit that would be great ... something like "DVC is a CLI tool that works on top of Git to get you this features", wdyt?

content/docs/sidebar.json
shcheklein: yep, I think it should be Data Registry

jorgeorpinel: Updated in https://github.com/iterative/dvc.org/pull/3447

content/docs/use-cases/ml-model-registries.md
jorgeorpinel: OK. I thought we wanted keep the whole term "Git ecosystem" for SEO. Much easier to phrase as just "Git". Done:

> DVC works on top of Git to enable these capabilities on your existing ecosystem.

jorgeorpinel:
Just missing the better diagram to make this ready for final review (on the first half). My aim is to have that today ⌛

jorgeorpinel:
Latest [diagram](https://github.com/iterative/dvc.org/pull/3333/files#diff-073c6f9ca8b9f82ba87258f833c73e0b046bc6da58b4a81ddbf9cc0d6ec14d34) pasted here since our deployments are temporarily down:

image

jorgeorpinel:
p.s. Probably still too complicated but I'd like to know what to prioritize (keep)

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,52 @@
...
+
+To achieve the remaining goals, the final outputs of these model development
+projects can be **packaged** and published in a global DVC repository. Existing
+model files stored remotely are reused without copying them, so building the
+registry is fast. A connection to the original projects is preserved, so you can
+always retrieve full lineage and provenance information.
+
+Distributed DVC repos and managed storage are already great for [sharing] data
+and models. And getting your models from the registry to production is also
+properly tooled: deploy manually with the `dvc` [CLI], integrate with code using
```
shcheklein, line 42: `tooled`  - this sounds strange to me, not very common usage

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,52 @@
...
+files can be pushed to [remote storage], an efficient and scalable access point
+for your models.
+
+To achieve the remaining goals, the final outputs of these model development
+projects can be **packaged** and published in a global DVC repository. Existing
+model files stored remotely are reused without copying them, so building the
+registry is fast. A connection to the original projects is preserved, so you can
+always retrieve full lineage and provenance information.
+
+Distributed DVC repos and managed storage are already great for [sharing] data
```
shcheklein, line 40: `Distributed DVC repos and managed storage`  - `DVC managed repos` or maybe even "DVC repos" ? (to simplify this)

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,52 @@
...
+[experiments]: /doc/user-guide/experiment-management
+[metrics and plots]: /doc/start/metrics-parameters-plots
+
+Many of these benefits are built into DVC: Your [modeling process] and
+performance metadata become **codified** in Git-based <abbr>DVC projects</abbr>.
+This makes it possible to manage them with standard GitOps workflows. Large
+files can be pushed to [remote storage], an efficient and scalable access point
+for your models.
+
+To achieve the remaining goals, the final outputs of these model development
```
shcheklein, line 34: At this moment we think that Studio should be solving this - combining models from multiple repos.

Saving performance metrics, evaluate and compare - this can be done in a single repo with DVC or in Studio

`A platform to export, integrate, deploy, or roll-back published models` - this is done by `gto` (which is published already), by `mlem` (deploy, and it's coming), `roll-back` (also by GTO). Btw this item is a bit too complicated. In our case this system provide interface to access a model of a specific version (on CI, or from a service). ... other tools do more things

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,52 @@
...
+registries is between development and applications. See also [Data Registries]._
+
+[data registries]: /doc/use-cases/data-registries
+
+DVC model registries give your team key collaborative capabilities:
+
+- Store, organize, and track the [versions] of different models effectively.
+- Full data provenance and lineage trails guarantee that your modeling is
+  reproducible.
+- Share and control access to early-stage [experiments] and production-ready
```
shcheklein, line 20: this is not clear to be honest

shcheklein:
Image - good, a few comments.

- A bit too complicated still, but we can already send it.
- Discovery belongs to DS and ML engineers
- Simplify "modeling an testing"
- Share access remotely - not clear

shcheklein (changes requested):
A few comments here and there, again, let's send an image (maybe after one more iteration).

Btw, gto was released (repo is public) - we can and should mention it now.

content/docs/use-cases/model-registry.md
jorgeorpinel: Simplified.

content/docs/use-cases/model-registry.md
jorgeorpinel: Yes I think I wanted to go for "tooling". Using this for now:

> Now you also have proper tools to get models from the registry to production: ...

WDYT?

content/docs/use-cases/model-registry.md
jorgeorpinel: Updated to:

> - Share and control access to any data artifact, whether these are early-stage [experiments] or production-ready models.

I think that's clearer. I'm just not sure it's necessary to mention that range of possibilities anymore... but it's nice to link to experiments from somewhere here. WDYT?

jorgeorpinel:
See the updated image [here](https://github.com/iterative/dvc.org/pull/3333/files#diff-073c6f9ca8b9f82ba87258f833c73e0b046bc6da58b4a81ddbf9cc0d6ec14d34).

> Share access remotely - not clear

Changed to "share & access remotely". Is that clearer? I thought the concept of sharing/ accessing would be general and clear enough.

shcheklein:
> See the updated image [here](https://github.com/iterative/dvc.org/pull/3333/files#diff-073c6f9ca8b9f82ba87258f833c73e0b046bc6da58b4a81ddbf9cc0d6ec14d34).

I would remove the dotted line, also, models could be published to the registry automatically (small nuance, not sure worth complicating the diagram).

content/docs/use-cases/model-registry.md
shcheklein: okay, got it, that's fine!

content/docs/use-cases/model-registry.md
jorgeorpinel: > we think that Studio should be solving this - combining models from multiple repos

Yep but that release is still a bit away I think?

We could focus on a hypothetical manual registry done purely with DVC for now. We'll definitely want to revisit this use case soon anyway (it's in the Q2 plan already).

---

Working on the GTO mentions... ⌛

jorgeorpinel:
Small image update

ml_model_registry_placement

content/docs/use-cases/model-registry.md
jorgeorpinel: > this item is a bit too complicated. In our case this system provide interface to access a model of a specific version (on CI, or from a service)

Simplified for now.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+
+Many of these benefits are built into DVC: Your [modeling process] and
+performance metadata become **codified** in Git-based <abbr>DVC projects</abbr>.
+This makes it possible to manage them with standard GitOps workflows along with
+code. Large model files are stored separately and can be pushed to [remote
+storage], an efficient and scalable access point.
+
+The relevant outputs of these model development projects can then be
+**packaged** together with their metadata, and published in a [global Git
+repository]. With DVC, existing model files stored remotely are reused, so
```
jorgeorpinel, line 37: We first mention "global Git repo" and link to the GTO section.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+building the registry is fast. A connection to the original projects is
+preserved, so you can always get the full lineage and provenance information.
+
+DVC repos are already great for [sharing] data and models. They can also help
+you get models from the registry to production: deploying manually with the
+`dvc` [CLI] or integrating in code using DVC's Python [API]. You can even
+automate their training and delivery [via CML]. In fact any downstream system
+can sync with you models lifecycle using _GTO_.
```
jorgeorpinel, line 45: We then properly introduce GTO (associated with the "model lifecycle" concpt) in the deployment/ integration explanations.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+## Model Registries based on annotated Git tags
+
+[GTO] (Git Tag Ops) is a free, open-source tool that let's you turn any Git
+repository into an artifact registry. Tag artifact versions and annotate them
+with meaningful metadata. Manage model states throughout their lifecycle:
+promoting for production, rolling back, and reviewing their history. And any
+Git-based system can [sync with these events].
+
+GTO integrates nicely into DVC repos by using [metafiles] as annotations for
+your artifacts.
```
jorgeorpinel, line 64: Finally this is the GTO section for now.

jorgeorpinel (commented):
> Working on the GTO mentions... ⌛

@shcheklein we discussed integrating GTO into the existing story without a new section but it hasn't proven very easy and meanwhile I improved my existing version with a separate section that integrates well (I think) to the DVC storyline. Please consider it meanwhile 👇🏼 

But I'll continue trying a more fluid (and probably shorter) mention without the extra section... ⌛

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
+# Machine Learning Model Registry
+
+A **model registry** is a central catalog for data artifacts produced in
```
shcheklein, line 3: data artifacts -> ML models

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+multiple data science projects. Different models can be discovered, audited,
```
shcheklein, line 4: `multiple` - we can get rid of this

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+shared, tested, and deployed from here. DVC works on top of Git to enable these
+capabilities on your existing ecosystem. And other tools from
```
shcheklein, line 6: ecosystem -> existing software engineering stack

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+[Iterative](https://iterative.ai/) provide even better features and usability.
```
shcheklein, line 7: better -> more?

shcheklein: they are not better ... they are additional

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+registries is between development and applications. See also [Data Registries]._
+
+[data registries]: /doc/use-cases/data-registries
+
+Model catalogs give your team key collaborative capabilities:
+
+- Store, organize, and track the [versions] of different models effectively.
+- Full data provenance and lineage trails guarantee that your modeling is
+  reproducible.
+- Share and control access to any data artifact, from early-stage [experiment]
```
shcheklein, line 19: data -> model

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+  results to production-ready models.
+- Save performance [metrics and plots] to quickly evaluate and compare models.
```
shcheklein, line 21: Save -> Store (and show?)

content/docs/use-cases/model-registry.md
jorgeorpinel: Changed to more features and better usability

content/docs/use-cases/model-registry.md
shcheklein: text is good, but I still think we don't need the whole section, it feels artificial and too complicated

content/docs/use-cases/model-registry.md
jorgeorpinel: I thought "data provenance" was a specific concept + a key term

content/docs/use-cases/model-registry.md
shcheklein: GTO is not about global repos at all. I think we still make too much emphasis on the global repo setup. This is an advanced use case that would replace in CLU an upcoming Studio MR. I would not make the whole paragraph about this.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+code. Large model files are stored separately and can be pushed to [remote
+storage], an efficient and scalable access point.
+
+The relevant outputs of these model development projects can then be
+**packaged** together with their metadata, and published in a [global Git
+repository]. With DVC, existing model files stored remotely are reused, so
+building the registry is fast. A connection to the original projects is
+preserved, so you can always get the full lineage and provenance information.
+
+DVC repos are already great for [sharing] data and models. They can also help
```
shcheklein, line 41: It's not like repos help get us models - it's DVC provides APIs / tools to get models described in repos ... that are natively compatible with CI systems (since DVC works on top git) - it's "Git native" to some extent :)

content/docs/use-cases/model-registry.md
shcheklein: I meant ... data artifact -> model artifact 

Or maybe I'm missing your point?

shcheklein (changes requested):
Put some comments. We are getting there :)

Last paragraph got complicated, the previous one need some re-focusing if that makes sense at all. Happy to brainstorm on this in Slack if it's needed.

content/docs/use-cases/model-registry.md
jorgeorpinel: Hmmm... The first bullet already reads "Store ... models effectively" -> Changed that to "Collect".
Changed this bullet to "Store and present your model artifacts" (more focused on discoverability since we mention "interface to access" in the last bullet.

content/docs/use-cases/model-registry.md
jorgeorpinel: Maybe this part is somewhat unclear? I'm still thinking mainly on a pure DVC model registry at this point (imported artifacts from other repos). You could `dvc get` models directly from this kind of registry or integrate them with `dvc.api.open`...

jorgeorpinel:
> Happy to brainstorm on this in Slack if it's needed.

Let me give it one last shot right now (integrating GTO into the main story) and I'll ping you on Slack ⌛

content/docs/use-cases/model-registry.md
jorgeorpinel: Yes, the current text focuses on a "global" centralized registry that collects models from many projects, similar to the base case presented in https://dvc.org/doc/use-cases/data-registries but for models. I realize GTO is designed with source repos in mind though, so indeed this link text doesn't make sense.

But more broadly, I guess I should focus more on the cataloging aspect and less on the centralizing aspect in general.. ⌛

jorgeorpinel: * now that we're incorporating GTO

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
+# Machine Learning Model Registry
+
+A **model registry** is a repository to catalog ML models and their versions.
```
shcheklein, line 3: `is a repository` - we start with being DVC-specific?

repository -> tool, service, place ...

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+Artifacts from your data science projects can be discovered, audited, shared,
```
shcheklein, line 4: Let's keep it simple - Models, not artifacts

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+DVC works on top of Git to enable these capabilities on your existing software
+engineering stack. Other tools from [Iterative](https://iterative.ai/) provide
+complementary features and more usability.
+
+![](/img/ml_model_registry_placement.jpg) _A common placement for model
+registries is between development and applications. See also [Data Registries]._
+
+[data registries]: /doc/use-cases/data-registries
+
+Model registries give your team key collaborative capabilities:
```
shcheklein, line 15: `collaborative capabilities` - model registries are not only about collaboration ... I think previous version was better or we had some other problems/doubts?

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Store and present your model artifacts to others, from early-stage
```
shcheklein, line 17: just models  ... not need to introduce additional terms, they don't give any additional information?

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+  [experiment] results to production-ready models.
+- Collect, organize, and track the [versions] of different models effectively;
+  with full data provenance and lineage trails based on [reproducible modeling].
+- Save performance [metrics and plots] or other metadata to evaluate or compare
+  models.
+- An interface to access specific model versions from other systems (e.g. in
+  CI/CD)
```
shcheklein, line 24: or in production ...

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+- Centralize ML models from multiple projects into a global repository, if
```
shcheklein, line 25: ppl won't understand this ...

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+projects</abbr>. This makes it possible to manage them with standard GitOps
+workflows along with code. Large model files are stored separately and
+efficiently, and can be pushed to [remote storage] -- a scalable access point.
+
+DVC repos are also great for [sharing] data and models. You can get the
+artifacts described by DVC into production in a few ways: by deploying with the
+`dvc` [CLI], or integrating into Python code using the [API]. You can even
+automate their training and delivery [via CML].
+
+To build the registry, the relevant artifacts from your ML projects (whether
```
shcheklein, line 44: Still, I don't think we need to go that much into packaging models into a single repo. To some extent GitHub with multiple repos becomes that place, We can mention this, but it should come after GTO, MLEM, Studio MR, etc

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,64 @@
...
+using DVC or not) can be **packaged** together with their metadata using
+Git-native mechanics. One option is to annotate the repo using tags with [GTO].
```
shcheklein, line 46: One option ... I don't see the connection between the packaging part and one option tbh

shcheklein (changes requested):
Put some additional comments. Sorry that it took a bit of time.

content/docs/use-cases/model-registry.md
jorgeorpinel: "repository" has a general meaning too 🙂 which is what we were going for here.

image

content/docs/use-cases/model-registry.md
jorgeorpinel: OK removed most "artifact" mentions now.

content/docs/use-cases/model-registry.md
jorgeorpinel: Removed "collaborative". The list hasn't changed much recently except for the last bullet about centralizing, which I added recently.

content/docs/use-cases/model-registry.md
jorgeorpinel: Removed. I suppose the centralization feature is implied in other bullets and explanations. Not really the goal anyway, just a possible byproduct of this pattern

content/docs/use-cases/model-registry.md
jorgeorpinel: > I don't think we need to go that much into packaging models into a single repo

I agree. It's no longer my intention to give this impression. In fact we already removed all explicit info about centralizing. Here "packaging" doesn't have to mean to move/import them into a central repo.

Let me rework this to avoid the confusion though ⌛

content/docs/use-cases/model-registry.md
shcheklein: in this case it's clearly overloaded

jorgeorpinel: > repository -> tool, service, place

OK I did change this to "construct" after all since "repo"'s definition (computing context) may imply "central" per ☝🏼 (and we want to avoid emphasizing that aspect per https://github.com/iterative/dvc.org/pull/3333#discussion_r861087231).

content/docs/use-cases/model-registry.md
jorgeorpinel: I rewrote the last 2 paragraphs (mainly) to present and expand the GTO strategy first, then mention packaging and presenting models together virtually (i.e. Studio MR) or by actually centralizing them. PTAL

content/docs/use-cases/model-registry.md
jorgeorpinel: > I don't see the connection between the packaging part and one option

These ideas are now in separate paragraphs. Clearer? Or do you now think the package manager analogy is too far fetched @shcheklein ?

jorgeorpinel:
@aguschin please check out the last few paragraphs where GTO is introduced. Lmk if you have any comments please!

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,65 @@
+# Machine Learning Model Registry
+
+A **model registry** is a construct to catalog ML models and their versions.
```
shcheklein, line 3: sorry, but "construct" is also very unusual ... why not tool?

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,65 @@
...
+[CLI], or integrating into Python code using the [API]. You can even automate
+their training and delivery [via CML].
+
+To build a Git-native registry, the ML models in your repos can be annotated
+using tags. [GTO] (Git Tag Ops) is a tool to do just that! The annotations carry
+meaningful metadata, such as a model file path (or [DVC metafile]), development
+stage, deployment environment, etc. This helps you manage your models lifecycle:
+promoting versions, rolling back, and reviewing their history.
+
+This approach is like **packaging** the models with their related information,
```
shcheklein, line 48: this is what MLEM can do

content/docs/use-cases/model-registry.md
jorgeorpinel: No strong opinion
```suggestion
A **model registry** is a tool to catalog ML models and their versions.
```

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,65 @@
...
+Model registries give your team key capabilities:
+
+- Store and present your models to others, from early-stage [experiment] results
+  to production-ready models.
+- Collect, organize, and track the [versions] of different models effectively;
+  with full data provenance and lineage trails based on [reproducible modeling].
+- Save performance [metrics and plots] or other metadata to evaluate or compare
+  models.
+- An interface to access specific model versions from other systems (e.g. in
+  CI/CD or production environments)
```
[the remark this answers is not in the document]
jorgeorpinel: OK I did some rephrasing and consolidating of the existing bullets in order to include ☝🏼 

> one is missing: "Productionize" 

This is what I'm getting: "(Re)place different models on certain stages (dev, shadow, prod, etc.) without touching the application layer." WDYT?

> Security and audit trail

Added them as a single (last) item. PTAL.

content/docs/use-cases/model-registry.md
shcheklein: okay, I guess I'm still lost :) what is that we trying to say in this paragraph? could you clarify please?

I think MLEM is def ready to be mentioned / linked.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
...
+A **model registry** is a tool to catalog ML models and their versions. Models
+from your data science projects can be discovered, audited, shared, tested, and
+deployed from here. [Iterative](https://iterative.ai/) tools like
+[DVC](https://dvc.org/doc), [GTO], and [MLEM] enable these capabilities on top
+of Git, so you can stick to an existing software engineering stack. No more
+division between ML and App operations!
+
+![](/img/ml_model_registry_placement.jpg) _A common placement for model
+registries is between development and applications. See also [Data Registries]._
+
```
jorgeorpinel, line 12: Figure improvement idea: make Git a central concept e.g.

image
> From GTO presentation

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,69 @@
+# Machine Learning Model Registry
```
jorgeorpinel, line 1: Should we call this **Git-native Model Registry** or **MRaC** (MR as code) instead? In the future

content/docs/use-cases/model-registry.md
jorgeorpinel: p.s. applies to benefits list, which currently covers model registries in general, no Git-as-single-source-of-truth focus yet.

content/docs/use-cases/model-registry.md
shcheklein: A bit too complicate. Business use case is "Model Registry" - everything else hides it to my mind

content/docs/use-cases/model-registry.md
jorgeorpinel: Well I rewrote the bottom paragraphs now so it's quite different, plus I fully mentioned MLEM (iteration 1). PTAL @shcheklein ! Resolving this review for now.

> p.s. removed Studio mention for now since the MR release is still a bit away (I think).

jorgeorpinel:
Review app page: https://dvcorg-casesmodelregistry.gtsb.io/doc/use-cases/model-registry

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+can fully **package** models by introducing [MLEM] in any Python ML code that
+uses common libraries like `sklearn` or `xgboost`. MLEM automagically codifies
+all the context needed to produce data and models in the Git repo. It's also
+able to push them to any cloud storage (by itself or via DVC), as well as to
+list and transfer models between locations, or even serve them directly!
+
+Each of these complementary methods brings your ML process to GitOps practices,
+especially to manage and deploy models with mature software delivery methods.
+This means for example Github Actions or Gitlab CI/CD, which can now
+automatically sync with the state of the artifacts in your model registry.
```
jorgeorpinel, line 62: WDYT about this text @iterative/mlem ?

jorgeorpinel: p.s. my self-review is that the body of the page has become a little long even when these paragraphs are extremely summarized. Maybe we should consider putting them a 2nd section so the structure is clearer (0. general MR benefits, 1. DVC layer, 2. other alternative/complementary layers i.e. GTO and MLEM for now). Or to use more diagrams, other bullet/numbered lists, etc. to make the text less "boring" so-to-speak. That said all that could be left for a near future iteration once Studio MR is released.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+[versions]: /doc/use-cases/versioning-data-and-model-files
+[reproducible modeling]: /doc/start/data-pipelines
+
+Many of these benefits are built into DVC: Your [modeling process] and
+performance information become **codified** in Git-based <abbr>DVC
+projects</abbr>. This makes it possible to manage them with standard Git
+workflows along with code. Large model files are stored separately and
+efficiently, and can be pushed to [remote storage] -- a scalable access point.
+
+DVC repos are also great for [sharing] any data artifact. You can get the models
```
shcheklein, line 41: it feels we don't need this paragraph, I don't feel it's value to be honest ... what is the main idea here?

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+
+An additional route to build a Git-native registry is to **annotate** ML models
+in your repos using tags. [GTO] (Git Tag Ops) is a tool to do just that! The
+annotations carry meaningful metadata, such as a model file path, deployment
+env, etc. and get codified in the repo. GTO also helps manage your models
+lifecycle: promoting versions, rolling back, and reviewing history.
+
+To finish closing the gap between model training and software development, you
+can fully **package** models by introducing [MLEM] in any Python ML code that
+uses common libraries like `sklearn` or `xgboost`. MLEM automagically codifies
```
shcheklein, line 54: let's not mention libraries, otherwise we would have to mention all of them

shcheklein: I would simplify this paragraph and try to merge with GTO

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+all the context needed to produce data and models in the Git repo. It's also
+able to push them to any cloud storage (by itself or via DVC), as well as to
+list and transfer models between locations, or even serve them directly!
+
+Each of these complementary methods brings your ML process to GitOps practices,
```
shcheklein, line 59: "methods" - feel like it's not the right term

overall "While each tool can be used independently ... .This ecosystem of tools ..."

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+especially to manage and deploy models with mature software delivery methods.
```
shcheklein, line 60: "mature" - drop it

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,70 @@
...
+This means for example Github Actions or Gitlab CI/CD, which can now
```
shcheklein, line 61: example in this very general paragraph seems artificial (or requires other examples)

content/docs/use-cases/model-registry.md
shcheklein: yes. agreed, it's too long now

content/docs/use-cases/model-registry.md
jorgeorpinel: Selling DVC. Specific features it provides to the MR.

content/docs/use-cases/model-registry.md
jorgeorpinel: Changed.

content/docs/use-cases/model-registry.md
jorgeorpinel: Done.

content/docs/use-cases/model-registry.md
jorgeorpinel: Changed examples for "continuous integration".

content/docs/use-cases/model-registry.md
jorgeorpinel: I kind of like having a paragraph for each tool.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,62 @@
...
+for [sharing].
+
+An additional route to build a Git-native registry is to **annotate** ML models
+in your repos using tags. [GTO] (Git Tag Ops) is a tool to do just that! The
+annotations carry meaningful metadata, such as a model file path, deployment
+env, etc. and get codified in the repo. GTO also helps manage your models
+lifecycle: promoting versions, rolling back, and reviewing history.
+
+And to fully **package** models, you can introduce [MLEM] in any Python code
+that uses common ML libraries. It automagically codifies all the context needed
```
shcheklein, line 49: let's try to replace codifies pls - it too technical I think

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,62 @@
...
+
+Many of these benefits are built into DVC: Your [modeling process] and
+performance information become **codified** in Git-based <abbr>DVC
+projects</abbr>. This makes it possible to manage them with standard Git
+workflows along with code. Large model files are stored separately and
+efficiently, and can be pushed to [remote storage] -- a scalable access point
+for [sharing].
+
+An additional route to build a Git-native registry is to **annotate** ML models
+in your repos using tags. [GTO] (Git Tag Ops) is a tool to do just that! The
```
shcheklein, line 43: `do just that` - can we make it sound a bit less like an ad :)

shcheklein (approved):
okay, some final changes + update the image and let's 🚢 it

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,62 @@
...
+annotations carry meaningful metadata, such as a model file path, deployment
+env, etc. and get codified in the repo. GTO also helps manage your models
+lifecycle: promoting versions, rolling back, and reviewing history.
```
aguschin, line 46: In fact, with git tags we register semantic versions and promote them to production. We `annotate` by writing to `artifacts.yaml`. Also, I would bring forward the use cases instead of providing some details, e.g.

```suggestion
An additional route to build a Git-native registry is to use git tags to **register**
and **promote** [semantic versions](https://semver.org) of ML models. 
[GTO] (Git Tag Ops) helps you do that. Besides helping you manage models lifecycle, 
GTO helps you **annotate** your models to define the registry-level names for your DVC outputs.
```

By `registry-level names` I mean human-friendly names you would like to see in MR, instead of some path. I'm sure there should be a better wording to explain this)

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,62 @@
...
+
+And to fully **package** models, you can introduce [MLEM] in any Python code
+that uses common ML libraries. It automagically codifies all the context needed
+to produce them, and is able to push them to any cloud storage (by itself or via
+DVC), list and transfer them among locations, or even serve them directly!
```
aguschin, line 51: 1. I think the first sentence may be simpler
2. I would add more about MLEM as deploy - to let people know their MR can be used to simplify deployment
3. MLEM doesn't codify what's needed **to produce** models, but to serve them.
4. **automagically** is nice 😂 

An option
```suggestion
And to fully **package** models, you can save them with [MLEM]
in your Python training scripts. It automagically saves all the context needed 
to run the model, and is able to push them to any cloud storage (by itself or via
DVC), list and transfer them among locations, or even serve and deploy them 
to your favorite deployment platforms!
```

`favorite deployment platforms` is exagerration, we can remove that if you think it's better to be more humble until we have more supported providers

content/docs/use-cases/model-registry.md
jorgeorpinel: Changed to "captures".

content/docs/use-cases/model-registry.md
jorgeorpinel: Rephrased.

content/docs/use-cases/model-registry.md
jorgeorpinel: Thanks! The suggested text is too technical, but I rewrote the GTO paragraph to mention the use cases more accurately. PTAL @aguschin 🙏🏼

content/docs/use-cases/model-registry.md
jorgeorpinel: Rephrased as well!

jorgeorpinel:
> \+ update the image

@shcheklein do you mean [properly designed](https://iterativeai.slack.com/archives/C03AJN5PC75/p1649364736160299)? That may take some time. Meanwhile we could iterate on it ([here's an idea](https://github.com/iterative/dvc.org/pull/3333#pullrequestreview-970119457)) and have a provisional one.

shcheklein:
@jorgeorpinel let's iterate then, the one from the idea - you mentioned that it's more like a second image, it's bit too low level for the first one I think

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,63 @@
...
+performance information become **codified** in Git-based <abbr>DVC
+projects</abbr>. This makes it possible to manage them with standard Git
+workflows along with code. Large model files are stored separately and
+efficiently, and can be pushed to [remote storage] -- a scalable access point
+for [sharing].
+
+To make a Git-native registry (on top of DVC or not), one option is to use [GTO]
+(Git Tag Ops). It tags ML model releases and promotions, and links them to
+artifacts in the repo using versioned **annotations**. This creates abstractions
+for your models, which lets you manage their lifecycle freely and directly from
```
aguschin, line 45: I would make "lifecycle" bold instead of annotations. Annotations are great, but main point for GTO to exist is to manage the model lifecycle in Git. The annotations are important to connect GTO to actual files (models) in the repo, but that doesn't bring value by itself.

content/docs/use-cases/model-registry.md
```diff
@@ -0,0 +1,63 @@
...
+Git.
+
+And to **package** the models, you can save them with [MLEM] in your Python
+training code. It automagically captures all the context needed to deploy them,
+and saves it to the repo. It can save your model files in cloud storage (by
+itself or with DVC), list and transfer them among locations, or even serve them
```
aguschin, line 51: I still miss more clear indication that MLEM helps you use your model downstream. Pack it, deploy it. But since this is the page about MR, maybe that's ok.

content/docs/use-cases/model-registry.md
aguschin: Everything else looks good! Thanks for the huge work @jorgeorpinel !

jorgeorpinel: Done.

content/docs/use-cases/model-registry.md
jorgeorpinel: Yeah this paragraph improves with my understanding of MLEM. Reviewing mlem.ai/doc contents has helped 😋 . I rephrased it again, PTAL @aguschin

jorgeorpinel:
> let's iterate then

@shcheklein improvements based on questions from design thread:

image

> https://github.com/iterative/dvc.org/pull/3333/files#diff-073c6f9ca8b9f82ba87258f833c73e0b046bc6da58b4a81ddbf9cc0d6ec14d34

shcheklein:
@jorgeorpinel good, but looks too complicated now ... can we simplify it significantly (less arrows, may be less icons here and there, etc?)

jorgeorpinel:
I rearranged things a little both with the figure (much simpler now) and benefits list + Added some links and admons (d6d01c68f3e52578d3a9adcd5d6eac2e1a408981).

Please the latest https://dvcorg-casesmodelregistry.gtsb.io/doc/use-cases/model-registry

shcheklein:
@jorgeorpinel it looks great! way simpler, I think we are ready to merge and then ask folks to re-draw it if needed.

shcheklein:
Good stuff 🎉

jorgeorpinel:
There is a "pretty" version of the image but not really improved in terms of the design, so I'm not publishing it yet. Just for the record:

image

jorgeorpinel:
Also should we link to https://iterative.ai/model-registry ? I just found out about that page!

PR #3325, opened 2022-03-01
live: remove summary

daavoo:
P.R. for https://github.com/iterative/dvclive/pull/217

PR #3345, opened 2022-03-09
ref: rewrap a couple usage blocks

jorgeorpinel:
> Per https://github.com/iterative/dvc.org/pull/3325#pullrequestreview-903978598 and https://github.com/iterative/dvc.org/pull/3325#pullrequestreview-903978835 originally

Some quality-related ideas on what criteria to use when formatting usage blocks. Potentially we can later match this in the core repo too (`-h` help output).

- [x] Establish clear-enough criteria -- see https://github.com/iterative/dvc.org/pull/3345#pullrequestreview-910746273
- [ ] Update core repo as well
- [ ] Update all other commands both in core and docs...


content/docs/command-reference/run.md
```diff
@@ -7,11 +7,12 @@ command.
...
-               [-p [<filename>:]<params_list>] [-m <path>] [-M <path>]
-               [--plots <path>] [--plots-no-cache <path>] [--live <path>]
-               [--live-no-cache <path>] [--live-no-html]
-               [-w <path>] [-f] [--outs-persist <filename>]
+               [--no-commit] [-d <path>] [-o <filename>]
+               [-O <filename>] [-p [<filename>:]<params_list>]
+               [-m <path>] [-M <path>] [--plots <path>]
+               [--plots-no-cache <path>] [--live <path>]
+               [--live-no-cache <path>] [--live-no-html] [-w <path>]
+               [-f] [--outs-persist <filename>]
```
jorgeorpinel, line 15: OK this is a wrapping that respects the order from the help output and doesn't scroll horizontally at min width on Windows (up to 72 chars), but it seems like on Mac it does (68 chars) 🤷🏼 

image

content/docs/command-reference/run.md
```diff
@@ -6,13 +6,16 @@ command.
...
+               [-d <path>] [-p [<filename>:]<params_list>]
+               [-o <filename>] [-O <filename>] [-c <filename>]
+               [--outs-persist <filename>]
+               [--outs-persist-no-cache <filename>]
+               [-m <path>] [-M <path>]
+               [--plots <path>] [--plots-no-cache <path>]
+               [--live <path>] [--live-no-cache <path>]
+               [--live-no-html] [-w <path>] [-f]
                [--always-changed] [--external] [--desc <text>]
                command
```
jorgeorpinel, line 20: OR: Here's a new way to organize the usage block. Its sorted manually by related flags (does not preserve the semi-random order from the help output) + it's all under 68 chars long -- no scroll bar even on Mac:

image

iesahin: I think this version is better.

iesahin (approved):

content/docs/command-reference/run.md
jorgeorpinel: Agree but we'd probably want the tool itself (core repo) to use the same order. WDYT @dberenbaum ?

daavoo (approved):
Is there a way to handle this criteria in the auto-formatter/linter?

content/docs/command-reference/run.md
iesahin: I'm not sure if the core uses [custom formatter](https://docs.python.org/3/library/argparse.html#formatter-class), but from the looks of it, it's not an easy feat to specify the sorting of these. I think we can have slightly different sorting between the docs and the tool.

jorgeorpinel:
Meaning the line width specific to code blocks? I don't think so unfortunately, in fact code blocks are allowed any line width by the linter. We should always sanity check the review app before merging -- that's how I detect code blocks that are scrolling horizontally.

content/docs/command-reference/run.md
jorgeorpinel: From what I remember it's just the order in which args are added in the code itself so it has to be done manually. I'd be happy to contribute the PR to match this one but after that we'd need a little project (issue) to check all the commands.

content/docs/command-reference/repro.md
```diff
@@ -7,9 +7,9 @@ correct order.
 ## Synopsis
 
 ```usage
-usage: dvc repro [-h] [-q | -v] [-f] [-s] [-m] [--dry] [-i]
-                 [-p] [-P] [-R] [--no-run-cache] [--force-downstream]
-                 [--no-commit] [--downstream] [--pull] [--glob]
+usage: dvc repro [-h] [-q | -v] [-f] [-s] [-m] [-i] [-p] [-P] [-R]
```
dberenbaum, line 10: So this groups all the short flags at the start? Looks good to me.

content/docs/command-reference/run.md
```diff
@@ -6,13 +6,16 @@ command.
...
+usage: dvc run [-h] [-q | -v] [-n <name>] [-f]
+               [--no-exec] [--no-run-cache] [--no-commit]
+               [-d <path>] [-p [<filename>:]<params_list>]
+               [-o <filename>] [-O <filename>] [-c <filename>]
+               [--outs-persist <filename>]
+               [--outs-persist-no-cache <filename>]
+               [-m <path>] [-M <path>]
+               [--plots <path>] [--plots-no-cache <path>]
+               [--live <path>] [--live-no-cache <path>]
+               [--live-no-html] [-w <path>]
```
dberenbaum, line 18: Is the idea to put all the dependency and output flags together and in some meaningful order?

content/docs/command-reference/run.md
```diff
@@ -6,13 +6,16 @@ command.
...
                [--always-changed] [--external] [--desc <text>]
```
dberenbaum, line 19: And these are stage-level options?

content/docs/command-reference/repro.md
jorgeorpinel: Yeah I think that on the core repo, currently new arguments are just added at the bottom so there's no logical order.

My criteria here isn't super defined but broadly: put general ones in the first line, esp. if they're shot flags; Then group by functionality in the order in which we think people will use them most -- in the case above (`repro`) realted to deps/params, then outs/metrics/plots, then DVCLive.

content/docs/command-reference/run.md
jorgeorpinel: Yep

content/docs/command-reference/run.md
```diff
@@ -6,13 +6,16 @@ command.
...
 ```usage
-usage: dvc run [-h] [-q | -v] [-n <name>] [--no-exec] [--no-run-cache]
-               [--no-commit] [-d <path>] [-o <filename>] [-O <filename>]
-               [-p [<filename>:]<params_list>] [-m <path>] [-M <path>]
-               [--plots <path>] [--plots-no-cache <path>] [--live <path>]
-               [--live-no-cache <path>] [--live-no-html]
-               [-w <path>] [-f] [--outs-persist <filename>]
-               [--outs-persist-no-cache <filename>] [-c <filename>]
+usage: dvc run [-h] [-q | -v] [-n <name>] [-f]
+               [--no-exec] [--no-run-cache] [--no-commit]
```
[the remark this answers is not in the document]
jorgeorpinel: They just looked like they can be grouped 😋 . Maybe they should be closer to the bottom of the list since they're probably not used very often?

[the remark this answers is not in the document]
dberenbaum: Agreed. I'd probably put them at the bottom not only because they're not used often but also because it makes the options list look similar to `dvc stage add`..

content/docs/command-reference/repro.md
dberenbaum: If we move `[--no-exec] [--no-run-cache] [--no-commit]` to the bottom for `run` and `stage add`, can we put the related `repro` options `[--pull] [--no-run-cache] [--no-commit]` at the end here?

daavoo: Should the criteria be translated back to dvc core repo?

dberenbaum: Yeah, once there is an agreement on the order of everything, we should update the core repo.

jorgeorpinel: > move [--no-exec] [--no-run-cache] [--no-commit] to the bottom for run

Done. Well, I moved them above the DVCLive ones, PTAL

> and stage add

Doesn't have them.

> put the related repro options [--pull] [--no-run-cache] [--no-commit] at the end

Done.

---

- [ ] BTW `run --no-exec` is equivalent to `repro --dry` right? May want to rename the former to "dry" as well.

iesahin (approved):

iesahin:
I approved this again but is this an epic?

content/docs/command-reference/run.md
```diff
@@ -6,14 +6,17 @@ command.
...
-               [--plots <path>] [--plots-no-cache <path>] [--live <path>]
+usage: dvc run [-h] [-q | -v] [-n <name>] [-f]
+               [-d <path>] [-p [<filename>:]<params_list>]
+               [-o <filename>] [-O <filename>] [-c <filename>]
+               [--external] [--outs-persist <filename>]
+               [--outs-persist-no-cache <filename>]
+               [-m <path>] [-M <path>]
+               [--plots <path>] [--plots-no-cache <path>]
+               [--no-exec] [--no-commit] [--no-run-cache]
+               [--always-changed] [--live <path>]
```
dberenbaum, line 17: My preference would be to start by copying what you have below for `dvc stage add` and then append the additional options (`--no-exec`, `--no-commit`, --no-run-cache`) at the very end. What do you think?

jorgeorpinel:
> is this an epic?

No but there are follow up tasks (I'll make an issue after this or something).

content/docs/command-reference/repro.md
```diff
@@ -7,9 +7,11 @@ correct order.
...
 
 ```usage
-usage: dvc repro [-h] [-q | -v] [-f] [-s] [-m] [--dry] [-i]
-                 [-p] [-P] [-R] [--no-run-cache] [--force-downstream]
-                 [--no-commit] [--downstream] [--pull] [--glob]
+usage: dvc repro [-h] [-q | -v] [-f] [-i]
+                 [-s] [-p] [-P] [-R]
+                 [--downstream] [--force-downstream]
+                 [--dry] [--no-commit] [--no-run-cache]
+                 [--pull] [-m] [--glob]
```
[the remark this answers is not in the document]
jorgeorpinel: Yeah `-m|M` are in their functionality groups (metrics, part of outputs). I only moved general and misc. short flags to the top.

Agree about `--glob`...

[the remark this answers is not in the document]
jorgeorpinel: > -m|M are in their functionality groups

Ah wait, repro only has `-m` so I guess it counts as a misc. option in this case. Committing suggestion!

content/docs/command-reference/run.md
jorgeorpinel: Good call. Standardized in cb88988e.

content/docs/command-reference/repro.md
jorgeorpinel: `repro` is not read-only.

PR #3282, opened 2022-02-15
Test merge-schedule

julieg18:
* testing if merge-schedule can merge a non approved pr

/schedule 2022-03-10

PR #3346, opened 2022-03-09
Fix broken merge-schedule gh action

julieg18:
* updates `if` statement in action to check for event type

### DIFF
--- a/.github/workflows/merge-schedule.yaml
+++ b/.github/workflows/merge-schedule.yaml
@@ -12,7 +12,8 @@ jobs:
   merge_schedule:
     runs-on: ubuntu-latest
     if:
-      ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+      ${{ github.event_name == 'schedule' ||
+      github.event.pull_request.head.repo.full_name == github.repository }}
     steps:
       - uses: gr2m/merge-schedule-action@v1
         with:


.github/workflows/merge-schedule.yaml
```diff
@@ -12,7 +12,8 @@ jobs:
   merge_schedule:
     runs-on: ubuntu-latest
     if:
-      ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+      ${{ github.event_name == 'schedule' ||
```
julieg18, line 15: Figured out why #3282 wasn't running! This action runs on `pull_request` (for finding prs that need to be merged) and on `schedule` (for actually merging the prs). This `if` statement was stopping the action from running on `schedule`. Hopefully, this update, which checks for `event_name`, should fix the action.

yathomasi: Do we need to have if statement at all 🤔 ? If I understand correctly, [it is mentioned](https://github.com/marketplace/actions/merge-schedule) that pull requests from forks are ignored and I believe the pull request is generated when we have `/schedule date` so the event name would be `schedule` itself.

julieg18: > Do we need to have if statement at all 🤔 ? 

Yes! The action still runs on forks and fails on them. See [#3215 convo](https://github.com/iterative/dvc.org/pull/3215#issuecomment-1035489731). So we put down the if statement (#3267) so that this action will get skipped on forks.

rogermparent (approved):

PR #3347, opened 2022-03-10
Fix: keyboard interaction for dropdown

yathomasi:
>I just realized that by using onPointerUp, keyboard users are not able to click on the dropdown. 

Added keyboard support for the dropdown. Update on #3317


julieg18 (approved):
Looks great!

Issue #3309, opened 2022-02-22
`start`: Update `dvc run` references in various GS pages


iesahin:

jorgeorpinel:
There's already check boxes for this in https://github.com/iterative/dvc.org/issues/2076 .

iesahin:
Yes, and I created this from that checkbox actually :)

iesahin:
It looks most of the references are in `data-pipelines` and `params-metrics-plots` document. The latter is already rewritten, and I'll rewrite the former.

jorgeorpinel:
Only https://dvc.org/doc/start/metrics-parameters-plots is missing now, I think.

iesahin:
#3050 replaces that page completely.

PR #3348, opened 2022-03-10
start: Replace `dvc run` references in data-pipelines

iesahin:
This replaces `dvc run` examples in `data-pipelines`. 

Related #3309

### DIFF
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -23,10 +23,10 @@ https://youtu.be/71IGzyH95UY
 
 ## Pipeline stages
 
-Use `dvc run` to create _stages_. These represent processes (source code tracked
-with Git) which form the steps of a _pipeline_. Stages also connect code to its
-corresponding data _input_ and _output_. Let's transform a Python script into a
-[stage](/doc/command-reference/run):
+Use `dvc stage add` to create _stages_. These represent processes (source code
+tracked with Git) which form the steps of a _pipeline_. Stages also connect
+code to its corresponding data _input_ and _output_. Let's transform a Python
+script into a [stage](/doc/command-reference/stage):
 
 <details>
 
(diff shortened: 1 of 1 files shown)


shcheklein (approved):
`dvc stage add` doesn't actually run it, right? We should check though that we don't have anywhere text that assumes that artifacts are already created, etc. We would need to run `dvc exp run` for that.

content/docs/start/data-pipelines.md
```diff
@@ -64,11 +64,11 @@ Please also add or commit the source code directory with Git at this point.
...
-          -o data/prepared \
-          python src/prepare.py data/data.xml
+$ dvc stage add -n prepare \
+                -p prepare.seed,prepare.split \
+                -d src/prepare.py -d data/data.xml \
+                -o data/prepared \
+                python src/prepare.py data/data.xml
 ```
 
 A `dvc.yaml` file is generated. It includes information about the command we ran
```
jorgeorpinel, line 74: > about the command we ran

Nothing ran though.

jorgeorpinel: > We would need to run dvc exp run for that.

Why not `repro` here @shcheklein ? Experiments complicates things under-the-hood and we don't want to go into those details in this page.

jorgeorpinel (changes requested):
☝🏼

jorgeorpinel:
> Related https://github.com/iterative/dvc.org/issues/3309

It doesn't close it?

content/docs/start/data-pipelines.md
iesahin: Fixed that. 

`repro` is done after more commands added. I also don't think we should mention `exp run` here.

jorgeorpinel (approved):

content/docs/start/data-pipelines.md
```diff
@@ -307,8 +306,8 @@ stages:
 
 </details>
 
-DVC pipelines (`dvc.yaml` file, `dvc run`, and `dvc repro` commands) solve a few
-important problems:
+DVC pipelines (`dvc.yaml` file, `dvc stage run`, and `dvc repro` commands) solve
```
jorgeorpinel, line 309: Typo, I think (committing)
```suggestion
DVC pipelines (`dvc.yaml` file, `dvc stage add`, and `dvc repro` commands) solve
```

PR #3352, opened 2022-03-11
Added custom code based slugger for making `<details>` blocks linkable

yathomasi:
This is another solution for making a unique link for the `<details>` block. This uses custom Slugger to provide the custom code/word to make the hash link unique and eliminates the possibility of links pointing to another place if we perform some page updates.

Some considerations with this methods are:
- the code must be provided in case of duplicates titles otherwise development will show an error and the build also fails
- Also, there is also an optional props `position` for the position of the code in the hash link (which is "end" by default, also accepts "start")
 
For an example implementation, I have updated `docs/start/data-and-model-versioning` which had two `<details>` components with the title `Expand to see what happens under the hood.` so I have updated one with `code` and also used optional `positoin` to place the code at the start of the hash id. 

Screen Shot 2022-03-11 at 19 52 14

The generated link will be `/doc/start/data-and-model-versioning#storing-and-sharing-Expand-to-see-what-happens-under-the-hood`.

Note: This preview deployment may fail since I have not updated all the docs having the same title.

Continue on PR #3329

### DIFF
--- a/content/docs/start/data-and-model-versioning.md
+++ b/content/docs/start/data-and-model-versioning.md
@@ -145,7 +145,7 @@ $ dvc push
 Usually, we also want to `git commit` and `git push` the corresponding `.dvc`
 files.
 
-<details>
+<details code="storing and sharing" position="start">
 
 ### 💡 Expand to see what happens under the hood.
 
--- a/plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
+++ b/plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
@@ -4,7 +4,8 @@ import React, {
   useRef,
   ReactNode,
   ReactElement,
-  useContext
+  useContext,
+  useMemo
 } from 'react'
 import cn from 'classnames'
 import { nanoid } from 'nanoid'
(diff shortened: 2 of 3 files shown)


rogermparent (commented):
I'm not too enthusiastic about this one, if a manual id is going to be provided then I imagine it should entirely take the place of the generated slug; a completely manual id gives the same benefits as a combined slug with simpler code and the added benefit of being able to change the details block's text without breaking existing links to it.

yathomasi:
> I'm not too enthusiastic about this one, if a manual id is going to be provided then I imagine it should entirely take the place of the generated slug; a completely manual id gives the same benefits as a combined slug with simpler code and the added benefit of being able to change the details block's text without breaking existing links to it.

Or I think we can do it this way:

 if we add `position` prop then it will be added to the respective position i.e `start` or `end` otherwise `code` will be taken entirely for id.

rogermparent:
> if we add `position` prop then it will be added to the respective position i.e `start` or `end` otherwise `code` will be taken entirely for id.

It would be an improvement, but I don't see any use case for a code+content-derived id that isn't already solved by a code-only id.

yathomasi:
close in favor of #3329.

Issue #1368, opened 2019-07-11
config: document the way config files play together, add some how-tos or examples?


kopytjuk:
EDIT by @shcheklein 

Repurposed from a few issues - both come from some misunderstanding/lack of proper docs on how config files (default vs local vs system vs global) play together.

https://github.com/iterative/dvc.org/issues/1368#issuecomment-634456922

------------------------

Currently, there are only two ways to handle secrets (e.g. connection strings, tokens, etc)

1. Use `.dvc/config.local` file. Since this file is not shared among developers in Git, a team has to agree on a local configuration to pull/push data.
2. Use a single environment variable, e.g. `AZURE_STORAGE_CONNECTION_STRING` - but if one has different storage destinations, one variable is not enough.

It would be very handy to use `.dvc/config` but with own placeholders for secrets:

```
['remote "my-azure-remote"']
url = azure://my-container/dvc/
connection_string = $MY_SECRET
```
If the right secret is set, we can share `.dvc/config` without revealing secrets in Git.

kopytjuk:
> It's not a well known or documented feature of DVC, but DVC merges sections with same name from different configs. What it means in our case ...

Hey, thank you for your idea - that "merging" behaviour was not known to me - seems like a good solution.

jorgeorpinel:
Checking back on this, for now I made a small update to the `dvc config` ref. (see https://github.com/iterative/dvc.org/pull/3359) but the larger change will be to address https://github.com/iterative/dvc.org/issues/340 which includes this one. Closing as dupe

Issue #340, opened 2019-05-15
guide: extract part of  the`config` cmd ref.


jorgeorpinel:
The [Configuration sections](https://dvc.org/doc/commands-reference/config#configuration-sections) is too long for the cmd ref. Probably the best route is to extract the parts of the current cmd ref description and all the config options into a new User Guide, and leave the rest in the cmd ref.

> From https://github.com/iterative/dvc.org/pull/331#pullrequestreview-236975854

---

UPDATE from https://github.com/iterative/dvc.org/issues/1368#issuecomment-634456922:

> It's not a well known or documented feature of DVC, but DVC merges sections with same name from different configs

^ This can be part of the intro or index to the configuration guide.

- [x] Include basic remote storage config info and/or link to https://dvc.org/doc/user-guide/data-management/remote-storage

ryokugyu:
@shcheklein anyone working on it? 
if not please assign me this issue.

shcheklein:
@ryokugyu no, but let's hold on for a while. We are still deciding on the exact structure. Are you looking for something next to contribute?

ryokugyu:
@shcheklein yes, i am looking for something to contribute.

shcheklein:
How about this one - https://github.com/iterative/dvc.org/issues/306 ? It's a good one, with a nice description and really valuable. Just mention that you are working on it if you decide to take it. Thanks :)

ryokugyu:
sure, let me dig it. i will need some help for this.
I am working on it.

ryokugyu:
@shcheklein is the structure decided?

shcheklein:
@ryokugyu no, not yet. @jorgeorpinel has simplified the config section significantly, so this is issue not that pressing now.

jorgeorpinel:
I've updated the issue description to include.

> Probably the best route is to extract the parts of the current cmd ref description and all the options into a new User Guide, and leave the rest in the cmd ref.

What do you guys think? Sorry took so long!

shcheklein:
@jorgeorpinel yep, agreed. I would start probably by writing User Guide and taking some parts from the commands reference. When it's ready we'll clean the CR sections and put the appropriate links.

jorgeorpinel:
@ryokugyu still around/interested?

ryokugyu:
@jorgeorpinel Still around 😄 Was busy with placement session, got placed today only! 😺  Let's work on this.

jorgeorpinel:
[...]

PR #3359, opened 2022-03-15
ref: `config` reads from all config files combined

jorgeorpinel:
per https://github.com/iterative/dvc.org/issues/1368#issuecomment-634456922


dberenbaum (approved):
No problems with the proposed changes. I'm not sure they do enough to address the issue or explain how config sections get merged, but maybe that's left to a future PR?

jorgeorpinel:
Yeah we can expand on this when we address https://github.com/iterative/dvc.org/issues/340.

Issue #3134, opened 2021-12-28
install: `pip` method (silently) requires Python 3.7


jorgeorpinel:
It probably makes sense for us to make it clearer what the minimum supported python version is, it's currently only stated in the venv/pipx note:

> We strongly recommend creating a virtual environment or using pipx (on Python 3.7+) to encapsulate your local environment.

It would be better if it was noted more explicitly that 3.7+ is a hard requirement.

But beyond that, IMO it should be safe for us to assume that anyone choosing the "install via pip" method (as opposed to using the binary or conda packages) should already understand how to set up a python/pip environment on their machine...

Users that aren't familiar with managing python themselves will probably be using the system packages (or conda) which handles all of that for them.

_Originally posted by @pmrowla in https://github.com/iterative/dvc/issues/7181#issuecomment-1001956032_

jorgeorpinel:
p.s. by "silently" I mean that if you have Python 3.6 `pip install dvc` will work but you'll get the last version which was supported on Python 3.6 (which is DVC 2.8.1) and you may not even notice so, or if you do you won't necessarily know why that happened.

pmrowla:
> p.s. by "silently" I mean that if you have Python 3.6 `pip install dvc` **will work** but you'll get the last version which was supported on Python 3.6 (which is DVC 2.8.1) and you may not even notice so, or if you do you won't necessarily know why that happened.

There's no way for us to really do anything about this in DVC though. 2.8.1 is a valid DVC version for pip to install in python 3.6, so there's nothing to really warn the user about.

jorgeorpinel:
> no way for us to really do anything about this

We can document it 🙂

Issue #2449, opened 2021-05-06
install: missing `[webdav]` and `[hdfs]` from `pip` instructions


natephysics:
In the following documentation, there's a missing option for pip install dvc[webdav]. 

https://github.com/iterative/dvc.org/blob/master/content/docs/install/linux.md

https://github.com/iterative/dvc.org/blob/master/content/docs/install/windows.md

shcheklein:
Yes, we have some extras that are not documented: 

https://github.com/iterative/dvc/blob/master/setup.py#L131

jorgeorpinel:
Does this also apply to `conda`? Cc @efiop thanks

efiop:
@jorgeorpinel webdav is not included into conda at all, as there is no corresponding dependency package there. We'll likely just wait for webdav4 and then introduce a conda package for it too.

Issue #2112, opened 2021-01-24
install: Minimum `pip` version should be specified at 19.0+ for dev envs


gth828r:
Hello, and thanks for the great project and excellent documentation!

I just walked through the development environment setup on Ubuntu 18.04 and I ran into some issues that prevented the environment from successfully being set up when following the documented instructions. I traced it down to a version requirement in pygit2 (see https://github.com/libgit2/pygit2/issues/999#issuecomment-614691503 and https://www.pygit2.org/install.html) that pip needs to be at version 19.0+. In my case, I was using the default Ubuntu 18.04 pip version, which is woefully behind that.

In my case, I simply ran the following in my virtual environment to get things back on track:
```
pip install --upgrade pip
```

The minimum pip required version of 19.0 should likely be specified in the development environment setup document in the user guide. Optionally, the above command could be recommended to arrive at an appropriate version.

shcheklein:
@gth828r thanks! good catch, we should mention it as a note in docs indeed

@efiop just to let you know about this, may be something should be done on the DVC end - e.g. can we require a specific pip version?

efiop:
@shcheklein I don't think we can, as pip isolates itself from the env of setup.py(or requirements) when installing stuff. Let's just add a note to the docs, same as pygit2 guys. We might need to add it to the docs for regular installation after 2.0 release too.

shcheklein:
@efiop sounds good, thanks!

jorgeorpinel:
Added note in https://github.com/iterative/dvc.org/pull/3360. Hoping pip 19 is still the minimum?

gth828r:
According to the [pygit2 documentation](https://www.pygit2.org/install.html) there is now some type of dependency on pip 20.3+:
> This requires pip 20.3 or later, otherwise pip will try to install the source distribution, and likely fail to find a matching version of libgit2.

I tried it out, and I believe **you need at least pip 20.3** on a fresh dvc dev environment install to get all of the dvc unit tests to pass.

When installing from pip 20.0.2 (default on Ubuntu 20.04), I saw a lot of red text and some warning messages at the end of the attempted pip install. The unit test summary was as follows:
```txt
18 failed, 2450 passed, 14 skipped, 2 xfailed, 3 warnings in 180.76s (0:03:00)
```

I then uninstalled dvc, upgraded pip to the latest (taking me to version to 22.0.4), and reinstalled dvc. The unit test summary:
```txt
[...]

PR #3360, opened 2022-03-15
install: missing `pip` optional paks

jorgeorpinel:
Closes https://github.com/iterative/dvc.org/issues/2449
Closes https://github.com/iterative/dvc.org/issues/2112
Closes https://github.com/iterative/dvc.org/issues/3134

### DIFF
--- a/content/docs/install/linux.md
+++ b/content/docs/install/linux.md
@@ -16,13 +16,9 @@ $ pip install dvc
 ```
 
 Depending on the type of the [remote storage](/doc/command-reference/remote) you
-plan to use, you might need to install optional dependencies: `[s3]`, `[azure]`,
-`[gdrive]`, `[gs]`, `[oss]`, `[ssh]`. Use `[all]` to include them all.
-
-> Please restart your terminal or re-source the shell configuration file
-> (`.bashrc`, `.zshrc`, etc.) if you get `Command 'dvc' not found` after
-> installation. If restarting the terminal isn't desirable (e.g. containers),
-> it's also possible to manually edit the `PATH` env variable.
+plan to use, you might need to install optional dependencies: `[s3]`,
+`[gdrive]`, `[gs]`, `[azure]`, `[ssh]`, `[hdfs]`, `[webdav]`, `[oss]`. Use
+`[all]` to include them all.
 
 <details>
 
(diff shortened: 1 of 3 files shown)


jorgeorpinel:
Just added the ones missing from `all` here: https://github.com/iterative/dvc/blob/main/setup.cfg#L80

Is that correct? I see other optional deps further down such as `ssh_gssapi` and `webhdfs`.

content/docs/install/linux.md
```diff
@@ -9,20 +9,17 @@
 > [virtual environment](https://python.readthedocs.io/en/stable/library/venv.html)
 > or using
 > [pipx](https://packaging.python.org/guides/installing-stand-alone-command-line-tools/)
-> (on Python 3.7+) to encapsulate your local environment.
+> (on Python 3.7+) to encapsulate your local environment.  
+> Note that `pip` version 19+ is required.
```
jorgeorpinel, line 13: Also, is pip 19 still the min. required here? Per https://github.com/iterative/dvc.org/issues/2112

jorgeorpinel: I was confused, #2112 related to setting up a dev env. I rolled this back. Or should we specify a `pip` version? I remember regular `pip` installation failing with the default `pip` in Ubuntu before because it was pretty old.

content/docs/user-guide/contributing/core.md
```diff
@@ -42,11 +42,12 @@ Get the latest development version. Fork and clone the repo:
...
 ```
 
-Make sure that you have Python 3 installed. Version 3.7 or higher is required to
-run style checkers on pre-commit. On macOS, we recommend using `brew` to install
-Python. For Windows, we recommend an official
+Make sure that you have Python 3.7 or higher installed. On macOS, we recommend
+using `brew` to install Python. For Windows, we recommend an official
 [python.org release](https://www.python.org/downloads/windows/).
 
+> ℹ️ Note that `pip` version 20.3+ is required.
```
jorgeorpinel, line 49: I think it's 20.3 due to https://www.pygit2.org/install.html but if it's another version lmk.

pmrowla: That should be correct I think

pmrowla (approved):

PR #3365, opened 2022-03-16
Use `rehype-react` to render blog post content

### LINKED ISSUE #3076: Use same/similar engine for blog and documentation
Currently, the blog and docs have different engines (or they generate the final html in different ways). Should we look into having the blog use the same engine (or an engine similiar) to the docs engine? Not only would this simplify our markdown engine code, it would also be easier to give the blog custom components like the `external-link`. cc @rogermparent

julieg18:
Fixes #3076 (and allows us to work on #3153)

### DIFF
--- a/src/components/Blog/Post/Markdown/index.tsx
+++ b/src/components/Blog/Post/Markdown/index.tsx
@@ -1,13 +1,21 @@
 import React from 'react'
+import rehypeReact from 'rehype-react'
 
 import * as styles from './styles.module.css'
 
 interface IMarkdownProps {
-  html: string
+  htmlAst: Node
 }
 
-const Markdown: React.FC<IMarkdownProps> = ({ html }) => (
-  <div className={styles.wrapper} dangerouslySetInnerHTML={{ __html: html }} />
-)
+// Rehype's typedefs don't allow for custom components, even though they work
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const renderAst = new (rehypeReact as any)({
+  createElement: React.createElement,
+  Fragment: React.Fragment
+}).Compiler
+
+const Markdown: React.FC<IMarkdownProps> = ({ htmlAst }) => {
+  return <div className={styles.wrapper}>{renderAst(htmlAst)}</div>
+}
 
 export default Markdown
--- a/src/components/Blog/Post/index.tsx
+++ b/src/components/Blog/Post/index.tsx
@@ -22,7 +22,7 @@ import SubscribeSection from '../../SubscribeSection'
 import * as styles from './styles.module.css'
 
 const Post: React.FC<IBlogPostData> = ({
-  html,
+  htmlAst,
   timeToRead,
   title,
   date,
@@ -38,7 +38,6 @@ const Post: React.FC<IBlogPostData> = ({
   const wrapperRef = useRef<HTMLDivElement>(null)
   const { width, height } = useWindowSize()
   const { y } = useWindowScroll()
-
   useCustomYtEmbeds()
 
   const isFixed = useMemo(() => {
@@ -92,7 +91,7 @@ const Post: React.FC<IBlogPostData> = ({
           />
 
           <div className={styles.content}>
-            <Markdown html={html} />
+            <Markdown htmlAst={htmlAst} />
           </div>
           {tags && (
             <div className={styles.tags}>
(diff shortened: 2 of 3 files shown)


src/components/Blog/Post/Markdown/index.tsx
```diff
@@ -1,13 +1,21 @@
...
-)
+// Rehype's typedefs don't allow for custom components, even though they work
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const renderAst = new (rehypeReact as any)({
+  createElement: React.createElement,
+  Fragment: React.Fragment
+}).Compiler
+
+const Markdown: React.FC<IMarkdownProps> = ({ htmlAst }) => {
+  return <div className={styles.wrapper}>{renderAst(htmlAst)}</div>
```
julieg18, line 18: Instead of using `html`, we use `htmlAst` allowing us to add custom components to blog like we do with the doc.

julieg18:
> I'm a little surprised how well everything just drops in and works, I made sure to check the html-centric external-link-plugin used on [heartbeat articles](https://dvc-org-use-rehype-reac-egxrsl.herokuapp.com/blog/february-22-heartbeat) and they seem to work perfectly fine!

Same here! I was thinking we were going to have to rewrite how the blog nodes were made but nope, everything looks good!

src/components/Blog/Post/Markdown/index.tsx
shcheklein: @rogermparent @julieg18 does it mean that we now do more JS work on the client side + potentially have to bring more dependencies?

julieg18: No, I don't think so. @rogermparent, correct me if I'm wrong, but I believe HTML rendering is all done during build time, whether we use `html` or `htmlAst`.

shcheklein: yep, related question also here is what is being prefetched / fetched when we hover the side panel in the docs and eventually click to get the new page - rendered HTML or some tree that still requires post-processing?

rogermparent: The answer is between the two:

- Initial page load is fully compiled HTML that is then rehydrated
- Subsequent page loads are done within Gatsby's React SPA, whether we use html or rehype

So the blogs would be using JS a little more, but regardless of the html content being an HTML string or Rehype we're already in SPA world so the difference is negligible. The only extra library pulled in is `rehype-react`, which would already be pulled in when viewing a doc and is 9kb minzipped and 20kb unpacked.

I'll run some tests and inspect the browser data, between this and the prior commit without `rehype` blogs, I've run them before but the difference is negligible.

rogermparent: Weirdly enough, rehype delivers less JS on average according to lighthouse.

rehype:
image

html:
image

shcheklein: thanks Roger!

> but regardless of the html content being an HTML string or Rehype we're already in SPA world so the difference is negligible.

agreed, but I think the initial decision behind using the `dangerousHTML` call was specifically to avoid this slippery slope of bringing more and more post processing into client (e.g. that's why we were "complicating" our plugins to operate build time vs runtime as far as I remember). Ideally we should probably see CPU load, etc (does it become less "snappy" for example on large pages, etc).

> which would already be pulled in when viewing a doc

I guess initially docs engine was using dangerous html as well, right? Not sure what was the reason to migrate it to AST.

No strong opinion on this- you folks decide based on data. Just sharing the context about those decisions the way I remember them.

rogermparent: `page-data.json` with rehype (there's also an html string in there, which we should remove since it's ~20kb of data we're not using)
image

`page-data.json` with HTML
image

I've copied the response strings into files to compare, and the rehype page-data saves about 12kb on the same response on the same page.

rogermparent: @shcheklein that's definitely a valid concern, but from what I can tell the cost/benefit analysis favors using `htmlAst` in both engines over using `html` in the blog. As far as I understand, the reason the docs switched to `htmlAst` is because the custom components used there (particularly `Abbr` tooltips since the other complex components were introduced after) would be incredibly difficult to build and maintain in a pure HTML engine.

shcheklein: @rogermparent 

> would be incredibly difficult to build and maintain in a pure HTML engine.

yep, it might be. Can we find when it was done out of curiosity?

PR #3368, opened 2022-03-17
Update blog post graphql query

julieg18:
* takes off "html" from the page query, as it is not being used anywhere


yathomasi (approved):
LGTM.

PR #3369, opened 2022-03-17
Add `safeLocalStorage` util

julieg18:
* adds a util that checks if localStorage is supported before using it
* hopefully will clear the `SecurityError` that sentry caught

### DIFF
--- a/plugins/gatsby-theme-iterative-docs/src/utils/front/safeLocalStorage.ts
+++ b/plugins/gatsby-theme-iterative-docs/src/utils/front/safeLocalStorage.ts
@@ -1,0 +1,20 @@
+const IS_SUPPORTED = ((): boolean => {
+  try {
+    const key = 'localstoragecheck'
+    localStorage.setItem(key, key)
+    localStorage.removeItem(key)
+    return true
+  } catch (e) {
+    return false
+  }
+})()
+
+const safeLocalStorage = {
+  clear: () => IS_SUPPORTED && localStorage.clear(),
+  getItem: (key: string) => IS_SUPPORTED && localStorage.getItem(key),
+  setItem: (key: string, value: string) =>
+    IS_SUPPORTED && localStorage.setItem(key, value),
+  removeItem: (key: string) => IS_SUPPORTED && localStorage.removeItem(key)
+}
+
+export default safeLocalStorage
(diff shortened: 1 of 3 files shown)


plugins/gatsby-theme-iterative-docs/src/utils/front/safeLocalStorage.ts
```diff
@@ -0,0 +1,20 @@
+const IS_SUPPORTED = ((): boolean => {
+  try {
+    const key = 'localstoragecheck'
+    localStorage.setItem(key, key)
+    localStorage.removeItem(key)
+    return true
+  } catch (e) {
+    return false
+  }
+})()
```
yathomasi, line 10: maybe simply do this?
```js
return typeof(Storage) !== "undefined"
```

plugins/gatsby-theme-iterative-docs/src/utils/front/useCustomYtEmbeds.ts
```diff
@@ -24,7 +25,7 @@ const setUpEmbedClickListeners = (embeds: NodeListOf<Element>) => {
         iframe.src = iframe?.src + `&autoplay=1`
       }
       hideAllEmbedOverlays(embeds)
-      localStorage.setItem('yt-embed-consent', 'true')
+      safeLocalStorage.setItem('yt-embed-consent', 'true')
```
yathomasi, line 28: or maybe use optional chaining(`?.`) instead?

plugins/gatsby-theme-iterative-docs/src/utils/front/safeLocalStorage.ts
julieg18: I think the issue isn't about the fact that `Storage` doesn't exist. It's about how attempting to use `Storage` will cause an error. 🤔 Which is why the function check actually attempts to use it.

plugins/gatsby-theme-iterative-docs/src/utils/front/useCustomYtEmbeds.ts
julieg18: As mentioned in the previous comment, the issue isn't that `localStorage` doesn't exist, it's about that using it will cause a `SecurityError` :)

yathomasi (approved):
This works @julieg18 🎉 .
I was able to reproduce it on safari by disabling all the cookies on our current website.
This solves the problem.

PR #3378, opened 2022-03-22
Fix typo in data-registries.md

alex000kim:
I assume in this example it should say `songs` instead of `images`.


jorgeorpinel (approved):
Huh, that seems right. Thanks!

Issue #1909, opened 2020-11-05
metrics: document behavior of accepting any parsable file


pared:
After iterative/dvc#4446 `dvc metrics` commands accept any viable target, metric file does not have to be specified in `dvc.yaml`.
`plots` has this functionality already documented it #1800.
Related to #1809, iterative/dvc#4761

jorgeorpinel:
Is this still relevant @pared @dberenbaum ? Thanks

pared:
@jorgeorpinel yes, it is
in https://dvc.org/doc/command-reference/metrics#description we imply that one needs to mark the output as `metric` in order to use `dvc metrics` command on it. This is not a requirement. I don't think that's super important to mention that it works on files not specified as metrics, though it would be good to mention it. It shouldn't be a big change, sorry for leaving it to hang for so long.

jorgeorpinel:
Looks like this is already mostly done per https://github.com/iterative/dvc.org/pull/3387#pullrequestreview-924066307 but thanks for #3387 as well! Lmk if you need help finishing that one up. Closing this.

PR #3387, opened 2022-03-24
metrics: info on non-output usage

pared:
Per https://github.com/iterative/dvc/pull/5339

Closes: #1909

### DIFF
--- a/content/docs/command-reference/metrics/index.md
+++ b/content/docs/command-reference/metrics/index.md
@@ -47,8 +47,10 @@ Path          Metric    HEAD      workspace  Change
 metrics.json  AUC       0.763981  0.801807   0.037826
 ```
 
-`dvc metrics` subcommands by default use the metrics files specified in
-`dvc.yaml` (if any), for example `summary.json` below:
+`dvc metrics` subcommands can be used on any
+[valid metrics files](#supported-file-formats). By default they use the metrics
+files specified in `dvc.yaml` (if there are any), for example `summary.json`
+below:
 
 ```yaml
 stages:
--- a/content/docs/command-reference/metrics/index.md
+++ b/content/docs/command-reference/metrics/index.md
@@ -48,8 +48,8 @@ metrics.json  AUC       0.763981  0.801807   0.037826
 ```
 
 `dvc metrics` subcommands can be used on any
-[valid metrics files](#supported-file-formats). By default they use the metrics
-files specified in `dvc.yaml` (if there are any), for example `summary.json`
+[valid metrics files](#supported-file-formats). By default they use the ones
+specified in `dvc.yaml` (if any), for example `summary.json`
 below:
 
 ```yaml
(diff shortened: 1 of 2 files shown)


content/docs/command-reference/metrics/index.md
```diff
@@ -63,10 +63,15 @@ stages:
...
-> information on the file format above.
+> Notes:
+>
+> - `cache: false` above specifies that `summary.json` is not tracked or
+>   <abbr>cached</abbr> by DVC (`-M` option of `dvc stage add`). These metrics
+>   files are normally committed with Git instead. See `dvc.yaml` for more
+>   information on the file format above.
+> - even though `dvc metrics` subcommands use the metrics files specifed in
+>   `dvc.yaml` by default, they can be used on any file, as long as they are
+>   specified via `targets` argument
```
jorgeorpinel, line 74: Hmmmm. I see. I think instead of changing this note we should incorporate the information to the paragraph before the `summary.json` sample block.

jorgeorpinel: I.e. instead of 

```markdown
`dvc metrics` subcommands by default use the metrics files specified in
`dvc.yaml` ...
```

We should have something like

```markdown

`dvc metrics` subcommands can be used on any
[valid metrics files](#supported-file-formats). By default they use the one
specified in `dvc.yaml` ...
```

jorgeorpinel (changes requested):
BTW I see that this info. is already in https://dvc.org/doc/command-reference/metrics/show and diff 🙂  (around `targets` explanation). So we should only address the point above here ☝🏼

content/docs/command-reference/metrics/index.md
```diff
@@ -47,8 +47,10 @@ Path          Metric    HEAD      workspace  Change
 metrics.json  AUC       0.763981  0.801807   0.037826
 ```
 
-`dvc metrics` subcommands by default use the metrics files specified in
-`dvc.yaml` (if any), for example `summary.json` below:
+`dvc metrics` subcommands can be used on any
+[valid metrics files](#supported-file-formats). By default they use the metrics
+files specified in `dvc.yaml` (if there are any), for example `summary.json`
```
pared, line 52: I changed `if any` tp `if there are any` - sounded better to me, but it might be just me, I can roll this back :)

content/docs/command-reference/metrics/show.md
```diff
@@ -27,8 +27,7 @@ target can even be directories, so that DVC recursively shows all metrics files
 inside.
 
 > Note that targets don't necessarily have to be defined in `dvc.yaml`. For that
-> reason, this command doesn't require an existing DVC project to run in; It
-> works in any Git repo.
+> reason, this command doesn't require an existing DVC project to run in.
```
pared, line 30: Git repo is not necessary, removing this mention (only in `show`, because in diff we actually need it). Is removal ok, or should I explicitly state that for `show` Git is not necessary?

content/docs/command-reference/metrics/index.md
jorgeorpinel: Nice, thanks. This works perfectly in terms of meaning. I'll rephrase the wording a little for readability.

content/docs/command-reference/metrics/show.md
jorgeorpinel: Hmmm. I guess it's best to be explicit just in case:
```suggestion
> reason, this command doesn't require an existing DVC project or Git repo to run in.
```

Issue #3383, opened 2022-03-24
details: unique anchors shouldn't change with content order


jorgeorpinel:
> Follow-up to https://github.com/iterative/dvc.org/issues/3314

Per https://github.com/iterative/dvc.org/pull/3329#issuecomment-1077311398, the way unique anchors are generated for `<details>` blocks (based on the first child's content + an incremental index if needed) can be problematic. Eg. a block is no 1 in the page, someone links to it, and later the content get's rearranged such that the block is now no 2. The link is now incorrect (goes to a previous block, not the one we want).

Maybe we should require an `id` or some other prop that's supposed to be unique and use it in the anchor somehow. This way they're predictable and stable.

Additionally, should we use a `title` prop for the anchor base instead of the first child's text?

jorgeorpinel:
p.s. not a high priority probably but it could eventually cause super low-visibility bugs. For now let's try not to repeat headers in `<details>` blocks! Cc @iesahin

jorgeorpinel:
> should we use a title prop for the anchor base instead of the first child's text?

Can you put emojis in `title` though? 😅 

![image](https://user-images.githubusercontent.com/1477535/159864995-1fc65298-9884-4cbc-9c5d-71c4e74b899b.png)

yathomasi:
> Maybe we should require an `id` or some other prop that's supposed to be unique and use it in the anchor somehow. This way they're predictable and stable.

#3352 was for the approach mentioned but was closed in favor of #3329. The considerations were [mentioned before the merge](https://github.com/iterative/dvc.org/pull/3329#issuecomment-1070953981). 
We made the assumption such changes within a page will be **rare** and also we can try to make the titles unique on their own.

> Additionally, should we use a `title` prop for the anchor base instead of the first child's text?

Yes, sure it will be consistent with other blocks as well. I will be submitting a pr on this.

> Can you put emojis in title values though? 😅

Yes. It should be working just fine. 

 PS: if we are still thinking, we should be using `id` for making the link unique then we should proceed with the idea early because we also wouldn't want the current links with numbering also expire don't work.

jorgeorpinel:
> The considerations were https://github.com/iterative/dvc.org/pull/3329#issuecomment-1070953981.

Yes but as you can see my realization and [reply to that](https://github.com/iterative/dvc.org/pull/3329#issuecomment-1077311398) happened after the merge 🙂  thus this issue.

> We made the assumption such changes within a page will be rare

[...]

PR #3388, opened 2022-03-24
Allowed title as props for `<details>`

yathomasi:
From #3383 
> Additionally, should we use a `title` prop for the anchor base instead of the first child's text?

Used title as props for `<details>` tag. 

Initially, I only used `title` as props but there are 2-3 particular cases that use `code` highlighting in the title. 
Screen Shot 2022-03-24 at 19 12 52
So, I also added back the first child's text as a fallback if no title is passed as props so that we still can use markdown as well.
Also, previously in this particular case, the code section seems to be parsed as `object object` as well 
resulting: https://dvc.org/doc/user-guide/external-dependencies#expand-to-see-resulting-object-object-file
This issue should also be fixed with the pr.

But, if markdown syntax is not so important and the title would be enough then the code implementation of the Details component will be very simple and preferable.


yathomasi:
> question if this approach breaks some titles and we have to maintain two options, why do we consider it superior and why do? let's keep the previous one?

markdown wouldn't be supported when passed as props but I think it would be okay without it too as it's used in just 2 titles to highlight `.dvc`. 
And, using `title` as props should make consistent with other markdown custom elements as they also support through props.

> Also, previously in this particular case, the code section seems to be parsed as `object object` as well
> resulting: https://dvc.org/doc/user-guide/external-dependencies#expand-to-see-resulting-object-object-file
> This issue should also be fixed with the pr.

I think this one needs to be addressed and which we didn't address on pervious pr.

rogermparent (commented):
Not a bad idea, but I think it would be better to separate the bugfixes like the `object object` thing and the prop-title feature into different PRs, the former is an easy fix we'll want asap while the latter needs some more discussion.

content/docs/command-reference/checkout.md
```diff
@@ -112,9 +112,7 @@ pipeline stages, such as the <abbr>DVC project</abbr> created for the
 [Get Started](/doc/start). Then we can see what happens with `git checkout` and
 `dvc checkout` as we switch from tag to tag.
 
-<details>
-
-### Click and expand to set up the project
+<details title="Click and expand to set up the project>
```
jorgeorpinel, line 115: Does this PR update all existing `<details>` blocks? Thanks

jorgeorpinel:
Ah shoot, due to #3391 there are some merge conflicts to solve now... ⌛

shcheklein:
Reminder:

> Still, I don't see any clear motivation for this tbh. And still it breaks things in some cases. So, my take - let's postpone / keep it as-is for now.

can we skip addressing the conflict? 

please please let's be mindful about the bigger priorities

content/docs/command-reference/import-url.md
```diff
@@ -168,9 +168,7 @@ produces a regular stage in `dvc.yaml`.
 To illustrate these examples we will be using the <abbr>project</abbr> explained
 in the [Get Started](/doc/start).
 
-<details>
-
-### Click and expand to set up example
+<details title="Click and expand to set up example">
```
jorgeorpinel, line 171: OK just noticed we're postponing this. Just a note that per #3391 these titles should no longer start in "Click for" nor end in "example". Thanks

jorgeorpinel:
Given this is not much of a priority + all the conflicts, should we close it or turn it into an issue? Thanks

PR #3371, opened 2022-03-18
Add admonitions feature to blog/docs

julieg18:
code:

```html
<admon type="warn">

text text text

</admon>
```

output:

image

Fixes #3153

PR #3389, opened 2022-03-24
Update admonitions, making `title` optional

### LINKED ISSUE #3381: admonitions titles take too much precious space
Screen Shot 2022-03-23 at 2 17 25 PM

It's 40% of the whole item. I would say icon alone should be enough. Or show the tooltip when you hover.

julieg18:
* makes title optional, taking it off by default

Code Example:

```
<admon>

I don't need a title since I'm just a block of text.

</admon>

<admon title="Custom Title" type="tip">

I need a custom title for some reason

</admon>
```

Result:
image

Fixes #3381

### DIFF
--- a/content/docs/index.md
+++ b/content/docs/index.md
@@ -5,6 +5,26 @@ Data Version Control, or DVC, is a data and ML
 advantage of the existing engineering toolset that you're already familiar with
 (Git, CI/CD, etc.).
 
+<admon>
+
+I don't need a title since I'm just a block of text.
+
+</admon>
+
+<admon title type="warn">
+
+- I need
+- a title
+- since I'm a list
+
+</admon>
+
+<admon title="Custom Title" type="tip">
+
+I need a custom title for some reason
+
+</admon>
+
 <cards>
 
   <card href="/doc/start" heading="Get Started">
--- a/content/docs/index.md
+++ b/content/docs/index.md
@@ -5,26 +5,6 @@ Data Version Control, or DVC, is a data and ML
 advantage of the existing engineering toolset that you're already familiar with
 (Git, CI/CD, etc.).
 
-<admon>
-
-I don't need a title since I'm just a block of text.
-
-</admon>
-
-<admon title type="warn">
-
-- I need
-- a title
-- since I'm a list
-
-</admon>
-
-<admon title="Custom Title" type="tip">
-
-I need a custom title for some reason
-
-</admon>
-
 <cards>
 
   <card href="/doc/start" heading="Get Started">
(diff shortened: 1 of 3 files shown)


content/docs/index.md
```diff
@@ -5,6 +5,26 @@ Data Version Control, or DVC, is a data and ML
 advantage of the existing engineering toolset that you're already familiar with
 (Git, CI/CD, etc.).
 
+<admon>
```
julieg18, line 8: Similar to #3371, this pr will remain a draft until this test is taken off :)

rogermparent (commented):
Looks nice! I'm not sure if this is exactly what @shcheklein was thinking, but the option for an alternate minimal admonition seems pretty useful. I could see possible concerns around consistency between the two, though.

julieg18:
> Looks nice! I'm not sure if this is exactly what @shcheklein was thinking, but the option for an alternate minimal admonition seems pretty useful. I could see possible concerns around consistency between the two, though.

We could take off the title entirely! I wasn't sure which (asked in #3381), but for now, I made it an option 🤔

daavoo:
I think the title might be useful on some occasions.

content/docs/index.md
julieg18: Taking off tests since we now have `admon` examples in the docs DVCLive pages!

iesahin:
Could we make the titles similar to `<details>` block? Instead of adding it as an attribute, we could update the text itself like:

```
<admon type="tip">

## Very important tip

This is very important

</admon>
```

This is for the sake of consistency. I don't see the current way as particularly useless and if this is too much work, we can postpone it.

jorgeorpinel:
> make the titles similar to <details> block? 

We're actually trying to make `<details>` work like `<admon>` instead @iesahin 😅 . See https://github.com/iterative/dvc.org/pull/3388

iesahin:
Ah yeah, I noticed that after writing this. I think it's better to have the titles as first elements as in current `details`, it makes it easier to change `details/admon` to a standard section.

For linking there may be a `slug` or `anchor` attribute in the tag apart from the title. We can use it for linking. This also relaxes us to come up with very short titles for these sections.

julieg18:
> Leaving title optional can't hurt. What I'm not sure we need is the default titles e.g. "Warning" in the description usage example. I think if title is used, a value should be provided, otherwise the syntax looks a bit strange to me.

Good point! I'll take off the default title option! 

As for using a `title` attribute vs taking a `###` heading from the content, both are easily doable. I'm not sure which one is better...

jorgeorpinel:
> For linking there may be a slug or anchor attribute
discuss the title vs ### heading in a separate issue

We're discussing that in https://github.com/iterative/dvc.org/issues/3383 , feel free to mention your idea there @iesahin  (may be the easiest way indeed).

jorgeorpinel (approved):
Assuming default `title` has been removed. Thanks @julieg18

PR #3393, opened 2022-03-25
Fix for edge cases for `id`

yathomasi:
Fix for titles that use the markdown.
Links with issue: 
- Trialing dash(`-`)
  - https://dvc.org/doc/command-reference/remote#what-is-a-local-remote-
- `object-object`
  - https://dvc.org/doc/user-guide/how-to/stop-tracking-data#expand-to-add-a-sample-data-object-object-file
  - https://dvc.org/doc/user-guide/external-dependencies#expand-to-see-resulting-object-object-file
  - https://dvc.org/doc/user-guide/external-dependencies#expand-to-see-resulting-object-object-file-1

### DIFF
--- a/plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
+++ b/plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
@@ -50,11 +50,18 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
     firstChild.props.children.length - 1
   ) as ReactNode[]
 
-  let slug = slugger.slug(triggerChildren.toString())
-  if (slug[0] === '️') {
-    slug = slug.slice(1)
-  }
-  const id = slug.startsWith('-') ? slug.slice(1) : slug
+  const title = (triggerChildren as any[]).reduce((acc, cur) => {
+    return (acc +=
+      typeof cur === 'string'
+        ? cur
+        : typeof cur === 'object'
+        ? cur?.props?.children?.toString()
+        : '')
+  }, '')
+
+  let slug = slugger.slug(title)
+  slug = slug.startsWith('-') ? slug.slice(1) : slug
+  const id = slug.endsWith('-') ? slug.slice(0, -1) : slug
 
   useEffect(() => {
     if (location.hash === `#${id}`) {


plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,18 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
     firstChild.props.children.length - 1
   ) as ReactNode[]
 
-  let slug = slugger.slug(triggerChildren.toString())
-  if (slug[0] === '️') {
-    slug = slug.slice(1)
-  }
-  const id = slug.startsWith('-') ? slug.slice(1) : slug
+  const title = (triggerChildren as any[]).reduce((acc, cur) => {
```
julieg18, line 53: Not sure why, but links with ⚙️ are now not working correctly: 

https://dvc-org-fix-details-obj-yq7a0c.herokuapp.com/doc/start/experiments#%EF%B8%8F-run-multiple-experiments-in-parallel

And it looks like both this branch and `dvc.org` details links don't work with  ℹ️:

https://dvc-org-fix-details-obj-yq7a0c.herokuapp.com/doc/start/experiments#%E2%84%B9%EF%B8%8F-more-information-about-metrics

julieg18 (commented):
Great job catching the edge cases!

yathomasi:
Github slugger is not slicing the ℹ️ emoji.
On: `/doc/start/experiments`
Link: 
⚠️ may expire later:
https://dvc-org-fix-details-obj-yq7a0c.herokuapp.com/doc/start/experiments#%E2%84%B9-more-information-about-hyperparameters

Other emojis are sliced but not this one. This can be fixed but now wondering if there could be other emojis as well.

julieg18:
> Github slugger is not slicing the ℹ️ emoji.
> Other emojis are sliced but not this one. This can be fixed but now wondering if there could be other emojis as well.

If `github-slugger` is having issues slicing emojis, maybe we should write some code that takes off any emojis before we give the link to `github-slugger`?

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,21 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
...
+        ? cur
+        : typeof cur === 'object'
+        ? cur?.props?.children?.toString()
+        : '')
+  }, '')
+
+  let slug = slugger.slug(title)
+  if (slug.includes('️')) {
+    slug = slug.replaceAll('️', '')
   }
```
rogermparent, line 65: Does this do anything? As far as I understand from memory and a few quick tests, `string.includes('')` is always true, and `string.replaceAll('','')` always returns the string unchanged.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
```diff
@@ -50,11 +50,21 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
...
-  const id = slug.startsWith('-') ? slug.slice(1) : slug
+  slug = slug.startsWith('-') ? slug.slice(1) : slug
+  const id = slug.endsWith('-') ? slug.slice(0, -1) : slug
```
rogermparent, line 67: Maybe a regex-based replacement would be good in order to catch multiple dashes on the ends of the string, but admittedly it's a little unlikely that titles with the current format would do so.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
yathomasi: There was an issue that made me scratch my head. It's with the `github-slugger` again. 
If you look into this you can see highlighting on my vscode snippet. It's not empty string but a character. 
Screen Shot 2022-03-29 at 11 33 18

> Not sure why, but links with ⚙️ are now not working correctly:
> 
> https://dvc-org-fix-details-obj-yq7a0c.herokuapp.com/doc/start/experiments#%EF%B8%8F-run-multiple-experiments-in-parallel

This means `github-slugger` generates that string for emoji and is not visible to the eye. So, that was the fix needed.

yathomasi:
> If github-slugger is having issues slicing emojis, maybe we should write some code that takes off any emojis before we give the link to github-slugger?

> Maybe a regex-based replacement would be good in order to catch multiple dashes on the ends of the string, but admittedly it's a little unlikely that titles with the current format would do so.

Since we have problems with the `github-slugger` and need to come around with fixes after using it as well. Should we consider using our own version and it can be simple like [slugger when we proposed a custom one](https://github.com/iterative/dvc.org/blob/e7d3b2ee6649b9b3d893db2631a1b0bac3f046b3/plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts#L38)?

cc: @iterative/websites

rogermparent:
> Since we have problems with the github-slugger and need to come around with fixes after using it as well. Should we consider using our own version 

I guess since we're having to work around `github-slugger`, at least using something else is in order. Probably something similar to the original custom slugger but with the `code` and `position` system replaced with an `id` override.

yathomasi:
> If `github-slugger` is having issues slicing emojis, maybe we should write some code that takes off any emojis before we give the link to `github-slugger`?

Also, I believe https://github.com/Flet/github-slugger/blob/master/regex.js is supposed to filter all the emojis.

julieg18:
> Since we have problems with the github-slugger and need to come around with fixes after using it as well. Should we consider using our own version and it can be simple like [slugger when we proposed a custom one](https://github.com/iterative/dvc.org/blob/e7d3b2ee6649b9b3d893db2631a1b0bac3f046b3/plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts#L38)?

> I guess since we're having to work around github-slugger, at least using something else is in order. Probably something similar to the original custom slugger but with the code and position system replaced with an id override.

Agreed that using a custom slugger would be best if `github-slugger` isn't suiting our use case and using an `id` override instead of `code`/`position` props!

julieg18 (approved):
Everything looks fixed now!

Issue #3154, opened 2022-01-06
Add hidden `skip to content` navigation for directly skipping to content on keyboard navigation.


yathomasi:
Since we are taking keyboard navigation into account, we can also discuss on adding hidden `skip to content` navigation. 

We can find this implemented on most of the sites. An example is this Github itself, you can see if you start keyboard navigation by pressing `tab` button.
A similar implementation can be found on Reddit, Youtube, CSS-tricks, dev.to, and so on.

_Originally posted by @yathomasi in https://github.com/iterative/dvc.org/issues/3150#issuecomment-1006502094_

Maybe we can create a pluggable component and use it on other website projects as well.

@iterative/websites @iterative/docs

PR #3395, opened 2022-03-25
Added hidden `skip to content` for keyboard navigation

yathomasi:
Added hidden `skip to content`, that can be used by keyboard users for directly navigating to main content. 

Since the `docs` have side navigation as well, I thought it would be better to skip that as well. 

There seems some interference with how we currently handle the scrolling hash link.

Related to #3154

### DIFF
--- a/src/components/MainLayout/index.tsx
+++ b/src/components/MainLayout/index.tsx
@@ -1,4 +1,6 @@
 import React, { useEffect } from 'react'
+import { SkipNavContent, SkipNavLink } from '@reach/skip-nav'
+import '@reach/skip-nav/styles.css'
 
 import { IPageProps } from '../Page'
 import LayoutHeader from '../LayoutHeader'
@@ -28,7 +30,8 @@ export type LayoutComponent = React.FC<
 const MainLayout: LayoutComponent = ({
   className,
   children,
-  modifiers = []
+  modifiers = [],
+  pageContext
 }) => {
   useEffect(() => {
     if (className) {
@@ -51,11 +54,13 @@ const MainLayout: LayoutComponent = ({
 
   return (
     <>
+      <SkipNavLink contentId="main-content" className={styles.skipLink} />
       <LayoutHeader modifiers={modifiers} />
       <div
         id="layoutContent"
         //  className={styles.pageContent}
       >
+        {!pageContext.isDocs && <SkipNavContent id="main-content" />}
         {children}
       </div>
       <LayoutFooter />
--- a/src/components/MainLayout/index.tsx
+++ b/src/components/MainLayout/index.tsx
@@ -8,6 +8,8 @@ import LayoutFooter from '../LayoutFooter'
 import { handleFirstTab } from 'gatsby-theme-iterative-docs/src/utils/front/accessibility'
 
 import * as styles from './styles.module.css'
+import { logEvent } from 'gatsby-theme-iterative-docs/src/utils/front/plausible'
+import { useLocation } from '@reach/router'
 
 export enum LayoutModifiers {
   Wide,
@@ -33,6 +35,7 @@ const MainLayout: LayoutComponent = ({
   modifiers = [],
   pageContext
 }) => {
+  const location = useLocation()
   useEffect(() => {
     if (className) {
       document.body.classList.add(className)
@@ -54,7 +57,13 @@ const MainLayout: LayoutComponent = ({
 
   return (
     <>
-      <SkipNavLink contentId="main-content" className={styles.skipLink} />
+      <SkipNavLink
+        contentId="main-content"
+        className={styles.skipLink}
+        onClick={() => {
+          logEvent('Skip To Content', { path: location.pathname })
+        }}
+      />
       <LayoutHeader modifiers={modifiers} />
       <div
         id="layoutContent"
--- a/src/components/MainLayout/styles.module.css
+++ b/src/components/MainLayout/styles.module.css
@@ -2,3 +2,11 @@
   font-weight: 400 !important;
   background-color: #fff;
 }
+
+.skipLink {
+  &:focus {
+    z-index: 999;
+    font-size: 1.125rem;
+    top: 8px;
+  }
+}
--- a/src/components/MainLayout/styles.module.css
+++ b/src/components/MainLayout/styles.module.css
@@ -4,9 +4,15 @@
 }
 
 .skipLink {
+  color: var(--color-orange);
+  font-size: 1.25rem;
+  text-decoration: none;
+  border-radius: 4px;
+  box-shadow: rgb(46 41 51 / 8%) 0 2px 4px, rgb(71 63 79 / 16%) 0 4px 8px;
+
   &:focus {
-    z-index: 999;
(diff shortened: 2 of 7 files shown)


src/components/MainLayout/styles.module.css
```diff
@@ -2,3 +2,11 @@
   font-weight: 400 !important;
   background-color: #fff;
 }
+
+.skipLink {
+  &:focus {
+    z-index: 999;
+    font-size: 1.125rem;
```
julieg18, line 9: Should we improve the styling a bit? Right now, it's just a white rectangle holding a non decorated link

image

Maybe add some border-radius, box-shadow, and link styling?

julieg18 (commented):
I love this idea! It can really help keyboard users, especially when navigating the docs!

yathomasi:
@iterative/websites 
Any more suggestions? 
Also, preference on this?
Defaut one (Current Deployment): 
Screen Shot 2022-03-28 at 22 03 02

Proposed: `outline-offset: value between -5px to -10px`
Screen Shot 2022-03-28 at 21 01 49

jorgeorpinel:
I tried it on https://dvc-org-skip-to-content-9glmvb.herokuapp.com/doc/start and it just scrolls to the bottom of the page. What is it skipping exactly?

Also, if you click the nav to some other page then `tab` no longer shows this feature. How do you access it in that case (or after any other nav usage e.g. via keyboard)?

yathomasi:
What's the interference you are referring to? Everything appears to be looking good on my end 🤔

I don't see it now. It seemed to have an issue when I opened the link with `#main-content` directly. But, since it's not a use-case so this can be neglected.

> I tried it on https://dvc-org-skip-to-content-9glmvb.herokuapp.com/doc/start and it just scrolls to the bottom of the page. What is it skipping exactly?

https://www.loom.com/share/b6b825b9e152482ab4fd736d0dbc9f18

This helps skip to the main content of the page so that the user doesn't have to go through all the navigation menus before reaching the content. This feature is useful for keyboard users and otherwise, it's not visible will not interfere with them.

> Also, if you click the nav to some other page then the tab no longer shows this feature. How do you access it in that case (or after any other nav usage e.g. via keyboard)?

If clicked on the nav, the focus shifts to that place so the tab should continue moving the focus from there. This will particularly helpful when the site first loads and the keyboard users can tab to reach the main content. Also, it's available on the website as a whole so it should be useful in other places as well. 
For eg: If the user open the support link: https://dvc-org-skip-to-content-9glmvb.herokuapp.com/support
Then, the keyboard user can tab in 2-3 go to reach `Discord Chat`.

jorgeorpinel:
> This helps skip to the main content of the page

Again, it's not clear what's happening from trying https://dvc-org-skip-to-content-9glmvb.herokuapp.com/doc/start or any other docs page basically. It just scrolls down a page. The main content is already on the top of the page so its unclear why we're scrolling at all

> This will particularly helpful when the site first loads

Maybe this should be specifically enabled in the landing pages you have in mind. Can you list them? According to our analytics most entry pages are in /doc though, where the current behavior isn't making much sense IMO.

julieg18:
> Again, it's not clear what's happening from trying https://dvc-org-skip-to-content-9glmvb.herokuapp.com/doc/start or any other docs page basically. It just scrolls down a page. The main content is already on the top of the page so its unclear why we even need to skip anything.

Without this feature, a keyboard user is forced to tab through all the nav links and all the sidebar links to get to the main content, which is quite a few links! 

Without this feature:

https://user-images.githubusercontent.com/43496356/160868263-1fd18aa0-facd-4c7e-9abc-de11864f1f65.mov

With this feature:

https://user-images.githubusercontent.com/43496356/160868889-0d485985-c5cf-427d-a66e-09a8fe5699e6.mov

jorgeorpinel:
I see. I was pressing SPACE instead of ENTER 😅  that's why it just scrolled down a page. I see there's no scrolling with ENTER -- it skips the nav.

Sometimes it does scroll down just a little and hides the main title though, is that a bug? It doesn't happen when the page just loaded (though you see it glitch really fast) but if you go all the way around (or click somewhere blank with the mouse) and try again without re-loading, it happens.

jorgeorpinel:
> This will particularly helpful when the site first loads

BTW is there any way to go back to the top easily with tab-key navigation? Otherwise realistically this feature doesn't help people who're navigating via the nav since the Windows never reloads.

shcheklein:
Quick question - how would users navigate to the navigation panel / or top menu though? Is it possible?

Keyboard navigation is very important for accessibility as far as  understand. Only after that I would think about optimizing it for the regular users.

Btw, on this page TAB doesn't skip any parts of the page. It depends on the focus, mouse behavior, etc - hard to tell to be honest without going deep.

julieg18:
> BTW is there any way to go back to the top easily with tab-key navigation? Otherwise realistically this feature doesn't help people who're navigating via the nav since the Windows never reloads on our engine.

Keyboard users usually restart the page when they want to go back to the top quickly :)

jorgeorpinel:
If you skip to content and then refresh the page, it skips the nav again since it's using the `#main-content` anchor. So once using this feature you can't go back to the top easily, it seems.

Another tricky deal is to get to the left side nav since you have to tab 16 times to get to 🏠 Home

jorgeorpinel:
> Another tricky deal is to get to the left side nav since you have to tab 16 times to get to 🏠 Home

Nvmd there's a trick to get there actually: press `/` to get to the search bar, then one TAB (assuming this will be obvious for people used to keyboard nav).

julieg18:
> If you skip to content and then refresh the page, it skips the nav again since it's using the #main-content anchor. So once using this feature you can't go back to the very top easily, it seems -- not sure if that's a big deal. You could edit the URL I guess (via Ctrl+L)

Correct about editing the URL! And "Skip to main content" links are a recommended and common website practice! Any keyboard user who lands on this link is most likely going to aware that this will open an anchor link and understand the best ways to get back to the top if desired.

jorgeorpinel:
> Any keyboard user who lands on this link is most likely going to be aware that this will open an anchor

GH for example doesn't go to an anchor. In fact I have no idea what it does haha -- doesn't seem to do anything at all. And it's indeed only available upon first load of the page.

jorgeorpinel (approved):
Oh well it seems to be a thing pretty similar to how we have it now so 🤷🏼

shcheklein:
Okay, disregard my comment I see that there is a button Skip to content now and there is an option to "skip" it :)

julieg18 (approved):
Great job!

src/components/MainLayout/index.tsx
```diff
@@ -51,11 +57,19 @@ const MainLayout: LayoutComponent = ({
 
   return (
     <>
+      <SkipNavLink
+        contentId="main-content"
+        className={styles.skipLink}
+        onClick={() => {
+          logEvent('Skip To Content', { path: location.pathname })
+        }}
```
yathomasi, line 65: Added log event to capture click of skip to content. This can help us to decide if we want this to be added to other websites as well.
cc: @iterative/websites

PR #3407, opened 2022-03-31
guide: venv `pip` for local dev env setup

jorgeorpinel:
Shouldn't we use `python3 -m pip` here?

### DIFF
--- a/content/docs/user-guide/contributing/core.md
+++ b/content/docs/user-guide/contributing/core.md
@@ -56,13 +56,13 @@ do that, we **strongly** recommend creating a
 $ cd dvc
 $ python3 -m venv .env
 $ source .env/bin/activate
-$ pip install -e ".[all,tests]"
+$ python3 -m pip install -e ".[all,tests]"
 ```
 
 Install coding style pre-commit hooks with:
 
 ```dvc
-$ pip install pre-commit
+$ python3 -m pip install pre-commit
 $ pre-commit install
 ```
 


content/docs/user-guide/contributing/core.md
```diff
@@ -56,13 +56,13 @@ do that, we **strongly** recommend creating a
 $ cd dvc
 $ python3 -m venv .env
 $ source .env/bin/activate
-$ pip install -e ".[all,tests]"
+$ python3 -m pip install -e ".[all,tests]"
```
jorgeorpinel, line 59: ☝🏼 

For context I personally had to do this on my Mac since `pip` was still from globals even after activating the venv.

jorgeorpinel: p.s. could be just `python` at this point but I thought `python3` looked more consistent in that sample block (same result).

pmrowla: This shouldn't be necessary, once the venv is activated the `pip` in the user's `$PATH` should be the one from the venv's `bin/pip`.

In your case it sounds like a particular issue w/your shell environment configuration (you probably needed to run zsh's `rehash` to force zsh to handle `$PATH` changes).

jorgeorpinel: Hmmm ah yes. Zsh seems to alias `pip` to `noglob pip` by default...

jorgeorpinel: p.s. but under the alias the `pip` used from $PATH is correct 🎉  so yeah just a confusion on my side.

```console
$ which -a pip
pip: aliased to noglob pip
/Users/ME/dvc/.env/bin/pip
/opt/homebrew/bin/pip
```

PR #3413, opened 2022-04-01
Added id as props for unique anchor

### LINKED ISSUE #3383: details: unique anchors shouldn't change with content order

yathomasi:
Fixes #3383 
The comments from previous pr https://github.com/iterative/dvc.org/pull/3352#issuecomment-1068236555 and issue https://github.com/iterative/dvc.org/issues/3383#issuecomment-1085153649 are taken into consideration.

So by default, the anchor id will be generated based on the title. And, if there are duplicate titles inside the same page(path) then a run time error popup will appear on the page.
We can set the unique id using the `id` as props on the `<details>` tag, this will clear the above error if the id is unique.

Currently, if the `<details>` block misses the header element as first-child an error popup is thrown. It will be similar to that.
We need to display the error popup in order to make sure the id either passed as props or generated from the title is unique.

PS: Again, I expect the preview deployment to fail because I have only added a unique id for the duplicate titles on `start/data-and-model-versioning`. I will be collaborating with @iterative/docs once the implementation is finalized.

### DIFF
--- a/content/docs/command-reference/diff.md
+++ b/content/docs/command-reference/diff.md
@@ -119,7 +119,7 @@ $ dvc diff
 
 ## Example: Comparing workspace with arbitrary commits
 
-<details>
+<details id="example-comparing-workspace-with-arbitrary-commits-expand">
 
 ### Click and expand to set up the example
 
@@ -149,7 +149,7 @@ files summary: 1 added, 0 deleted, 0 modified
 
 ## Example: Comparing tags or branches
 
-<details>
+<details id="example-comparing-tags-or-branches-expand">
 
 ### Click and expand to set up the example
 
@@ -223,7 +223,7 @@ It outputs:
 
 ## Example: Renamed files
 
-<details>
+<details id="example-renamed-files-expand">
 
 ### Click and expand to set up the example
 
--- a/content/docs/command-reference/diff.md
+++ b/content/docs/command-reference/diff.md
@@ -119,7 +119,7 @@ $ dvc diff
 
 ## Example: Comparing workspace with arbitrary commits
 
-<details id="example-comparing-workspace-with-arbitrary-commits-expand">
+<details id="example-arbitrary-commits-set-up">
 
 ### Click and expand to set up the example
 
@@ -149,7 +149,7 @@ files summary: 1 added, 0 deleted, 0 modified
 
 ## Example: Comparing tags or branches
 
-<details id="example-comparing-tags-or-branches-expand">
+<details id="example-tags-branches-set-up">
 
 ### Click and expand to set up the example
 
@@ -223,7 +223,7 @@ It outputs:
 
 ## Example: Renamed files
 
-<details id="example-renamed-files-expand">
+<details id="example-renamed-files-set-up">
 
 ### Click and expand to set up the example
 
--- a/content/docs/install/linux.md
+++ b/content/docs/install/linux.md
@@ -79,7 +79,7 @@ $ snap install --classic dvc
 
 ## Install from repository
 
-<details id="from-repo-on-debian-or-ubuntu">
+<details id="from-repo-on-debian-ubuntu">
 
 ### On Debian/Ubuntu
 
@@ -94,7 +94,7 @@ $ sudo apt install dvc
 
 </details>
 
-<details id="from-repo-on-fedora-or-centos">
+<details id="from-repo-on-fedora-centos">
 
 ### On Fedora/CentOS
 
@@ -115,7 +115,7 @@ Get the binary package from the big "Download" button on the [home page](/), or
 from the [release page](https://github.com/iterative/dvc/releases/) on GitHub.
 Then install it with the following command.
 
-<details id="from-pkg-on-debian-or-ubuntu">
+<details id="from-pkg-on-debian-ubuntu">
 
 ### On Debian/Ubuntu
 
@@ -125,7 +125,7 @@ $ sudo apt install ./dvc_0.62.1_amd64.deb
 
 </details>
 
-<details id="from-pkg-on-fedora-or-centos">
+<details id="from-pkg-on-fedora-centos">
 
 ### On Fedora/CentOS
 
--- a/content/docs/install/linux.md
+++ b/content/docs/install/linux.md
@@ -22,7 +22,7 @@ plan to use, you might need to install optional dependencies: `[s3]`,
 `[gdrive]`, `[gs]`, `[azure]`, `[ssh]`, `[hdfs]`, `[webdav]`, `[oss]`. Use
 `[all]` to include them all.
 
-<details>
+<details id="example-pip-with-support-for-amazon-s3-storage">
 
 ### Example: with support for Amazon S3 storage
 
@@ -53,7 +53,7 @@ Depending on the type of the [remote storage](/doc/command-reference/remote) you
 plan to use, you might need to install optional dependencies: `dvc-s3`,
 `dvc-azure`, `dvc-gdrive`, `dvc-gs`, `dvc-oss`, `dvc-ssh`.
 
-<details>
+<details id="example-conda-with-support-for-amazon-s3-storage">
 
 ### Example: with support for Amazon S3 storage
 
@@ -79,7 +79,7 @@ $ snap install --classic dvc
 
 ## Install from repository
 
-<details>
+<details id="from-repo-on-debian-or-ubuntu">
 
 ### On Debian/Ubuntu
 
@@ -94,7 +94,7 @@ $ sudo apt install dvc
 
 </details>
 
-<details>
+<details id="from-repo-on-fedora-or-centos">
 
 ### On Fedora/CentOS
 
@@ -115,7 +115,7 @@ Get the binary package from the big "Download" button on the [home page](/), or
 from the [release page](https://github.com/iterative/dvc/releases/) on GitHub.
 Then install it with the following command.
 
-<details>
+<details id="from-pkg-on-debian-or-ubuntu">
 
 ### On Debian/Ubuntu
 
@@ -125,7 +125,7 @@ $ sudo apt install ./dvc_0.62.1_amd64.deb
 
 </details>
 
-<details>
+<details id="from-pkg-on-fedora-or-centos">
 
 ### On Fedora/CentOS
 
--- a/content/docs/start/data-and-model-versioning.md
+++ b/content/docs/start/data-and-model-versioning.md
@@ -59,7 +59,7 @@ $ git commit -m "Add raw data"
 
 The data, meanwhile, is listed in `.gitignore`.
 
-<details id="expand-dvc-add">
+<details id="add-expand-to-see-what-happens-under-the-hood">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -145,7 +145,7 @@ $ dvc push
 Usually, we also want to `git commit` and `git push` the corresponding `.dvc`
 files.
 
-<details id="expand-dvc-push">
+<details id="push-expand-to-see-what-happens-under-the-hood">
 
 ### 💡 Expand to see what happens under the hood.
 
--- a/content/docs/start/data-and-model-versioning.md
+++ b/content/docs/start/data-and-model-versioning.md
@@ -59,7 +59,7 @@ $ git commit -m "Add raw data"
 
 The data, meanwhile, is listed in `.gitignore`.
 
-<details>
+<details id="expand-dvc-add">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -145,7 +145,7 @@ $ dvc push
 Usually, we also want to `git commit` and `git push` the corresponding `.dvc`
 files.
 
-<details>
+<details id="expand-dvc-push">
 
 ### 💡 Expand to see what happens under the hood.
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -79,7 +79,7 @@ DVC uses these metafiles to track the data used and produced by the stage, so
 there's no need to use `dvc add` on `data/prepared`
 [manually](/doc/start/data-and-model-versioning).
 
-<details id="pipeline-stages-expand">
+<details id="stage-expand-to-see-what-happens-under-the-hood">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -172,7 +172,7 @@ $ dvc stage add -n featurize \
 
 The `dvc.yaml` file is updated automatically and should include two stages now.
 
-<details id="dependency-graphs-expand">
+<details id="pipeline-expand-to-see-what-happens-under-the-hood">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -271,7 +271,7 @@ it also doesn't rerun `train`! The previous run with the same set of inputs
 
 </details>
 
-<details id="reproduce-expand">
+<details id="repro-expand-to-see-what-happens-under-the-hood">
 
 ### 💡 Expand to see what happens under the hood.
 
--- a/content/docs/start/data-pipelines.md
+++ b/content/docs/start/data-pipelines.md
@@ -79,7 +79,7 @@ DVC uses these metafiles to track the data used and produced by the stage, so
 there's no need to use `dvc add` on `data/prepared`
 [manually](/doc/start/data-and-model-versioning).
 
-<details>
+<details id="pipeline-stages-expand">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -172,7 +172,7 @@ $ dvc stage add -n featurize \
 
 The `dvc.yaml` file is updated automatically and should include two stages now.
 
-<details>
+<details id="dependency-graphs-expand">
 
 ### 💡 Expand to see what happens under the hood.
 
@@ -271,7 +271,7 @@ it also doesn't rerun `train`! The previous run with the same set of inputs
 
 </details>
 
-<details>
+<details id="reproduce-expand">
 
 ### 💡 Expand to see what happens under the hood.
 
--- a/content/docs/user-guide/external-dependencies.md
+++ b/content/docs/user-guide/external-dependencies.md
@@ -189,7 +189,7 @@ Importing 'https://data.dvc.org/get-started/data.xml' -> 'data.xml'
 The command above creates the import `.dvc` file `data.xml.dvc`, that contains
 an external dependency (in this case an HTTPs URL).
 
-<details id="dvc-import-url-resulting-dot-dvc-file">
+<details id="import-url-expand-to-see-resulting-dvc-file">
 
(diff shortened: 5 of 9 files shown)


plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (this.lowercase) {
+      slug = slug.toLowerCase()
+    }
+    if (this.slugs.includes(slug)) {
+      throw new Error(`Duplicate slug: ${slug}`)
+    }
+    this.slugs.push(slug)
+    return slug
+  }
+  slugify = (str: string) => {
```
julieg18, line 25: Is there a reason why the `slugify` function is written in a different way in comparison with the other functions?

plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
```diff
@@ -0,0 +1,43 @@
...
+    return str
+      .replace(/[^\w\s-]/g, '')
+      .trim()
+      .replace(/[-\s]+/g, this.separator)
+      .replace(this.separator + this.separator, this.separator)
+      .replace(/(^\-+|\-+$)/g, '')
+  }
+  add(slug: string) {
```
julieg18, line 33: Should we use the `add` function inside of the `slug` function? That way the `throw new Error` code is only in one spot?

plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
```diff
@@ -0,0 +1,43 @@
+class Slugger {
```
julieg18, line 1: The `Slugger` class is very well written! At a glance, I was able to easily tell what everything does!

julieg18 (commented):
Code-wise, everything looks pretty good! Just had a couple small comments :)

From testing the code, the docs team would need to add `id` props to `<details>` blocks in these files:
* `content/docs/command-reference/diff.md`
* `content/docs/install/linux.md`
* `content/docs/install/macos.md`
* `content/docs/install/windows.md`
* `content/docs/start/data-and-model-versioning.md`
* `content/docs/start/data-pipelines.md`
* `content/docs/user-guide/external-dependencies.md`

plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
yathomasi: Nope, Thanks for spotting will make it consistent.

plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
yathomasi: Actually, ended up not using it so can drop this function.

plugins/gatsby-theme-iterative-docs/src/utils/front/Slugger.ts
yathomasi: Thanks, also took https://github.com/Flet/github-slugger/blob/master/index.js for reference.

jorgeorpinel:
> if there are duplicate titles inside the same page(path) then a run time error popup will appear on the page
I expect the preview deployment to fail because I have only added a unique id for the duplicate titles on start/data-and-model-versioning

I'm curious about this. Is that necessary? For example if we have repeated titles e.g. `### Something` and `#### Something` there's no build error even when technically it's a violation of HTML rules. Maybe a warning is enough?

That said we do need to update all `<details>` blocks regardless to move first-child text headers into `title` or `id`.

> set the unique id using the id as props

So `<details id="something">` will have anchor `#something`? Or `#{title}-something`

content/docs/start/data-and-model-versioning.md
```diff
@@ -59,7 +59,7 @@ $ git commit -m "Add raw data"
 
 The data, meanwhile, is listed in `.gitignore`.
 
-<details>
+<details id="expand-dvc-add">
```
yathomasi, line 62: As Roger commented, it would be better to take `id` to override the anchor link.
The anchor for this `id` will be `#expand-dvc-add`

> I'm curious about this. Is that necessary? For example if we have repeated titles e.g. ## Something and ### Something there's no build error even when technically it's a violation of HTML rules. Maybe a warning is enough?

Since, the whole idea is to have an anchor link unique, and robust(i.e doesn't change on moving content block to another section). To make sure of that we need to throw an error, instead of letting it pass.

screenshot-localhost_8000-2022 04 02-07_13_28

I think this will be descriptive enough for us to fix that.

> That said we do need to update all `<details>` blocks regardless to move first-child text headers into title or id.

For title as props, we have a separate discussion over #3388. 
So, 
- the title will still be the first child
- the title will be used to generate the unique anchor link
- if there are two or more same titles on the page, there will be an error that will need manual intervention to make that unique
- `id` as props will override the use of the title to generate a unique anchor link
- No, we don't need to update all the `<details>` blocks but only the ones that have duplicate titles
- Also, setting `id` can also be helpful in the context where the title is long so can pass shorter `id` which will override the link

yathomasi:
I have updated the `id` props for the `<details>` component so that the build gets **deployed**. Suggestions for better value for `id` are welcome.

content/docs/start/data-and-model-versioning.md
jorgeorpinel: > would be better to take id to override the anchor link

Agree, sounds like the easiest implementation and usage.

> To make sure of that we need to throw an error
this will be descriptive enough for us to fix that
I still think a warning is enough but not a strong opinion (up to you).

jorgeorpinel: > would be better to take id to override the anchor link

Agree, sounds like the easiest implementation and usage.

> To make sure of that we need to throw an error
this will be descriptive enough for us to fix that

I still think a warning is enough but not a strong opinion. Seems a bit out of scope for this PR to me but again: up to you.

rogermparent:
> Suggestions for better value for `id` are welcome.

Do you mean the name of the prop? I think `id` is perfect since it translates exactly into the resulting HTML code _and_ makes sense on its own.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
```diff
@@ -58,11 +60,17 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
...
   }, '')
-
-  let slug = slugger.slug(title)
-  slug = slug.replaceAll('️', '').replaceAll('ℹ', '')
-  const id = slug.replace(/(^\-+|\-+$)/g, '')
-
+  try {
+    id = id
+      ? useMemo(() => slugger.slug(id), [id])
+      : useMemo(() => slugger.slug(title), [title])
```
rogermparent, line 66: Does `useMemo` do anything here? This also seems to sort of break a rule of hooks that all hooks must always be called, but only technically because I having the same hook always called means the whole "hooks are preserved in the order they're called" thing that necessitates the rule is satisfied.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
```diff
@@ -58,11 +60,17 @@ const Details: React.FC<{ slugger: GithubSlugger }> = ({
...
+  } catch (error) {
+    throw new Error(
+      `Duplicate ${id ? `id: "${id}"` : `title: "${title}"`} \n in page: ${
+        location.pathname
+      }`
```
rogermparent, line 71: I think just the id would be fine such that the `try/catch` can be removed and the underlying Slugger's error message would be what's shown; the id is what's really what is being checked and it's not too difficult to figure out which title created the offending id.

rogermparent (approved):
Looks pretty good! Some concerns on my end but from local testing it's totally working even when trying to cause an issue with edge cases like changing ids in the dev server.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
yathomasi: There was an issue with the slugger function throwing error for the same title during rerenders. So, this would skip generating slug for the same title or id during rerenders. Since this is component level this will not have side effects with other `<details>` component titles. 

> all hooks must always be called

So, if you mean if one gets another will not get called. In that case, we can have the conditional statement inside the useMemo function.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
yathomasi: Yes, the error generated from the Slugger should be enough and I agree they can be figured out, but here I try to be more descriptive with `id` or `title` and `current-pathname` on the error description so that the #docs team can find it easy.

content/docs/command-reference/diff.md
```diff
@@ -119,7 +119,7 @@ $ dvc diff
 
 ## Example: Comparing workspace with arbitrary commits
 
-<details>
+<details id="example-comparing-workspace-with-arbitrary-commits-expand">
```
yathomasi, line 122: > Do you mean the name of the prop?

Actually, I meant these values for `ids`. It's better if @iterative/docs team can go through them once.

plugins/gatsby-theme-iterative-docs/src/components/Documentation/Markdown/index.tsx
rogermparent: > So, if you mean if one gets another will not get called. In that case, we can have the conditional statement inside the useMemo function.

Yep, that's usually how `useMemo` calls with conditionals are handled.

yathomasi: > Your explanation makes me think of how it's an interesting case that the page could error out if the writer goes back and forth on an id, or moves the same one around. It's enough of an edge case to be overlooked, especially since it can be cleared with a page refresh, but it's something worth looking into for the future

Umm, it errors out if we don't use the `useMemo` here. And, it's how the `Slugger` is implemented that would cause that error since each time it gets called the slug generated gets added to the Slug list to maintain uniqueness. 
Do we have a better way to handle this 🤔 ?

rogermparent: The reason I bring up turning this into a remark plugin is that I believe this issue with the `Slugger` failing in dev mode is linked to the id processing being done in React, whereas in remark the whole page would get reprocessed with a new slugger instance if I understand correctly.

That said, that would be a pretty big refactor to fix a pretty niche edge case that would only ever show up in development, so no need to go for it right now.

content/docs/command-reference/diff.md
jorgeorpinel: For now they should be the same as now. In this case `click-and-expand-to-set-up-the-example` (link: https://dvc.org/doc/command-reference/diff#click-and-expand-to-set-up-the-example)

jorgeorpinel:
```suggestion
<details id="example-arbitrary-commits-set-up-the-example">
```

content/docs/command-reference/diff.md
```diff
@@ -149,7 +149,7 @@ files summary: 1 added, 0 deleted, 0 modified
 
 ## Example: Comparing tags or branches
 
-<details>
+<details id="example-comparing-tags-or-branches-expand">
```
jorgeorpinel, line 152:
```suggestion
<details id="example-tags-branches-set-up-the-example">
```

content/docs/command-reference/diff.md
```diff
@@ -223,7 +223,7 @@ It outputs:
 
 ## Example: Renamed files
 
-<details>
+<details id="example-renamed-files-expand">
```
jorgeorpinel, line 226:
```suggestion
<details id="example-renamed-files-set-up-the-example">
```

content/docs/install/linux.md
```diff
@@ -79,7 +79,7 @@ $ snap install --classic dvc
 
 ## Install from repository
 
-<details>
+<details id="from-repo-on-debian-or-ubuntu">
```
jorgeorpinel, line 82:
```suggestion
<details id="from-repo-on-debian-ubuntu">
```

content/docs/install/linux.md
```diff
@@ -94,7 +94,7 @@ $ sudo apt install dvc
 
 </details>
 
-<details>
+<details id="from-repo-on-fedora-or-centos">
```
jorgeorpinel, line 97:
```suggestion
<details id="from-repo-on-fedora-centos">
```

content/docs/install/linux.md
```diff
@@ -115,7 +115,7 @@ Get the binary package from the big "Download" button on the [home page](/), or
 from the [release page](https://github.com/iterative/dvc/releases/) on GitHub.
 Then install it with the following command.
 
-<details>
+<details id="from-pkg-on-debian-or-ubuntu">
```
jorgeorpinel, line 118:
```suggestion
<details id="from-pkg-on-debian-ubuntu">
```

content/docs/install/linux.md
```diff
@@ -125,7 +125,7 @@ $ sudo apt install ./dvc_0.62.1_amd64.deb
 
 </details>
 
-<details>
+<details id="from-pkg-on-fedora-or-centos">
```
jorgeorpinel, line 128:
```suggestion
<details id="from-pkg-on-fedora-centos">
```

content/docs/start/data-and-model-versioning.md
jorgeorpinel, line 62:
```suggestion
<details id="add-expand-to-see-what-happens-under-the-hood">
```

content/docs/start/data-and-model-versioning.md
```diff
@@ -145,7 +145,7 @@ $ dvc push
 Usually, we also want to `git commit` and `git push` the corresponding `.dvc`
 files.
 
-<details>
+<details id="expand-dvc-push">
```
jorgeorpinel, line 148:
```suggestion
<details id="push-expand-to-see-what-happens-under-the-hood">
```

content/docs/start/data-pipelines.md
```diff
@@ -79,7 +79,7 @@ DVC uses these metafiles to track the data used and produced by the stage, so
 there's no need to use `dvc add` on `data/prepared`
 [manually](/doc/start/data-and-model-versioning).
 
-<details>
+<details id="pipeline-stages-expand">
```
jorgeorpinel, line 82:
```suggestion
<details id="stage-expand-to-see-what-happens-under-the-hood">
```

content/docs/start/data-pipelines.md
```diff
@@ -172,7 +172,7 @@ $ dvc stage add -n featurize \
 
 The `dvc.yaml` file is updated automatically and should include two stages now.
 
-<details>
+<details id="dependency-graphs-expand">
```
jorgeorpinel, line 175:
```suggestion
<details id="pipeline-expand-to-see-what-happens-under-the-hood">
```

content/docs/start/data-pipelines.md
```diff
@@ -271,7 +271,7 @@ it also doesn't rerun `train`! The previous run with the same set of inputs
 
 </details>
 
-<details>
+<details id="reproduce-expand">
```
jorgeorpinel, line 274:
```suggestion
<details id="repro-expand-to-see-what-happens-under-the-hood">
```

content/docs/user-guide/external-dependencies.md
```diff
@@ -189,7 +189,7 @@ Importing 'https://data.dvc.org/get-started/data.xml' -> 'data.xml'
 The command above creates the import `.dvc` file `data.xml.dvc`, that contains
 an external dependency (in this case an HTTPs URL).
 
-<details>
+<details id="dvc-import-url-resulting-dot-dvc-file">
```
jorgeorpinel, line 192:
```suggestion
<details id="import-url-expand-to-see-resulting-dvc-file">
```

content/docs/user-guide/external-dependencies.md
```diff
@@ -227,7 +227,7 @@ Importing 'model.pkl (git@github.com:iterative/example-get-started)'
 The command above creates `model.pkl.dvc`, where the external dependency is
 specified (with the `repo` field).
 
-<details>
+<details id="dvc-import-resulting-dot-dvc-file">
```
jorgeorpinel, line 230:
```suggestion
<details id="import-expand-to-see-resulting-dvc-file">
```

jorgeorpinel (commented):
Suggestions based on https://github.com/iterative/dvc.org/pull/3413#discussion_r845601788 below. WDYT?

content/docs/command-reference/diff.md
yathomasi: maybe remove example from prefix?
`<details id="example-arbitrary-commits-set-up-the-example">`
to
`<details id="arbitrary-commits-set-up-the-example">`

content/docs/command-reference/diff.md
yathomasi: @jorgeorpinel maybe remove `example` from the prefix?
`<details id="example-tags-branches-set-up-the-example">`
to
<details id="tags-branches-set-up-the-example">

content/docs/command-reference/diff.md
jorgeorpinel: Good call but from the end? `example-arbitrary-commits-set-up`

For consistency, since all my suggestions use prefixes.

jorgeorpinel: Good call but from the end? `example-arbitrary-commits-set-up`

For consistency, since all my suggestions use prefixes.

content/docs/command-reference/diff.md
jorgeorpinel: Similar answer: `example-tags-branches-set-up ` ?

content/docs/command-reference/diff.md
jorgeorpinel: UPDATE: I've edited all my similar suggestions removing `-the-example`

PR #3423, opened 2022-04-07
Added proxy api for Plausible on current express server

yathomasi:
Fixes https://github.com/iterative/iterative.ai/issues/423

Created proxy API for plausible basically to handle these two requests
```
pl/js/script.js -> https://plausible.io/js/script.js
pl/event    -> https://plausible.io/api/event
```

Todo: 
- [x] get this API implementation reviewed and finalize
- [x] replace current plausible API with this proxy API

**Update**: 
Script: https://dvc-org-proxy-plausible-cd4crn.herokuapp.com/api/pl/js/script.js
Event API: https://dvc-org-proxy-plausible-cd4crn.herokuapp.com/api/pl/event

### DIFF
--- a/src/server/middleware/api/plausible.js
+++ b/src/server/middleware/api/plausible.js
@@ -38,10 +38,15 @@ async function handlePlausibleRequest(req, res) {
           'content-type': 'application/json'
         }
       })
-      res.set({
-        'content-type': response.headers.get('content-type')
-      })
-      return res.status(response.status).send(await response.text())
+      if (response.ok) {
+        res.set({
+          'content-type': response.headers.get('content-type')
+        })
+        return res.status(response.status).send(await response.text())
+      } else {
+        if (!isProduction) console.error(error)
+        return res.status(500).end()
+      }
     }
     res.status(404).end()
   } catch (error) {
--- a/src/server/middleware/api/plausible.js
+++ b/src/server/middleware/api/plausible.js
@@ -41,7 +41,7 @@ async function handlePlausibleRequest(req, res) {
       res.set({
         'content-type': response.headers.get('content-type')
       })
-      return response.body.pipe(res)
+      return res.status(response.status).send(await response.text())
     }
     res.status(404).end()
   } catch (error) {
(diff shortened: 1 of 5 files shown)


rogermparent (commented):
The core concept looks solid to me! Most of our APIs have a caching system using `node-cache`, could that be valuable here as well?

src/server/middleware/api/plausible.js
```diff
@@ -0,0 +1,80 @@
+const { isProduction } = require('../../utils')
+const NodeCache = require('node-cache')
+require('isomorphic-fetch')
+
+const ScriptName = '/pl/script.js'
```
julieg18, line 5: Why are these variables capitalized? If they're constants, shouldn't they be completely uppercase 🤔

julieg18 (commented):
Looks good to me, but I have little experience in NodeJS. @rogermparent, what do you think?

src/server/middleware/api/plausible.js
rogermparent: That said, maybe these constants should be shared with the other files that require them? [DefaultSEO](https://github.com/iterative/dvc.org/pull/3423/files#diff-aa1a28d5dbb813758bec2678c00fcb7cf9104a2b5cfa9bbe8313270feaeec84a) could use this link.

rogermparent (commented):
I'm admittedly not the strongest in Express, but nothing looks outright obviously here. I think a change to the consts is all we need to make this mergeable!

src/server/middleware/api/plausible.js
yathomasi: Thanks for pointing it out.

yathomasi:
It looks like the API is not working on gatsby cloud deployment. If we are going to use Gatsby Cloud, it would be better to use [Gatsby Functions](https://www.gatsbyjs.com/docs/reference/functions/)

rogermparent:
We've mostly just enabled Gatsby Cloud to give content editors deploy previews while Heroku integration is busted. We certainly could use functions, but I figured we'd only go down that road if we wanted a full switch over.

rogermparent (approved):
Even if the deploy previews can't test it, I think it's fine to bite the bullet and try this in prod as long as we ensure it works soon after it's merged.

Issue #2572, opened 2021-06-18
guide: Params, Plots and Metrics


iesahin:
The current documentation includes information on plots and metrics in the command reference for [`dvc plots`](https://dvc.org/doc/command-reference/plots) and [`dvc metrics`](https://dvc.org/doc/command-reference/metrics).  We need a ~~self-contained~~ UG page for all features regarding the plots and metrics. 

- [ ]  Extract details from https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#stages
- [ ] Extract some details from [params](https://dvc.org/doc/command-reference/params) and [metrics](https://dvc.org/doc/command-reference/metrics) ref. indices. 
  See also https://github.com/iterative/dvc.org/issues/4202
- Maybe clarify the concept of *params** as well.
- ~~Keep in mind conceptual and terminology decisions in https://github.com/iterative/dvc.org/issues/2956~~

### Related

- #2472 contains a page that describes offline usage of Vega libraries
- #2402 is the issue around offline usage
- #2216 contains some information about using plots/metrics without pipelines
- #1442 is terminology issue around plots and metrics
- #1755 asks for clarification where to use DVC or Git to track the plots and metrics
- #1909 is about the capability of `metrics` and `plots` to work with any parsable file
- #1175 is also a terminology issue
- #3818
- #2644

daavoo:
Related?: https://github.com/iterative/dvc.org/issues/2644

jorgeorpinel:
From https://github.com/iterative/dvc.org/issues/2815

> We briefly cover those 3 types of files on the [`Project Structure -> Pipeline Files -> Stages` section](https://dvc.org/doc/user-guide/project-structure/pipelines-files#stages). 
However, it feels like the current info in each command reference could be moved to the user guide, potentially extending it.

jorgeorpinel:
> Explain params metafiles
> metrics and plots metafiles

Should these be tasks in https://github.com/iterative/dvc.org/issues/2911 since params, metrics and plots are mainly experiment tracking mechanisms? (Correct me if I'm wrong.)

iesahin:
Conceptually these are rather closer to experiment management. However, I think _metrics_  that's relevant to `dvc exp` commands are covered (implicitly) in `dvc exp show`. `dvc metrics` command is related to pipelines in its current usage.

I think having a separate metrics/plots guide that describes these in pipelines _and_ experiments context may be a better approach.

jorgeorpinel:
[...]

PR #3431, opened 2022-04-08
`params`: support for tracking all params in a file

skshetry:
See https://github.com/iterative/dvc/pull/7432.


content/docs/command-reference/params/index.md
```diff
@@ -43,7 +43,9 @@ to define [stage](/doc/command-reference/run) dependencies more granularly:
 changes to other parts of the params file will not affect the stage. Parameter
 dependencies also prevent situations where several stages share a regular
 dependency (e.g. a config file), and any change in it invalidates all of them
-(see `dvc status`), causing unnecessary re-executions upon `dvc repro`.
+(see `dvc status`), causing unnecessary re-executions upon `dvc repro`. However,
+you can track all the parameters in a file by setting _parameter value_ to be
+`null`.
```
skshetry, line 48: I am not quite sure on how to include this information here.

dberenbaum: I think an example either here or in the examples section below would clarify better than any explanation.

dberenbaum: It would help to show both `dvc.yaml` and how `dvc stage add` can be used.

Some rationale for it might also help, like explaining that you may want to granularly track all parameters in a file and it would be tedious to list them all and keep them updated as new parameters are added.

dberenbaum, line 48: Yeah, this is tricky. What about this?

```suggestion
you can track all the parameters in a file by specifying the value of
`params.yaml:` to be `null` or empty.
```

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -51,8 +51,13 @@ If it writes files or dirs, they can be defined as <abbr>outputs</abbr>
 ### Parameter dependencies
 
 [Parameters](/doc/command-reference/params) are a special type of stage
```
dberenbaum, line 53: I'm not sure we need any of the changes to this user guide page. It otherwise looks like a very brief overview of parameters, and this seems like too much detail.

jorgeorpinel:
Is there a syntax for `run/ stage add -p` to indicate the whole default params file? What about a custom one? If so those refs would need an update probably.

content/docs/command-reference/params/index.md
jorgeorpinel: No need to include this info. in this paragraph I think. This is a general explanation about parameter dependencies.

How about adding the note or even a new paragraph and sample block to https://dvc.org/doc/user-guide/project-structure/pipelines-files#parameter-dependencies instead?

Agree about updating `stage add` (as I mentioned in another comment) and on mentioning motivation when presenting this function

content/docs/user-guide/project-structure/pipelines-files.md
jorgeorpinel: Oh I see you did update the schema ref. already @skshetry 👍🏼 

@dberenbaum I do think we need to update it since the schema changed. The detail given now may be too much but we can address that...

jorgeorpinel: > the schema changed 

p.s. does https://github.com/iterative/dvcyaml-schema need an update?

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -51,8 +51,13 @@ If it writes files or dirs, they can be defined as <abbr>outputs</abbr>
...
-dependency. They consist of a name/value pair to find in a YAML, JSON, TOML, or
-Python parameters file (`params.yaml` by default). Example:
+dependency. They consist of a list of values, in which each value can be:
+
+1. a string which can be found in `params.yaml` file,
+2. a dictionary, with key being the file path and value being the list of
+   parameters to find in a given file to track or set to empty/`null` which will
+   track all the parameters in the file. The given file path must be a valid
+   YAML, JSON, TOML, or Python parameters file.
```
jorgeorpinel, line 60:
```suggestion
dependency. They consist of a list of params to track in one of these formats:

1. A param key/value pair that can be found in `params.yaml` (default
   params file);
2. A dictionary named by the file path to a custom params file, and with a
   list of param key/value pairs to find in it;
3. An empty set (give no value or use `null`) named by the file path to a
   params file: to track all the params in it dynamically. 
  
> Note that file paths used must be to valid YAML, JSON, TOML, or Python
> parameters file.
```
^ That last note should probably be an admonition.

content/docs/user-guide/project-structure/pipelines-files.md
daavoo: > > the schema changed
> 
> p.s. does https://github.com/iterative/dvcyaml-schema need an update?

Done in https://github.com/iterative/dvcyaml-schema/pull/12

jorgeorpinel:
- [ ] Something else to check that we cover:

> Parameter names (paths) cannot contain file extensions supported for parameter files cannot.

_From https://github.com/iterative/dvc/issues/4112#issuecomment-1096155819_

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -51,8 +51,17 @@ If it writes files or dirs, they can be defined as <abbr>outputs</abbr>
...
-dependency. They consist of a name/value pair to find in a YAML, JSON, TOML, or
-Python parameters file (`params.yaml` by default). Example:
+dependency. They consist of a list of params to track in one of these formats:
+
+1. A param key/value pair that can be found in `params.yaml` (default
+   params file);
+2. A dictionary named by the file path to a custom params file, and with a
+   list of param key/value pairs to find in it;
+3. An empty set (give no value or use `null`) named by the file path to a
+   params file: to track all the params in it dynamically.
```
jorgeorpinel, line 61: > The detail given now may be too much but we can address that... (from https://github.com/iterative/dvc.org/pull/3431#pullrequestreview-936901013)

Well... I ended up adding even more details here. This should at some point get extracted to a separate guide (https://github.com/iterative/dvc.org/issues/2572) but OK to keep for now, I think.

content/docs/user-guide/project-structure/pipelines-files.md
```diff
@@ -63,6 +72,9 @@ stages:
     params:
       - threshold
       - passes
+      - custom_params.yaml: # track only specified parameters
+          - epochs
+      - config.json: # track all parameters
```
jorgeorpinel, line 77:
```suggestion
      - threshold  # track specific parameter (from params.yaml)
      - passes
      - custom_params.yaml:  # track only specified parameters
          - epochs
      - config.json:  # track all parameters
```

content/docs/command-reference/params/index.md
jorgeorpinel: Will update the other refs as agreed. Deleting this for now...
```suggestion
(see `dvc status`), causing unnecessary re-executions upon `dvc repro`.
```

jorgeorpinel: I modified the sample instead (similar to the `dvc.yaml` ref.) and generalized the text in the description. See 0ef8c607

jorgeorpinel:
Trying to take this over but I still don't know the answer to https://github.com/iterative/dvc.org/pull/3431#issuecomment-1095946365 above. Will assume not for now. Cc @skshetry @dberenbaum thanks

jorgeorpinel:
Nvmd! The answer is here 😋 : https://github.com/iterative/dvc/issues/4112#issuecomment-1096865302

> it needs to be `-p params.yaml:` if passed as a command line argument

jorgeorpinel:
@skshetry PTAL

daavoo (approved):

shcheklein:
@skshetry @jorgeorpinel merging this?

jorgeorpinel:
> merging this?

Yeah I was expecting approval from @skshetry since I took this over but haven't heard back. There's approval anyway so lets merge 🚀

PR #3442, opened 2022-04-12
start: Reorganize into Trails (Candidate #1)

iesahin:
This PR splits the GS content into little pages that can be interlinked. It's planned to have 3 introduction points. (Data, Pipelines, Experiments)

PR #3050, opened 2021-11-30
Add GS:Visualization and Plots

iesahin:
This is mostly for discussion after the Notion experimentation went "we wanted the best, but it turned out like always" :)

I'll develop this as we discuss the scope of changes. 

Closes #2925 
Closes #3309

PR #3444, opened 2022-04-13
start: Reorganize into Trails (Candidate #1, Take #2)

iesahin:
Due to branching error in #3442, I renewed the PR. 

This PR updates the whole Get Started section into a "trail format" we discussed in #2496. It's an example to present the GS section into a _less scary_ way.


shcheklein (changes requested):
Let's first keep even simpler - two trails, no need to change any content (split into separate section - it looks a bit strange). We can iterate after that if needed.

jorgeorpinel:
Yeah, it was interesting to see all the small pages, but I don't think our engine is currently very conducive to using such short one-pagers -- maybe if we had a better <PREVIOUS, NEXT> navigation UI.

Also it's very hard to review the PR with so many file changes so for a quick proposal just 3 pages with the existing content in a different order would be better indeed.

I realize it will be pretty long for just 2 trails but let's try it (we can call it candidate #2) and take it from there.

p.s. there's a problem in the formatting of the [GS index page](https://dvc-org-iesahin-gs-reor-g57txx.herokuapp.com/doc/start).

jorgeorpinel:
> it's very hard to review the PR with so many file changes

I guess we can review [the deployment](https://dvc-org-iesahin-gs-reor-g57txx.herokuapp.com/doc/start). As for the order of sections (focusing on left-side nav), here's my general feedback:

- A bit too granular, some of these pages can be a single section e.g. Adding Files + Making Changes; or Storing & sharing + Retrieving; or all the `import`-related ones
- I like the more practical renaming of some of the nav titles 👍🏼 
- I'm surprised there's no overlap between trails (no repeated sections). I guess that's good?

But also, let's merge https://github.com/iterative/dvc.org/pull/3050 first!

iesahin:
Closing in favor of #3455

PR #3456, opened 2022-04-20
Try heroku deploy action

rogermparent:
I imagine this won't work since I don't think we have Heroku API keys in our repo's secrets, but let's see what happens.

[Here's the docs for the environment variables we need](https://github.com/mheap/github-action-pr-heroku-review-app#available-configuration)
- `HEROKU_API_TOKEN` comes from someone's account,  which it seems like I can do
- `HEROKU_PIPELINE_ID` seems to be what's displayed in the URL when looking at the review apps index, i.e. `https://dashboard.heroku.com/pipelines/${ID}`


rogermparent:
Closing both because this is incomplete and I want to see if the action can delete review apps as well

rogermparent:
The action does remove review apps when the PR is closed!

PR #3459, opened 2022-04-22
api: Add `params_show`

daavoo:
P.R. for https://github.com/iterative/dvc/pull/7613


content/docs/api-reference/get_params.md
```diff
@@ -0,0 +1,90 @@
...
+
+When multiple `dvc params` files are associated with a stage, this function
+takes care of aggregating and selecting all the relevant keys, regardless of the
+file format used.
+
+## Parameters
+
+- **`stage`** (required) - Name of the dvc.yaml stage to get params from.
+
+- `flatten` - Whether to return flattened dict or not. _Default_: False.
```
shcheklein, line 34: how essential it is to have this option? people could use some lib to do this if they want, wdyt?

content/docs/api-reference/get_params.md
```diff
@@ -0,0 +1,90 @@
...
+import json
+import dvc.api
+
+params = dvc.api.get_params("merge-params")
+
+with open("merged.json", "w") as f:
+    json.dump(params,f, indent=4)
+```
+
+If we run `dvc repro`, this will generate the following output in `merged.json`:
```
shcheklein, line 81: it's not clear why `dvc repro` is here

content/docs/api-reference/get_params.md
```diff
@@ -0,0 +1,90 @@
+# dvc.api.get_params()
+
+Returns all the <abbr>params</abbr> associated with the given
```
shcheklein, line 3: does it make sense to return all params by default and make stage name an optional target?

content/docs/api-reference/get_params.md
```diff
@@ -0,0 +1,90 @@
...
+```py
+def get_params(stage: str, flatten: bool = False) -> Dict:
+```
+
+#### Usage:
+
+```py
+import dvc.api
+
+params = dvc.api.get_params("train")
```
shcheklein, line 15: how will it behave if params come from Python? btw, I assume that it reads the workspace (rather than dvc.lock), is it correct?

content/docs/api-reference/get_params.md
daavoo: IMO, it's not essential to have it exposed. 

I exposed it because the default behavior of this public API (flatten `False`) is different from the default of the internal `Repo` API (flatten `True`). 

However, given that we don't document the `Repo` API at all this might not matter at all.

content/docs/api-reference/get_params.md
daavoo: I wanted to showcase that `dvc.api.get_params` can be used inside a stage while it's running. Not sure if this is necessary for the docs but running `dvc repro` was the only way to showcase that.

content/docs/api-reference/get_params.md
daavoo: > how will it behave if params come from Python?

Only values that DVC can parse will be returned. Complex expressions (https://dvc.org/doc/command-reference/params#examples-python-parameters-file) won't be parsed.

> btw, I assume that it reads the workspace (rather than dvc.lock), is it correct?

Correct.

content/docs/api-reference/get_params.md
daavoo: That kind of deviates from the use case I had in mind for this API, which is retrieving only the parameters relevant to the current stage being run.

Retrieving all DVC params is, IMO, a different use case and can be done with `dvc.api.Repo().params.show()`. Would a note with this snippet be enough?

shcheklein: I see what you mean regarding the use case, but since we are adding this as a public interface we should be also thinking about some general scenarios. In this case it feels natural to expect that target is optional (DVC CLI does the same in a lot of cases), but since it can be changed later (almost) w/o breaking compatibility we can go like this for now. Your call. (I would not include a snippet though with Repo API - it's not public yet?)

content/docs/api-reference/get_params.md
dberenbaum: Should it also be possible to read params from a particular revision and/or repo? It seems consistent with the rest of the API.

content/docs/api-reference/get_params.md
```diff
@@ -0,0 +1,90 @@
+# dvc.api.get_params()
+
+Returns all the <abbr>params</abbr> associated with the given
+<abbr>stage</abbr>.
```
dberenbaum, line 4: `dvc params diff` has a `--targets` argument which takes paths and not stages. Can we have some consistency between targets in `dvc params diff` and the first arg to this method since they both read params? I understand why it makes sense to use stages in your intended use case. Could both `dvc params diff` and `get_params()` accept either a stage or a path like some other dvc commands do? WDYT?

daavoo: Do you mean that we should extend `dvc params diff` to accept also accept a `stage` name as a target?

dberenbaum: Potentially, yes, although we could start by including paths in `get_params()`. Not sure if I like having an arg that could be either stages or paths, though. WDYT?

content/docs/api-reference/get_params.md
dberenbaum: Personally, I was interested to see if we can cover both the CLI-like use cases and the repro scenario to keep the API small and make public the CLI-like features that aren't documented. If it contradicts your use case, it makes sense to leave it out. In the case of making the stage optional, or including optional kwargs for repo or rev, would it break any existing functionality? Does it add a lot more complexity? If so, it's fine to leave as is.

content/docs/api-reference/get_params.md
dberenbaum: Given your comment above, I think it's fine to leave this out. I was thinking it could be useful as a generic way to read multiple different different parameter file formats directly into a dict (even outside a dvc repo), but I don't think it's necessary.

content/docs/api-reference/get_params.md
dberenbaum: The use case I had in mind is, in conjunction with a `get_metrics()`, method, this would enable users to do custom analysis of experiments without having to manually clone the repo and use the CLI.

daavoo: > Personally, I was interested to see if we can cover both the CLI-like use cases and the repro scenario to keep the API small and make public the CLI-like features that aren't documented.

Makes sense.

> would it break any existing functionality? Does it add a lot more complexity? 

I don't think so but I will implement and see.

> The use case I had in mind is, in conjunction with a get_metrics(), method, this would enable users to do custom analysis of experiments without having to manually clone the repo and use the CLI.

Got it

jorgeorpinel: > maybe we should merge and add these in a follow-up

Referring to https://github.com/iterative/dvc/pull/7613 right? Not sure we want to document a WIP feature or not in great detail.

jorgeorpinel:
Hi. Is this ready for review?

content/docs/sidebar.json
```diff
@@ -488,6 +488,10 @@
     "label": "Python API Reference",
     "source": "api-reference/index.md",
     "children": [
+      {
+        "slug": "params_show",
+        "label": "params_show()"
+      },
       {
         "slug": "get_url",
```
jorgeorpinel, line 496: Is there any sorting to this list BTW? Unclear to me why this comes first. If it's by relevance, this should probably be 2nd to last?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+
+```py
+import dvc.api
+
+params = dvc.api.params_show()
+```
+
+## Description
+
+Retrieves the <abbr>params</abbr> tracked in a <abbr>DVC repository</abbr>.
```
jorgeorpinel, line 26: What if they're in a params file but not used in any stages. Are they still "tracked" ?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+
+Without arguments, this function will retrieve all <abbr>params</abbr> from all
+parameter files, for the current revision.
```
jorgeorpinel, line 29: What about unstaged changes? Does this read the actual files in the working tree, or the ones in `HEAD`?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+- **`rev`**: (str, optional): Name of the Git
+  [revision](https://git-scm.com/docs/revisions) to retrieve <abbr>params</abbr>
+  from. Example of git revision can be a branch or tag name, a commit hash or a
+  dvc experiment name. Defaults to `HEAD`. If `repo` is not a Git repo, this
+  option is ignored.
+
+- **`deps`**: (bool, optional): Whether to retrieve only <abbr>params</abbr>
+  that are <abbr>stage</abbr> dependencies or not. Defaults to `False`.
+
+## Example: No arguments.
```
jorgeorpinel, line 58:
```suggestion
## Example: default behavior (no filters)
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+Retrieves the <abbr>params</abbr> tracked in a <abbr>DVC repository</abbr>.
+
+Without arguments, this function will retrieve all <abbr>params</abbr> from all
+parameter files, for the current revision.
+
+See the options below to restrict the <abbr>params</abbr> retrieved.
+
+## Parameters
+
+- **`*targets`** (str, optional): Names of the parameter files to retrieve
```
jorgeorpinel, line 35: Please only bold required args, and replicate existing format in other pages. Thanks
```suggestion
- `*targets` - Names of the parameter files to retrieve e.g. "params.py, myparams.toml"
```
☝🏼 Just an idea. Please fix sample value somewhere in the paragraph, it's unclear what format this arg expects.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+  },
+  "train": {
+    "seed": 20170428,
+    "n_est": 50,
+    "min_split": 0.01
+  }
+}
+```
+
+## Example: Filtering with `stages`.
```
jorgeorpinel, line 87:
```suggestion
## Example: Filter by `stages`
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+  },
+  "train": {
+    "seed": 20170428,
+    "n_est": 50,
+    "min_split": 0.01
+  }
+}
+```
+
+## Example: Using `rev`.
```
jorgeorpinel, line 132:
```suggestion
## Example: Specify a project version
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+  directory tree). It can be a URL or a file system path. Both HTTP and SSH
+  protocols are supported for online Git repos (e.g. [user@]server:project.git).
+
+- **`stages`** (Union[str, Iterable[str]], optional): Name(s) of the stages to
+  retrieve <abbr>params</abbr> from. Defaults to `None`. If no stages are
+  provided, all parameters from all stages will be retrieved.
+
+- **`rev`**: (str, optional): Name of the Git
+  [revision](https://git-scm.com/docs/revisions) to retrieve <abbr>params</abbr>
+  from. Example of git revision can be a branch or tag name, a commit hash or a
```
jorgeorpinel, line 51: Let's reuse some text [from `get_url()`](https://dvc-org-api-get-params-2lnv5kl.herokuapp.com/doc/api-reference/get_url#parameters), I think.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+## Example: Using `rev`.
+
+> Working on https://github.com/iterative/example-get-started
+
+```py
+import json
+import dvc.api
+params = dvc.api.get_params(rev="tune-hyperparams")
```
jorgeorpinel, line 139: Needs a bit more explanation I think. Unclear what's happening or why do this from just the command and output samples (which look similar to the other examples).

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+  }
+}
+```
+
+## Example: Using `targets`.
+
+> Working on `multi-params-files` folder of
+> https://github.com/iterative/pipeline-conifguration
+
+You can pass a single target:
```
jorgeorpinel, line 166:
```suggestion
## Example: Filter by parameter files
> Working on
> https://github.com/iterative/pipeline-conifguration/tree/main/multi-params-files

You can pass a single param file as target:
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+      "size": 1000
+    }
+  },
+  "configs/params_dev.yaml:train": {
+    "epochs": 10
+  }
+}
+```
+
+## Example: Git URL as `repo`.
```
jorgeorpinel, line 232:
```suggestion
## Example: Show parameters from a remote DVC repository
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+
+```py
+import json
+import dvc.api
+params = dvc.api.get_params(
+    repo="https://github.com/iterative/demo-fashion-mnist")
```
jorgeorpinel, line 238: This one also needs some explanation/motivation I think. Or if it's basically the same as the others, we probably don't need the example 🙂 (the motivation can be in the **Parameters** section).

jorgeorpinel (commented):
Thanks @daavoo, here's a review. We can probably take it over after this round of questions.

content/docs/sidebar.json
daavoo: My bad, I didn't think of the order. 

> Is there any sorting to this list BTW?

I think the usual sort is alphabetical

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
+# dvc.api.params_show()
+
+Returns all the <abbr>params</abbr> associated with the given <abbr>DVC
+repository</abbr>.
```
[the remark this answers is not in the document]
daavoo: No, it won't work

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,253 @@
...
+
+```py
+def params_show(
+    *targets: str,
+    repo: Optional[str] = None,
+    stages: Optional[Union[str, Iterable[str]]] = None,
```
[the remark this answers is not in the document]
daavoo: > It's still actually optional, right? Python syntax may be confusing here.

Optional is used in type hints to indicate that `None` is also a valid value. From https://docs.python.org/3/library/typing.html#typing.Optional:

> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

content/docs/api-reference/params_show.md
daavoo: > What if they're in a params file but not used in any stages. Are they still "tracked" ?

They are not tracked and they won't be retrieved unless the untracked params file is passed as `targets`.

Not sure which alternative could be used to better reflect all the posibilities

content/docs/api-reference/params_show.md
daavoo: > Does this read the actual files in the working tree, or the ones in HEAD?

It reads the actual files in the working tree. Should we replace `current revision` with `working tree` or `workspace`?

daavoo: I replaced with `working tree`

content/docs/api-reference/params_show.md
daavoo: Motivation is that you can use the API to retrieve params from a repository without having to clone and/or navigate to it locally

content/docs/api-reference/params_show.md
daavoo: Added an introductory line.

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
daavoo: Not sure about `for the <abbr>stages</abbr>` part, because stages only come into play when using the `stages` argument

content/docs/sidebar.json
jorgeorpinel: Yeah up to you.

Loosely related: should https://dvc.org/doc/api-reference be alphabetical too? (just move `make_checkpoint` up a little)

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
jorgeorpinel: > it won't work

Is that expected? `dvc params` CLI works with non-Git DVC projects I believe.

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
jorgeorpinel: Thanks for the reminder!

> an optional argument, which is one that has a default 

But so `*targets` is also optional, right? Even when it has no default value. Nothing is sent in the Usage block below.

content/docs/api-reference/params_show.md
jorgeorpinel: > unless the untracked params file is passed as targets

- [ ] Good to know, we should add this info to the `targets` arg description.

And don't worry, I'll take this text over once all my doubts have been cleared. Thanks

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
daavoo: I misunderstood. I thought you asked about non-dvc projects. 

Non-Git DVC projects do work

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
daavoo: > But so *targets is also optional, right?

It is optional, but the recommended syntax for the `*args` typing is to include only the valid type (in this case, str). `Optional[str]` would mean that `params_show(None)` would be a valid call, which is not.

content/docs/api-reference/params_show.md
daavoo: Done

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
daavoo: > dvc params CLI works with non-Git DVC projects I believe.

It doesn't because CLI only has `dvc params diff`, which makes sense to don't be supported in non-Git.
`metrics show` does work and `params show` (if/when added to CLI), should work

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,258 @@
+# dvc.api.params_show()
+
+Load <abbr>parameters</abbr> (name and values) tracked in a <abbr>DVC
+project</abbr>.
+
+```py
+def params_show(
+    *targets: str,  # Optional
+    repo: Optional[str] = None,
```
jorgeorpinel, line 9: WDYT about this comment to make it extra clear @daavoo, @dberenbaum ? Thanks

content/docs/api-reference/params_show.md
jorgeorpinel: No, that's clear enough. I didn't realize it's already mentioned. Thanks!

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,267 @@
...
+
+```json
+{
+  "split": 0.2,
+  "seed": 20170428
+}
+```
+
+Without arguments, this function will retrieve all params from all tracked
+parameter files (used in any `dvc.yaml` file). This applies to the current
```
jorgeorpinel, line 37: Please confirm this is correct @daavoo , thanks.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,240 @@
...
+params = dvc.api.get_params(stages="featurize")
+print(json.dumps(params, indent=2))
+```
+
+```json
+{
+  "featurize": {
+    "max_features": 200,
+    "ngrams": 2
+  }
```
jorgeorpinel, line 86: @daavoo in https://github.com/iterative/dvc/pull/7613#issuecomment-1158729997 you said "we don't include the stage name ... when loading a single stage's params" yet this example does show it. Can you confirm what's the behavior? Thanks

jorgeorpinel: Ooooh never mind, I read that comment again and I see it's just the structure of the params file.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,237 @@
...
+  (e.g. `[user@]server:project.git`). _Default_: The current project (found by
+  walking up from the current working directory tree).
+
+- `stages` - name or names of the stages to retrieve params from. If `None`, all
+  parameters from all stages will be retrieved. _Default_: `None`.
+
+- `rev` - Git commit (any [revision](https://git-scm.com/docs/revisions) such as
+  a branch or tag name, a commit hash or an
+  [experiment](/doc/command-reference/exp) name). If `repo` is not a Git repo,
+  this option is ignored. _Default_: `None` (current working tree will be used)
```
jorgeorpinel, line 60: > _Default_: `None` (current working tree will be used)

Is this the same for all the other functions BTW? Currently they state that the default is `HEAD`.

jorgeorpinel:
I've taken this over, PTAL @daavoo . Still needs an approval @dberenbaum @shcheklein . Thanks.

content/docs/api-reference/params_show.md
daavoo: 👍

content/docs/api-reference/params_show.md
daavoo: > Is this the same for all the other functions BTW? Currently they state that the default is HEAD.

Yes, it is the same for all the other functions, HEAD can be kind of misleading there

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
...
+```py
+import dvc.api
+
+params = dvc.api.params_show()
+```
+
+## Description
+
+Retrieves <abbr>params</abbr> keys and values from a <abbr>DVC project</abbr>
+and returns a JSON structure such as:
```
dberenbaum, line 27: Isn't it a Python dict and not a JSON structure?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
...
+```json
+{
+  "train": {
+    "batch_size": 128,
+    "hidden_units": 64,
+    "dropout": 0.4,
+    ...
+```
+
+## Example: Filter by stage name(s)
```
dberenbaum, line 167: Can this be moved up, preferably to the top or at least right below the `targets` example? Also, could you add some context for how this can be used inside a stage, or is that better left for a user guide doc?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
...
+Load <abbr>parameters</abbr> (name and values) tracked in a <abbr>DVC
+project</abbr>.
+
+```py
+def params_show(
+    *targets: str,  # Optional
+    repo: Optional[str] = None,
+    stages: Optional[Union[str, Iterable[str]]] = None,
+    rev: Optional[str] = None,
+    deps: bool = False,
```
daavoo, line 12: @dberenbaum @jorgeorpinel 

Should we drop `deps`? 

It doesn't feel that useful after adding support for `stages`

content/docs/api-reference/params_show.md
daavoo: > Also, could you add some context for how this can be used inside a stage, or is that better left for a user guide doc?

I included a minimal example on the very first version of the P.R. but it felt too much for API Reference.

content/docs/api-reference/params_show.md
jorgeorpinel: Extracted to #3681

content/docs/api-reference/params_show.md
jorgeorpinel: My bad. Fixed.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
...
+  [experiment](/doc/command-reference/exp) name). If `repo` is not a Git repo,
+  this option is ignored. _Default_: `None` (current working tree will be used)
+
+- `deps` - whether to retrieve only params that are stage dependencies. Accepts
+  `True` or `False` (_default_).
+
+## Example: Filter by parameter file(s)s
+
+> Working on
+> https://github.com/iterative/pipeline-conifguration/tree/main/multi-params-files
```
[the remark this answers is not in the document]
jorgeorpinel: Fixed here for now but

a) once renamed, conifguration would redirect to configuration I think (by GH);
b) more importantly: yes, let's definitely not use unofficial example repos in the docs.

> we don't appear to have any proper example
include a minimal example setup

We can keep them al hypothetical here. No need to make cmd ref example sreproducible, it's fine ⌛

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
+# dvc.api.params_show()
+
+Load <abbr>parameters</abbr> (name and values) tracked in a <abbr>DVC
+project</abbr>.
+
+```py
+def params_show(
+    *targets: str,  # Optional
+    repo: Optional[str] = None,
+    stages: Optional[Union[str, Iterable[str]]] = None,
```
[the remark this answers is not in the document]
jorgeorpinel: This should be the real function signature. Did it change?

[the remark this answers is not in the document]
jorgeorpinel: This is the signature [from what I can tell](https://github.com/iterative/dvc/pull/7613/files#diff-54fd2c4af8ff96a0ddfb2a22c1c56a659c41ae7eb263a6fb943b6d246cb34025R9-R15). I don't think we should alter it for docs (other than that one commend I added).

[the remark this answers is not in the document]
jorgeorpinel: That said if you plan to change the signature in the core repo then sure. Lmk

content/docs/api-reference/params_show.md
jorgeorpinel: Sure, we can have an ad-hoc order for examples. Moved all the way up (and ordered the remaining ones to match the parameter order).

> could you add some context for how this can be used inside a stage, or is that better left for a user guide doc?

Who is "you" here? 😬 Let's put it here as we don't have any guides with API usage and probably won't soon(?)

> included a minimal example

Again, no need to make things fully reproducible. I think @dberenbaum means some more sample code that explains what you would do with the params dict -- in this case set the model training method params I suppose?

content/docs/api-reference/params_show.md
jorgeorpinel: The only use case I see is to combine param file `targets` and `deps=True` to discover which params in a param file are used by the DVC project. If that's an intended feature we should probably add an example (or change the last one).

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
daavoo: Note that those are all keyword arguments so the order doesn't really affect the behavior. We could also update the order in core, if it makes more sense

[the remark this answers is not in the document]
dberenbaum: I don't mind updating in core, but to me it's more important to update here to start with more common use cases.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+## Example: Load specific parameter file(s)
+
+> Working on
+> https://github.com/iterative/pipeline-conifguration/tree/main/multi-params-files
+
+You can pass any valid param file path as target to load all of the parameters
+defined in it:
```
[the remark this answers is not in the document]
daavoo: It is accurate. Although I would suggest including a note here clarifying that the param file path doesn't necessarily need to be tracked by DVC

content/docs/api-reference/params_show.md
daavoo: > If that's an intended feature we should probably add an example (or change the last one).

I believe it is the original idea behind `deps` in the CLI. 

Note that you don't necessarily need to combine it with `targets`, it could be also useful to discover which params from *any* param file are used by the DVC project

content/docs/api-reference/params_show.md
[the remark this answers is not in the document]
jorgeorpinel: OK. I moved `stages` one position up. Why not all the way up then? Since deem the stage-filtering example most important... Which goes back to https://github.com/iterative/dvc/pull/7613#pullrequestreview-1009761660

> I don't mind updating in core

I think that's ideal because at some point this may get copy-pasted again from there (on some update) and our custom docs order is gone.

content/docs/api-reference/params_show.md
jorgeorpinel: If we don't see much use in retrospect, no need for an example. Yet OK to keep to reflect the CLI, IMO. Open an issue in core to deprecate `deps` from both CLI and API maybe?

content/docs/api-reference/params_show.md
jorgeorpinel: > more sample code that explains what you would do with the params dict

I tried doing this in 2a70b7dcf. PTAL @daavoo @dberenbaum 

In review app: https://dvc-org-api-get-params-2lnv5kl.herokuapp.com/doc/api-reference/params_show#example-filter-by-stage-names

content/docs/api-reference/params_show.md
dberenbaum: I think it will only retrieve params from `dvc.yaml` in the current directory, right? See https://github.com/iterative/dvc/issues/7926.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
+Without arguments, this function will retrieve all params from all tracked
+parameter files (used in any `dvc.yaml` file). This applies to the current
+project version when using Git (including any changes in the working tree).
+
+The function parameters (below) let you restrict what's retrieved.
+
+## Parameters
+
+- `*targets` - paths to one or more valid parameter file(s) to retrieve params
+  from. For example, `"params.py, myparams.toml"`. If no `targets` are provided,
```
dberenbaum, line 45:
```suggestion
  from. For example, `"params.py", "myparams.toml"`. If no `targets` are provided,
```

TBH I would suggest removing that sentence. I don't see similar examples for targets in the CLI, and I think people should know how Python positional args work?

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
+  [experiment](/doc/command-reference/exp) name). If `repo` is not a Git repo,
+  this option is ignored. _Default_: `None` (current working tree will be used)
+
+- `deps` - whether to retrieve only params that are stage dependencies. Accepts
+  `True` or `False` (_default_).
+
+## Example: Filter by stage name(s)
+
+> Working on https://github.com/iterative/example-get-started, file
+> `src/featurization.py`.
```
dberenbaum, line 68: I think we are better off without referencing an example repo because:

1. Users probably aren't familiar with the repo like we are and shouldn't have to check out the repo to understand the context.
2. We probably should update the example repo with `dvc.api.params_show()`, and then we will have to update this ref again.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
+## Example: Use a remote DVC repository
+
+You can use the `repo` argument to retrieve parameters from any <abbr>DVC
+repository</abbr> without having to clone it locally.
+
+```py
+import json
+import dvc.api
+params = dvc.api.get_params(
+    repo="https://github.com/iterative/demo-fashion-mnist")
```
dberenbaum, line 158: Can we use example-get-started here since it's stable and already a docs dependency?

content/docs/api-reference/params_show.md
dberenbaum: To clarify, all the example content underneath looks good. I just don't think we should reference that it's based on example-get-started.

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
+`stages` can be a single name (string):
+
+```git
++import json
+-import yaml
+
+-params = yaml.safe_load(open("params.yaml"))["featurize"]
++import dvc.api
++
++params = dvc.api.get_params(stages="featurize")
```
daavoo, line 79:
```suggestion
+params = dvc.api.params_show(stages="featurize")
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
++max_features = params["featurize"]["max_features"]
++ngrams = params["featurize"]["ngrams"]
+```
+
+Or an iterable of strings:
+
+```py
+import json
+import dvc.api
+params = dvc.api.get_params(stages=["featurize", "train"])
```
daavoo, line 94:
```suggestion
params = dvc.api.params_show(stages=["featurize", "train"])
```

content/docs/api-reference/params_show.md
```diff
@@ -0,0 +1,198 @@
...
+## Example: Specify a project version
+
+> Working on https://github.com/iterative/example-get-started
+
+You can retrieve params from arbitrary Git commits, for example a branch name:
+
+```py
+import json
+import dvc.api
+params = dvc.api.get_params(rev="tune-hyperparams")
```
daavoo, line 179:
```suggestion
params = dvc.api.params_show(rev="tune-hyperparams")
```

content/docs/api-reference/params_show.md
jorgeorpinel: PTAL

content/docs/api-reference/params_show.md
jorgeorpinel: > all the example content underneath looks good

But rn the example uses a git-diff style so it's supposed to be a change based on something:

image

@daavoo mind rewriting this part so no reference is needed yet the usage of the function is explained in the code snippet? Prob no need for git-diff block after all.

content/docs/api-reference/params_show.md
jorgeorpinel: Agree.

PR #3467, opened 2022-04-27
gc: note on cleaning data in the shared cache / remote case

### LINKED ISSUE #3464: how: mention how to `gc` in share-a-dvc-cache
See https://dvc.org/doc/user-guide/how-to/share-a-dvc-cache

Running `gc` with shared cache setup can be dangerous (i.e. https://github.com/iterative/dvc/issues/7610)

We should mention the appropriate options (i.e. https://github.com/iterative/dvc/issues/7610#issuecomment-1104832471)

shcheklein:
Closes #3464 


daavoo (commented):
Added a reference to this content in the user guide.

TBH I'm not sure whether the whole section belongs to the user guide more than in the command reference

shcheklein:
ping @jorgeorpinel @iesahin :)

content/docs/command-reference/gc.md
```diff
@@ -44,8 +44,29 @@ If the `--cloud` (`-c`) flag is used, this command deletes unused data from the
...
+Cloud deletion is irreversible unless there is another DVC remote or a manual
+backup with the same data.
+
+</admon>
+
+### Cleaning shared cache or shared remote
+
+If a single remote or a single [cache is shared] among different projects and
+those projects have overlapped DVC-tracked files, removing files via `gc` in one
+project may break some links in another project.
```
jorgeorpinel, line 58:
```suggestion
### Cleaning shared cache (or remote)

If a [cache is shared] among different projects that
track some of the same files, using `dvc gc` in one
project will break those overlapping data links in the other projects.
```

I'm removing "remote" because we don't really mention combining this with the `-r` option, which would further complicate the explanation. Hopefully people can infer this from the title.

content/docs/command-reference/gc.md
```diff
@@ -44,8 +44,29 @@ If the `--cloud` (`-c`) flag is used, this command deletes unused data from the
...
+
+To keep overlapping files `dvc gc` has a flag `--projects` (`-p`) that takes
+multiple path to DVC projects. DVC _keeps_ data referenced in those projects.
+Make sure that all the commits (including branches) that are referencing any
+files to be preserved have been pulled before running this command.
```
jorgeorpinel, line 63:
```suggestion
To prevent this, use the `--projects` (`-p`) option. It takes one or more
paths to the DVC projects whose data should be preserved.
Make sure that all the commits and branches that reference
files you want to keep have been pulled in those other projects first.
```

content/docs/command-reference/gc.md
```diff
@@ -44,8 +44,29 @@ If the `--cloud` (`-c`) flag is used, this command deletes unused data from the
...
+
+For example, if we have `N` projects that have some overlapping files and we'd
+like to remove all data in one of those (let's say we don't need one of the
+projects anymore), we would need first to clone all the remaining `N-1` projects
+(including branches) and pass their path to the `dvc gc -p` command while
+running it from the project we'd like to delete the data from.
```
jorgeorpinel, line 69:
```suggestion
For example, if we have 4 projects with some overlapping files and we'd
like to remove all the data that's only needed for one of them (e.g. when we no longer need that projects),
we would first clone all the other projects and fetch all their branches,
then pass their paths to the `dvc gc -p` command from
the project we want to clear.
```

jorgeorpinel (commented):
Wording updates for the explanation on collecting shared caches:

content/docs/command-reference/gc.md
jorgeorpinel: p.s. Wow this is a complicated process. I wonder whether there could be a better design e.g. where external caches"knew" what projects use them so you could simply supply `--project` to collect only the data used in the present project cc @dberenbaum

content/docs/command-reference/gc.md
jorgeorpinel: Maybe we can move this to an actual example instead of providing an example in a subsection of the description. Or skip the example.

content/docs/command-reference/gc.md
shcheklein: `shared remote` can be good for search

content/docs/command-reference/gc.md
```diff
@@ -145,3 +164,6 @@ Let's check the size now:
 $ du -sh .dvc/cache/
 3.1G    .dvc/cache/
 ```
+
+[cache is shared]: /doc/user-guide/how-to/share-a-dvc-cache
+[dvc experiments]: /doc/user-guide/experiment-management#experiments
```
jorgeorpinel, line 169: 💅🏼 Ideally we should place these relinks at the end of the **Options** section (before the **Examples** H2), where they're used.

content/docs/command-reference/gc.md
shcheklein: may be share cache (or remote) is enough - not sure (I prefer usually explicit vs optimized wording)

shcheklein: anyway, Jorge, please take it over - commit suggestions as you see and let's merge this. This kind of things we should be in 5 minutes ideally :)

content/docs/command-reference/gc.md
shcheklein: mentioning 4 and not mentioning 3 after has less sense. Then it's better to use `we have a few project` ... (I was trying to make a bit more explicit). Again, not a strong opinion at all ... let's merge all this stuff

content/docs/user-guide/how-to/share-a-dvc-cache.md
```diff
@@ -87,3 +91,11 @@ If you're using Git, commit the changes to your project's config file (usually
 $ git add .dvc/config
 $ git commit -m "config external/shared DVC cache"
 ```
+
+<admon type="warn">
+
+When using a shared cache, running `dvc gc` in one project may break some links
+in another project. See more info about
+[cleaning a shared cache](/doc/command-reference/gc#cleaning-shared-cache-or-shared-remote).
```
jorgeorpinel, line 99:
```suggestion
Using `dvc gc` with a shared cache may delete data needed
in another project! See more info. about
[cleaning a shared cache](/doc/command-reference/gc#cleaning-shared-cache-or-shared-remote)
safely.
```

shcheklein: `dvc gc` doesn't have shared cache (at least it's harder to read it this way) ...` using `dvc gc` in a project(s) with a shared cache ...`

content/docs/command-reference/gc.md
jorgeorpinel: changed to "several"

jorgeorpinel (approved):
Thanks @shcheklein @daavoo !

content/docs/command-reference/gc.md
dberenbaum: Yup, something similar was mentioned by @iesahin about remote in https://github.com/iterative/dvc/discussions/6653. We could open a separate issue for that specific request. It's probably not something we are likely to prioritize soon since this is a command that's hopefully used sparingly, and there are a lot of complexities in solving it, but I agree it's not ideal as is.

PR #3469, opened 2022-04-27
Keep footer at bottom of page

### LINKED ISSUE #2430: Footer not staying at bottom of page in 404 page
There is a gap between the footer and the bottom of the page, starting at `610px` screen width:

image

julieg18:
* updates our `MainLayout` component and `Footer` components

Fixes #2430

### DIFF
--- a/src/components/LayoutFooter/styles.module.css
+++ b/src/components/LayoutFooter/styles.module.css
@@ -1,6 +1,7 @@
 .wrapper {
   width: 100%;
   background-color: #40364d;
+  margin-top: auto;
 }
 
 .div {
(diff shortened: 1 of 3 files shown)


julieg18:
While it's working locally, I'm getting this on the gatsby build: 

image

@rogermparent, is it just a bug with Gatsby deployment?

rogermparent:
I think Gatsby Cloud has been having some intermittent issues, I had a similar problem yesterday for an hour or so. I can access the site now so I'm guessing the issue will be resolved for you as well.

rogermparent (approved):
Nice work! I've been meaning to add this style for a while, but keep forgetting to do so since it's encountered so infrequently.

PR #3473, opened 2022-04-28
gdrive remote service accounts can use env var

shcheklein:
For service accounts it is also an option to pass keys via an env variable.

### DIFF
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -199,7 +199,7 @@ authentication is needed.
    key in CI/CD systems, production setup, read-only file systems, etc. The
    content of this variable should be a string with JSON that has the same
    format as in the keys file described above. If both this variable and
-   `gdrive_service_account_json_file_path` are provide,
+   `gdrive_service_account_json_file_path` are provided,
    `GDRIVE_CREDENTIALS_DATA` takes priority and
    `gdrive_service_account_json_file_path` is ignored.
 
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -199,8 +199,8 @@ authentication is needed.
    key in CI/CD systems, production setup, read-only file systems, etc. The
    content of this variable should be a string with JSON that has the same
    format as in the keys file described above. If both this variable and
-   `gdrive_service_account_json_file_path` are provide, DVC reads the
-   environment variable first.
+   `gdrive_service_account_json_file_path` are provide, `GDRIVE_CREDENTIALS_DATA`
+   takes priority and `gdrive_service_account_json_file_path` is ignored.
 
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -195,6 +195,13 @@ authentication is needed.
                  gdrive_service_account_json_file_path path/to/file.json
    ```
 
+   Alternatively, a `GDRIVE_CREDENTIALS_DATA` can be set to pass service account
+   key in CI/CD systems, production setup, read-only file systems, etc. The
+   content of this variable should be a string with JSON that has the same
+   format as in the keys file described above. If both this variable and
+   `gdrive_service_account_json_file_path` are provide, DVC reads the
+   environment variable first.
```
daavoo, line 203: I think that providing `gdrive_service_account_json_file_path` is no longer required after https://github.com/iterative/dvc/pull/7213

daavoo: > I read the section as saying that either gdrive_service_account_json_file_path or GDRIVE_CREDENTIALS_DATA is required.

I understood the phase wrong 🙏

shcheklein: @dberenbaum @daavoo thanks! Done, I've changed the message a bit - pls take a look.

shcheklein: > Finally, should we promote GDRIVE_CREDENTIALS_DATA as the primary method 

Make sense, but let's keep it as is for now.  Service accounts can be used (and are used) locally as well ...

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -195,6 +195,14 @@ authentication is needed.
                  gdrive_service_account_json_file_path path/to/file.json
    ```
 
+   Alternatively, a `GDRIVE_CREDENTIALS_DATA` can be set to pass service account
+   key in CI/CD systems, production setup, read-only file systems, etc. The
+   content of this variable should be a string with JSON that has the same
+   format as in the keys file described above. If both this variable and
+   `gdrive_service_account_json_file_path` are provide,
```
dberenbaum, line 202:
```suggestion
   `gdrive_service_account_json_file_path` are provided,
```

dberenbaum:
One more thought that doesn't need to block this PR: should the "Authorization" section be moved above "Using service accounts"? I find it confusing to go through a section about service accounts and then have a section after it focused back on user accounts without being clear that it's a different use case from the above section.

shcheklein:
> One more thought that doesn't need to block this PR

Yep, it's a good point also, but let me do this a separate PR though.

shcheklein:
@jorgeorpinel @iesahin ping :)

daavoo (approved):

shcheklein:
@dberenbaum here is the followup https://github.com/iterative/dvc.org/pull/3473 :)

PR #3486, opened 2022-04-29
Update service account creation flow

RCdeWit:
Noticed the description on how to create a service account was somewhat outdated. The flow is now somewhat different on GCP; updated the docs to reflect that.


shcheklein (approved):

shcheklein:
thanks @RCdeWit !

PR #3495, opened 2022-04-30
Follow up to improve gdrive and move auth before service accounts

shcheklein:
Followup based on https://github.com/iterative/dvc.org/pull/3473#issuecomment-1112659711

-----------------------------

### DIFF
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -167,9 +167,9 @@ $ dvc remote modify myremote gdrive_client_secret 'client-secret'
 
 <admon type="info">
 
-This covers simple authentication, which gives DVC access to GDrive
-on behalf of a user account. This is ideal to use to run DVC locally, for example.
-If some automation is needed (e.g. CI/CD) we recommend
+This covers simple authentication, which gives DVC access to GDrive on behalf of
+a user account. This is ideal to use to run DVC locally, for example. If some
+automation is needed (e.g. CI/CD) we recommend
 [using a service account](#using-service-accounts) instead.
 
 </admon>
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -250,6 +250,14 @@ authentication is needed.
                  gdrive_service_account_json_file_path path/to/file.json
    ```
 
+   Alternatively, a `GDRIVE_CREDENTIALS_DATA` can be set to pass service account
+   key in CI/CD systems, production setup, read-only file systems, etc. The
+   content of this variable should be a string with JSON that has the same
+   format as in the keys file described above. If both this variable and
+   `gdrive_service_account_json_file_path` are provided,
+   `GDRIVE_CREDENTIALS_DATA` takes priority and
+   `gdrive_service_account_json_file_path` is ignored.
+
 3. Share the Google Drive folders that you want to use with the service account.
    Navigate to your Google Drive folder's sharing options and add the service
    account as an editor (read/write) or viewer (read-only):
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -165,11 +165,14 @@ $ dvc remote modify myremote gdrive_client_secret 'client-secret'
 
 ## Authorization
 
-There are two ways to give DVC access to a GDrive directory or a disk. The one
-descibed in this section gives DVC a way to access GDrive on behalf of a user
-account. It's the simplest way to use, and it's fine to use to run DVC locally.
-If some automation is needed (e.g. CI/CD) we recommend to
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -163,56 +163,14 @@ $ dvc remote modify myremote gdrive_client_secret 'client-secret'
...
-![](/img/gdrive-share-with-service-account.png)
-
 ## Authorization
 
+There are two ways to give DVC access to a GDrive directory or a disk. The one
+descibed in this section gives DVC a way to access GDrive on behalf of a user
+account. It's the simplest way to use, and it's fine to use to run DVC locally.
+If some automation is needed (e.g. CI/CD) we recommend to
+[use a service account](#using-service-accounts) instead.
+
```
shcheklein, line 173: One extra paragraph above

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -254,3 +212,43 @@ authentication process. If `GDRIVE_CREDENTIALS_DATA` is set, the
 `gdrive_user_credentials_file` value (if provided) is ignored.
 
 > Please note our [Privacy Policy (Google APIs)](/doc/user-guide/privacy).
+
+## Using service accounts
```
shcheklein, line 216: This section is unchanged and only moved to the bottom

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -163,56 +163,14 @@ $ dvc remote modify myremote gdrive_client_secret 'client-secret'
...
-
-![](/img/gdrive-share-with-service-account.png)
-
 ## Authorization
 
+There are two ways to give DVC access to a GDrive directory or a disk. The one
+descibed in this section gives DVC a way to access GDrive on behalf of a user
+account. It's the simplest way to use, and it's fine to use to run DVC locally.
+If some automation is needed (e.g. CI/CD) we recommend to
+[use a service account](#using-service-accounts) instead.
```
jorgeorpinel, line 172:
```suggestion
<admon type="info">

This covers simple authentication, which gives DVC access to GDrive
on behalf of a user account. This is ideal to use to run DVC locally, for example.
If some automation is needed (e.g. CI/CD) we recommend
[using a service account](#using-service-accounts) instead.

</admon>
```

content/docs/user-guide/setup-google-drive-remote.md
jorgeorpinel: Not exactly unchanged: It's missing the `GDRIVE_CREDENTIALS_DATA` paragraph, but that's already in **Authentication** which is now before this section. Is this the intention? (To have it only above)

jorgeorpinel (approved):

dberenbaum:
Looking at it again, I think "Authorization" should even be above "Using a custom Google Cloud Project" (sorry for not catching this earlier). So the structure would be:

1. Quick start
2. URL format - provides more detail on the quick start scenario
3. Authorization - provides more detail on the quick start scenario and introduces auth options for advanced scenarios below
4. Using a custom Google Cloud Project - explains advanced best practice that is needed to use service accounts
5. Using service accounts - explains advanced best practice for automation

Thoughts?

Happy to make the changes myself if it's easier once we all agree.

content/docs/user-guide/setup-google-drive-remote.md
dberenbaum: Right, especially with the new paragraph about specifying that the "Authorization" section is about user accounts, it seems unclear to me that `GDRIVE_CREDENTIALS_DATA` can also be used for service accounts.

shcheklein: > Not exactly unchanged:

good catch, that's a bug, I'll fix it. It's not intentional at all.

shcheklein:
> Looking at it again, I think "Authorization" should even be above "Using a custom Google Cloud Project"

I think that was intentional, so that it's not considered as _advanced_ but rather as recommended (in most cases you would want to have an independent app so that your rate limits don't overlap with other DVC users).

Also, it's an optional configuration step, while Auth follows both of them.

So, I don't see changing the order as a fixing something step, but I'm fine with that if you guys decide to this.

dberenbaum:
Sounds good. I'm not sure the GCP -> Auth -> Service Accounts order is any better than the current GCP -> Service Accounts -> Auth order, so I'm open to closing, merging, or reordering as Auth -> GCP -> Service Accounts. @jorgeorpinel If you have any opinion, happy to go with that.

jorgeorpinel:
Sounds like service accounts are "more advanced" so it makes sense to have them last. If it's the recommended method then maybe we should be more explicit about that and move them back up. But it's probably not worth going back to it now... We have to maintain this periodically anyway so I'm sure we'll have another chance in the future.

> TBH ideally I'd prefer if we didn't have to be so specific with 3rd party tool docs.

PR #3503, opened 2022-05-03
Plausible proxy update

### LINKED ISSUE #3501: Plausible proxy through express server is not working as expected
We added a Plausible proxy with #3423 

Testing the website, the `script` is loaded and `event` POST requests are sent to the server with 202 `Ok` responses from the server. 

The script is available on [/api/pl/script.outbound-links.js](https://dvc.org/api/pl/script.outbound-links.js)
And, the event is sent to [/api/event](https://dvc.org/api/event) with POST request.

yathomasi:
Using [`http-proxy-middleware`](https://github.com/chimurai/http-proxy-middleware) for handling proxy middleware for Plausible.

Fixes #3501


yathomasi:
Current gatsby cloud build doesn't support proxy. For my local testing, it seems to be working. I can see the stats are pumped up specially `Unique Visitors` which seems to be one that's not updated.

julieg18 (commented):
Looks good! Hopefully, this fixes things!

julieg18 (approved):

PR #3525, opened 2022-05-09
Add action for `master` automatic deployment

julieg18:
This pr is a draft since I can't add secrets to dvc.org.

### DIFF
--- a/.github/workflows/deploy-master.yaml
+++ b/.github/workflows/deploy-master.yaml
@@ -1,0 +1,19 @@
+name: Deploy Master
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: akhileshns/heroku-deploy@v3.12.12
+        with:
+          heroku_api_key: ${{secrets.HEROKU_API_KEY}}
+          heroku_app_name: 'dvc-org'
+          heroku_email: 'juliannagal18@gmail.com'
+          buildpack: 'https://github.com/heroku/heroku-buildpack-static.git'
+          dontautocreate: true


.github/workflows/deploy-master.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: akhileshns/heroku-deploy@v3.12.12
+        with:
+          heroku_api_key: ${{secrets.HEROKU_API_KEY}}
```
julieg18, line 15: Since I don't have admin access on `dvc.org`, I wasn't able to add my account API key to the secrets. @iterative/websites, anyone with admin access that can help me add this?

yathomasi (approved):
I see it's working for iterative.ai. Good work @julieg18 🎉 
Looks good to me.

PS: I also don't have access.

.github/workflows/deploy-master.yaml
```diff
@@ -0,0 +1,19 @@
...
+          heroku_app_name: 'dvc-org'
+          heroku_email: 'juliannagal18@gmail.com'
+          buildpack: 'https://github.com/heroku/heroku-buildpack-static.git'
```
rogermparent, line 18: I think this buildpack can cause issues with the custom Express server, which is primarily an issue on dvc.org because of how we host our APIs. Worth a try, but we'll have to make sure those are working on these deploys. Might need to be changed to [`heroku/nodejs`](https://elements.heroku.com/buildpacks/heroku/heroku-buildpack-nodejs), not sure if it must be a Git link or if the official buildpacks are exempt, but the docs offer both.

julieg18: I'll give it a try! We do have an `app.json`, making this key invalid 🤔

PR #3564, opened 2022-05-18
Create Blog “2022-05-18-the-right-to-be-forgotten-with-dvc”

jendefig:
Automatically generated by Netlify CMS


yathomasi:
> @yathomasi Can we automatically set up the checklist message that goes into the opening? Can we automatically put the blog tag on?

It doesn't seem to be customizable. I couldn't find any documented way but trying to find one.

PR #3579, opened 2022-05-21
docs: update docs for YAML 1.2 compatibility mitigation

bobertlo:
See https://github.com/iterative/dvc/issues/5971

adds note to parameter values section metioning scientific notation for SEO
use ruamel.yaml in example instead of PyYAML and add note

### DIFF
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -58,6 +58,10 @@ as the tree path to find those values. Supported types are: string, integer,
 float, and arrays (groups of params). Note that DVC does not ascribe any
 specific meaning to these values.
 
+> YAML 1.2 stores very large and very small numbers in scientific notation, but
+> the popular PyYAML library uses an older version of the format. To avoid
+> introducing subtle bugs, the ruamel.yaml library should be used instead.
+
 DVC saves parameter names and values to `dvc.lock` in order to track them over
 time. They will be compared to the latest params files to determine if the stage
 is outdated upon `dvc repro` (or `dvc status`).
@@ -115,15 +119,20 @@ The `train.py` script will have some code to parse and load the needed
 parameters. For example:
 
 ```py
-import yaml
+from ruamel.yaml import YAML
 
 with open("params.yaml", 'r') as fd:
-    params = yaml.safe_load(fd)
+    yaml = YAML()
+    params = yaml.load(fd)
 
 lr = params['lr']
 epochs = params['train']['epochs']
 layers = params['train']['layers']
 ```
+
+> Note that the popular PyYAML library does not support YAML 1.2. The
+> ruamel.yaml library should be used instead to avoid subtle differences in
+> number handling.
 
 You can find that each parameter was defined in `dvc.yaml`, as well as saved to
 `dvc.lock` along with the values. These are compared to the params files when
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/params/index.md
```diff
@@ -115,16 +119,20 @@ The `train.py` script will have some code to parse and load the needed
 parameters. For example:
 
 ```py
-import yaml
+from ruamel.yaml import YAML
```
dberenbaum, line 122: It looks like there's a couple other cases of `import yaml` in `stage add` and `run` docs. Would you mind making similar changes there?

content/docs/command-reference/params/index.md
```diff
@@ -58,6 +58,10 @@ as the tree path to find those values. Supported types are: string, integer,
 float, and arrays (groups of params). Note that DVC does not ascribe any
 specific meaning to these values.
 
+> YAML 1.2 stores very large and very small numbers in scientific notation. Some
```
dberenbaum, line 61: I like that you specify a common problem here with scientific notation. In the note below, I like that you specify that PyYAML doesn't support YAML 1.2. Maybe you can combine them and have pretty much the same note in both places?

It would also be great to suggest solutions in both places (ruamel.yaml, JSON or other file formats).

dberenbaum:
Thank you @bobertlo! I left some comments but also open to merging as is and iterating since it's already an improvement.

bobertlo:
Sure! I had similar thoughts about the structure of the changes and wasn’t
sure how to approach it.

How would you feel about making “parameter values” a (linkable) heading
with a full description and in a note offering solutions and the code
examples could just mention PyYAML briefly after the examples and link back?

I could send a revision back for review with that later and I will look for
the other examples using PyYAML at least to include them in this review.

content/docs/command-reference/params/index.md
dberenbaum: It might be nice since it saves a `pip install`, but it might also look like it's required to use the DVC API.

@daavoo Let's use `ruamel.yaml` and make a separate issue to discuss `dvc.utils.serialize`.

dberenbaum:
@bobertlo I'm not sure without seeing it but could be nice. As I mentioned before, I'm open to merging as is or in any form that's better than what we have today (no warning about YAML incompatibility).

bobertlo:
Maybe it would be good to just merge this and the other points can be
addressed separately.

My first concern is catching the “dvc parameters scientific notation”
Google search with a solution ;)

dberenbaum (approved):

bobertlo:
I did my best to reconcile some of the ideas mentioned and got the other two places where `import yaml` was included in examples.

Feel free to make any language edits you feel are helpful if needed.

content/docs/command-reference/params/index.md
bobertlo: I thought a little more on this and realized `ruamel.yaml` is a DVC dependency, so it would also be installed. Whether it is better to provide a specific external library or an internal DVC function is a tough question, and neither looks as good as `import yaml` unfortunately.

skshetry: > It might be nice since it saves a `pip install`, but it might also look like it's required to use the DVC API.
> 
> @daavoo Let's use `ruamel.yaml` and make a separate issue to discuss `dvc.utils.serialize`.

It's an internal API, let's not do anything.

PR #3585, opened 2022-05-24
ref: exp push/pull experiment metavar

dberenbaum:
Small fix to `exp push/pull`


shcheklein (changes requested):
I think we can update the text a bit:

` as well as an experiment name or hash` -> ` as well as one or more experiment names or hashes`?

Can please we also review a few other places where we refer to the `experiment` and see if we need to adjust the text around?

shcheklein (approved):

PR #3599, opened 2022-05-28
nav: updates

jorgeorpinel:
- Standard and optimized sidebar.json structure
- Fix some problems and typos related to the changes

> Relates to #144 and prepares the nav for VSCode Extension docs

- [ ] Apply similar changes to MLEM.ai/doc


content/docs/sidebar.json
```diff
@@ -97,16 +88,22 @@
...
-        "slug": "fast-data-caching-hub"
+        "slug": "ml-experiment-tracking",
+        "label": "ML Experiment Tracking"
       },
-      "experiment-tracking",
-      "model-registry",
-      "data-registry"
+      {
+        "slug": "ml-model-registry",
+        "label": "ML Model Registry"
```
jorgeorpinel, line 105: Hi @rogermparent qq: where do we add words for default capitalization from sidebar slugs? I'd like to make `ml`->"ML". Thanks

content/docs/sidebar.json
```diff
@@ -524,33 +506,11 @@
     "type": "external"
   },
   {
-    "label": "CML",
     "slug": "cml",
-    "source": "cml/index.md",
+    "label": "CML",
     "icon": "cml",
-    "children": [
```
jorgeorpinel, line 531: `cml` -> "CML" would also be good to auto-capitalize. Plus how is this external link implemented now? The nav still has a substructure. Seems like it's hard-coded somewhere? We can remove that 🙂 cc @rogermparent thanks

content/docs/sidebar.json
julieg18: Currently the code is written in a way that only `dvc` can be uppercased. But if we want to be able uppercase more than just DVC, it should be doable with a little code refactoring! I'll open a issue for it.

content/docs/sidebar.json
julieg18: If I'm understanding correctly, we have a line in [`redirects.json`](https://github.com/iterative/dvc.org/blob/master/redirects-list.json#L78) that converts the `/doc/cml` to `https://cml.dev/doc`.

content/docs/sidebar.json
julieg18: #3602 has been merged and `ml` should become `ML` automatically :)

jorgeorpinel: Awesome, thanks Julie! Should this be a feature of the docs engine in general? CML and MLEM for example could use this 🙂

julieg18: Yes! Already opened a pr for adding this feature to our doc engine theme (See gatsby-theme-iterative#22)

jorgeorpinel: `dvclive` - > "DVCLive" ? Or is that too much

shcheklein: a bit too much, Jorge (how many places do we have)?

jorgeorpinel: A few, same as the other ones. But yes never mind, it complicates things conceptually.

content/docs/sidebar.json
```diff
@@ -523,75 +499,49 @@
     "type": "external"
   },
   {
-    "label": "CML",
     "slug": "cml",
```
jorgeorpinel, line 502: For some reason removing the "CML" label hides the nav entry @julieg18 . All other auto-upper-case seem to work fine. Probably related with the hardcoded redirect...

image

jorgeorpinel:
So anyway! This was a clean-up / proof of concept / test. I need to split it later (not a priority). Thanks though!

content/docs/sidebar.json
jorgeorpinel: Good to know, thanks (fixed in a74facb74).

> We should def. document the doc engine features at some point 😬 rel. https://github.com/iterative/gatsby-theme-iterative/issues/17

content/docs/sidebar.json
```diff
@@ -523,75 +499,49 @@
...
+            "slug": "prepare-repositories",
+            "label": "Prepare Your Repositories"
           },
           {
-            "label": "Views and Experiments",
             "slug": "views",
+            "label": "Views and Experiments",
             "source": false,
             "children": [
               "what-is-a-view",
```
jorgeorpinel, line 529: This whole sub-section is too buried, especially while the docs are inside dvc.org/doc.

- [ ] Probably we could re-combine some of these pages and put them directly under the Studio UG.

PR #3602, opened 2022-05-30
Let sidebar uppercase more than `dvc` in labels made from slugs

### LINKED ISSUE #3601: Allow more than `dvc` to be uppercased in our `sidebar.json` slugs
Currently, our sidebar automatically uppercases the word `dvc` in our `sidebar.json` slugs (aka the slug [`running-dvc-on-windows`](https://github.com/iterative/dvc.org/blob/master/content/docs/sidebar.json#L179) gets the title of [`Running DVC on Windows`](https://dvc.org/doc/user-guide/running-dvc-on-windows)).

We want to be able to uppercase more than just dvc though ([`api`, `cml`](https://github.com/iterative/dvc.org/pull/3599#discussion_r884224938), [`ml`](https://github.com/iterative/dvc.org/pull/3599#discussion_r884190953), etc.).

Should be doable with a little `sidebar.js` refactoring!

julieg18:
Fixes #3601

### DIFF
--- a/content/docs/sidebar.json
+++ b/content/docs/sidebar.json
@@ -620,7 +620,6 @@
         "slug": "dvclive-with-dvc"
       },
       {
-        "label": "ML Frameworks",
         "slug": "ml-frameworks",
         "source": "dvclive/ml-frameworks/index.md",
         "children": [
@@ -667,7 +666,6 @@
         ]
       },
       {
-        "label": "API Reference",
         "slug": "api-reference",
         "source": "api-reference/index.md",
         "children": [
--- a/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
+++ b/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
@@ -23,11 +23,7 @@
 const { titleCase } = require('title-case')
 const sidebar = require('../../../sidebar')
 
-const UPPERCASE_KEYWORDS = ['dvc', 'cml', 'api', 'mlem', 'ml', 'ldb']
-const UPPERCASE_KEYWORDS_REGEX = new RegExp(
-  String.raw`${UPPERCASE_KEYWORDS.join('|')}`,
-  'g'
-)
+const UPPERCASE_KEYWORDS_REGEX = /dvc|cml|api|mlem|ml|ldb/g
 const PATH_ROOT = '/doc'
 const FILE_ROOT = '/docs/'
 const FILE_EXTENSION = '.md'
--- a/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
+++ b/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
@@ -37,8 +37,6 @@ function uppercaseSlugKeywords(slug) {
   return newSlug
 }
 
-console.log(uppercaseSlugKeywords('ml-mlem-dvc-test-ldb-api-cml'))
-
 function slugTitleCase(slug) {
   return titleCase(uppercaseSlugKeywords(slug).replace(/-/g, ' '))
 }
--- a/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
+++ b/plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
@@ -23,7 +23,7 @@
 const { titleCase } = require('title-case')
 const sidebar = require('../../../sidebar')
 
-const uppercaseKeywords = ['dvc', 'cml', 'api', 'ml']
+const uppercaseKeywords = ['dvc', 'cml', 'api', 'mlem', 'ml', 'ldb']
 const PATH_ROOT = '/doc'
 const FILE_ROOT = '/docs/'
 const FILE_EXTENSION = '.md'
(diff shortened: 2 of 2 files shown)


plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
```diff
@@ -23,12 +23,22 @@
...
 const sidebar = require('../../../sidebar')
 
+const uppercaseKeywords = ['dvc', 'cml', 'api', 'ml']
 const PATH_ROOT = '/doc'
 const FILE_ROOT = '/docs/'
 const FILE_EXTENSION = '.md'
 
-function dvcTitleCase(slug) {
-  return titleCase(slug.replace(/dvc/g, 'DVC').replace(/-/g, ' '))
+function uppercaseSlugKeywords(slug) {
```
julieg18, line 31: If this change is approved, I can open a pr to update our docs theme :)

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
```diff
@@ -23,12 +23,22 @@
 const { titleCase } = require('title-case')
 const sidebar = require('../../../sidebar')
 
+const uppercaseKeywords = ['dvc', 'cml', 'api', 'ml']
```
julieg18, line 26: @jorgeorpinel, these are the keywords that will be uppercased. Does the list look good?

content/docs/sidebar.json
```diff
@@ -620,7 +620,6 @@
         "slug": "dvclive-with-dvc"
       },
       {
-        "label": "ML Frameworks",
```
julieg18, line 623: Since the sidebar code automatically uppercases `ml` and `api`, these two labels can be taken off.

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
shcheklein: let's add MLEM, LDB

julieg18: List is now `['dvc', 'cml', 'api', 'mlem', 'ml', 'ldb']` :)

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
```diff
@@ -23,12 +23,24 @@
...
+function uppercaseSlugKeywords(slug) {
+  let newSlug = slug
+  uppercaseKeywords.forEach(word => {
+    const regex = new RegExp(String.raw`${word}`, 'g')
+    newSlug = newSlug.replace(regex, word.toUpperCase())
+  })
+  return newSlug
+}
+
+console.log(uppercaseSlugKeywords('ml-mlem-dvc-test-ldb-api-cml'))
```
shcheklein, line 40: do we need this?

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
```diff
@@ -23,12 +23,24 @@
...
+const uppercaseKeywords = ['dvc', 'cml', 'api', 'mlem', 'ml', 'ldb']
 const PATH_ROOT = '/doc'
 const FILE_ROOT = '/docs/'
 const FILE_EXTENSION = '.md'
 
-function dvcTitleCase(slug) {
-  return titleCase(slug.replace(/dvc/g, 'DVC').replace(/-/g, ' '))
+function uppercaseSlugKeywords(slug) {
+  let newSlug = slug
+  uppercaseKeywords.forEach(word => {
```
shcheklein, line 33: can we optimize it a bit please? combine into one regex? `regex = [words].join("|")` (this is not an actual lang, just to communicate the idea). Also, pre-compile the regex (not sure is JS has a cache of complied regexps - need to check).

shcheklein (changes requested):
Good stuff, @julieg18  - some optimization I think is possible that makes sense + some cleanup (?)

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
julieg18: Whoops, that was accidentally committed! Deleted now!

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
julieg18: Good idea! Forgot that we can easily just use one regex since `replace` has different options for the second argument. 

I also moved the regex outside of the function so that it isn't recompiled on every function run :)

plugins/gatsby-theme-iterative-docs/src/utils/shared/sidebar.js
```diff
@@ -23,12 +23,21 @@
...
+const UPPERCASE_KEYWORDS_REGEX = /dvc|cml|api|mlem|ml|ldb/g
 const PATH_ROOT = '/doc'
 const FILE_ROOT = '/docs/'
 const FILE_EXTENSION = '.md'
 
-function dvcTitleCase(slug) {
-  return titleCase(slug.replace(/dvc/g, 'DVC').replace(/-/g, ' '))
+function uppercaseSlugKeywords(slug) {
+  let newSlug = slug
+  newSlug = newSlug.replace(UPPERCASE_KEYWORDS_REGEX, match => {
```
shcheklein, line 33: can it be as simple as `return slug.replace ....`?

shcheklein (approved):

jorgeorpinel:
Last Q. There's also the opposite feature i.e. some words are kept lower case, like `and`. Where is that defined? Thanks

jorgeorpinel:
Ah thanks for the info. Will mention in https://github.com/iterative/gatsby-theme-iterative/issues/17

PR #3608, opened 2022-05-31
ref: ruamel.yaml admons

jorgeorpinel:
Follow-up to https://github.com/iterative/dvc.org/pull/3579

### DIFF
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -58,10 +58,6 @@ as the tree path to find those values. Supported types are: string, integer,
 float, and arrays (groups of params). Note that DVC does not ascribe any
 specific meaning to these values.
 
-> YAML 1.2 stores very large and very small numbers in scientific notation, but
-> the popular PyYAML library uses an older version of the format. To avoid
-> introducing subtle bugs, the ruamel.yaml library should be used instead.
-
 DVC saves parameter names and values to `dvc.lock` in order to track them over
 time. They will be compared to the latest params files to determine if the stage
 is outdated upon `dvc repro` (or `dvc status`).
--- a/content/docs/command-reference/params/index.md
+++ b/content/docs/command-reference/params/index.md
@@ -130,9 +130,12 @@ epochs = params['train']['epochs']
 layers = params['train']['layers']
 ```
 
-> Note that the popular PyYAML library does not support YAML 1.2. The
-> ruamel.yaml library should be used instead to avoid subtle differences in
-> number handling.
+<admon>
+
+We use [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) which supports YAML
+1.2 (unlike the more popular PyYAML).
+
+</admon>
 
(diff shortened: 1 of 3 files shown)


content/docs/command-reference/params/index.md
```diff
@@ -58,10 +58,6 @@ as the tree path to find those values. Supported types are: string, integer,
 float, and arrays (groups of params). Note that DVC does not ascribe any
 specific meaning to these values.
 
-> YAML 1.2 stores very large and very small numbers in scientific notation, but
-> the popular PyYAML library uses an older version of the format. To avoid
-> introducing subtle bugs, the ruamel.yaml library should be used instead.
```
jorgeorpinel, line 63: This note is quite interesting (thanks @bobertlo) but doesn't seem necessary. This is a [general explanation about params](http://localhost:8000/doc/command-reference/params#description) (which should be extracted to a user guide eventually) and DVC in general is language-agnostic, so this very Python-specific note isn't required IMO.

content/docs/command-reference/params/index.md
```diff
@@ -130,9 +126,12 @@ epochs = params['train']['epochs']
...
 
-> Note that the popular PyYAML library does not support YAML 1.2. The
-> ruamel.yaml library should be used instead to avoid subtle differences in
-> number handling.
+<admon>
+
+We use [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) which supports YAML
+1.2 (unlike the more popular PyYAML).
+
+</admon>
```
jorgeorpinel, line 134: The other notes I just edited a little and made into proper admonitions

image

content/docs/command-reference/params/index.md
dberenbaum: I see your point and will leave it up to you, but I think limiting this info to the examples defeats the purpose of the PR by @bobertlo. We aim to be language agnostic and this note is not really about DVC, but IMO we should err on the side of including warning about common issues that users face even if they are not strictly necessary.

bobertlo: For context, I posted this because the mentioned subtle bug breaks user code by not parsing the scientific notation generated by experiments. After reading the error messages, the top google result for 'dvc params scientific notation' lead me an open issue from a year ago discussing whether or not to fix the bug. So this blurb is specifically for SEO and user experience.

jorgeorpinel: I think in any case we should find the note a better place. I'll take a look...

Mind approving meanwhile @dberenbaum ? 🙂 thanks

dberenbaum (approved):

content/docs/command-reference/params/index.md
jorgeorpinel: Mentioned in https://github.com/iterative/dvc.org/issues/3635 for now.

jorgeorpinel: p.s. after https://github.com/iterative/dvc.org/pull/3459 this may not be relevant anyway...

PR #3621, opened 2022-06-04
nav: clean up Studio nav + misc. updates

jorgeorpinel:
In prep for large changes for MR feature

In review app: https://dvc-org-nav-studio-hv40outbutj.herokuapp.com/doc/studio

PR #3622, opened 2022-06-05
studio/guide: merge How it Works; flatten Views and Exps

jorgeorpinel:
> Depends on https://github.com/iterative/dvc.org/pull/3621

- Merge similar info. from Studio docs homes, How it Works page (gone), and Prepare your repos
  - [x] Merge all info about ~~views~~ projects into one guide
  - [ ] Merge all info about experiments into another guide

- [ ] Un-burry guides under UG -> ~~Views~~ Projects and Experiments

  > Extracted from https://github.com/iterative/dvc.org/pull/3621#pullrequestreview-995958843


tapadipti:
@jorgeorpinel ok, let's revisit this later.

shcheklein:
@tapadipti @jorgeorpinel any updates here?

tapadipti:
@shcheklein No, we have not discussed this recently.

@jorgeorpinel it would be good to restructure the docs a bit to "un-bury guides" as you have stated. But in the form in this PR, the pages are going to be quite long. Eg, I looked at the `content/docs/studio/user-guide/projects.md` page, and it's very long.

I think it'll be good to tackle this in steps - first, let's split the `projects` and `experiments` contents. Then, we can improve further.

Also, the following types of changes can be a different PR, coz they aren't about the structure:

image

Let me know what you think about when we can work this out. I'm currently working on the Live Metrics docs, and the feature is expected to go live in 1 week or so. So any restructuring would have to be done either within 1-2 days or would have to wait until the Live Metrics docs are released.

jorgeorpinel:
It'll be easier to start over at this point since there are so many conflicts per other changes to Studio docs.

PR #3625, opened 2022-06-06
Use link check from the CLI instead of an action

rogermparent:


rogermparent:
I didn't realize the large diff issue wasn't due to GitHub Actions, but `child_process.exec`'s buffer limitations. I can think of a few fixes to that, all involving a small upgrade to the link check package itself. I still think this is the way to go, but I'll shelve it for now until we add link-check to mlem.ai and get the VSCode extension released.

rogermparent:
Superceded by #3700

PR #3643, opened 2022-06-13
Resize images for blogpost previews to preserve aspect ratio

RCdeWit:
Thumbnails for blogpost header images are resized to a different aspect ratio. This results in some content being lost on the left and right sides, which can be problematic if there are logos or text in that area.

This change to the CSS resizes the images in a way that preserves the aspect ratio.

**Before:**

Screenshot 2022-06-13 at 10 26 00

**After:**

Screenshot 2022-06-13 at 10 26 18


RCdeWit:
As discussed here, by the way:
https://github.com/iterative/vscode-dvc/issues/1631#issuecomment-1153248185

julieg18 (approved):
Looks much better!

maxagin:
Is it possible to reduce the spacing between the blocks? If this is not too much to ask before the release :) 

Screen Shot 2022-06-13 at 2 46 20 PM
171778157-8dd5e24f-2106-444e-b0c3-51423001b4f0

PR #3647, opened 2022-06-13
Updated image for May community gems to one without the title

flippedcoder:
We had to update the image to use the one that doesn't have the text on it because those are used in the social media posts.


jendefig (approved):
Thank you! 🙏

PR #3651, opened 2022-06-13
Add Info about VS Code Extension

julieg18:
- [x] Adds VS Code to header and footer
- [x] Updates Banner
- [x] Updates `/features`

### DIFF
--- a/src/@dvcorg/gatsby-theme-iterative/components/LayoutFooter/index.tsx
+++ b/src/@dvcorg/gatsby-theme-iterative/components/LayoutFooter/index.tsx
@@ -126,6 +126,12 @@ const footerListsData: Array<IFooterListData> = [
         text: 'MLEM',
         icon: <MlemSVG className={styles.productIcon} />,
         target: '_blank'
+      },
+      {
+        href: 'https://marketplace.visualstudio.com/items?itemName=Iterative.dvc',
+        text: 'VS Code Extension',
+        icon: <LogoSVG className={styles.productIcon} />,
+        target: '_blank'
       }
     ]
   }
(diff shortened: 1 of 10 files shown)


jorgeorpinel:
Unrelated but why do we list DVC in other tools?

image

image

May be good to remove that now that the list is getting longer.

jorgeorpinel:
Should use consistent icon/logo I think (including the external link icon on hover).

image

image

jorgeorpinel:
Seems too prominent in /features? No strong opinion as that whole page is in need of redesign anyway...

image

jorgeorpinel (approved):
Some design questions but should work IMO. Thanks @julieg18

jorgeorpinel:
p.s. did we decide not to put it in the Download button after all?

yathomasi:
> Indirectly related, but why do we list DVC under "other tools"?

I think it would be better to have all the tools in the same place. Maybe we can rename it to "our tools". If we add more tools then we can reorganize them into multiple columns.

jorgeorpinel:
> Maybe we can rename it to "our tools"

Agree.

julieg18:
> Seems too prominent in /features? No strong opinion as that whole page is in need of https://github.com/iterative/dvc.org/issues/144#issuecomment-1004522646 anyway...

Agreed! We originally just wanted to add another feature to the list but it looked odd with uneven rows, which is why we chose to just put more focus on it as a top feature. We could discuss other options, but this is going to be merged in about an hour 😅 And as you already mentioned, we'll be refactoring this whole page in the future! 

> p.s. did we decide not to put it in the Download button after all?

Yes, maybe we'll add it in the future, but it's not needed for the release. 

> Should use consistent icon/logo I think (including the external link icon on hover).

Maybe! I don't see why the footer/nav would need to match besides the logo though... Footers don't typically have special animation/effects added to them... As for logo, we could move the links to have the VSC link to be right under DVC without a logo as well but it will need to be in a follow-up pr. 

> Indirectly related, but why do we list DVC under "other tools"?
> I think it would be better to have all the tools in the same place. Maybe we can rename it to "our tools". If we add more tools then we can reorganize them into multiple columns.

Good idea! I can change the naming in a follow-up pr :)

PR #3655, opened 2022-06-14
gdrive: update setup guide for LocalWebserver auth

dtrifiro:
dvc used Pydrive2's `CommandLineAuth` mechanism, which used the now-deprecated OAuth2 out-of-band authorization.
This was replaced with `LocalWebserver` authentication, which no longer requires manual setup of client id/secret.

See https://github.com/iterative/dvc-objects/pull/52 for more context

### DIFF
--- a/content/docs/user-guide/setup-google-drive-remote.md
+++ b/content/docs/user-guide/setup-google-drive-remote.md
@@ -111,9 +111,7 @@ We highly recommend this for heavy use and advanced needs because:
 
 DVC uses the [Google Drive API](https://developers.google.com/drive) to connect
 to your Google Drive. This requires a Google Cloud _project_ that allows Drive
-API connections, and its
-[OAuth](https://developers.google.com/identity/protocols/OAuth2) credentials
-(**client ID** and **client secret**).
+API connections.
 
 1. Sign into the [Google API Console](https://console.developers.google.com).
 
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -141,22 +139,6 @@ API connections, and its
 
    ![](/img/gdrive-create-credentials.png)
 
-6. The newly generated **client ID** and **client secret** should be shown to
```
shcheklein, line 144: @dtrifiro I'm not sure I understand why we want to remove this? Creating a custom APP is still better compared to the built -in one. Also, it's not related to the changes we are doing, or I'm missing something.

content/docs/user-guide/setup-google-drive-remote.md
```diff
@@ -141,22 +139,6 @@ API connections, and its
...
-GDrive remote), for example:
-
-```dvc
-$ dvc remote modify myremote gdrive_client_id 'client-id'
-$ dvc remote modify myremote gdrive_client_secret 'client-secret'
-```
-
 > Note that Google Drive API usage limits/quotas apply per _project_ client and
 > can be reviewed in the
 > [OAuth consent screen](https://console.developers.google.com/apis/credentials/consent).
```
jorgeorpinel, line 144: Is this still applicable?

jorgeorpinel:
Maybe we should deprecate this page in favor of blog posts... We keep having to maintain specific 3rd party docs.

jorgeorpinel:
p.s. that said it does have OK [statistics](https://plausible.io/dvc.org?period=30d&page=%2Fdoc%2Fuser-guide%2Fsetup-google-drive-remote) with over 2% of all site traffic and a low 5% bounce rate.

content/docs/user-guide/setup-google-drive-remote.md
dtrifiro: Looking at this again, this does not make any sense 🤦 Closing this PR as it's not required

dtrifiro:
Closing this as it is pointless: we still need to setup client id/secret for custom OAuth apps

PR #3663, opened 2022-06-14
vs-code: docs

jorgeorpinel:
> To work on after #3665


jorgeorpinel:
Discussed a few times with @shcheklein and this doesn't seem like a priority. We can retake in the future.

PR #3666, opened 2022-06-14
Add "vscode" icon to docs sidebar

julieg18:
See https://github.com/iterative/dvc.org/pull/3665#discussion_r897094431.


src/@dvcorg/gatsby-theme-iterative/components/Documentation/Layout/SidebarMenu/icons/index.ts
```diff
@@ -3,11 +3,13 @@ import defaultIcons, {
 } from '@dvcorg/gatsby-theme-iterative/src/components/Documentation/Layout/SidebarMenu/icons'
 import { ReactComponent as CMLIcon } from './cml-icon.svg'
 import { ReactComponent as StudioIcon } from './studio-icon.svg'
+import { ReactComponent as VsCodeIcon } from '../../../../../../../../static/img/vscode-icon.svg'
```
julieg18, line 6: Would look better in a grayed-out version, but we can only use the officially logo legally :/

jorgeorpinel (dismissed):
I don't think we can use their icon here at all actually.

image

> https://code.visualstudio.com/brand

Can we just get a grayed-out version of the DVC logo in there for now, instead? So the entry stands out.

shcheklein:
@julieg18 @jorgeorpinel let's keep VS Code but grey it out, similar to Studio

julieg18:
> @julieg18 @jorgeorpinel let's keep VS Code but grey it out, similar to Studio

I thought we couldn't modify the logo? https://code.visualstudio.com/brand

shcheklein:
@julieg18 yes, but we can't keep it colorful, I think it's fine if use it as silhouette/shape icon, especially in case like this - we are not branding any product, etc.

At least you can get it from the https://simpleicons.org/?q=vs%20code and use it - they don't restrict it.

Or the way we do GH icons for navigation in different colors a bit for better contrast.

Hard for me to tell exactly, but it feels fine tbh. Clearly, we can't keep it "blue", so even for them it would be better to have it in full grey.

rogermparent (approved):
Ignoring the question of being allowed to use the logo, the code looks good!

shcheklein (approved):

Issue #3633, opened 2022-06-10
Server Error for `Hydration failed ...` and `Text content does not match ...`


yathomasi:
We are starting to get multiple errors on production due to the recent upgrade to React 18. 

**Severity:**
- mild 
- the website doesn't seem to break but it's complaining with errors on the console
- works fine with javascript disabled

**Common Errors:**
- There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering. 
- Hydration failed because the initial UI does not match what was rendered on the server.
- Text content does not match server-rendered HTML.

**Cause:**
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#other-breaking-changes
>Stricter hydration errors: Hydration mismatches due to missing or extra text content are now treated like errors instead of warnings. React will no longer attempt to “patch up” individual nodes by inserting or deleting a node on the client in an attempt to match the server markup, and will revert to client rendering up to the closest <Suspense> boundary in the tree. This ensures the hydrated tree is consistent and avoids potential privacy and security holes that can be caused by hydration mismatches.

**Gatsby Discussion:**
- https://github.com/gatsbyjs/gatsby/discussions/31943
- https://github.com/vercel/next.js/discussions/35773( similar discussion ongoing on Next.js)

**Possible Approach:**
- Find what content is causing the error( DEV_SSR seemed to be a solution but it's WIP: https://github.com/gatsbyjs/gatsby/pull/35685)

**Resources:**
- https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#other-breaking-changes
- https://lihautan.com/hydrating-text-content

**Sites:**
- [x] dvc.org
- [x] cml.dev
- [x] mlem.ai

julieg18:
dvc.org pages that are showing these console errors
* any of the doc pages
* home page
* the community page

rogermparent:
I've managed to pin this down to our Download button's "change depending on OS" functionality! That would explain why I was only able to replicate the error on the home and docs pages.

@julieg18 I couldn't replicate on the community page at all, maybe there was another error that looked like this one? Also, I was able to get a local replication via `yarn build && yarn serve`, as `yarn develop` doesn't have any hydration and behaves like a traditional React app.

I'm working on a fix.

LekoArts:
You can try out `DEV_SSR` with `gatsby@next` now again.

yathomasi:
[...]

PR #3675, opened 2022-06-18
Only update User OS after hydration

rogermparent:
rough fix for #3633 

Main problem is that it has to update after first render which is a problem shared with live GitHub stars

We could either remove the description or make its update transition friendlier, anything with less drawbacks will take an order of magnitude more effort to implement.

### DIFF
--- a/src/components/DownloadButton/index.tsx
+++ b/src/components/DownloadButton/index.tsx
@@ -4,22 +4,38 @@ import cn from 'classnames'
 import TwoRowsButton from '../TwoRowsButton'
 import Link from '@dvcorg/gatsby-theme-iterative/src/components/Link'
 
-import isClient from '@dvcorg/gatsby-theme-iterative/src/utils/front/isClient'
 import { logEvent } from '@dvcorg/gatsby-theme-iterative/src/utils/front/plausible'
 
 import * as styles from './styles.module.css'
+import { OS, useUserOS } from '../../utils/front/useUserOS'
 
 const VERSION = `2.11.0`
 
-enum OS {
-  UNKNOWN = 'unknown',
-  OSX = 'osx',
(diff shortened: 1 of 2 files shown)


yathomasi:
@rogermparent Great work pinning it down 👏 

> Main problem is that it has to update after first render which is a problem shared with live GitHub stars

If I understand correctly, updating the OS after the first render is okay because we can only get the client OS once the script is loaded to the client browser(I believe it's also the reason for the hydration issue). Also, with our existing implementation, it should also be doing the same.

> make its update transition friendlier

That would be great but I think right now silencing the error takes priority.

shcheklein:
@rogermparent it's still a draft?

rogermparent:
Sorry, I thought I had changed it to ready but must have been looking at the wrong tab or something

shcheklein:
Let's ship it, but in general it takes surprisingly long to update the screen (drop down and GitHub). 

What had changed to cause this problem btw?

rogermparent:
We've always had this problem, it's just that React 18 made hydration mismatches throw an error instead of log a warning. The whole process of hydration means that a static HTML file is sent first, renders fast, and then React kick in to hydrate where it expects the results to match up with that initial HTML, which makes UI bits that are different for every user more difficult than other website implementations.

shcheklein (approved):

shcheklein:
thanks Roger, It's almost immediate now on prod, do you expect it taking that much longer after the deploy?

rogermparent:
> It's almost immediate now on prod, do you expect it taking that much longer after the deploy?

I doubt it'd be much longer than it currently is on prod, as the GH stars currently are populated on first render and they're pretty quick on prod with the download button not much further ahead.

PR #3677, opened 2022-06-20
Use Gatsby Script tag to add Reddit script

rogermparent:
Gatsby 4.15 introduced a built-in `Script` component, which acts much like the HTML `script` element but does optimizations such as making scripts loaded with the Gatsby component load after Hydration, improving initial page load time.

It turns out this is the only `script` tag in the repo which isn't in the theme. There's another in `gatsby-ssr` but that one is meant to not be deferred. Probably won't gain us a ton of points, but figured I may as well push it.


yathomasi (approved):
Looks good to me.

PR #3680, opened 2022-06-21
Move Script tag outside of helmet

rogermparent:
After finding something that implements the reddit embed API, I realized #3677 actually broke it. Seems like the `Script` tag is incompatible with `Helmet`, but that's not an issue since the `Script` tag already moves its scripts to the bottom of the page regardless of where they are.

See demo of broken reddit embed [on this page](http://dvc.org/blog/july-19-dvc-heartbeat#how-do-you-manage-your-machine-learning-experiments-discussion-on-reddit-is-full-of-insights).

broken:
image

fixed:
image


yathomasi (approved):
Looks good.

PR #3636, opened 2022-06-11
.github: new issue template config

jorgeorpinel:
Trying https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

Closes #2503

- Could try this for the PR template too: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

PR #3685, opened 2022-06-23
fix issue template

jorgeorpinel:
> Fix https://github.com/iterative/dvc.org/pull/3636

Per https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax


jorgeorpinel:
Hmmm this is not what I wanted to push...

jorgeorpinel:
This was a test for pull request template 🤦🏼

jorgeorpinel:
See https://github.com/iterative/dvc.org/pull/3689

PR #3690, opened 2022-06-23
blog: June Community Gems

flippedcoder:
Tentative publish date: 06/29/22

Things left to update:
- [x] image (make sure to use https://tinypng.com/)
- [x] comments link
- [x] July office hours description/link (no office hours next month)
- [ ] technical style (code blocks formatted correctly with dvc, python, yaml, etc)


content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,200 @@
...
+commentsUrl: https://discuss.dvc.org/t/may-22-community-gems/1184
+tags:
+  - DVC Remotes
+  - Pipelines
+  - CML
+  - GCP
+  - Community
+---
+
+## [Is there something like `git add-commit -m "some message"` for DVC?](https://discord.com/channels/485586884165107732/563406153334128681/981498675689828362)
```
RCdeWit, line 23: Had to look up the original question; it sounds like Ramnath wants to do all of this with one command. I'm not sure the explanation now answers that question. Maybe go for "No, but you could create an alias"?

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,200 @@
...
+
+## [How does `dvc get-url` interact with the cache compared to `dvc import-url`?](https://discord.com/channels/485586884165107732/563406153334128681/981862313076346920)
+
+This is an awesome question from @Gema Parreno!
+
+When you run `dvc get-url`, it downloads the file/directory to your local file
+system. It's _not_ tracking the downloaded data with a `.dvc` file. It's just
+pulling that data from some source to your file system.
+
+On the other hand, when you run `dvc import-url`, the local `cache` folder
```
RCdeWit, line 60: Clear explanation, might just copy this for the course. I've seen quite a bit of confusion around this distinction; maybe provide an example of when you'd use either option?

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,200 @@
...
+```dvctable
+ ─────────────────────────────────────────────────────────────────
+  neutral:**Experiment**   metric:**avg_prec**   metric:**roc_auc**   param:**train.n_est**   param:**train.min_split**
+ ─────────────────────────────────────────────────────────────────
+  workspace           -         -   300           75
+  mlem-serve    0.76681   0.38867   300           75
+ ─────────────────────────────────────────────────────────────────
+```
+
+If you're interested in more advanced visualizations, you should try out
```
RCdeWit, line 114: This seems like a prime candidate to plug the VSCode extension as well! Do the same but with a mouse click. 😄

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+In DVC, you can update the params and run `dvc exp run --set-param <param>`.
+Here's an example of what that might look like:
+
+```dvc
+$ dvc exp run --set-param params.py:TrainTestSplit.SPLIT_METHOD="proportional"
+```
+
+_Note:_ It may not be able to update Python parameters correctly.
```
RCdeWit, line 180: Depending on what? Are we unsure whether this works at all, or is DVC flaky in this instance? 😅
Maybe we could provide a nudge towards using a `params.yaml`? Or is `params.py` a common approach?

daavoo: > Depending on what?

From https://dvc.org/doc/command-reference/params#examples-python-parameters-file :

> ⚠️ Note that complex expressions (unsupported by [ast.literal_eval](https://docs.python.org/3/library/ast.html#ast.literal_eval)) won't be parsed as DVC parameters.

>  Maybe we could provide a nudge towards using a params.yaml? Or is params.py a common approach?

We already nudge towards `params.yaml` by making it the default. 
Outside of that, I think we should let users choose any of the supported formats and explicitly state limitations.

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,215 @@
...
+commentsUrl: https://discuss.dvc.org/t/may-22-community-gems/1184
+tags:
+  - DVC Remotes
+  - Pipelines
+  - CML
+  - GCP
+  - Community
+---
+
+## [Is there something like `git add-commit -m "some message"` for DVC?](https://discord.com/channels/485586884165107732/563406153334128681/981498675689828362)
```
dberenbaum, line 23: I'm not sure the question is that clear. Later on, they state: "Sometimes if the updated files are "scattered", and you want to take snapshot of all modified files (which were earlier added/ which dvc started tracking), it would be convenient to have a shorthand for that."

If that's the issue, `dvc commit` would already handle this. As you already wrote below:

> If you already have data tracked by DVC, the
> `dvc commit` command adds all the changes to those files or directories without
> having to name each target.

I don't think the proposed alias below is needed for DVC (it doesn't work quite like Git in this sense).

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,215 @@
...
+
+Thanks for the question @Ramnath T!
+
+While there isn't a single command to handle this, you can get similar behavior
+with DVC with an alias, like the following:
+
+```dvc
+$ alias dvc-add-commit='dvc add . & dvc commit'
```
dberenbaum, line 31: I don't think this alias works (`dvc add .` will fail) and it wouldn't make much sense anyway. You wouldn't want to `dvc add` your whole workspace, and `dvc commit` would be redundant if you did. Just `dvc commit` should really be enough here.

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,215 @@
...
+
+When you run `dvc get-url`, it downloads the file/directory to your local file
+system. It's _not_ tracking the downloaded data with a `.dvc` file. It's just
+pulling that data from some source to your file system. If you want to download
+a file or directory without needing a DVC project, you can use the `dvc get-url`
+command.
+
+On the other hand, when you run `dvc import-url`, the local `cache` folder
+inside of `.dvc` will be updated. This is similar to running `dvc get-url` and
+`dvc add` together.
```
dberenbaum, line 73: Except that `dvc import-url` also saves a link to the original file/directory location so that if it changes, you can download the updated file/directory.

content/blog/2022-06-29-june-community-gems.md
```diff
@@ -0,0 +1,215 @@
...
+representation.
+
+In summary, the image file is only stored in the shared cache once unless it's
+modified in one of the directories.
+
+## [Is it possible to limit which columns show for experiments in the metrics table?](https://discord.com/channels/485586884165107732/563406153334128681/985448515402616842)
+
+Nice question from @DylanTF!
+
+You can use `dvc exp --drop` (or `--keep`) to decide what to hide (or show). For
```
dberenbaum, line 96:
```suggestion
You can use `dvc exp show --drop` (or `--keep`) to decide what to hide (or show). For
```

dberenbaum, line 96: Consider suggesting `--only-changed` here since it will automatically drop most columns that aren't useful.

dberenbaum (commented):
@flippedcoder I had some comments on the first question. Other than that, they are minor suggestions that you can decide whether to include or not. Thanks!

content/blog/2022-06-29-june-community-gems.md
flippedcoder: It does run in a project I have set up, but I see what you mean about it not making sense to add the whole workspace. I just took out the alias tidbit and went back to what I had originally.

content/blog/2022-06-29-june-community-gems.md
flippedcoder: I updated the question to be a little more general. Let me know if it makes more sense.

flippedcoder:
@dberenbaum Thanks for the feedback! I made a few updates that'll hopefully clear up that first question. If it looks good to you though, could you go ahead and approve?

PR #3692, opened 2022-06-27
install: add note regarding arm macos support

pmrowla:
Document that pip (and homebrew which uses pip) is the only supported installation method for ARM macs. Currently the conda packages and PKG binary installer are only built for x86.

### DIFF
--- a/content/docs/install/macos.md
+++ b/content/docs/install/macos.md
@@ -2,8 +2,8 @@
 
 <admon type="tip">
 
-To use DVC [as a Python library](/doc/api-reference), please
-install [with `pip`](#install-with-pip) or [with `conda`](#install-with-conda).
+To use DVC [as a Python library](/doc/api-reference), please install
+[with `pip`](#install-with-pip) or [with `conda`](#install-with-conda).
 
 </admon>
 


content/docs/install/macos.md
```diff
@@ -3,6 +3,11 @@
 > To use DVC [as a Python library](/doc/api-reference), please
 > [install with pip](#install-with-pip) or [with conda](#install-with-conda).
```
jorgeorpinel, line 4:
```suggestion
<admon type="tip">

To use DVC [as a Python library](/doc/api-reference), please
[install with pip](#install-with-pip) or [with conda](#install-with-conda).

</admon>
```

content/docs/install/macos.md
```diff
@@ -3,6 +3,11 @@
...
 
+> ⚠️ To use DVC on ARM (Apple Silicon) macOS, you must
+> [install with homebrew](#install-with-homebrew) or
+> [with pip](#install-with-pip). The PKG binary and conda installation methods
+> can only be used on x86 (Intel) macOS and are not supported for ARM macOS.
```
jorgeorpinel, line 9:
```suggestion
<admon type="warn">

To use DVC on ARM (Apple Silicon) macOS, you must install
[with homebrew](#install-with-homebrew) or [with `pip`](#install-with-pip).
The PKG binary and `conda` installation methods can only be used on x86 (Intel)
macOS.

</admon>
```

jorgeorpinel (approved):
Taking the opportunity to use proper admonitions (comitting)

content/docs/install/macos.md
```diff
@@ -1,7 +1,20 @@
 # Installation on macOS
 
-> To use DVC [as a Python library](/doc/api-reference), please
-> [install with pip](#install-with-pip) or [with conda](#install-with-conda).
+<admon type="tip">
+
+To use DVC [as a Python library](/doc/api-reference), please
+[install with pip](#install-with-pip) or [with conda](#install-with-conda).
+
+</admon>
```
jorgeorpinel, line 8: - [ ] We should apply this to all installation pages...

content/docs/install/macos.md
```diff
@@ -1,7 +1,20 @@
 # Installation on macOS
 
-> To use DVC [as a Python library](/doc/api-reference), please
-> [install with pip](#install-with-pip) or [with conda](#install-with-conda).
+<admon type="tip">
+
+To use DVC [as a Python library](/doc/api-reference), please
+[install with pip](#install-with-pip) or [with conda](#install-with-conda).
```
jorgeorpinel, line 6:
```suggestion
install [with `pip`](#install-with-pip) or [with `conda`](#install-with-conda).
```

dberenbaum:
@pmrowla Is this still the case? I was able to install the pkg on my m1.

pmrowla:
> @pmrowla Is this still the case? I was able to install the pkg on my m1.

The pkg installer is only built for x86, you can install it on an m1 but it gets installed and run through rosetta (the x86 emulation layer).

The conda package for dvc is noarch/universal now, but installing on conda requires that all of DVC's dependencies also have either noarch or arm builds (see https://github.com/conda-forge/dvc-feedstock/issues/250)

PR #3700, opened 2022-06-28
Update link-check to 0.12 using CLI and comment reporting

rogermparent:
This PR upgrades `link-check` to 0.12, which should squash the "diffs are too big" problem/bug.

In order to squash an issue where `node:buffer` couldn't be imported in the GH Action, this PR now just uses link check through the CLI and its stdout to make a special review comment which gets edited after each check.

screenshot of a failing version of this PR's content

https://github.com/iterative/dvc.org/pull/3700#issuecomment-1184964728

screenshot of checks list

We no longer have three checks for link check!


.github/workflows/link-check-deploy.yml
```diff
@@ -12,9 +12,10 @@ jobs:
       github.event.deployment.ref != 'main' &&
       github.event.deployment_status.state == 'success'
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
rogermparent, line 15: I'll follow up with a change to update `actions/checkout` in all our other actions as well, from what I've seen it's a no-downsides performance upgrade.

shcheklein:
@rogermparent ping

rogermparent:
Sorry, this has fallen by the wayside a bit, I'll bump it up in my priorities this week!

.github/workflows/link-check-deploy.yml
```diff
@@ -2,47 +2,58 @@ name: Check new links against deployment
...
 # Checks refuses to show the check on deployment_status
 on:
-  - deployment
   - deployment_status
 jobs:
   run:
-    name: Initialize
+    name: Run
     runs-on: ubuntu-latest
+    continue-on-error: true
```
rogermparent, line 10: Lets us continue even if an existing report comment isn't found

.github/workflows/link-check-deploy.yml
```diff
@@ -2,47 +2,58 @@ name: Check new links against deployment
...
-        uses: iterative/link-check.action@v0.11
+        run: |
+          set +e
+          body="<h1 id=\"link-check\">Link Check Report</h1>
+
+          $(
+            npx repo-link-check \
+            -d -c config/link-check/config.yml \
+            -r ${{ github.event.deployment.payload.web_url }}
+          )"
```
rogermparent, line 28: `set +e` prevents the script from exiting when `repo-link-check` exits with code 2. `+e` is a normal default for interactive terminals, but `-e` is used in some CI like GitHub Actions.

.github/workflows/link-check-deploy.yml
```diff
@@ -2,47 +2,58 @@ name: Check new links against deployment
...
+          body="${body//'%'/'%25'}"
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo "::set-output name=report::$body"
+          exit 0
```
rogermparent, line 33: From [`create-or-update-comment` docs](https://github.com/marketplace/actions/create-or-update-comment#setting-the-comment-body-from-a-file), escapes the command result for GH Action output

.github/workflows/link-check-deploy.yml
```diff
@@ -2,47 +2,58 @@ name: Check new links against deployment
...
         with:
-          diff: main
-          configFile: config/link-check/config.yml
-          rootURL: '${{ github.event.deployment.payload.web_url }}'
-          output: checksAction
-      - uses: LouisBrunner/checks-action@v1.0.0
-        if: ${{ success() }}
+          script: |
+            const findPR = require('./.github/workflows/deployment-report.js')
+            return await findPR({ github, context, core })
```
rogermparent, line 43: This action is really handy for rapidly prototyping weird stuff with the GitHub API. You can import modules so most of the code can be properly linted instead of being stuck in a YAML block.

.github/workflows/link-check-deploy.yml
```diff
@@ -2,47 +2,58 @@ name: Check new links against deployment
...
-          token: ${{ secrets.GITHUB_TOKEN }}
-          check_id: ${{ steps.build_check.outputs.check_id }}
-          status: completed
-          conclusion: ${{ steps.check.outputs.conclusion }}
-          output: ${{ steps.check.outputs.output }}
-      - uses: LouisBrunner/checks-action@v1.0.0
-        if: ${{ failure() }}
+          issue-number: ${{ steps.pr-number.outputs.result }}
+          comment-author: 'github-actions[bot]'
+          body-includes: "<h1 id=\"link-check\">Link Check Report</h1>"
```
rogermparent, line 51: Adds a way to navigate to the report comment with [an anchor for #user-content-link-check](#user-content-link-check), and also helps further prevent false positives on which comment to edit.

.github/workflows/deployment-report.js
```diff
@@ -0,0 +1,12 @@
...
+  const owner = context.payload.repository.owner.login
+  const repo = context.payload.repository.name
+  const { data: pulls } = await github.rest.pulls.list({
+    state: 'open',
+    owner,
+    repo
+  })
+  const actionPull = pulls.find(pull => pull.head.sha === sha)
+  return actionPull.number
+}
```
rogermparent, line 12: The `deployment_status` event we use from Heroku doesn't actually include info about the pull request, so we have to duck into the GitHub API to get that PR via the current commit's `sha`.

.github/workflows/link-check-deploy.yml
yathomasi: Could this also skip any other error occurrence and skip without reporting the error?

yathomasi: Could this also skip any other error occurrence without reporting the error?

yathomasi (approved):
Looking alright to me. Great work @rogermparent. 
I can see the report generated and updated fine.

shcheklein (commented):
Roger, have you tried to use CML for this? Could you share your experience?

rogermparent:
I took a look, but hit the ["Known Issues" portion](https://cml.dev/doc/ref/send-comment#faqs-and-known-issues) of the `send-comment` docs that states the following:
> commit_id has been locked.
> 
> This [error](https://github.community/t/comment-api-does-not-describe-commit-id-has-been-locked/159853/2) is caused by using the default GitHub token with [cml send-comment --update](https://cml.dev/doc/ref/send-comment#--update). Use a [personal access token (PAT)](https://cml.dev/doc/self-hosted-runners?tab=GitHub#personal-access-token) instead.

.github/workflows/link-check-deploy.yml
rogermparent: It seems like it does, I'll need to improve that next. If an error occurs, the script will just write the "Link Check Report" `h1` with no content. I'm pretty sure that can be fixed by getting stderr output into the content body in the bash script part of the action.

PR #3716, opened 2022-07-01
Introduce vale linter

daavoo:
Add a new workflow for running https://vale.sh/ on P.R.

Added a bunch of rules ("styles") but I have no idea which ones make sense for our docs. Please review them

It can be tested locally by installing https://vale.sh/ and running from the root:

```
vale content/docs/start
```

E: Note that there's a [VS Code extension](https://marketplace.visualstudio.com/items?itemName=errata-ai.vale-server) too. - @jorgeorpinel 

Example of error and warning on markdown file:

https://github.com/iterative/dvc.org/pull/3716/files#diff-b1dd216e145355375d79e34394c0c99b634227f74ee3d1534406363c4c042d96


.github/workflows/vale.yaml
```diff
@@ -0,0 +1,16 @@
...
+  vale:
+    name: runner / vale
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: errata-ai/vale-action@reviewdog
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        with:
+          files: "content/docs"
```
jorgeorpinel, line 14: Is it possible to run it only on the files that are changed? Or even just on the changes? Wait I guess that's what it's doing now right?

.vale.ini
```diff
@@ -0,0 +1,7 @@
+StylesPath = .github/styles
+
+MinAlertLevel = suggestion
+Vocab = Base
+
+[*.md]
+BasedOnStyles = Vale, alex, proselint, Readability, write-good, dvc
```
jorgeorpinel, line 7:
```suggestion
BasedOnStyles = Vale, alex, proselint, Readability, write-good, dvc
```

.vale.ini
```diff
@@ -0,0 +1,7 @@
+StylesPath = .github/styles
```
jorgeorpinel, line 1: I'd just use `styles/`, like in the [vale-boilerplate](https://github.com/errata-ai/vale-boilerplate).

.vale.ini
jorgeorpinel, line 7: Most of these are available [packages](https://vale.sh/docs/topics/packages/), I think 🙂 that should mean there's no need to copy them into `styles/` in this repo, I think.

.vale.ini
jorgeorpinel: p.s. I guess they also implicitly recommend [using the CI folders](https://vale.sh/docs/topics/config/#stylespath)...

jorgeorpinel:
Very promising @daavoo ! It will take some time to go through the existing packages and rules to create a good base config, and then possibly write our own rules as well of course. Then check which repos could use this, talk to the product least, etc... It will take some time but I'll follow up on this.

Keeping open as a draft to play around with for now.

.vale.ini
daavoo: > That should mean there's no need to copy them into styles/ in this repo (need to check).

 I didn't expect the packages to work for us out of the box and there is no way to modify the packages (i.e. add/edit/remove rules) unless we copy the contents.
 
 I think the best would be to create a single `dvc` package with rules cherry picked (copy-pasted / edited) from existing packages + our custom rules.

.github/styles/dvc/admon.yml
```diff
@@ -0,0 +1,10 @@
+---
+extends: substitution
+message: Consider using '%s' instead of '%s'.
+ignorecase: true
+level: warning
+action:
+  name: replace
+swap:
+  'Note that' : an <admon> block
```
daavoo, line 9: @jorgeorpinel Here is an example custom rule that I added. Checks for `Not that` and suggest using `<admon>` block

jendefig: This is really exciting @daavoo!  It would be AMAZING if we could automate the grammar and style issues and free up reviewers time in this regard.  This would be a major leveraged win for all involved!

.github/styles/alex/Suicide.yml
```diff
@@ -0,0 +1,24 @@
...
+level: warning
+action:
+  name: replace
+swap:
+  commit suicide: die by suicide
+  committed suicide: died by suicide
+  complete suicide: die by suicide
+  completed suicide: died by suicide
+  epidemic of suicides: rise in suicides
+  failed attempt: suicide attempt|attempted suicide
```
casperdcl, line 14: "My failed attempt at installing GPU drivers -> My suicide attempt at ..."

wtf

.github/styles/alex/Condescending.yml
```diff
@@ -0,0 +1,16 @@
+---
+extends: existence
+message: Using '%s' may come across as condescending.
+link: https://css-tricks.com/words-avoid-educational-writing/
+level: error
```
casperdcl, line 5: should be `warning` not error. This entire file in laughable. The vast majority of technical content use these terms in a non-condescending way, e.g. "It is not obvious how to easily create a new ML project from scratch. We attempt to make this more simple here" throws 3 errors really?

.github/styles/alex/Race.yml
```diff
@@ -0,0 +1,87 @@
...
+  red indians: Native American People
+  redskin: Native American
+  redskins: Native American People
+  sand niggers: Arabs|Middle Eastern People
+  savage: simple|indigenous|hunter-gatherer
+  shlomo: Jewish person
+  shyster: Jewish person
+  sophisticated culture: complex culture
+  sophisticated technology: complex technology
+  spade: a Black person
```
casperdcl, line 73: :wat: Is the implication here you're meant to replace gardening tools with slave labour? fml these lists are sooooo wrong.

.github/styles/alex/Race.yml
```diff
@@ -0,0 +1,87 @@
...
+  pocahontas: Native American
+  pow wow: conference|gathering|meeting
+  powwow: conference|gathering|meeting
+  primitive: simple|indigenous|hunter-gatherer
+  red indian: Native American
+  red indians: Native American People
+  redskin: Native American
+  redskins: Native American People
+  sand niggers: Arabs|Middle Eastern People
+  savage: simple|indigenous|hunter-gatherer
```
casperdcl, line 68: what if "savage" not used to describe hunter-gatherers (overwhelmingly more common)? As in "my savage review of `.github/styles`?"

casperdcl (changes requested):
- really very against hosting `.github/styles` here.
  + Instead there should be a flake8-like convention - config file(s) should state categories to include/exclude + overrides... rather than the entire upstream lists.
  + How do we pull updates from upsteam lists?
- Also these lists are mostly super controversial. Numerous examples too upsetting to list (some exampled given below). The linter's highly misguided suggestions will likely be super triggering. Can we have a minimal no-brainer subset?

daavoo:
> * nstead there should be a flake8-like convention - config file(s) should state categories to include/exclude + overrides... rather than the entire upstream lists.
> 
> * How do we pull updates from upsteam lists?

This is a limitation of the tool itself -> https://github.com/errata-ai/vale/issues 
If there are better open source alternatives, I don't really care which tool to use.

See https://github.com/iterative/dvc.org/pull/3716#discussion_r918830373 for the reasoning of the setup, given the tool limitations.

>     Also these lists are mostly super controversial. Numerous examples too upsetting to list (some examples given below). The linter's highly misguided suggestions will likely be super triggering. 
> Can we have a minimal no-brainer subset instead?

Note that these are pre-defined packages I copy-pasted so we can remove what we don't need but reuse the rest (as there is no better way of doing it per the limitations mentioned above). 

We can absolutely have a minimal subset. I just opened this to showcase the setup and push the discussion. 
I expect the @iterative/docs team to refine and prune the rules to match our criteria and use case. Don't hesitate to edit or remove rules directly on the P.R.

.github/styles/alex/Race.yml
jendefig: 😆 Ok.  *Definite* issues here.  But can't we just have it run Grammarly?  Hard stop.  

That's all we need in my opinion.  That's what is the big time suck to all involved. 

Per what Jorge put above.  We may have to develop our own styles (which will be a project and take time).

daavoo: > what if "savage" is not used to describe hunter-gatherers (overwhelmingly more common)? As in "my savage review of .github/styles?"

@casperdcl All your comments make sense, those are rules built into the default packages provided. The reason for having `.github/styles` is to be able to modify/remove rules that don't apply, like your comments in this P.R, by just removing the `yaml` file.

daavoo: >  But can't we just have it run Grammarly?

@jendefig The short answer is no / not easy to run on the GitHub P.R.

> That's all we need in my opinion. That's what is the big time suck to all involved.

The best I can suggest is trying to edit the blog posts in VSCode and use the [Grammarly](https://marketplace.visualstudio.com/items?itemName=znck.grammarly) and [GitHub](https://code.visualstudio.com/docs/editor/github) extensions. 

This requires some investment in getting familiar with the editor and used to the new workflow. 

However, after setup you are able to view/respond to GitHub comments and get Grammarly suggestions in the same place you edit/write the blog:

Captura de Pantalla 2022-09-20 a las 10 17 07

SoyGema: Thanks for the recommendation @daavoo !

jendefig: This is great!  Thanks for pointing this out @daavoo!!!! 🙏 I have installed it in VS Code. It's asking me to log in, and I have, with the same Gmail as GitHub, but it's not giving the automated fix for the issue, but the suggestion at least so that's progress!

daavoo: @jendefig I don't know how to sync accounts, I am using the anonymized (no login) version. 

I can click on `Quick Fix` at the bottom left (or press the shortcut Comand + `.` if the cursor is on top of the error) to see and apply suggestions. 

It is still a worse UX than the web, but it's something. Maybe syncing the account enables real auto-fixing?

shcheklein:
Closing as stale

PR #3719, opened 2022-07-01
ref: include import-url scenario in dvc update

dberenbaum:
Parts of the doc look like they predate `import-url`, so I added some text to clarify that it also updates external URLs from `import-url`.

### DIFF
--- a/content/docs/command-reference/update.md
+++ b/content/docs/command-reference/update.md
@@ -1,8 +1,8 @@
 # update
 
 Update files or directories imported from external <abbr>DVC repositories</abbr>
-or [URLs](/doc/command-reference/import-url#description),
-and the corresponding import `.dvc` files.
+or [URLs](/doc/command-reference/import-url#description), and the corresponding
+import `.dvc` files.
 
 ## Synopsis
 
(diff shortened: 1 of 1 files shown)


content/docs/command-reference/update.md
```diff
@@ -18,7 +18,7 @@ positional arguments:
 
 After creating import stages (`.dvc` files) with `dvc import` or
 `dvc import-url`, the data source can change. Use `dvc update` to bring these
-imported file or directory up to date.
+imported files or directories up to date.
```
dberenbaum, line 21: Unrelated typo

shcheklein (approved):

content/docs/command-reference/update.md
```diff
@@ -1,7 +1,7 @@
 # update
 
-Update files or directories imported from external <abbr>DVC
-repositories</abbr>, and the corresponding import `.dvc` files.
+Update files or directories imported from external <abbr>DVC repositories</abbr>
+or URLs, and the corresponding import `.dvc` files.
```
daavoo, line 4: Maybe we should have a `<abbr>` also for URLs or link to the type of URL supported in this case (https://dvc.org/doc/command-reference/import-url#description)

daavoo (approved):

PR #3726, opened 2022-07-04
live: Updates for log standardization.

daavoo:
Per https://github.com/iterative/dvclive/releases/tag/0.9.0

Closes https://github.com/iterative/dvclive/issues/263

### DIFF
--- a/content/docs/dvclive/ml-frameworks/catalyst.md
+++ b/content/docs/dvclive/ml-frameworks/catalyst.md
@@ -39,7 +39,7 @@ Where:
 
 - `{Live.dir}` is the
   [`dir` attribute of `Live`](/doc/dvclive/api-reference/live#attributes).
-- `{split}` can be either `train` or `eval`.
+- `{split}` can be either `train` or `valid`.
 - `{metric}` is the name provided by the framework.
 
 <admon type="tip">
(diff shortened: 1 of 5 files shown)


dberenbaum (approved):

content/docs/dvclive/ml-frameworks/catalyst.md
```diff
@@ -28,8 +28,19 @@ runner.train(
...
+```py
+{Live.dir}/scalars/{split}/{metric}.tsv
+```
+
+Where:
+
+- `{Live.dir}` is the
+  [`dir` attribute of `Live`](/doc/dvclive/api-reference/live#attributes).
+- `{split}` can be either `train` or `eval`.
+- `{metric}` is the name provided by the framework.
```
dberenbaum, line 43: Where does this catalyst standardization come from? I don't see it in https://github.com/iterative/dvclive/pull/259/files.

daavoo: Its the default format used by the framework

daavoo: 😓 Catalyst uses `valid` instead of `eval` so I forgot to include it in the standardization

Issue #3718, opened 2022-07-01
Reference VS Code (and Studio?) for DVC plots


dberenbaum:
### Report

Some places where it would be helpful to mention VS Code (and maybe Studio):

https://dvc.org/doc/command-reference/plots

This might seem odd in a CLI ref, but right now it states:

> DVC generates plots as static HTML webpages that can be open with a web browser. They can also be saved as SVG or PNG image files from the browser.

Since VS Code provides a better experience, why not mention it?

https://dvc.org/doc/start/experiments/visualization

@jorgeorpinel @shcheklein Thoughts?

jorgeorpinel:
Agree to both (VS Code and Studio).

Probably in https://dvc.org/doc/start/experiments/visualization first and foremost.

> might seem odd in a CLI ref

It's OK to add to the CLI ref too, since most of the plots info is there still (will get extracted [eventually](https://github.com/iterative/dvc.org/issues/2572))

jorgeorpinel:
Or maybe in the parent (https://dvc.org/doc/start/experiments) even.

jorgeorpinel:
Started #3735 for the VS Code links.

The pending question is: do we add Studio links in all (or most) places where VS Code is linked from? Or just the plots-related ones for now (i.e. in the same PR) ?

jorgeorpinel:
Wait this looks mostly like a dupe of https://github.com/iterative/dvc.org/issues/3717.

Issue #3717, opened 2022-07-01
Add references to VS Code for DVC plots


dberenbaum:
### Report

Some places where it would help:

https://dvc.org/doc/command-reference/plots

This might be odd in a CLI ref, but for example, the current docs say:
 
> DVC generates plots as static HTML webpages that can be open with a web browser. They can also be saved as SVG or PNG image files from the browser.

Seems like it would be useful to at least mention that they can be visualized in VS Code, since it's a better experience than the static HTML.

https://dvc.org/doc/start/data-management/metrics-parameters-plots

https://dvc.org/doc/start/experiments/visualization

@jorgeorpinel @shcheklein Thoughts?

shcheklein:
Yes, 💯 , I agree.

shcheklein:
@jorgeorpinel could you take this one?

jorgeorpinel:
Sure. I started [working on it](https://github.com/iterative/dvc.org/pull/3735) and left this follow-up Q from https://github.com/iterative/dvc.org/issues/3718#issuecomment-1175613958:

> ... do we add Studio links in all (or most) places where VS Code is linked from? Or just the plots-related ones for now (i.e. in the same PR) ? Lmk

> https://dvc.org/doc/start/data-management/metrics-parameters-plots
https://dvc.org/doc/start/experiments/visualization

BTW I added it to the parent pages for now (see PR). Do we want to repeat it in those specific sub-pages too? (Note the whole Get Started needs [yet another reorg](https://github.com/iterative/dvc.org/issues/3678)...)

jorgeorpinel:
Continuing discussion from https://github.com/iterative/dvc.org/pull/3735#issuecomment-1176695901:

> For visualizing plots or experiments, we may want to treat VS Code as a first-class option and integrate them more seamlessly into the docs...
> 1. From https://dvc.org/doc/command-reference/plots ... mention that plots can be visualized in VS Code, Studio, or static HTML. Static HTML feels like an option of last resort now
> 2. A lot of https://dvc.org/doc/user-guide/experiment-management/comparing-experiments could be skipped by a user if they use VS Code
> 3. In https://dvc.org/doc/start/experiments/visualization, it's not that clear how these plots are useful... Showing the image alongside other plots or compared to other experiments in VS Code could help clarify

[...]

PR #3735, opened 2022-07-05
vs-code: more links

jorgeorpinel:
~~Rel. https://github.com/iterative/dvc.org/issues/3718#issuecomment-1175589529~~ https://github.com/iterative/dvc.org/issues/3717

### DIFF
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -49,6 +49,15 @@ Data-series plots utilize [Vega-Lite](https://vega.github.io/vega-lite/) for
 rendering (declarative JSON grammar for defining graphics). Image-type plots are
 rendered using `<img>` tags directly.
 
+<admon icon="fire">
+
+**New!** DVC Plots are available [inside the VS Code IDE], a quick and intuitive
+UI for DVC Experiments.
+
+[inside the vs code ide]: /doc/vs-code-extension
+
+</admon>
+
 ## Supported file formats
 
 Image-type plots are included in HTML as-is, without additional processing.
--- a/content/docs/start/experiments/index.md
+++ b/content/docs/start/experiments/index.md
@@ -14,6 +14,15 @@ ones that we need to Git.
 
 https://youtu.be/FHQq_zZz5ms
 
+<admon icon="fire">
+
+**New!** You can track and compare you ML experiments with DVC directly [from
+Visual Studio Code], a leading IDE in the industry.
+
+[from visual studio code]: /doc/vs-code-extension
+
+</admon>
+
 In this section, we explore the basic features of DVC experiment management with
 the [`example-dvc-experiments`][ede] project.
 


content/docs/command-reference/plots/index.md
```diff
@@ -49,6 +49,15 @@ Data-series plots utilize [Vega-Lite](https://vega.github.io/vega-lite/) for
 rendering (declarative JSON grammar for defining graphics). Image-type plots are
 rendered using `<img>` tags directly.
 
+<admon icon="fire">
+
+**New!** DVC Plots are available [inside the VS Code IDE], a quick and intuitive
+UI for DVC Experiments.
+
+[inside the vs code ide]: /doc/vs-code-extension
```
jorgeorpinel, line 57: Should we link directly to https://github.com/iterative/vscode-dvc/blob/main/extension/resources/walkthrough/plots.md in this case?

content/docs/start/experiments/index.md
```diff
@@ -14,6 +14,15 @@ ones that we need to Git.
 
 https://youtu.be/FHQq_zZz5ms
 
+<admon icon="fire">
+
+**New!** You can track and compare you ML experiments with DVC directly [from
+Visual Studio Code], a leading IDE in the industry.
+
+[from visual studio code]: /doc/vs-code-extension
```
jorgeorpinel, line 22: Unrelated but note: these relative links are redirected to https://marketplace.visualstudio.com/items?itemName=Iterative.dvc (actually the engine renders that external link directly so no redirect happens technically).

> This was in case we decided to write official docs for the extension in this site, which is still a possibility, I think.

dberenbaum:
Thanks @jorgeorpinel! Looks good although I have some questions about the best way to integrate VS Code references. I'm open to either merging as is and continuing discussion after or having discussion before merging.

If I were reading the docs, these would look like ads to me that I might check out later after I've read the rest of the doc. For visualizing plots or experiments, we may want to treat VS Code as a first-class option and integrate them more seamlessly into the docs.

Some examples:

1. From https://dvc.org/doc/command-reference/plots#description:

> DVC generates plots as static HTML webpages that can be open with a web browser. They can also be saved as SVG or PNG image files from the browser.

I would mention that plots can be visualized in VS Code, Studio, or static HTML. Static HTML feels like an option of last resort now that we have so many better ways to visualize plots.

2. A lot of https://dvc.org/doc/user-guide/experiment-management/comparing-experiments could be skipped by a user if they use VS Code. We could just tell them their experiments can be shown, filtered, and sorted there.

3. In https://dvc.org/doc/start/experiments/visualization, it's not that clear how these plots are useful. For example, there is a section on how to display an image file without much context about why you would want/need DVC to do that. Showing the image alongside other plots or compared to other experiments in VS Code could help clarify how plots visualization is useful in experimentation.

jorgeorpinel:
Good points @dberenbaum 

If you don't mind approving this for now so we can merge it and iterate on top of that. There's several of these admonitions in other pages already anyway.

jorgeorpinel:
> these would look like ads to me that I might check out later after I've read the rest of the doc

They're basically ads as we have them now.

We don't have much documentation on what experiments are or how they work (or any DVC feature in detail for that matter) in the VS Code side and it's unclear whether we should. To learn about that you still have to refer to the dvc.org/doc .

> For visualizing plots or experiments, we may want to treat VS Code as a first-class option and integrate them more seamlessly into the docs

I see where you're going. This PR does not attempt that. Let's discuss in #3717 ... Thanks

shcheklein:
Yep, I also have a feeling that it looks like an add. Not in a bad sense (we don't advertise a commercial product after all), but in terms of how useful it is and how disruptive it is in the document. I would also expect more "embedded" links.

dberenbaum:
@shcheklein WDYT about merging as is and then continuing discussion? I'm fine with whatever you and @jorgeorpinel  think is best here for now.

shcheklein:
I'm totally fine with that @dberenbaum .

dberenbaum (approved):

PR #3744, opened 2022-07-07
using GCP remote blog: minor cleanup

shcheklein:


shcheklein:
@flippedcoder I made all the code blocks look the same and highlighted, ptal

also, images - tbh it seems you took a very large portion of the screen. They look consistent, but very hard to read. We can subtitle (click to open), we can also highlight/zoom the proper place on the screen to pay attention to. I would make way easier to read.

and one more question (sorry that it's so late after we merged this, no need to fix this probably - may be add a note?) - why do show the service account workflow? By default I though that just your personal account is enough, if GCP tools are setup locally you don't need extra credentials, etc? Service accounts are needed usually to automate things or for extra layer of security.

shcheklein:
cc @flippedcoder

flippedcoder:
> why do show the service account workflow?

I think when the idea originally came up, there were some DVC users getting confused in that part of the flow so this is just here to help them out.

flippedcoder (approved):

shcheklein:
@flippedcoder yes, but there should be then a better motivation / explanation when service accounts make sense. They have a cost of using them.

shcheklein:
@flippedcoder and this blog post looks like a general / recommended way of doing auth. People will be coming and reading this w/o even realizing there are other (usually simpler ways) to do this + w/o any help navigating through the different options.

PR #3745, opened 2022-07-07
Fix typo

### LINKED ISSUE #3747: Small typo in checkout.md
### Report

There's a small typo in the `dvc fetch` description.

<https://dvc.org/doc/command-reference/checkout#examples>

> We run `dvc fetch` with the `-aT` flags to get the DVC-tracked data from all Git branches and tags from [remote storage](/doc/command-reference/remote) to the <abbr>cache</abbr>. This way it's all available for the `checkout` examples below.

It should be:

We run `dvc fetch` with the `-aT` flags to get the DVC-tracked data ~~from~~ **for** all Git branches and ...

huang06:
Fix #3747

### DIFF
--- a/content/docs/command-reference/checkout.md
+++ b/content/docs/command-reference/checkout.md
@@ -124,7 +124,7 @@ $ cd example-get-started
 $ dvc fetch -aT
 ```
 
-We run `dvc fetch` with the `-aT` flags to get the DVC-tracked data from all Git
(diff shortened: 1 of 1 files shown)


jorgeorpinel (commented):
Hmmmm. Thanks @huang06 but I'm not sure this needed fixing 🙂 

DVC reads metafiles **from** different Git revisions, not _for_ them. "For" could also work (figurative speech), but "from" is more correct.

Sorry

PR #3767, opened 2022-07-14
Remove static directory after heroku build

rogermparent:
I don't think we use this directory in production as all contents are copied to `public`. This will help us with our slug size issue on our websites, particularly with blogs.

The resulting slug is much smaller, and will probably scale just fine since it doesn't include infinitely scaling images.
```
-----> Compressing...

       Done: 192.1M
```


yathomasi (approved):
That's great.

Issue #3765, opened 2022-07-13
Upgrade Heroku deployment stack to latest


yathomasi:
### Report

<img width="1222" alt="Screen Shot 2022-07-13 at 07 39 04" src="https://user-images.githubusercontent.com/20840228/178633783-11e6403a-91f6-48d0-b0d5-42f729f53236.png">

PR #3769, opened 2022-07-14
Upgrade heroku stack

yathomasi:
Upgrade Heroku stack. Part of ( #3765 )


rogermparent (approved):
Seems like everything still works!

PR #3728, opened 2022-07-05
Rename view to project

tapadipti:
Related to https://github.com/iterative/studio/pull/3842

Changes to docs in Studio to reflect that **views** have been renamed to **projects**. Several pages have been changed, but each change is minimal (only changed **views** to **projects**).

Screenshots are yet to be replaced. They still use the old Studio UI version which is currently in prod (which uses the term **views**, not **projects**).

PR #3772, opened 2022-07-18
hotfix: redirects

jorgeorpinel:
Per https://github.com/iterative/dvc.org/pull/3728#pullrequestreview-1041204655:

> We shouldn't change existing (301) redirects, especially if they're not quite old (1y+). We should only add more (subsequent) redirects

### DIFF
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -88,7 +88,7 @@
   "^/doc/studio/user-guide/run-experiments$                                               /doc/studio/user-guide/views/run-experiments",
   "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project 302",
   "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings 302",
-  "^/doc/studio/user-guide/views(/.*)$                                                    /doc/studio/user-guide/projects$1 302",
+  "^/doc/studio/user-guide/views(/.*)?$                                                    /doc/studio/user-guide/projects$1 302",
 
   "^/(.+)/$                                                                               /$1"
 ]
--- a/redirects-list.json
+++ b/redirects-list.json
@@ -83,16 +83,15 @@
   "^/doc/vs-code-extension$                                                               https://marketplace.visualstudio.com/items?itemName=Iterative.dvc 302",
 
   "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/views/view-settings",
-  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings",
   "^/doc/studio/user-guide/explore-experiments$                                           /doc/studio/user-guide/views/explore-experiments",
   "^/doc/studio/user-guide/visualize-experiments$                                         /doc/studio/user-guide/views/visualize-experiments",
   "^/doc/studio/user-guide/run-experiments$                                               /doc/studio/user-guide/views/run-experiments",
-  "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project",
-  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings",
-  "^/doc/studio/user-guide/views/explore-experiments$                                     /doc/studio/user-guide/projects/explore-experiments",
-  "^/doc/studio/user-guide/views/visualize-experiments$                                   /doc/studio/user-guide/projects/visualize-experiments",
-  "^/doc/studio/user-guide/views/run-experiments$                                         /doc/studio/user-guide/projects/run-experiments",
-  "^/doc/studio/user-guide/views(/.*)$                                                    /doc/studio/user-guide/projects$1",
+  "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project 302",
+  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings 302",
+  "^/doc/studio/user-guide/views/explore-experiments$                                     /doc/studio/user-guide/projects/explore-experiments 302",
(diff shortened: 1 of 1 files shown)


redirects-list.json
```diff
@@ -82,14 +82,16 @@
...
+  "^/doc/studio/user-guide/visualize-experiments$                                         /doc/studio/user-guide/views/visualize-experiments",
+  "^/doc/studio/user-guide/run-experiments$                                               /doc/studio/user-guide/views/run-experiments",
+  "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project 302",
+  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings 302",
+  "^/doc/studio/user-guide/views/explore-experiments$                                     /doc/studio/user-guide/projects/explore-experiments 302",
+  "^/doc/studio/user-guide/views/visualize-experiments$                                   /doc/studio/user-guide/projects/visualize-experiments 302",
+  "^/doc/studio/user-guide/views/run-experiments$                                         /doc/studio/user-guide/projects/run-experiments 302",
+  "^/doc/studio/user-guide/views(/.*)$                                                    /doc/studio/user-guide/projects$1 302",
```
jorgeorpinel, line 94: Note that I'm making all the redirects to /doc/studio/user-guide/projects... 302s so we can just remove them in https://github.com/iterative/dvc.org/pull/3621, since they're pretty recent (hopefully not too many browsers have cached them as 301s).

Or should I leave them and add yet another redirect in #3621 ?

shcheklein (approved):

tapadipti (approved):

redirects-list.json
```diff
@@ -82,14 +82,16 @@
...
 
-  "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/projects/project-settings",
-  "^/doc/studio/user-guide/explore-experiments$                                           /doc/studio/user-guide/projects/explore-experiments",
-  "^/doc/studio/user-guide/visualize-experiments$                                         /doc/studio/user-guide/projects/visualize-experiments",
-  "^/doc/studio/user-guide/run-experiments$                                               /doc/studio/user-guide/projects/run-experiments",
-
-  "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project",
-  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings",
-  "^/doc/studio/user-guide/views/(.*)$                                                    /doc/studio/user-guide/projects/$1",
+  "^/doc/studio/view-settings$                                                            /doc/studio/user-guide/views/view-settings",
```
yathomasi, line 85: Why are we not directly redirecting it to the final version instead of multiple redirects 🤔?
i.e
 ```
"^/doc/studio/view-settings$     /doc/studio/user-guide/projects/project-settings 302",
```

Comment (jorgeorpinel) [reply redirects-list.json:85]: Because the 301 redirect to `/doc/studio/user-guide/views/` has been published for a while already (recently removed [in #3728](https://github.com/iterative/dvc.org/pull/3728/files#diff-a6828b19e9794c1fbb6799ea95263fb7487b676d054932b5f6e3b315eb12a6a2)) and as stated in the OP, we shouldn't change existing (301) redirects. That's because 301s get cached indefinitely by browsers so once we publish them they should be considered "final" (which I think means we need to wait about a year before removing them), especially if they've been published for a while as is the case here.

> BTW this list needs an urgent clean up I think...

Comment (jorgeorpinel) [reply redirects-list.json:85]: p.s. in this case maybe the cached redirects did get caught by the `^/doc/studio/user-guide/views/(.*)$` redirect but still, as a best practice we should avoid changing existing 301 redirects until after a long time has passed, I think. Up to you @yathomasi lmk what to do here please.

File: redirects-list.json (hunk 1 above)
Comment (jorgeorpinel) [inline redirects-list.json:94]: I guess these will get caught by the last one! Removing
```suggestion
  "^/doc/studio/user-guide/views(/.*)$                                                    /doc/studio/user-guide/projects$1 302",
```

redirects-list.json
```diff
@@ -82,14 +82,13 @@
...
+  "^/doc/studio/user-guide/explore-experiments$                                           /doc/studio/user-guide/views/explore-experiments",
+  "^/doc/studio/user-guide/visualize-experiments$                                         /doc/studio/user-guide/views/visualize-experiments",
+  "^/doc/studio/user-guide/run-experiments$                                               /doc/studio/user-guide/views/run-experiments",
+  "^/doc/studio/user-guide/views/create-view$                                             /doc/studio/user-guide/projects/create-project 302",
+  "^/doc/studio/user-guide/views/view-settings$                                           /doc/studio/user-guide/projects/project-settings 302",
+  "^/doc/studio/user-guide/views(/.*)$                                                    /doc/studio/user-guide/projects$1 302",
```
jorgeorpinel, line 91: After sanity check this was the only error:
```suggestion
  "^/doc/studio/user-guide/views(/.*)?$                                                    /doc/studio/user-guide/projects$1 302",
```

redirects-list.json
jorgeorpinel: I merged this to get moving (needed for #3621) but lmk what you think @yathomasi and we can address redirects separately (see https://github.com/iterative/dvc.org/issues/3776)

yathomasi: Umm, I think the existing cached 301 redirects will still be working even if we point the redirects directly to the final version.
Existing:
`^/doc/studio/view-settings$` -> `/doc/studio/user-guide/views/view-settings` 
`/doc/studio/user-guide/views/view-settings` -> `/doc/studio/user-guide/projects/project-settings 302`

Proposed:
`^/doc/studio/view-settings$` -> `/doc/studio/user-guide/projects/project-settings 302`
`/doc/studio/user-guide/views/view-settings` -> `/doc/studio/user-guide/projects/project-settings 302`

So, the link `^/doc/studio/view-settings$` will not have to jump into multiple redirects and the previous redirects will still be working.

PR #3414, opened 2022-04-05
guide: ML Pipelines (1): Defining Pipelines & Stages

iesahin:
Related to #2883

- [x] [Index page](https://dvc-org-iesahin-ug-pipe-qykfpz.herokuapp.com/doc/user-guide/machine-learning-pipelines/defining-pipelines): General explanation about ML Pipelines
- [x] [Defining Pipelines](https://dvc-org-iesahin-ug-pipe-qykfpz.herokuapp.com/doc/user-guide/machine-learning-pipelines/defining-pipelines) page. UPDATE: Jump to https://github.com/iterative/dvc.org/pull/3414#issuecomment-1238505396. Covers:
  - Stages
  - Dependency graph (DAG)
  - `dvc.yaml` intro
  - Commands (`cmd`) details
  - Dependencies (overview)
- [x] Remove explanations on these topics from other pages (mainly `run`, `stage add`, and `repro` refs)
- [x] Update many links accordingly
- [x] Add links from appropriate basic concept tooltips
- [x] Update Glossary accordingly?

In review app: https://dvc-org-iesahin-ug-pipe-qykfpz.herokuapp.com/doc/user-guide/data-pipelines

--- 

- [ ] Merge https://github.com/iterative/dvc.org/pull/3899 when this is approved (then merge to `main` together).

PR #3787, opened 2022-07-21
guide: `dvc.yaml` doc updates

jorgeorpinel:
Extracted from https://github.com/iterative/dvc.org/pull/3414#issuecomment-1189898267

This page should focus on the schema/spec of the file and it's sections more than guidance and explaining concepts, since we'll have a guide for that now (see https://github.com/iterative/dvc.org/pull/3414).


jorgeorpinel:
Something is off. Why are there 14 files changed? ... Starting over...

PR #3805, opened 2022-07-25
Fix link-check-all and improve link-check-deploy

rogermparent:
https://github.com/iterative/mlem.ai/pull/144 is the primary PR, https://github.com/iterative/dvc.org/pull/3805 and https://github.com/iterative/cml.dev/pull/281 are the same thing for different sites

This PR fixes `link-check-all` by giving it the same treatment as `link-check-deploy` and using the CLI instead of a GitHub Action to run the check. It also makes some improvements to `link-check-deploy`'s workflow, biggest of which is using [an existing action](https://github.com/marketplace/actions/find-current-pull-request) to find the current PR instead of using a custom github script snippet.


yathomasi (approved):
LGTM!

PR #3811, opened 2022-07-26
node: unrestrict node version to allow 18.x

skshetry:


yathomasi (approved):
Thanks, @skshetry for the upgrade. We were having a problem running with the Apple M1 device previously so we didn't upgrade it. Now, I can verify it's working fine.

jorgeorpinel:
Nice! I believe this needs docs updates though? E.g. https://dvc.org/doc/contributing/docs#development-environment

skshetry:
@rogermparent @yathomasi, is it possible to make `sharp` an `optionalDependency`? I like to be on the latest version, so it's always painful to use dvc.org. I guess it's sharp that usually holds us back to support the latest version.

yathomasi:
> is it possible to make `sharp` an `optionalDependency`?

I am afraid it's not. Gatsby itself uses `sharp` to process and optimize the images.

The best approach would be to use the node version manager locally. 
1. https://github.com/nvm-sh/nvm
2. https://volta.sh/ ( I prefer this)

PR #3821, opened 2022-07-27
update GDrive user guide example output

shcheklein:


efiop (approved):
Ah, so this did indeed change https://github.com/iterative/dvc/issues/8058. Great. Thank you!

PR #3834, opened 2022-07-29
ref: Hydra support for `exp run --set-param`

daavoo:
P.R. for https://github.com/iterative/dvc/pull/8067

### DIFF
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -242,7 +242,7 @@ Path         Param              HEAD    workspace
 params.yaml  prepare.new_param  -       foo
 ```
 
-- `~<param_name>=<param_value>`
+- `~<param_name>`
 
 Removes existing parameters:
 
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -26,9 +26,13 @@ directories, etc.
 `dvc exp run` has the same general behavior as `dvc repro` when it comes to
 `targets` and stage execution (restores the dependency graph, etc.).
 
-> This includes committing any changed data <abbr>dependencies</abbr> to the
-> <abbr>DVC cache</abbr> when preparing the experiment, which can take some
-> time. See the [Options](#options) section for the differences.
+<admon type="info">
+
+This includes committing any changed data <abbr>dependencies</abbr> to the
+<abbr>DVC cache</abbr> when preparing the experiment, which can take some time.
+See the [Options](#options) section for the differences.
+
+</admon>
 
 Use the `--set-param` (`-S`) option as a shortcut to change
 <abbr>parameter</abbr> values [on-the-fly] before running the experiment.
@@ -38,15 +42,22 @@ flag. To actually run them, use `dvc exp run --run-all`. Queued experiments are
 run sequentially by default, but can be run in parallel using the `--jobs`
 option.
 
-> ⚠️ Parallel runs are experimental and may be unstable. Make sure you're using
-> a number of jobs that your environment can handle (no more than the CPU
-> cores).
+<admon type="warn">
+
+Parallel runs are experimental and may be unstable. Make sure you're using a
+number of jobs that your environment can handle (no more than the CPU cores).
+
+</admon>
 
 It's also possible to run special [checkpoint experiments] that log the
 execution progress (useful for deep learning ML). The `--rev` and `--reset`
 options have special uses for these.
 
-> 📖 See the [Running Experiments] guide for more details on all these features.
+<admon type="info">
+
+See the [Running Experiments] guide for more details on all these features.
+
+</admon>
 
 [Review] your experiments with `dvc exp show`. Successful ones can be [made
 persistent] by restoring them via `dvc exp branch` or `dvc exp apply` and
@@ -70,11 +81,11 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
 - `-S [<filename>:]<override_pattern>`,
   `--set-param [<filename>:]<override_pattern>` - set the value of existing
   `dvc params` for this experiment. `filename` can be any valid params file
-  (`params.yaml` by default). This will override the param values coming from
-  the params file.
+  (`params.yaml` by default). This will override the param file before running
+  the experiment.
 
-  Valid `override_pattern` values are defined in the
-  [Hydra Override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object).
+  Check the [examples](#example-modify-parameters-on-the-fly) for valid values
+  of `<override_pattern>`.
 
 - `-n <name>`, `--name <name>` - specify a [unique name] for this experiment. A
   default one will generated otherwise, such as `exp-f80g4` (based on the
@@ -186,25 +197,85 @@ experiment we just ran (`exp-44136`).
 `dvc exp run--set-param` (`-S`) saves you the need to manually edit the params
 file before running an experiment.
 
+<admon type="warn">
+
+Note that `--set-param` doesn't update your `dvc.yaml`. If you are tracking each
+parameter individually, make sure to update your `dvc.yaml` accordingly when
+appending or removing parameters.
+
+</admon>
+
+You can optionally provide a prefix `[<filename>:]` in order to edit a specific
+params file. If not provided, `params.yaml` will be used as default.
+
+You can use the different syntaxes of `<override_pattern>` for modifying the
+parameters:
+
+- `<param_name>=<param_value>`
+
+Overrides existing parameters:
+
 ```dvc
 $ dvc exp run -S prepare.split=0.25 -S featurize.max_features=2000
 ...
-Reproduced experiment(s): exp-18bf6
 ```
-
-To see the results, you can use `dvc exp diff`. It compares both params and
-metrics to the previous project version:
 
 ```dvc
-$ dvc exp diff
-Path         Metric    Value    Change
-scores.json  avg_prec  0.58187  -0.022184
-scores.json  roc_auc   0.93634  -0.024464
-
-Path         Param                   Value    Change
-params.yaml  featurize.max_features  2000     -1000
-params.yaml  prepare.split           0.25     0.05
+$ dvc params diff
+Path         Param                   HEAD    workspace
+params.yaml  featurize.max_features  1000    2000
+params.yaml  prepare.split           0.20    0.25
 ```
 
-> Notice that experiments run as a series don't build up on each other. They are
-> all based on `HEAD`.
+- `+<param_name>=<param_value>`
+
+Appends new parameters:
+
+```dvc
+$ dvc exp run -S +prepare.new_param=foo
+...
+```
+
+```dvc
+$ dvc params diff
+Path         Param              HEAD    workspace
+params.yaml  prepare.new_param  -       foo
+```
+
+- `~<param_name>=<param_value>`
+
+Removes existing parameters:
+
+```dvc
+$ dvc exp run -S ~prepare.seed
+...
+```
+
+```dvc
+$ dvc params diff
+Path         Param         HEAD      workspace
+params.yaml  prepare.seed  20170428  -
+```
+
+- `++<param_name>=<param_value>`
+
+Appends or overrides parameters:
+
+```dvc
+$ dvc exp run -S ++prepare.new_param=foo ++prepare_split=0.1
+...
+```
+
+```dvc
+$ dvc params diff
+Path         Param              HEAD    workspace
+params.yaml  prepare.new_param  -       foo
+params.yaml  prepare.split      0.2     0.1
+```
+
+<admon type="tip">
+
+Experiments run as a series don't build up on each other. They are all based on
+`HEAD`.
+
+</admon>
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -208,8 +208,7 @@ appending or removing parameters.
 You can optionally provide a prefix `[<filename>:]` in order to edit a specific
 params file. If not provided, `params.yaml` will be used as default.
 
-You can use the different syntaxes of `<override_pattern>` for modifying the
-parameters:
+`<override_pattern>` supports different syntaxes for modifying the parameters:
 
 - `<param_name>=<param_value>`
 
--- a/content/docs/command-reference/exp/run.md
+++ b/content/docs/command-reference/exp/run.md
@@ -8,7 +8,7 @@ Run or resume a
 ```usage
 usage: dvc exp run [-h] [-q | -v] [-f]
                    { repro options ... } [-n <name>]
-                   [-S [<filename>:]<params_list>]
+                   [-S [<filename>:]<override_pattern>]
                    [--queue] [--run-all] [-j <number>] [--temp]
                    [-r <experiment_rev>] [--reset]
                    [targets [targets ...]]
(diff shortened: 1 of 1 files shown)


dberenbaum:
One tricky issue here is that we update `params.yaml` or other params files with the `+` and `~` syntax to add/remove parameters, but we don't update `dvc.yaml`.

For example, assume you have a `dvc.yaml` like:

```yaml
stages:
  train:
    cmd: python train.py
    params:
    - params.yaml:
      - train.lr
...
```

* `dvc exp run --set-param=~train.optim` will fail because the param will be dropped from `params.yaml` but `dvc.yaml` still expects it.
* `dvc exp run --set-param=+train.optim=Adam` will add the param to params.yaml but not track it as a new parameter.'

So this syntax only really works when specifying a whole params file or a key like `train` for a group of params.

daavoo:
> One tricky issue here is that we update `params.yaml` or other params files with the `+` and `~` syntax to add/remove parameters, but we don't update `dvc.yaml`.
> 
> For example, assume you have a `dvc.yaml` like:
> 
> ```yaml
> stages:
>   train:
>     cmd: python train.py
>     params:
>     - params.yaml:
>       - train.lr
> ...
> ```
> 
>     * `dvc exp run --set-param=~train.optim` will fail because the param will be dropped from `params.yaml` but `dvc.yaml` still expects it.
> 
>     * `dvc exp run --set-param=+train.optim=Adam` will add the param to params.yaml but not track it as a new parameter.'
> 
> 
> So this syntax only really works when specifying a whole params file or a key like `train` for a group of params.

Good point. 
Do you think it's worth adding an explicit note in the ref? We could add an `<admon>` warning?

The ref doesn't really lie, as we state that we update the values in the params file and don't mention `dvc.yaml`. 
Maybe it's better to cover it as part of a user guide?

dberenbaum:
Do we need to specify if there is any particular syntax that is not supported? For example, in the current PR, I don't think https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-defaults-list is applicable (although it probably will be in an upcoming one). 

Should we have examples or is it repeating Hydra docs too much? Possible examples:
* `+newparam=1`
* `~dropparam`
* `custom_file.json:+newparam=1`
* `+newparam=${oldparam}`
* `param=[1,2,3]`

dberenbaum:
Leaving high-level product thoughts here instead of https://github.com/iterative/dvc/pull/8067. Overall, it looks great as a first step. Seems more flexible and stable than what we have now.

My only additional concern is to not force users immediately to the Hydra docs to start using `--set-param` since that's pretty daunting. I think we should make clear the basic syntax at least (`--set-param filename:key=value`) without referring users to Hydra.

dberenbaum:
> Do you think it's worth adding an explicit note in the ref (I was avoiding including info that we will end up extracting to user guide)? We could add an `<admon>` warning

No strong opinion as long as we don't lose track of documenting it.

daavoo:
@dberenbaum Updated with examples for basic syntax and warning about `dvc.yaml` requiring update

content/docs/command-reference/exp/run.md
```diff
@@ -67,11 +78,14 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
-  file (`params.yaml` by default). This will override the param values coming
-  from the params file.
+- `-S [<filename>:]<override_pattern>`,
+  `--set-param [<filename>:]<override_pattern>` - set the value of existing
+  `dvc params` for this experiment. `filename` can be any valid params file
+  (`params.yaml` by default). This will override the param file before running
+  the experiment.
+
+  Check the [examples](#example-modify-parameters-on-the-fly) for valid values
+  of `<override_pattern>`.
```
dberenbaum, line 88: Sorry, I didn't mean that we should avoid mentioning Hydra. I think it's important to mention that Hydra override syntax is supported as you did before, but I thought it might be helpful to first note how it works without any special Hydra syntax.

dberenbaum: Before you had written:

> Valid `override_pattern` values are defined in the
  [Hydra Override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object).

Would you say that's still correct or do you think we need to specify what is and isn't valid?

content/docs/command-reference/exp/run.md
```diff
@@ -183,25 +197,84 @@ experiment we just ran (`exp-44136`).
 `dvc exp run--set-param` (`-S`) saves you the need to manually edit the params
 file before running an experiment.
 
+<admon type="warn">
+
+Note that `--set-param` doesn't update your `dvc.yaml`. If you are tracking each
+parameter individually, make sure to update your `dvc.yaml` accordingly when
+appending or removing parameters.
```
dberenbaum, line 204: Do you think it would be helpful to link to https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#parameter-dependencies? No strong opinion to add it.

Note: we could probably better call out support for groups of params there.

content/docs/command-reference/exp/run.md
```diff
@@ -183,25 +197,84 @@ experiment we just ran (`exp-44136`).
...
+
+</admon>
+
+You can optionally provide a prefix `[<filename>:]` in order to edit a specific
+params file. If not provided, `params.yaml` will be used as default.
+
+`<override_pattern>` supports different syntaxes for modifying the parameters:
```
dberenbaum, line 211: Again, I would mention Hydra.

What do you think about the examples? I'm not sure if they are helpful or too repetitive of Hydra docs.

I think we should have an example like `custom_file.json:+newparam=1` because it's easy for me to think it should be `+custom_file.json:newparam=1`.

dberenbaum:
Looks good @daavoo! Can you resolve the merge conflict and do one more iteration? I think we could merge as is, but since there is no rush here, let's try to work on it a bit more.

content/docs/command-reference/exp/run.md
```diff
@@ -180,27 +192,36 @@ experiment we just ran (`exp-44136`).
...
-
-Path         Param                   Value    Change
-params.yaml  featurize.max_features  2000     -1000
-params.yaml  prepare.split           0.25     0.05
+$ dvc params diff
+Path              Param      HEAD    workspace
+custom_file.json  new_param  -       foo
 ```
 
-> Notice that experiments run as a series don't build up on each other. They are
```
daavoo, line 205: Removed this note at the bottom of the page that felt a little out of context @jorgeorpinel . Not sure when/why it was added

daavoo:
> Looks good @daavoo! Can you resolve the merge conflict and do one more iteration? I think we could merge as is, but since there is no rush here, let's try to work on it a bit more.



> Looks good @daavoo! Can you resolve the merge conflict and do one more iteration? I think we could merge as is, but since there is no rush here, let's try to work on it a bit more.

Tried to address the comments. PTAL

content/docs/command-reference/exp/run.md
dberenbaum: > I tried to link to the `Modifying the Config Object` section. Maybe we just need to specify that `Modifying the Defaults List` is not supported

I think it will be after https://github.com/iterative/dvc/pull/8093, right? It's not even unsupported as much as inapplicable here I think. IMO it's not worth mentioning.

content/docs/command-reference/exp/run.md
```diff
@@ -63,11 +71,15 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
 > In addition to the following, `dvc exp run` accepts the options in `dvc repro`
 > except for `--glob`, `--no-commit`, and `--no-run-cache`.
 
-- `-S [<filename>:]<param_name>=<param_value>`,
-  `--set-param [<filename>:]<param_name>=<param_value>` - set the value of
```
dberenbaum, line 67: What do you think about keeping the basic `<param_name>=<param_value>` syntax in the description somewhere?

dberenbaum:
@jorgeorpinel Do you want to take a final pass at this or not?

dberenbaum:
@alex000kim Maybe you want to review this one? Note that this is only for part 1 of the Hydra integration, which only includes https://github.com/iterative/dvc/pull/8067, not https://github.com/iterative/dvc/pull/8093.

content/docs/command-reference/exp/run.md
```diff
@@ -63,11 +71,18 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
+  `dvc params` for this experiment. `filename` can be any valid params file
+  (`params.yaml` by default). This will override the param file before running
+  the experiment.
+
+  For example, to override the value an existing param, you can use
+  `--set-param <param_name>=<param_value>`.
+
+  Valid `<override_pattern>` values are defined in the
+  [Hydra Override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object).
+  Patterns modifying hydra's defaults list are not supported.
```
shcheklein, line 85: Hydra's

content/docs/command-reference/exp/run.md
```diff
@@ -63,11 +71,18 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
 
 - `-n <name>`, `--name <name>` - specify a [unique name] for this experiment. A
   default one will generated otherwise, such as `exp-f80g4` (based on the
```
alex000kim, line 88: `default one will _be_ generated otherwise`

content/docs/command-reference/exp/run.md
```diff
@@ -63,11 +71,18 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
+  `--set-param [<filename>:]<override_pattern>` - set the value of existing
+  `dvc params` for this experiment. `filename` can be any valid params file
+  (`params.yaml` by default). This will override the param file before running
+  the experiment.
+
+  For example, to override the value an existing param, you can use
+  `--set-param <param_name>=<param_value>`.
+
+  Valid `<override_pattern>` values are defined in the
+  [Hydra Override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object).
```
alex000kim, line 84: Below we refer to the same page as `Hydra Override Grammar`.

content/docs/command-reference/exp/run.md
```diff
@@ -180,27 +195,36 @@ experiment we just ran (`exp-44136`).
 ## Example: Modify parameters on-the-fly
 
 `dvc exp run--set-param` (`-S`) saves you the need to manually edit the params
-file before running an experiment.
+file before running an experiment. It can override (`foo.bar=value`), append
```
alex000kim, line 198: Can we use more obvious (from an ML point of view) examples for parameter names instead of `foo.bar`?

content/docs/command-reference/exp/run.md
```diff
@@ -180,27 +195,36 @@ experiment we just ran (`exp-44136`).
...
+with the exception of patterns that modify Hydra's defaults list.
+
+</admon>
+
+You can optionally provide a prefix `[<filename>:]` in order to edit a specific
+`dvc params` file. If not provided, `params.yaml` will be used as default.
 
 ```dvc
-$ dvc exp run -S prepare.split=0.25 -S featurize.max_features=2000
+$ dvc exp run -S custom_file.json:+newparam=foo
```
alex000kim, line 213: Same as above: 
I find `newparam=foo` too generic and it's not clear why this syntax is useful

content/docs/command-reference/exp/run.md
```diff
@@ -180,27 +195,36 @@ experiment we just ran (`exp-44136`).
...
+$ dvc params diff
+Path              Param      HEAD    workspace
+custom_file.json  new_param  -       foo
 ```
 
-> Notice that experiments run as a series don't build up on each other. They are
-> all based on `HEAD`.
+<admon type="warn">
+
+Note that `--set-param` doesn't update your `dvc.yaml`. When appending or
```
alex000kim, line 225: "Note that `--set-param` (or `-S`) doesn't ..."

content/docs/command-reference/exp/run.md
daavoo: Tried to use some more meaningful names

jorgeorpinel:
Hi. I probably won't be reviewing all ref PRs in the future but I'm still trying to check as much as possible for some time to compile best practices for each kind of doc. So

> we update params.yaml or other params files with the + and ~ syntax to add/remove parameters, but we don't update dvc.yaml... So this syntax only really works when specifying a whole params file or a key like train for a group of params.

Yeah this is intriguing. I guess for now let's try to avoid examples that would need a dvc.yaml update (defeating the purpose of "on-the-fly" param setting).

> avoiding including info that we will end up extracting to user guide

There's no PR for that yet right? That's higher level so I should try to be more involved.

> Possible examples...

Yes these are all great. I don't think we covered them all in the example after all?

content/docs/command-reference/exp/run.md
```diff
@@ -42,7 +46,11 @@ It's also possible to run special [checkpoint experiments] that log the
 execution progress (useful for deep learning ML). The `--rev` and `--reset`
 options have special uses for these.
 
-> 📖 See the [Running Experiments] guide for more details on all these features.
+<admon type="info">
+
+See the [Running Experiments] guide for more details on all these features.
+
+</admon>
```
jorgeorpinel, line 53: This book icon links are OK I think. We still use them. Or maybe we need a new `<admon icon="book">` ⏳

content/docs/command-reference/exp/run.md
```diff
@@ -63,14 +71,21 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
-- `-S [<filename>:]<param_name>=<param_value>`,
-  `--set-param [<filename>:]<param_name>=<param_value>` - set the value of
-  existing `dvc params` for this experiment. `filename` can be any valid params
-  file (`params.yaml` by default). This will override the param values coming
-  from the params file.
+- `-S [<filename>:]<override_pattern>`,
+  `--set-param [<filename>:]<override_pattern>` - set the value of existing
+  `dvc params` for this experiment. `filename` can be any valid params file
+  (`params.yaml` by default). This will override the param file before running
+  the experiment.
```
jorgeorpinel, line 78: "override" could be easily misread as "overwrite" here. I think the previous wording was clearer, or did the behavior change?

content/docs/command-reference/exp/run.md
```diff
@@ -63,14 +71,21 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
+
+  For example, to override the value an existing param, you can use
+  `--set-param <param_name>=<param_value>`.
```
jorgeorpinel, line 81: 💅🏼 
```suggestion
  `--set-param name=value`.
```

content/docs/command-reference/exp/run.md
```diff
@@ -63,14 +71,21 @@ committing them to the Git repo. Unnecessary ones can be [cleared] with
...
+
+  Valid `<override_pattern>` values are defined in the
+  [Hydra Override Grammar](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object).
+  Patterns modifying Hydra's defaults list are not supported.
```
jorgeorpinel, line 85: - Clarify that this is a somewhat optional/ advanced feature.
- Use md ref link 💅🏼

content/docs/command-reference/exp/run.md
```diff
@@ -179,28 +194,37 @@ experiment we just ran (`exp-44136`).
...
+file before running an experiment. It can override (`train.epochs=10`), append
+(`+train.weight_decay=0.01`), or remove (`~model.dropout`) parameters.
+
+<admon type="tip">
+
+DVC supports the
+[Hydra Override Grammar](https://hydra.cc/docs/advanced/override_grammar/basic/),
+with the exception of patterns that modify Hydra's defaults list.
+
+</admon>
```
jorgeorpinel, line 207: All of this info should probably be somewhere in the **Description**. An Example is too specific to be the only place where general behavior is described.

content/docs/command-reference/exp/run.md
```diff
@@ -179,28 +194,37 @@ experiment we just ran (`exp-44136`).
...
+
+You can optionally provide a prefix `[<filename>:]` in order to edit a specific
+`dvc params` file. If not provided, `params.yaml` will be used as default.
```
jorgeorpinel, line 211: Same? Or maybe in the corresponding option description in this case

content/docs/command-reference/exp/run.md
```diff
@@ -179,28 +194,37 @@ experiment we just ran (`exp-44136`).
...
-> Notice that experiments run as a series don't build up on each other. They are
-> all based on `HEAD`.
+<admon type="warn">
+
+Note that `exp run --set-param` (`-S`) doesn't update your `dvc.yaml`. When
+appending or removing <abbr>parameters</abbr>, make sure to update the
+[`params` section](https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#parameter-dependencies)
+of your `dvc.yaml` accordingly.
+
+</admon>
```
jorgeorpinel, line 230: Same

jorgeorpinel (commented):
> Don't hesitate to post additional comments and I will address them in subsequent P.R.

OK 😬

content/docs/command-reference/exp/run.md
daavoo: https://github.com/iterative/gatsby-theme-iterative/pull/76

Issue #3832, opened 2022-07-28
guide: working with SSH remote storage


jorgeorpinel:
### Report

See details in https://discuss.dvc.org/t/how-do-i-use-dvc-with-ssh-remote/279 (top forum question)

jorgeorpinel:
Not sure this belongs as it's own guide, part of https://github.com/iterative/dvc.org/issues/2856, a how-to, or for the blog (e.g. [for S3](https://dvc.org/blog/aws-remotes-in-dvc)).

PR #3862, opened 2022-08-03
Add s3 guide

sixhobbits:
Linked to part of https://github.com/iterative/dvc.org/issues/2866
Or https://github.com/iterative/dvc.org/issues/3832

> Perhaps we could try to have guides like [the one for Google Drive](https://dvc.org/doc/user-guide/setup-google-drive-remote) (with screenshots, etc.) for other popular remotes? (no stats to support but AWS and Azure appear to be the most commonly asked about)

<details>

We at Ritza.co love helping companies like DVC produce useful content for their users. Feel free to reach out if you would like more :) 

</details>


jorgeorpinel:
This is pretty cool @sixhobbits , thanks! And we do want to specifically cover S3 remotes in more detail somewhere in our docs (rel https://github.com/iterative/dvc.org/issues/3832).

But we would prefer to avoid documenting Amazon S3 as much as possible at the same time (mention an operation and link to their docs instead). Keep in mind we do have a [recent blog post about this](https://dvc.org/blog/aws-remotes-in-dvc) already, where these screenshots are OK because it's dated (no need to maintain it). Maybe for the docs we should use `aws s3` CLI examples instead?

Also, it seems like there was a learning process to write this, which is greatly appreciated because of the lack of bias (fresh eyes); that said, we can probably skip some steps that may be evident for someone who reaches this guide (e.g. DVC installation), or just mention some things and link to existing docs

sixhobbits:
makes sense! Thanks for the prompt feedback. Going to close this one and if we have left over capacity in future we can look at doing a version with command line examples instead. Keep us in mind if you're ever looking to augment your docs team :)

PR #3864, opened 2022-08-04
Replace protected tweet embed with a static representation

rogermparent:
The tweet that this PR changes recently became protected, likely due to the author marking their account protected for other reasons. This is a problem for us because `gatsby-remark-embedder` throws an error if it runs into a protected tweet, and even if we disabled that and went full client-side we still wouldn't be able to embed it.

This PR puts out the fire by changing the tweet in question into a static representation with a link to the original. The style is modeled after the original, and the change is nearly imperceptible.

https://user-images.githubusercontent.com/9111807/182909746-4f7555b5-613c-4d95-aea3-8aff160fe2e7.mp4

### DIFF
--- a/config/link-check/excluded-links.yml
+++ b/config/link-check/excluded-links.yml
@@ -18,3 +18,4 @@
 - 'https://timheuer.com/blog/skipping-ci-github-actions-workflows/'
 - 'https://www.zazzle.com/t_shirt-235920696568133954'
 - 'https://medium.com**'
+- 'https://twitter.com/flavioclesio/status/*'
--- a/content/blog/2021-03-15-March-21-dvc-heartbeat.md
+++ b/content/blog/2021-03-15-March-21-dvc-heartbeat.md
@@ -177,13 +177,19 @@ image="/uploads/images/2021-03-15/arxiv.png"/>
 
 ## Tweet Love ❤️
 
-From a Porutguese speaking community member in Finland...
+From a Portuguese speaking community member in Finland...
 
 > "The @DVCorg surely it is among the best tools of the ecosystem of the last 3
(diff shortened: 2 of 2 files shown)


content/blog/2021-03-15-March-21-dvc-heartbeat.md
```diff
@@ -177,13 +177,19 @@ image="/uploads/images/2021-03-15/arxiv.png"/>
 
 ## Tweet Love ❤️
 
-From a Porutguese speaking community member in Finland...
+From a Portuguese speaking community member in Finland...
```
rogermparent, line 180: this jumped out at me and I figured I should address it while we're already changing the past here.

jendefig (approved):
Thanks so much @rogermparent!

config/link-check/excluded-links.yml
```diff
@@ -18,3 +18,4 @@
 - 'https://timheuer.com/blog/skipping-ci-github-actions-workflows/'
 - 'https://www.zazzle.com/t_shirt-235920696568133954'
 - 'https://medium.com**'
+- 'https://twitter.com/flavioclesio/status/*'
```
shcheklein, line 21: probably no need to include it if we are replacing it anyways

rogermparent: We still have a link to it under the static tweet, which is what tripped the checker before the most recent commit

shcheklein: my take on this - we should remove it completely (link is indeed broken, we should not have it in the content in the first place probably now)

rogermparent: Yup! #3866 was opened for this. Right now it's keeping the quote, but we could remove the whole thing to be safe.

Issue #3865, opened 2022-08-04
Improve Twitter integration


rogermparent:
With the most recent example being #3864, our current Twitter-embedding solution of `gatsby-remark-embedder`+`gatsby-plugin-twitter` has some downsides:

- Builds break if a tweet is deleted or protected, and they persist unstyled in cache
![image](https://user-images.githubusercontent.com/9111807/182912024-e8b731b8-978b-4288-8812-1a331c28eead.png)
- Loads a whole extra external JS embed bundle from Twitter
- Just looks kinda ugly

(it might just be me, but it really just dumps text like if you copy-pasted it while adding a couple links in. We're downloading a library for _this_?)

![image](https://user-images.githubusercontent.com/9111807/182933798-da52614d-98f9-4dd2-9efe-82c64669af89.png)

As far as an alternative implementation, I can think of two main routes off the top of my head:

## Use a custom component with pasted content

Pros:
- simplest implementation code-wise
- doesn't break regardless of what happens to the tweet
- pretty easy to do overall
- Would need a custom component to add boilerplate that keeps us in line with [Twitter's Tweet display guidelines](https://developer.twitter.com/en/developer-terms/display-requirements)

Cons:
- doesn't update if tweet changes, is updated, or deleted.
- Slightly more difficult on editors than other options

## [`gatsby-source-twitter`](https://www.gatsbyjs.com/plugins/gatsby-source-twitter/)

Pros:
- can maintain the slightly easier workflow of pasting a link instead of pasting a tweet
- can be heavily customized
- (can be) fully static

Cons:
- a significant amount of dev work that could balloon depending on requirements

shcheklein:
> Just looks kinda ugly

Yep, It should not be looking like this (it was not like this for sure)- it's a bug somewhere. It should be like this:


<img width="743" alt="Screen Shot 2022-08-04 at 11 54 10 AM" src="https://user-images.githubusercontent.com/3659196/182931092-81718d46-eb3a-4bde-a926-399966858dba.png">


> Builds break if a tweet is embedded or protected

it's expected, right? in this case we should be removing tweet from the blogs post most likely. What do you mean by `embedded` btw?

> Just paste the tweet in as a blockquote

we can't do this, Twitter has certain rules about embedding their tweets (put a logo, etc). 

----------

[...]

PR #3866, opened 2022-08-04
Revert and remove static tweet

rogermparent:
Reverts iterative/dvc.org#3864

from https://github.com/iterative/dvc.org/issues/3865#issuecomment-1205653477

> > Just paste the tweet in as a blockquote
> 
> we can't do this, Twitter has certain rules about embedding their tweets (put a logo, etc).

> In this case I would just keep a quote and link to person's name.

I merged #3864 before I read this, so this PR changes to a static quote with a name and no tweet link.


yathomasi:
According to this Twitter [display-requirements](https://developer.twitter.com/en/developer-terms/display-requirements), we still need to add a logo.

jendefig:
@yathomasi I'm okay with risking Twitter's wrath for a post that is more than a year old...

jendefig (approved):

PR #3830, opened 2022-07-28
guide: make plots guide

dberenbaum:
This is a quick iteration to address https://github.com/iterative/dvc.org/issues/3818. I made a new guide page since I thought shoving that all into the dvcyaml/project structure would be way too much/messy and hard to find.

Also related: https://github.com/iterative/dvc.org/issues/2572

In review app: https://dvc-org-guide-plots-u9q3eipvdr.herokuapp.com/doc/user-guide/plots

PR #3869, opened 2022-08-08
guide: add high-level plots guide content

dberenbaum:
Based on the comments in https://github.com/iterative/dvc.org/pull/3830, I added new high-level content to the plots guide here and compared to that PR to get feedback on whether it seems like a useful guide with that added info.


shcheklein (approved):

dberenbaum:
@shcheklein Should this be merged into `guide-plots` or into `main`?

shcheklein:
@dberenbaum looks like Restyle is complaining, so this could not be merged into main. Besides that I don't have any strong preference to be honest.

PR #3875, opened 2022-08-12
cmd-ref: update plots templates

dberenbaum:
Updates to `plots templates` for https://github.com/iterative/dvc/pull/8129


pared (approved):

content/docs/user-guide/visualizing-plots.md
```diff
@@ -276,6 +276,8 @@ The `linear` template is the default. It can be changed with the `--template`
 (`-t`) option of `dvc plots show` and `dvc plots diff`. The argument provided to
 `--template` can be a (built-in) template name or a path to a [custom template].
 
+[custom templates]: /doc/command-reference/plots/templates
+
```
jorgeorpinel, line 280: It's already in line 299

```suggestion
```

But I just noticed that line 307 should reuse the ref link (currently it repeats the whole path).

dberenbaum: Actually, the ref above should be `custom template` (without the `s`) and is broken now. Not sure if there's a way to consolidate here if they all need different text to show. Otherwise I can just fix the link above.

jorgeorpinel: Yes we should consolidate...

content/docs/user-guide/visualizing-plots.md
```diff
@@ -276,6 +276,8 @@ The `linear` template is the default. It can be changed with the `--template`
 (`-t`) option of `dvc plots show` and `dvc plots diff`. The argument provided to
 `--template` can be a (built-in) template name or a path to a [custom template].
 
+[custom templates]: /doc/command-reference/plots/templates
```
jorgeorpinel, line 279: > Yes we should consolidate...

Like this

```suggestion
`--template` can be a (built-in) template name or a path to a
[custom template][plots-templates].

[plots-templates]: /doc/command-reference/plots/templates
```

jorgeorpinel: And reuse in line 307 too, ideally.

jorgeorpinel: p.s. it's also a bit strange that the ref is in line 299, interrupting a bullet list.

dberenbaum:
Sorry @jorgeorpinel, I think this is outdated after #3902. Closing.

dberenbaum:
😓 Sorry, too many plots PRs between this,  #3902, and #3903. Reopening.

jorgeorpinel:
Not sure what happened here but it seems like you managed @dberenbaum . Thanks!

PR #3878, opened 2022-08-16
Use internal-use tag in link-check

rogermparent:
# https://github.com/iterative/dvc.org/pull/3878 = https://github.com/iterative/mlem.ai/pull/155 = https://github.com/iterative/cml.dev/pull/300 = https://github.com/iterative/iterative.ai/pull/542

These PRs change the version of link check to be pinned to a git tag on the link check repo called `internal-use`, which currently points to https://github.com/iterative/link-check/pull/22, making the default check only show failing links instead of showing all links including passing ones.

This `internal-use` git tag can be moved to different versions and updated independently of the default branch, which could be useful for temporarily testing features on our sites with the link checker.

### DIFF
--- a/.github/workflows/link-check-all.yml
+++ b/.github/workflows/link-check-all.yml
@@ -5,4 +5,4 @@ on:
     - cron: '0 0 * * *'
 jobs:
   run:
-    uses: iterative/link-check/.github/workflows/link-check-all.yml@v0.13.0
+    uses: iterative/link-check/.github/workflows/link-check-all.yml@internal-use
--- a/.github/workflows/link-check-deploy.yml
+++ b/.github/workflows/link-check-deploy.yml
@@ -3,4 +3,4 @@ on:
   deployment_status:
 jobs:
   run:
-    uses: iterative/link-check/.github/workflows/link-check-deployment-status.yml@v0.13.0
+    uses: iterative/link-check/.github/workflows/link-check-deployment-status.yml@internal-use


yathomasi:
> This `internal-use` git tag can be moved to different versions and updated independently of the default branch, which could be useful for temporarily testing features on our sites with the link checker.

Yes, this seems a good idea for temporarily testing. For the production use case, shouldn't we follow the version approach as it will be easier in long run for maintainability?

rogermparent:
> For the production use case, shouldn't we follow the version approach as it will be easier in long run for maintainability?

I wouldn't say "easier", using individual versions over a tag that updates all at once sacrifices ease of updating to ensure an update doesn't break anything. I think in the case of link check being a CI library that we share across many repos, the cost/benefit analysis moves more toward having one mechanism that updates the workflows in all our repos.

Alternatively, now that I think about it, we could leave link check upgrade PRs to tools like renovate if we can get it set up to do so.

yathomasi:
Thanks for putting it clearly. Yes, the point was about possible breakability and version approach being better to handle those. Definitely, we should focus on automating workflow updates on all our repos, and using tools like renovate seems a better approach. 
It's just to me moving tag approach is almost the same task as the version approach and it seems to be best for temporarily testing.

rogermparent:
I'll shelve this and start introducing renovate, probably with some special config to grab link-check updates ASAP

PR #3715, opened 2022-07-01
ref: document `dvc queue`

pmrowla:
Initial command reference changes for `dvc queue` and `dvc exp`
Related to https://github.com/iterative/dvc.org/issues/3658

TODO:
- [x] `dvc exp run` changes/deprecations
- [x] `dvc queue` examples

PR #3894, opened 2022-08-25
`queue`: edits to new docs

jorgeorpinel:
Follow up to https://github.com/iterative/dvc.org/pull/3715#issuecomment-1203509146


content/docs/user-guide/experiment-management/cleaning-experiments.md
```diff
@@ -210,16 +210,16 @@ Removed experiments: exp-bb09c
 
 ## Removing queued experiments
 
-When you've created experiments to be run in the queue with
-`dvc exp run --queue` and later decide not to run them, you can remove them with
-`dvc exp remove --queue`.
+When you've created experiments to be run in the queue with `dvc queue start`
+and later decide not to run them, you can remove them with
+`dvc queue remove --queued`.
```
jorgeorpinel, line 215: @pmrowla should we make this update in other places? The docs still mention `exp run --queue` a lot.

content/docs/command-reference/queue/start.md
```diff
@@ -12,39 +13,25 @@ usage: dvc queue start [-h] [-q | -v] [-j <number>]
 
 ## Description
 
-Starts one or more task queue worker processes. Each worker process will consume
-and execute one queued experiment task at a time in the background, until either
-`dvc queue stop` is used or the queue is empty.
+Starts one or more workers (`--jobs`) to process experiments. Each worker will
+consume and execute one queued experiment tasks at a time in the background,
+until either `dvc queue stop` is used or the queue is empty.
```
jorgeorpinel, line 18: - [ ] For a possible future follow up: review terms "task queue" and "worker process" everywhere. They're probably unnecessary IMO.

content/docs/command-reference/queue/remove.md
```diff
@@ -17,15 +17,14 @@ positional arguments:
 
 ## Description
 
-Removes the specified queued or completed experiment tasks from the queue. For
-completed tasks, this will also remove any associated output logs.
+Removes the specified queued or completed experiment `task`(s) from the queue.
+For completed tasks, this will also remove any associated output logs.
```
jorgeorpinel, line 21: Do we need to specify "queued or completed"? What other states are there? Only running, I suppose? If so, could phrase like "removes queued exp... Doesn't work with running experiments -- see `queue kill`".

content/docs/user-guide/experiment-management/cleaning-experiments.md
pmrowla: The issue here was that `dvc exp run --queue` was explaining the way the user "created experiments with ..." but `dvc queue start` is the way the user would "run experiments with ...".

So the original intention was more like

> When you've created experiments with `dvc exp run --queue` and later decide not to run them, ...

I guess the question is whether or not we can just say "added experiments to the queue" or "queued experiments" instead of providing the commands in these parts of the user guide?

content/docs/command-reference/queue/remove.md
pmrowla: There will be other possible states after https://github.com/iterative/dvc/pull/8158

The previous definition of "running" was just "not queued and not completed", but after that PR, `RUNNING` state in DVC UI specifically means "executing pipeline". And there are other new states to represent setup and cleanup of the experiment.

So for the docs purposes, I think "queued or completed" is still the correct description here. It may also now be better to use something like "active" as the generic "not queued or completed" term instead of "running" (when it comes to things like "Doesn't work with running experiments")

dberenbaum: There are also failed experiments that probably should be added to the types of tasks that can be removed.

Active/inactive are new terms but seem intuitive. I'm fine with whatever you both decide.

content/docs/user-guide/experiment-management/cleaning-experiments.md
dberenbaum, line 215: I thought we agreed that `dvc exp remove --queue` will stay and we will generally recommend `exp` over low-level `queue` commands? Not that important, just surprised to see this suggestion.

content/docs/user-guide/experiment-management/cleaning-experiments.md
```diff
@@ -210,16 +210,16 @@ Removed experiments: exp-bb09c
...
 
 ```dvc
 $ dvc exp run --queue -S param=10
 Queued experiment '7b83744' for future execution.
 $ dvc exp run --queue -S param=20
 Queued experiment '68808d5' for future execution.
-$ dvc exp show
+$ dvc queue start
```
dberenbaum, line 222: Confused by this one. I assume it's a typo and shouldn't be changed?

dberenbaum:
Thanks @jorgeorpinel! Would you mind taking a look at my few comments and then hopefully we can merge and call this done?

content/docs/user-guide/experiment-management/cleaning-experiments.md
jorgeorpinel: > the question is whether or not we can just say "added experiments to the queue" or "queued experiments" instead of providing the commands

Great Q @pmrowla. I think that in the User Guide (as is the case), we should indeed use descriptive phrasing more, but link to the corresponding sections in the same or other guides.

> However, in Refs it's better to link to other commands so it stays self-contained (linking to guides additionally in admonitions is also good, for those who want/need more reading).

jorgeorpinel: I rolled this back and rephrased a little (675947351) to preserve the original intention (explained by @pmrowla in https://github.com/iterative/dvc.org/pull/3894#discussion_r954598608). PTAL

@dberenbaum But it wasn't that clear to me that a) we won't recommend `queue start` over `exp run --all` at least until new aliases are decided; and b) why was `exp remove --queue` changed to `queue remove --queued` [in the original PR](https://github.com/iterative/dvc.org/pull/3715/files#diff-213c105f41cfb7e3abe29cba4fd80d9cf459bb94703b0a49708b23cf6b6cf5c5L171-R172) then? (Should we roll that back?)

content/docs/user-guide/experiment-management/cleaning-experiments.md
jorgeorpinel: Oops. Rolled back.

content/docs/command-reference/queue/remove.md
dberenbaum: > So again, maybe better to phrase like "Removes non-active experiments from the queue. See `queue kill` to interrupt active ones" @dberenbaum ?

Sounds good, thanks!

content/docs/user-guide/experiment-management/cleaning-experiments.md
```diff
@@ -210,9 +210,8 @@ Removed experiments: exp-bb09c
 
 ## Removing queued experiments
 
-When you've created experiments to be run in the queue with
-`dvc exp run --queue` and later decide not to run them, you can remove them with
-`dvc exp remove --queue`.
+When you've queued experiments with `dvc exp run --queue` and later decide not
+to run them, you can remove them with `dvc exp remove --queue`.
```
dberenbaum, line 214: Maybe mention both `dvc exp remove` and `dvc queue remove` or keep it generic as suggested above. However, whatever we decide, it should be consistent with the examples below (see line 236 below).

content/docs/command-reference/queue/remove.md
jorgeorpinel: Updated in afd08c33f.

> explicitly document the states, maybe grouped by active vs. not

This is pending. Feel free to make an issue/task/note somewhere. Thanks

content/docs/user-guide/experiment-management/cleaning-experiments.md
jorgeorpinel: Rolled back line 236.

> keep it generic as suggested above

That seems like a bigger task that applies to many places probably. I'll leave it up to you to create an issue, etc. if needed!

dberenbaum (approved):
Thanks @jorgeorpinel!

Issue #2486, opened 2021-05-18
regular: fix expired and broken links


shcheklein:
See https://github.com/iterative/dvc.org/actions/workflows/link-check-all.yml

```
Run iterative/link-check.action@v0.7
* content/blog/2020-07-16-devops-for-data-scientists.md
  - http://engineering.microsoft.com/devops/ (404)

* content/blog/2020-11-11-november-20-dvc-heartbeat.md
  - https://torontomachinelearning.com/ (409)
  - https://torontomachinelearning.com/ (409)

* content/blog/2020-12-30-december-20-community-gems.md
  - https://github.com/iterative/cml/blob/master/docker/Dockerfile (404)

* content/blog/2021-02-22-cml-runner-prerelease.md
  - https://github.com/iterative/cml/blob/master/docker/Dockerfile (404)

* content/blog/2021-04-16-april-21-dvc-heartbeat.md
  - https://weworkremotely.com/remote-jobs/iterative-senior-frontend-engineer (404)

* content/docs/cml/self-hosted-runners.md
  - https://github.com/iterative/cml/blob/master/docker/Dockerfile (404)

* content/docs/cml/usage.md
  - https://github.com/iterative/cml/blob/master/docker/Dockerfile (404)
```

jorgeorpinel:
Seems like the most important ones (last 2) are about CML cc @casperdcl 

Do we want to maintain old blogs like this though? I guess these are not so old so why not fix them indeed...

julieg18:
The link-check is showing some more broken doc links (Went ahead and tested them to confirm they weren't false negatives):

```
* content/docs/command-reference/status.md
  - /doc/command-reference/reproduce = https://dvc.org/doc/command-reference/reproduce (404)

* content/docs/dvclive/dvclive-with-dvc.md
  - /doc/dvclive/api-reference/get_step = https://dvc.org/doc/dvclive/api-reference/get_step (404)
 
* content/docs/dvclive/ml-frameworks/mmcv.md
  - https://github.com/iterative/dvclive/blob/master/dvclive/mmcv.py (404)

* content/docs/dvclive/ml-frameworks/pytorch-lightning.md
  - https://pytorch-lightning.readthedocs.io/en/latest/common/weights_loading.html#automatic-saving (404)
```
cc @iterative/docs

jorgeorpinel:
Also, @daavoo what about this broken link?

```
* content/docs/dvclive/ml-frameworks/mmcv.md
[...]

PR #3895, opened 2022-08-25
Update mmcv.md

daavoo:
Fix broken link, per https://github.com/iterative/dvc.org/issues/2486#issuecomment-1226624396


jorgeorpinel (approved):
Gracias @daavoo

PR #3918, opened 2022-09-02
guide: apply merge driver to all .dvc files

dberenbaum:
This is minor but when looking into https://github.com/iterative/dvc/issues/8233, I noticed that we recommend adding `mydataset.dvc` to `.gitattributes`. Can we recommend `*.dvc` instead and save the user from having to understand Git wildcards? Is there a risk to doing so?


jorgeorpinel (approved):
I assume this is OK. Thanks @dberenbaum

PR #3920, opened 2022-09-05
Use reusable package for server

yathomasi:
Published server package at: https://www.npmjs.com/package/@dvcorg/websites-server

Replaced existing server code with reusable server package.


src/gatsby/hooks/stars.ts
```diff
@@ -18,7 +18,7 @@ export default function useStars(): number | null {
   // Run an IIFE to update from the server on the client side.
   useEffect(() => {
     ;(async (): Promise<void> => {
-      const res = await fetch(`/api/github/stars`)
+      const res = await fetch(`/api/github/stars?repo=dvc`)
```
rogermparent, line 21: Maybe we should be able to specify a default repo within the site's server definition, and if we allow multiple repos because a site needs it we should restrict it to just those repos to minimize the possibility of unintended api usage.

rogermparent (approved):
Sorry for the late review, this is looking good! I pulled the PR down and tested it locally, everything seems to run well! I have a small nitpick with the API, but other than that I think we're good to start using this since it'll make server improvements across all websites much easier.

src/gatsby/hooks/stars.ts
yathomasi: Right now we query for single repo and by default, it's `dvc` on the server side i.e for this particular case it's okay if we don't provide the query `repo=dvc`. 
If we want to query for multiple repos, we can modify the logic but I don't think we don't have such a use case so we should be fine.

Also, it would be nice to set the default repo on server initialization and pass it to the server application if we have repeated use cases. So, again I feel it's okay as it is unless I am missing something.

rogermparent: The repeated use cases are each site with stars, right? dvc.org, mlem.ai, and cml.dev have stars, and will need to set this parameter. I'm sure we can define this in `src/server.js` considering each website has its own anyway, and then we're exposing the least amount of data necessary on these APIs for seemingly no benefit.

We can and should definitely ignore the prospect of multiple star sources, but I bring it up because that's one of the only reasons to specify the repo in a query parameter instead of internally.

PR #3931, opened 2022-09-09
Update DocSearch

yathomasi:
Related https://github.com/iterative/cml.dev/pull/320


rogermparent (approved):
We'll want to update the theme so we can use less copy-and-paste code in the consumer repos here, but this works in the meantime.

yathomasi:
> We'll want to update the theme so we can use less copy-and-paste code in the consumer repos here, but this works in the meantime.

The simplest solution would be using the environmental variables to obtain the key values. We already have an [issue](https://github.com/iterative/gatsby-theme-iterative/issues/27) open for it, we just need to bump the priority.

Issue #3927, opened 2022-09-07
Document dvcfs API


dberenbaum:
### Report

[Dvcfs](https://github.com/iterative/dvc/tree/main/dvc/fs) can be a useful public api that is more flexible than the methods in the current dvc api. Some use cases for it:

* It can support walking through a repo or directory and getting the url or opening the files inside.
* It can also be used to pass arbitrary credentials or other configuration to connect to remote storage.

jorgeorpinel:
Have we considered simplifying API refs and moving the details to auto-generated sites like https://docs.iterative.ai/dvc-task/reference/dvc_task/ ?

skshetry:
@jorgeorpinel, that’s a “Developer docs”, not a user facing one. It does not make sense to me to redirect to a different site for what is part of dvc’s offering.

But I am not against auto-generation if the _engine_ can support.

shcheklein:
We had that discussion for `dvc.api` and decided against it (primarily bc team wanted to keep docstrings really simple - e.g. no full examples like we have, etc, etc). Do we plan in this case bring those into docstring?

skshetry:
We do have docstrings now in `dvc.api` thanks to @daavoo. Docstrings are useful when you are working in IDEs/Editors. Personally, I wouldn't want to leave my text-editor/REPL to docs as much as possible because it involves context switching from the task I have in hand. So I prefer docstrings.

But I also think the docs have a different responsibility compared to docstrings, such as introducing the dvcfs to new users, suggesting scenarios that it can solve, etc that @dberenbaum mentioned above. That said, we also need an API reference too in the docs (which could be generated from the docstrings).

shcheklein:
Yes, I was talking specifically about API reference. If you feel that docstring that we have (and plan to have for dvcfs) cover enough of a doc like this https://dvc.org/doc/api-reference/get_url , we should def auto generate. Since docstrings are 100% needed (I would never want to jump into browser also).

jorgeorpinel:
> > simplifying API refs and moving the details to auto-generated sites like https://docs.iterative.ai/dvc-task/reference/dvc_task/ ?
>
> that’s a “Developer docs”, not a user facing one

Not sure it's a meaningful distinction @skshetry: The user is a "developer" here (API ref).

[...]

PR #3932, opened 2022-09-09
dvcfs: initial docs

skshetry:
Fixes partially https://github.com/iterative/dvc.org/issues/3927.

First draft of `dvcfs`.


content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+```py
+>>> fs = DvcFileSystem("https://github.com/iterative/example-get-started.git", rev="main")
+>>> fs.get("data", "data", recursive=True)
+```
+
+This downloads all the files in "data" directory - be it Git-tracked or
+DVC-tracked into a local directory "data". Similarly, DVC might fetch files from
+remote if they don't exist in the cache.
+
+## API Reference
```
skshetry, line 61: I am not really seeing much value from "API reference" right now compared to Scenarios/Recipes that we can cover.

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+
+```py
+>>> from dvc.fs.dvc import _DvcFileSystem as DvcFileSystem
+# opening a local repository
+>>> fs = DvcFileSystem("/path/to/local/repository")
+# opening a remote repository
+>>> remote_fs = DvcFileSystem("https://github.com/iterative/example-get-started.git", rev="main")
+```
+
+## Examples
```
skshetry, line 23: Instead of starting with the verbose docs, I am going with "Show me the code" approach.

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+
+DvcFileSystem provides a single view of all the files/directories in your
+repository, be it Git-tracked or DVC-tracked, or untracked (in case of a local
+repository). DvcFileSystem is smart to reuse the files in your cache directory
+(if present) and can otherwise stream/fetch the files from your default remote.
+
+### Basic Usage
+
+```py
+>>> from dvc.fs.dvc import _DvcFileSystem as DvcFileSystem
```
skshetry, line 16: This is temporary. Eventually, this will look something like:
```suggestion
>>> from dvc.api import DvcFileSystem
```

Or, better:
```suggestion
>>> from dvc.fs import DvcFileSystem
```

skshetry:
@dberenbaum, I'd say yes. This is a bare-bones version, as I wanted to start simple. And there are open questions regarding docs organization and API reference.

I think in these docs, I am going to approach this with more code examples/recipes instead of API references.

Here's the PR: https://github.com/iterative/dvc/pull/8277 that this is based on (I made some changes to the constructor of the filesystem).

shcheklein (commented):
I like the approach. I also don't see a benefit of repeating the whole fsspec (even though it doesn't have proper docs at all). We might need to iterate of the text on the first two paragraphs and merge it.

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
+# DvcFileSystem
+
+DvcFileSystem provides a pythonic file interface (aka
```
shcheklein, line 3: aka -> fsspec-compatible

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+[fsspec](https://filesystem-spec.readthedocs.io/)) for a DVC repo and provides a
```
shcheklein, line 4: It implements read-only operations like ...

shcheklein: (too many "provides")

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+read-only filesystem-like operations like `ls`, `du`, `glob`, `get`, `download`,
+etc.
+
+DvcFileSystem provides a single view of all the files/directories in your
+repository, be it Git-tracked or DVC-tracked, or untracked (in case of a local
+repository). DvcFileSystem is smart to reuse the files in your cache directory
```
shcheklein, line 10: Internally it avoids downloading DVC-tracked files unless it's needed. It's using DVC cache [make it abbr] to avoid downloading objects every time and support streaming for supported remote storages [link] for operations like ...

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+(if present) and can otherwise stream/fetch the files from your default remote.
+
+### Basic Usage
```
shcheklein, line 13: do we need this subtitle?

content/docs/api-reference/dvcfs.md
shcheklein, line 23: let's remove this subtitle also and just go use case by use case? (if we don't have that much besides examples, we don't need this structure + we can make it everything second level, so that it's easier to navigate), wdyt?

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+>>> from dvc.fs.dvc import _DvcFileSystem as DvcFileSystem
+# opening a local repository
+>>> fs = DvcFileSystem("/path/to/local/repository")
+# opening a remote repository
+>>> remote_fs = DvcFileSystem("https://github.com/iterative/example-get-started.git", rev="main")
+```
+
+## Examples
+
+### Listing all DVC-tracked files
```
shcheklein, line 25: suggestion: make title look like command/calls - `ls`: list DVC-tracked files

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+
+```py
+>>> fs = DvcFileSystem("https://github.com/iterative/example-get-started.git", rev="main")
```
shcheklein, line 28: may be worth mentioning in those examples important aspects of their behavior - e.g. it doesn't download files, streams, etc. Can be brief

content/docs/api-reference/dvcfs.md
dberenbaum: When do we expect this change?

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+repository, be it Git-tracked or DVC-tracked, or untracked (in case of a local
+repository). DvcFileSystem is smart to reuse the files in your cache directory
+(if present) and can otherwise stream/fetch the files from your default remote.
+
+### Basic Usage
+
+```py
+>>> from dvc.fs.dvc import _DvcFileSystem as DvcFileSystem
+# opening a local repository
+>>> fs = DvcFileSystem("/path/to/local/repository")
```
dberenbaum, line 18: Does this represent the current state of the repo? It looks like `_DvcFileSystem.__init__()` expects a `Repo` instance, and it doesn't seem to work for me with a `str` arg:

```python
>>> fs = DvcFileSystem(".")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dave/miniforge3/envs/dvc/lib/python3.10/site-packages/fsspec/spec.py", line 76, in __call__
    obj = super().__call__(*args, **kwargs)
  File "/Users/dave/Code/dvc/dvc/fs/dvc.py", line 118, in __init__
    self.hash_jobs = repo.fs.hash_jobs
AttributeError: 'str' object has no attribute 'fs'
```

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+This downloads all the files in "data" directory - be it Git-tracked or
+DVC-tracked into a local directory "data". Similarly, DVC might fetch files from
+remote if they don't exist in the cache.
+
+## API Reference
+
+As DvcFileSystem is based on [fsspec](https://filesystem-spec.readthedocs.io/),
+it is compatible with most of the APIs that it offers. Please check the fsspec's
```
dberenbaum, line 64:
```suggestion
it is compatible with most of the read-only methods that it offers. Please check the fsspec's
```

content/docs/api-reference/dvcfs.md
dberenbaum, line 3: Maybe it can be further simplified to stress the main points that it's read-only and fsspec comapitable. I think it should start with something like, "DVCFileSystem is a read-only [fsspec](https://filesystem-spec.readthedocs.io/)-compatible interface." Then you could go on to describe the typical operations and everything else.

content/docs/api-reference/dvcfs.md
skshetry: See https://github.com/iterative/dvc.org/pull/3932#issuecomment-1242232523.

content/docs/api-reference/dvcfs.md
skshetry: This week. It’s not difficult, we just need to decide.

I find `dvc.fs` namespace more natural, but it’s controversial compared to `dvc.api`.
Let me know if you have any opinions/preference.

content/docs/api-reference/dvcfs.md
skshetry: Initially I wrote like that, but I later thought that the end users does not care about fsspec or need to care about it to use it.

daavoo: The `read-only` part is important

content/docs/api-reference/dvcfs.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Basic Usage
+
+```py
+>>> from dvc.fs.dvc import _DvcFileSystem as DvcFileSystem
+# opening a local repository
+>>> fs = DvcFileSystem("/path/to/local/repository")
+# opening a remote repository
+>>> remote_fs = DvcFileSystem("https://github.com/iterative/example-get-started.git", rev="main")
+```
```
daavoo, line 21:
```suggestion
>>> remote_fs = DvcFileSystem(
    "https://github.com/iterative/example-get-started.git", 
    rev="main")
```
```
The `rev` arg gets cropped in the block and it's not visible without scrolling. I think its important to show it given that it's not present in the fsspec API

File: content/docs/api-reference/dvcfs.md (hunk 3 above)
Comment (dberenbaum) [reply content/docs/api-reference/dvcfs.md:16]: No strong preference as long as it's not `_DvcFileSystem`!

File: content/docs/api-reference/dvcfs.md (hunk 4 above)
Comment (dberenbaum) [reply content/docs/api-reference/dvcfs.md:3]: > Initially I wrote like that, but I later thought that the end users does not care about fsspec or need to care about it to use it.

Good point, but if we are not documenting the available methods, we need to point out clearly that it implements the read-only methods available in fsspec.

Comment (dberenbaum) [reply content/docs/api-reference/dvcfs.md:3]: If you want to first focus on it being a filesytem-like interface and read-only, you could say something like: "DvcFileSystem provides read-only operations like `ls`, `du`, `glob`, `get`, `download`, etc. for a DVC repo. The API implements most read-only methods of [fsspec](https://filesystem-spec.readthedocs.io/)."

File: content/docs/sidebar.json [hunk 13]
```diff
@@ -520,6 +520,10 @@
     "label": "Python API Reference",
     "source": "api-reference/index.md",
     "children": [
+      {
+        "slug": "dvcfs",
+        "label": "DvcFileSystem"
```
jorgeorpinel, line 525: Not seeing `dvcfs` as an actual identifier in docs or code. Maybe
```suggestion
        "slug": "dvcfilesystem",
        "label": "DvcFileSystem"
```
Or
```suggestion
        "slug": "dvc.fs",
        "label": "DvcFileSystem"
```

content/docs/api-reference/dvcfs.md
jorgeorpinel: Goes back to https://github.com/iterative/dvc.org/issues/3927#issuecomment-1245044754. If we have the API ref auto-generated then I agree. Or do you suggest skipping it completely and let people/IDEs check docstrings?

content/docs/api-reference/dvcfs.md
jorgeorpinel: For now reusing the familiar `## Example: {Title}` pattern would be a first improvement (visible in the right-hand ToC).

content/docs/api-reference/dvcfs.md
jorgeorpinel: Wouldn't this be capitalized `DVCFileSystem`?

content/docs/api-reference/dvcfs.md
jorgeorpinel: > the end users does not care about fsspec or need to care about it to use it.

So why do we even mention it?

jorgeorpinel:
Many comments already so I'll just summarize some thoughts here:

- Nothing against the example-based approach but
  - It's completely different to the current functions we have listed; Should we group those under `dvc.api` in the nav and use `dvc.fs` for this one?
  - Are the examples exhaustive or comprehensive? (How do we decide what to cover?)

- In general the exact relation with `fsspec` is currently vague. Is it literally `fsspec`? The link at the bottom goes to `fsspec.spec.AbstractFileSystem`, is that what we implement? This link seems a better intro doc on their side BTW: https://filesystem-spec.readthedocs.io/en/latest/usage.html#use-a-file-system

content/docs/api-reference/dvcfs.md
skshetry: It is fsspec-based filesystem. We do need to mention it, I just don't think we should lead with that in the intro. 🙂

jorgeorpinel: Sounds good. Please see https://github.com/iterative/dvc.org/pull/3932#issuecomment-1246342519 below. Thanks

content/docs/api-reference/dvcfs.md
shcheklein: fsspec is supposed to be a documented API that this library implements, we don't want to repeat everything here. Also, it makes sense to start small, with examples. I hope a lot of things here will be self descriptive through examples.

content/docs/api-reference/dvcfs.md
shcheklein: yep, probably should be ... for consistency with DVCLive

skshetry: In camel casing, the abbreviations are usually treated the same as lowercase words.

From [Wikipedia/Camel_case#Programming_and_coding](https://en.wikipedia.org/wiki/Camel_case#Programming_and_coding):
> Programming identifiers often need to contain [acronyms and initialisms](https://en.wikipedia.org/wiki/Acronyms_and_initialisms) that are already in uppercase, such as "old HTML file". By analogy with the title case rules, the natural camel case rendering would have the abbreviation all in uppercase, namely "oldHTMLFile". However, this approach is problematic when two acronyms occur together (e.g., "parse DBM XML" would become "parseDBMXML") or when the standard mandates lower camel case but the name begins with an abbreviation (e.g. "SQL server" would become "sQLServer"). For this reason, some programmers prefer to treat abbreviations as if they were lowercase words and write "oldHtmlFile", "parseDbmXml" or "sqlServer".

Also, [Google Javascript Style Guide](https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined):
> lowercase everything (including acronyms), then uppercase only the first character of:
> a) … each word, to yield upper camel case, or
> b) … each word except the first, to yield lower camel case

skshetry:
>   * It's completely different to the current functions we have listed; Should we group those under `dvc.api` in the nav and use `dvc.fs` for this one?

It's now under `dvc.api`.

>   * Are the examples exhaustive or comprehensive? (How do we decide what to cover?)

I don't think we have to be strict about it, we can always update it and iterate.

> * In general the exact relation with `fsspec` is currently vague. Is it literally `fsspec`? The link at the bottom goes to `fsspec.spec.AbstractFileSystem`, is that what we implement?

yes, `fsspec` provides a standard interface, "dvcfs" implements it.

content/docs/api-reference/dvcfs.md
shcheklein: Good points, @skshetry how do you feel about consistency though? (DVCLive). I bothers me a bit. Are there popular examples in Python world with all upper acronyms?

skshetry: MLFlow?

skshetry: [`socketserver.TCPServer`](https://docs.python.org/3/library/socketserver.html#socketserver.TCPServer) and [`http.HTTPServer`](https://docs.python.org/3/library/http.server.html#module-http.server) comes to mind from Python.

skshetry: Also urllib's HTTPResponse (Django uses `HttpResponse`). Also `json.JSONDecoder`. But `HTTP`/`JSON`/`TCP` are common acronyms.

skshetry: [PEP8](https://peps.python.org/pep-0008/#descriptive-naming-styles):
> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

[Pocoo Styleguide: flask/click, etc.](https://flask.palletsprojects.com/en/1.1.x/styleguide/#naming-conventions):
> Class names: CamelCase, with acronyms kept uppercase (HTTPWriter and not HttpWriter)

content/docs/api-reference/dvcfs.md
skshetry: It's not the API that is read-only, but the filesystem and it's operation in general. I have also myself struggled to explain this, but tbh I don't think we need to mention anything other than "... is a read-only filesystem".

content/docs/api-reference/dvcfs.md
skshetry: Thanks, I think I'll split them into two lines rather than indenting it.

content/docs/api-reference/dvcfs.md
skshetry: Like @daavoo said, it's equivalent to `ls --recursive`, so I don't want to force commands in the title. I do mention that after the snippets now. :)

content/docs/api-reference/dvcfs.md
skshetry: I think it's better to even skip it and raise a `EROFS`.

content/docs/api-reference/dvcfs.md
daavoo: Afaik, all fsspec implementations (which is Python) are upper case when using acronyms: `SSHFileSystem`, `GCSFileSystem`,  `IPFSFileSystem`, `OSSFileSystem`, etc

daavoo: > Are there popular examples in Python world with all upper acronyms?

scikit-learn has tons of examples using acronyms with all caps: `PLSCanonical`, `KNNImputer`, `RBFSampler`, `SGDClassifier`, `RANSACRegressor`, etc.

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
+# DvcFileSystem
+
+_New in DVC 2.26_
```
skshetry, line 3:
```suggestion
_New in DVC 2.27_
```

content/docs/api-reference/dvcfs.md
dberenbaum: What is an `EROFS`?

I must be missing your point here because I don't see what's confusing or ambiguous. Is dvcfs compatible with the read-only methods of https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem?

skshetry: EROFS is an errno which is raised for "Read Only FileSystem". What I am trying to say is that the filesystem in general is read-only, not the methods. Take `fs.open()` for example. It works with `mode="rb"` but `mode="wb"` raises `EROFS`. So I am trying to distinguish between the method `open` vs the operations: writing/reading, etc.

skshetry:
@jorgeorpinel @shcheklein, I am leaving for a vacation next week, so it'd be great if you could review them, I can iterate on these docs this week.

shcheklein:
@skshetry I see that @dberenbaum has approved it. Are there unresolved things / any blockers / you need help with something?

skshetry:
I can always get more feedback. 🙂  Is it fine to merge this then?

shcheklein:
Yes, I think so, one approval should be enough.

What do you think about keeping the ticket open like Dave (and I guess Jorge suggested)?

skshetry:
> What do you think about keeping the ticket open like Dave (and I guess Jorge suggested)?

Done, thanks.

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
+# DVCFileSystem
+
+_New in DVC 2.27_
+
+DVCFileSystem provides a pythonic file interface (
+[fsspec-compatible](https://filesystem-spec.readthedocs.io/)) for a DVC repo. It
+is a read-only filesystem, hence it does not support any write operations, like
+`put_file`, `cp`, `rm`, `mv`, `mkdir` etc.
```
jorgeorpinel, line 8: 💅🏼 
```suggestion
`DVCFileSystem` is a Python class that provides a file interface for <abbr>DVC repositories</abbr>
(compatible with [fsspec](https://filesystem-spec.readthedocs.io/)). It
is read-only, hence it does not support any write operations like
`put_file`, `cp`, `rm`, `mv`, `mkdir` etc.
```

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
...
+
+DVCFileSystem provides a unified view of all the files/directories in your
+repository, be it Git-tracked or DVC-tracked, or untracked (in case of a local
+repository). It can reuse the files in DVC <abbr>cache</abbr> and can otherwise
+stream from
+[supported remote storage](/doc/command-reference/remote/add#supported-storage-types).
```
jorgeorpinel, line 14:
```suggestion
`DVCFileSystem` objects construct a unified view of all the files and directories in your
repo, be it Git-tracked, DVC-tracked, or untracked (in the case of local
repos). It can reuse files in the <abbr>DVC cache</abbr>, and it can otherwise
stream from [supported remote storage].

[supported remote storage]: /doc/command-reference/remote/add#supported-storage-types
```

> It can reuse files in the DVC cache

What does this mean though? Reuse them for what purpose?

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
...
+
+```py
+>>> from dvc.api import DVCFileSystem
+# opening a local repository
+>>> fs = DVCFileSystem("/path/to/local/repository")
+# opening a remote repository
+>>> url = "https://github.com/iterative/example-get-started.git"
+>>> fs = DVCFileSystem(url, rev="main")
```
jorgeorpinel, line 22: 💅🏼 
````suggestion
```py
from dvc.api import DVCFileSystem
# Opening a local repository
fs = DVCFileSystem("/path/to/local/repository")
# Opening a remote repository
url = "https://github.com/iterative/example-get-started.git"
fs = DVCFileSystem(url, rev="main")
````

File: content/docs/api-reference/dvcfilesystem.md [hunk 18]
```diff
@@ -0,0 +1,126 @@
...
+```
+
+The optional positional argument can be a URL or a local path to the DVC
+project. If unspecified, the DVC project in current working directory is used.
```
jorgeorpinel, line 26: Does it work with any DVC project or does it have to be a Git repo? (We currently use the term "repo" specifically in a few places.)

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
...
+
+The optional `rev` argument can be passed to open a filesystem from a certain
+Git commit (any [revision](https://git-scm.com/docs/revisions) such as a branch
+or a tag name, a commit hash, or an [experiment name]).
+
+[experiment name]: /doc/command-reference/exp/run#-n
```
jorgeorpinel, line 32: Would it be better to put the [class signature](https://github.com/iterative/dvc/blob/391bb29be29fe9ac3036180f503590c97b71c850/dvc/fs/dvc.py#L368) up somewhere? With param names, etc. (may need to make a mixed definition combined with [FileSystem](https://github.com/iterative/dvc/blob/570e3a5359b41520ea01952df11602e40d69f793/dvc/objects/fs/base.py#L55))

jorgeorpinel (commented):
Just some suggested follow-ups post-merge, concentrated on the intro (the subsections are pretty good I think 👍🏼 -- may just need double checking grammar, etc.)

content/docs/api-reference/dvcfilesystem.md
```diff
@@ -0,0 +1,126 @@
...
+This downloads all the files in "data" directory - be it Git-tracked or
+DVC-tracked into a local directory "data". Similarly, DVC might fetch files from
+remote if they don't exist in the cache.
+
+## API Reference
+
+As DVCFileSystem is based on [fsspec](https://filesystem-spec.readthedocs.io/),
+it is compatible with most of the APIs that it offers. Please check the fsspec's
+[API Reference](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem)
+for more details.
```
jorgeorpinel, line 126: Last note: we could add the term "methods" or something here, to clarify we mean the reference of the elements of the class/object, not some larger API.

PR #3933, opened 2022-09-09
ref: pull --glob only works on workspace paths

dberenbaum:


content/docs/command-reference/pull.md
```diff
@@ -124,7 +124,8 @@ used to see what files `dvc pull` would download.
 
 - `--glob` - allows pulling files and directories that match the
   [pattern](https://docs.python.org/3/library/glob.html) specified in `targets`.
-  Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`, and `**`
+  Shell style wildcards supported: `*`, `?`, `[seq]`, `[!seq]`, and `**`. Only
+  matches against files present in the workspace.
```
shcheklein, line 128: I'm not sure I understand this tbh :) (I don't know the context where it comes from)

dberenbaum:
@shcheklein PTAL at https://github.com/iterative/dvc/issues/8188. Sorry for not making it easier to find that in the PR description!

jorgeorpinel:
I wonder if we need similar notes in other cmd refs. See https://github.com/iterative/dvc/issues/8188#issuecomment-1244407579

content/docs/command-reference/pull.md
jorgeorpinel: Specifically for pull this seems more like a bug than something to document. What good is it to "pull" something that's already present?

shcheklein:
If I understand it correctly then `--glob` is broken (original intention was to support sparse checkouts). Should we for now disable it? Is there any use to it in this form?

Is it the same if I do `dvc pull --glob data/*.json` (some files within an output that is still not part of the workspace).

Even if decide to do keep it. We should be way more explicit, this explanation was very confusing and it won't solve the problem (of people keep coming and asking about this).

pmrowla:
The `--glob` implementation comes from a user PR that probably should not have been merged, but was deemed good enough at the time since it filled that particular user's use case. IIRC they just wanted to be able to pull/fetch the latest version of specific files they already had in their workspace (because another collaborator had `dvc push`ed changes for those files).

content/docs/command-reference/pull.md
pmrowla: Someone else may have pushed changes for files that I have in my workspace, and I want to pull the latest version of those specific files, but not everything in the entire workspace.

karajan1001:
Context: 

We didn't implement a complete `glob` functionality, and users believed that we did. This is why we are continuously receiving issue reports for this. This PR fixs tells the users the current status of `glob` but sometime later we still make the `glob` work in a more intuitive (complete) way.

shcheklein:
@karajan1001 @dberenbaum  thanks folks, I think (I hope) I finally understand the context 😅 My issue with this PR is that it doesn't help with the `tells the users the current status of glob` (you can proxy it by how much time and explanation it was needed to get me us to speed :) ). Three suggestions (can be implemented at once or one at a time - your call):

- Hide it for now (as I mentioned I see it as broken functionality that would be hard to explain no matter what we do and we'll keep answering these questions)
- Be very explicit. Mention - "Warning: this command doesn't do what you expect, please refer to this ticket ... to see the details and the progress. Summary: when it expands glob spec to get files names to download it analyzes only files that already physically present on the disk at the moment. It can be useful to update files that were already pulled, but won't work for files that are still in the remote storage". + We def need to update the DVC help message, mention that there are limits and unexpected behavior
- Put a warning message in the UI, when you run this command

dberenbaum:
@karajan1001 Have you noticed whether the reports are all/mostly about `pull` or about other commands also? I imagine glob is still useful for other commands even if it seems pretty limited for `pull`.

content/docs/command-reference/pull.md
jorgeorpinel: OK to note this then:
```suggestion
  works with file names that are already in the workspace (pull their latest versions).
```
But is it also a bug or important enhancement to address? If we can `dvc ls` we should be able to match `dvc pull` to any tracked file mentioned in .dvc or dvc.yaml files, IMO.

jorgeorpinel:
> glob is still useful for other commands

Is it? https://github.com/iterative/dvc/issues/8188#issuecomment-1244407579

karajan1001:
> @karajan1001 Have you noticed whether the reports are all/mostly about `pull` or about other commands also? I imagine glob is still useful for other commands even if it seems pretty limited for `pull`.

As mentioned in https://github.com/iterative/dvc/issues/8188#issuecomment-1244407579
Only `pull`, `push`, `add` and `repro` have a `glob` flag, and for the other three commands their target is all locally. 
I don't think `dvc add --glob` would miss any file that didn't exist in the workspace. And for `pull` the target might be missing in the workspace.

jorgeorpinel:
> or the other three commands their target is all local

OK so are we hiding the option instead for now? Up to you @dberenbaum, please consider https://github.com/iterative/dvc.org/pull/3933#discussion_r971700324 if you want to merge this (lmk if you need an approval).

PR #3934, opened 2022-09-09
ref: reorg and wrap `import-url` usage block

jorgeorpinel:
Prevents horizontal scrolling in https://dvc.org/doc/command-reference/import-url#synopsis.

PR #3930, opened 2022-09-08
import/import-url/update: --no-download improvements

dtrifiro:
Follow-up to https://github.com/iterative/dvc.org/pull/3773#pullrequestreview-1098609995

PR #3773, opened 2022-07-18
import-url/update: add --no-download flag

dtrifiro:
Add documentation for new `dvc import-url` `dvc update` flags  (`--no-download`)

related: https://github.com/iterative/dvc/pull/8024 (per iterative/dvc#7918)

PR #3935, opened 2022-09-09
ref: reorg and wrap `import` usage block

jorgeorpinel:
Prevents horizontal scrolling in https://dvc.org/doc/command-reference/import#synopsis.

> Edited from GH so couldn't combine with https://github.com/iterative/dvc.org/pull/3934, sorry.

### DIFF
--- a/content/docs/command-reference/import.md
+++ b/content/docs/command-reference/import.md
@@ -8,9 +8,10 @@ Download a file or directory tracked by another DVC or Git repository into the
 ## Synopsis
 
 ```usage
-usage: dvc import [-h] [-q | -v] [-j <number>]
-                  [-o <path>] [--file <filename>]
-                  [--rev <commit>] [--no-exec | --no-download] [--desc <text>]
+usage: dvc import [-h] [-q | -v] [--file <filename>]
+                  [-o <path>] [--rev <commit>]
+                  [--no-exec | --no-download]
+                  [-j <number>] [--desc <text>]
                   url path
 
 positional arguments:


jorgeorpinel:
Edited from GH so couldn't combine with #3934, sorry.

content/docs/command-reference/import.md
```diff
@@ -8,9 +8,10 @@ Download a file or directory tracked by another DVC or Git repository into the
 ## Synopsis
 
 ```usage
-usage: dvc import [-h] [-q | -v] [-j <number>]
-                  [-o <path>] [--file <filename>]
-                  [--rev <commit>] [--no-exec | --no-download] [--desc <text>]
+usage: dvc import [-h] [-q | -v] [--file <filename>]
```
shcheklein, line 11: if we change order, should we also update in DVC itself?

shcheklein (approved):

content/docs/command-reference/import.md
jorgeorpinel: Yes, that would be ideal cc @dberenbaum

dberenbaum: @jorgeorpinel Is this minor cleanup or is there a larger goal? If it's minor cleanup, would you mind submitting the accompanying core change PR in these cases?

jorgeorpinel: Def. minor. Yes, sorry I know I should just didn't have time but also didn't want to leave it as it is now and it's quick to fix on the docs side. Unfortunately we didn't catch this during which ever PR added flags that made the usage block overflow. I'll add it to the best/std practices (maybe we can make a check list) for writing/updating cmd refs.

dberenbaum: Thanks @jorgeorpinel! Do you plan to open the core PR (no rush, just clarifying)? I think it makes sense to set the precedent that whoever contributes a fix owns propagating the changes to both core and docs repos.

jorgeorpinel: I'll take care of it 🙂

dberenbaum: Thanks @jorgeorpinel!

jorgeorpinel: p.s. I just noticed that we did catch this but post-merge. See https://github.com/iterative/dvc.org/pull/3773#discussion_r964450724. There's a follow-up to that too: https://github.com/iterative/dvc.org/pull/3930 -- reviewing now.

PR #3940, opened 2022-09-13
plots: mention dir support and update templates

dberenbaum:
Minor updates to address https://github.com/iterative/dvcyaml-schema/pull/16#discussion_r968801509 (and an unrelated change for `dvc plots templates`).

I don't see a good place to mention support for `plots outputs` dir support specifically. Since it's a subset of both `top-level plots` and `outputs`, both of which support dirs, I don't know if an explicit mention is needed, but open to suggestions.

### DIFF
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -19,8 +19,9 @@ positional arguments:
                 in a single image.
     modify      Modify display properties of data-series plots
                 defined in stages (has no effect on image plots).
-    templates   Write built-in plots templates to a directory
-                (.dvc/plots by default).
+
+    templates   List built-in plots templates or show JSON
+                specification for one.
 ```
 
 ## Description
--- a/content/docs/command-reference/plots/index.md
+++ b/content/docs/command-reference/plots/index.md
@@ -19,7 +19,6 @@ positional arguments:
                 in a single image.
     modify      Modify display properties of data-series plots
                 defined in stages (has no effect on image plots).
-
     templates   List built-in plots templates or show JSON
                 specification for one.
 ```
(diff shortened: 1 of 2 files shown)


content/docs/command-reference/plots/index.md
```diff
@@ -19,8 +19,9 @@ positional arguments:
                 in a single image.
     modify      Modify display properties of data-series plots
                 defined in stages (has no effect on image plots).
-    templates   Write built-in plots templates to a directory
-                (.dvc/plots by default).
+
```
shcheklein, line 22: minor: do we need an extra space?

shcheklein (approved):

content/docs/command-reference/plots/index.md
jorgeorpinel: I don't think so
```suggestion
```

jorgeorpinel (approved):
Thanks @dberenbaum 

This is great but [the comment](https://github.com/iterative/dvcyaml-schema/pull/16#discussion_r968801509) originally was about stage plots. Do we also need to update that?

jorgeorpinel:
> the comment originally was about stage plots. Do we also need to update that?

I guess it's implied in in https://dvc.org/doc/user-guide/visualizing-plots, but we could probably be more explicit:

image

Maybe a good opportunity to expand on https://dvc.org/doc/command-reference/plots too ⌛

dberenbaum:
@jorgeorpinel See the comment in https://github.com/iterative/dvc.org/pull/3940#issue-1371922926 (I think "stage plots" were renamed to "plots outputs," right?):

> I don't see a good place to mention support for `plots outputs` dir support specifically. Since it's a subset of both `top-level plots` and `outputs`, both of which support dirs, I don't know if an explicit mention is needed, but open to suggestions.

jorgeorpinel:
OK, up to you.

> I think "stage plots" were renamed to "plots outputs," right?

I don't remember this but now I think "stage plot" (or maybe "stage-level") is a better term so it sounds comparable to top-level plots.

Currently docs only mention the term "stage plot" twice though.

_I also noticed some broken links to a `#stage-outputs` anchor so I guess you're right. Fixing that BTW..._

dberenbaum:
TBH I agree and originally used the name "stage plots," but at this point I think we're over-optimizing here.

PR #3941, opened 2022-09-14
start: cleanup nav

jorgeorpinel:
In preparation for https://github.com/iterative/dvc.org/pull/3943

> Ideally approve/merge along with https://github.com/iterative/dvc.org/pull/3942.


content/docs/sidebar.json
```diff
@@ -37,21 +37,12 @@
...
-          {
-            "label": "Data and Model Access",
-            "slug": "access"
-          },
-          {
-            "label": "Data Pipelines",
-            "slug": "pipelines"
-          },
+          "data-and-model-access",
+          "data-pipelines",
```
jorgeorpinel, line 45: - Significantly simplifies this file
- Uses more complete URLs (good for SEO I think)

content/docs/start/index.md
```diff
@@ -69,7 +69,7 @@ in two independent trails:
   be attached to pipelines. These let you capture, navigate, and evaluate ML
   projects without leaving Git. Think "Git for machine learning".
 
-### Experiments Trail
+### Experimentation
 
 - [**Experiments**](/doc/start/experiments) enable exploration, iteration, and
   comparison across many ML experiments. Track your experiments with automatic
```
jorgeorpinel, line 75: Should we call this 'tail' something longer? Maybe "Machine Learning Experimentation" ?

redirects-list.json
```diff
@@ -38,19 +38,19 @@
...
-  "^/doc/start/sharing-data-and-model-files$                                              /doc/start/data-and-model-versioning#storing-and-sharing 302",
+  "^/doc/start/data-versioning$                                                           /doc/start/data-management",
+  "^/doc/start/data-and-model-versioning(/.*)?$                                           /doc/start/data-management 302",
+  "^/doc/start/sharing-data-and-model-files$                                              /doc/start/data-management#storing-and-sharing 302",
+  "^/doc/start/data-access$                                                               /doc/start/data-management/data-and-model-access",
+  "^/doc/start/data-and-model-access(/.*)?$                                               /doc/start/data-management/data-and-model-access 302",
+  "^/doc/start/data-pipelines(/.*)?$                                                      /doc/start/data-management/data-pipelines 302",
+  "^/doc/start/metrics-parameters-plots(/.*)?$                                            /doc/start/data-management/metrics-parameters-plots 302",
```
jorgeorpinel, line 47: Updated and simplified redirects as well (using final locations for previous 301 ones, plus fortunately many were 302s so we can change those any time).

jorgeorpinel:
Also, I made a separate PR to update all the links to keep this PR lean. PTAL: https://github.com/iterative/dvc.org/pull/3942

content/docs/sidebar.json
rogermparent: Removing `source` from things might break the "Edit on GitHub" links unless we update the sidebar logic to better accept them. Might not, but we need to make sure to check.

shcheklein:
Preview link https://dvc-org-start-cleanup-d780abpa.herokuapp.com/doc (not working here for some reason)

shcheklein (changes requested):
Let's not mix many things at once (e.g. data pipelines - we actually go with Pipeline now, right?)

do we need to remove a few index files here?

jorgeorpinel:
> Preview link https://dvc-org-start-cleanup-d780abpa.herokuapp.com/doc (not working here for some reason

Something with Heroku's SSL certs proabbly. You can chose to "visit unsafe site" manually:

image

content/docs/sidebar.json
jorgeorpinel: Hmmm thanks for letting me know @rogermparent. I did not know that (probably something we can improve on in the engine...) I'll reinstate the `source`s for now ⏳

jorgeorpinel: > unless we update the sidebar logic to better accept them

https://github.com/iterative/gatsby-theme-iterative/issues/77

> Might not, but we need to make sure to check

P.s. I did check and the button links were broken indeed (in the review app).

jorgeorpinel:
> Let's not mix many things at once

Sorry @shcheklein, not sure what you mean. This is a nav cleanup with 2 URL changes. Should I make a PR for each one?

> e.g. data pipelines - we actually go with Pipeline now

In the user guide yes but in Get Started [the title is already Data Pipelines](https://dvc.org/doc/start/data-management/pipelines), this PR just fixed the URL so it reflects the title. We can change the title if preferred but that would be a separate change IMO. Let's discuss including that in https://github.com/iterative/dvc.org/pull/3943 ?

jorgeorpinel:
> We can change it, but that would be a separate change IMO.

Or I can return the file to `pipelines.md` and update the title to just **Pipelines** as well. Either route works for me as long as they're consistent. Regardless, it's a change (perhaps larger).

Lmk please!

content/docs/sidebar.json
yathomasi: The button on the review app were broken because the link generated based on `master/main` tree. I checked by changing the path to current branch tree and it's working. 

The logic is intact to link the file source using a simple string slug. And, the `source` field is also getting populated accordingly.
Screen Shot 2022-09-14 at 15 32 52

rogermparent: That would be the case for some of them, but the result of the logic on some nested items, for example `docs/start`, render the link as `https://github.com/iterative/dvc.org/blob/main/content/docs/start.md` where the file is actually at `https://github.com/iterative/dvc.org/blob/main/content/docs/start/index.md`. We can likely just make assume the file is as `index.md` and that'd be all we need to support 90+% of source-less cases.

jorgeorpinel: Well I already rolled this back for now. Feel free to take it to https://github.com/iterative/gatsby-theme-iterative/issues/77 😬

yathomasi: Okay, I checked on the rolled-back version of the preview app. 

Yes, we do need to define the source for items with `index.md`.

shcheklein:
@jorgeorpinel okay, I see I got confused since it also cleanups a lot of stuff along the way ... if this doesn't break any then LGTM ;)

shcheklein (approved):

jorgeorpinel:
I did include other closely related goals (to my ming). 3 in total. It's def. not easy to understand just from the diff which is why I left comments ([1](https://github.com/iterative/dvc.org/pull/3941#pullrequestreview-1106682956), [2](https://github.com/iterative/dvc.org/pull/3941#pullrequestreview-1106683287), [3](https://github.com/iterative/dvc.org/pull/3941#pullrequestreview-1106683991)). Glad it's clearer now. Thanks!

PR #3942, opened 2022-09-14
start: update links from all docs

jorgeorpinel:
- Many of these updates were missed when the last GS update happened, it seems (they're currently hitting redirects).
- Others are updated per https://github.com/iterative/dvc.org/pull/3941.
  - [x] Do not merge this before parent PR ☝🏼


jorgeorpinel:
To reiterate, this is just link updates to avoid redirects (all are passing ☝🏼)

Just needs an approve + merge.

The failing check is wrong (Prettier bug).

shcheklein (approved):

PR #3944, opened 2022-09-14
document labels/type/meta

skshetry:
See https://github.com/iterative/dvc/issues/8214.

As use-cases/scenarios are still unclear, I have attempted here to only document basic stuff. And as mentioned in the above issue, we are not going to document `data ls` for a while. :)


content/docs/command-reference/add.md
```diff
@@ -174,8 +175,13 @@ not.
...
 
-- `--desc <text>` - user description of the data (optional). This doesn't affect
-  any DVC operations.
+- `--desc <text>` - user description of the data.
+
+- `--type <str>` - user assigned type of the data.
+
+- `--label <text>` - user assigned label(s) to add to the data.
+
+- `--meta key=value` - custom metadata to add to the data.
```
jorgeorpinel, line 184: Product questions

- I think they're still optional? Probably worth mentioning
  Maybe having said optional the "user" wording is no longer necessary: I don't think it's clear, as everything in the pipeline is defined by the user.
- Why are some `<text>` and some `<str>`? Distinction unclear (or needs more detail)
- Should we use `<>` around `key=value`?
- "metadata ... to the data" seems redundant, can we rephrase?

Syntax-wise

- Pls hyphenate "user-assigned" if we keep that term.
- No periods `.` needed (not full sentences).

Thanks, I'll leave the rest here to the DVC team as it's mainly a technical reference type contribution! But please @ mention me if needed.

dberenbaum: > * I think they're still optional? Probably worth mentioning

Just for clarification: almost all of the CLI flags are optional, so I questioned why this was needed in the help text of the command. However, agree that we should mention it in the dvc.yaml reference fields where we do otherwise specify what is optional.

content/docs/user-guide/project-structure/dvc-files.md
```diff
@@ -67,6 +67,9 @@ The following subfields may be present under `outs` entries:
...
+| `type`                          | User assigned type of the data.                                                                                                                                                                                                                                                                                                       |
+| `labels`                        | User assigned labels to add to the data.                                                                                                                                                                                                                                                                                              |
+| `meta`                          | Custom metadata about the data.                                                                                                                                                                                                                                                                                                       |
```
dberenbaum, line 72: Here I think `(Optional)` should be mentioned?

content/docs/command-reference/add.md
skshetry: In dvc.yaml too, everything is optional except for `cmd`.

skshetry: In `.dvc` file, everything is optional.

content/docs/user-guide/project-structure/dvcyaml-files.md
```diff
@@ -530,6 +530,9 @@ libraries in the environment.
...
 | `desc`       | (Optional) User description for this output. This doesn't affect any DVC operations.                                                                                                                                                                                              |
+| `type`       | User assigned type of the data.                                                                                                                                                                                                                                                   |
+| `labels`     | User assigned labels to add to the data.                                                                                                                                                                                                                                          |
+| `meta`       | Custom metadata about the data.                                                                                                                                                                                                                                                   |
```
dberenbaum, line 535: `(Optional)` makes sense here, too.

content/docs/command-reference/add.md
dberenbaum: Good point. I saw it was mentioned for other fields so was aiming for consistency. Can we remove `(Optional)` in the docs for the other fields where it's mentioned?

@jorgeorpinel Any objections?

jorgeorpinel: Makes sense! Maybe just indicate "required" in the few metafile fields/ cmd args/ fn params that are. Goes beyond the scope of this PR though

dberenbaum: I think for consistency, it makes sense as part of this PR scope.

skshetry: > Maybe having said optional the "user" wording is no longer necessary: I don't think it's clear, as everything in the pipeline is defined by the user.

- `user` is important here. Other keywords have certain meanings or is associated with something. These are however not. Besides, `desc` already has that wording.

> Why are some <text> and some <str>? Distinction unclear (or needs more detail)

It's fine to leave this out. `text` is multiline string. It could be `<label>`, but since we don't have understanding of those metavars, I think it's fine to keep just `<str>`.

> Should we use <> around key=value?

I don't think it's necessary. It should be `<key>=<value>` actually, if we typed it out. But simple is better, no need to complicate here.

> "metadata ... to the data" seems redundant, can we rephrase?

What do you propose?

Pls hyphenate "user-assigned" if we keep that term.

Sure, I can do that.

> No periods . needed (not full sentences).

We already have some phrases/sentences with periods. It's better to be consistent. I am not fixing it at this time.

skshetry: Let's update in a separate PR. It touches other parts too, eg: `stage` definition.

dberenbaum: Sounds good, @skshetry . Could you make those changes?

dberenbaum: I added to this PR to save from having to make another one.

dberenbaum (approved):

PR #3951, opened 2022-09-15
Renovate group iterative/dvcorg packages

yathomasi:
I am hoping this way we get our packages grouped together which will make it easier to review and also it will be available sooner.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,10 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base", "group:allNonMajor"]
+  "extends": ["config:base", "group:allNonMajor"],
+  "packageRules": [
+    {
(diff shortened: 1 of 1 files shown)


rogermparent:
This should also allow us to more easily allow in the "all minor updates" PRs while letting us inspect theme updates specifically since they have the possibility to break things. Good idea!

rogermparent:
Seems like we're not getting a PR for the latest theme version, but it's still not in the all minor versions PR. Did we misconfigure something here?

yathomasi:
> Seems like we're not getting a PR for the latest theme version, but it's still not in the all minor versions PR. Did we misconfigure something here?

Actually, we were not getting the PR for the theme on dvc.org(it was there for other websites) and that was the main reason for putting this PR. 

Maybe we could add 
```
"matchUpdateTypes": ["minor", "patch", "pin", "digest", "rollback"],
"prCreation": "immediate",
```
to make it promptly create PR.

PR #3943, opened 2022-09-14
start: clean up pages (index)

jorgeorpinel:
For #3678

- [x] Do not merge before parent PR, https://github.com/iterative/dvc.org/pull/3942.

- Next: https://github.com/iterative/dvc.org/pull/3955

Issue #3678, opened 2022-06-20
start: less content in sub-index pages


jorgeorpinel:
The only issue with adding this middle click-level is that it's possible to "get lost" in the navigation. Example: I click Get Started and read the general index. Then I only open the Data Management element in the nav (click the little arrow) \*without opening its index page\* and see that there's pages inside. I then assume the content starts in the first sub-page (Accessing Data) and click on that. I have now missed all of the basic info on data mgmt. Cc @shcheklein 

![image](https://user-images.githubusercontent.com/1477535/165393264-a7bad7f0-1f45-4454-a7fe-528785dd0414.png)

_Originally posted by @jorgeorpinel in https://github.com/iterative/dvc.org/pull/3455#discussion_r859150197_

---

Resurfaced recently in https://github.com/iterative/dvc.org/issues/3676#issuecomment-1160838221:

> Clearly this index page (the first part of it?) should belong here https://dvc.org/doc/start/data-management
And https://dvc.org/doc/start/data-management should become a separate section

jorgeorpinel:
@iesahin could you still look into this?

PR #3955, opened 2022-09-15
start: clean up pages (Data Versioning)

jorgeorpinel:
For https://github.com/iterative/dvc.org/issues/3678

- Prev: https://github.com/iterative/dvc.org/pull/3943
- Next: https://github.com/iterative/dvc.org/pull/3956

In review app: https://dvc-org-start-pages-ver-idxbzm.herokuapp.com/doc/start/data-management

- [x] Update links - https://github.com/iterative/dvc.org/pull/3973


content/docs/start/data-management/data-versioning.md
```diff
@@ -0,0 +1,260 @@
+---
+title: 'Get Started: Data Versioning'
+description: 'Get started with data and model versioning in DVC. Learn how to
+use a regular Git workflow for datasets and ML models, without storing large
+files in Git.'
+---
+
+# Get Started: Data Versioning
```
jorgeorpinel, line 8: Nothing changed below, this comes straight from the existing [Data Management index page](https://dvc.org/doc/start/data-management) (after **Expand to get an example dataset**).

jorgeorpinel: Just not sure whether to call this Data **and Model** Versioning. The next page is called "Data and Model Access" so it would get a little repetitive (in the nav at least). The next one is just "Data Pipelines".

content/docs/start/data-management/index.md
```diff
@@ -1,11 +1,10 @@
...
-description: 'Get started with data and model versioning in DVC. Learn how to
-use a regular Git workflow for datasets and ML models, without storing large
-files in Git.'
+title: 'Get Started: Data Management'
+description: 'Get started with data management in DVC, including versioning
+access, pipelines, as well as metrics, parameters, and plots.'
 ---
 
-# Get Started: Data Versioning
+# Get Started: Data Management
```
jorgeorpinel, line 7: Here I just renamed the index page.

jorgeorpinel: p.s. not sure whether it makes sense to keep the video in this page though.

jorgeorpinel:
In review app: https://dvc-org-start-pages-ver-idxbzm.herokuapp.com/doc/start/data-management

shcheklein:
This [page](https://dvc-org-start-pages-ver-idxbzm.herokuapp.com/doc/start/data-management) doesn't make much sense to me tbh.

- The `Expand to get dataset` comes on what is looking an intro page out of the blue and the very end.
- We can safely move the whole page content where it was before - the first page of the Data Versioning. Partially to the index page.


--------

Overall re get started- let's please get back to do more trails, rebuilding the experiments trail, decide what to do with Pipelines (name Data Management is a bit misleading). I think we need to focus on things like this first before optimizing pages.

First step (since might take time to do more complicated things) - let's clarify that data management includes pipelines, for example. That actually each trail is comprehensive more or less and its name defines an angle from which we describe _all_ fearures.

jorgeorpinel:
> This page doesn't make much sense to me tbh.

Agreed. I guess I can just completely remove that whole index page but then title "Data Management" will only be shown in nav (no actual page for search engine results) -- is that OK?

> let's clarify that data management includes pipelines

I always thought we needed a separate trail for pipelines but I believe you said it could fit into data management? TBH I'm not sure what's the rationale in order to clarify this: how do we explain pipelines as part of data management?

> each trail is comprehensive more or less and its name defines an angle from which we describe all fearures

Good idea. Will make a separate PR for this ⏳

content/docs/start/data-management/index.md
```diff
@@ -1,294 +1,17 @@
...
 
-How cool would it be to make Git handle arbitrarily large files and directories
-with the same performance it has with small code files? Imagine cloning a
-repository and seeing data files and machine learning models in the workspace.
-Or switching to a different version of a 100Gb file in less than a second with a
-`git checkout`. Think "Git for data".
+Let's look at DVC's features from the perspective of data and machine learning
+model management. This includes automatic caching; versioning on top of Git
+(without storing in the Git repo); sharing, exploring, and accessing remotely,
+among other tasks.
```
jorgeorpinel, line 12: OK so now the content in this page is completely new, just the title for the trail is kept...

content/docs/start/data-management/data-versioning.md
```diff
@@ -0,0 +1,294 @@
...
+
+How cool would it be to make Git handle arbitrarily large files and directories
+with the same performance it has with small code files? Imagine cloning a
+repository and seeing data files and machine learning models in the workspace.
+Or switching to a different version of a 100Gb file in less than a second with a
+`git checkout`. Think "Git for data".
```
jorgeorpinel, line 14: ... Everything else was moved to the Data Versioning page.

jorgeorpinel: One could argue that [Storing and sharing](https://dvc.org/doc/start/data-management#storing-and-sharing) and [Retrieving](https://dvc.org/doc/start/data-management#retrieving) should stay in DM though...

jorgeorpinel:
An alternative approach to solving #3678 is in https://github.com/iterative/dvc.org/pull/3956. Please lmk which one to use in both cases and I'll update one or the other PR accordingly.

shcheklein (approved):

PR #3956, opened 2022-09-15
start: clean up pages (Experiments)

jorgeorpinel:
Closes https://github.com/iterative/dvc.org/issues/3678.

- Prev: https://github.com/iterative/dvc.org/pull/3955

In review app: https://dvc-org-start-pages-exp-tfyhfk.herokuapp.com/doc/start/experimentation

- [x] Update links - https://github.com/iterative/dvc.org/pull/3974


shcheklein:
Tbh, I don' feel the motivation behind this. It complicates things even further. What would be more reasonable is to take the whole page and move as child - as-is (name it "Intro" or somehow else). Since both pages contain content of the same level. And there is no need in having an index I think.

jorgeorpinel:
> take the whole page and move as child - as-is

OK, moved as-is to https://dvc-org-start-pages-exp-tfyhfk.herokuapp.com/doc/start/experimentation/experiments now.

content/docs/start/experimentation/experiments.md
```diff
@@ -1,5 +1,8 @@
 ---
 title: 'Get Started: Experiments'
+description: 'Get started with experiment management in DVC. Learn how to
+capture multiple project versions automatically without bloating the Git repo.
+Iterate quickly by comparing them easily and finding best performers.'
```
jorgeorpinel, line 5: Well, I did add a SEO description but this is not visible.

shcheklein (approved):

jorgeorpinel:
Curious about your opinion too @dberenbaum . Both this pr and https://github.com/iterative/dvc.org/pull/3955 are approved but the approach in each one was different: that one keeps the **Data Management** index page while this one does not keep an **Exp Mgmt** index page (it's just an empty nav entry).

Pros:
- Less pages to maintain
- User goes straight into the action

Cons:
- The empty nav entry probably doesn't influence SEO as an actual page could
- The index page could be used to summarize the trail + emphasize that it's comprehensive (covers all applicable DVC features)

jorgeorpinel:
OK well I'm going to merge this for now. All the redirects are 302s anyway, we can revisit.

dberenbaum:
Sorry, I missed this mention. No strong opinion on whether to have an index page. LGTM.

jorgeorpinel:
Rel. https://github.com/iterative/dvc.org/pull/4000

PR #3959, opened 2022-09-16
Remove allowTransparency from iframe and update non-major dependencies

rogermparent:
Based off #3907 with a lint fix


src/components/Support/index.tsx
```diff
@@ -47,7 +47,6 @@ const SupportPage: React.FC = () => (
                       src="https://discordapp.com/widget?id=485586884165107732&theme=light"
                       width="350"
                       height="500"
-                      allowTransparency
```
rogermparent, line 50: This is the prop in question

rogermparent:
actually, this prop doesn't even apply on the `iframe` in production when it's defined. I think it's safe to just roll this into the renovate PR.

Issue #3953, opened 2022-09-15
plots: add bar_horizontal and bar_horizontal_sorted templates


dberenbaum:
### Report

@tibor-mach @pared Would one of you mind adding these new templates to the docs? Thanks!

tibor-mach:
I can do that...but dunno how to contribute to the docs :-)

tibor-mach:
You don't use Sphinx, right?

pared:
@tibor-mach 
we are using this repository, adding those templates would require changing the pages responsible for examples:
`content/docs/command-reference/plots/show.md` and including those templates in:
`content/docs/user-guide/visualizing-plots.md`
If you want to try, we can help you iterate on that. Its basically editing markdown files. :)

dberenbaum:
Thanks @tibor-mach! You can also take a look at https://dvc.org/doc/contributing/docs.

tibor-mach:
PR ready. Just out of curiosity - why don't you use Sphinx (given that dvc is entirely in python and sphinx can work stuff like doctests which are useful to keep documentation up to date)? 

Not to say that you do it wrong, I'm just curious about why you opted for the setup you have :-)

pared:
Regretfully I am unable to answer this question, @shcheklein @jorgeorpinel ?

jorgeorpinel:
Hi. Just for the record, when was this introduced, was it in https://github.com/iterative/dvc/pull/8282 ? Thanks

jorgeorpinel:
p.s. what about vertical bar charts? I had the impression those were also available.

jorgeorpinel:
> why don't you use Sphinx

@tibor-mach we are considering automating some of the command references but most of our docs are not connected to docstrings directly.

tibor-mach:
@jorgeorpinel no, it was [here](https://github.com/iterative/dvc-render/issues/8)

@jorgeorpinel We had a discussion about the plot options in general. It is very easy to make plots vertical as well, or to make them vertical and sorted or ... the problem is that you can then quickly end up with so many vega templates that have practically the same code (with a single line being different). Of course, I am not a part of your team, so not my call, but personally I'd consider a "feature freeze" on the templates and try to come up with a way to make plots more "modular" than with static templates...

[...]

PR #3964, opened 2022-09-17
horizontal bar plot docs

tibor-mach:
Solves #3953


content/docs/command-reference/plots/show.md
```diff
@@ -492,3 +492,46 @@ file:///Users/usr/src/dvc_plots/index.html
 ```
 
 ![](/img/plots_show_confusion_normalized.svg)
+
```
shcheklein, line 495: @tibor-mach something funny is happening here 🤔  Did you run the pre-commit hook?

shcheklein:
@iterative/websites folks we need to figure out deployments, restyled, etc for the external contributors when we have time. Something happened to the restyled. Can we use the logic similar to the one we did in PyDrive2 that @0x2b3bfa0 discovered (using pull_request_target and deployments to approve runs - https://github.com/iterative/PyDrive2/blob/main/.github/workflows/test.yml ?)

content/docs/command-reference/plots/show.md
tibor-mach: I didn't because the `.pre-commit-config.yaml` seems to be missing in the repo. There is a `.pre-commit-hooks.yaml` file but it is not a valid config...so I am bit confused about how I can run it.

shcheklein: Yes, in this case it's installed if you run `yarn`.

shcheklein: Here are a bit more details - https://dvc.org/doc/contributing/docs#development-environment, but you don't need to run everything. And we are figuring out how to simplify the workflow for the external contributors. We had an external tool (Restyled) that was taking care of running linters and creating a fixed PR on top of a contribution.

tibor-mach: I did run `yarn` though. I set up a locally hosted version of the website with `yarn develop` and had it opened as I was editing the code.

I now also ran all the liters etc. manually. Everything seems to be fine. Building it also works fine...

I do get the following warning and error when I run `yarn develop` though. Not sure if this might have something to do with the problem here (I am not exactly sure what the problem is as it works on my machine ... :-) )

```
(process:22910): GModule-CRITICAL **: 21:18:59.521: g_module_make_resident: assertion 'module != NULL' failed

(process:22910): VIPS-WARNING **: 21:18:59.521: unable to load "/usr/lib/vips-modules-8.13/vips-jxl.so" -- libjxl_threads.so.0.6: cannot open shared object file: No such file or directory
```

shcheklein:
Okay, it works fine https://dvc-org-main-ffxxiia534cu28qo8.herokuapp.com/doc/command-reference/plots/show#example-horizontal-bar-plot

I think GH highlights it for a different reason (malformed JSON above on the page)

Linters also work fine and there are no problems.

content/docs/command-reference/plots/show.md
```diff
@@ -492,3 +492,42 @@ file:///Users/usr/src/dvc_plots/index.html
...
+![](/img/plots_show_bar_horizontal.svg)
+
+When using the `bar_horizontal` template the plot is sorted by the vertical axis
+(in our example this means alphabetically, by feature name).
+
+If you want it to be sorted by the horizontal axis (by feature importance in the
+example) you can use the `bar_horizontal_sorted` template instead:
+
+```dvc
+$ dvc plots show importances.csv -t bar_horizontal_sorted
```
shcheklein, line 528: we need `\` at the end of this command

tibor-mach: Fixed. Btw, this is probably missing with the confusion matrix example above as well (I copied most of the structure from there)

shcheklein: yes, I saw this. Feel free to fix as part of this PR or I'll fix it when this one is merged.

shcheklein (approved):

pared (approved):

shcheklein (approved):

rogermparent:
Maybe re-creating the external applications we use like Restyled within our lint pipeline could cut down on weirdness: If GHA is down, CI is down, with no other delays.

We don't have to port everything at once, but I definitely think Restyled shouldn't be too difficult to replicate in our CI- I'd imagine we just have to commit and/or open a PR with the results of `prettier --write` and/or `eslint --fix` if anything changes.

shcheklein:
@rogermparent yes, I agree and we discussed this with @yathomasi - we have a lot of stuff in place to make it quick I think. CML can help a lot with creating and updating PRs - it's literally one-two lines, and we can reuse the same mechanism for approvals as we do in PyDrive2 and CML (abusing the deployments). We should probably create a ticket for this.

yathomasi:
> We should probably create a ticket for this.

We already have one: https://github.com/iterative/dvc.org/issues/3970

content/docs/user-guide/visualizing-plots.md
```diff
@@ -279,6 +279,9 @@ DVC has the following built-in plot templates:
   [example](/doc/command-reference/plots/show#example-confusion-matrix)
 - `confusion_normalized` - confusion matrix with values normalized to <0, 1>
   range
+- `bar_horizontal` - horizontal bar plot, see
+  [example](/doc/command-reference/plots/show#example-horizontal-bar-plot)
+- `bar_horizontal_sorted` - horizontal bar plot sorted by bar size
```
jorgeorpinel, line 284: Minor: I think "bar chart" is much more widely used. But surprisingly (for me) term "bar plot" is [not insignificant](https://trends.google.com/trends/explore?q=bar%20plot,bar%20chart) out there 🤷🏼

PR #3968, opened 2022-09-19
api docs: use `f` instead of `fd`, and file-like object instead of file descriptor

skshetry:
`fd` means file-descriptor which is not what `open` or `fs.open()` returns. It returns a [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object). 

Both `open()` and `Path.open()` uses `f` in Python's docs. 
1. https://docs.python.org/3/library/functions.html#open
2. https://docs.python.org/3/library/pathlib.html#pathlib.Path.open

`fsspec` does the same: https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.open.

### DIFF
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -73,15 +73,15 @@ from dvc.api import make_checkpoint
 while True:
     try:
         if os.path.exists("int.txt"):
-            with open("int.txt", "r") as fd:
-                i_ = int(fd.read()) + 1
+            with open("int.txt", "r") as f:
+                i_ = int(f.read()) + 1
         else:
             i_ = 0
 
         # ... do something meaningful
 
-        with open("int.txt", "w") as fd:
-            fd.write(f"{i_}")
+        with open("int.txt", "w") as f:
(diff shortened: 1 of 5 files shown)


content/docs/api-reference/make_checkpoint.md
```diff
@@ -73,15 +73,15 @@ from dvc.api import make_checkpoint
...
 
         # ... do something meaningful
 
-        with open("int.txt", "w") as fd:
-            fd.write(f"{i_}")
+        with open("int.txt", "w") as f:
+            f.write(f"{i_}")
 
         if i_ % 100 == 0:
             make_checkpoint()
```
skshetry, line 87: This code bothers me a lot. 😄

shcheklein (approved):

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: Any suggestions? Cc @daavoo

skshetry: I have a patch, trying to propose one.

PR #3969, opened 2022-09-19
docs: fix code formatting/indentations

skshetry:
- Adding hanging indent where matters
- moving end-bracket to a newline (without indent) like how `black` does
- reducing indentations from 8 to 4
- trying to fit some code into a single line (I did not see any reason to hang some of them).

### DIFF
--- a/content/docs/api-reference/open.md
+++ b/content/docs/api-reference/open.md
@@ -17,9 +17,9 @@ def open(path: str,
 import dvc.api
 
 with dvc.api.open(
-        'get-started/data.xml',
-        repo='https://github.com/iterative/dataset-registry'
-        ) as f:
+    'get-started/data.xml',
+    repo='https://github.com/iterative/dataset-registry'
+) as f:
     # ... f is a file-like object that can be processed normally.
 ```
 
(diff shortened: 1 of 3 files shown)


content/docs/api-reference/open.md
```diff
@@ -177,11 +174,7 @@ providing a `remote` argument:
 ```py
 import dvc.api
 
-with dvc.api.open(
-        'activity.log',
-        repo='location/of/dvc/project',
-        remote='my-s3-bucket'
-        ) as f:
+with dvc.api.open('activity.log', remote='my-s3-bucket') as f:
```
skshetry, line 177: Removed `repo` here for readability. It's about the `remote` argument, so I think it's okay to skip` repo=` kwarg.

content/docs/api-reference/open.md
```diff
@@ -154,10 +154,7 @@ CSV dataset:
 import csv
 import dvc.api
 
-with dvc.api.open(
-        'clean.csv',
-        rev='v1.1.0'
-        ) as f:
+with dvc.api.open('clean.csv', rev='v1.1.0') as f:
```
skshetry, line 157: This could fit in one single line, so we don't need to hang them.

shcheklein (approved):

PR #3971, opened 2022-09-20
plots: clarify (more) that stage-level plots can be dirs

jorgeorpinel:
per https://github.com/iterative/dvc.org/pull/3940#issuecomment-1251843556

### DIFF
--- a/content/docs/user-guide/visualizing-plots.md
+++ b/content/docs/user-guide/visualizing-plots.md
@@ -217,7 +217,7 @@ details.
 [full format specification]:
   /doc/user-guide/project-structure/dvcyaml-files#top-level-plot-definitions
 
-### Plot Outputs
+### Plot outputs
 
 When defining [pipelines], some <abbr>outputs</abbr> (both files and
 directories) can be placed under a `plots` list for the corresponding stage.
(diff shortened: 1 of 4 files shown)


content/docs/user-guide/visualizing-plots.md
```diff
@@ -217,11 +217,11 @@ details.
 [full format specification]:
   /doc/user-guide/project-structure/dvcyaml-files#top-level-plot-definitions
 
-### Plot Outputs
+### Plot outputs
```
jorgeorpinel, line 220: Unrelated but should we call this section "Stage-level plots" or just add the word (stage) in there? E.g.

```suggestion
### Plot outputs (stage-level)
```

dberenbaum:
Found an old issue that this should fix: https://github.com/iterative/dvc/issues/5216. It mentions `dvc run`, so we should update that and `stage add` for the `--plots/--plot-no-cache options`. Do you want to add that here? Happy to take that if it's easier.

jorgeorpinel:
Good find @dberenbaum ! I'm not supposed to spend much time working on cmd refs though. Trying to focus on higher level content. But if this is approved so far please merge it or take it over within your team 🙂

dberenbaum (approved):

PR #3983, opened 2022-09-22
make_checkpoint: rewrite code for every100th iterations

skshetry:
IANADS so any help on improving this code/docs would be great.

### DIFF
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -59,8 +59,9 @@ stages:
           checkpoint: true
 ```
 
-The code in `train.py` will train a model upto certain epochs. At every 100 iterations,
-it saves the model to `model`, evaluates it and makes a checkpoint for [DVC experiments]:
+The code in `train.py` will train a model upto certain epochs. At every 100
+iterations, it saves the model to `model`, evaluates it and makes a checkpoint
+for [DVC experiments]:
 
 [dvc experiments]: /doc/user-guide/experiment-management#experiments
 
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -60,8 +60,8 @@ stages:
 ```
 
 The code in `train.py` will train a model up to a number of epochs. Every 100
-iterations, it saves the `model`, evaluates it, and makes a checkpoint
-for [DVC experiments]:
+iterations, it saves the `model`, evaluates it, and makes a checkpoint for [DVC
+experiments]:
 
 [dvc experiments]: /doc/user-guide/experiment-management#experiments
 
--- a/content/docs/api-reference/make_checkpoint.md
+++ b/content/docs/api-reference/make_checkpoint.md
@@ -52,42 +52,28 @@ Let's consider the following `dvc.yaml` file:
 
 ```yaml
 stages:
-  every100:
-    cmd: python iterate.py
+  train:
+    cmd: python train.py
     outs:
-      - int.txt:
+      - model:
           checkpoint: true
 ```
 
-The code in `iterate.py` will execute continuously increment an integer number
-saved in `int.txt` (starting at 0). At 0 and every 100 loops, it makes a
-checkpoint for [DVC experiments]:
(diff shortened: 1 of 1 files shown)


content/docs/api-reference/make_checkpoint.md
```diff
@@ -52,42 +52,28 @@ Let's consider the following `dvc.yaml` file:
...
-                i_ = int(f.read()) + 1
-        else:
-            i_ = 0
-
-        # ... do something meaningful
-
-        with open("int.txt", "w") as f:
-            f.write(f"{i_}")
+for epoch in range(epochs):
+    train(model, x_train, y_train)
```
skshetry, line 71: I think `model`/`x_train`/`y_train`/`x_test`/`y_test` are common namings that we don't have to show them in the code.

content/docs/api-reference/make_checkpoint.md
```diff
@@ -52,42 +52,29 @@ Let's consider the following `dvc.yaml` file:
...
+The code in `train.py` will train a model upto certain epochs. At every 100
+iterations, it saves the model to `model`, evaluates it and makes a checkpoint
+for [DVC experiments]:
 
 [dvc experiments]: /doc/user-guide/experiment-management#experiments
 
 ```py
-import os
-
 from dvc.api import make_checkpoint
```
skshetry, line 69: I wonder why we named it `make_checkpoint()` instead of just `checkpoint()`.

content/docs/api-reference/make_checkpoint.md
```diff
@@ -115,15 +102,8 @@ Experiment results have been applied to your workspace.
 > DVC checkpoints to behave as expected.
```
skshetry, line 102: There may be other errors too, so i am not sure whether we should show example of catching `KeyboardInterrupt`.

content/docs/api-reference/make_checkpoint.md
```diff
@@ -52,42 +52,29 @@ Let's consider the following `dvc.yaml` file:
...
-      - int.txt:
+      - model:
           checkpoint: true
 ```
 
-The code in `iterate.py` will execute continuously increment an integer number
-saved in `int.txt` (starting at 0). At 0 and every 100 loops, it makes a
-checkpoint for [DVC experiments]:
+The code in `train.py` will train a model upto certain epochs. At every 100
+iterations, it saves the model to `model`, evaluates it and makes a checkpoint
```
jorgeorpinel, line 63:
```suggestion
The code in `train.py` will train a model up to a number of epochs. Every 100
iterations, it saves the `model`, evaluates it, and makes a checkpoint
```

content/docs/api-reference/make_checkpoint.md
jorgeorpinel: I think the code sample was taken from an actual checkpoints example repo which is a demo meant to run continuously until you Ctrl + C.

dberenbaum: See https://github.com/iterative/dvc-checkpoints-mnist/pull/2

dberenbaum:
My overall take here: I don't have a strong opinion on whether to merge this. It's a more realistic example, but the original example was simplistic on purpose to clarify how it works.

IMO it would be better to focus on improving checkpoints functionality before worrying too much about improving the docs. See https://github.com/iterative/dvc/issues/7381.

content/docs/api-reference/make_checkpoint.md
skshetry: @dberenbaum, I think that could be and should be explained separately. I'd keep it simple for `every 100th iteration` example.

skshetry:
> but the original example was simplistic on purpose to clarify how it works.

I disagree that it was simplistic because of all the reading/writing stuff it is doing, which is confusing. Also, it's 22 lines of code.

content/docs/api-reference/make_checkpoint.md
pmrowla: I used `make_checkpoint` because `checkpoint` is a noun in this context and IMO method names should almost always be verbs.

pmrowla (approved):

PR #3985, opened 2022-09-22
fix: outdated link anchor `#stage-plots`

jorgeorpinel:


dberenbaum (approved):
LGTM. I'd also be fine with changing it back to "stage plots." My only concern is spending too much time on this.

PR #3989, opened 2022-09-23
doc: add doc about foreach target

skshetry:
Syncing with https://github.com/iterative/dvc/pull/8352. I am skipping `repro` and `exp run` because they are well discussed in `Options` sections, and we wanted to avoid describing too much and took the example-based approach in the docs.

See https://github.com/iterative/dvc.org/pull/1983. The support for foreach group name in the target was extended in https://github.com/iterative/dvc/pull/8210.

### DIFF
--- a/content/docs/command-reference/checkout.md
+++ b/content/docs/command-reference/checkout.md
@@ -12,7 +12,7 @@ usage: dvc checkout [-h] [-q | -v] [--summary] [-d] [-R] [-f]
 
 positional arguments:
   targets       Limit command scope to these tracked files/directories,
-                .dvc files, or stage names.
+                .dvc files and stage or foreach-group names.
 ```
 
 ## Description
--- a/content/docs/command-reference/commit.md
+++ b/content/docs/command-reference/commit.md
@@ -9,7 +9,8 @@ usage: dvc commit [-h] [-q | -v] [-f] [-d] [-R]
                   [targets [targets ...]]
 
 positional arguments:
-  targets        Limit command scope to these stages or .dvc files.
+  targets        Limit command scope to these tracked files/directories,
+                 .dvc files and stage or foreach-group names.
                  Using -R, directories to search for stages or .dvc
                  files can also be given.
 ```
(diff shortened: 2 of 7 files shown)


content/docs/command-reference/checkout.md
```diff
@@ -12,7 +12,7 @@ usage: dvc checkout [-h] [-q | -v] [--summary] [-d] [-R] [-f]
 
 positional arguments:
   targets       Limit command scope to these tracked files/directories,
-                .dvc files, or stage names.
+                .dvc files and stage or foreach-group names.
```
daavoo, line 15: I am not sure if people would know what `foreach group` means. 

It is not used in https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#foreach-stages which are the only existing docs for `foreach`

dberenbaum:
Thanks, @skshetry!

> I am skipping `repro` and `exp run` because they are well discussed in the [**Options**](https://dvc.org/doc/command-reference/repro#options) sections, and we wanted to avoid describing too much and took the example-based approach in the docs.

Should we add a `foreach` example in the options section then? I don't see `foreach` mentioned anywhere. Also, same with `stage list`.

content/docs/command-reference/checkout.md
jorgeorpinel: Agreed. I don't think this change is needed TBH: foreach groups ultimately become stages so it's covered with "or stage names". No need to be that super explicit about every possible syntax in the help output, especially for such an advanced feature (not that used from what I understand) -- explaining that you can target them in the foreach docs should be enough IMO.

jorgeorpinel:
> add a foreach example in the options section then? 

Adding that to `targets` (under https://dvc.org/doc/command-reference/repro#options) as well as possibly a stand-alone example could help indeed since this topic is only covered in the dvc.yaml ref atm and we want cmd refs to be as comprehensive/ self-contained as possible.

> p.s. note that `targets` arg in the Options section doesn't have a 🔗 anchor like all the options, probably because it doesn't correspond to an actual option -- if that's a problem we can check what to do about it with the websites team. Cheers

dberenbaum:
Closed in favor of https://github.com/iterative/dvc.org/pull/4010

PR #3994, opened 2022-09-26
dvclive: docs for log_param/log_params

dtrifiro:
documentation for `log_param`/`log_params` (https://github.com/iterative/dvclive/pull/292)

### DIFF
--- a/content/docs/dvclive/api-reference/live/index.md
+++ b/content/docs/dvclive/api-reference/live/index.md
@@ -23,8 +23,8 @@ live = Live()
 
 ## Description
 
-A `Live()` instance is required in order to log machine learning metrics and
-other metadata.
+A `Live()` instance is required in order to log machine learning parameters,
+metrics and other metadata.
 
 <admon type="warn">
 
@@ -72,6 +72,8 @@ other metadata.
 ## Methods
 
 - `Live.log()`
+- `Live.log_param()`
+- `Live.log_params()`
 - `Live.log_image()`
 - `Live.log_plot()`
 - `Live.make_report()`
--- a/content/docs/dvclive/api-reference/live/index.md
+++ b/content/docs/dvclive/api-reference/live/index.md
@@ -71,12 +71,12 @@ metrics and other metadata.
 
 ## Methods
 
+- `Live.get_step()`
 - `Live.log()`
+- `Live.log_image()`
 - `Live.log_param()`
 - `Live.log_params()`
-- `Live.log_image()`
 - `Live.log_plot()`
 - `Live.make_report()`
-- `Live.get_step()`
 - `Live.next_step()`
 - `Live.set_step()`
--- a/content/docs/dvclive/api-reference/live/log_param.md
+++ b/content/docs/dvclive/api-reference/live/log_param.md
@@ -1,0 +1,53 @@
+# Live.log_param()
+
+Logs the given parameter `val` associating it with the given `name`.
+
+```py
+ def log_param(name: str, val: ParamLike):
+```
+
+Also see `Live.log_params()`.
+
+## Usage
+
+```py
+from dvclive import Live
+
+live = Live()
+
+live.log_param("lr", 1e-3)
+live.log_param("n_iter", 50)
+```
+
+## Description
+
+On each `live.log_param(name, val)` call, DVCLive will write the `name`
+parameter to `{Live.dir}/params.yaml` with the corresponding `val`:
+
+```dvc
+$ cat dvclive/params.yaml
+lr: 1e-3
+n_iter: 50
+```
+
+## Parameters
+
+- `name` - Name of the parameter being logged.
+
+- `val` - The value to be logged.
+
+## Exceptions
+
+- `dvclive.error.InvalidParameterType` - thrown if the parameter value is not
+  among supported types. Supported types include:
+
+  ```python
+  ParamLike = Union[
+      int,
+      float,
+      str,
+      bool,
+      List["ParamLike"],
+      Dict[str, "ParamLike"]
+  ]
+  ```
--- a/content/docs/dvclive/api-reference/live/log_params.md
+++ b/content/docs/dvclive/api-reference/live/log_params.md
@@ -1,0 +1,63 @@
+# Live.log_params()
+
+Logs the given group of parameters `params`.
+
+```py
+ def log_params(params: Dict[ParamLike]):
+```
+
+Also see `Live.log_param()`.
+
+## Usage
+
+```py
+from dvclive import Live
+
+live = Live()
+
+params = {
+    "num_classes": 10,
+    "metrics": ["accuracy", "mae"],
+    "training": {
+        "epochs": 42,
+        "batch_size": 128,
+        "optimizer": {
+            "name": "adam",
+            "loss": "categorical_crossentropy",
+            "learning_rate": 1e-3,
+        },
+    },
+}
+live.log_params(params)
+```
+
+## Description
+
+On each `live.log_params(params)` call, DVCLive will write keys/values pairs in
+the `params` dict to `{Live.dir}/params.yaml`:
+
+```dvc
+$ cat dvclive/params.yaml
+lr: 1e-3
+n_iter: 50
+```
+
+## Parameters
+
+- `params` - Dictionary with name/value pairs of parameters to be logged.
+
+## Exceptions
+
+- `dvclive.error.InvalidParameterType` - thrown if any of the parameter values
+  in the `params` dict is not among supported types. Supported types include:
+
+  ```python
+  ParamLike = Union[
+      int,
+      float,
+      str,
+      bool,
+      List["ParamLike"],
+      Dict[str, "ParamLike"]
+  ]
+  ```
--- a/content/docs/dvclive/dvclive-with-dvc.md
+++ b/content/docs/dvclive/dvclive-with-dvc.md
@@ -28,6 +28,8 @@ from dvclive import Live
 
 live = Live("training_metrics")
 
+...
+
 for epoch in range(NUM_EPOCHS):
     train_model(...)
     metrics = evaluate_model(...)
--- a/content/docs/dvclive/get-started.md
+++ b/content/docs/dvclive/get-started.md
@@ -1,7 +1,7 @@
 # Get Started
 
-DVCLive is a Python library for logging machine learning metrics and other
-metadata in simple file formats, which is fully compatible with DVC.
+DVCLive is a Python library for logging machine learning parameters, metrics and
+other metadata in simple file formats, which is fully compatible with DVC.
 
 <admon type="tip">
 
@@ -21,6 +21,34 @@ live = Live()
 ```
 
 See [`Live()`](/doc/dvclive/api-reference/live) for details.
+
+### (Optionally) Log parameters
+
+<toggle>
+<tab title="Single parameter">
+
+```python
+live.log_param("num_classes", 10)
+```
+
+See `Live.log_param()`.
+
+</tab>
+<tab title="Multiple parameters">
+
+```python
+params = {
+    "num_classes": 10,
+    "metrics": ["accuracy", "mae"],
+    "optimizer": "adam"
+}
+live.log_params(params)
+```
+
+See `Live.log_params()`.
+
+</tab>
+</toggle>
 
 ### Log data
 
@@ -76,6 +104,8 @@ from dvclive import Live
 
 live = Live()
 
+live.log_param(NUM_EPOCHS)
+
 for epoch in range(NUM_EPOCHS):
     train_model(...)
     metrics = evaluate_model(...)
@@ -96,6 +126,7 @@ $ tree
 ├── dvclive
 │   ├── images
 │   │   └── img.png
+│   ├── params.yaml
 │   ├── plots
 │   │   └── roc.json
 │   └── scalars
@@ -107,7 +138,8 @@ The contents of the `dvclive` folder and `dvclive.json` will vary depending on
 the type of data you have logged and whether you have updated the step value or
 not.
 
-See `Live.log()`, `Live.log_image()` and `Live.log_plot()` for more details.
+See `Live.log()`, `Live.log_param()`/`Live.log_params()`, `Live.log_image()` and
+`Live.log_plot()` for more details.
 
 ### Metrics report
 
--- a/content/docs/dvclive/ml-frameworks/catalyst.md
+++ b/content/docs/dvclive/ml-frameworks/catalyst.md
@@ -16,7 +16,7 @@ int the callbacks list passed to your
 ```git
 +from dvclive.catalyst import DvcLiveCallback
 
-. . .
+...
 
 runner.train(
     model=model,
--- a/content/docs/dvclive/ml-frameworks/fastai.md
+++ b/content/docs/dvclive/ml-frameworks/fastai.md
@@ -16,7 +16,7 @@ int the callbacks list passed to your
 ```git
 +from dvclive.fastai import DvcLiveCallback
 
-. . .
+...
 
 learn = tabular_learner(data_loader, metrics=accuracy)
 learn.fit_one_cycle(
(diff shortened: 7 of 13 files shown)


content/docs/dvclive/dvclive-with-dvc.md
```diff
@@ -28,6 +28,10 @@ from dvclive import Live
 
 live = Live("training_metrics")
 
+live.log_param("num_epochs", NUM_EPOCHS)
```
dberenbaum, line 31: I suggest not touching the `DVCLive with DVC` page until we complete some more steps to better integrate it into DVC. I don't think it makes much sense right now under stage params.

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -72,6 +75,8 @@ other metadata.
 ## Methods
 
 - `Live.log()`
+- `Live.log_param()`
+- `Live.log_params()`
```
dberenbaum, line 79: I think you need to add these to `sidebar.json`.

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -38,6 +38,9 @@ other metadata.
 - `dir` - Location of the directory to store
   [outputs](/doc/dvclive/get-started#outputs).
 
+- `params_path` - `{Live.params_path}`. Location of the parameters logged via
+  [`log_param`](/doc/dvclive/api-reference/live/log_param#description)/[`log_params`](/doc/dvclive/api-reference/live/log_params#description).
+
 - `summary_path` - `{Live.dir}.json`. Location of the
```
dberenbaum, line 44: cc @daavoo Should this be called a "metrics summary" now to distinguish from params? Should we update the attribute name?

content/docs/dvclive/dvclive-with-dvc.md
dberenbaum: Instead, I think this material should be added to `Get Started`.

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -38,6 +38,9 @@ other metadata.
 - `dir` - Location of the directory to store
   [outputs](/doc/dvclive/get-started#outputs).
 
+- `params_path` - `{Live.params_path}`. Location of the parameters logged via
```
dberenbaum, line 41: Should it be `{Live.params_path}.yaml` to match `summary_path` (sorry this wasn't raised in the core PR)?

content/docs/dvclive/api-reference/live/log_params.md
```diff
@@ -0,0 +1,54 @@
+# Live.log_params()
+
+Logs the given group of parameters `params`, saving them as yaml.
```
dberenbaum, line 3: Not sure it's worth pointing out the yaml format at the top, although making it clear somewhere in this doc makes sense.

content/docs/dvclive/api-reference/live/log_params.md
```diff
@@ -0,0 +1,54 @@
...
+    "lr": 1e-3,
+    "n_iter": 50
+}
+live.log_params(params)
+```
+
+## Description
+
+On each `live.log_params(params)` call, DVCLive will write keys/values pairs in
+the `params` dict to `{Live.params_file}`:
```
dberenbaum, line 28:
```suggestion
the `params` dict to `{Live.params_path}`:
```

Also, if this becomes `{Live.params_path}.yaml`, I think that might be enough to indicate the output format.

content/docs/dvclive/api-reference/live/log_params.md
```diff
@@ -0,0 +1,54 @@
...
+
+```py
+from dvclive import Live
+
+live = Live()
+
+params = {
+    "lr": 1e-3,
+    "n_iter": 50
+}
```
dberenbaum, line 21: What do you think about including a nested param value here?

content/docs/dvclive/api-reference/live/log_params.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Parameters
+
+- `params` - Dictionary with name/value pairs of parameters to be logged.
+
+## Exceptions
+
+- `dvclive.error.InvalidParameterType` - thrown if any of the parameter values
+  in the `params` dict is not among supported types. Supported types include:
+  `{ParamLike=}`.
```
dberenbaum, line 51: Seems like the ParamLike definition is missing here?

content/docs/dvclive/dvclive-with-dvc.md
dtrifiro: Agree

content/docs/dvclive/api-reference/live/log_params.md
dtrifiro: Yeah, that's fine

content/docs/dvclive/api-reference/live/index.md
dtrifiro: +1

Maybe we should also consider rethinking the summary/report names, they're somehow similar and it can be confusing

content/docs/dvclive/api-reference/live/log_params.md
dtrifiro: > Should it be {Live.params_path}.yaml to match summary_path (sorry this wasn't raised in the core PR)?
...
> Also, if this becomes {Live.params_path}.yaml, I think that might be enough to indicate the output format.

I'd like to have something like that, but I'd expect `dvclive.params_path` to be a full path (e.g. `dvclive/params.yaml`) Having something like `dvclive/params` would be pretty odd.

content/docs/dvclive/api-reference/live/log_params.md
dtrifiro: How about something like this? It showcases quite a few possibilities, although it's a bit large snippet

```python
params = {
    "num_classes": 10,
    "metrics": ["accuracy", "mae"],
    "training": {
        "epochs": 42,
        "batch_size": 128,
        "optimizer": {
            "name": "adam",
            "loss": "categorical_crossentropy",
            "learning_rate": 1e-3,
        },
    },
}
```

content/docs/dvclive/api-reference/live/index.md
dberenbaum: Sorry, got confused here. Can't it just be `params.yaml` since we got rid of the custom path?

dberenbaum: In which case, does it even need to be a documented attribute/property?

dtrifiro: I'd go with `{Live.dir}/params.yaml`

content/docs/dvclive/api-reference/live/log_params.md
dtrifiro: Added the definition, although I'm not sure it's worth it to have the exact definition in the docs. Maybe it will be enough to have something like:

> Supported types include basic python data types such as int, float, bool as well as maps and lists.

dberenbaum:
@dtrifiro Let me know when it's ready for another review, thanks!

dtrifiro:
@dberenbaum ready for review

content/docs/dvclive/api-reference/live/log_param.md
```diff
@@ -0,0 +1,70 @@
...
+live = Live()
+
+live.log_param("lr", 1e-3)
+live.log_param("n_iter", 50)
+```
+
+<admon type="note">
+
+When using `Live(resume=True)`, calling `log_param()` after resuming a session
+will be a no-op.
```
daavoo, line 25: Sorry for coming back to this at the docs stage, but what was the motivation for this?

Why having different params after `resume=True` is not considered valid?

cc @dberenbaum

dtrifiro: I'd find it a bit odd to start one experiment with a set of parameters and resume it with another. In the context of experiment tracking, what set of parameters would be considered valid?

dtrifiro: I agree having a no-op might be odd. Alternatives include
- overwriting `params.yaml` with the most recent one
- reading the existing `params.yaml` and comparing the parameters to the ones being logged and warning or failing when parameters do not match.
- ...

daavoo: Note that `resume=True` was mainly introduced for `dvc checkpoints`, where one of the motivations is to resume from a checkpoint after modifying some parameters (https://dvc.org/doc/user-guide/experiment-management/checkpoints#starting-from-an-existing-checkpoint).

Outside `dvc checkpoints`, `resume=True` is just something I am not sure we should handle at all 😅, but we do. 

Your concern is valid but I guess we should let the users decide how to handle it. It is only a problem if you execute the training script without persisting the experiment.
If you `resume=True` from a persisted experiment, then the traceability is clear and it makes sense to allow overriding params.

daavoo: > Alternatives include

@dtrifiro what if we change https://github.com/iterative/dvclive/blob/afa21a17c89c048a0e93c0dfbc7c57f8fbed5c83/src/dvclive/live.py#L230 to :

```python
if param_name in self._params and not self._resume:
```

dtrifiro: This could be ok, although it might allow to log the same parameter twice when using resume.

dtrifiro: This will only raise when logging the same parameter twice on step 0, although I feel like it's over-complicating this a bit
```python
            if param_name in self._params and not (
                self._resume and self.get_step()
            ):
```

dberenbaum: > Note that `resume=True` was mainly introduced for `dvc checkpoints`, where one of the motivations is to resume from a checkpoint after modifying some parameters (https://dvc.org/doc/user-guide/experiment-management/checkpoints#starting-from-an-existing-checkpoint).

This is a good point, so if it's easy to accommodate, let's do it. I don't mind generally allowing users to override the dvclive values even if it could lead to dangerous/unexpected results. Users can mostly manage this themselves, and it's probably better to provide flexibility than enforce some strict ideas of correctness.

However, I don't feel the use case is that important, so let's not go crazy addressing this workflow. The idea of changing parameters mid-experiment has always seemed a little contrived to me and not something I have seen users asking for.

dberenbaum: What would the behavior be if you remove https://github.com/iterative/dvclive/blob/afa21a17c89c048a0e93c0dfbc7c57f8fbed5c83/src/dvclive/live.py#L223-L227? Any parameters that are already logged would fail, but new parameters could still be logged (before or after resume)? I think that's fine if so. Again, I don't think we need to worry too much about either enforcing correctness or addressing every possible workflow for now.

dberenbaum: Also fine: dropping `ParameterAlreadyLoggedError` altogether and leaving this entirely to users. I probably prefer that most. This kind of error is a little more important in metrics logging where users may forget to increment the step. 

Sorry for the lengthy responses here.

content/docs/dvclive/get-started.md
```diff
@@ -22,6 +22,51 @@ live = Live()
 
 See [`Live()`](/doc/dvclive/api-reference/live) for details.
 
+### Log parameters
```
dberenbaum, line 25:
```suggestion
### (Optionally) Log parameters
```
Technically, it's all optional 😄 , but it might be worth making explicit that you don't have to log params.

content/docs/dvclive/get-started.md
```diff
@@ -22,6 +22,51 @@ live = Live()
...
+live.log_param("training",
+    "epochs": 42,
+        "batch_size": 128,
+        "optimizer": {
+            "name": "adam",
+            "loss": "categorical_crossentropy",
+            "learning_rate": 1e-3,
+        },
+    }
+)
```
dberenbaum, line 42: All this detail is great for the API ref, but it's too much here. Let's leave it at the simple example to save space.

content/docs/dvclive/get-started.md
```diff
@@ -22,6 +22,51 @@ live = Live()
...
+        "epochs": 42,
+        "batch_size": 128,
+        "optimizer": {
+            "name": "adam",
+            "loss": "categorical_crossentropy",
+            "learning_rate": 1e-3,
+        },
+    },
+}
+live.log_params(params)
```
dberenbaum, line 62: Again, we can simplify for get started. It doesn't need to be more than 3 lines at most.

content/docs/dvclive/get-started.md
```diff
@@ -22,6 +22,51 @@ live = Live()
...
+```
+
+</tab>
+</toggle>
+
+See `Live.log_param()` and `Live.log_params()`.
```
dberenbaum, line 68: If we are doing the tabs, can we embed these links in the tabs like for `Log data` below?

content/docs/dvclive/get-started.md
```diff
@@ -76,6 +121,10 @@ from dvclive import Live
 
 live = Live()
 
+model_params = {...}
+live.log_param(NUM_EPOCHS)
+live.log_params(model_params)
```
dberenbaum, line 126: Let's drop this. We don't need both param methods in this example.

content/docs/dvclive/api-reference/live/log_param.md
dtrifiro: >What would the behavior be if you remove [...] Any parameters that are already logged would fail, but new parameters could still be logged (before or after resume)? I think that's fine if so

The idea behind it is to be being able to have an idempotent training script, which you can run to resume training in case files already exist. If we drop that, we'll break this workflow (logging the same parameter twice will error out), unless we decide to drop `ParameterAlreadyLoggedError`, though I fear this might make for some annoying and hard to discover bugs when logging (by mistake) different values to the same key.

dberenbaum: I think we are overthinking this and should start by dropping `ParameterAlreadyLoggedError`. I think users can handle checking whether they log different values to the same key and would not expect DVCLive to be responsible for checking this.

dtrifiro: Got rid of it both here and in dvclive: https://github.com/iterative/dvclive/pull/316

content/docs/dvclive/api-reference/live/log_param.md
```diff
@@ -0,0 +1,60 @@
...
+lr: 1e-3
+n_iter: 50
+```
+
+<admon type="tip">
+
+The parameters file `{Live.dir}/params.yaml` is usable by `dvc params`,
+`dvc metrics` and `dvc exp show`/`dvc exp diff`.
+
+</admon>
```
dberenbaum, line 38: It's not that clear how to use it with any of these commands, or why we would suggest both `dvc params` and `dvc metrics`. What do you think about leaving this out until we add better support for tracking it in DVC?

content/docs/dvclive/api-reference/live/index.md
```diff
@@ -71,10 +71,12 @@ other metadata.
...
 ## Methods
 
+- `Live.get_step()`
 - `Live.log()`
 - `Live.log_image()`
+- `Live.log_param()`
+- `Live.log_params()`
 - `Live.log_plot()`
 - `Live.make_report()`
-- `Live.get_step()`
```
jorgeorpinel, line 78: Why was `get_step` moved up?

dberenbaum: I think it was made alphabetical, but good catch that it doesn't match the sidebar. If you're fine with it, we can update the sidebar also. Probably easier for users to find things that way than trying to guess some order of importance.

jorgeorpinel: Sure up to you. I was just pointing out a possible discrepancy I saw.

Issue #1943, opened 2020-11-14
start: provide code samples, make it more actionable


JIoJIaJIu:
It would be great that every chapter of [Get Start](https://dvc.org/doc/start) provides instruction how to prepare a needful repo structure to proceed further
At this moment it's hard to init needful structure if you start not from the first chapter

In one chapter such approach is applied - https://dvc.org/doc/start/data-management/data-pipelines
> Expand to download example code.

- [x] but it's not enough and the user needs to do few additional steps to proceed further:
* run `git init`, covered with `Please also add or commit the source code directory with Git at this point`
* run `dvc init`
* fetch `data/data.xml`

- [ ] Also if I would like to proceed [Experiments chapter](https://dvc.org/doc/start/experiment-management/experiments) I need to have `model.pkl` there. But adding `model.pkl` is marked optional a the previous chapter [Data Pipelines/Dependency graphs (DAGs)/Expand to add more stages](https://dvc.org/doc/start/data-management/data-pipelines#expand-to-add-more-stages). So the user can easier skip the step and after get an error in **Experiments**.

Providing at the start of every chapter the steps how to initialize the repo will make chapters independent and help to avoid few errors

jorgeorpinel:
> instruction how to prepare a needful repo structure to proceed further

Yeah, we used to have this in the previous version of the GS. Not sure why we no longer have those since it can be hidden in an expandable section indeed. Let's do this unless you remember a reason for this @shcheklein ?

> Also if I would like to proceed Experiments chapter I need to have model.pkl there. But adding model.pkl is marked optional in the previous chapter

Also true, we should definitely address this part. Could be addressed by the same solution like Guro mentioned.

shcheklein:
We had it back and forth a few times already. Doing it completely reproducible and chapters independent makes it a hands-on tutorial like (which get started is not). It means longer read time, more distractions for users who want to get the idea quick, etc. 

The idea with the recent changes was to use Katacoda extensively to provide such playground (it still out of date btw, which is on of the biggest priorities). 

If Katacoda is not visible we should put links into the expandable sections. It should solve this problem.

> Also true, we should definitely address this part. Could be addressed by the same solution like Guro mentioned.

it's not marked optional. The problem should be addressed in a similar way- change language at the very beginning, may be at the very top of every chapter. Make it super clear "how to use the get started".

shcheklein:
Katacoda - we had another complain recently, btw on Discord.

shcheklein:
[...]

PR #3998, opened 2022-09-27
start: complete setup for data pipelines

jorgeorpinel:
Per #1943

> ... the user needs to do few additional steps...
> - ~~run git init~~
> - ~~run dvc init~~
> - ~~fetch data/data.xml~~
> UPDATE: Or just checkout appropriate tag from example repo

### DIFF
--- a/content/docs/start/data-management/data-pipelines.md
+++ b/content/docs/start/data-management/data-pipelines.md
@@ -32,7 +32,18 @@ into a [stage](/doc/command-reference/stage):
 
 ### ⚙️ Expand to download example code.
 
-Get the sample code, initialize DVC, and save everything to Git:
+
+<admon type="info">
+
+To actually try this example, begin after following the
+[Data Versioning](/doc/start/data-management/data-versioning) chapter.
+You can get there by cloning the `2-track-data` tag of our
+[example-get-started](https://github.com/iterative/example-get-started)
+repo.
+
+</admon>
+
+Get the sample code like this:
 
 ```dvc
 $ wget https://code.dvc.org/get-started/code.zip
@@ -47,12 +58,6 @@ $ tree
     ├── prepare.py
     ├── requirements.txt
     └── train.py
-
-$ git init  # if needed
-$ dvc init
-$ git add .
-$ git commit -a -m "Create evaluation stage"
-$ dvc pull data/data.xml
 ```
 
 Now let's install the requirements:
@@ -63,7 +68,6 @@ Now let's install the requirements:
 
 ```dvc
 $ pip install -r src/requirements.txt
-$ git commit
 ```
 
 </details>
--- a/content/docs/start/data-management/data-pipelines.md
+++ b/content/docs/start/data-management/data-pipelines.md
@@ -32,14 +32,12 @@ into a [stage](/doc/command-reference/stage):
 
 ### ⚙️ Expand to download example code.
 
-
 <admon type="info">
 
 To actually try this example, begin after following the
-[Data Versioning](/doc/start/data-management/data-versioning) chapter.
-You can get there by cloning the `2-track-data` tag of our
-[example-get-started](https://github.com/iterative/example-get-started)
-repo.
+[Data Versioning](/doc/start/data-management/data-versioning) chapter. You can
+get there by cloning the `2-track-data` tag of our
+[example-get-started](https://github.com/iterative/example-get-started) repo.
 
 </admon>
 
--- a/content/docs/start/data-management/data-pipelines.md
+++ b/content/docs/start/data-management/data-pipelines.md
@@ -50,6 +50,7 @@ $ tree
 
 $ git init  # if needed
 $ dvc init
+$ git add .
 $ git commit -a -m "Create evaluation stage"
 $ dvc pull data/data.xml
 ```
--- a/content/docs/start/data-management/data-pipelines.md
+++ b/content/docs/start/data-management/data-pipelines.md
@@ -61,7 +61,7 @@ Now let's install the requirements:
 
 ```dvc
 $ pip install -r src/requirements.txt
-$ git commit 
+$ git commit
 ```
 
 </details>
--- a/content/docs/start/data-management/data-pipelines.md
+++ b/content/docs/start/data-management/data-pipelines.md
@@ -50,7 +50,8 @@ $ tree
 
 $ git init  # if needed
 $ dvc init
-$ git commit .dvc params.yaml src
+$ git commit -a -m "Create evaluation stage"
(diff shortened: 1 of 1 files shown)


content/docs/start/data-management/data-pipelines.md
```diff
@@ -47,6 +47,10 @@ $ tree
     ├── prepare.py
     ├── requirements.txt
     └── train.py
+
+$ git init  # if needed
+$ dvc init
+$ git commit .dvc params.yaml src
```
jorgeorpinel, line 53: Ah wait I forgot about fetch.
```suggestion
$ git commit -a -m "Create evaluation stage"
$ dvc fetch data/data.xml
```
Or should it be pull?
```suggestion
$ git commit -a -m "Create evaluation stage"
$ dvc pull data/data.xml
```
Please commit the right one.

jorgeorpinel:
p.s. not sure this is wanted per https://github.com/iterative/dvc.org/issues/1943#issuecomment-728193024:

> Doing it completely reproducible and chapters independent makes it a hands-on tutorial like (which get started is not). It means longer read time, more distractions for users who want to get the idea quick, etc.

content/docs/start/data-management/data-pipelines.md
```diff
@@ -47,6 +47,11 @@ $ tree
     ├── prepare.py
     ├── requirements.txt
     └── train.py
+
+$ git init  # if needed
+$ dvc init
+$ git commit -a -m "Create evaluation stage"
```
dberenbaum, line 53:
```suggestion
$ git add .
$ git commit -a -m "Create evaluation stage"
```

content/docs/start/data-management/data-pipelines.md
```diff
@@ -47,6 +47,11 @@ $ tree
     ├── prepare.py
     ├── requirements.txt
     └── train.py
+
+$ git init  # if needed
+$ dvc init
```
shcheklein, line 52: folks, I think it might confuse users even more ... also I it won't make it "actionable", you need a lot of steps (dvc add / import at the very least to start). The previous section serves as an actionable item for this.

content/docs/start/data-management/data-pipelines.md
```diff
@@ -57,10 +62,9 @@ Now let's install the requirements:
 
 ```dvc
 $ pip install -r src/requirements.txt
+$ git commit
```
shcheklein, line 65: why do we need `git commit` here? It's unusual flow

overall, let's not overcomplicate pls. First priority is still simplicity and ability to get the "idea". It'll always contradict with making it more and more reproducible and correct.

content/docs/start/data-management/data-pipelines.md
```diff
@@ -32,7 +32,7 @@ into a [stage](/doc/command-reference/stage):
 
 ### ⚙️ Expand to download example code.
 
-Get the sample code like this:
+Get the sample code, initialize DVC, and save everything to Git:
```
shcheklein, line 35: again, I'm not sure I like this

content/docs/start/data-management/data-pipelines.md
dberenbaum: > p.s. not sure this is wanted per [#1943 (comment)](https://github.com/iterative/dvc.org/issues/1943#issuecomment-728193024):
> 
> > Doing it completely reproducible and chapters independent makes it a hands-on tutorial like (which get started is not). It means longer read time, more distractions for users who want to get the idea quick, etc.


> folks, I think it might confuse users even more ... also I it won't make it "actionable", you need a lot of steps (dvc add / import at the very least to start). The previous section serves as an actionable item for this.

So it looks like @jorgeorpinel opened the PR to generate this discussion of whether reproducibility is wanted, and @shcheklein your opinion is that it is not wanted, correct?

It makes sense, except having the section to download the code and install requirements doesn't make sense to me in that case. Why not refer back to previous sections and leave it at that?

shcheklein: > of whether reproducibility is wanted, and @shcheklein your opinion is that it is not wanted, correct?

it's needed. My point that we have enough for this (maybe do some additional hidden section with precise `git commit` commands, etc). Making every section independent goes way beyond what we had in mind for the get started and will complicate things.

Also, I don't believe this should be a priority. Even if we do this at best we'll get some small improvement. It's not worth it. We should be doing major updates (whatever it means).

shcheklein: > So it looks like @jorgeorpinel opened the PR to generate this discussion of whether reproducibility is wanted

it was not clear from the description btw, so I was reading this as an improvement/bug fix PR:

Screen Shot 2022-09-27 at 1 08 27 PM

jorgeorpinel: Opened as a quick way to make sure #3998 was closed properly (fully done) as I groomed GS-related tickets. This is definitely not high priority.

Removed `bug` label and updated description. Thanks for the other Feedback as well! Will address when possible

jorgeorpinel (commented):
180° turn: let's just mention the Data Versioning chapter (and corresponding example repo tag) as preparation then?

p.s. maybe the Pipelines chapter should come before Data and Model Access BTW?

content/docs/start/data-management/data-pipelines.md
```diff
@@ -32,6 +32,17 @@ into a [stage](/doc/command-reference/stage):
...
 ### ⚙️ Expand to download example code.
 
+
+<admon type="info">
+
+To actually try this example, begin after following the
+[Data Versioning](/doc/start/data-management/data-versioning) chapter.
+You can get there by cloning the `2-track-data` tag of our
+[example-get-started](https://github.com/iterative/example-get-started)
+repo.
```
jorgeorpinel, line 42: I guess this part is also needed? Please commit if so:
```suggestion
repo and using `dvc pull` to get the DVC-tracked data.
```

dberenbaum:
> Also, I don't believe this should be a priority. Even if we do this at best we'll get some small improvement. It's not worth it. We should be doing major updates (whatever it means).

@jorgeorpinel What do you think about closing this PR? It seems like we went into detailed review without resolving the question of whether it's worth doing at all?

jorgeorpinel:
If you don't approve it feel free to close it to avoid further discussion on a non-priority @dberenbaum . Otherwise, if the change can be valuable, please approve+merge (since the work has been done after all). No worries either way!

dberenbaum:
Thanks @jorgeorpinel! Despite the narrow scope, I still think there's discussion that would be needed here about what should be added and deleted, so going to close to not distract from priorities.

PR #4000, opened 2022-09-27
start: reader instructions (in index pages)

jorgeorpinel:
Per https://github.com/iterative/dvc.org/issues/1943#issuecomment-728193024:

> Change language at the very beginning, may be at the very top of every chapter. Make it super clear "how to use the get started"...
Also, don't forget that we have the whole project prepared - iterative/example-get-started that can be taken, any tag can be git checkout-ed and it's an ideal starting point.

In review app: ⏳ 

---

Jump to https://github.com/iterative/dvc.org/pull/4000#issuecomment-1293021600


content/docs/start/experiment-management/index.md
```diff
@@ -0,0 +1,17 @@
...
+  'Get started with experiment management and visualization in DVC, without
+  bloating Git history.'
+---
+
+# Get Started: Experiment Management
+
+Let's look at DVC's features from the perspective of machine learning experiment
+management. This includes bootstrapping, running and queueing experiments
+(possibly in parallel), as well as visually comparing the results to commit the
+best ones to Git.
```
jorgeorpinel, line 13: Added this intro to the Exp Mgmt chapter since we'd be using the index page after all.

> Related context in https://github.com/iterative/dvc.org/pull/3956#issuecomment-1253067819.

content/docs/start/index.md
```diff
@@ -45,20 +45,22 @@ $ git commit -m "Initialize DVC"
...
 Now you're ready to DVC!
 
-DVC's multiple feature sets are best understood from different angles. Pick a
-trail below to see an overview of all features from that perspective:
+The value of DVC's many feature sets is best understood from different angles.
+Pick a trail below to see an overview of all features from that perspective:
 
 ### Data Management
 
+- [Introduction](/doc/start/data-management)
```
jorgeorpinel, line 53: Added links to the indices as "introduction".

content/docs/start/index.md
```diff
@@ -45,20 +45,22 @@ $ git commit -m "Initialize DVC"
...
   storing large files in the repo (think "Git for data"). Data is stored
   separately, which allows for efficient sharing.
 
-- **[Data and model access]** shows how to use data artifacts from outside of
-  the project and how to import data artifacts from another DVC project. This
-  can help to download a specific version of an ML model to a deployment server
-  or import a model to another project.
+- **[Data and model access]** goes over using data artifacts from outside of the
+  project and importing them from another DVC project. This can help to download
+  a specific version of an ML model to a deployment server or import a model to
```
jorgeorpinel, line 62: Updated some of these descriptions to match the cards at the bottom of the trail index pages.

content/docs/start/index.md
```diff
@@ -11,20 +11,24 @@ running `dvc init` inside a Git project:
...
 $ cd example-get-started
 $ git init
 ```
 
+<admon type="info">
+
+This name is actually used in our
+[example-get-started](https://github.com/iterative/example-get-started) repo.
+
+</admon>
```
jorgeorpinel, line 30: I also realized this is only applicable to the Data Mgmt trail now so I updated the note here (made it more general) and added hidden tips about each example repo into the trail index pages (context: https://github.com/iterative/dvc.org/issues/2496#issuecomment-1255431925).

content/docs/start/data-management/index.md
```diff
@@ -15,3 +15,56 @@ We can also build and version pipelines to capture our data workflows stage by
...
+You'll find useful notes and tips throughout our docs such as this one:
+
+<admon type="tip">
+
+The steps and results of some of these chapters are captured in our
+[example-get-started] repo on GitHub. Feel free to clone it and `git checkout`
+any of it's [tags] to explore further.
+
+[example-get-started]: https://github.com/iterative/example-get-started
+[tags]: https://github.com/iterative/example-get-started/tags
```
jorgeorpinel, line 43: Should we maybe turn the tags that correspond to certain chapters and sections into Releases in the GH repo? I.e. from some https://github.com/iterative/example-get-started/tags to https://github.com/iterative/example-get-started/releases in this case. May be easier to mention releases than all the tags (not so easy to explore and match to the GS).

content/docs/start/data-management/index.md
```diff
@@ -31,7 +31,9 @@ you can click on to expand:
 
 ### Click for an example!
 
-You'll find useful notes and tips throughout our docs such as this one:
+You'll find useful notes and tips throughout our docs such as the onw below.
```
shcheklein, line 34: onw -> one

shcheklein (changes requested):
Sorry for stressing it one more time, but I think we are optimizing things that are not very important.

Re the PR - can we also simplify it? do an a common one paragraph description that is common for all trails? It feels sad that we need to explain in such length. Get started should be self-descriptive, simple, etc.

content/docs/start/data-management/index.md
jorgeorpinel:
```suggestion
You'll find useful notes and tips throughout our docs such as the one below.
```

jorgeorpinel:
> Sorry for stressing it one more time, but I think we are optimizing things that are not very important.

You're right to insist on this @shcheklein. It's not always clear what is most impactful. And some of our plans (at least Q3) imply shuffling existing content around and secondary enhancements like this one.

That said, clarifying the structure and usage of the GS could be pretty impactful at least for current readers, though it probably doesn't directly help take us to the next level. But again, I consider that I'm following our plans here.

> sad that we need to explain in such length. Get started should be self-descriptive

Good points. I thought about this and didn't think that having relatively long index pages was much of a problem but we could def. take the opposite approach: shorten and centralize instructions in the general GS index page (also then we probably don't need secondary index pages at all) ⏳

shcheklein:
>  And some of our plans (at least Q3) imply shuffling existing content around and secondary enhancements like this one.

Yes, but one thing is shuffle UG, another one is GS. GS existing content is already in a decent shape. To get it to the next level we need some fundamental changes- new projects, new trails, going back from trails to something simpler, redesigning completely the main entry page, etc, etc. That level of things can move the needle.

jorgeorpinel:
Many of the updates became were irrelevant after other GS changes. This is now boiled down to a single file (not sure it completely addresses the OP though).

content/docs/start/index.md
```diff
@@ -12,20 +12,24 @@ running `dvc init` inside a Git project:
...
 
 ```dvc
 $ mkdir example-get-started
 $ cd example-get-started
 $ git init
 ```
 
+<admon type="info">
+
+This name is actually used in our
```
jorgeorpinel, line 28:
```suggestion
This directory name is actually used in our
```

content/docs/start/index.md
```diff
@@ -74,15 +78,54 @@ trail below to see an overview of all features from that perspective:
...
 
 - **[Visualization]** helps you compare experiment results visually, track your
   plots, and generate them with library integrations.
 
 [experiments]: /doc/start/experiment-management/experiments
 [visualization]: /doc/start/experiment-management/visualization
+
+<admon type="tip">
+
+The are captured in our [example-dvc-experiments] repo (see its [tags]).
```
jorgeorpinel, line 106:
```suggestion
These are captured in our [example-dvc-experiments] repo (see its [tags]).
```

content/docs/start/index.md
```diff
@@ -74,15 +78,54 @@ trail below to see an overview of all features from that perspective:
...
+
+## Following the Get Started
+
+Although presented in a logical order, each page in the trails above is an
+independent chapter you can go to directly. There may be references to other
+chapters or docs, but you can skip them and still get a good idea of the
+features in question, even by only reading page sections.
+
+For better learning, you can also try each step yourself and get to similar
+results. Some of the required steps for this may be inside collapsed sections
```
jorgeorpinel, line 121:
```suggestion
Each page in the trails above is more or less independent,
especially if you're only reading them to get a general idea of the
features in question. For better learning, try each step yourself from the beginning of any trail.
Some of the preparation steps may be inside collapsed sections
```

yathomasi:
I had to close the issue as something strange had been happening. The deployments were retriggered even though the deployment failed and no changes were pushed. I have been noticing the alert and directly checked the log on heroku to see the issue failure issue was due to `Github Token` issue. I didn't bother to check the PR as that would have been solved by update `main`. 
I just noticed that no update has been pushed but there are multiple failed attempts to build the app. I merged with the main and now although the build was successful, new build is triggered again. Also, Gatsby Cloud seems to be applied here. I guess it's picked up from the previous setup.

shcheklein (approved):

shcheklein:
Btw, I think `Pick a trail below to see an overview of all features from that perspective:` is what confuses people. Most likely they think that each link below on the page is an independent trail. I would suggest a minor change - Data Management -> Data Management Trail on the page (same for experiments) + we can make it clear in the links - `in the part 1 of the trail ....` ...

jorgeorpinel:
Good observation @shcheklein . PTAL at https://github.com/iterative/dvc.org/pull/4197.

Issue #2883, opened 2021-10-01
guide: Modeling Pipelines


jorgeorpinel:
> From https://github.com/iterative/dvc.org/issues/144#issuecomment-844711839

- [x] What are data pipelines, including DAG concept (extract from https://dvc.org/doc/command-reference/dag)

Structure:

> See #2857 for the original structure ideas.

- [x] [Defining Pipelines](https://dvc.org/doc/user-guide/pipelines/defining-pipelines)
  - [x] Focus on the pipelining process and ~~codification~~ link to dvc.yaml ref. for formalities
  - [x] ~~Defining stages (extract from `run` and `stage` refs.)~~
    - [ ] Dependencies and (general) parameters
    - [ ] Outputs and metrics/plots
  - [ ] Should provide some actual `dvc.yaml` example/templates to copy&paste.
  - [ ] Multiple dvc.yaml files vs. multiple pipelines (`repro --recursive`)
  - [ ] Describing formally outs, deps, stage doesn't make sense here
  - Mention `exp init`? (Link to appropriate page/section) [Maybe](https://github.com/iterative/dvc.org/issues/2883#issuecomment-1240994630)
- [ ] Pipeline reproduction
  - [ ] Extract some details from https://dvc.org/doc/command-reference/repro ?
- [ ] Experimental pipelines ([discussion](https://github.com/iterative/dvc.org/issues/2883#issuecomment-1244806361))
- [ ] Operationalizing pipelines?

Other tasks

- [ ] https://github.com/iterative/dvc.org/issues/4267
- [ ] Mention VS Code "as an editor that supports schema definition, etc."
- [ ] Update some of the [stage] links around this term that currently go to `run/stage` refs (some should link to the concept page instead)
- [ ] Recover some info deleted in [this change to Running Experiments](https://github.com/iterative/dvc.org/pull/3182/files#diff-213c105f41cfb7e3abe29cba4fd80d9cf459bb94703b0a49708b23cf6b6cf5c5) ?
- [ ] https://github.com/iterative/dvc.org/issues/3670

jorgeorpinel:
Updated [plan](https://github.com/iterative/dvc.org/issues/2883#issue-1013656881) for this epic per https://github.com/iterative/dvc.org/pull/3414#issuecomment-1238505396. PTAL @dberenbaum @shcheklein

dberenbaum:
Thanks @jorgeorpinel!

[...]

PR #4011, opened 2022-09-30
guide: reorg sidebar

dberenbaum:
Proposed restructure of the guide based on discussion at the bottom of #144. I only updated the sidebar so we can iterate on that before propagating the changes elsewhere.


content/docs/sidebar.json
```diff
@@ -119,6 +119,19 @@
           }
         ]
       },
+      {
+        "label": "Data Management",
+        "slug": "data-management",
+        "children": [
+          "large-dataset-optimization",
+          "defining-pipelines",
```
shcheklein, line 127: The only thing that raises some concern from my end. It should not be in the data management to my mind. I think pipelines deserve their own place in the UG. It's just content should be very different as we were discussing in the PR for the pipelines.

shcheklein (approved):
Good stuff, I would reconsider the pipelines part. Otherwise LGTM (maybe we need some redirects)

content/docs/sidebar.json
dberenbaum: I struggled to keep them separate because:

1. I wasn't sure where to put `external-dependencies` (and `managing-external-data`/external outputs to a lesser extent) since they are about external data but only really make sense in the context of pipelines.
2. Pipelines in general is tightly coupled with data management now (I would like to address this on the product side, but it's not going to happen this year).
3. From a use case perspective, I can see an argument for keeping them together (a typical data engineering workflow needs data versioning and pipelines).

Still, I'm happy to keep pipelines separate and agree it seems like a more natural structure (see my comments at the bottom of #144). Where would you suggest to put `external-dependencies` and `managing-external-data`?

shcheklein: > since they are about external data but only really make sense in the context of pipelines.

Hmm, it think the both could be used independently though? Especially external outputs (like versioning an external directory).

Even if that is the case, then we should move them into pipelines, out of the data management.

dberenbaum: Ah, I didn't catch your point about imports before. In that case, we can duplicate both sections:

```
      {
        "label": "Data Management",
        "slug": "data-management",
        "children": [
            "large-dataset-optimization",
            {
              "label": "Importing External Data",
              "source": "/docs/user-guide/pipelines/external-dependencies"
            },
            {
            }
        ]
      },
      {
        "label": "Pipelines",
        "slug": "pipelines",
        "children": [
            "defining-pipelines",
            "external-dependencies",
            {
              "label": "External Outputs",
              "source": "/docs/user-guide/data-management/managing-external-data"
            }
        ]
      },
```

shcheklein: yep, don't know if that is possible though :) (would be a nice feature it is possible)

redirects-list.json
```diff
@@ -73,6 +73,15 @@
...
   "^/doc/user-guide/dvcignore$                                                            /doc/user-guide/project-structure/dvcignore-files",
   "^/doc/user-guide/contributing(/.*)?$                                                   /doc/contributing$1",
+  "^/doc/user-guide/contributing(/.*)?$                                                   /doc/contributing$1",
+  "^/doc/user-guide/large-dataset-optimization$                                           /doc/user-guide/data-management/large-dataset-optimization",
+  "^/doc/user-guide/external-dependencies$                                                /doc/user-guide/data-management/importing-external-data",
+  "^/doc/user-guide/pipelines/external-dependencies$                                      /doc/user-guide/data-management/importing-external-data",
+  "^/doc/user-guide/managing-external-data$                                               /doc/user-guide/data-management/managing-external-data",
+  "^/doc/user-guide/pipelines/external-outputs$                                           /doc/user-guide/data-management/managing-external-data",
```
jorgeorpinel, line 81: These redirects are a little weird. Even if we want to duplicate page entries like that, the navigation via redirect can be confusing (it moves you unexpectedly in the nav).

> they are about external data but only really make sense in the context of pipelines

I'd just put them under pipelines for now and link from certain data mgmt pages as appropriate.

content/docs/sidebar.json
jorgeorpinel: I don't think it's ideal to have page duplicates like that. I'd consider it a patch for either a product or docs content issue that should be resolved.

content/docs/sidebar.json
```diff
@@ -128,36 +146,26 @@
...
         "slug": "how-to",
         "source": false,
         "children": [
+          "running-dvc-on-windows",
+          "setup-google-drive-remote",
           "stop-tracking-data",
           "update-tracked-data",
           "add-deps-or-outs-to-a-stage",
           "resolve-merge-conflicts",
           "share-a-dvc-cache"
```
jorgeorpinel, line 166: Minor (could be a follow-up) in terms of order and page naming:

```suggestion
          "run-dvc-on-windows",
          "stop-tracking-data",
          "update-tracked-data",
          "resolve-merge-conflicts",
          "add-deps-or-outs-to-a-stage",
          "setup-google-drive-remote",
          "share-a-dvc-cache"
```

content/docs/sidebar.json
```diff
@@ -128,36 +146,26 @@
           "running-experiments",
           "hydra-composition",
           "comparing-experiments",
+          "visualizing-plots",
```
jorgeorpinel, line 149: > Possibly follow-up

Should the [Parallel Coordinates Plot](https://dvc.org/doc/user-guide/experiment-management/comparing-experiments#parallel-coordinates-plot) section be moved or linked from the [Visualizing Plots](https://dvc-org-ug-reorg-zcajwdgw5kgni.herokuapp.com/doc/user-guide/experiment-management/visualizing-plots) page?

content/docs/user-guide/pipelines/index.md
```diff
@@ -1,19 +0,0 @@
-# Pipelines
-
-If you find yourself repeating sequence of actions to get or update the results
-of your project, then you may already have a pipeline. For example, a data
-science workflow could involve:
-
-1. Gathering data for training and validation
-2. Extracting useful features from the training dataset
-3. (Re)training an ML model
-4. Evaluating the results against the validation set
```
jorgeorpinel, line 10: Up to you but [this](https://dvc.org/doc/user-guide/pipelines) seemed like a nice intro. Keep in mind we plan to add more pages to this section (see https://github.com/iterative/dvc.org/issues/2883).

redirects-list.json
```diff
@@ -73,6 +73,15 @@
   "^/doc/user-guide/dvc-internals(/.*)?$                                                  /doc/user-guide/project-structure/internal-files$1",
   "^/doc/user-guide/dvcignore$                                                            /doc/user-guide/project-structure/dvcignore-files",
   "^/doc/user-guide/contributing(/.*)?$                                                   /doc/contributing$1",
+  "^/doc/user-guide/contributing(/.*)?$                                                   /doc/contributing$1",
```
jorgeorpinel, line 76: Repeated redirect
```suggestion
  "^/doc/user-guide/contributing(/.*)?$                                                   /doc/contributing$1",
```

redirects-list.json
jorgeorpinel: Let's at least make them 302s for now?
```suggestion
  "^/doc/user-guide/external-dependencies$                                                /doc/user-guide/data-management/importing-external-data",
  "^/doc/user-guide/pipelines/external-dependencies$                                      /doc/user-guide/data-management/importing-external-data 302",
  "^/doc/user-guide/managing-external-data$                                               /doc/user-guide/data-management/managing-external-data",
  "^/doc/user-guide/pipelines/external-outputs$                                           /doc/user-guide/data-management/managing-external-data 302",
```

redirects-list.json
```diff
@@ -73,6 +73,14 @@
...
+  "^/doc/user-guide/external-dependencies$                                                /doc/user-guide/data-management/importing-external-data",
+  "^/doc/user-guide/pipelines/external-dependencies$                                      /doc/user-guide/data-management/importing-external-data",
+  "^/doc/user-guide/managing-external-data$                                               /doc/user-guide/data-management/managing-external-data",
+  "^/doc/user-guide/pipelines/external-outputs$                                           /doc/user-guide/data-management/managing-external-data",
+  "^/doc/user-guide/running-dvc-on-windows$                                               /doc/user-guide/how-to/running-dvc-on-windows",
+  "^/doc/user-guide/setup-google-drive-remote$                                            /doc/user-guide/how-to/setup-google-drive-remote",
+  "^/doc/user-guide/visualizing-plots$                                                    /doc/user-guide/experiment-management/visualizing-plots",
   "^/doc/understanding-dvc(/.*)?$                                                         /doc/user-guide/what-is-dvc",
```
jorgeorpinel, line 84: Unrelated but we can update this now
```suggestion
  "^/doc/understanding-dvc(/.*)?$                                                         /doc/user-guide/overview",
```

jorgeorpinel: Or maybe even remove it altogether (it's been there for a long time)

```suggestion
```

jorgeorpinel (approved):
A bunch of link updates should follow this. Fun fun... Lmk if you need help!

jorgeorpinel:
I think this would basically close #144 ?

content/docs/sidebar.json
```diff
@@ -119,6 +119,24 @@
           }
         ]
       },
+      {
+        "slug": "data-management",
+        "source": false,
+        "children": [
+          "large-dataset-optimization",
+          "importing-external-data",
+          "managing-external-data"
```
jorgeorpinel, line 128: An alternative to consider (from https://github.com/iterative/dvc.org/issues/144#issuecomment-1206648437) which may be better since we don't love the product feature as-is now (I think):

> External Deps/Managing External Data: ... or make them only reachable via links.

jorgeorpinel: p.s. update the H1 in `importing-external-data.md`

content/docs/user-guide/pipelines/index.md
dberenbaum: Sorry, this was a mistake from when I initially got rid of the pipelines guide. Adding back ⏳

dberenbaum:
> A bunch of link updates should follow this. Fun fun... Lmk if you need help!

I guess it's not that bad, but it seems like the redirects should make it a pretty low priority, right?

An auto update script when adding redirects would be 🔥 .

content/docs/sidebar.json
dberenbaum: > or make them only reachable via links.

I like it better than having no solution for using data directly from the cloud, so I'm going to keep it in.

content/docs/sidebar.json
dberenbaum: TBH I think they're fine as is. For example, I intentionally put windows and google drive at the top since they are blockers for doing almost anything for those users.

content/docs/sidebar.json
dberenbaum: I don't think it should be moved since it's more related to comparing experiments, but linking could make sense.

content/docs/sidebar.json
jorgeorpinel: OK but (minor)
- should it be (how to) `run-dvc-on-windows` (not running)
- the one for GDrive may be moved in the future again right? Into data management/ remote setup (see https://github.com/iterative/dvc.org/issues/2866)

redirects-list.json
dberenbaum: > I'd just put them under pipelines for now and link from certain data mgmt pages as appropriate.

I think I'd rather leave them in data management and remove from pipelines here -- we already link to external dependencies in https://dvc.org/doc/user-guide/pipelines/defining-pipelines#simple-dependencies.

dberenbaum:
Good feedback @jorgeorpinel and @shcheklein! If one of you want to do one more pass, I'll merge if everything looks good.

jorgeorpinel:
LGTM. I started another PR for the links: https://github.com/iterative/dvc.org/pull/4015

redirects-list.json
jorgeorpinel: Do use 302s for now though, please!

dberenbaum: Done, thanks!

PR #4026, opened 2022-10-07
ref: simplifications around "pipelines"/"DAG" concepts

jorgeorpinel:


shcheklein:
@jorgeorpinel please resolve the conflicts, please add some summary on what is going on here and where should we pay attention to, thanks!

jorgeorpinel:
@shcheklein I think this was meant to optimize the overlap between the 3 related cmd refs mentioned in the title.

Due to the messy diff in this one, it may be easier to compare the existing [cmd ref pages](https://dvc.org/doc/command-reference) vs. the new ones (listed [in the OP](https://github.com/iterative/dvc.org/pull/4026#issue-1400581950)).

This is p2 though, so I'm finishing other things first. We can always close this PR (temporarily or permanently) or ask the eng team to review it when they can (since it's purely changes to the cmd ref).

jorgeorpinel:
Update: now that the `run` ref is gone, this should be easier to review 🙂 

Will leave it up to your team @dberenbaum (sorry to keep pinging you from multiple places).

content/docs/command-reference/stage/add.md
```diff
@@ -37,82 +37,72 @@ are ignored by `dvc stage add`.
...
-  (outside of the <abbr>workspace</abbr>) are also supported (except metrics and
-  plots).
+- [external dependencies] and [external outputs] (outside of the
+  <abbr>workspace</abbr>) are also supported (except metrics and plots).
 
-- Outputs are deleted from the workspace before executing the command (including
-  at `dvc repro`) if their paths are found as existing files/directories (unless
-  `--outs-persist` is used). This also means that the stage command needs to
-  recreate any directory structures defined as outputs every time its executed
-  by DVC.
```
dberenbaum, line 103: @jorgeorpinel Do you mind explaining why you dropped this note?

content/docs/command-reference/stage/add.md
```diff
@@ -37,82 +37,72 @@ are ignored by `dvc stage add`.
...
+- Stage commands need to recreate any directory structures defined as outputs
+  every time its executed by DVC.
```
dberenbaum, line 90:
```suggestion
- Outputs are deleted from the workspace before executing the command (unless
  `--outs-persist` is used). Stage commands need to recreate any directory structures defined as outputs
  every time its executed by DVC.
```

@jorgeorpinel I will probably go with something like this unless there's a strong reason to drop the first sentence completely.

content/docs/command-reference/stage/add.md
jorgeorpinel: Yeah I was just making it shorter I guess. Idk this was a long time ago. I've reinstated the first part so it's explicit again. PTAL

content/docs/command-reference/stage/add.md
jorgeorpinel: Oh, I already rephrased in a similar way. Sure up to you 🙂

PR #4028, opened 2022-10-07
Bump theme to v0.1.17

rogermparent:
Just speeding up what Renovate otherwise would do.


yathomasi (approved):
Looks like renovate configuration is not working. I hope [PR](https://github.com/iterative/dvc.org/pull/4030) will fix that.

PR #4033, opened 2022-10-11
Add conditional for running link check deploy in consumer repo

rogermparent:


.github/workflows/link-check-deploy.yml
```diff
@@ -3,4 +3,7 @@ on:
   deployment_status:
 jobs:
   run:
+    if:
+      github.event.deployment.ref != ${{ github.event.repository.default_branch }} &&
+      github.event.deployment_status.state == 'success'
```
rogermparent, line 8: Same conditional from inside the reusable action

yathomasi (approved):
It would be better if these would work from the reusable workflow themselves.
 
Since the  `deployment` or `deployment_status` event occurs in the current workflow there is a `wrokflow_call` event on reusable action, which could be the reason we may not access those values. And, if we want to pass these, I think we would need to pass them as inputs. 

Anyway, this seems to be the way to go, and looking into [payload data](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-example-10) conditional looks good to me.

PR #3789, opened 2022-07-21
guide: updates to `dvc.yaml` spec

jorgeorpinel:
Extracted from https://github.com/iterative/dvc.org/pull/3414#issuecomment-1189898267

This page should focus on the schema/spec of the file and it's sections more than guidance and explaining concepts, since we'll have a guide for that now (see https://github.com/iterative/dvc.org/pull/3414).

PR #4039, opened 2022-10-12
hotfix: dvc.yaml spec format

jorgeorpinel:
The format under https://dvc.org/doc/user-guide/project-structure/dvcyaml-files#metrics-and-plots-outputs is currently all broken due to an open `<admon>` tag:

image

### DIFF
--- a/content/docs/user-guide/project-structure/dvcyaml-files.md
+++ b/content/docs/user-guide/project-structure/dvcyaml-files.md
@@ -201,7 +201,7 @@ stages:
 `cache: false` is typical here, since they're small enough for Git to store
 directly.
 
-<admon>
+</admon>
 
 The commands in `dvc metrics` and `dvc plots` help you display and compare
 metrics and plots.


shcheklein (approved):

jorgeorpinel:
OK that seems to fix it (checked https://dvc-org-jorgeorpinel-pa-mn3m0z.herokuapp.com/doc/user-guide/project-structure/dvcyaml-files#foreach-stages). Merging

jorgeorpinel:
Do we need a post-mortem? I assume we forgot to sanity check some recent PR.

dberenbaum:
Good catch! Looks like it was introduced in https://github.com/iterative/dvc.org/pull/3789.

jorgeorpinel:
Yeah I imagined it was my bad 😅 we should just remember to sanity check review apps... I almost always do it but forgot this time. Or we could consider checks for open tags or other XML validation...

Issue #3896, opened 2022-08-25
ref: formalize `dvc.yaml` page + better structure


jorgeorpinel:
In general, https://dvc.org/doc/user-guide/project-structure/dvcyaml-files is turning into a reference closer to the cmd ref than guides, esp. since there are actual guides for certain topics in there (like pipelining or plots). But the content is still mixed with some guide-like sections and other more formal spec/definition sections. Should be closer to the latter.

> Maybe also move it into /doc/files-reference along with a few other pages?

Some specific ideas from the OP:

---

...we should be describing things in their order. In every section we can put a small snippet again to get some context back what part of dvc.yaml we are talking about. 

Section names could be a bit more formal e.g. "`cmd`: stage command"

_Originally posted by @shcheklein in https://github.com/iterative/dvc.org/pull/3414#discussion_r954298890_

PR #4043, opened 2022-10-13
term: metafile vs DVC file

jorgeorpinel:
👀 👇🏼

### DIFF
--- a/content/docs/user-guide/project-structure/dvcyaml-files.md
+++ b/content/docs/user-guide/project-structure/dvcyaml-files.md
@@ -10,7 +10,7 @@ schema explained below. We encourage you to get familiar with it so you may
 modify, write, or generate them by your own means.
 
 `dvc.yaml` files are designed to be small enough so you can easily version them
-with Git along with other DVC <abbr>metafiles</abbr> and your project's code.
+with Git along with other <abbr>DVC files</abbr> and your project's code.
 
 ## Stages
 


content/docs/user-guide/project-structure/dvcyaml-files.md
```diff
@@ -10,7 +10,7 @@ schema explained below. We encourage you to get familiar with it so you may
 modify, write, or generate them by your own means.
 
 `dvc.yaml` files are designed to be small enough so you can easily version them
-with Git along with other DVC <abbr>metafiles</abbr> and your project's code.
+with Git along with other <abbr>DVC files</abbr> and your project's code.
```
jorgeorpinel, line 13: Changing this because "metafile" is not among the official basic concepts (see the [glossary](https://dvc.org/doc/user-guide/glossary)). The term is still "DVC file".

jorgeorpinel:
But most importantly, this PR could serve as a discussion on which term to use for this concept, because we currently have a mix of "dvc file(s)" (44 instances with 14 being `<abbr>` tooltips i.e. refer to the glossary) and "metafile" (24 instances throughout docs).

My plan for some time was to migrate to the latter, but per recent discussions (in the core repo, can't find the ticket but it involved @skshetry @shcheklein) it seems that we may not prefer that one either.

The problem with "DVC file" is that is may sound/look the same as .dvc file, which is only one of the file types we support. Possibly rel. https://github.com/iterative/dvc.org/issues/3896:

> Maybe also move it into /doc/files-reference along with a few other pages?

Cc @dberenbaum

content/docs/user-guide/project-structure/dvcyaml-files.md
jorgeorpinel: > "metafile" is not among the official basic concepts

Should it though?

shcheklein (approved):

